PR #7336, opened 2021-01-25
M3 4781 firewall rule clone

Jskobos:
## Description

Add "Clone" to the Firewall rule action menu; directly clones a rule and adds it to the queued changes.

Holding off on the open-the-drawer piece, still not sure how to handle that part but this works and we might as well get it in and get some feedback on it.


johnwcallahan (approved):

DevDW (approved):
LGTM 🚢 

I don't think the drawer needs to open initially; seems like the current flow is fine, but we'll see what the feedback on it is

PR #7340, opened 2021-01-26
M3-4800: Fix scheduled migration banner font color on dark mode

tiffwong:
This is what it looked like before:
Screen Shot 2021-01-26 at 12 04 11 PM

[http://localhost:3000/linodes/0](http://localhost:3000/linodes/0) has a mock migration banner for ease of testing


Jskobos (commented):
This works, but this component is old and out of date. I think it would be more semantic/up to spec if we:

- Wrap the messaging in a Typography (has the same effect as the class you added)
- Replace the existing Typography with a button and use ...theme.applyLinkStyles.
- Change createStyles to makeStyles
- Remove the `compose` (won't need it after `makeStyles` and then use `useSnackbar`.

johnwcallahan (approved):

Jskobos (approved):

PR #7342, opened 2021-01-26
M3-4850: Backups CTA Adjustments

DevDW:
## Description
Remove the "Enable Backups" button in banner and use user preferences instead of local storage to dismiss banner. And CMR cleanup -- `BackupsCTA_CMR.tsx` --> `BackupsCTA.tsx`

## Type of Change


Jskobos (approved):

packages/manager/src/features/Backups/BackupsCTA.tsx
```diff
@@ -12,56 +12,46 @@ import {
...
 
-type ClassNames = 'root' | 'container' | 'buttonsContainer' | 'dismiss';
-
-interface Props {
-  dismissed: () => void;
-}
+type ClassNames = 'root' | 'enableButton' | 'enableText' | 'closeIcon';
 
 const styles = (theme: Theme) =>
   createStyles({
```
tiffwong, line 24: You can use `makeStyles`/`useStyles` here instead

tiffwong (approved):
Other than the one cleanup item I mentioned, everything else looks good!

johnwcallahan (changes requested):
Overall this looks much cleaner, although it doesn't handle the case when the preference isn't found in either user preferences or local storage (e.g. new users). In that case I'm unable to see the CTA. The PreferenceToggle defaults to the first option you give it if it's not specified in preferences/localStorage, so in this case `dismissed` is defaulted to `true`. 

Also the text and X button should be aligned with each other (centered in the panel). 

Screen Shot 2021-01-28 at 12 34 11 PM

johnwcallahan (commented):
Thanks for the user preferences change, that's working really well.

The text still looks a bit low to me, though I'm not sure what the problem is. Maybe @tiffwong could weigh in here.
Screen Shot 2021-01-29 at 9 54 21 AM

tiffwong:
@johnwcallahan I checked and it is aligned so maybe its an optical illusion?

johnwcallahan:
@tiffwong maybe something with line-height? The text sits lower than the "X" button, which seems perfectly aligned.
Screen Shot 2021-02-02 at 5 27 25 PM

tiffwong:
@johnwcallahan I removed some of the text to compare the "X" closer to full height letters like "k" and "l" and it looks even to me

Screen Shot 2021-02-02 at 5 46 26 PM
Screen Shot 2021-02-02 at 5 44 47 PM

PR #7343, opened 2021-01-27
M3-4879 Fix: "Compatible Images" field on StackScript detail

Jskobos:
We weren't requesting images from StackScript detail view,
which meant that if you navigated directly there the "Compatible Images"
section would be blank.

There was some defaulting "No compatible Images found" text meant to be
displayed in that situation, but that wasn't working either.

- Converted StackScript.tsx to a function component
- Used useImages and useReduxLoad to get the Images data
- Fixed the empty state
- Updated and refactored the tests that this broke

I considered using React-Query, but images are a harder case for
this since withImages is still used in some class components. Going
to punt that until later so we don't have 2 parallel data channels
for them.

### DIFF
--- a/packages/manager/src/components/StackScript/StackScript.test.tsx
+++ b/packages/manager/src/components/StackScript/StackScript.test.tsx
@@ -1,49 +1,16 @@
-import { shallow } from 'enzyme';
+import { screen } from '@testing-library/react';
 import * as React from 'react';
-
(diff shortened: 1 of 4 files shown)


ndukatz (approved):
tested coming in directly and through regular nav. getting images each time. compatible images are being displayed each time.

tiffwong (approved):
💯

PR #7344, opened 2021-01-27
M3-4778: Add "Edit" button on StackScript Detail page

DevDW:
## Description
Add "Edit" button on StackScript Detail page; minor updates to associated test & a function name change in `StackScripts.tsx`.

## Type of Change

## Note to Reviewers
Open to other ideas for the placement of the "Edit" button, but the upper right is consistent with what we have for Billing Contact, for example

To-Do:
- [x] Check on Community StackScripts having "Edit" button
- [x] Use hooks where possible in `StackScript.tsx`
- [x] More detailed checks to display "Edit" button; add `account/users` to React Query

### DIFF
--- a/packages/manager/src/components/StackScript/StackScript.tsx
+++ b/packages/manager/src/components/StackScript/StackScript.tsx
@@ -1,19 +1,26 @@
+import { Grant } from '@linode/api-v4/lib/account';
 import { StackScript } from '@linode/api-v4/lib/stackscripts';
 import { stringify } from 'qs';
 import { pathOr } from 'ramda';
 import * as React from 'react';
-import { Link } from 'react-router-dom';
+import { connect } from 'react-redux';
+import { Link, RouteComponentProps, withRouter } from 'react-router-dom';
+import { compose } from 'recompose';
+import Button from 'src/components/Button';
 import CopyTooltip from 'src/components/CopyTooltip';
 import Chip from 'src/components/core/Chip';
 import Divider from 'src/components/core/Divider';
-import { Theme, makeStyles } from 'src/components/core/styles';
+import { makeStyles, Theme } from 'src/components/core/styles';
 import Typography from 'src/components/core/Typography';
 import DateTimeDisplay from 'src/components/DateTimeDisplay';
 import Grid from 'src/components/Grid';
 import H1Header from 'src/components/H1Header';
 import ScriptCode from 'src/components/ScriptCode';
+import { isRestrictedUser as _isRestrictedUser } from 'src/features/Profile/permissionsHelpers';
+import { canUserModifyAccountStackScript } from 'src/features/StackScripts/stackScriptUtils';
 import { useImages } from 'src/hooks/useImages';
 import { useReduxLoad } from 'src/hooks/useReduxLoad';
+import { MapState } from 'src/store/types';
 
 const useStyles = makeStyles((theme: Theme) => ({
   root: {
@@ -21,6 +28,12 @@ const useStyles = makeStyles((theme: Theme) => ({
     '.detailsWrapper &': {
       padding: theme.spacing(4)
     }
+  },
+  headerLabel: {
+    marginLeft: '0.25em'
+  },
+  editBtn: {
+    minWidth: 'fit-content'
   },
   deployments: {
     marginTop: theme.spacing(1)
@@ -75,8 +88,13 @@ export interface Props {
   data: StackScript;
 }
 
-export const SStackScript: React.FC<Props> = props => {
+type CombinedProps = Props & StateProps & RouteComponentProps<{}>;
+
+export const SStackScript: React.FC<CombinedProps> = props => {
   const {
+    history,
+    stackScriptGrants,
+    isRestrictedUser,
     data: {
       username,
       deployments_total,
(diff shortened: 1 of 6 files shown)


packages/manager/src/components/StackScript/StackScript.tsx
```diff
@@ -42,6 +51,14 @@ const styles = (theme: Theme) =>
         padding: theme.spacing(4)
       }
     },
+    header: {
+      display: 'flex',
+      flexBasis: '100%',
+      width: '100%',
+      '& > button': {
+        justifyContent: 'flex-end'
```
tiffwong, line 59: Instead of targeting the button here, it's better to apply a class directly on the button

Also instead of `justifyContent: 'flex-end'`, use `minWidth: 'fit-content'` instead. Otherwise, the button will look off center when click and hold the button

tiffwong (changes requested):
I left a few cleanup items but I have a few more questions. 

Is the "Edit" button supposed to show up for Community StackScripts?

Additionally, the name of the StackScript is no longer aligned with the other content:
Screen Shot 2021-01-27 at 5 24 54 PM

DevDW:
> Is the "Edit" button supposed to show up for Community StackScripts?

My initial thought is that it shouldn't, but I'll get confirmation on that

Jskobos:
> > Is the "Edit" button supposed to show up for Community StackScripts?
> 
> My initial thought is that it shouldn't, but I'll get confirmation on that

It definitely should not. Only show Edit if the current user has the ability to edit the StackScript. This might require checking grants for Account scripts, since there are per-user grants for each script.

packages/manager/src/components/StackScript/StackScript.tsx
```diff
@@ -119,7 +154,32 @@ export const SStackScript: React.FC<Props> = props => {
...
   return (
     <div className={classes.root}>
-      <H1Header title={label} data-qa-stack-title={label} />
+      <Grid container justify="space-between">
+        <H1Header
+          className={classes.headerLabel}
+          title={label}
+          data-qa-stack-title={label}
+        />
+        {!is_public ||
```
Jskobos, line 163: This doesn't quite work, since you can publish your own stackscripts and subsequently edit them.

Jskobos: You may have to pass down a `type='community'` or `type='account'` to this component, I think Tiffany is running into something very similar so maybe discuss with her.

Jskobos (changes requested):

DevDW:
@Jskobos I made an attempt at passing it down via props from `StackScripts.tsx` --> `StackScriptsDetail.tsx` --> `StackScript.tsx`, but it wasn't quite fruitful. I commented out some of the code associated with that attempt so you can get an idea of what I was trying. 

On the other hand, in `StackScriptRow.tsx`, I persisted the `category` on the `History` object, and was able to use that in `StackScript.tsx`. With that approach it seems to work if a user navigates to a StackScript from the Account or Community tabs, but it wouldn't work if a user went to a direct URL.

packages/manager/src/queries/accountUsers.ts
```diff
@@ -0,0 +1,12 @@
+import { User, getUsers } from '@linode/api-v4/lib/account';
```
Jskobos, line 1: Query should live in src/queries/users.ts

packages/manager/src/components/StackScript/StackScript.tsx
```diff
@@ -126,6 +124,22 @@ export const SStackScript: React.FC<CombinedProps> = props => {
     }
   );
 
+  let unrestrictedUserHasAccess;
+  if (!isRestrictedUser && accountUsersData) {
```
Jskobos, line 128: Some of this is duplicative, there's already a canUserModifyAccountStackScript function (imported in this file ^^^ ) that will help with these calculations. Probably something like this will work:

```
const showTheButton = isRestrictedUser ? canModifyAccountStackScript(...) : newFunctionCalledIsAccountStackScript(users, stackscript.username)

```

Jskobos: From StackScript detail, I don't think a restricted user can possibly know if he can't edit it because he has read_only access or because it's from someone else's account (I mean they can figure this out but there's nothing in the response or elsewhere in the API that will help here)

Jskobos: ...so probably ix-nay on the disabled button

packages/manager/src/components/StackScript/StackScript.tsx
```diff
@@ -89,9 +106,40 @@ export const SStackScript: React.FC<Props> = props => {
   } = props;
 
   const classes = useStyles();
+  const history = useHistory();
+  const { data: accountUsersData } = useAccountUsers();
+
   const { images: imagesData } = useImages('public');
   useReduxLoad(['images']);
 
+  const { isRestrictedUser, stackScriptGrants } = useSelector(
```
johnwcallahan, line 115: Could the `useAccountManagement` hook be used here instead of this custom selector?

Jskobos (approved):
looks good 👍

tiffwong (approved):

PR #7345, opened 2021-01-27
M3-4880 Fix: Disk selection when opening Rebuild from /linodes

Jskobos:
This was a hooks dependency issue. The dialog state was being
reset when the dialog opened, as intended, but since only [open]
was included in the deps array, stale values of the device map
were used. In addition, when the dialog was opened without /disks
having been previously requested, the completed request wouldn't trigger
the effect.

Used the usePrevious hook to run the reset logic whenever `open`
changes _and_ the calculated deviceMap is different from the previous
value. This involves a deep object comparison on every render while
the dialog is open, but it seems ok in practice.

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDialog.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDialog.tsx
@@ -1,7 +1,7 @@
 import { rescueLinode } from '@linode/api-v4/lib/linodes';
 import { APIError } from '@linode/api-v4/lib/types';
 import { withSnackbar, WithSnackbarProps } from 'notistack';
-import { assoc, clamp, pathOr } from 'ramda';
+import { assoc, clamp, equals, pathOr } from 'ramda';
 import * as React from 'react';
 import { connect } from 'react-redux';
 import { RouteComponentProps } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


johnwcallahan (changes requested):
I haven't fully comprehended the problem/solution here, but I believe I'm still seeing the bug by opening my Linode's Details page from a refresh and then selecting "Rescue".

Jskobos:
> I haven't fully comprehended the problem/solution here, but I believe I'm still seeing the bug by opening my Linode's Details page from a refresh and then selecting "Rescue".

I can't reproduce the bug anywhere on Linode detail...even on develop, so I thought it was just a context bug. Can you please screenshot what you're seeing? Might be something related that needs to be fixed

johnwcallahan:
@Jskobos on this branch, this is what I see when from the Linode Landing page I click the hamburger -> Rescue:

Screen Shot 2021-01-28 at 9 55 15 AM

It's most of the assignments, but not all.

If I go to that Linode's Details page, refresh the page, click the hamburger -> Rescue, I only see two of the assignments: 

Screen Shot 2021-01-28 at 9 55 28 AM

johnwcallahan (approved):
I guess a side effect of using placeholderData is that there's a double /regions request until we move all the region-dependent stuff to RQ, which seems acceptable to me (it's disk cached).

johnwcallahan:
Left this review on the wrong PR ^.

tiffwong (approved):

PR #7349, opened 2021-01-28
M3-4833: Change table header font color

tiffwong:
The new grey color has been applied to table headers and the "Download CSV" link on the Linode Landing page

This PR also includes a fix make sure all table headers have a blue background and white text on hover which was previously missing on the Linode Landing page and a few other tables


DevDW (approved):
Looks good, the table header hover behavior appears to be consistent across the app now. Does the contrast meet the a11y standards Matthew mentioned on the ticket?

Jskobos (approved):

PR #7350, opened 2021-01-28
Firewalls: update submission of IP addresses

johnwcallahan:
A change made in the API means that an empty address list is not accepted. Therefore if there are no ipv6 addresses for example, we don't include the ipv6 field at all.

I also made `ports` optional since that was changed in the API as well.

**To test,** use the dev API and check FW rule submission/editing thoroughly.

### DIFF
--- a/packages/api-v4/src/firewalls/types.ts
+++ b/packages/api-v4/src/firewalls/types.ts
@@ -21,7 +21,7 @@ export interface FirewallRules {
 
 export interface FirewallRuleType {
   protocol: FirewallRuleProtocol;
-  ports: string;
+  ports?: string;
   addresses?: null | {
     ipv4?: null | string[];
     ipv6?: null | string[];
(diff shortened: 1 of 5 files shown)


Jskobos (changes requested):
Screen Shot 2021-01-29 at 9 08 44 AM

Did we ever resolve what was supposed to happen here? In any case looks like sending an empty object for addresses won't work. Probably should delete the field from the payload if it's empty (though the error message makes it seem like the API isn't handling it properly either).

johnwcallahan:
@Jskobos good catch. Addresses are now required, so I pushed up that change.

Jskobos (approved):
Sorry I forgot to re-review. I can still get to the bad place by selecting "IP/Netmask", leaving the text field blank, and submitting the form. I'm going to approve because honestly that's valid, but wanted to raise the issue.

DevDW (approved):

PR #7351, opened 2021-01-28
M3-4844: New Linode logo

tiffwong:
Replace existing logo with new logo from marketing
- The `logoCollapsed` class wasn't doing anything so I repurposed it to cover part of the Linode text from showing when the menu is collapsed
- I also deleted previous versions of the logo that wasn't being used anywhere


DevDW (approved):
To me it looks a tad empty at the top of the collapsed menu when you hover over it now since the text is hidden. What do you think about centering the logo in that instance?

tiffwong:
@DevDW it shouldn't be covered on hover so I'll fix that!

Jskobos (approved):

PR #7352, opened 2021-01-28
M3-4835: New Accordion expand/collapse icons

tiffwong:
Swap the previous expand/collapse icons used in the `Accordion` component and the same icons used in `/support/tickets`


Jskobos (approved):
I think you can remove the minus-square and plus-square svgs (you have to delete the non-CMR AddNewLink)

DevDW (approved):

PR #7353, opened 2021-01-28
M3-4840 Change: Notification drawer events

Jskobos:
## Description

- 25 most recent events are displayed in the drawer (number is a guesstimate that will change when the drawer is finalized)
- In progress events appear at the top of the list with a progress bar
- Older events are below, using messaging from Events landing
- Unseen events are bold
- When the drawer is closed, all events are marked as seen.
- Deleted a bunch of stuff.

This matches the designs, but still will likely see some tweaks.

## Discussion Questions

- The display for progress events is inching closer to non-progress events. We could probably make them basically the same, except that progress events have a progress bar and "remaining".  I haven't done that yet because:
    a) I deviated from the designs a bit already and want to get feedback before I go much further
    b) The new progress messages include extra information (e.g. what region a Linode is migrating to). If we use the old 
         messages, we can consider dropping this info, adding it to the old eventMessageGenerator, or just rethinking what we 
         would like to include. In any case I think it's too large a question for this PR.


tiffwong (approved):
Looks good! Only question I have is if the icons should be active/green if they're unseen

johnwcallahan (approved):
Works really well. Two styling things I noticed which can be addressed now or later:

Icons look a little high to me:
Screen Shot 2021-02-01 at 11 38 03 AM

Only things that are actually clickable should have the underline on hover:
Screen Shot 2021-02-01 at 11 38 19 AM

DevDW (approved):

PR #7355, opened 2021-01-29
M3-4854: Fix StackScripts search

tiffwong:
Removing the type from `querySearch` fixes the search redirecting to /accounts

**Update:**
- If the username matches the one in the profile, the username link from the StackScripts Detail page redirects to `/accounts`
- If not, it'll redirect to `/community/?query=username[insert username]` instead
- Change the table headers to the new gray color
- Made the link underline color match the text instead of blue

### DIFF
--- a/packages/manager/src/components/StackScript/StackScript.tsx
+++ b/packages/manager/src/components/StackScript/StackScript.tsx
@@ -112,10 +112,7 @@ export const SStackScript: React.FC<Props> = props => {
     return imageChips.length > 0 ? imageChips : <>No compatible images found</>;
   }, [images, imagesData]);
 
-  const queryString = stringify({
-    type: 'community',
-    query: `username:${username}`
-  });
+  const queryString = stringify({ query: `username:${username}` });
 
   return (
     <div className={classes.root}>
@@ -126,7 +123,10 @@ export const SStackScript: React.FC<Props> = props => {
         data-qa-stack-author={username}
       >
         by&nbsp;
-        <Link to={`/stackscripts/${queryString}`} data-qa-community-stack-link>
+        <Link
+          to={`/stackscripts/community?${queryString}`}
+          data-qa-community-stack-link
+        >
           {username}
         </Link>
       </Typography>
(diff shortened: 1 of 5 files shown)


packages/manager/src/components/StackScript/StackScript.tsx
```diff
@@ -126,7 +126,7 @@ export const SStackScript: React.FC<Props> = props => {
         data-qa-stack-author={username}
       >
         by&nbsp;
-        <Link to={`/stackscripts?${queryString}`} data-qa-community-stack-link>
+        <Link to={`/stackscripts/${queryString}`} data-qa-community-stack-link>
```
Jskobos, line 129: Clicking on one of my stackscripts takes me to something like `http://localhost:3000/stackscripts/type=community&query=username%3Aprod-test-004` which doesn't lead anywhere

Jskobos (changes requested):

packages/manager/src/components/StackScript/StackScript.tsx
tiffwong: Is this from clicking on your stackscript from /accounts?

Jskobos: Clicking on my username from StackScript detail

tiffwong: What's the behavior supposed to be if you click on a username (that's not yours) from a community stackscript?

Jskobos: Should go back to the relevant landing page with the search query set to that username. So that url above needs to be `/stackscripts/community/?query=username....`

DevDW (approved):
Search is working now 👌

packages/manager/src/components/StackScript/StackScript.tsx
Jskobos: OK that wasn't a good example...if you click on your own script or the username of someone on your account, it has to go to `/account?query=username...`.

tiffwong: When I tried the search query for my username on /accounts, there's no results 🤔 

Also I think the original redirected to /community, what's the best way to get the category type?

Jskobos (approved):

PR #7356, opened 2021-01-29
M3-4836 & M3-4860: Vertically center hamburger menu and clean up NodeBalancers buttons

tiffwong:
M3-4836: Vertically center the hamburger menu in the top nav

M3-4860: Remove plus icon from "Add a Node" link, change "Delete" button type and swap places with the "Save" button


Jskobos (approved):
The alignment of the three text nodes looks odd to me:
Screen Shot 2021-02-01 at 10 23 09 AM

tiffwong:
@Jskobos it's cause "Add a Node" and "Delete" are secondary button types which results in them being centered and looking weird near one another 🤷‍♀️

DevDW (approved):

PR #7363, opened 2021-02-02
M3 4851 linode storage tab tests

ndukatz:
Linode Storage Tab Tests, and I snuck in a cypress upgrade and linode api utility refactor. This probably should have been 2-3 separate PRs. I will split it up if anyone thinks it would be beneficial.


packages/manager/cypress/integration/smoke-visual-icons.spec.ts
```diff
@@ -58,7 +56,7 @@ describe.skip('Checking icons render correctly', () => {
           url: '*/linode/instances/*',
           method: 'GET',
           response: {
-            data: [makeLinodeDataWithStatus('running')]
+            // data: [makeLinodeDataWithStatus('running')]
```
ndukatz, line 59: this is just to avoid errors (this test is not running right now)

ndukatz:
looks like the backups test is still a little flaky, or something else is going on. I will look into that today and If I find the issue I'll sneak it in here as well.

Jskobos (approved):
The yarn.lock change is humongous for a minor version bump 🤷

packages/manager/cypress/support/api/linodes.ts
```diff
@@ -79,83 +35,35 @@ const makeLinodeCreateReq = (linode, password) => {
...
     : {
-        root_pass: strings.randomPass(),
+        root_pass:
+          requestType === RequestType.PASSWORD ? value : RequestType.PASSWORD,
         label: makeLinodeLabel(),
+        status: requestType === RequestType.STATUS ? value : RequestType.STATUS,
         type: 'g6-standard-2',
-        region,
-        image: 'linode/debian10',
+        region: requestType === RequestType.REGION ? value : RequestType.REGION,
```
johnwcallahan, line 47: This pattern is sort of confusing... `RequestType.REGION` is the literal string `"us-east"`, so for example if you wanted to create a Linode in Dallas this works (but is strange):

```ts
makeLinodeCreateReq('us-east', 'us-central');
```

I'd maybe do something like this:

```ts
const makeLinodeCreateReq = (data: Partial<Linode>) => {
  const linodeData = {
    ...linode,
    ...data
  };
  return postReturn(linodeData);
}
```

(With corresponding changes in `createLinode`).

That way you can do:

```ts
createLinode() // create a Linode with the default data only
createLinode({ region: 'us-central' }) // create a Linode in a specific region (defaults for all other fields)
createLinode({ region: 'us-central', backups_enabled: true }) // etc.


```

johnwcallahan (commented):
The new tests are working well. I left a comment about the new utility function behavior.

> The yarn.lock change is humongous for a minor version bump

@Jskobos I think this is picking up changes from the Node version bump. Right now if you run `yarn install` on develop you'll see similar results. In develop right now, take a look at `ts-jest` in the /packages/manager package.json:

```json
"ts-jest": "~26.4.4"
```

But the version in yarn.lock doesn't match:

```yaml
ts-jest@^24.1.0:
  version "24.1.0"
```

Not really sure what the correct approach is.

packages/manager/cypress/support/api/linodes.ts
ndukatz: > This pattern is sort of confusing... `RequestType.REGION` is the literal string `"us-east"`, so for example if you wanted to create a Linode in Dallas this works (but is strange):
> 
> ```ts
> makeLinodeCreateReq('us-east', 'us-central');
> ```
> 
> I'd maybe do something like this:
> 
> ```ts
> const makeLinodeCreateReq = (data: Partial<Linode>) => {
>   const linodeData = {
>     ...linode,
>     ...data
>   };
>   return postReturn(linodeData);
> }
> ```
> 
> (With corresponding changes in `createLinode`).
> 
> That way you can do:
> 
> ```ts
> createLinode() // create a Linode with the default data only
> createLinode({ region: 'us-central' }) // create a Linode in a specific region (defaults for all other fields)
> createLinode({ region: 'us-central', backups_enabled: true }) // etc.
> ```

Thank you! I agree it is weird. I will try to implement this.

johnwcallahan (approved):

PR #7367, opened 2021-02-03
Add: Firewalls beta notification

Jskobos:
## Description

Add some info about the fact that Firewalls is still in beta (since everyone will be able to see it as a normal feature starting next Wednesday). All copy and style is ad hoc and up for discussion.


johnwcallahan (commented):
The badge in the primary nav looks great. 

FWIW we have a MainContentBanner component that we used for the BLM banner. Obviously the styles are way off but these components are very similar functionally. 

My 2 cents: maybe check with Matthew on styles... seems like the Firewall Icon could be fun here. Maybe some more margin between the banner and main content:
Screen Shot 2021-02-03 at 4 57 40 PM

tiffwong (approved):

johnwcallahan (approved):

PR #7371, opened 2021-02-04
Prevent SelectPlanPanel wrap in Linode Create flow

tiffwong:
Change to card view at an earlier breakpoint to prevent the label from wrapping

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
@@ -1,5 +1,6 @@
-import * as classnames from 'classnames';
 import { LinodeTypeClass } from '@linode/api-v4/lib/linodes';
+import * as classnames from 'classnames';
+import { LDClient } from 'launchdarkly-js-client-sdk';
 import { isEmpty, pathOr } from 'ramda';
 import * as React from 'react';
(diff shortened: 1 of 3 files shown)


DevDW (approved):

packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
```diff
@@ -182,172 +177,156 @@ export class SelectPlanPanel extends React.Component<CombinedProps> {
...
-                {tableHeader}
-                <TableBody role="radiogroup">
-                  {plans.map(this.renderSelection)}
-                </TableBody>
-              </Table>
-            )}
-          </Grid>
-        </Hidden>
+        {isCreate ? (
+          <Hidden lgUp>{plans.map(this.renderSelection)}</Hidden>
```
Jskobos, line 320: Maybe something like `<Hidden lgUp={isCreate} mdUp{!isCreate}>...` would work?

Jskobos (approved):

packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
tiffwong: Brilliant 😱

PR #7372, opened 2021-02-04
M3-4884: Increase body text line-height 

tiffwong:
Increase line height for body text to 18px and remove duplicate body variant (replace all usages of `body2` with `body1`)


Jskobos (approved):
Clicked around and couldn't see anything that looked off.

DevDW (approved):
LGTM 🚢

PR #7381, opened 2021-02-08
M3-4906: Reduce vertical spacing between Label & Description fields on StackScript Create/Edit page

DevDW:
## Description
Make the vertical spacing between `StackScript Label` and `Description` on the StackScript Create/Edit page more consistent with the spacing between the other fields.

## Type of Change


Jskobos (commented):
I'm not sure we need this...if you check out the feature/firewall-updates the spacing is already improved (from a fix John put in there for the icon alignment). It's all margin at that point, too, which I can bring to 0 with the `noMarginTop` prop to TextField: 

Screen Shot 2021-02-08 at 1 36 20 PM

DevDW:
@Jskobos I agree, this is unneeded now given the changes in feature/firewall-updates. I'll close this PR.

PR #7383, opened 2021-02-08
M3-4784 Add: port presets in FirewallRuleDrawer

Jskobos:
## Description

This is mostly working; I'm going to hold off on finishing touches etc. until we've gotten some feedback on the approach, since some of the possibilities I've discussed offline would require throwing out most of this logic.

### DIFF
--- a/packages/api-v4/src/firewalls/firewalls.schema.ts
+++ b/packages/api-v4/src/firewalls/firewalls.schema.ts
@@ -34,6 +34,11 @@ export const ipAddress = string().test({
   test: validateIP
 });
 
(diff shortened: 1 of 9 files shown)


johnwcallahan (changes requested):
Works great. One small thing: the Ports input should be cleared when opening the drawer to add a new rule (currently it uses the last value).

Small suggestion: what do you think about renaming "Port Range" to "Custom Port Range"? Just trying to find a way to link the "Custom" select item to the field below.

DevDW (approved):
LGTM, everything seems to be working well 👌

johnwcallahan (approved):

PR #7384, opened 2021-02-08
M3-4891: Remove sidebar upload panel in Object Storage

tiffwong:
Remove icon and prevent the upload panel from being displayed in the sidebar for all viewports

**Update:** The scope has expanded to include a few UI adjustments:
- Removed the left icon when making an upload
- Removed the filename truncation
- Changed overwrite message and button
- Updated progress bar color


Jskobos (approved):
I liked the icon 😢 ...this looks good though, tables have margin against the edge of the viewport down to about 960 px then they're flush, the drop zone is using this style too and it looks a bit weird having the dotted line right at the edge of the screen.

johnwcallahan (changes requested):
There needs to be a max height with overflow set to scroll. Currently the upload box takes over the whole page:
Screen Shot 2021-02-09 at 12 22 01 PM

I think there are some other things we need to consider here too, as the layout kind of breaks in general when uploading stuff:
Screen Shot 2021-02-09 at 12 24 15 PM

johnwcallahan:
> I think there are some other things we need to consider here too, as the layout kind of breaks in general when uploading stuff

E.g. we should consider moving the size and status icons to the left side (by the filename) and left-align the Cancel/Overwrite buttons instead of centering. Let's sync up with Matthew on this.

johnwcallahan (approved):

PR #7385, opened 2021-02-08
M3-4894: Entity Transfer routing

tiffwong:
Added routing for:
- [/account/entity-transfers](http://localhost:3000/account/entity-transfers)
- [/account/entity-transfers/create](http://localhost:3000/account/entity-transfers/create)


packages/manager/src/features/EntityTransfers/EntityTransfersCreate/EntityTransfersCreate.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+interface Props {}
+
+export const EntityTransfersCreate: React.FC<Props> = props => {
+  return (
+    <>
+      <DocumentTitleSegment segment="Create a Transfer" />
+      <Breadcrumb
+        pathname={location.pathname}
```
Jskobos, line 13: Seems like we might want the full crumb here? account / transfers / create

packages/manager/src/features/EntityTransfers/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <Switch>
+      <Route component={EntityTransfersCreate} path={`${path}/create`} />
+      <Route component={EntityTransfersLanding} path={path} exact strict />
+      <Redirect to={path} />
+    </Switch>
+  );
+};
+
+export default withRouter(EntityTransfers);
```
johnwcallahan, line 33: I'd opt for the React Router hooks rather than the HOC for function components.

packages/manager/src/features/TopMenu/UserMenu/UserMenu.tsx
```diff
@@ -265,6 +265,12 @@ export const UserMenu: React.FC<{}> = () => {
         href: '/account/users',
         hide: _isRestrictedUser
       },
+      // Restricted users can't view the Transfers tab regardless of their grants
+      {
+        display: 'Transfers',
+        href: '/account/entity-transfers',
+        hide: _isRestrictedUser
```
johnwcallahan, line 272: I know I mentioned that restricted users can't use entity transfers, but now I'm wondering if a restricted user with _account access_ actually might be able use it. Please check with the API team for clarification, because if so you'll need to use `_hasAccountAccess` instead.

Jskobos (approved):
Maybe we want to hide the tab for restricted users, though that isn't what we do for /account/users for some reason

packages/manager/src/features/TopMenu/UserMenu/UserMenu.tsx
tiffwong: @johnwcallahan I asked and restricted users will not have access to account transfers in any capacity

DevDW (approved):
I agree with @Jskobos on hiding the Transfers tab for restricted users, but this LGTM otherwise 👍

johnwcallahan (approved):

PR #7364, opened 2021-02-02
M3-4839: Add Notifications section in Notifications drawer & a Past Due notice therein

DevDW:
## Description
Add Notifications section and a "Past Due" item therein if the customer has a balance (shown starting on the 3rd of each month -- update Billing Header on the Account Page with this logic as well).

## Type of Change
- Non breaking change ('update', 'change')

## To-Do:
- [x] Feed all of the data into the Notifications section that should be there
- [x] Add logic for "Past Due" item
- [ ] Clear Notification section items from badge total once user closes Notification drawer
- [ ] Formatting of Notification section items
- [ ] Cleanup

The logic for having hyperlinks where appropriate in the items, etc. will be done in a follow-up PR

PR #7386, opened 2021-02-08
M3-4839: Notifications section in Notifications drawer follow-up

DevDW:
## Description
Follow-up to https://github.com/linode/manager/pull/7364 to improve formatting of Notification section items and more closely match the mockup:

Mockup_02

## Type of Change

### DIFF
--- a/packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
+++ b/packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
@@ -63,7 +63,6 @@ const interceptNotification = (notification: Notification): Notification => {
     /** replace "this facility" with the name of the datacenter */
     return {
       ...notification,
-      severity: adjustSeverity(notification),
       label: notification.label
         .toLowerCase()
         .replace('this facility', convertedRegion || 'one of our facilities'),
@@ -76,8 +75,7 @@ const interceptNotification = (notification: Notification): Notification => {
   if (notification.type === 'ticket_abuse') {
     return {
       ...notification,
-      message: notification.message.replace('!', '.'),
-      severity: adjustSeverity(notification)
+      message: notification.message.replace('!', '.')
     };
   }
 
@@ -119,22 +117,16 @@ const formatNotificationForDisplay = (
   countInTotal: true
 });
 
-// For communicative purposes in the UI, in some cases we want to upgrade or downgrade the severity of certain notifications compared to what the API returns. Example: the API has ticket_abuse as having a severity of major, but we want to show those notifications as critical.
+// For communicative purposes in the UI, in some cases we want to adjust the severity of certain notifications compared to what the API returns. If it is a maintenance notification of any sort, we display them as major instead of critical. Otherwise, we return the existing severity.
 const adjustSeverity = ({
   severity,
   type
 }: Notification): NotificationSeverity => {
-  if (
-    checkIfMaintenanceNotification(type) ||
-    (severity === 'major' && type !== 'ticket_abuse')
-  ) {
+  if (checkIfMaintenanceNotification(type)) {
     return 'major';
   }
-  if (severity === 'critical' || type === 'ticket_abuse') {
-    return 'critical';
-  }
 
-  return 'minor';
+  return severity;
 };
 
 export default useFormattedNotifications;
--- a/packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
+++ b/packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
@@ -118,7 +118,7 @@ const formatNotificationForDisplay = (
 });
 
 // For communicative purposes in the UI, in some cases we want to adjust the severity of certain notifications compared to what the API returns. If it is a maintenance notification of any sort, we display them as major instead of critical. Otherwise, we return the existing severity.
-const adjustSeverity = ({
+export const adjustSeverity = ({
   severity,
   type
 }: Notification): NotificationSeverity => {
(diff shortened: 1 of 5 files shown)


packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
```diff
@@ -65,24 +66,37 @@ const interceptNotification = (notification: Notification): Notification => {
...
+      ...notification,
+      message: notification.message.replace('!', '.'),
+      severity: reduceSeverity(notification)
+    };
+  }
+
   /** there is maintenance on this Linode */
   if (
-    notification.type === 'maintenance' &&
+    (notification.type === 'maintenance' ||
```
johnwcallahan, line 79: You defined a nice `isMaintenanceNotification` variable in RenderNotification above. I'd turn that into a function, move it somewhere else, and use it here too.

packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
```diff
@@ -65,24 +66,37 @@ const interceptNotification = (notification: Notification): Notification => {
...
+      notification.type === 'migration_scheduled' ||
+      notification.type === 'migration_pending') &&
     notification.entity &&
     notification.entity.type === 'linode'
```
johnwcallahan, line 83: `notification.entity && notification.entity.type === 'linode'` can be shortened to `notification.entity?.type === 'linode`. (Appears in multiple places).

If `notification.entity` doesn't exist, `notification.entity?.type` will be `undefined` and thus the comparison will be false.

packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
```diff
@@ -65,24 +66,37 @@ const interceptNotification = (notification: Notification): Notification => {
...
-      severity: 'major',
-      message: `${
-        linodeAttachedToNotification
-          ? `Linode ${linodeAttachedToNotification}`
-          : `This Linode`
-      }
-            has scheduled maintenance`
+      severity: reduceSeverity(notification),
+      message: linodeAttachedToNotification
+        ? notification!.body!.replace(
```
johnwcallahan, line 95: It's almost always a bad idea to use the assertion operator. This is currently resulting in a crashing bug on test account 1 because it is currently receiving this notification, which doesn't have a body:

Screen Shot 2021-02-09 at 10 39 24 AM

Jskobos (approved):
Looks good. The drawer still needs to close when a notification is clicked, that can be done in a separate PR.

packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
```diff
@@ -98,4 +112,23 @@ const formatNotificationForDisplay = (
   countInTotal: true
 });
 
+const reduceSeverity = ({ severity, type }: Notification) => {
+  if (
+    ['maintenance', 'maintenance_scheduled', 'migration_pending'].includes(
```
johnwcallahan, line 117: Here's another instance of the `isMaintenanceNotification` logic you could reuse.

packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
```diff
@@ -98,4 +112,23 @@ const formatNotificationForDisplay = (
...
+      type
+    ) ||
+    (severity === 'major' && type !== 'ticket_abuse')
+  ) {
+    return 'major';
+  }
+  if (severity === 'critical' || type === 'ticket_abuse') {
+    return 'critical';
+  }
+  if (severity === 'minor') {
```
johnwcallahan, line 127: I think this is redundant... at this point we're going to return `minor` anyway, so why do the additional `if` check?

johnwcallahan (changes requested):
Layout, messaging, etc. looks great. Still a crashing bug to address and some smaller code things to clean up.

packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
johnwcallahan: Meant to leave an additional comment for this function: I know it's based on old code but we should add a few unit tests to verify the pathways are working out like we expect. Also please add some comments... the code is easy to read but the _why_ is unclear.

packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
johnwcallahan: It was a migration_scheduled notification that was crashing it, not a ticket_important notification as I first indicated.

Screen Shot 2021-02-09 at 1 24 01 PM

packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
```diff
@@ -98,4 +111,19 @@ const formatNotificationForDisplay = (
   countInTotal: true
 });
 
+// For communicative purposes in the UI, in some cases we want to upgrade or downgrade the severity of certain notifications compared to what the API returns. Example: the API has ticket_abuse as having a severity of major, but we want to show those notifications as critical.
```
johnwcallahan, line 114: Thanks, this comment made it clearer.

Do we actually want to downgrade emergency maintenance tickets to `major`, though? That's a departure from what's currently in develop, which I think is more appropriate:

Screen Shot 2021-02-09 at 1 45 01 PM

If that's the case, all this function is doing is upgrading ticket_abuse to `critical`, so the function body can be something like:

```ts
return type === 'ticket_abuse' ? 'critical' : severity;
```

In that case we should rename the function since it's misleading (makes me think it's _reducing_ the severity level instead of the intent, which is to serve as a reducer). 

Also, the return type should be `NotificationSeverity` instead of a plain string.

DevDW: I may be mistaken, but during the last demo I thought Chris mentioned that Linode maintenance alerts shouldn't be so scary. If that shouldn't apply to the emergency maintenance (is that just `maintenance_pending` types?) notices I'll get this fixed

johnwcallahan (approved):
Thanks for making these changes, the adjustments look good. I have a handful of remaining questions we can discuss with the team offline.

PR #7388, opened 2021-02-10
M3-4900 Add: Create new entity transfer (Skeleton)

Jskobos:
## Description

Putting this up to get some feedback on the general approach before I get too far down the rabbit hole.

## FAQ

Q: Does this match the designs?
A: No

Q: Does it include all the important functionality?
A: lol no

Q: Does it use real data?
A: also no

Q: Is it hideous?
A: God yes

Q: Is the code clean and tested at least?
A: Stop hounding me

Q: What does it do, exactly?
A: You should be able to select all/some entities, deselect them, and submit the form to generate a payload.

## Questions for reviewers

- Is the abstraction for the tables worth it? I'm hoping to simplify the process for adding other entity types down the road but the savings is kind of marginal.
- What about the table row with a checkbox? Same reason, we want to do bulk actions everywhere eventually so I thought it would be helpful to have, but the actual savings are on the border between worth it/not worth it.

### DIFF
--- a/packages/manager/src/features/EntityTransfers/EntityTransfersCreate/transferReducer.ts
+++ b/packages/manager/src/features/EntityTransfers/EntityTransfersCreate/transferReducer.ts
@@ -1,10 +1,58 @@
+import produce from 'immer';
 export type TransferableEntity = 'linodes';
 export const TRANSFERABLE_ENTITIES: TransferableEntity[] = ['linodes'];
 
+export type Entity = { label: string; id: string | number };
 export type TransferEntity = Record<string, string>;
-export type TransferState = Record<TransferableEntity, TransferEntity[]>;
+export type TransferState = Record<TransferableEntity, TransferEntity>;
 
 export const defaultTransferState: TransferState = TRANSFERABLE_ENTITIES.reduce(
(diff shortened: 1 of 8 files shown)


packages/manager/src/features/EntityTransfers/EntityTransfersCreate/transferReducer.ts
```diff
@@ -0,0 +1,58 @@
+import produce from 'immer';
```
Jskobos, line 1: Of course there's a reducer!

johnwcallahan (approved):
Way to go for putting this together so quickly; this works exactly how I'd expect and the architecture makes sense to me. 

> Is the abstraction for the tables worth it?

Yes, especially once we add styling.

> What about the table row with a checkbox?

Yes, same reason. Depending on how specific it gets styled, I might consider putting the component in the EntityTransfers feature directory instead of the components directory, but that could be debated either way.

Reducer looks good.

tiffwong (approved):
Woah, great job getting this set up!

DevDW (approved):
This is looking good; I think the infrastructure for the down-the-road pieces will prove useful 👍

PR #7389, opened 2021-02-10
Add: Account transfer methods for JS client and mock service worker

Jskobos:
## Description

New methods for the account transfer feature, based on a spec doc from the API. Also includes mocks. These methods may not be complete/final, but they should be enough for us to work with.


johnwcallahan (approved):

DevDW (approved):
Looks good 🚢

PR #7390, opened 2021-02-10
Notifications section of Notifications drawer follow-up Pt. II

DevDW:
## Description
As discussed in the stand-up today, 
1) filter out `payment_due` (since we use a custom one) and `ticket_abuse` and bounced email notifications (since we have global notifications for those), 
2) fix the alignment of notification items that have minor severity, 
3) handle issues with casing and links included in `notification.message` as returned by the API.

## Type of Change

## To-Do:

- [x] Handle casing issues
- [x] Handle issues with links included in `notification.message` as returned by the API


packages/manager/src/features/NotificationCenter/NotificationData/RenderNotification.tsx
```diff
@@ -178,4 +162,50 @@ const linkifiedMaintenanceMessage = (
...
+        <Link to={'https://status.linode.com/'}>https://status.linode.com</Link>{' '}
+        for further information. Thank you for your patience and understanding.
+      </Typography>
+    );
+  }
+
+  return notification.message;
+};
+
+const hrefRegex = /<a href=('|")(.+)('|")>(.+)<\/a>/i;
```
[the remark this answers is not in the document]
Jskobos: (As a rule of thumb if you ever find yourself dissecting strings, it's a good time to ask yourself, "Am I sure there's not a better way?" Sometimes there isn't! But it ought to be a last resort.)

packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -387,6 +387,36 @@ export const handlers = [
     //   body: null
     // });
 
+    const generalGlobalNotice = {
+      type: 'notice',
+      entity: null,
+      when: null,
+      // eslint-disable-next-line xss/no-mixed-html
+      message:
+        "We've updated our policies. See <a href='https://manager.linode.com/account/policy'>https://manager.linode.com/account/policy</a> for more information.",
```
Jskobos, line 396: All of the actual API notifications follow this pattern where the href and link body are the same, but I don't think it's safe to build your logic around this assumption. (The ideal behavior is for the API to not return inline html in the first place, but 🤷 )

packages/manager/src/features/NotificationCenter/NotificationData/RenderNotification.test.tsx
```diff
@@ -0,0 +1,27 @@
+import { handleEmbeddedHTML } from './RenderNotification';
+
+const plainStringMessage =
+  'You have a notification consisting of a plain string with no embedded HTML.';
+// eslint-disable-next-line xss/no-mixed-html
+const messageWithEmbeddedHTML =
+  "We've updated our policies. See <a href='https://manager.linode.com/account/policy'>https://manager.linode.com/account/policy</a> for more information.";
+
+describe('Rendering notifications', () => {
```
Jskobos, line 9: Thanks for including tests

Jskobos (changes requested):
The only required change is the handling of HTML; release is on Monday so maybe it's best to just tackle that for now and do a follow-up for some of the more stylistic cleanup/efficiency stuff later.

Jskobos (approved):
Thanks for the quick fix. I'd say you can delete the test file. And something else we should discuss in the future is adding a link to "Important Ticket" notifications, like these from test account 1: 

Screen Shot 2021-02-12 at 2 41 46 PM

johnwcallahan (approved):

PR #7391, opened 2021-02-11
M3-4900 Change: Styling for transfer checkout bar

Jskobos:
## Description

Includes styling along with the max-width and scroll behavior discussed offline. The fake data has been updated so that if you click the select all box you can see the scroll behavior.

### DIFF
--- a/packages/manager/src/features/EntityTransfers/EntityTransfersCreate/LinodeTransferTable.tsx
+++ b/packages/manager/src/features/EntityTransfers/EntityTransfersCreate/LinodeTransferTable.tsx
@@ -16,7 +16,7 @@ interface Props {
   handleToggle: (linode: Entity) => void;
 }
 
-const linodes = linodeFactory.buildList(10);
+const linodes = linodeFactory.buildList(200);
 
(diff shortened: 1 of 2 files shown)


johnwcallahan (approved):
The height of the Checkout Bar rows could come down a bit (they're currently 45px and the spec is 40px) but everything else looks good. 

[for later:] One potential risk I see is that with this component it's hard to tell if there are more items to scroll down to (at least for people like me, who have scrollbars hidden by default on MacOS). We've done the "fade" pattern before but there are other options (arrows, etc.) @tiffwong let's chat with Matthew about this.

DevDW (approved):
Looks good aside from the "Generate Transfer Token" button getting a tad messy at mobile view:

Screen Shot 2021-02-12 at 10 45 31 AM

Jskobos:
> Looks good aside from the "Generate Transfer Token" button getting a tad messy at mobile view:
> 
> Screen Shot 2021-02-12 at 10 45 31 AM

I haven't done any mobile anything for this entire page yet, going to wait until the pieces are mostly done and in the right places.

PR #7392, opened 2021-02-11
Styling for transfer controls

Jskobos:


johnwcallahan (commented):
Looks good at larger viewports; mobile needs a little work. Also, in the mockup there is a little more spacing between the token input and the "Review Details" button (16px).

johnwcallahan (approved):

DevDW (approved):
I think we could use some left padding for the input field at tablet & mobile views and for the "Make a Transfer" button at tablet views:

Screen Shot 2021-02-12 at 10 20 15 AM

Screen Shot 2021-02-12 at 10 21 09 AM

Jskobos:
I'm not finished the mobile stuff here but going to tackle that later.

PR #7393, opened 2021-02-12
M3-4900 Table style for Create Transfer page

Jskobos:
## Description

Searchbar isn't there yet, data is still static, and mobile is a work in progress (because the checkout bar is in a separate PR). But the table itself should match the designs now.

### DIFF
--- a/packages/manager/src/features/EntityTransfers/EntityTransfersCreate/TransferTable.tsx
+++ b/packages/manager/src/features/EntityTransfers/EntityTransfersCreate/TransferTable.tsx
@@ -42,6 +42,9 @@ const useStyles = makeStyles((theme: Theme) => ({
       width: 20,
       height: 20
     }
+  },
+  checkEmpty: {
+    '& svg': { color: '#cccccc' }
   }
 }));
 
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/EntityTransfers/EntityTransfersCreate/TransferTable.tsx
```diff
@@ -3,19 +3,48 @@ import Table from 'src/components/Table/Table_CMR';
...
+    }
+  },
+  check: {
+    '& svg': {
+      width: 20,
+      height: 20
+    }
+  },
+  checkEmpty: {
+    '& svg': { color: '#cccccc' }
```
Jskobos, line 47: This is really dumb and if @tiffwong or anyone else can think of a better way that would be awesome.

DevDW (approved):
Table looks good, just needs some left padding at smaller viewports for the table name and explanatory text above it:

Screen Shot 2021-02-12 at 11 09 44 AM

Also, are we holding out on updating the explanatory text until the text is finalized?

johnwcallahan (approved):

PR #7394, opened 2021-02-12
M3-4898 Receive Transfer Confirmation

Jskobos:
## Description

- Add an override for the styling in src/components/Dialog
- Fill in the Accept a Transfer dialog
- It'll work with any random token with the mocks.

Screen Shot 2021-02-12 at 1 31 05 PM

## Todo

- [x] Refactor and clean up the time remaining calculation
- [ ] I'd like to do some validation on the token before opening the modal (need API input)
- [x] Tests

### DIFF
--- a/packages/manager/src/features/EntityTransfers/EntityTransfersLanding/ConfirmTransferDialog.tsx
+++ b/packages/manager/src/features/EntityTransfers/EntityTransfersLanding/ConfirmTransferDialog.tsx
@@ -35,15 +35,16 @@ const useStyles = makeStyles((theme: Theme) => ({
     fontSize: '1rem'
   },
   entityTypeDisplay: {
-    marginTop: theme.spacing(),
     marginBottom: theme.spacing()
   },
   summary: {
-    fontSize: '1rem'
+    fontSize: '1rem',
+    marginBottom: 4
   },
   list: {
     listStyleType: 'none',
-    paddingLeft: theme.spacing(2)
+    paddingLeft: 0,
+    margin: 0
   }
 }));
 
@@ -210,7 +211,7 @@ export const DialogContent: React.FC<ContentProps> = React.memo(props => {
         <ul className={classes.list}>
           {Object.keys(entities).map(thisEntityType => {
             // According to spec, all entity names are plural and lowercase
-            // (NB: This will cause problems for NodeBalancers if/when they are added to the payload)
+            // (NB: This may cause problems for NodeBalancers if/when they are added to the payload)
             const entityName = capitalize(thisEntityType).slice(0, -1);
             return (
               <li key={thisEntityType}>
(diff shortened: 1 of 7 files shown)


packages/manager/src/features/EntityTransfers/EntityTransfersLanding/EntityTransfersLanding.tsx
```diff
@@ -6,6 +6,13 @@ import TransferControls from './TransferControls';
 export const EntityTransfersLanding: React.FC<{}> = _ => {
   const [confirmDialogOpen, setConfirmDialogOpen] = React.useState(false);
   const [token, setToken] = React.useState('');
+
+  const handleCloseDialog = () => {
+    setConfirmDialogOpen(false);
+    // I don't love the UX here but it seems better than leaving a token in the input
+    setTimeout(() => setToken(''), 150);
```
Jskobos, line 13: Thoughts on this plz

tiffwong: I don't think it's the worst, the sudden change in the button color is a little jarring but I'm not sure if there'a better way of handling it 💁‍♀️

tiffwong (approved):

packages/manager/src/features/EntityTransfers/EntityTransfersLanding/ConfirmTransferDialog.tsx
```diff
@@ -1,25 +1,264 @@
...
-export type CombinedProps = Props;
-
 export const ConfirmTransferDialog: React.FC<Props> = props => {
   const { onClose, open, token } = props;
+  const classes = useStyles();
+  const { enqueueSnackbar } = useSnackbar();
+  const { data, isLoading, isError, error } = useTransferQuery(
+    token ?? '',
+    open
+  );
```
DevDW, line 64: It seems like this is firing off on every input change to the token `TextField`. Should it only do so when the user opens this modal?

DevDW (approved):
LGTM overall, I raised a couple of points/questions in the comments

packages/manager/src/features/EntityTransfers/EntityTransfersLanding/ConfirmTransferDialog.tsx
Jskobos: It should only make a request when the dialog is open...which is what's happening for me. Could you please share a video so I can see how to reproduce this?

packages/manager/src/features/EntityTransfers/EntityTransfersLanding/EntityTransfersLanding.tsx
Jskobos: I'm not able to reproduce this one either, if I'm understanding it correctly. Could you put up a video for this as well? 🙇

DevDW:
@jskobos I went to record the token remaining in the input field issue, but wasn't able to reproduce it this time 🤔 and on the React Query front, I think I may have misinterpreted the individual queries in the RQ DevTools as each being a request, but it is in fact only the final one that has a request tied to it, and that happens only when you open the confirmation modal (since `data` gets populated for that request). So false alarms, this should be good to go 🚢

PR #7395, opened 2021-02-13
M3-4896/M3-4897: Transfers Pending, Received, & Sent tables

DevDW:
## Description
This PR adds the Transfers Pending, Received, and Sent tables to the Transfers landing page.

## Type of Change

## To-Do:
- [x] Unit tests
- [x] Adjustments based on feedback

### DIFF
--- a/packages/manager/src/features/EntityTransfers/TransfersTable.tsx
+++ b/packages/manager/src/features/EntityTransfers/TransfersTable.tsx
@@ -8,11 +8,10 @@ import TableBody from 'src/components/core/TableBody';
 import TableHead from 'src/components/core/TableHead';
 import TableRow from 'src/components/core/TableRow';
 import DateTimeDisplay from 'src/components/DateTimeDisplay';
-import PaginationFooter from 'src/components/PaginationFooter';
+// import PaginationFooter from 'src/components/PaginationFooter';
 import Table from 'src/components/Table/Table_CMR';
 import TableCell from 'src/components/TableCell';
 import TableContentWrapper from 'src/components/TableContentWrapper';
-import { usePagination } from 'src/hooks/usePagination';
 import capitalize from 'src/utilities/capitalize';
 import { pluralize } from 'src/utilities/pluralize';
 // import ActionMenu from './TransfersPendingActionMenu';
@@ -53,11 +52,10 @@ export const TransfersTable: React.FC<CombinedProps> = props => {
   const { transferType, isLoading, error, transfers } = props;
 
   const classes = useStyles();
-  const { page, pageSize, handlePageChange } = usePagination();
 
   // const [cancelPendingDialogOpen, setCancelPendingDialogOpen] = React.useState(false);
 
-  const transfersCount = transfers?.length ?? 0;
+  // const transfersCount = transfers?.length ?? 0;
 
   const transferTypeIsPending = transferType === 'pending';
   const transferTypeIsSent = transferType === 'sent';
@@ -197,7 +195,7 @@ export const TransfersTable: React.FC<CombinedProps> = props => {
           </TableBody>
         </Table>
       </Accordion>
-      {transfersCount > pageSize ? (
+      {/* {transfersCount > pageSize ? (
         <PaginationFooter
           count={transfersCount}
           handlePageChange={handlePageChange}
@@ -207,7 +205,7 @@ export const TransfersTable: React.FC<CombinedProps> = props => {
           eventCategory="Entity Transfer Table"
           fixedSize
         />
-      ) : null}
+      ) : null} */}
     </div>
   );
 };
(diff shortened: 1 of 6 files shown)


DevDW:
Going ahead and removing the WIP label here. A bit of accidental scope creep  -- I started to get into the Cancel Transfer modal stuff, but that's a separate ticket. Since I started some of that work, I grabbed that ticket, and commented out some of the stuff I did already in this PR and put "Cancel" text as a placeholder where the inline action will be.

Jskobos (changes requested):
Really good work on this. I left a bunch of opinionated cleanup stuff that you can ignore, the only thing I think we have to change is handling of large numbers of transfers (pagination or whatnot). It'll probably be a while until someone has enough transfers for this to matter, but who knows what people will do.

Other general UI feedback:
- It's nice that the entities column scales with entity types but I wonder if we really want rows that thick (down the road):
Screen Shot 2021-02-16 at 8 35 12 AM

- I'd like to think about setting explicit widths for all of these columns so that they line up across tables even when the columns themselves differ, it would look a little cleaner.
- I suggest this below but maybe a global loading/error state rather than three tables mirroring state based on the same request. I'm not as confident about this though.

Jskobos (approved):
Looks good pending the commenting stuff we discussed offline. One thing is that a Pending transfer also appears in sent transfers. You can maybe filter those out really quickly here, but the filtering we're talking about for using separate requests can handle this in the longer term.

packages/manager/src/features/EntityTransfers/TransfersTable.tsx
```diff
@@ -0,0 +1,223 @@
...
+  return (
+    <div className={classes.root}>
+      <Accordion
+        heading={`Transfers ${capitalize(transferType)}`}
+        defaultExpanded
+      >
+        <Table className={classes.table}>
+          <TableHead>
+            <TableRow>
+              <TableCell key="transfer-token-table-header-token">
```
johnwcallahan, line 143: Just curious: why the `keys` on TableCells?

packages/manager/src/features/EntityTransfers/TransfersTable.tsx
```diff
@@ -0,0 +1,223 @@
...
+                Token
+              </TableCell>
+              {transferTypeIsPending || transferTypeIsSent ? (
```
johnwcallahan, line 146: `<Hidden />` props are booleans, so you should do something like this to simplify the ternaries here:

Screen Shot 2021-02-17 at 10 15 13 AM

packages/manager/src/features/EntityTransfers/TransfersTable.tsx
```diff
@@ -0,0 +1,223 @@
...
+          >
+            {transfer.token}
+          </button>
+        </TableCell>
+        <Hidden smDown>
+          <TableCell className={classes.cellContents}>
+            <DateTimeDisplay value={transfer.created} />
+          </TableCell>
+        </Hidden>
+        {transferTypeIsPending ? (
```
johnwcallahan, line 86: See note below; you can remove the duplication and ternary by giving the condition (`transferTypeIsPending`) to `xsDown` on the Hidden component.

packages/manager/src/features/EntityTransfers/TransfersTable.tsx
```diff
@@ -0,0 +1,223 @@
...
+  // const transfersCount = transfers?.length ?? 0;
+
+  const transferTypeIsPending = transferType === 'pending';
+  const transferTypeIsSent = transferType === 'sent';
+
+  // const cancelPendingTransfer = (token: string) => {
+  //   setCancelPendingDialogOpen(true);
+  // }
+
+  const renderTransferRow = (transfer: EntityTransfer, idx: number) => {
```
johnwcallahan, line 67: This function is being called _a lot_ (e.g. once per row for every keystroke in the "Receive a Transfer" input up top. This specific example can be solved other ways... @Jskobos maybe `ConfirmTransferDialog` should be moved to TransferControls so we don't have to keep track of `token` state on the Landing page.)

But in any case, this should probably be its own memoized Function Component (instead of a plain function). I'd go so far as to give it flat props (`token="" created="" linodes={}` etc) so that it can be memoized efficiently.

johnwcallahan (approved):
Good work; left some feedback which should be addressed before release, but we should go ahead and merge this PR so we can connect the flow.

packages/manager/src/features/EntityTransfers/TransfersTable.tsx
Jskobos: I thought about that, figured we might need that state on the landing page for whatever reason. But it seems not, so I can move it down a level

packages/manager/src/features/EntityTransfers/TransfersTable.tsx
DevDW: these are artifacts from when I was mapping over an array of headers to generate these `<TableCell />`'s -- can remove them in the follow-up

PR #7396, opened 2021-02-13
POC: Firewall Rule reordering

johnwcallahan:
Use the react-beautiful-dnd library to enable reordering for Firewall Rules.

Not done in this PR:
- Styling (see mockups from Matthew: we need to add an icon and check styling of "dragging" state.)
- Updating FW when order changes. To do this I need to update the data structure a bit.

To reiterate: this PR contains client-only re-ordering of FW rules.

Another note: I didn't do the performance optimization the docs suggest (memoized wrapper) because a) to do so we'd need to change _a lot_ of our component tree, and b) in practice I didn't notice any performance regressions with 20+ rules.


Jskobos (approved):

tiffwong (approved):
This is so neat!

PR #7397, opened 2021-02-15
M3-4900 Add real data and pagination to the Transfer table

Jskobos:
## Description

The pagination footer is the style currently in use, and doesn't match the designs. As Matthew says in the comment on the design, the new footer style shouldn't be a blocker for getting this feature out.

### DIFF
--- a/packages/manager/src/features/EntityTransfers/EntityTransfersCreate/TransferTable.tsx
+++ b/packages/manager/src/features/EntityTransfers/EntityTransfersCreate/TransferTable.tsx
@@ -48,7 +48,6 @@ const useStyles = makeStyles((theme: Theme) => ({
     '& svg': { color: '#cccccc' }
   },
   footer: {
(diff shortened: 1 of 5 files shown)


packages/manager/src/features/EntityTransfers/EntityTransfersCreate/TransferTable.tsx
```diff
@@ -45,19 +46,35 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
   checkEmpty: {
     '& svg': { color: '#cccccc' }
+  },
+  footer: {
+    // '& button': { border: 'none' }
```
tiffwong, line 51: This can be removed

tiffwong (approved):
Wow, this looks great! 💯

DevDW (approved):
LGTM 🚢

PR #7399, opened 2021-02-15
Create Transfer Success Modal

Jskobos:
## Description

If you could create a transfer, which you can't, you would see this modal afterwards. The data here is fake, this is just to get the modal working and looking correct.

## Note to Reviewers

The modal will open automatically on this branch if you go to /account/transfers.


johnwcallahan (approved):
Looks great. One minor thing: there's no hover state for the close button ("X").

tiffwong (approved):
It looks great!

Only comment I have is if we can align the "X" button to the right?

Screen Shot 2021-02-16 at 4 23 01 PM

PR #7400, opened 2021-02-15
M3-4900 Connect the wires

Jskobos:
## Description

You should be able to create a transfer. Until #7397 is merged you'll have to use the mock service worker; until #7399 is merged this diff will look very messy.


Jskobos:
The modal needs pluralization but I think otherwise we're good to review.

DevDW (commented):
Aside from the pluralization you already noted, when I block requests to `*entity-transfers*` while on the Create page and go to generate a token, the error makes the page get jumbled up:

Screen Shot 2021-02-16 at 5 29 24 PM

The rest looks good though 👍

DevDW (approved):

tiffwong (approved):

PR #7403, opened 2021-02-16
M3-4792: Linode Detail > Backups buttons

tiffwong:
Add variant for button with outline and apply it to "Cancel Backups" button. Also disable "Take Snapshot" and "Save Schedule" button until a change has been detected

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeBackup/LinodeBackup.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeBackup/LinodeBackup.tsx
@@ -438,6 +438,20 @@ class _LinodeBackup extends React.Component<CombinedProps, State> {
     });
   };
 
+  inputHasChanged = (
+    initialValue: LinodeBackupSchedule,
+    newValue: LinodeBackupSchedule
+  ) => {
+    if (
+      (newValue.day === 'Scheduling' || newValue.day === initialValue.day) &&
+      (newValue.window === 'Scheduling' ||
+        newValue.window === initialValue.window)
+    ) {
+      return true;
+    }
+    return false;
+  };
+
   handleSnapshotNameChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     this.setState({ snapshotForm: { label: e.target.value } });
   };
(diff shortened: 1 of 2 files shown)


packages/manager/src/features/linodes/LinodesDetail/LinodeBackup/LinodeBackup.tsx
```diff
@@ -681,7 +683,11 @@ class _LinodeBackup extends React.Component<CombinedProps, State> {
           <Button
             buttonType="primary"
             onClick={this.saveSettings}
-            disabled={isReadOnly(permissions)}
+            disabled={
+              isReadOnly(permissions) ||
+              defaultDaySelection?.label === 'Choose a day' ||
```
Jskobos, line 688: This will work for new backups, but if I've already selected a window, the button will be enabled as soon as I get here, even before I make any changes. Maybe that's the intended behavior but otherwise ought to check whether the current value matches the default.

Jskobos (approved):
I'm not sure what the intended behavior is for the backups window button, that may have to be adjusted. The outline looks good though.

packages/manager/src/features/linodes/LinodesDetail/LinodeBackup/LinodeBackup.tsx
```diff
@@ -438,6 +438,20 @@ class _LinodeBackup extends React.Component<CombinedProps, State> {
     });
   };
 
+  inputHasChanged = (
+    initialValue: LinodeBackupSchedule,
+    newValue: LinodeBackupSchedule
+  ) => {
+    if (
```
Jskobos, line 445: This is good 👍 

Maybe it would make more sense to return false if either the values haven't changed, or at least one of them is 'Scheduling', since submitting the form with e.g. "Thursday" and "Scheduling" results in an error. You could make a case for this way too though, since that might be kind of confusing

packages/manager/src/features/linodes/LinodesDetail/LinodeBackup/LinodeBackup.tsx
```diff
@@ -438,6 +438,20 @@ class _LinodeBackup extends React.Component<CombinedProps, State> {
...
+      (newValue.day === 'Scheduling' || newValue.day === initialValue.day) &&
+      (newValue.window === 'Scheduling' ||
+        newValue.window === initialValue.window)
+    ) {
+      return true;
```
Jskobos, line 450: Technically you could simplify this a bit, just return the result of the conditionals above directly

DevDW (approved):

PR #7404, opened 2021-02-16
Search bar

Jskobos:
## Description

Add a search bar that uses API filtering to let you uh search for things in the Create Transfer view.


johnwcallahan (approved):
This is awesome. Really shows you how useful the caching in React Query is, i.e. it doesn't have to re-request after a search term has been used once.

johnwcallahan:
Oh, one thought. If we're going to let you search by tag, seems like we need to show tags in the table. Not sure we want to go there, though. 

I didn't think of this until now, but as a future enhancement the labels in the table rows should probably be clickable and take you to the details page in a new tab, maybe.

DevDW (approved):
This is working great! 🚢

PR #7405, opened 2021-02-17
Add event action types and basic messaging for transfer events

Jskobos:
## Description

Add some basic messaging for transfer-related events. We can eventually try to link to the transfers page with the corresponding detail modal open, but since that doesn't exist yet it just goes to the landing page.

Also adds the event actions to the JS client types. None of these are progress events so they're pretty simple.


tiffwong (approved):

johnwcallahan (approved):
Pushed up handling for `entity_transfer_cancel`, otherwise LGTM.

PR #7407, opened 2021-02-17
Change: Move token management to TransferControls

Jskobos:
The confirm transfer dialog was originally put on the Transfers landing page
but it turns out we don't need that state there. Moving the dialog and its state
management inside TransferControls makes it easier to memoize things and should
prevent some renders.


johnwcallahan (approved):
Works as advertised; thanks for grabbing this.

tiffwong (approved):

DevDW (approved):
🚢

PR #7408, opened 2021-02-17
Transfer create query

Jskobos:
## Description

Use a query for creating transfers, so that a newly created query gets added to the landing tables. There's a small bug I'm still trying to work out (see below).


packages/manager/src/features/EntityTransfers/EntityTransfersCreate/EntityTransfersCreate.tsx
```diff
@@ -58,18 +52,12 @@ export const EntityTransfersCreate: React.FC<{}> = _ => {
...
   const handleCreateTransfer = (payload: CreateTransferPayload) => {
-    setErrors(undefined);
-    setCreating(true);
-    createEntityTransfer(payload)
-      .then(transfer => {
-        // Transfer is the new transfer object; send it off to the modal.
-        setCreating(false);
+    createTransfer(payload, {
+      onSuccess: transfer => {
+        queryClient.invalidateQueries('entity-transfers');
```
Jskobos, line 57: This seems like it should do what we want, but it doesn't. To check:
1. Have a pending transfer (or several) already.
1. Navigate directly to the create page (don't use the button, or else you can reload the page). 
2. Create a transfer
3. Observe: the new transfer is the only one in the "Pending" list

Jskobos: EDIT: fixed this with help from John. All should be well.

packages/manager/src/queries/entityTransfers.ts
```diff
@@ -1,32 +1,43 @@
...
     getAllEntityTransfersRequest,
     queryPresets.longLived
   );
 };
+
+export const useTransferQuery = (token: string, enabled: boolean = true) => {
+  return useQuery<EntityTransfer, APIError[]>(
+    [queryKey, token],
+    () => getEntityTransfer(token),
+    { ...queryPresets.shortLived, enabled, retry: false }
```
johnwcallahan, line 32: re: retry; this was bothering me too. I wonder if a more comprehensive solution is to _not_ retry on 404s. Something to think about for later.

johnwcallahan (approved):

tiffwong (approved):
The pending transfers seem to display as expected 👍

DevDW (approved):
Working well 👌

PR #7410, opened 2021-02-17
M3-4905: Cancel Transfer modal

DevDW:
## Description
Add the Cancel Transfer modal with a fully functional cancellation flow.

## Type of Change


Jskobos (commented):
When a cancellation is successful, it should move to the Sent Transfers table (without needing a refresh). The way to do this is with a query, please see the PR I have open that addresses this problem for creating a transfer.

Jskobos (approved):
This is good to go as discussed offline. Will address the table updates bit later.

tiffwong (approved):
Approved pending the comment Jared made about the canceled transfers moving to the sent table

PR #7411, opened 2021-02-17
Fix: Make event progress bars always full-width

Jskobos:
## Description

This was bugging me.


ndukatz (approved):
I agree that it looks much better this way.

tiffwong (approved):
🎉

PR #7413, opened 2021-02-18
Fix: Create Transfer page mobile layout

Jskobos:
## Description

Keep the layout from breaking on tablet and smaller viewports.


packages/manager/src/features/EntityTransfers/EntityTransfersCreate/EntityTransfersCreate.tsx
```diff
@@ -18,10 +18,35 @@ import {
...
+  root: {
+    [theme.breakpoints.down('md')]: {
+      margin: 0,
+      justifyContent: 'center'
+    }
+  },
+  crumb: {
+    [theme.breakpoints.down('md')]: {
+      paddingLeft: theme.spacing()
+    }
```
tiffwong, line 34: This should only be applied to the `xs` and `md` breakpoints since the `sm` breakpoint (600 - 960px) already has a built in padding
Screen Shot 2021-02-18 at 10 29 48 AM

tiffwong (approved):
Other than the one breakpoint comment I have, everything else looks good!

PR #7416, opened 2021-02-18
Cancel Transfer modal follow-up

DevDW:
## Description
Ensure that the Transfers Pending & Transfers Sent tables update appropriately after a transfer is canceled; this is done by refreshing the `entity-transfers` query.

## Type of Change

### DIFF
--- a/packages/manager/src/features/EntityTransfers/EntityTransfersLanding/ConfirmTransferCancelDialog.tsx
+++ b/packages/manager/src/features/EntityTransfers/EntityTransfersLanding/ConfirmTransferCancelDialog.tsx
@@ -1,5 +1,3 @@
-import { cancelTransfer } from '@linode/api-v4/lib/entity-transfers';
-import { APIError } from '@linode/api-v4/lib/types';
 import { useSnackbar } from 'notistack';
 import * as React from 'react';
 import ActionsPanel from 'src/components/ActionsPanel';
@@ -8,7 +6,8 @@ import ConfirmationDialog from 'src/components/ConfirmationDialog';
 import { makeStyles } from 'src/components/core/styles';
 import Typography from 'src/components/core/Typography';
 import Notice from 'src/components/Notice';
-import { getAPIErrorOrDefault } from 'src/utilities/errorUtils';
+import { queryClient } from 'src/queries/base';
+import { useCancelTransfer } from 'src/queries/entityTransfers';
 
 const useStyles = makeStyles(() => ({
   actions: {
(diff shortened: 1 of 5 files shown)


packages/manager/src/queries/entityTransfers.ts
```diff
@@ -41,3 +47,9 @@ export const useCreateTransfer = () => {
     creationHandlers(queryKey, 'token')
   );
 };
+
+export const useCancelTransfer = () => {
+  return useMutation<{}, APIError[], string>(token => {
+    return cancelTransfer(token);
+  }, mutationHandlers(queryKey, 'token'));
```
Jskobos, line 54: I don't think you actually need the mutation handlers here. The cancelTransfer method, unlike most of our PUT requests, doesn't return anything. The mutation handler is always getting `{}`, which doesn't have a `token`, so nothing is going to happen.

It's an open question whether you want to have a separate query for this or not, you could alternatively import the `queryClient` directly in the consumer and invalidate the query in your existing .then() block. Or you could keep this, which handles the error states etc., and just not set any of these handlers.

packages/manager/src/features/EntityTransfers/EntityTransfersLanding/ConfirmTransferCancelDialog.tsx
```diff
@@ -25,42 +24,28 @@ export interface Props {
...
     // This should never happen.
     if (!token) {
       return;
     }
-    setSubmissionErrors(null);
-    setSubmitting(true);
-    cancelTransfer(token)
-      .then(() => {
+
+    cancelTransfer(token, {
```
Jskobos, line 38: You may not end up keeping this (see my comments in the query definition). If you do, you have to handle two things:

1. you need a .catch() block even if it doesn't do anything (by which I mean `.catch(_ => null)`), or else an error will be uncaught and thus reported to Sentry (I only figured this part out this morning)
2. the error state persists across modals, so if you get an error then close the dialog and open it for another transfer, the error is still there.

On balance I'd say all of this argues in favor of doing normal JS client/then/catch stuff and just calling `queryClient.invalidateQueries()` in the .then() block. But you could make an argument either way, you can go with whichever way you think works out better.

Jskobos (changes requested):

Jskobos (approved):
Looks good now, I pushed up a small change I meant to do on my PR as well but forgot about.

tiffwong (approved):

PR #7417, opened 2021-02-18
Remove hardcoded maintenance window

johnwcallahan:
## Description

This removes the hardcoded two-hour maintenance window in the helper text on Linodes Landing.

The two-hour window isn’t accurate in many cases (it could be 5-6 hours). Instead of giving the illusion of precision, this simply explains that the maintenance window opens at the specified time (which is accurate).

API has a ticket to see if the actual end time can be returned to us, so that’s a more dynamic solution for down the line.

Also did some CMR cleanup and adjusted the width of the tooltip so the date was easier to read.


ndukatz (approved):
Looks much better!

Jskobos (approved):
Changes look good. While I was checking I saw some preexisting weirdness in MaintenanceBanner.tsx, the grammar on 156 is awkward ("undergoing maintenance on 2021-01-01 8:00") and the rawDate and humanizedDate are always the same. Might as well adjust those while we're here.

DevDW (approved):

PR #7418, opened 2021-02-18
Transfers Pending/Received/Sent follow-up

DevDW:
## Description
Follow-up to https://github.com/linode/manager/pull/7395 addressing feedback:
- Expand Transfers table accordion by default only if the table is not empty
- Clean up `key`'s on `<TableCell />`'s where appropriate
- Make use of `<Hidden />`'s boolean props to simplify conditional logic
- Make `RenderTransferRow` into its own memoized component

## Type of Change

## To-Do:
- [x] Pagination


packages/manager/src/features/EntityTransfers/EntityTransfersLanding/EntityTransfersLanding.tsx
```diff
@@ -61,31 +102,40 @@ export const EntityTransfersLanding: React.FC<{}> = _ => {
         transfer={transfer}
         onClose={handleCloseSuccessDialog}
       />
-      {isLoading ? (
+      {pendingTransfersLoading ||
+      receivedTransfersLoading ||
+      sentTransfersLoading ? (
```
DevDW, line 107: The purpose of showing this loading spinner if any transfer query is loading is to prevent the situation where a populated table is collapsed because the transfersCount isn't > 0 yet because the query didn't complete

DevDW:
Still looking into fixing the pagination, not sure yet what's not connected properly

packages/manager/src/features/EntityTransfers/EntityTransfersLanding/EntityTransfersLanding.tsx
```diff
@@ -32,25 +31,67 @@ export const EntityTransfersLanding: React.FC<{}> = _ => {
...
+      status: 'received'
+    }
+  );
+  const receivedTransfers = Object.values(
+    receivedTransfersData?.entityTransfers ?? {}
   );
 
-  const pendingTransfers = allEntityTransfers?.filter(
-    transfer => transfer.status === 'pending'
+  // Fetch the Sent Transfers
```
Jskobos, line 78: This duplication and all the proppiness below is understandable. I'm wondering though if it would work to move a lot of this down to TransfersTable (not generally a great idea but it might help here) since the only thing that's going to be different between the three tables is the X-Filter.

So in the table, you'd use usePagination and useEntityTransfers, then accept a filter from the parent (can be an object like `{ is_sender: true}` or whatever. Do all the pagination and whatnot from inside the table.

It will definitely be cleaner this way, but will it work? I have no idea!

packages/manager/src/features/EntityTransfers/EntityTransfersLanding/EntityTransfersLanding.tsx
Jskobos: This is fine but if you do what I'm suggesting above you'll have to play around with an extra instance of the query (`const { isLoading } = useEntityTransfersQuery()` or something maybe) or ditch the concept altogether.

packages/manager/src/features/EntityTransfers/TransfersTable.tsx
```diff
@@ -206,14 +121,26 @@ export const TransfersTable: React.FC<CombinedProps> = props => {
...
+                    handleCancelPendingTransferClick
+                  }
+                />
+              ))}
             </TableContentWrapper>
           </TableBody>
         </Table>
       </Accordion>
-      {/* {transfersCount > pageSize ? (
+      {transfersCount > pageSize ? (
```
Jskobos, line 143: This is the misconnected wire. Need the total length of the data here (data.results).

packages/manager/src/features/EntityTransfers/TransfersTable.tsx
```diff
@@ -223,7 +150,7 @@ export const TransfersTable: React.FC<CombinedProps> = props => {
           eventCategory="Entity Transfer Table"
           fixedSize
```
Jskobos, line 151: Should this be fixedSize though? the create tables are the way they are to avoid things getting complicated in a tabbed searchable table, but we should check with Matthew to see if it's better to show or hide the page size controls here.

Jskobos (commented):
This is pretty close.

The problem is that you're using `data.entityTransfers.etc.length` to determine whether or not to display the pagination controls. (I commented inline on the actual place where this is happening). The process ends up being:
1. Request data from the API with a page_size of 25
2. Get back a list of 25 things (data.length === 25)
3. Tell TransferTable that if data.length > page_size (i.e. if 25 > 25) show the pagination controls.

You already have the information you need in `data.results`, need to grab that and use it for the comparison.

packages/manager/src/features/EntityTransfers/EntityTransfersLanding/EntityTransfersLanding.tsx
Jskobos: Or don't pass the filter and instead use the existing `transferType` to infer what the filter ought to be.

packages/manager/src/features/EntityTransfers/TransfersTable.tsx
DevDW: Checked with Matthew and John, consensus was to give it the normal treatment so I have the page size controls displaying now

packages/manager/src/features/EntityTransfers/EntityTransfersLanding/EntityTransfersLanding.tsx
DevDW: @Jskobos I tried this approach and it works, but the issue with tables that are populated being collapsed upon page load occurred again, and having three separate loading spinners for each table to get around that issue didn't look great to me. I think using the current pattern is the lesser evil compared to that, but maybe there's another way around it. I stashed my changes using that approach in case we want to revisit it.

packages/manager/src/features/EntityTransfers/EntityTransfersLanding/EntityTransfersLanding.tsx
```diff
@@ -32,64 +31,131 @@ export const EntityTransfersLanding: React.FC<{}> = _ => {
...
+  const {
+    data: receivedTransfersData,
+    isLoading: receivedTransfersLoading,
+    error: receivedTransfersError
+  } = useEntityTransfersQuery(
+    {
+      page: paginationReceivedTransfers.page,
+      page_size: paginationReceivedTransfers.pageSize,
+      handlePageChange: paginationReceivedTransfers.handlePageChange,
+      handlePageSizeChange: paginationReceivedTransfers.handlePageSizeChange
```
Jskobos, line 71: John pointed this out to me, the url this sends to the api is pretty funny: `Request URL: https://api.linode.com/v4beta/account/entity-transfers?page=1&page_size=25&handlePageChange=function+(newPage)+%7B%0A++++++++setPage(newPage)%3B%0A++++%7D` 

No need to send functions through the query. All it needs to care about is the page and the page size.

packages/manager/src/features/EntityTransfers/RenderTransferRow.tsx
```diff
@@ -0,0 +1,112 @@
...
+  }
+}));
+
+interface Props {
+  token: string;
+  created: string;
+  entities: TransferEntities;
+  expiry?: string;
+  status?: string;
+  transferTypeIsPending?: boolean;
```
Jskobos, line 32: Bit odd to have 2 separate props for this. Just pass in the type. Also these are typed as `boolean | undefined` but they're always `boolean`

packages/manager/src/features/EntityTransfers/TransfersTable.tsx
```diff
@@ -50,26 +43,36 @@ interface Props {
...
+  handlePageChange: (v: number, showSpinner?: boolean | undefined) => void;
+  handlePageSizeChange: (v: number) => void;
 }
 
 type CombinedProps = Props;
 
 export const TransfersTable: React.FC<CombinedProps> = props => {
-  const classes = useStyles();
+  const {
+    transferType,
```
Jskobos, line 57: This is a `string` but it ought to be an enum

Jskobos (changes requested):
Still needs a rebase (most of the commit history is stuff that's already present in develop from the feature branch).

packages/manager/src/features/EntityTransfers/EntityTransfersLanding/EntityTransfersLanding.tsx
johnwcallahan: Re: duplication. It might be good to combine the usePagination and useEntityTransfer queries to reduce this duplication.

packages/manager/src/features/EntityTransfers/TransfersTable.tsx
```diff
@@ -220,24 +161,35 @@ export const TransfersTable: React.FC<CombinedProps> = props => {
...
-          count={transfersCount}
-          handlePageChange={handlePageChange}
-          handleSizeChange={() => null} // Transfer tables are going to be sticky at 25
-          page={page}
-          pageSize={pageSize}
-          eventCategory="Entity Transfer Table"
-          fixedSize
-        />
-      ) : null} */}
+        {results > pageSize ? (
```
Jskobos, line 183: It's not quite as simple as moving this up to inside the `<Accordion>` but somehow these have to go away when the accordion is collapsed.

Jskobos (commented):
Pagination is working well now. Few things:

1. The pagination controls are outside the accordion, and are visible after a table is collapsed.
2. One quirk of having a global loading state is that if I change the page size, all three tables disappear for a bit even though only one of them is changing. Not a dealbreaker but there may be a way to juggle dataLastUpdated and isLoading to get around it.
3. The usePagination hook takes an optional preferenceKey argument that might be nice to use here so that each table's size is remembered.

DevDW:
I moved the pagination footer inside the accordion and partially fixed the UI issue with it, but the last row still has a chunk that looks like a partial row right below it:

Screen Shot 2021-02-25 at 1 24 08 PM

@tiffwong any insights on how to fix this?

packages/manager/src/features/EntityTransfers/EntityTransfersLanding/EntityTransfersLanding.tsx
```diff
@@ -31,23 +31,35 @@ export const EntityTransfersLanding: React.FC<{}> = _ => {
     }
   }, [location]);
 
-  const paginationPendingTransfers = usePagination();
-  const paginationReceivedTransfers = usePagination();
-  const paginationSentTransfers = usePagination();
+  const initialPage = 1;
+  const transfersTablePreferenceKey = 'entity-transfers-tables';
```
Jskobos, line 35: This could be specific to each table. Use case: I set the "received" table to page_size 50 because I'm weird; the next time I view the page, all tables will have page_size: 50.

Jskobos:
> I moved the pagination footer inside the accordion and partially fixed the UI issue with it, but the last row still has a chunk that looks like a partial row right below it:
> 
> Screen Shot 2021-02-25 at 1 24 08 PM
> 
> @tiffwong any insights on how to fix this?

As a start I'd try merging in develop now that the new pagination controls have been merged. might fix it by itself

tiffwong (approved):

Jskobos (approved):

PR #7419, opened 2021-02-19
M3 4406 stackscripts e2e

ndukatz:
There are parts of this that will need a refactor in the future (some of this could be mocked in order to make it run faster, but it will take a good amount of work to set up more accessible mocked linode utilities that could be used throughout all of the tests, so this will take some refactoring in a separate PR. The API utilites could also use some help at some point since a lot of them are basically the same).  But for now this should work. This tests deploying a linode from a custom stackscript, as well as creating a linode from a community stackscript.


packages/manager/cypress/support/api/stackscripts.ts
```diff
@@ -0,0 +1,22 @@
+import { deleteById, getAll, isTestEntity } from './common';
+
+export const getStackscripts = (page = 1) =>
+  getAll(`/linode/stackscripts?page=${page}`);
+
+export const deleteStackscriptById = (stackscriptId: number) =>
+  deleteById('linode/stackscripts', stackscriptId);
+
+export const deleteAllTestStackscripts = () => {
+  getStackscripts().then(resp => {
```
johnwcallahan, line 10: This is iterating through all 16 pages of StackScripts (most of which are community StackScripts). All you really want are your own StackScripts that start with the test prefix, so you should use an API filter like:

```
{ "+and": [ { "username": <username> }, { "label": { "+contains": <test_prefix> } } ] }
```

packages/manager/cypress/integration/stackscripts/stackscripts.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+      fbtVisible(linode.response?.body.label);
+    });
+    cy.contains('RUNNING', { timeout: 300000 });
+    cy.reload();
+    fbtVisible(`${disk} Disk`);
+    deleteAllTestLinodes();
+    deleteAllTestStackscripts();
+  });
+
+  it('use community stackscript to deploy linode', () => {
```
johnwcallahan, line 67: This test is relying on two conditions outside of our control:

1. That `m28n-init-2` is the first StackScript listed on the Community tab.
2. That `m28-init-2` has two user defined fields: ServerID and Service Secret.

Further, I don't think this test currently adds value since it basically does the same thing as deploying a Linode from your own StackScript (in the test above).

I think we could safely remove this test and maybe make the account StackScript test more robust by adding a few UDFs.

packages/manager/cypress/integration/stackscripts/stackscripts.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+    fbtVisible(ssLabel);
+    getVisible(`[data-qa-table-row="${ssLabel}"]`).within(() => {
+      fbtClick('Deploy New Linode');
+    });
+    createLinodeFromSs();
+    cy.wait('@createLinode', { timeout: 300000 }).then(linode => {
+      cy.visit(`/linodes/${linode.response?.body.id}/storage`);
+      fbtVisible(linode.response?.body.label);
+    });
+    cy.contains('RUNNING', { timeout: 300000 });
```
johnwcallahan, line 60: Do we really need to check that the Linode reaches a "Running" state and that the disk is what we put in the POST request? You could shave off ~1 min by removing these checks and instead just verifying the POST request was successful.

johnwcallahan (changes requested):

johnwcallahan (approved):

Jskobos (approved):

PR #7421, opened 2021-02-19
Bump typescript to 4.X and prettier along with it

Jskobos:
## Description

I started this bc I thought it would let us do something cool, but it turns out we still can't do that. TS4 has some nice features though, so here's a PR. Please make sure all the linters, typecheckers, etc. on your local still work, and that the app is fine.

The new prettier version defaults `trailingComma` to `es5`, which is going to add a lot of commas as files get edited (and hence passed through Prettier). We can update that setting if we want, but I kind of like it.


johnwcallahan (approved):
I like trailing commas too.

tiffwong (approved):

PR #7422, opened 2021-02-19
Marketplace Add: Q1 Partner Release

hmorris3293:
## Description
This PR adds the fakespec & svgs for our Q1 Partner release for Marketplace

## Type of Change
 * Non breaking change ('update', 'change')
 
## Note to Reviewers
Full release for prod will need to be coordinated with Marketing.


packages/manager/src/features/OneClickApps/FakeSpec.ts
```diff
@@ -1126,5 +1164,43 @@ export const oneClickApps: OCA[] = [
...
+      {
+        title: 'https://github.com/SelfhostedPro/Yacht/',
+        href: 'https://github.com/SelfhostedPro/Yacht/'
+      }
+    ],
+    logo_url: 'assets/yacht_color.svg'
+  },
+  {
+    name: 'Zabbix',
+    description: `Host multiple sites on a single server while managing apps, firewall, databases, backups, system users, cron jobs, SSL and email–  all in an intuitive interface.`,
```
Jskobos, line 1189: This is the same description as serverwand above.

packages/manager/src/features/OneClickApps/FakeSpec.ts
```diff
@@ -118,6 +118,25 @@ export const oneClickApps: OCA[] = [
     ],
     logo_url: 'assets/csgo_color.svg'
   },
+  {
+    name: 'Cyberpanel',
```
Jskobos, line 122: Needs to be capitalized.

Jskobos (changes requested):
What about Tomcat and Aberrit Node.js? There are scripts for them but no info here. Have they been pushed to the next release?

packages/manager/src/features/OneClickApps/FakeSpec.ts
hmorris3293: fixed!

packages/manager/src/features/OneClickApps/FakeSpec.ts
hmorris3293: fixed!

hmorris3293:
@Jskobos Aberrit NodeJS & Tomcat did not make it into this release, so it is just the 4 apps

Jskobos (approved):

Jskobos:
The new apps have color icons used in the card view, they should be B&W

tiffwong (approved):

PR #7425, opened 2021-02-22
Add semi-final copy for transfer acceptance confirmation text

Jskobos:


tiffwong (approved):

DevDW (approved):
Seems like something is slightly off with the linter, there are a lot of trailing commas that were added

Jskobos:
> Seems like something is slightly off with the linter, there are a lot of trailing commas that were added

That was this: https://github.com/linode/manager/pull/7421

PR #7426, opened 2021-02-22
M3-4890: Improve responsiveness for NodeBalancers and update buttons

tiffwong:
Instead of hiding a column, I decided to switch the inline actions to the ActionMenu at an earlier breakpoint. The NodeBalancer table uses the EntityTable component which makes it difficult to change the breakpoint to an earlier one

Also added some button changes for M3-4930


packages/manager/src/features/NodeBalancers/NodeBalancerConfigPanel.tsx
```diff
@@ -25,85 +25,20 @@ import { getErrorMap } from 'src/utilities/errorUtils';
...
-  | 'passiveChecks';
+type ClassNames = 'divider' | 'passiveChecks' | 'actionsPanel';
 
 const styles = (theme: Theme) =>
   createStyles({
     divider: {
       marginTop: theme.spacing(1),
       marginBottom: theme.spacing(1)
     },
-    backendIPAction: {
```
tiffwong, line 45: These were all unused so I removed them

Jskobos (approved):
Pushed up a semi-related fix. The table looks good across all sizes

DevDW (approved):

DevDW:
I'm guessing somewhere there was a change in linter preferences somehow, trailing commas were added in here also 🤔

PR #7427, opened 2021-02-23
Hide Transfers behind a flag

Jskobos:


packages/manager/src/features/Account/AccountLanding.tsx
```diff
@@ -53,6 +58,8 @@ const AccountLanding: React.FC<Props> = props => {
     props.history.push(tabs[index].routeName);
   };
 
+  let idx = 0;
```
Jskobos, line 61: Did this to avoid magic number stuff, maybe the cure is worse than the disease tho

tiffwong (approved):

johnwcallahan (approved):

PR #7430, opened 2021-02-23
M3-4943 Change: Selecting a Linode auto-selects the region when creating a Volume

Jskobos:
## Description

Instead of adding plumbing to work with the /Volumes/LinodeSelect component that's only really used here, I used src/features/linodes/LinodeSelect instead and adjusted things to make it work out.

### DIFF
--- a/packages/manager/src/components/EnhancedSelect/variants/RegionSelect/RegionSelect.tsx
+++ b/packages/manager/src/components/EnhancedSelect/variants/RegionSelect/RegionSelect.tsx
@@ -9,7 +9,6 @@ import SG from 'flag-icon-css/flags/4x3/sg.svg';
 import US from 'flag-icon-css/flags/4x3/us.svg';
 import { groupBy } from 'ramda';
 import * as React from 'react';
-import { compose } from 'recompose';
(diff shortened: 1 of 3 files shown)


packages/manager/src/components/EnhancedSelect/variants/RegionSelect/RegionSelect.tsx
```diff
@@ -177,13 +178,17 @@ const SelectRegionPanel: React.FC<Props> = props => {
   } = props;
 
   const onChange = React.useCallback(
-    (selection: RegionItem) => {
+    (selection: RegionItem | null) => {
+      if (selection === null) {
+        handleSelection('');
```
Jskobos, line 183: I don't like this `''` but it seems to be the convention we've settled on for these region selects as a default value 🤷

DevDW (approved):

tiffwong (approved):
This is great! Would it make more sense to put the Linode select before the Region select?

Jskobos:
> This is great! Would it make more sense to put the Linode select before the Region select?

Maybe, the Linode field is optional though. The form is pretty old, we could put it on the radar for UX review.

PR #7432, opened 2021-02-23
M3-4945: Align "Docs" and "Create" button

tiffwong:
## Description

Moved the DocumentationButton component down 2px so it's aligned with the "Create" button text

Also replaced the component with the CMR version


packages/manager/src/components/DocumentationButton/DocumentationButton.tsx
```diff
@@ -1,32 +1,28 @@
...
 const useStyles = makeStyles((theme: Theme) => ({
   root: {
-    padding: `0 ${theme.spacing(1) + 4}px`,
-    marginBottom: 0,
+    alignItems: 'center',
+    fontFamily: theme.font.normal,
+    fontSize: '.875rem',
     lineHeight: 'normal',
+    margin: 0,
+    marginTop: 2,
```
Jskobos, line 13: Is this line the magic one? hard to tell with these linter diffs

Jskobos (approved):
I think I can tell the difference?

packages/manager/src/components/DocumentationButton/DocumentationButton.tsx
tiffwong: Yup 😓

DevDW (approved):

PR #7433, opened 2021-02-23
M3-4940 Change: Block device assignment form

Jskobos:
## Description

In the create/edit config dialog, update the block assignment
logic to match what we do with Rescue mode: only show the currently
selected/assigned devices plus one empty "None" slot. Add a button
to increment the number of devices (up to the limit).


DevDW (approved):
LGTM. Should we be scrolling to the error when there is one? (if so, we don't seem to be doing so currently)

tiffwong (approved):

PR #7436, opened 2021-02-24
M3-4909: Implement new pagination controls

tiffwong:
**Before:**
Screen Shot 2021-02-24 at 10 51 37 AM

**After:**
Screen Shot 2021-02-24 at 10 51 00 AM


packages/manager/src/App.tsx
```diff
@@ -297,7 +297,7 @@ const mapStateToProps: MapState<StateProps, Props> = state => ({
   nodeBalancersError: path(['read'], state.__resources.nodeBalancers.error),
   appIsLoading: state.initialLoad.appIsLoading,
   featureFlagsLoading: state.featureFlagsLoad.featureFlagsLoading,
-  euuid: state.__resources.profile.data?._euuidFromHttpHeader
+  euuid: state.__resources.profile.data?._euuidFromHttpHeader,
```
ndukatz, line 300: I've noticed that a lot of these changes are either adding commas or parenthesis. I just want to make sure that was intentional because they don't all appear to be necessary.

ndukatz (approved):
Looks much better and works well. Some of the changes not related to this PR look to have been changed to add commas and parenthesis in a lot of places. I made a comment under one of the examples above. I just want to make sure those are intentional. Other than that, looks great!

Jskobos (approved):
Controls look good, tests are much better. On light mode the page size input matches the background color of the Paper, but on dark mode it's a bit different:
Screen Shot 2021-02-25 at 1 47 43 PM

Not sure if this is intentional/a necessary change.

packages/manager/src/App.tsx
tiffwong: These are courtesy of the new eslint updates

DevDW (approved):

PR #7437, opened 2021-02-24
Explicitly read transfer.expiry in UTC

Jskobos:
## Description

I think that because API dates don't include an explicit timezone, fromISO() was inferring an incorrect zone. To see this:

1. Create a new transfer
2. From another account, plug the token into the Receive Transfer box
3. Observe: the message says it will expire in >24 hours (I was seeing 28).

### DIFF
--- a/packages/manager/src/features/EntityTransfers/EntityTransfersLanding/ConfirmTransferDialog.tsx
+++ b/packages/manager/src/features/EntityTransfers/EntityTransfersLanding/ConfirmTransferDialog.tsx
@@ -268,7 +268,8 @@ export const getTimeRemaining = (time?: string) => {
   }
 
   const minutesRemaining = Math.floor(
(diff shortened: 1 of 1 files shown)


DevDW (approved):

packages/manager/src/features/EntityTransfers/EntityTransfersLanding/ConfirmTransferDialog.tsx
```diff
@@ -268,7 +268,8 @@ export const getTimeRemaining = (time?: string) => {
   }
 
   const minutesRemaining = Math.floor(
-    DateTime.fromISO(time).diffNow('minutes').toObject().minutes ?? 0
+    DateTime.fromISO(time, { zone: 'UTC' }).diffNow('minutes').toObject()
```
johnwcallahan, line 271: I believe you could use `parseAPIDate` for this if you wanted.

johnwcallahan (approved):

PR #7439, opened 2021-02-24
Update types and factories for VLAN Adjustments

DevDW:
## Description
Update the types and factories as per the new VLAN API adjustments.

## Type of Change]


DevDW:
Going ahead and removing the WIP label.

With regards to the hard-coded IDs in `linodeConfigs.ts` and `serverHandlers.ts`, I tried using `Factory.each((i) => i)` but it was either causing a TS error or mismatching the data in such a way that the Network Interfaces column was empty

packages/manager/src/factories/linodeInterfaces.ts
```diff
@@ -0,0 +1,14 @@
+import * as Factory from 'factory.ts';
+import { LinodeInterface } from '@linode/api-v4/lib/linodes/types';
+
+export const linodeInterfacesFactory = Factory.Sync.makeFactory<LinodeInterface>(
```
Jskobos, line 4: All of these endpoints are going away, I don't think it's necessary to add a factory for them. (related, all JS client types and methods for /interfaces requests should be deleted as part of this PR).

packages/api-v4/src/linodes/types.ts
```diff
@@ -147,9 +147,15 @@ export type LinodeStatus =
   | 'stopped';
 
 export type InterfaceSlot = Record<string, InterfaceBody>; // e.g. { "eth0": { "id": 111 }}
+export type InterfacePurpose = 'public' | 'vlan' | 'internal' | 'multivlan';
```
Jskobos, line 150: internal and multivlan are not available for customers to use, they probably shouldn't be included or documented here.

packages/api-v4/src/linodes/types.ts
```diff
@@ -147,9 +147,15 @@ export type LinodeStatus =
   | 'stopped';
 
 export type InterfaceSlot = Record<string, InterfaceBody>; // e.g. { "eth0": { "id": 111 }}
```
Jskobos, line 149: This comment is no longer correct

packages/manager/src/factories/linodeConfigs.ts
```diff
@@ -41,5 +43,22 @@ export const linodeConfigFactory = Factory.Sync.makeFactory<Config>({
     sde: null,
   },
   kernel: 'linode/grub2',
-  interfaces: {},
+  interfaces: {
+    eth0: {
+      id: 3,
+      devum: 1,
```
Jskobos, line 49: This is an internal field

packages/manager/src/factories/linodeConfigs.ts
```diff
@@ -41,5 +43,22 @@ export const linodeConfigFactory = Factory.Sync.makeFactory<Config>({
...
+      label: 'Test eth0',
+      purpose: 'public',
+      ipam_address: '',
+      create_dt: createDate,
```
Jskobos, line 53: this is a db field

packages/manager/src/factories/linodeConfigs.ts
```diff
@@ -41,5 +43,22 @@ export const linodeConfigFactory = Factory.Sync.makeFactory<Config>({
     sde: null,
   },
   kernel: 'linode/grub2',
-  interfaces: {},
+  interfaces: {
+    eth0: {
+      id: 3,
```
Jskobos, line 48: this will not come back from the API

packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -195,6 +196,17 @@ export const handlers = [
     const ips = linodeIPFactory.build();
     return res(ctx.json(ips));
   }),
+  rest.get('*/instances/*/interfaces', async (req, res, ctx) => {
```
Jskobos, line 199: Remove please

packages/api-v4/src/linodes/types.ts
Jskobos: (and InterfaceSlot is a bad name for the new pattern; and it's an array not a Record)

packages/manager/src/factories/linodeConfigs.ts
Jskobos: (it's also "devnum")

Jskobos (changes requested):
- The payload for creating a Linode is different in the new system, the types and schema will have to be updated
- It will no longer be possible to PUT/POST/DELETE at networking/vlans, those should be removed
- Interface payloads and response values only have 3 fields: label, purpose, and ipam_address. Everything else is internal and not returned by the API
- Config schemas will need to be updated

packages/manager/src/factories/linodeConfigs.ts
Jskobos: the shape needs to be updated as well, it's no longer { eth0: {}...etc } it's an array.

packages/api-v4/src/linodes/linodes.schema.ts
```diff
@@ -7,18 +7,24 @@ const stackscript_data = array()
...
+    .max(64, 'Label must be between 1 and 64 characters.')
+    .matches(
+      /[a-z0-9-]+/,
+      'Interface labels cannot contain special characters.'
+    ),
+    purpose: mixed()
+      .oneOf([null, 'public', 'vlan'], 'Purpose must be null, public, or vlan.'),
+    ipam_address: string()
+      .matches(
+        /^\d{3}\.\d{3}\.\d{1,3}\.\d{1,3}$/,
```
Jskobos, line 25: The spec says the ipam_address has to be a range, so this won't work. It's also a risky way to validate IP addresses. I think the ipaddr.js library is already part of the JS client build, you can use ipaddr.parseCIDR here (I think, see firewalls.schema.ts for examples).

Jskobos (approved):

johnwcallahan (approved):

PR #7440, opened 2021-02-25
M3-4937 Change: Refactor LinodeConfigs drawer to an FC that uses Formik

Jskobos:
## Description

Got started adding the Network Interfaces section here and decided to make life more difficult for myself, so here we are. Functionality should be unchanged (except for the empty interfaces section I'll add to later).

### DIFF
--- a/packages/manager/.prettierrc
+++ b/packages/manager/.prettierrc
@@ -1,3 +1,4 @@
 {
-  "singleQuote": true
+  "singleQuote": true,
+  "trailingComma": "es5"
 }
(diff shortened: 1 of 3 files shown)


DevDW (approved):
Looks good, just a couple of points:

1. It seems like the error notice can be moved down slightly, lots of blank space below it that I don't think we do elsewhere:

Screen Shot 2021-02-26 at 12 31 09 PM

2. The error isn't scrolling into view for me (tried in Chrome and Firefox)

tiffwong (commented):
On top of decreasing the spacing between the error and label that Dajahi mentioned, can we align the red bar to the text and increase the width of the label error message?

Screen Shot 2021-02-26 at 2 15 28 PM

ndukatz (approved):
I am not seeing the same issues as the other two screenshots - not sure why. I tested in Chrome and Safari. All UI E2E tests passed on this branch as well. 
image

PR #7442, opened 2021-02-26
Run prettier with new config

Jskobos:
## Description

I had to commit this with `--no-verify`, because running the linter across the entire project turns up a bunch of errors with the new config that aren't trivial to resolve. I fixed the trivial ones but it's better to handle this on a case by case basis.

One drawback of this approach is that we're going to see another era of the "Format all the (other) things" in git blame and history.


tiffwong (approved):
Everything I checked looks good to me! 🚢

DevDW (approved):
I did some spot checking of unit tests, parts of the app, and types to ensure type checking is working as expected, and everything LGTM

ndukatz (approved):
Test failure is due to changes unrelated to this pr.

johnwcallahan (approved):

PR #7443, opened 2021-02-26
Source maintenance information from /maintenance endpoint

johnwcallahan:
## Description

Use the `/account/maintenance` endpoint instead of notifications to populate the maintenance display for Linodes.

The UI has not changed. The main changes:

- Add /account/maintenance endpoint and types to the JS client.
- Add accountMaintenance query.
- Add wrapper around LinodesLanding so we could use the new accountMaintenance query and other hooks.
- Add `addMaintenanceToLinodes` function (corresponds to the existing `addNotificationsToLinodes` function).
- Update mocks and tests.

Todo:
- [x] Handle Linode Details banner

## Note to Reviewers

I added mocks, so run with those enabled to test.


Jskobos (commented):
The query, client stuff, types, etc. all looks good. Few questions:

1. My test account has a bunch of pending maintenance (for 3/21) for a bunch of Linodes that no longer exist. I don't think this'll affect your PR because you're matching on entity.type and entity.id, but it seems weird?
2. What about the notification section in the drawer? It's hard to see how this will work just with mocks, but there might be incongruities if we're sourcing maintenance-related information from two different places. I remember Nick saying that you definitely still get an `important_ticket` notification in these scenarios, which seems ok.
3. I'm not sure if this is mocking strangeness or if we're going to end up overkilling the banners in the real world:

Screen Shot 2021-03-01 at 8 49 06 AM

johnwcallahan:
> My test account has a bunch of pending maintenance 

That's an API bug; fix going out with the next release (like you mentioned it shouldn't affect Cloud since we're matching IDs). 

> What about the notification section in the drawer?

I thought about adding a "fake" notification to the drawer based on the presence of maintenance items. I decided against it since there will be an "Important Ticket" notification that links them to a ticket with all the details. If that's acceptable, we can take another look at this when we add the dedicated maintenance page. I figure we might want a notification that links there.

> I'm not sure if this is mocking strangeness... 

This was a bug... since I removed handling for `maintenance` notifications from LinodesDetailHeader/Notifications.tsx, it fell back to rendering a `<ProductNotification />`. Pushed up a change to address this.

ndukatz (commented):
It all looks good to me. I am just curious what the difference between what triggers "This Linode's physical host will be undergoing maintenance" and status as "Maintenance Scheduled" vs what only triggers "This Linode will be affected by critical maintenance!" and keeps the status as is.

ndukatz (approved):

Jskobos (approved):

DevDW (approved):

PR #7444, opened 2021-03-01
Fix: Account balance display

Jskobos:
Recent changes to how we display past due balances lead to confusion
for some customers. Specifically, a balance due was being hidden
during the first 3 calendar days of the month. This is a problem because:

- Some customers have a custom grace period much longer than 3 days
- Customers with old balances were seeing them disappear

Solution (for now!):

- No longer hide the balance due during the first 3 days of the month
  This is a temporary fix, the balance will still be displayed in red
  and as "Past Due." Customers with longstanding balances were seeing
  this disappear at the beginning of the month.
- Use the notification past_due instead of hiding it and faking a similar one.
  For now, we'll still hide this for the first three days of the month as a
  stopgap.
- Use the natural severity (Major) of this notification rather than making
  it artificially Critical

### Note to Reviewers

Sorry about the commas. Test with the mocks, and reach out for login info for accounts that are currently affected by this.

### DIFF
--- a/packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
+++ b/packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
@@ -1,5 +1,4 @@
 import { ActivePromotion } from '@linode/api-v4/lib/account/types';
-import { DateTime } from 'luxon';
 import * as React from 'react';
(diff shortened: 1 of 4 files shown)


tiffwong (approved):

DevDW (approved):
🚢

johnwcallahan (approved):

PR #7445, opened 2021-03-01
Fix: Received Transfers table update

Jskobos:
When receiving a transfer, the table is not updated immediately
after submitting the confirmation form. In addition, the X-Filtering
for the tables is incorrect.

~There seem to be some API problems related to filtering that caused
us to not notice this, and this PR won't actually work as written.
It should work, though, when the API is fixed.~

### Note to reviewers

The API fix has been merged to development, you should be able to test this PR using dev services.

### DIFF
--- a/packages/manager/src/features/EntityTransfers/EntityTransfersLanding/EntityTransfersLanding.tsx
+++ b/packages/manager/src/features/EntityTransfers/EntityTransfersLanding/EntityTransfersLanding.tsx
@@ -3,7 +3,10 @@ import * as React from 'react';
 import { useHistory, useLocation } from 'react-router-dom';
 import { DocumentTitleSegment } from 'src/components/DocumentTitle';
 import usePagination from 'src/hooks/usePagination';
-import { useEntityTransfersQuery } from 'src/queries/entityTransfers';
+import {
+  TRANSFER_FILTERS,
+  useEntityTransfersQuery,
+} from 'src/queries/entityTransfers';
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/EntityTransfers/EntityTransfersLanding/EntityTransfersLanding.tsx
```diff
@@ -81,12 +81,13 @@ export const EntityTransfersLanding: React.FC<{}> = (_) => {
       page_size: paginationReceivedTransfers.pageSize,
     },
     {
-      status: 'received',
+      '+and': [{ is_sender: false }, { status: { '+neq': 'pending' } }],
```
Jskobos, line 84: This should have triggered an error if the API was working correctly, since "received" is not a valid status for a transfer.

Jskobos: Update: filtering on nonexistent enum types doesn't actually cause an error, but the API team is going to revisit this and see if it ought to be the expected behavior.

johnwcallahan (approved):

DevDW (approved):
Confirmed that the Transfers Received table now gets updated upon a successful receipt 👍 

The only thing I noticed was that in the case where the Transfers Received table was previously unpopulated, the table stays collapsed after the first transfer received gets added to it

PR #7448, opened 2021-03-01
M3-4938: Add network interface information to Configurations table

DevDW:
## Description
Removes some old VLANs logic from the previous feature iteration; adds Config interfaces mocks; tweaks Network Interfaces column to show VLAN info if the interface is connected to a VLAN.

## Type of Change

## Note to Reviewers
Please use the Mock Service Worker as the mocks are updated with the new data shape for Linode Config interfaces

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeAdvanced/ConfigRow.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeAdvanced/ConfigRow.tsx
@@ -4,7 +4,6 @@ import { Volume } from '@linode/api-v4/lib/volumes';
 import { isEmpty } from 'ramda';
 import * as React from 'react';
 import { makeStyles } from 'src/components/core/styles';
-import { Link } from 'src/components/Link';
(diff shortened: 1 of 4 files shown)


packages/manager/src/features/linodes/LinodesDetail/LinodeAdvanced/ConfigRow.tsx
```diff
@@ -180,21 +163,16 @@ export const ConfigRow: React.FC<CombinedProps> = (props) => {
...
 
-  const vlan = vlans[linodeInterface.vlan_id];
-
-  // I don't think this would ever happen, but here's a fallback anyway.
-  if (!vlan) {
-    return linodeInterface.description;
-  }
+  const interfaceLabel = configInterface.label;
+  const ipamAddress = configInterface.ipam_address;
+  const hasIPAM = !isEmpty(ipamAddress);
```
Jskobos, line 175: `isEmpty` will work here probably but it's intended for arrays and objects. You can do a Boolean check, that should work just as well.

Jskobos (approved):

ndukatz (approved):

PR #7449, opened 2021-03-02
M3-4927 Add: Action input to firewall rule drawer and table

Jskobos:
## Description

Keeping this small because I know @johnwcallahan is also messing with the Firewall rules table. The next step will be to add a default policy display/dropdown to each table.

## Note to Reviewers

The API stuff hasn't been merged yet, so please use the mocks to test.


tiffwong (commented):
Using dev to test and I get this error when I try to add a rule 
Screen Shot 2021-03-02 at 5 10 50 PM

Jskobos:
> Using dev to test and I get this error when I try to add a rule
> Screen Shot 2021-03-02 at 5 10 50 PM

That's expected, the backend isn't ready to accept the updated shape yet.

johnwcallahan (approved):
To my eye there should be an additional 8px of margin between "Sources" and the new radio button. 

It's the only input without a label; maybe we should contextualize it somehow?

DevDW (approved):
Looks good to go 🚢

PR #7451, opened 2021-03-03
Allow saving reordered FW rules

johnwcallahan:
## Description

This PR allows saving of FW rules that have been reordered. 

I went back and forth with how to represent these modifications in our data structure. The challenge is that:

- Each rule is represented as list of "revisions"!
- These lists of revisions are themselves stored in an array. The order of the array corresponds to the order of rules.
- So, there's no good way to indicate that the order has been modified without pushing a new revision to each rule stack.

~I opted to change the reducer state to~:

```ts
export interface RuleEditorState {
  revisionLists: ExtendedFirewallRule[][]; // <-- existing reducer state
  hasModifiedOrder: boolean;
}
```

~When the order is modified, **no new revisions are pushed to the rule stacks.** Instead, `revisionLists` is reordered and we set `hasModifiedOrder` to `true`.~

**EDIT: ** this was changed to @Jskobos's much better idea.

We also capture the original index of each rule when the state is initialized so that we can "Discard Changes" and return to the original order.

## Note to Reviewers

You'll likely run into API issues trying to save rules; one way to solve is to drop some code into the JS client to delete the new keys the API isn't expecting (inbound_policy, outbound_policy, and on rules: action).

Please beat this up. Add rules, delete them, modify them, reorder them, etc. etc.

### DIFF
--- a/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.test.tsx
+++ b/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.test.tsx
@@ -106,6 +106,7 @@ describe('utilities', () => {
       protocol: 'TCP',
       status: 'NEW',
       action: 'ACCEPT',
+      originalIndex: 0,
       addresses: {
         ipv4: ['1.2.3.4'],
         ipv6: ['::0'],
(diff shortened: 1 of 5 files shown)


Jskobos (approved):
Looks fine, I haven't been able to break it. Some small thoughts:

- Instead of `hasModifiedOrder`, what about `hasModified` returns `thisRule.status !== "NOT_MODIFIED" || idx !== originalIndex`? That would help with two tiny things: 1) if you drag the rules back to the initial order there's no indication that there are no longer pending changes; 2) the drag-drop is so smooth there's no feedback when you change the order. If all rules are highlighted if their index !== originalIndex that might be a better experience.
- Undo on a rule that has been e.g. created and then reordered might be a bit counterintuitive to users...the rule goes away when maybe it should have moved to its original position. I don't think this is a problem and there isn't much we can do about it, just something to keep an eye on for feedback.

johnwcallahan:
@Jskobos 🤦  great suggestion, pushed.

DevDW (approved):
This is working well 🚢

PR #7452, opened 2021-03-03
M3-4964: Style reorder-able Firewall Rules

tiffwong:
Add styles for dragging Firewall Rules and also removed the old `TableEditor` component and draggable icon that's not being used


Jskobos (approved):
looks slick

johnwcallahan (approved):

PR #7454, opened 2021-03-04
M3-4929: Enforce masks for FW Rule IPs

DevDW:
## Description
When adding or editing a FW rule with IP / Netmask sources, enforce masks by appending `/32` for IPv4 addresses and `/128` for IPv6 addresses if one has not been provided by the user.

## Type of Change

## To-Do:
- [x] ~~More stringent and precise logic in `enforceIPMasks()`~~
- [x] ~~Possibly have a faded, placeholder text-style appearance for the appended IP mask~~
- [x] Cleanup


johnwcallahan (changes requested):
This is close, nice work. 

- From a UX perspective, the new mask logic should be run `onBlur` vs. `onChange`. This fixes several bugs that you make note of in the comments. To do this you'll need to create "handleBlur" functions in MultipleIPInput and FirewallRuleDrawer.
- When you run this logic on blur, you'll no longer need to check the minimum length. 
- There are a few eslint warnings (see screenshots).
- You don't need to check `mask?.length > 1` (which should use >= anyway)... you can just check `if (mask) {}`. 


Screen Shot 2021-03-04 at 8 56 15 AM

Screen Shot 2021-03-04 at 8 56 25 AM

packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.tsx
```diff
@@ -832,3 +843,28 @@ export const validateForm = (
...
+      // The user provided a mask already
+      return extendedIP;
+    }
+
+    try {
+      const parsed = parseIP(base);
+      const type = parsed.kind();
+
+      const appendedMask = type === 'ipv4' ? '/32' : '/128';
+      const ipWithMask = ipAddress + appendedMask;
```
johnwcallahan, line 863: Maybe this should be `base + appendedMask`. That solves a few oddities like what happens when entering "10.0.0.1//".

packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.tsx
```diff
@@ -832,3 +843,28 @@ export const validateForm = (
 
   return errors;
 };
+
+export const enforceIPMasks = (ips: ExtendedIP[]) => {
```
johnwcallahan, line 847: Still need a return type here.
Screen Shot 2021-03-04 at 1 48 34 PM

packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.tsx
```diff
@@ -362,6 +362,12 @@ const FirewallRuleForm: React.FC<FirewallRuleFormProps> = React.memo(
       [formTouched, setIPs]
     );
 
+    const handleOnBlurIP = (_ips: ExtendedIP[]) => {
```
johnwcallahan, line 365: Minor: to match naming convention this should probably be called `handleIPBlur`.

packages/manager/src/components/MultipleIPInput/MultipleIPInput.tsx
```diff
@@ -70,6 +81,17 @@ export const MultipleIPInput: React.FC<Props> = (props) => {
     onChange(newIPs);
   };
 
+  const handleBlur = (e: any, idx: number) => {
```
johnwcallahan, line 84: We can do better than `any`! 🙂 

Hint: hover over `e` on L155 to see what the type is:
Screen Shot 2021-03-04 at 2 22 46 PM

packages/manager/src/components/MultipleIPInput/MultipleIPInput.tsx
```diff
@@ -70,6 +81,17 @@ export const MultipleIPInput: React.FC<Props> = (props) => {
...
+    const newIPs = [...ips];
```
johnwcallahan, line 85: Very minor:

There's a condition here where we `return` nothing. It's ever-so-slightly more readable if you make that clear up front to the reader of this code, so they don't waste time looking at L85 and 86.

```ts
    if (!onBlur) {
      return;
    }

    const newIPs = [...ips];
    newIPs[idx].address = e.target.value;
    onBlur(newIPs);
```

johnwcallahan (changes requested):
Thanks for the adjustments. Very close, just two things that should be addressed (missing return type and unnecessary `any` type), one question/suggestion, and a few minor thoughts.

Jskobos (commented):
There are some small parallel adjustments that should be made to reflect the fact that only ranges are allowed:

1. On line 49 of the drawer, the IP_ERROR_MESSAGE should be updated (it says "address or range" which is no longer kosher)
2. The validateIPs helper function maybe doesn't need to do the whole checking for a mask thing and trying parseIP if there's no mask. It wouldn't hurt to leave it in I guess, though at least the comments in there should be updated.
3. Unless Matthew has advised against it, I would still prefer to have a placeholder inside the IP inputs (192.0.2.1/32 or something else from inside the reserved range for documentation).

packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.tsx
DevDW: Thanks, it looks like my ESLint had stopped working for some reason so I wasn't seeing these problems indicated. Updated VSC and things are back in order. 👍  Added a return type

packages/manager/src/components/MultipleIPInput/MultipleIPInput.tsx
```diff
@@ -81,17 +81,24 @@ export const MultipleIPInput: React.FC<Props> = (props) => {
...
+    if (!onBlur) {
       return;
     }
+
+    onBlur(newIPs);
   };
 
+  // We want the placeholder to show in the IP fields in FirewallRuleDrawer.
+  // Presently that is the only place that passes the onBlur prop to this component.
+  const ipFieldPlaceholder = onBlur ? '192.0.2.1/32' : '';
```
Jskobos, line 100: Slightly more maintainable to pass an optional placeholder prop from the consumers of this component and default to '' if it's not passed. You're implying a relationship between blur handlers and placeholders that happens to hold for this particular case but isn't an actual relationship.

packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.tsx
```diff
@@ -611,15 +610,10 @@ export const validateIPs = (
     if (!options?.allowEmptyAddress && !address) {
       return { address, error: 'Please enter an IP address.' };
     }
-    // We accept plain IPs as well as ranges (i.e. CIDR notation). Ipaddr.js has separate parsing
-    // methods for each, so we check for a netmask to decide the method to use.
-    const [, mask] = address.split('/');
+    // We accept IP ranges (i.e., CIDR notation). By the time this function is ran,
```
Jskobos, line 613: *run

Jskobos (approved):
looks good 👍

johnwcallahan (approved):

PR #7457, opened 2021-03-04
M3-4965: Enable keyboard shortcuts with Firewall Rules

tiffwong:
Added focus states to the rows and allow keyboard dragging by pressing the spacebar on the selected row to pick up and drop it in a new position (use up and down arrows to move the row)


Jskobos (approved):
When I spacebar again to "drop" the row in its new position, some other row ends up with the focus state. It's not a big deal but it's a little jarring.

tiffwong:
@Jskobos do you mean this?
Screen Shot 2021-03-04 at 5 47 15 PM

If so, the lighter blue row (second row in this case) is our styling for when a change has been made and the darker blue is the actual focus state. I tried to make the focus state an outline but didn't have any luck but if that makes more sense, I can look into it some more!

ndukatz (approved):
Works well! I see what Jared means, but I also see why the focus state would remain on the original row.

johnwcallahan (approved):

PR #7458, opened 2021-03-04
Firewalls feedback

Jskobos:
## Description

- fix extra column in “Outbound” table
- add label to default rules
- fix “Add IP” button style
- hide “Description” column,
- Add sentence explaining a Linode can only be added to FW at a time
- Fixed an unrelated styling bug in TextField error text displays.


packages/manager/src/components/TextField.tsx
```diff
@@ -107,6 +107,7 @@ const styles = (theme: Theme) =>
       },
     },
     errorTextLong: {
+      width: '100%',
```
Jskobos, line 110: I can't find anywhere that is broken by this but it seems risky. Definitely a bug as is though, try attaching a Linode to a bunch of Firewalls in the Create Firewall drawer.

DevDW (approved):
LGTM

As a minor point, in the Add a Rule drawer, if you end up selecting "Drop" for the action before a preset, the default label will still have "accept" in it. Might be a nice touch if the default label reflected the selection of "Drop" (i.e., it'd have "drop" instead of "accept" in the default label)

tiffwong (approved):
Looks great!

PR #7460, opened 2021-03-04
Adjusting the column widths for Entity Transfer tables

tiffwong:


packages/manager/src/features/EntityTransfers/RenderTransferRow.tsx
```diff
@@ -53,25 +80,46 @@ export const RenderTransferRow: React.FC<CombinedProps> = (props) => {
...
           className={classes.link}
           onClick={() => handleTokenClick(token, entities)}
         >
           {token}
         </button>
       </TableCell>
-      <Hidden smDown>
-        <TableCell className={classes.cellContents}>
+      {transferTypeIsPending || transferTypeIsSent ? (
+        <Hidden smDown>
```
Jskobos, line 100: I think `smDown={<that conditional>}` does the same thing without having to put the cell in the markup twice.

Jskobos (approved):
Looks good

johnwcallahan (approved):

PR #7461, opened 2021-03-04
Add validation package

Jskobos:
## Description

Laying the groundwork for a change we've discussed for a long time but haven't moved on. The idea is to add a new package to our workspace, @linode/validation, that will contain all of our Yup schemas. I'll explain this in detail in the forthcoming README.md, but tldr;

- It's not possible to turn client-side validation off, which is a weird choice for a public package
- The validation doesn't match the API's exactly (we can't validate uniqueness and some things we've decided are too complicated and passed them off to the backend)
- We're double-validating in many cases (whenever we use Formik's validationSchema option); the same schema is run on form submit by Formik, then run again in the JS client.
- The JS client bundle is a bit too big, and this is a big part of it, which is a shame given that not everyone necessarily wants to run client validation.

We'll probably want to organize some of the files by directory but for now I just copied them all over to see if the config files would work.

## Note to Reviewers

Run build (test won't work) to make sure it builds; I imported these files in api-v4/src/account.ts as a POC.

Future steps include:
-  Removing all schemas from the JS client (temporarily just replacing them with imports from /validation)
- Adding an options argument to all the client methods (lot of typing 😢 ) so that consumers can specify a schema if they want to have it applied
- Replacing Manager imports from api-v4/schema with validation/schema
- Testing etc.

Note that absolutely nothing should change as a result of this PR, in either Manager or the JS client.


packages/validation/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "@linode/validation",
+  "version": "0.1.0",
+  "description": "Yup validation schemas for use with the Linode APIv4",
```
DevDW, line 4: "Yup validation schemas for use with ~~the~~ Linode APIv4"

tiffwong (approved):
I was able to run yarn build successfully 👍

DevDW (approved):
I'm able to get localhost up after running `serve -p 3000 packages/manager/build` or `npx serve packages/manager/build` 👌 . The configurations look good to me aside from the idea I raised earlier about adding `forceConsistentCasingInFileNames` to tsconfig

johnwcallahan (approved):

PR #7465, opened 2021-03-05
Remove usernames from all entity transfer events

Jskobos:
## Description

We can add some of these back when we've worked out the details with the API.


tiffwong (approved):

ndukatz (approved):
Seems to work well

PR #7469, opened 2021-03-08
Display some text on empty state in transfer checkout bar

Jskobos:
## Description

Step 1 of making the empty state of the make transfers page less weird, particularly when using Dark theme


ndukatz (approved):
Looks better already

DevDW (approved):

PR #7471, opened 2021-03-08
Add: queries and types for StatusPage.io API

Jskobos:
## Description

Add queries for maintenance events and incidents. We may need different specific requests (the ones I've chosen are for all active maintenance and all unresolved incidents, which seems to match our needs) but that should be a pretty simple adjustment, either changing URLs or copy-pasting to get new queries.


packages/manager/src/queries/statusPage/types.ts
```diff
@@ -0,0 +1,72 @@
...
+  status: IncidentStatus;
+  created_at: string;
+  updated_at: string;
+  monitoring_at: string;
+  resolved_at: string;
+  impact: IncidentImpact;
+  shortlink: string;
+  started_at: string;
+  page_id: string;
+  incident_updates: any[];
```
DevDW, line 61: `IncidentUpdate[]` instead of `any[]`?

DevDW (approved):
The types, RQ, and request error handling look good to me

packages/manager/src/queries/statusPage/types.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+export interface Incident {
+  id: string;
+  name: string;
+  status: IncidentStatus;
+  created_at: string;
+  updated_at: string;
+  monitoring_at: string;
+  resolved_at: string;
```
johnwcallahan, line 56: Same here (can be `null`).

johnwcallahan (approved):
New code looks good.

The lack of documentation around types doesn't exactly fill me with confidence... when we do the implementation we should be as defensive as possible in case we get something back we don't expect.

PR #7472, opened 2021-03-08
UI global updates

tiffwong:
List of changes:
- Change hover state for table headers
- Change label links in tables to regular weight
- Change hover state background color for primary buttons
- Remove hover state for secondary buttons
- Add focus and hover state background color for tabs (only ones where the background is the light grey color)
- Make exposed actions on the Linode Details view match the styles used in the Linode Landing table


Jskobos (approved):

DevDW (approved):
LGTM overall 👍 , just a couple of points:

1. Is it intentional that the hover state background color for primary buttons in dark mode remains unchanged from what we currently have outside this PR (i.e., the background color gets lighter on hover)?

2. In dark mode the tab hover state background color doesn't blend into the bottom border as it does in light mode, which makes it look slightly visually disrupted to me
Screen Shot 2021-03-09 at 2 58 32 PM
Screen Shot 2021-03-09 at 2 58 20 PM

PR #7479, opened 2021-03-09
Configs and skeleton for validation package

Jskobos:
--- Copied from the first attempt at this, #7461 ---

## Description

Laying the groundwork for a change we've discussed for a long time but haven't moved on. The idea is to add a new package to our workspace, @linode/validation, that will contain all of our Yup schemas. I'll explain this in detail in the forthcoming README.md, but tldr;

- It's not possible to turn client-side validation off, which is a weird choice for a public package
- The validation doesn't match the API's exactly (we can't validate uniqueness and some things we've decided are too complicated and passed them off to the backend, which is fine for our internal use but confusing for external users).
- We're double-validating in many cases (whenever we use Formik's validationSchema option); the same schema is run on form submit by Formik, then run again in the JS client.
- The JS client bundle is a bit too big, and this is a big part of it, which is a shame given that not everyone necessarily wants to run client validation.

We'll probably want to organize some of the files by directory but for now I just copied them all over to see if the config files would work.

## Note to Reviewers

Run build (test won't work) to make sure it builds; I imported these files in api-v4/src/account.ts as a POC.

Future steps include:
-  Removing all schemas from the JS client (temporarily just replacing them with imports from /validation)
- Adding an options argument to all the client methods (lot of typing 😢 ) so that consumers can specify a schema if they want to have it applied
- Replacing Manager imports from api-v4/schema with validation/schema
- Testing etc.
- Publish to the @linode npm org so that external users can use the schemas

Note that absolutely nothing should change as a result of this PR, in either Manager or the JS client.


.github/workflows/ci.yml
```diff
@@ -4,11 +4,27 @@ on:
   push:
     branches:
       - master
+      - ci/validation
```
johnwcallahan, line 7: can remove before merge.

johnwcallahan (approved):
CI works, building the new package works, building/starting the app works. Is adding a dev Webpack config in the next phase? I think we should have that in place when we start moving all the schemas over.

Nice work on this, and thanks @stvnjacobs for getting the workflow file working!

ndukatz (approved):
Looks good from what I can tell. E2E passed

tiffwong (approved):

PR #7270, opened 2020-12-30
M3-4760: Change border-radius of buttons to 1px

tiffwong:
Includes fixes for "Add a Tag" cell font and `UserMenu` hover state

PR #7484, opened 2021-03-10
Fix collision on FW -> Linodes

johnwcallahan:
Fix this collision:

image (2)

In doing so, I restored the passing down of classNames to the Button in AddNewLink_CMR. This reverses a change made here: https://github.com/linode/manager/pull/7270/files#diff-14508ed4b0441d547fadf5aa1f4c123ee37bb0abc5e2a5b97029bf80c6d73405L30.

### DIFF
--- a/packages/manager/src/features/Firewalls/FirewallDetail/Devices/FirewallLinodesLanding.tsx
+++ b/packages/manager/src/features/Firewalls/FirewallDetail/Devices/FirewallLinodesLanding.tsx
@@ -25,7 +25,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     fontSize: '16px',
   },
   link: {
-    margin: '8px 8px 8px 0px',
+    margin: '8px 8px 8px auto !important',
   },
   cmrSpacing: {
     [theme.breakpoints.down('md')]: {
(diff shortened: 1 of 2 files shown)


Jskobos (approved):

ndukatz (approved):

packages/manager/src/features/Firewalls/FirewallDetail/Devices/FirewallLinodesLanding.tsx
```diff
@@ -134,21 +134,23 @@ const FirewallLinodesLanding: React.FC<CombinedProps> = (props) => {
...
         flexDirection="row"
-        alignItems="flex-end"
-        justifyContent="flex-end"
+        alignItems="center"
+        justifyContent="space-between"
+        flexWrap="wrap"
+        marginTop="10px"
       >
+        <Typography
+          className={classnames({
```
DevDW, line 146: could do 

```
className={`${classes.message} ${classes.cmrSpacing}`}
```

and delete `import classnames ...`

DevDW (approved):

packages/manager/src/features/Firewalls/FirewallDetail/Devices/FirewallLinodesLanding.tsx
tiffwong: Or you can even add the `cmrSpacing` styles to `classes.message` since that's the only place it's currently being used

tiffwong (approved):
Only thing I noticed is that the button should be aligned to the right edge when its > 960px, otherwise it looks good!
Screen Shot 2021-03-10 at 4 56 24 PM

johnwcallahan:
@DevDW @tiffwong good feedback, thank you.

PR #7487, opened 2021-03-11
M3-4966: Remove "Status" column from Community StackScripts landing page

DevDW:
## Description
Since Community StackScripts are all public, this removes the "Status" column from the table on the Community StackScripts tab on the StackScripts landing page.

Areas elsewhere in the app where StackScripts appear (Create flow; Rebuild, etc. from Community StackScript; etc.) should be unaffected.

## Type of Change

### DIFF
--- a/packages/manager/src/features/StackScripts/StackScriptBase/StackScriptBase.tsx
+++ b/packages/manager/src/features/StackScripts/StackScriptBase/StackScriptBase.tsx
@@ -35,6 +35,7 @@ import {
   generateCatchAllFilter,
   generateSpecificFilter,
 } from '../stackScriptUtils';
+import { StackScriptsRequest } from '../types';
 import withStyles, { StyleProps } from './StackScriptBase.styles';
 
 type CurrentFilter = 'label' | 'deploys' | 'revision';
@@ -79,12 +80,7 @@ type CombinedProps = StyleProps &
     publicImages: Record<string, Image>;
     currentUser: string;
     category: string;
-    request: (
-      user: string,
-      pageArgs: { page: number; page_size: number },
-      filter: any,
-      grants?: Grant[]
-    ) => Promise<ResourcePage<StackScript>>;
+    request: StackScriptsRequest;
   };
 
 interface HelperFunctions {
--- a/packages/manager/src/features/StackScripts/StackScriptBase/StackScriptBase.tsx
+++ b/packages/manager/src/features/StackScripts/StackScriptBase/StackScriptBase.tsx
@@ -1,6 +1,9 @@
 import { Grant } from '@linode/api-v4/lib/account';
 import { Image } from '@linode/api-v4/lib/images';
-import { StackScript } from '@linode/api-v4/lib/stackscripts';
+import {
+  StackScript,
+  StackScriptsRequest,
+} from '@linode/api-v4/lib/stackscripts';
 import { APIError, ResourcePage } from '@linode/api-v4/lib/types';
 import classnames from 'classnames';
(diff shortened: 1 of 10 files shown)


packages/manager/src/features/StackScripts/StackScriptBase/StackScriptBase.tsx
```diff
@@ -79,7 +79,12 @@ type CombinedProps = StyleProps &
     publicImages: Record<string, Image>;
     currentUser: string;
     category: string;
-    request: Function;
```
DevDW, line 82: TypeScript was not pleased with the use of `Function` here:

Screen Shot 2021-03-10 at 11 00 44 PM

packages/manager/src/features/StackScripts/StackScriptBase/StackScriptBase.tsx
```diff
@@ -79,7 +79,12 @@ type CombinedProps = StyleProps &
...
+    request: (
+      user: string,
+      pageArgs: { page: number; page_size: number },
+      filter: any,
+      grants?: Grant[]
+    ) => Promise<ResourcePage<StackScript>>;
```
Jskobos, line 87: This is fine, should probably be consistent with type declarations in the parents, e.g. in SelectStackScriptPanelContent.tsx:
```
  request: (
    username: string,
    params?: any,
    filter?: any,
    stackScriptGrants?: Grant[]
  ) => Promise<ResourcePage<StackScript>>;
```

Might be even better to declare it as a type in a StackScripts/types.ts file somewhere and then use it as needed. And I'd like to see if we can start using `unknown` instead of `any` in cases like this to see if it breaks something.

Jskobos (approved):

packages/manager/src/features/StackScripts/StackScriptBase/StackScriptBase.tsx
```diff
@@ -1,6 +1,9 @@
 import { Grant } from '@linode/api-v4/lib/account';
 import { Image } from '@linode/api-v4/lib/images';
-import { StackScript } from '@linode/api-v4/lib/stackscripts';
+import {
```
Jskobos, line 3: If you trace the method that's being passed around all over the place to its source, you'll find it's in stackScriptUtils.ts. The function includes a bunch of our custom logic, which external users of the JS client have no use for or knowledge of. These types should live inside the manager directory.

We have an existing system where Manager-specific types live inside the corresponding /features/ subdirectory, while types that are directly derived from the API live inside api-v4/{entity}/types.ts. See for example https://github.com/linode/manager/blob/develop/packages/manager/src/features/Volumes/types.ts.

Jskobos: I implied this when I suggested "a StackScripts/types.ts file somewhere" (the casing was deliberate!) but I should have been more specific.

Jskobos (changes requested):

packages/manager/src/features/StackScripts/StackScriptBase/StackScriptBase.tsx
DevDW: Thanks for clarifying -- I undid the changes in the JS client, made a new `features/StackScripts/types.ts` file with the type, and used that type in the handful of StackScript files from the earlier commit plus `stackScriptUtils.ts`

ndukatz (approved):
Looks good!

tiffwong (approved):
Styling looks good!

Jskobos (approved):

PR #7488, opened 2021-03-11
M3 4414 clone linode e2e

ndukatz:
Initial Linode Cloning test. The changes on MonitorStatus.tsx are just from formatting.


packages/manager/cypress/support/api/common.ts
```diff
@@ -65,6 +65,7 @@ export const testNamePrefix = 'cy-test-';
 export const isTestEntity = (entity) =>
   entity.tags?.includes(testTag) ||
   entity.label?.startsWith(testNamePrefix) ||
-  entity.summary?.includes(testTag);
+  entity.summary?.includes(testTag) ||
+  entity.label?.includes('clone');
```
Jskobos, line 69: This is an old regression, the label of a cloned linode should reflect the original. I pushed up a change to fix this, you should be able to remove this line.

Jskobos (approved):

DevDW (approved):

PR #7489, opened 2021-03-11
Billing Summary updates

johnwcallahan:
## Description

UX changes replacing the Billing Summary Header:

Screen Shot 2021-03-15 at 9 01 00 AM

The "Promotions" panel will not appear unless there is an active promotion. 

When testing, please try various promotions with and without monthly cap, etc. Also try: negative balance, 0 balance, positive balance. 

Still todo: 

- [x] Documentation button
- [x] Helper copy 
- [x] Unit tests


ndukatz (approved):
Looks really good so far!

DevDW (approved):
Looks to be working well and I think this layout is an improvement over the previous iteration; hopefully everyone finds it to be clearer in communicating the balance/credit, promotions being applied, etc.

tiffwong (approved):
This looks great!

packages/manager/src/features/Billing/BillingDetail.tsx
```diff
@@ -78,7 +78,7 @@ export const BillingDetail: React.FC<CombinedProps> = (props) => {
           <Grid item xs={12} md={12} lg={12} className={classes.main}>
             <BillingSummary
               balance={account?.data?.balance ?? 0}
-              promotion={account?.data?.active_promotions?.[0]}
+              promotions={account?.data?.active_promotions}
               balanceUninvoiced={account?.data?.balance_uninvoiced ?? 0}
               mostRecentInvoiceId={mostRecentInvoiceId}
```
Jskobos, line 83: This is still in the child's props interface but doesn't seem to be used anymore.

packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
```diff
@@ -165,190 +103,184 @@ export const BillingSummary: React.FC<Props> = (props) => {
...
+              <Box display="flex" alignItems="center">
+                <Typography variant="h3">Promotions</Typography>
+                <HelpIcon
+                  className={classes.helpIcon}
+                  text="Some helper text about promotions."
+                />
+              </Box>
+              <Divider className={classes.divider} />
+              {promotions?.map((thisPromo) => (
+                <PromoDisplay key={thisPromo.summary} {...thisPromo} />
```
Jskobos, line 187: Are promo summaries guaranteed to be unique? Our factory promos are causing a bunch of key errors in the console, I doubt it's possible irl but we ought to check.

Jskobos (approved):
Looks much better this way. I think the route matching should be addressed before merge, but everything else is non-blocking feedback. Approving pending that change.

I don't think this situation is possible now but imo we should just let this `overflow: scroll`:

Screen Shot 2021-03-16 at 8 56 54 AM

PR #7492, opened 2021-03-12
M3-5003: Consistent button and dialog naming conventions

tiffwong:
## Description

Matthew and I came up with some general guidelines to remove redundancy and make everything a little more consistent

### Buttons
- "Create [Entity]"
- "Add" is reserved for entities that generally need to be attached to another entity

### Dialogs/Drawers
- "[Action] [Entity] [entity name]"
- The primary button should be "[Action] [Entity]"
- If the action is to edit something, the primary button should be "Save Changes"

## Note to Reviewers
The next step is to make sure all the buttons are right aligned with the rightmost button being the primary action. We also want to look into the wording in the destructive dialog content


ndukatz (approved):

packages/manager/src/components/LandingHeader/LandingHeader.tsx
```diff
@@ -62,9 +60,7 @@ export const LandingHeader: React.FC<Props> = (props) => {
             onClick={onAddNew}
             style={{ width: createButtonWidth ?? defaultCreateButtonWidth }}
           >
-            {createButtonText
-              ? createButtonText
-              : `Create ${startsWithVowel ? 'an' : 'a'} ${entity}`}
+            {createButtonText ? createButtonText : `Create ${entity}`}
```
Jskobos, line 63: `createButtonText ?? 'Create ${entity}`

Jskobos (approved):
I think you may have missed a few:

- create lke cluster breadcrumb
- managed CTA buttons "Add a Credential", "Add a Monitor", etc.

packages/manager/src/components/LandingHeader/LandingHeader.tsx
tiffwong: "Create Cluster" was intentionally left like that but I did make the changes for Managed, thanks for catching that!

PR #7493, opened 2021-03-12
Upgrade storybook

tiffwong:
Run `yarn storybook` to make sure it's running and check to make sure nothing is broken in Cloud Manager


Jskobos (approved):

packages/manager/src/components/PaginationControls/PaginationControls.stories.tsx
```diff
@@ -1,31 +1,31 @@
-import { storiesOf } from '@storybook/react';
-import * as React from 'react';
-import PaginationControls from './PaginationControls';
+// import { storiesOf } from '@storybook/react';
```
DevDW, line 1: can we do something like

```
const Controls: React.FC<{}> = (props) => {
  const [currentPage, setCurrentPage] = React.useState<number>(1);

  const handleClick = (page: number) => {
    setCurrentPage(page);
  };

  return (
    <PaginationControls
      count={250}
      page={currentPage}
      onClickHandler={handleClick}
      pageSize={25}
    />
  );
};

export default {
  title: 'Pagination Controls',
};

export const Example = () => <Controls />;
```

to get this working again? (looks like TagsPanel.stories.tsx can also be fixed)

DevDW (approved):
Storybook and the app are both working 👍 a couple of stories are still commented out though

johnwcallahan (commented):
Storybook is running for me, but the styles got wonky:

Screen Shot 2021-04-27 at 8 48 40 AM

Screen Shot 2021-04-22 at 9 11 06 AM

codyfinn (approved):

PR #7496, opened 2021-03-15
M3-4947 Add: status banners on all /support pages

Jskobos:
## Description

- Refactor routing of all /support links so that they share a common
rendered root.
- Add banners to the aforementioned root for all incidents

### Todo
- [ ] Integration tests with the msw

## Note to Reviewers

The banners are dismissible but the dismissals aren't saved anywhere (the banners come back if you reload the page). I'm waiting to use the notifications hooks in #7482 once that's been merged.

You can test with the mocks, or there are some real scheduled maintenance events atm too.


packages/manager/src/features/Help/HelpLanding.tsx
```diff
@@ -24,7 +24,7 @@ const styles = (theme: Theme) =>
 type CombinedProps = WithStyles<ClassNames>;
 
 export class HelpLanding extends React.Component<CombinedProps, {}> {
-  render() {
+  render(): JSX.Element {
```
Jskobos, line 27: I think we need to disable this rule, this is going to get very tedious

packages/manager/src/features/Help/StatusBanners.tsx
```diff
@@ -0,0 +1,138 @@
...
+      <Typography className={classes.header}>
+        <Link to={href}>
+          <strong>
+            {title}
+            {status ? `: ${capitalize(status)}` : ''}
+          </strong>
+        </Link>
+      </Typography>
+      <Typography
+        dangerouslySetInnerHTML={{ __html: sanitizeHTML(message) }}
```
johnwcallahan, line 131: Do status body's have HTML?

packages/manager/src/features/Help/StatusBanners.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const status = _status ?? '';
+  const classes = useStyles();
+
+  const [hidden, setHidden] = React.useState(false);
+
+  if (hidden) {
+    return null;
+  }
+
+  const handleDismiss = () => {
```
johnwcallahan, line 106: I'd either just inline this or use useCallback.

johnwcallahan (commented):
Screen Shot 2021-03-16 at 11 57 27 AM

The font size looks a little big especially compared to existing banners. 

Also, we should only show active incidents here (we'll show maintenance in the upcoming "Maintenance" page, which we don't have tickets for yet).

DevDW (approved):
I agree with John regarding the font size for these banners, but the rest looks good to me. Approved pending the limiting of the incidents displayed to only active ones.

packages/manager/src/features/Help/StatusBanners.tsx
Jskobos: apparently so, links to "open a support ticket" are common for unresolved incidents

Jskobos: We could do the find-and-replace thing to avoid the dangerous stuff, but we don't control the banners, raw html looks horrible in them, and I think the status page api is a pretty safe source. Opted for innerHTML on balance but please feel free to argue the other side.

packages/manager/src/features/Help/StatusBanners.tsx
Jskobos: this is just a placeholder, the logic is built into the dismissibleNotifications hook so I'll just kill this when that piece is ready to pull in

packages/manager/src/features/Help/StatusBanners.tsx
johnwcallahan: Yeah, that seems OK to me.

johnwcallahan (changes requested):
Just found one more thing: we probably shouldn't display incident updates with a status of "postmortem":

Screen Shot 2021-03-16 at 5 16 42 PM

In fact, maybe we should apply a (liberal) character limit just in case. Maybe an ellipsis in brackets that takes you to status.linode.com? `[...]`

Also, as a follow-on, it'd be awesome if we could make the support ticket links smarter (rather than refreshing the whole app.) We could even pre-fill them. That's a really low-priority bonus, though.

tiffwong (approved):

johnwcallahan (approved):

PR #7501, opened 2021-03-17
Pass default values to formik.resetForm

Jskobos:
## Description

Apparently calling `formik.resetForm()` does not, in fact, reset the form.

To test:

1. create a config
2. edit it and change some things
3. go to create a new config
4. the inputs from editing are still there


DevDW (approved):
Confirmed that this fixes the bug 👍

johnwcallahan (approved):

PR #7505, opened 2021-03-17
M3-4971: Prevent wrapping of Linode Detail graph legends

DevDW:
## Description
Aiming to prevent the wrapping of Linode Detail graph legends. This approach adds the `noWrap` prop to the `<TableCell />` the legend keys are rendered in, which seems to fix the issue while not causing any adverse effects where the `LineGraph` component is used elsewhere (Longview, etc.)

Before (this can currently be seen as you shrink to around the tablet view, let's say ~1080px width):
Screen Shot 2021-03-17 at 2 09 10 PM

After:
Screen Shot 2021-03-17 at 2 09 28 PM

## Type of Change

### DIFF
--- a/packages/manager/src/components/LineGraph/LineGraph.tsx
+++ b/packages/manager/src/components/LineGraph/LineGraph.tsx
@@ -320,7 +320,7 @@ const LineGraph: React.FC<CombinedProps> = (props: CombinedProps) => {
                 const { data: metricsData, format } = legendRows[idx];
                 return (
                   <TableRow key={idx}>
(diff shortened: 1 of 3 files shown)


johnwcallahan (approved):
Nice fix. A related question comes up: I wonder if we should make these graphs collapse sooner. In general things look a little wonky around 1000-1200px, with values/units getting wrapped, etc. 

Screen Shot 2021-03-17 at 4 17 12 PM

ndukatz (approved):

tiffwong (approved):

PR #7506, opened 2021-03-17
Remove default Firewall rules

johnwcallahan:
## Description

...and change default inbound Policy to "ACCEPT".


Jskobos (approved):
The tests need to be updated, but the change looks good. We're ok with this, right?

Screen Shot 2021-03-18 at 8 31 23 AM

johnwcallahan:
> We're ok with this, right?

Good catch, I think the text "No rules" works better. Just pushed up a change.

tiffwong (approved):

PR #7508, opened 2021-03-18
M3-5015: Fix alignment on Longview Landing filter input

tiffwong:
The alignment was slightly off when the width was > 1280px


Jskobos (approved):

ndukatz (approved):
Looks much better!

PR #7511, opened 2021-03-18
M3-4981: Add Google as TPA provider

tiffwong:
## Description

~Waiting on login before merging/release, copy needs to be reviewed as well~

~Still waiting on login for Google to actually work but I added everything behind a feature flag. In the meantime, please test the enabling/disabling flow for TPA via GitHub~

**Update:** New endpoint has been added to allow switching between TPA providers without having to go through the reset password flow along with renaming the tab to "Login & Authentication".

Merge both tabs under ~Password~ Login & Authentication and only show relevant components depending on the auth settings. Also updated the button to toggle back to using Linode credentials to resemble the TPA provider options

## How to test

The actual endpoint doesn't work yet but please test switching between GitHub and Linode credentials and making sure Google is only displayed when the flag is on.


packages/manager/src/features/Profile/AuthenticationSettings/ThirdPartyContent.tsx
```diff
@@ -63,68 +83,93 @@ export const ThirdPartyContent: React.FC<CombinedProps> = (props) => {
 
   const thirdPartyEnabled = props.authType !== 'password';
 
-  const displayName =
-    providers.find((thisProvider) => thisProvider.name === props.authType)
-      ?.displayName || '';
+  // If the flag is off, remove Google from the list of TPAProviders
+  if (!flags.tpaGoogle) {
+    const index = providers.findIndex((provider) => provider.name === 'google');
```
Jskobos, line 88: This can be a `.filter()` to skip the multiple steps business.

packages/manager/src/features/Profile/AuthenticationSettings/AuthenticationSettings.tsx
```diff
@@ -48,55 +56,45 @@ export const AuthenticationSettings: React.FC<CombinedProps> = (props) => {
...
-          tabs={tabs}
-          initTab={initialTab}
-        />
+        <>
+          <Paper className={classes.root}>
+            <Typography className={classes.tpa} variant="h3">
+              Log-In Method
+            </Typography>
+            <ThirdPartyContent authType={authType} />
+            {thirdPartyEnabled && <ThirdParty authType={authType} />}
```
Jskobos, line 71: these names are confusing me 😕

packages/manager/src/features/Profile/AuthenticationSettings/ThirdPartyContent.tsx
```diff
@@ -63,68 +83,93 @@ export const ThirdPartyContent: React.FC<CombinedProps> = (props) => {
...
+                className={
+                  thirdPartyEnabled && props.authType === thisProvider.name
+                    ? classes.enabled
+                    : ''
+                }
+                onClick={() => {
+                  setProvider(thisProvider.name);
+                  setDialogOpen(true);
+                }}
+                disabled={thirdPartyEnabled}
```
Jskobos, line 117: I know there's color issues, but it's hard to tell from looking that the thing is actually disabled. If "Linode" is my auth type and I click on GitHub, I get a modal. If "Google" is my auth type and I click on GitHub, nothing happens and it isn't visually clear (only the cursor is different?) how the situation has changed.

packages/manager/src/features/Profile/AuthenticationSettings/ThirdPartyContent.tsx
```diff
@@ -63,68 +83,93 @@ export const ThirdPartyContent: React.FC<CombinedProps> = (props) => {
...
+            }}
+            disabled={!thirdPartyEnabled}
+          >
+            <div>
+              <LinodeLogo className={classes.providerIcon} />
+            </div>
+            <div className={classes.providerWrapper}>
+              <div>
+                Linode
+                {!thirdPartyEnabled && props.authType === 'password' && (
```
Jskobos, line 155: This check is redundant, if authType is password then mechanically thirdPartyEnabled is false

packages/manager/src/features/Profile/AuthenticationSettings/ThirdPartyContent.tsx
```diff
@@ -63,68 +83,93 @@ export const ThirdPartyContent: React.FC<CombinedProps> = (props) => {
...
+                  <span className={classes.enabledText}>(Enabled)</span>
+                )}
+              </div>
+              {!thirdPartyEnabled && props.authType === 'password' && (
```
Jskobos, line 159: Same here

packages/manager/src/features/Profile/AuthenticationSettings/ThirdPartyDialog.tsx
```diff
@@ -45,23 +42,16 @@ const ThirdPartyDialog: React.FC<CombinedProps> = (props) => {
     <ConfirmationDialog
       className={classes.dialog}
       open={open}
-      title={`Enable ` + `${displayName}` + ` Authentication`}
+      title="Change Log-in Method?"
```
Jskobos, line 45: I like log in (verb) and login (noun) but ymmv

packages/manager/src/features/Profile/AuthenticationSettings/ThirdPartyContent.tsx
Jskobos: You could even make the flag a `string[]` instead of a boolean, then do `providers.filter(thisProvider => flags.tpaProviders.includes(thisProvider.name))` or something similar.

Jskobos (changes requested):
Looking good, some things need to be cleared up and/or cleaned up. And I think dark mode is a bit off:

Screen Shot 2021-03-29 at 2 29 47 PM

packages/manager/src/features/Profile/AuthenticationSettings/TPAProviders.tsx
```diff
@@ -0,0 +1,256 @@
...
+  );
+
+  return (
+    <>
+      <Paper className={classes.root}>
+        <Typography variant="h3">Login Method</Typography>
+        <Typography className={classes.copy}>
+          You can use your Linode credentials or another provider such as Google
+          or GitHub to log in to your Linode account. More information is
+          availible in{' '}
```
DevDW, line 140: "available"

DevDW (commented):
Looking good so far, noticed a typo

packages/manager/src/features/Profile/AuthenticationSettings/TPADialog.tsx
```diff
@@ -0,0 +1,106 @@
...
+      <Typography className={classes.copy} variant="body1">
+        This will disable your login via{' '}
+        {currentProvider?.displayName ?? 'username and password'}.
+      </Typography>
+    </ConfirmationDialog>
+  );
+};
+
+const changeLogin = (provider?: ProviderOptions) => {
+  return provider === undefined
```
Jskobos, line 65: Maybe add a comment here to clarify what's happening (you're using the fact that if `provider` is `undefined` it means they just clicked on "Linode" but it's a bit implicit. Another option would be to default `provider` to `linode` throughout this file and use it that way.

packages/manager/src/features/Profile/AuthenticationSettings/TPADialog.tsx
```diff
@@ -0,0 +1,106 @@
...
+        buttonType="primary"
+        loading={loading}
+        onClick={() => {
+          onClose();
+          changeLogin(provider);
+        }}
+        data-qa-confirm
+        data-testid={'dialog-confirm'}
+      >
+        Change login
```
Jskobos, line 100: Capital L maybe

packages/manager/src/features/Profile/AuthenticationSettings/TPAProviders.tsx
```diff
@@ -0,0 +1,256 @@
...
+        </Typography>
+        <Grid container className={classes.providers}>
+          <Grid item md={4}>
+            <Button
+              className={classnames({
+                [classes.provider]: true,
+                [classes.enabled]: !thirdPartyEnabled,
+              })}
+              onClick={() => {
+                setProvider(undefined);
```
Jskobos, line 154: Maybe you could set provider to `linode` in this case to avoid the undefined stuff in TPADialog

Jskobos: Another option would be to make a manual provider type for Linode so that you're not handling it differently (this button and grid code is mostly repetitive and there are few `currentProvider.displayName ?? 'username and password'` things that might be possible to clean up this way.

Jskobos (approved):
Looks good, I left some cleanup stuff but nothing blocking.

packages/manager/src/components/ConfirmationDialog/ConfirmationDialog.tsx
```diff
@@ -3,27 +3,29 @@ import Dialog, { DialogProps } from 'src/components/core/Dialog';
...
-
-const styles = (theme: Theme) =>
-  createStyles({
-    root: {},
-    error: {
-      color: '#C44742',
-      marginTop: theme.spacing(2),
+const useStyles = makeStyles((theme: Theme) => ({
+  root: {
+    '& .MuiPaper-root': {
```
johnwcallahan, line 11: These are global changes. Are they intended only for the TPA confirmation dialogs, or for all of them app-wide?

packages/manager/src/features/Profile/AuthenticationSettings/TPADialog.tsx
johnwcallahan: It does look a little inconsistent, but this is actually the direction that we want to move... only entity types (as proper nouns) being capitalized.

packages/manager/src/features/Profile/AuthenticationSettings/TPAProviders.tsx
```diff
@@ -0,0 +1,256 @@
...
+  const currentProvider = providers.find(
+    (thisProvider) => thisProvider.name === props.authType
+  );
+
+  return (
+    <>
+      <Paper className={classes.root}>
+        <Typography variant="h3">Login Method</Typography>
+        <Typography className={classes.copy}>
+          You can use your Linode credentials or another provider such as Google
```
johnwcallahan, line 138: At least for the release, we shouldn't hardcode "Google" here, since we're releasing without the feature flag enabled for Google.

johnwcallahan (approved):
Good work on this. Let's merge it so we can test the whole flow in dev. There are a couple of things to address before release, including whether or not the confirmation dialog minWidth change was intended to be global, as well as the hardcoding of "Google" in the description.

Also, please create a tech-story ticket to add thorough unit tests.

PR #7512, opened 2021-03-18
M3-4948 Fix: Don't clear support modal contents from local storage on close

Jskobos:
Use a separate onCancel handler for the support ticket modal's Cancel
button. If the user closes the modal by clicking outside it, hitting the X button,
or pressing ESC (most of these can be done accidentally), close the drawer but
don't clear the values from local storage. In this way, if the user opens the drawer
their contents are still present.

If the user explicitly hits the "Cancel" button, though, take this opportunity to
clear the contents. Contents will also clear when a ticket has been created successfully.

We can also consider disabling backdrop and escape click handlers for the support dialog (or any dialog, if accidental closures become annoying), it would be easy to do.

### DIFF
--- a/packages/manager/src/features/Support/SupportTickets/SupportTicketDrawer.tsx
+++ b/packages/manager/src/features/Support/SupportTickets/SupportTicketDrawer.tsx
@@ -305,7 +305,7 @@ export const SupportTicketDrawer: React.FC<CombinedProps> = (props) => {
 
   const onCancel = () => {
     props.onClose();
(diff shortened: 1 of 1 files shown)


johnwcallahan (changes requested):
Looks good, but there's that thing where we briefly see the placeholder text after clicking cancel. Maybe add the a dreaded timeout?

johnwcallahan (approved):

DevDW (approved):

PR #7513, opened 2021-03-18
Demo feedback

Jskobos:
## Description

Assorted items from demo prep this morning:

- M3-5017: Show IDs (and entity labels as a POC) for abuse ticket notifications in the drawer
- M3-5018: Remove "right" from "right now" in the migration banner/confirmation modal
- M3-????: Make the abuse ticket banner a warning not an error


DevDW (approved):
LGTM 👍 

One thing I noticed is that there's still a severity disagreement between the bounced email global notification (red w/ exclamation point) and the notification in the drawer (yellow w/ exclamation point), should those be brought into agreement as well?

tiffwong (approved):
Agreed with Dajahi's comment on making sure the severity levels are consistent but otherwise, it looks great!

johnwcallahan (approved):

PR #7514, opened 2021-03-18
Change: Make LKE upgrade banner dismissible

Jskobos:
## Description

Newest installment in "all banners should be dismissible: a continuing saga." LKE version upgrade banners are now dismissible, using the hooks from useDismissibleBanners. Once dismissed, they won't reappear until a new version of LKE is available.

### DIFF
--- a/packages/manager/src/features/Firewalls/FirewallLanding/FirewallBanner.tsx
+++ b/packages/manager/src/features/Firewalls/FirewallLanding/FirewallBanner.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import Typography from 'src/components/core/Typography';
 import DismissibleBanner from 'src/components/DismissibleBanner';
 import { dcDisplayNames } from 'src/constants';
 import { useRegionsQuery } from 'src/queries/regions';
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/Firewalls/FirewallLanding/FirewallBanner.tsx
```diff
@@ -15,11 +16,12 @@ export const FirewallBanner: React.FC<{}> = (_) => {
...
 
   return (
-    <DismissibleBanner
-      preferenceKey="firewall-beta-notification"
-      message={`Cloud Firewalls are currently in beta. Firewalls can
-        only be used with Linodes in ${regionDisplayList}.`}
-    />
+    <DismissibleBanner preferenceKey="firewall-beta-notification">
+      <Typography>
+        Cloud Firewalls are currently in beta. Firewalls can only be used with
```
Jskobos, line 21: A consequence of this refactor is that this banner may pop back up for people who have dismissed it. I think that's fine tbh but I could wire it to avoid the issue if necessary.

DevDW (approved):

tiffwong (approved):

PR #7515, opened 2021-03-18
M3-5020: Styling improvements to Network Interfaces

tiffwong:


DevDW (approved):
The ConfigRow padding looks good. The dividers in the Network Interfaces section work, although this case feels odd to me:

Screen Shot 2021-03-18 at 3 55 52 PM

The dividers within the section are the same dividers that are used between the sections, so visually it could potentially cause some confusion

Jskobos:
`git fetch upstream && git rebase -i upstream/develop` plz, drop the four commits from DW and I

Jskobos (commented):
Looks good, I think @DevDW is right that there ought to be a divider between eth0 and eth1 as well. Matthew also mentioned making the help icons grey instead of blue in the modal, that doesn't have to be done in this PR but if you wanted to knock those out while you're in there that would be great too.

Jskobos (approved):
Looks good, at the risk of being ocd is the border between eth1 and eth2 doubled?

Screen Shot 2021-03-19 at 11 09 16 AM

tiffwong:
@Jskobos I just checked the specs and it all looks the same 🤔

PR #7516, opened 2021-03-18
M3-4941: Attach a VLAN section in Linode Create flow

DevDW:
## Description
Create new `AttachVLAN` component that uses `InterfaceSelect.tsx` and appears above the Add-ons in the Linode Create flow

## Type of Change

## To-Do:
- [x] Include VLAN data w/ Linode creation payload (as opposed to the "Attach VLAN" button in the designs)
- [x] Error handling
- [x] Helper text


DevDW:
I'll go ahead and open this up for review, but a few things are still being worked on (helper text finalization, error handling, tighten things up). The VLAN label and IPAM address get added to the Linode Create payload now.

Jskobos (changes requested):
This is really good, the approach is exactly the right idea. Can't check the error handling yet obviously but the payload is correct.

I think we need to be a little more clever with the payload, though. Having a single select is better for UX, but if I choose to attach a VLAN, the `interfaces` array has only a single object, which means (I think) that the VLAN will end up attached to eth0 and the Linode will not be available on the public internet. We should essentially `[...publicInterfaceObject, ...this.state.interfaces]` or similar so that we always send a public interface in the payload.

packages/manager/src/features/linodes/LinodesCreate/AddonsPanel.tsx
```diff
@@ -131,6 +115,11 @@ const AddonsPanel: React.FC<CombinedProps> = (props) => {
   return (
     <Paper className={classes.root} data-qa-add-ons>
       <div className={classes.inner}>
+        <AttachVLAN
```
Jskobos, line 118: This will need to be behind a feature flag, using the showVlans logic you've removed.

packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
```diff
@@ -306,6 +308,16 @@ export class LinodeCreate extends React.PureComponent<
         // StackScripts
         stackscript_id: this.props.selectedStackScriptID,
         stackscript_data: this.props.selectedUDFs,
+
+        // VLANs
+        interfaces: [
+          defaultPublicInterface, // to make sure the eth0 slot is not occupied by a VLAN
+          {
```
Jskobos, line 315: This is better, but it introduces a new problem: it will always send 2 interfaces, even if the user doesn't choose a vlan.

packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
```diff
@@ -287,6 +287,17 @@ export class LinodeCreate extends React.PureComponent<
...
   createLinode = () => {
+    let interfaces = [defaultPublicInterface]; // the purpose of defaultPublicInterface is to make sure the eth0 slot is not occupied by a VLAN.
+    if (Boolean(this.props.vlanLabel)) {
+      interfaces.push({
+        purpose: 'vlan',
+        label: this.props.vlanLabel,
+        ipam_address: this.props.ipamAddress,
+      });
+    } else {
+      interfaces = [];
```
Jskobos, line 298: This isn't necessary, if you submit just the defaultPublicInterface you end up with the same result (i.e. a config where eth0 is a public interface).

Jskobos (approved):

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/InterfaceSelect.tsx
```diff
@@ -113,11 +113,9 @@ export const InterfaceSelect: React.FC<Props> = (props) => {
                 placeholder="Create or select a VLAN"
                 variant="creatable"
                 createOptionPosition="first"
-                defaultValue={vlanOptions.find(
-                  (thisVlan) => thisVlan.value === label
-                )}
+                value={vlanOptions.find((thisVlan) => thisVlan.value === label)}
                 onChange={handleLabelChange}
-                isClearable={false}
+                isClearable
```
Jskobos, line 118: This got overwritten in the merge resolution I think

tiffwong (changes requested):
The label for IPAM Address wraps when there's plenty of space so I'd recommend adding `whiteSpace: nowrap` to the label
Screen Shot 2021-03-23 at 5 34 06 PM

The width of the input boxes should also be increased for mobile
Screen Shot 2021-03-23 at 5 32 19 PM

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/InterfaceSelect.tsx
```diff
@@ -12,6 +12,22 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+  vlanLabelField: {
+    width: 202,
+    height: 35,
+    marginRight: theme.spacing(),
+  },
+  ipamAddressLabel: {
+    '& label': {
+      whiteSpace: 'nowrap',
+    },
+    [theme.breakpoints.down(1280)]: {
```
tiffwong, line 27: Replace `1280` with `'md'` just in case we change our breakpoints in the future

tiffwong (approved):
I left one cleanup item but everything else looks good!

PR #7517, opened 2021-03-18
Demo prep feedback: Fix casing for eth0 in Configurations table

DevDW:
## Description
Change the eth0 details ("Public", "Public, Private") from all capital letters.

## Type of Change


Jskobos (approved):

johnwcallahan (commented):
Casing looks good, but I actually have a note from design review that we should remove "Private" and make it "Public Internet" instead.

johnwcallahan (approved):

PR #7518, opened 2021-03-22
Fix schema and error handling for interfaces

Jskobos:
## Description

Fix the linodeInterfaceSchema and error paths in the InterfaceSelect

You should be able to submit the Add Config form (using alpha services).

### DIFF
--- a/packages/api-v4/.eslintrc.js
+++ b/packages/api-v4/.eslintrc.js
@@ -6,7 +6,7 @@ module.exports = {
     // BUT we decided not to use this feature due to a very important performance impact
     // project: 'tsconfig.json',
     ecmaVersion: 2020,
-    warnOnUnsupportedTypeScriptVersion: true
+    warnOnUnsupportedTypeScriptVersion: true,
   },
   plugins: ['@typescript-eslint', 'sonarjs', 'prettier'],
   extends: [
@@ -15,7 +15,7 @@ module.exports = {
     // like eslint:recommended, except it only turns on rules from our TypeScript-specific plugin.
     'plugin:@typescript-eslint/recommended',
     'plugin:sonarjs/recommended',
-    'plugin:prettier/recommended' // Enables eslint-plugin-prettier and eslint-config-prettier. This will display prettier errors as ESLint errors. Make sure this is always the last configuration in the extends array.
+    'plugin:prettier/recommended', // Enables eslint-plugin-prettier and eslint-config-prettier. This will display prettier errors as ESLint errors. Make sure this is always the last configuration in the extends array.
   ],
   rules: {
     // prepend `_` to an arg you accept to ignore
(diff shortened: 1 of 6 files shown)


packages/api-v4/.eslintrc.js
```diff
@@ -68,20 +68,41 @@ module.exports = {
...
+          {
+            types: {
+              String: true,
+              Boolean: true,
+              Number: true,
+              Symbol: true,
+              '{}': false,
+              Object: false,
+              object: false,
+              Function: false,
```
Jskobos, line 100: This is the point of all the changes here, we use `: Function[]` in a few places and until we can do a typing overhaul it ought to pass the linter.

packages/manager/src/utilities/formikErrorUtils.ts
```diff
@@ -3,22 +3,24 @@ import { getAPIErrorOrDefault } from './errorUtils';
...
+  callback: (error: unknown) => void,
   fieldErrors: APIError[] = []
 ) => {
-  const mappedFieldErrors = fieldErrors.reduce(
-    (result, { field, reason }) =>
-      field ? { ...result, [field]: reason } : result,
-    {}
-  );
+  const mappedFieldErrors = fieldErrors
+    .reverse()
```
Jskobos, line 10: @johnwcallahan I did this so that errors that come first in the Yup schema end up displayed to the user, as is they get overwritten. Since we usually stack the schema rules in importance order, we sometimes end up with `Label can't contain special characters` when a required field is empty. I don't think this can possibly break anything but it's a global-ish change, please tell me what you think.

packages/api-v4/src/linodes/linodes.schema.ts
```diff
@@ -1,9 +1,11 @@
-import { array, boolean, lazy, mixed, number, object, string } from 'yup';
+import { array, boolean, mixed, number, object, string } from 'yup';
 import { parseCIDR } from 'ipaddr.js';
-// import zxcvbn from 'zxcvbn';
-// import { MINIMUM_PASSWORD_STRENGTH } from 'src/constants';
 
 const validateIP = (ipAddress: string) => {
+  if (ipAddress === '') {
+    // ipam_address is technically required, but empty strings are valid
```
Jskobos, line 6: maybe they shouldn't be! but they're getting through the back end for now

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
```diff
@@ -607,7 +607,8 @@ const LinodeConfigDialog: React.FC<CombinedProps> = (props) => {
                     key={`eth${idx}-interface`}
                     slotNumber={idx}
                     readOnly={readOnly}
-                    error={formik.errors[`interfaces[${idx}]`]}
+                    labelError={formik.errors[`interfaces[${idx}].label`]}
```
Jskobos, line 610: API doesn't actually support this yet, but Yup does and the API has said they're working on it.

tiffwong (approved):

DevDW (approved):

PR #7520, opened 2021-03-22
M3-5014 Fix: IP sharing bug

Jskobos:
## Description

IP sharing (and transfer) rely on having all of a user's Linodes in order to display the ip addresses as options. For large customers, this means that if you land directly on a detail page, and open the sharing modal, your options will be limited (or nonexistent), since only a single request for Linodes has been made.

I refactored the old code in IPSharing so that we could use useReduxLoad; a longer-term solution will be to use a query and request all Linodes in the selected region instead. There was some questionable code in there so I thought the refactor was worth it. I also removed IPSharingPanel, which afaict is an older, unused version of the same file.

To test, please check all parts of the IP sharing flow, with both a normal and a "large" account (>500 Linodes). I used log-in-as-customer for this, but if you go this route you're only checking to see that the correct options are displayed, please don't submit the form.


DevDW (approved):
The functionality all seems to be working as intended 👍 . It appears that the hover state of the "Add IP Address" button got thrown off a bit compared to Prod though:

Prod:
Screen Shot 2021-03-24 at 11 45 58 AM

This branch:
Screen Shot 2021-03-24 at 11 46 20 AM

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -90,308 +70,315 @@ interface Props {
...
-    );
-  };
-
-  onIPSelect = (ipIdx: number) => (e: Item<string>) => {
+  const onIPSelect = (ipIdx: number, e: Item<string>) => {
     if (ipIdx === undefined) {
       return;
     }
-    const newIPsToShare = clone(this.state.ipsToShare);
+    const newIPsToShare = clone(ipsToShare);
```
johnwcallahan, line 108: I feel like this should use the function state updater pattern like `onIPDelete` uses, if nothing else but for consistency, but also since the new value is based on the old state value.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -90,308 +70,315 @@ interface Props {
...
-            hideLabel
-          />
-        </Grid>
-      </Grid>
-    );
-  };
-
-  onIPSelect = (ipIdx: number) => (e: Item<string>) => {
+  const onIPSelect = (ipIdx: number, e: Item<string>) => {
     if (ipIdx === undefined) {
```
johnwcallahan, line 105: I'd be in favor of removing this (hopefully) unnecessary check and trust TS and our code.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -90,308 +70,315 @@ interface Props {
...
     });
   };
 
-  onSubmit = () => {
+  const onSubmit = () => {
     const finalIPs = uniq(
-      this.state.ipsToShare.filter(
-        (ip: string) => ip !== IPSharingPanel.selectIPText
-      )
+      ipsToShare.filter((ip: string) => ip !== selectIPText)
```
johnwcallahan, line 141: Obviously this is carryover, but it's pretty weird that we treat `selectIPText` like an IP and then strip it out at the end. Shouldn't `selectIPText` just be a placeholder in the Select component instead of a selectable option?

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -90,308 +70,315 @@ interface Props {
...
-        });
+        props.refreshIPs();
+        setErrors(undefined);
+        setSubmitting(false);
+        setSuccessMessage('IP Sharing updated successfully');
       })
       .catch((errorResponse) => {
         const errors = getAPIErrorOrDefault(
           errorResponse,
           'Unable to complete request at this time.'
```
johnwcallahan, line 158: We should change this to something more stylistically aligned with our modern error messages.

johnwcallahan (approved):
Looks good, left a few additional refactoring comments.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -90,308 +70,306 @@ interface Props {
...
+          isClearable={false}
+          placeholder="Select an IP"
+          label="Select an IP"
+          inputId={`ip-select-${idx}`}
+          hideLabel
+          overflowPortal
+        />
+      </Grid>
+      {handleDelete ? (
+        <Grid item className={classes.removeCont}>
```
tiffwong, line 357: If you add `sm={2}` here, it won't wrap as early

tiffwong (approved):
Looks good! Left one cleanup item but otherwise, LGTM

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -90,308 +70,306 @@ interface Props {
...
+                      handleSelect={onIPSelect}
+                      labels={ipChoiceLabels}
+                      getRemainingChoices={remainingChoices}
                     />
-                  </div>
-                )}
-              </React.Fragment>
-            )}
+                  ))}
+                  {remainingChoices('').length > 1 && (
```
johnwcallahan, line 217: I'm pretty sure this should be `>= 1` now... currently if you've got only two choices left, you can only select one.

johnwcallahan (changes requested):
One more quick thing.

johnwcallahan (approved):

PR #7521, opened 2021-03-23
VLAN assorted fixes

Jskobos:
## Description

- Make interfaceSelect filterable by region (since vlans are region-limited)
- Make sure when loading a config for editing in the drawer that there are
enough empty interfaces (purpose === 'none') added to it to match the UI pattern.


DevDW (approved):
These fixes LGTM

ndukatz (approved):

PR #7522, opened 2021-03-23
Patch/hide configs

Jskobos:
## Description

Hide the interfaces section of the config modal if vlans tag isn't active
Fix some root device stuff that I broke when I refactored a month ago


DevDW (approved):
Looks good 🚢

tiffwong (approved):

ndukatz (approved):
GTG

PR #7525, opened 2021-03-24
Change: Make Domain banner dismissible

Jskobos:
## Description

Make the "Your domains aren't being served" banner disappear. This involved adding an optional `expiry` field to our mini-database and refactoring a bunch of stuff around it. The idea is:

- If a notification is dismissed normally, it is considered "stale" after 60 days. The next time a user dismisses anything, the stale stuff will be removed from the mini-database (this is clean-up to avoid bloat inside user preferences)
- If a notification is dismissed with a specified expiry date, it is not considered "dismissed" if the expiry has passed (so regardless of cleanup, an expired dismissal will reappear in the UI; in this case, the Domain banner will reappear after 30 days). Dismissals past their expiry are also considered "stale" and will be cleaned up as described above.

## Note to Reviewers

To test:
1. Have an account with > 1 Domain and 0 Linodes (mocks won't quite work with user preferences, so that's out)
2. Go to /domains; you should see a banner
3. Dismiss it; make sure it stays dismissed when you reload the page
4. Use the super-secret user preferences editor to find the domain banner in the dismissedNotifications section (it'll be the one with an `expiry` key) and set the expiry to the past.
5. The banner should reappear.

### DIFF
--- a/packages/manager/src/features/Domains/DomainBanner.tsx
+++ b/packages/manager/src/features/Domains/DomainBanner.tsx
@@ -1,5 +1,6 @@
 import { DateTime } from 'luxon';
 import * as React from 'react';
+import Typography from 'src/components/core/Typography';
 import Link from 'src/components/Link';
 import Notice from 'src/components/Notice';
 import { makeStyles, Theme } from 'src/components/core/styles';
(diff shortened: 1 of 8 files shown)


johnwcallahan (changes requested):
This is working well, but we're storing dates in two different formats:

Screen Shot 2021-03-25 at 10 11 20 AM


Also, maybe it'd be helpful to include a "label" field or something on the DismissibleNotification object for easy debugging.

DevDW (approved):
Working nicely and the tests are thorough; approved pending addressing John's feedback

johnwcallahan (approved):

PR #7527, opened 2021-03-24
M3-4963: StackScripts Refinements

tiffwong:
## Description

StackScripts is a nightmare but I managed to wrangle the UI into something more consistent with our current style. Noticeable changes include decreasing the font sizes, updating the image chips, and better responsiveness.

## Note to Reviewers

Please check the Linode Create From StackScripts page and both the tabs on the StackScripts Landing page at different viewports. 

I'll also be releasing another PR ([https://github.com/linode/manager/pull/7528](https://github.com/linode/manager/pull/7528)) to cleanup the remaining CMR files for StackScripts.

### DIFF
--- a/packages/manager/src/features/StackScripts/StackScriptPanel/StackScriptRow.tsx
+++ b/packages/manager/src/features/StackScripts/StackScriptPanel/StackScriptRow.tsx
@@ -9,11 +9,7 @@ import TableCell from 'src/components/TableCell/TableCell_CMR';
 import TableRow from 'src/components/TableRow/TableRow_CMR';
 import StackScriptsActionMenu from 'src/features/StackScripts/StackScriptPanel/StackScriptActionMenu';
 import { StackScriptCategory } from 'src/features/StackScripts/stackScriptUtils';
-import {
-  ClassNames,
-  displayTagsAndShowMore,
-  styles,
-} from '../StackScriptRowHelpers';
+import { ClassNames, styles } from '../StackScriptRowHelpers';
 
 export interface Props {
   label: string;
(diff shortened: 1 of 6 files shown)


packages/manager/src/features/StackScripts/StackScriptPanel/StackScriptRow.tsx
```diff
@@ -77,22 +73,37 @@ export const StackScriptRow: React.FC<CombinedProps> = (props) => {
...
         </TableCell>
-        <TableCell data-qa-stackscript-images className={classes.chip}>
-          {displayTagsAndShowMore(images)}
+      </Hidden>
+      <Hidden mdDown>
+        <TableCell data-qa-stackscript-images className={classes.images}>
+          {images.map((image, id) => {
+            return (
+              <div key={id} className={classes.image}>
+                {image}
```
Jskobos, line 102: I think `images.join(',')` would work just as well here and be easier to read

Jskobos (approved):
Seems weird that the images font size is smaller than the other cells 🤷

ndukatz (approved):

packages/manager/src/features/StackScripts/StackScriptRowHelpers.tsx
```diff
@@ -14,97 +11,76 @@ export type ClassNames =
...
       display: 'flex',
       flexWrap: 'wrap',
-      height: '100%',
       alignItems: 'center',
+      height: '100%',
       width: 70,
     },
     libRadioLabel: {
       cursor: 'pointer',
+      fontSize: '0.875rem',
```
DevDW, line 34: It could just be me, but the font size feels very small... maybe that's just because the font sizes have been so large in this section historically though. I tried 0.9rem for this and `libTitle` and bumped `libDescription` and `images` up a bit too and it felt better to me.

DevDW (approved):
Nice improvements for the responsiveness 👍  left one comment about the font size.

Also, in the Create flow, what do you think about possibly having "Show Details" shift to be under the description in the mobile view? That would prevent having to scroll to the right:

Screen Shot 2021-03-30 at 11 07 29 AM

packages/manager/src/features/StackScripts/StackScriptRowHelpers.tsx
tiffwong: Thanks for the feedback! I brought it up to Matthew and we decided to leave it as is for now since we want to prevent the row from wrapping to a third line for images 😬

PR #7529, opened 2021-03-24
Add GA event when clicking the Billing Docs link

johnwcallahan:
## Description

Add GA event when a customer clicks ~the "How Linode Billing Works" button~ all documentation buttons. ~To do this I added support for a `callback` function on DocsLink.~

~It's technically _not_ exactly a callback since the passed function is called before the link is opened in a new tab, so I'm open to suggestions on the naming/method.~


Jskobos (approved):
Event works, I left some 🤔 but good to go

johnwcallahan:
@Jskobos addressed your feedback... now these events are reported when a user clicks any documentation button.

Jskobos (approved):

DevDW (approved):

PR #7532, opened 2021-03-25
UI fixes for Linode pagination and graphs

tiffwong:
Removed the extra padding surrounding the pagination in the Linodes Landing group by tag view and fixed the alignment for the Linode Detail graphs


Jskobos (approved):
I can't tell exactly what changed in the graphs, but they look good and fewer grids is always great to see

DevDW (approved):

PR #7533, opened 2021-03-26
Make OBJ banner dismissible

Jskobos:
## Description

Same as Domains, it'll come back after 30 days


DevDW (approved):
Good to go, but the color scheme for dark mode needs adjusting

Screen Shot 2021-03-26 at 5 17 12 PM

johnwcallahan (approved):

PR #7534, opened 2021-03-26
Patch/hide interfaces in bad regions

Jskobos:
## Description

- Hide the interfaces section in the config modal if the Linode's region doesn't support vlans
- Remove interfaces from the payload in this case ^^^
- Conditionally add wrapper style for vlan input (only apply in create flow)

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
@@ -1,4 +1,10 @@
-import { Config, Disk, Interface, Kernel } from '@linode/api-v4/lib/linodes';
+import {
+  Config,
+  Disk,
+  Interface,
+  Kernel,
+  LinodeConfigCreationData,
+} from '@linode/api-v4/lib/linodes';
 import { APIError } from '@linode/api-v4/lib/types';
 import { Volume } from '@linode/api-v4/lib/volumes';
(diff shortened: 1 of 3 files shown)


ndukatz (approved):
Works well when testing against staging API

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
```diff
@@ -239,7 +253,11 @@ const LinodeConfigDialog: React.FC<CombinedProps> = (props) => {
 
     formik.setSubmitting(true);
 
-    const configData = convertStateToData(values);
+    const configData = convertStateToData(values) as LinodeConfigCreationData;
+
+    if (!regionHasVLANS) {
+      delete configData.interfaces;
```
DevDW, line 259: This works with no issues, but it looks like the `interfaces` property is still there in the payload somehow with an empty array 🤔 

Screen Shot 2021-03-26 at 1 20 46 PM

DevDW (approved):
Looks good; things are hidden in the situations they should be and visible in the situations they should be

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
Jskobos: That's the response 🤷  nothing we can do about that. I double checked the actual request payload and there's nothing there

PR #7535, opened 2021-03-26
Changes for VLAN attachment in Linode Create flow

DevDW:
## Description
Changes:
1. Revised `showVlans` logic to match what's in the Config modal
2. Pass selected region from AddonsPanel --> AttachVLAN --> InterfaceSelect so only VLANs for the given region populate the dropdown
3. New utility file for checking if a region supports VLANs
4. Only show "Attach a VLAN" section in Create flow if the selected region supports VLANs
5. Only include `interfaces` in the Create Linode payload if the selected region supports VLANs

## Type of Change

### DIFF
--- a/packages/manager/src/utilities/doesRegionSupportVLANs.ts
+++ b/packages/manager/src/utilities/doesRegionSupportVLANs.ts
@@ -1,0 +1,14 @@
+import { Region } from '@linode/api-v4/lib/regions';
+
+export const doesRegionSupportVLANs = (
+  region: string,
+  regionsData: Region[]
+) => {
+  const regionMetaData = regionsData.find((thisRegion) => {
+    return thisRegion.id === region;
+  });
+  if (!regionMetaData) {
+    return false;
+  }
+  return regionMetaData.capabilities.includes('Vlans');
+};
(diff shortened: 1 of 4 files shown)


packages/manager/src/utilities/doesRegionSupportVLANs.ts
```diff
@@ -0,0 +1,14 @@
+import { Region } from '@linode/api-v4/lib/regions';
```
Jskobos, line 1: Thanks for adding a utility! Once this release is done with, a fun enhancement here would be to use currying to make it more generic and widely useful, something like:
```
const doesRegionSupportVlans = doesRegionSupportFeature('Vlans');
const doesRegionSupportKubernetes = doesRegionSupportFeature('Kubernetes');
```

Jskobos (approved):
Flow and flagging both look good. 👍  
Screen Shot 2021-03-26 at 2 47 53 PM
I think we still need to make a style pass through here to make sure it looks right in both places.

DevDW:
> I think we still need to make a style pass through here to make sure it looks right in both places.

Screenshots of how they look as of the most recently-pushed change:

Create flow:
Screen Shot 2021-03-29 at 10 00 31 AM

Config modal:
Screen Shot 2021-03-29 at 9 51 29 AM

tiffwong (commented):
Everything looks great! Only thing I notice is that the the interface component in the Linode Create flow will cause the window to scroll when its < 400px instead of wrapping
Screen Shot 2021-03-31 at 10 33 05 AM

PR #7536, opened 2021-03-29
Regions query

Jskobos:
## Description

Follow-up to a previous effort to replace our Redux wrapper around Regions with React-Query. This time around we completely remove the Redux stuff and (🤞 ) the app works as before.

Notes:

- I left the regions.container in place, only now it's an HOC that injects ReactQuery stuff into the passed component. This is (as documented) for class components that we can't/won't refactor to FCs. It shouldn't be something we frequently fall back on. We should make a WithQuery wrapper that works for all of our containers, but for this PR it was easiest to do something that left existing usage of withRegions unchanged.
- Only a single request to Regions should be made on app load (in production there are two, one from the legacy Redux implementation and one from React-Query.


packages/manager/src/features/Kubernetes/ClusterList/ClusterRow.tsx
```diff
@@ -122,7 +123,9 @@ export const ClusterRow: React.FunctionComponent<CombinedProps> = (props) => {
           <DateTimeDisplay value={cluster.created} />
         </TableCell>
       </Hidden>
-      <TableCell data-qa-cluster-region>{cluster.region}</TableCell>
+      <TableCell data-qa-cluster-region>
+        {dcDisplayNames[cluster.region] ?? cluster.region}
```
Jskobos, line 127: This is unrelated but it's a fix we've discussed, this was a long-ago oversight.

packages/manager/src/features/NodeBalancers/NodeBalancerCreate.tsx
```diff
@@ -817,14 +817,6 @@ const mapStateToProps: MapState<StateProps, CombinedProps> = (state) => ({
 
 const connected = connect(mapStateToProps);
 
-const withRegions = regionsContainer(({ data, loading, error }) => ({
-  regionsData: data
-    .filter((region) => region.id !== 'ap-northeast-1a') // Don't show Tokyo1 as an option
```
Jskobos, line 822: This hasn't been returned for a very long time (Tokyo 1 hasn't existed in ages)

johnwcallahan (approved):
The code looks good, and I have not been able to break it.

PR #7538, opened 2021-03-29
M3-4961 Change: Hide unneeded fields on LinodeDetail for bare metal plans

Jskobos:
## Description

This combines M3-4958, M3-4960, and M3-4961, with some remaining work (but this was getting out of hand so I'm splitting it up):
- Add isThisBareMetal? logic to LinodeDetailNavigation
- Hide configs, backups, and storage tabs if ^^^ is true
- Hide CPU and Disk IO alerts in the /settings tab for Bare Metal (and remove them from the payload)
- Hide Resize, Migrate, and Clone actions from the action menu for BM linodes (in both detail and list view)

This is where things got a bit dicey, we were passing different varieties and shapes of types around the Linodes flow in ways that were hard to track down. I tried to make things more consistent; might want to try again and see if flattening out the props is workable. There was also a bunch of old VLANs stuff (from back when there was a vlanDetail view with a LinodesLanding table embedded in it) that was making things confusing so I got rid of it. It's possible that vlanDetail will come back, though it would have to be pretty different (there's no vlan/id/detach or /attach methods anymore, which were sort of the point of that table).

Still to come:
- Root password form update
- M3-4959 Graphs on LinodeDetail/summary


packages/manager/src/features/linodes/LinodesLanding/LinodeActionMenu.tsx
```diff
@@ -150,17 +162,6 @@ export const LinodeActionMenu: React.FC<CombinedProps> = (props) => {
     );
   };
 
-  const {
```
Jskobos, line 153: don't need two of these props destructurings

DevDW (approved):
Everything LGTM -- I didn't observe any regressions, and the cleanup of the old VLAN logic definitely improves readability.

tiffwong (approved):
LGTM

PR #7543, opened 2021-03-31
M3-5010 Fix: IP transfer not showing all options

Jskobos:
## Description

When loading LinodeDetails directly, we only request a single page of 100 Linodes (for our "is this a big account?" checks). If a large account tries to share or swap IP addresses, in some cases only a few Linodes in the target region are included in the 100-Linode response, with no way for the user to view other options.

To get around this, I removed reliance on the Redux state in IP swap and transfer modals  and used the new Linodes query. It's based on the region and a search term. How it should work is:

1. user opens Transfer modal (or sharing, same effect)
2. A single request is made to get Linodes in the current region (max size of 500)
3. If the user has > 500 Linodes in the region, they can type in the input, at which point we make a debounced request to retrieve Linodes with a matching label.

We still should circle back to the sharing modal, and use networking/ips instead of Linodes, but I felt this solution was already a big improvement over getAll(). In addition, it's a separate loading state which doesn't trigger the LinodeDetail-level spinner.


johnwcallahan (approved):
I tested this pretty thoroughly, haven't been able to find any issues.

ndukatz (approved):

PR #7545, opened 2021-03-31
Allow creating a Linode without an image

Jskobos:
When creating a Linode without an image, interfaces are not allowed.
Remove interfaces from the payload in this case to avoid an error
from the API. Disable the VLAN select and show helper text to explain
what needs to be done.


johnwcallahan (commented):
Could we conditionally give the `vlanLabel` only if the VLAN input isn't disabled? 

Currently you can get yourself into this state which makes it look like you're creating a Linode with a VLAN:
Screen Shot 2021-03-31 at 2 39 41 PM

johnwcallahan (approved):

tiffwong (approved):

PR #7550, opened 2021-04-01
M3-4961 Change: Separate password change flow for bare metal Linodes

Jskobos:
## Description

Bare metal Linodes don't have disks that are managed through the API, so they have to use a new endpoint (linode/instances/:id/password) to reset their root password. In planning I recommended using a separate component for this to avoid a bunch of prop-driven stuff, but it turns out there's a lot of shared logic between the two flows that would be repeated and I couldn't think of a great way around it. So I kept the existing stuff in place and a) hid the disks dropdown then b) used a different method on submit.

### DIFF
--- a/packages/api-v4/src/linodes/linodes.ts
+++ b/packages/api-v4/src/linodes/linodes.ts
@@ -128,7 +128,7 @@ export const deleteLinode = (linodeId: number) =>
  * through the API.
  */
 
-export const resetLinodePassword = (linodeId: number, root_pass: string) =>
+export const changeLinodePassword = (linodeId: number, root_pass: string) =>
   Request<{}>(
     setURL(`${API_ROOT}/linode/instances/${linodeId}/password`),
     setData({ root_pass }),
--- a/packages/api-v4/src/linodes/linodes.ts
+++ b/packages/api-v4/src/linodes/linodes.ts
@@ -121,7 +121,7 @@ export const deleteLinode = (linodeId: number) =>
 /**
  * resetLinodePassword
  *
- * This method is distinct from resetLinodeDiskPassword,
+ * This method is distinct from changeLinodeDiskPassword,
  * in that it resets the root password on the Linode itself
  * rather than on a specific disk. This situation only applies
  * to bare metal instances, which don't have disks that are managed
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeSettingsPasswordPanel.tsx
```diff
@@ -73,27 +74,34 @@ class LinodeSettingsPasswordPanel extends React.Component<
...
+          set(lensPath(['errors']), errors),
+          set(lensPath(['submitting']), false)
+        ),
+        () => {
+          scrollErrorIntoView('linode-settings-password');
+        }
+      );
+    };
+
+    if (isBareMetalInstance) {
```
johnwcallahan, line 99: Missing resetLinodePassword call here?

johnwcallahan (commented):
I'm unable to test this right now since I don't have any bare metal instances, but the code looks OK aside from one thing.

tiffwong (approved):
Tested with mocks, everything looks good

packages/api-v4/src/linodes/linodes.ts
```diff
@@ -117,3 +117,20 @@ export const deleteLinode = (linodeId: number) =>
...
+ * rather than on a specific disk. This situation only applies
+ * to bare metal instances, which don't have disks that are managed
+ * through the API.
+ */
+
+export const changeLinodePassword = (linodeId: number, root_pass: string) =>
+  Request<{}>(
+    setURL(`${API_ROOT}/linode/instances/${linodeId}/password`),
+    setData({ root_pass }),
+    setMethod('DELETE')
```
DevDW, line 135: Should this be a `POST`?

packages/api-v4/src/linodes/linodes.ts
```diff
@@ -117,3 +117,20 @@ export const deleteLinode = (linodeId: number) =>
     setURL(`${API_ROOT}/linode/instances/${linodeId}`),
     setMethod('DELETE')
   );
+
+/**
+ * resetLinodePassword
+ *
+ * This method is distinct from resetLinodeDiskPassword,
```
DevDW, line 124: `changeLinodeDiskPassword` ?

DevDW (approved):
LGTM aside from a couple of comments

PR #7551, opened 2021-04-01
M3-5016: Fix wrapping on Firewall Rule table

tiffwong:
Prevent the label column and the "Port Range" header from wrapping on top of dropping the protocol column sooner


DevDW (approved):

ndukatz (approved):
Looks good

PR #7560, opened 2021-04-05
M3-4969: Convert StackScripts Drawer to Dialog

tiffwong:
Convert the StackScript drawer that opens when the "Show Details" button is clicked on the Linode Create flow via StackScripts into a full page dialog

I also refactored the component while I was in there


packages/manager/src/features/StackScripts/StackScriptDialog.tsx
```diff
@@ -0,0 +1,90 @@
...
+      maxWidth="md"
+    >
+      {loading ? (
+        <CircleProgress />
+      ) : (
+        <>
+          {error && (
+            <Notice error text="There was an error loading this StackScript." />
+          )}
+          {stackScript && <_StackScript data={stackScript} />}
```
Jskobos, line 69: This is fine, a possible improvement would be to make the dialog equivalent of DrawerContent now that we're using more dialogs and fewer drawers. I did something sort of like this for LinodeConfigDialog, maybe we could extract that to src/components and make it more general? This pattern is going to keep coming up.

Jskobos (approved):

DevDW (approved):
🚢

packages/manager/src/features/StackScripts/StackScriptDialog.tsx
tiffwong: Will address this in a follow-up PR

PR #7561, opened 2021-04-06
For Release: Schema and typing revisions to allow interface.label to be null

DevDW:
## Description
There was a bug where you could not successfully edit a configuration that had a VLAN interface. That was due to some incorrect validation. This PR revises the schema and allows `null` to be a valid type for `interface.label` in several places.

## Type of Change

## Note to Reviewers
Please ensure that Linode configurations with and without VLANs can be edited and created as expected, and that Linodes can be created with or without VLANs.

### DIFF
--- a/packages/api-v4/src/linodes/linodes.schema.ts
+++ b/packages/api-v4/src/linodes/linodes.schema.ts
@@ -1,16 +1,19 @@
 import { array, boolean, mixed, number, object, string } from 'yup';
 import { parseCIDR } from 'ipaddr.js';
 
-const validateIP = (ipAddress: string) => {
+const validateIP = (ipAddress: string | null) => {
   if (ipAddress === '') {
     // ipam_address is technically required, but empty strings are valid
     return true;
   }
-  // We accept IP ranges (i.e., CIDR notation).
-  try {
-    parseCIDR(ipAddress);
-  } catch (err) {
-    return false;
+
+  if (ipAddress !== null) {
+    // We accept IP ranges (i.e., CIDR notation).
+    try {
+      parseCIDR(ipAddress);
+    } catch (err) {
+      return false;
+    }
   }
 
   return true;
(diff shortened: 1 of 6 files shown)


ndukatz (approved):
Looks good, Seems to be working well now

packages/api-v4/src/linodes/linodes.schema.ts
```diff
@@ -1,16 +1,19 @@
...
     // ipam_address is technically required, but empty strings are valid
     return true;
   }
-  // We accept IP ranges (i.e., CIDR notation).
-  try {
-    parseCIDR(ipAddress);
-  } catch (err) {
-    return false;
+
+  if (ipAddress !== null) {
```
Jskobos, line 10: This can be combined with the preceding if statement

johnwcallahan, line 10: I'd instead combine this with the condition on L5, e.g. `if (!ipAddress)`. We should also remove the comment that says ipam_address is required.

packages/manager/src/features/linodes/LinodesDetail/LinodeAdvanced/ConfigRow.tsx
```diff
@@ -109,7 +109,7 @@ export const ConfigRow: React.FC<CombinedProps> = (props) => {
 
         return (
           <li
-            key={interfaceEntry.label + idx}
+            key={interfaceEntry.purpose + idx}
```
Jskobos, line 112: Unlikely this will affect anything but I think `interfaceEntry.label ?? 'public'` is a bit safer

packages/api-v4/src/linodes/linodes.schema.ts
```diff
@@ -24,19 +27,21 @@ export const linodeInterfaceSchema = array().of(
...
+          .min(1, 'Label must be between 1 and 64 characters.')
+          .max(64, 'Label must be between 1 and 64 characters.')
+          .matches(
+            /[a-z0-9-]+/,
+            'Interface labels cannot contain special characters.'
+          ),
+        otherwise: string().notRequired(),
+      })
+      .nullable(true),
+    ipam_address: string().nullable().test({
```
Jskobos, line 44: Is this `nullable()` call doing anything since null is already handled by the test?

johnwcallahan (approved):

tiffwong (approved):

packages/api-v4/src/linodes/linodes.schema.ts
DevDW: did some testing without it and it doesn't seem to be needed now so I removed it

PR #7563, opened 2021-04-06
M3-4985: StackScripts Details breadcrumb alignment

tiffwong:
Shift the username up 1px so that it's aligned with "StackScripts" and the title

Also refactored the component for funsies and fixed the background color

### DIFF
--- a/packages/manager/src/features/StackScripts/StackScriptsDetail.tsx
+++ b/packages/manager/src/features/StackScripts/StackScriptsDetail.tsx
@@ -21,12 +21,12 @@ import DocumentationButton from 'src/components/DocumentationButton';
 import Grid from 'src/components/Grid';
 import NotFound from 'src/components/NotFound';
 import _StackScript from 'src/components/StackScript';
-import withProfile from 'src/containers/profile.container';
 import { StackScripts as StackScriptsDocs } from 'src/documentation';
 import {
   hasGrant,
   isRestrictedUser as _isRestrictedUser,
 } from 'src/features/Profile/permissionsHelpers';
+import useProfile from 'src/hooks/useProfile';
 import { MapState } from 'src/store/types';
 import { getAPIErrorOrDefault, getErrorMap } from 'src/utilities/errorUtils';
 import { getStackScriptUrl } from './stackScriptUtils';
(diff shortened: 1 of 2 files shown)


packages/manager/src/features/StackScripts/StackScriptsDetail.tsx
```diff
@@ -39,147 +31,89 @@ import { MapState } from 'src/store/types';
...
+        setLoading(false);
+        setStackScript(stackScript);
       })
       .catch((error) => {
-        this.setState({ error, loading: false });
+        setLoading(false);
+        setErrors(error);
       });
-  }
+  }, [props.match.params]);
```
johnwcallahan, line 114: This is causing an extra function call/request since props.match.params is an object and the equality check doesn't do what we want.

If you destructure stackScriptId above the useEffect definition, then include `stackScriptId` as a dependency (instead of props.match.params) it should do what you want.

packages/manager/src/features/StackScripts/StackScriptsDetail.tsx
```diff
@@ -561,8 +250,7 @@ const enhanced = compose<CombinedProps, {}>(
   setDocs([StackScriptsDocs]),
   withProfile<ProfileProps, {}>((ownProps, { profileData: profile }) => ({
     username: profile?.username,
-  })),
-  withStyles(styles)
+  }))
```
johnwcallahan, line 253: While we're at it we could refactor this to make use of e.g. `useProfile` hooks instead of containers. You don't have to do this now, though.. could open up a tech ticket.

johnwcallahan (changes requested):
Thanks for the refactor – really nice work. Left a few minor comments; I think the one about `useEffect` is the only one required before merge.

DevDW (commented):
Looking good overall; in addition to John's feedback, I also noticed that very long StackScript names push the Docs & "Deploy New Linode" and the Edit buttons to new lines:

Screen Shot 2021-04-08 at 11 28 27 AM

"Deploy New Linode" is right up against the main content section, so that probably calls for some bottom spacing, and the Edit button has a left indentation

johnwcallahan (approved):
Great work on this, thanks for getting those little refactors in.

DevDW (approved):
🚢

PR #7565, opened 2021-04-06
M3-5040 Add: Delete cluster button/modal in LKE detail view

Jskobos:
## Description

Currently it's only possible to delete a cluster from the action menu on the cluster list view. This adds a button to do the same thing from a cluster's detail view.

Also included in the ticket is restyling the tags section of the summary panel to match what's used in Linode detail view. This is tricky since it's the same component used in DomainRecordWrapper. I don't love what I did here, I tried to match the appearance and behavior of the TagCell component, which we can't use here because it requires a fixed width to do its fade-out magic. Discussed this with Matthew and decided on this approach for now.

### DIFF
--- a/packages/manager/src/components/Tag/Tag_CMR.tsx
+++ b/packages/manager/src/components/Tag/Tag_CMR.tsx
@@ -10,6 +10,7 @@ import {
   withStyles,
   WithStyles,
 } from 'src/components/core/styles';
+import { truncateEnd } from 'src/utilities/truncate';
 
 type Variants =
   | 'white'
@@ -157,6 +158,7 @@ export interface Props extends ChipProps {
   asSuggestion?: boolean;
   closeMenu?: any;
   component?: string;
+  maxLength?: number;
 }
 
 type CombinedProps = Props & RouteComponentProps<{}> & WithStyles<CSSClasses>;
@@ -181,18 +183,24 @@ class Tag extends React.Component<CombinedProps, {}> {
       colorVariant,
       classes,
       className,
+      label,
       history,
       location,
       staticContext,
       match, // Don't pass route props to the Chip component
       asSuggestion,
       closeMenu,
+      maxLength,
       ...chipProps
     } = this.props;
+
+    // If maxWidth is set, truncate display to that length.
+    const _label = maxLength ? truncateEnd(label, maxLength) : label;
 
     return (
       <Chip
         {...chipProps}
+        label={_label}
         className={classNames({
           ...(className && { [className]: true }),
           [classes[colorVariant!]]: true,
--- a/packages/manager/src/components/Tag/Tag_CMR.tsx
+++ b/packages/manager/src/components/Tag/Tag_CMR.tsx
@@ -26,7 +26,9 @@ type CSSClasses = 'label' | 'root' | 'deleteButton' | Variants;
 
 const styles = (theme: Theme) =>
   createStyles({
-    label: {},
+    label: {
+      maxWidth: 350,
+    },
     root: {
       height: 30,
       paddingLeft: 0,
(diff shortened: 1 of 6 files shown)


packages/manager/src/components/Tag/Tag_CMR.tsx
```diff
@@ -26,7 +26,9 @@ type CSSClasses = 'label' | 'root' | 'deleteButton' | Variants;
 
 const styles = (theme: Theme) =>
   createStyles({
-    label: {},
+    label: {
+      maxWidth: 350,
```
Jskobos, line 30: This ought to help un-break the layout in places, but it's up to the consumer to truncate the text of extremely long tag names. I tried doing this via CSS but that requires a fixed width, which tags don't have.

packages/manager/src/components/TagsPanel/TagsPanelRedesigned.tsx
```diff
@@ -410,21 +413,21 @@ class TagsPanelRedesigned extends React.Component<CombinedProps, State> {
...
-                  key={eachTag}
-                  label={eachTag}
-                  tagLabel={eachTag}
-                  onDelete={disabled ? undefined : this.handleDeleteTag}
+                <Tag
+                  key={`tag-item-${thisTag}`}
+                  colorVariant="lightBlue"
+                  label={truncateEnd(thisTag, 30)}
+                  style={
+                    this.props.align === 'left'
```
Jskobos, line 423: @tiffwong this works ok but I feel ashamed to have written it. Do you have any better ideas?

tiffwong: I added the styles to the LKE panel instead which should do the trick

tiffwong (approved):

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeSummaryPanel.tsx
```diff
@@ -15,105 +15,120 @@ import MiniKube from 'src/assets/icons/mini-kube.svg';
...
+        justifyContent: 'flex-start !important',
+      },
+      // Tags Panel wrapper
+      '& > div:last-child': {
+        marginBottom: 2,
+      },
     },
-  });
+  },
+  deleteButton: {
```
DevDW, line 124: Not sure if the ticket is referring to another button, but it seems to be calling for, on hover, a solid blue background w/ white text?

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeSummaryPanel.tsx
```diff
@@ -390,14 +414,29 @@ export const KubeSummaryPanel: React.FunctionComponent<CombinedProps> = (
...
+                  tags={cluster.tags}
+                  updateTags={handleUpdateTags}
+                />
+              </Grid>
+            </Grid>
           </Grid>
+
+          <button
+            className={classes.deleteButton}
+            onClick={() => openDialog(cluster.id)}
```
DevDW, line 436: Should we be redirecting back to the Kubernetes landing page as well? Right now it just hangs on an empty page for the now-deleted cluster

DevDW (commented):
Cluster deletion is working; left a few comments on secondary functionality and styling points

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeSummaryPanel.tsx
Jskobos: I saw that too, but when I showed Matthew this over Slack he said it should be a plain text link. I'll ask him to check out the final version and can adjust if needed.

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeSummaryPanel.tsx
Jskobos: Yes, we definitely should.

DevDW (approved):
🚢

PR #7566, opened 2021-04-06
M3-4970: Linode Details graphs UI refinements

tiffwong:
Changes include:
- New title
- Gray background
- Remove "Toggle Graphs" label and icon (aka the caterpillar)
- Abbreviate legend labels and decrease font size
- Decrease graph height

Also updated the dark mode colors used in the graphs that previously blended in due to the transparency

The legend converts to a smaller view too quickly imo and the dark mode colors (border vs background color) aren't ideal but everything is legible so I'll have a separate PR address these two issues

Before:
Screen Shot 2021-04-06 at 4 13 07 PM

After:
Screen Shot 2021-04-09 at 5 14 16 PM

### DIFF
--- a/packages/manager/src/components/LineGraph/LineGraph.tsx
+++ b/packages/manager/src/components/LineGraph/LineGraph.tsx
@@ -140,14 +140,23 @@ const LineGraph: React.FC<CombinedProps> = (props: CombinedProps) => {
       scales: {
         yAxes: [
           {
+            // Defines a fixed width for the Y-axis labels
+            afterFit(axes) {
+              axes.width = 32;
+            },
             gridLines: {
               borderDash: [3, 6],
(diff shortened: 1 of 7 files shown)


Jskobos (commented):
1. Unrelated: the tags input cell totally breaks everything in this view if there are a lot of tags and you're viewing at a lower screen width
2. Maybe it's just my monitor but the legends look more spaced out for me than in your screenshot:
Screen Shot 2021-04-07 at 8 31 22 AM

It looks a bit wasteful, the graph y axis labels are smooshed together but half the height of the panel is taken up by luxuriously spread out legend entries

ndukatz (approved):
looks good!

Jskobos (approved):

PR #7567, opened 2021-04-06
M3-5021: Domains related UI stuff

tiffwong:
Left align the inline "Enable/Disable" actions and change the "Delete" button in the Edit drawer

### DIFF
--- a/packages/manager/src/features/Domains/DomainActionMenu.tsx
+++ b/packages/manager/src/features/Domains/DomainActionMenu.tsx
@@ -12,6 +12,7 @@ import InlineMenuAction from 'src/components/InlineMenuAction';
 
 const useStyles = makeStyles(() => ({
   button: {
+    justifyContent: 'flex-start',
     minWidth: 66,
   },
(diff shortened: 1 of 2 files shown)


Jskobos (approved):

packages/manager/src/features/Domains/DomainActionMenu.tsx
```diff
@@ -52,7 +66,7 @@ export const DomainActionMenu: React.FC<CombinedProps> = (props) => {
     onClone(domain, id);
   };
 
-  const actions: Action[] = [
+  const actions = [
```
DevDW, line 69: Instead of removing the type here, maybe add something like https://github.com/linode/manager/blob/3477c69b6dbfb76b7c3a642eca4d5900e6df678e/packages/manager/src/features/linodes/LinodesLanding/LinodeActionMenu.tsx#L330 

and type this as it, that way the `className` will be accounted for

DevDW (approved):

PR #7570, opened 2021-04-08
Assorted VLAN feedback

johnwcallahan:
## Description

M3-5049: In Config Dialog, save button should be labeled "Save Changes"
M3-5051: In the Config Dialog, the targets for toggles are too wide
M3-5052: In the Linode Create flow and Config Dialog, change VLAN "Label" field to "VLAN"
M3-5055: Remove (or disable) "None" as an option for eth0 network interface

One thing I haven't addressed in M3-5052 is that the error message for this field will still say e.g. "Label is required". I didn't want to make this change in the API wrapper / validation layer since this is Cloud-specific. Still thinking about how to best do this.

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/InterfaceSelect.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/InterfaceSelect.tsx
@@ -153,7 +153,7 @@ export const InterfaceSelect: React.FC<Props> = (props) => {
         <Grid item xs={12} sm={6}>
           <Select
             options={
-              // Do not display "None" as an option for eth0, since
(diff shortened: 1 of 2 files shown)


ndukatz (commented):
I am seeing this error when trying to create a linode on my personal account so I'm not sure if that's expected or not. but everything looks good on the vlan account I tested on
Screen Shot 2021-04-08 at 4 59 00 PM

ndukatz:
> I am seeing this error when trying to create a linode on my personal account (no vlans) so I'm not sure if that's expected or not. but everything looks good on the vlan account I tested on
> Screen Shot 2021-04-08 at 4 59 00 PM

Never mind, I see that was caused by a different PR.

ndukatz (approved):

DevDW (approved):

PR #7572, opened 2021-04-08
Fix for linodeInterfaceSchema test

DevDW:
## Description
Seems that the Linode Create flow was broken on `develop` because the linodeInterfaceSchema test was failing when there were no interfaces. This change addresses that case 

## Type of Change

## Note to Reviewers
Please confirm that Linode creation and editing and Linode Configuration creation and editing with and without VLANs function as intended


johnwcallahan (approved):
Thanks for the speedy fix!

Jskobos (approved):

PR #7575, opened 2021-04-09
M3-5045: Images Create page

DevDW:
## Description
- Create new "Images / Create" page at `/images/create`
- Make "/images*" to "/images" redirect & Create button behavior conditional on `machineImagesEnabled`

## Type of Change

## To-Do:
- [x] Tabs
- [x] Cleanup

### DIFF
--- a/packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
+++ b/packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
@@ -46,10 +46,13 @@ const styles = (theme: Theme) =>
     },
   });
 
+export interface Props {
+  label?: string;
+  description?: string;
+  changeLabel: (e: React.ChangeEvent<HTMLInputElement>) => void;
+  changeDescription: (e: React.ChangeEvent<HTMLInputElement>) => void;
+}
 interface State {
-  description: string;
-  imageID: string;
-  label: string;
   selectedLinode: number;
   selectedDisk: string | null;
   disks: Disk[];
@@ -58,7 +61,8 @@ interface State {
   submitting: boolean;
 }
 
-type CombinedProps = WithStyles<ClassNames> &
+type CombinedProps = Props &
+  WithStyles<ClassNames> &
   RouteComponentProps<{}> &
   WithSnackbarProps &
   ProfileProps &
@@ -67,9 +71,6 @@ type CombinedProps = WithStyles<ClassNames> &
 class CreateImageTab extends React.Component<CombinedProps, State> {
   mounted: boolean = false;
   state = {
-    description: '',
-    imageID: '',
-    label: '',
     selectedLinode: 0,
     selectedDisk: '',
     disks: [],
@@ -148,25 +149,16 @@ class CreateImageTab extends React.Component<CombinedProps, State> {
     this.changeSelectedDisk(diskID);
   };
 
-  setLabel = (e: React.ChangeEvent<HTMLInputElement>) => {
-    const value = e.target.value;
-
-    this.setState({
-      label: value,
-    });
-  };
-
-  setDescription = (e: React.ChangeEvent<HTMLInputElement>) => {
-    const value = e.target.value;
-    this.setState({
-      description: value,
-    });
-  };
-
   onSubmit = () => {
-    const { history, createImage, enqueueSnackbar } = this.props;
+    const {
+      history,
+      createImage,
+      enqueueSnackbar,
+      label,
+      description,
+    } = this.props;
 
-    const { label, description, selectedDisk } = this.state;
+    const { selectedDisk } = this.state;
 
     this.setState({ errors: undefined, notice: undefined, submitting: true });
     const safeDescription = description ?? '';
@@ -219,14 +211,20 @@ class CreateImageTab extends React.Component<CombinedProps, State> {
   };
 
   render() {
-    const { availableLinodesToImagize, canCreateImage, classes } = this.props;
+    const {
+      availableLinodesToImagize,
+      canCreateImage,
+      classes,
+      label,
+      description,
+      changeLabel,
+      changeDescription,
+    } = this.props;
     const {
       disks,
       errors,
       notice,
       submitting,
-      label,
-      description,
       selectedLinode,
       selectedDisk,
     } = this.state;
@@ -304,7 +302,7 @@ class CreateImageTab extends React.Component<CombinedProps, State> {
           <TextField
             label="Label"
             value={label}
-            onChange={this.setLabel}
+            onChange={changeLabel}
             error={Boolean(labelError)}
             errorText={labelError}
             disabled={!canCreateImage}
@@ -316,7 +314,7 @@ class CreateImageTab extends React.Component<CombinedProps, State> {
             multiline
             rows={4}
             value={description}
-            onChange={this.setDescription}
+            onChange={changeDescription}
             error={Boolean(descriptionError)}
             errorText={descriptionError}
             disabled={!canCreateImage}
@@ -350,7 +348,7 @@ interface ProfileProps {
   availableLinodesToImagize: number[] | null;
 }
 
-export default compose<CombinedProps, {}>(
+export default compose<CombinedProps, Props>(
   styled,
   withRouter,
   withImages(),
--- a/packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
+++ b/packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
@@ -219,7 +219,7 @@ class CreateImageTab extends React.Component<CombinedProps, State> {
   };
 
   render() {
-    const { availableImages, canCreateImage, classes } = this.props;
+    const { availableLinodesToImagize, canCreateImage, classes } = this.props;
     const {
       disks,
       errors,
@@ -267,14 +267,16 @@ class CreateImageTab extends React.Component<CombinedProps, State> {
           disabled={!canCreateImage}
           handleChange={(linode) => this.handleLinodeChange(linode.id)}
           filterCondition={(linode) =>
-            availableImages ? availableImages.includes(linode.id) : true
+            availableLinodesToImagize
+              ? availableLinodesToImagize.includes(linode.id)
+              : true
           }
           updateFor={[
             selectedLinode,
             linodeError,
             classes,
             canCreateImage,
-            availableImages,
+            availableLinodesToImagize,
           ]}
         />
 
@@ -345,7 +347,7 @@ const styled = withStyles(styles);
 
 interface ProfileProps {
   canCreateImage: boolean;
-  availableImages: number[] | null;
+  availableLinodesToImagize: number[] | null;
 }
 
 export default compose<CombinedProps, {}>(
@@ -361,7 +363,7 @@ export default compose<CombinedProps, {}>(
     // Unrestricted users can create Images from any disk;
     // Restricted users need read_write on the Linode they're trying to Imagize
     // (in addition to the global add_images grant).
-    availableImages: profile?.restricted
+    availableLinodesToImagize: profile?.restricted
       ? profile?.grants?.linode
           .filter((thisGrant) => thisGrant.permissions === 'read_write')
           .map((thisGrant) => thisGrant.id) ?? []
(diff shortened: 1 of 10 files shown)


packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
```diff
@@ -0,0 +1,355 @@
...
+  componentDidMount() {
+    this.mounted = true;
+  }
+
+  componentWillUnmount() {
+    this.mounted = false;
+  }
+
+  componentDidUpdate(prevProps: CombinedProps) {
+    if (!this.props.selectedLinode && prevProps.selectedLinode) {
```
Jskobos, line 93: This stuff is spooky and we should avoid it if we can. I know it's preexisting, though I can't remember why we (I) decided to manage drawer state in ImagesLanding instead of the actual drawer. See below

packages/manager/src/features/Images/ImagesCreate/ImageCreate.tsx
```diff
@@ -0,0 +1,130 @@
...
+    return Boolean(matchPath(p, { path: props.location.pathname }));
+  };
+
+  const navToURL = (index: number) => {
+    setRoute(props.location.pathname.replace('/create/', ''));
+    props.history.push(tabs[index].routeName);
+  };
+
+  // Logic for <CreateImageTab /> props
+  const defaultCreateTabState = {
```
Jskobos, line 52: I'm not sure about this! Looks like machine images use a different API endpoint, but label and description are still part of the payload. Bundling them together with selected disk makes the update methods more complicated and makes it harder to reuse the state in the other tab.

A simple improvement would be to make these 3 separate state variables. 2 strings and 1 number isn't too many to juggle. Another possibility would be to manage Linode and disk selection entirely inside the CreateImageTab component, since that's the only place where they'll be used. That would also let you get rid of the componentDidUpdate stuff I mentioned above.

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
```diff
@@ -0,0 +1,355 @@
...
+
+interface ProfileProps {
+  canCreateImage: boolean;
+  availableImages: number[] | null;
+}
+
+export default compose<CombinedProps, Props>(
+  styled,
+  withRouter,
+  withImages(),
```
Jskobos, line 339: Are these injected props being used in here? It's possible but I can't really see where

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
```diff
@@ -0,0 +1,355 @@
...
+  withSnackbar,
+  SectionErrorBoundary,
+  withProfile<ProfileProps, {}>((undefined, { profileData: profile }) => ({
+    canCreateImage:
+      Boolean(!profile?.restricted) ||
+      Boolean(profile?.grants?.global.add_images),
+    // Unrestricted users can create Images from any disk;
+    // Restricted users need read_write on the Linode they're trying to Imagize
+    // (in addition to the global add_images grant).
+    availableImages: profile?.restricted
```
Jskobos, line 349: This looks like a bad name from way back when. It's a list of Linodes, not Images.

packages/manager/src/features/Images/ImagesCreate/ImageCreate.tsx
```diff
@@ -0,0 +1,130 @@
...
+    },
+    {
+      title: 'Upload Image',
+      routeName: `${props.match.url}/upload-image`,
+    },
+  ];
+
+  React.useEffect(() => {
+    setRoute(props.location.pathname.replace('/create/', ''));
+  }, [route, props.location.pathname]);
```
Jskobos, line 40: What is this doing?

packages/manager/src/features/Images/ImagesCreate/ImageCreate.tsx
```diff
@@ -0,0 +1,130 @@
...
+  description?: string;
+  selectedDisk: string | null;
+  selectedLinode?: number;
+}
+
+const CreateImageTab = React.lazy(() => import('./CreateImageTab'));
+const ImageUpload = React.lazy(() => import('../ImageUpload'));
+
+export const ImageCreate: React.FC<CombinedProps> = (props) => {
+  const [route, setRoute] = React.useState<string>('');
```
Jskobos, line 24: This `route` variable is only used as a dependency in the useEffect below. What is its purpose?

packages/manager/src/features/Images/ImagesCreate/ImageCreate.tsx
```diff
@@ -0,0 +1,130 @@
...
+  React.useEffect(() => {
+    setRoute(props.location.pathname.replace('/create/', ''));
+  }, [route, props.location.pathname]);
+
+  const matches = (p: string) => {
+    return Boolean(matchPath(p, { path: props.location.pathname }));
+  };
+
+  const navToURL = (index: number) => {
+    setRoute(props.location.pathname.replace('/create/', ''));
```
Jskobos, line 47: Same comment, when I delete all of the setRoute stuff I can't see a change in behavior. Which case is this handling?

Jskobos (changes requested):
Good work on this, I think the MainContent routing should be fixed (unless I'm missing something), otherwise most of what I've left is cleanup.

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
DevDW: The `createImage` method is coming from this. I commented this line out and creation failed because `createImage` was undefined

packages/manager/src/features/Images/ImagesCreate/ImageCreate.tsx
DevDW: I was using `ObjectStorageLanding.tsx` as a model and thought these pieces were necessary, but it seems they aren't. The behavior looks unchanged without them to me as well. I removed these pieces

Jskobos (approved):

packages/manager/src/features/Images/ImagesCreate/ImageCreate.tsx
```diff
@@ -0,0 +1,80 @@
...
+    const value = e.target.value;
+    setLabel(value);
+  };
+
+  const handleSetDescription = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const value = e.target.value;
+    setDescription(value);
+  };
+
+  const tabs = [
```
johnwcallahan, line 29: We've been using the `<NavTabs />` pattern, which would clean up this component quite a bit. See Profile.tsx for example usage. I believe that would also fix the following bug: 

1. Go to `/images/create/random-bad-path`
2. Observe: it _doesn't_ redirect you to the URL of the first tab.

packages/manager/src/features/Images/ImagesCreate/ImageCreateContainer.tsx
```diff
@@ -0,0 +1,25 @@
...
+      <DocumentTitleSegment segment="Create an Image" />
+      <Grid container spacing={0} className="m0">
+        <Grid item xs={12} className="p0">
+          <Breadcrumb
+            pathname={'/images/create'}
+            labelTitle="Create"
+            data-qa-create-image-header
+          />
+        </Grid>
+        <ImageCreate />
```
johnwcallahan, line 19: Followup: check with Matthew and the docs team, but we should probably put a Docs link here like we usually do.

packages/manager/src/features/Images/index.tsx
```diff
@@ -6,26 +6,42 @@ import {
...
+  );
+
+  return (
+    <React.Suspense fallback={<SuspenseLoader />}>
+      <Switch>
+        <Route component={ImagesLanding} path={path} exact />
+        {machineImagesEnabled ? (
+          <Route component={ImageCreate} path={`${path}/create`} />
+        ) : null}
+      </Switch>
```
johnwcallahan, line 41: We should add a redirect so that junk URLs take you back to the landing page, (like "/images/bad-path")

```tsx
<Redirect to="/images" />
```

packages/manager/src/features/Images/ImagesCreate/ImageCreate.tsx
```diff
@@ -0,0 +1,80 @@
...
+    /* NB: These must correspond to the routes, inside the Switch */
+    {
+      title: 'Create Image',
```
johnwcallahan, line 32: I'm not super sold on this tab name. Seems weird to have the breadcrumb read "Images / Create" and the first tab to be "Create Image" (seems redundant). Make "From Disk"? @tiffwong thoughts here? Check with Matthew as well.

johnwcallahan (changes requested):
Good work, this routing stuff can be gnarly.

packages/manager/src/features/Images/ImagesCreate/ImageCreate.tsx
DevDW: `<NavTabs />` works wonderfully, thanks for bringing this to my attention

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
```diff
@@ -36,6 +36,7 @@ const styles = (theme: Theme) =>
     root: {},
     helperText: {
       paddingTop: theme.spacing(1) / 2,
+      width: '70%',
```
DevDW, line 39: Did 70% instead of 75% because it looked a bit better to me

packages/manager/src/features/Images/ImagesCreate/ImageCreate.tsx
```diff
@@ -26,53 +23,32 @@ export const ImageCreate: React.FC<CombinedProps> = (props) => {
     setDescription(value);
   };
 
-  const tabs = [
-    /* NB: These must correspond to the routes, inside the Switch */
+  const tabs: NavTab[] = [
```
DevDW, line 26: I landed on using "From Linode" and "From Uploaded Source" after balancing the feedback from prep & from Slack, happy to continue brainstorming and tweaking though, especially if anyone has strong objections.

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -449,7 +451,7 @@ export const ImagesLanding: React.FC<CombinedProps> = (props) => {
   }
 
   const onCreateButtonClick = () => {
-    if (machineImagesEnabled) {
+    if (account.data?.capabilities.includes('Machine Images')) {
```
DevDW, line 454: This change means you'll need to test the Create page using dev until the feature is rolled out on prod more broadly

packages/manager/src/features/Images/ImagesCreate/ImageCreate.tsx
johnwcallahan: I don't think "From Uploaded Source" works. I think "Upload Image" worked well, because that's exactly what you'll be doing on the tab.

My proposal:

`Disk Image | Upload Image`

and the URLs should be:

`images/create/disk`
`images/create/upload`

packages/manager/src/features/Images/ImagesCreate/ImageCreate.tsx
```diff
@@ -0,0 +1,56 @@
...
+type CombinedProps = RouteComponentProps<{}>;
+
+const CreateImageTab = React.lazy(() => import('./CreateImageTab'));
+const ImageUpload = React.lazy(() => import('../ImageUpload'));
+
+export const ImageCreate: React.FC<CombinedProps> = (props) => {
+  const [label, setLabel] = React.useState<string>('');
+  const [description, setDescription] = React.useState<string>('');
+
+  const handleSetLabel = (e: React.ChangeEvent<HTMLInputElement>) => {
```
johnwcallahan, line 16: Why are `handleSetLabel` and `handleSetDescription` operating in this component (the parent) rather than in CreateImageTab? Is that just holdover from the old ImagesDrawer component? Once everything is merged and things are cleaned up, we should make sure this state is in the most appropriate component.

Jskobos: The idea was we would share state between the tabs, since those fields are the same regardless of flow. In #7586 I saw this maybe didn't work out as well as I thought, we'll have to see what happens.

DevDW: I oscillated between moving them down a level and moving them back up between 5 and 6 commits ago, but I'll move them back down in a follow-up to this PR

johnwcallahan (approved):

PR #7578, opened 2021-04-12
M3-5043 Change: ImagesLanding split into 2 tables

Jskobos:
## Description

- Using client-side sorting (for now, until the API work is complete), split
Images into 2 tables, one for manual (normal images) and one for automatic (created
by Linode when a Linode is deleted).
- Manual images no longer have an Expiry column, since they never expire
- Add a Status column, present only when the machineImages flag is active
- When image.status is pending, only one action is possible ("Cancel", which
just deletes the image)
- When image.status is not "available", disable "deploy" actions

## Note to Reviewers

I used our normal flagging logic to keep with convention and because it's in the ticket's AC, but I think I might feel a bit safer if we showed the Status column if the API returns a "status" field on the Image. It amounts to the same thing but we'd be tying the display to exactly what we're concerned with.

To test, try the following combos:

1. Feature flag is on (you can use devtools for this) && customer tag is present on the account; use dev/alpha API
2. Flag is off, and customer tag is not present (normal customers)
3. For the handling of `pending_upload` and `creating` statuses, use the provided mocks.

### DIFF
--- a/packages/manager/src/components/ActionMenu_CMR/ActionMenu_CMR.tsx
+++ b/packages/manager/src/components/ActionMenu_CMR/ActionMenu_CMR.tsx
@@ -62,7 +62,6 @@ const useStyles = makeStyles((theme: Theme) => ({
     '&[data-reach-menu-items]': {
       padding: 0,
       minWidth: 200,
-      width: 'fit-content',
       background: '#3683dc',
       border: 'none',
       fontSize: 14,
--- a/packages/manager/src/components/ActionMenu_CMR/ActionMenu_CMR.tsx
+++ b/packages/manager/src/components/ActionMenu_CMR/ActionMenu_CMR.tsx
@@ -61,7 +61,8 @@ const useStyles = makeStyles((theme: Theme) => ({
   itemsOuter: {
     '&[data-reach-menu-items]': {
       padding: 0,
-      width: 200,
+      minWidth: 200,
+      width: 'fit-content',
       background: '#3683dc',
       border: 'none',
       fontSize: 14,
--- a/packages/manager/src/features/Images/ImageRow.tsx
+++ b/packages/manager/src/features/Images/ImageRow.tsx
@@ -11,23 +11,6 @@ import { formatDate } from 'src/utilities/formatDate';
 import ActionMenu, { Handlers } from './ImagesActionMenu';
 
 const useStyles = makeStyles((theme: Theme) => ({
-  label: {
-    width: '30%',
-    [theme.breakpoints.down('sm')]: {
-      width: '45%',
-    },
-    [theme.breakpoints.down('xs')]: {
-      width: '65%',
-    },
-  },
-  size: {
-    [theme.breakpoints.down('sm')]: {
-      width: '15%',
-    },
-    [theme.breakpoints.down('xs')]: {
-      width: '20%',
-    },
-  },
   loadingStatus: {
     marginBottom: theme.spacing() / 2,
   },
@@ -54,12 +37,13 @@ const ImageRow: React.FC<CombinedProps> = (props) => {
     id,
     label,
     size,
+    status,
     ...rest
   } = props;
 
   return isImageUpdating(event) ? (
     <TableRow key={id} data-qa-image-cell={id}>
-      <TableCell className={classes.label} data-qa-image-label>
+      <TableCell data-qa-image-label>
         <ProgressDisplay
           className={classes.loadingStatus}
           text="Creating"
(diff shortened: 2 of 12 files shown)


packages/manager/src/features/Images/ImageRow.tsx
```diff
@@ -11,23 +11,6 @@ import { formatDate } from 'src/utilities/formatDate';
 import ActionMenu, { Handlers } from './ImagesActionMenu';
 
 const useStyles = makeStyles((theme: Theme) => ({
-  label: {
```
Jskobos, line 14: These were overriding the ones in ImagesLanding; could have deleted either one.

packages/manager/src/features/Images/ImagesActionMenu.tsx
```diff
@@ -33,38 +35,54 @@ export const ImagesActionMenu: React.FC<CombinedProps> = (props) => {
...
+      : [
+          {
+            title: 'Edit',
+            onClick: () => {
+              onEdit(label, description ?? ' ', id);
+            },
+          },
+          {
+            title: 'Deploy New Linode',
+            // @todo remove first half of this conditional when the feature is GA
```
Jskobos, line 65: Maybe we need a tooltip here too

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -80,43 +152,32 @@ const defaultDialogState = {
 export const ImagesLanding: React.FC<CombinedProps> = (props) => {
   useReduxLoad(['images']);
 
+  const classes = useStyles();
+
+  // Feature flag logic for machine images
+  const { account } = useAccount();
```
Jskobos, line 158: See note, I'd prefer to flag off of the presence or absence of `status` in the Images response.

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -258,7 +322,7 @@ export const ImagesLanding: React.FC<CombinedProps> = (props) => {
     return (
       <ActionsPanel>
         <Button buttonType="cancel" onClick={closeDialog} data-qa-cancel>
-          Cancel
+          {dialogAction === 'cancel' ? 'Keep Image' : 'Cancel'}
```
Jskobos, line 325: Trying to make this coherent, tbh I think always having "Delete" is fine, since that's what's happening anyway.

packages/manager/src/features/Images/ImageRow.tsx
```diff
@@ -73,20 +57,25 @@ const ImageRow: React.FC<CombinedProps> = (props) => {
     </TableRow>
   ) : (
     <TableRow key={id} data-qa-image-cell={id}>
-      <TableCell className={classes.label} data-qa-image-label>
-        {label}
-      </TableCell>
+      <TableCell data-qa-image-label>{label}</TableCell>
       <Hidden xsDown>
+        {status ? <TableCell>{status.replace('_', ' ')}</TableCell> : null}
```
DevDW, line 62: `capitalize(status.replace('_', ' '))` or something similar

DevDW (approved):
LGTM; some improvements to the tablet/mobile styling can be made either here or in a follow-up (e.g., the contents of the "Status", "Created", and "Expiry" fields wrap pretty early even when there's still a lot of table space to the right)

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -383,18 +458,41 @@ export const ImagesLanding: React.FC<CombinedProps> = (props) => {
...
         open={dialog.open}
-        title={`Delete Image ${dialog.image}`}
+        title={`${capitalize(dialogAction)} Image ${dialog.image}`}
         onClose={closeDialog}
         actions={getActions}
       >
         {dialog.error && <Notice error text={dialog.error} />}
-        <Typography>Are you sure you want to delete this Image?</Typography>
+        <Typography>
+          Are you sure you want to {dialogAction} this Image?
```
johnwcallahan, line 494: Maybe the "Cancel" button should be "Cancel Upload" and the dialog text here should be something like "Are you sure you want to cancel this Image's upload?".

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -383,18 +458,41 @@ export const ImagesLanding: React.FC<CombinedProps> = (props) => {
...
+      </Paper>
+      <Paper className={classes.imageTable}>
+        <div className={classes.imageTableHeader}>
+          <Typography variant="h3">Automatic Images</Typography>
+          <Typography className={classes.imageTableSubheader}>
+            These images are created automatically when a Linode is deleted.
+            They will be deleted after the indicated expiration date.
+          </Typography>
+        </div>
+        <EntityTable
```
johnwcallahan, line 479: Maybe more descriptive empty text (instead of the default "Not data to display")? Same thing for Manual Images.

johnwcallahan (commented):
Keying off of `status` instead of the flag works for me. Currently I'm getting this when the LD flag is enabled but not the customer tag:

Screen Shot 2021-04-13 at 3 18 14 PM

Recently we've done some work on the StackScripts List to put all actions in the hamburger menu to avoid inconsistent actions. I wonder if we should do the same thing here:

Screen Shot 2021-04-13 at 3 36 31 PM

packages/manager/src/features/Images/ImagesActionMenu.tsx
tiffwong: A tooltip would be helpful to explain why the action is disabled

tiffwong (approved):
Currently also experiencing the same thing John is when the LD flag is off 
Screen Shot 2021-04-14 at 11 52 04 AM

Approved pending the feedback John and Dajahi left (responsiveness can be address in another ticket)

packages/manager/src/components/ActionMenu_CMR/ActionMenu_CMR.tsx
```diff
@@ -61,7 +61,8 @@ const useStyles = makeStyles((theme: Theme) => ({
   itemsOuter: {
     '&[data-reach-menu-items]': {
       padding: 0,
-      width: 200,
+      minWidth: 200,
+      width: 'fit-content',
```
johnwcallahan, line 65: What does this do? 

I didn't actually notice any problems on FF, but according to the [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/fit-content) this needs a vendor prefix:

```css  
width: -moz-fit-content
```

johnwcallahan (approved):
I wrote down a few minor UX enhancements to discuss with Matthew at demo prep, but I think this is good to go. 

(here's my list):

```
- Include tooltip next to "Pending Upload" status?
- Images with “Pending Upload” status: size is 0 MB… does that make sense? Does it need more context?
- Could use helper text below "Manual Images" (like we do for "Automatic Images") 
- Maybe more space between Manual Images and Automatic Images table
```

PR #7579, opened 2021-04-12
M3- 5060 Add event handling for image_upload events

Jskobos:


packages/manager/src/eventMessageGenerator_CMR.tsx
```diff
@@ -41,6 +41,8 @@ export const eventMessageGenerator = (
       return `reboot with ${e.secondary_entity?.label}`;
     case 'linode_shutdown':
       return 'shutdown';
+    case 'linode_delete':
```
Jskobos, line 44: I think this can have progress, so it ought to be here

DevDW (approved):

tiffwong (approved):

PR #7582, opened 2021-04-13
M3-1801: Fix tags error for restricted users w/ Linode creation permissions

DevDW:
## Description
Currently, restricted users with Linode creation permissions who go to create a Linode see this:

Screen Shot 2021-04-13 at 6 14 54 PM

This is because calling the `getTags` method returns an error. Based on the API, these users are able to create Linodes with (new) tags but can't interact with existing tags on other entities.

I added an optional `fromLinodeCreateFlow` prop that gets passed from LinodeCreate --> LabelAndTagsPanel --> TagsInput, where the `defaultError` is not set if the error reason is "Unauthorized" and `fromLinodeCreateFlow` is true.

## Type of Change

### DIFF
--- a/packages/manager/src/components/TagsInput/TagsInput.tsx
+++ b/packages/manager/src/components/TagsInput/TagsInput.tsx
@@ -23,7 +23,6 @@ export interface Props {
   hideLabel?: boolean;
   name?: string;
   tagError?: string;
-  fromLinodeCreateFlow?: boolean;
   value: Item[];
   onChange: (selected: Item[]) => void;
   disabled?: boolean;
@@ -70,8 +69,8 @@ class TagsInput extends React.Component<Props, State> {
           { reason: 'There was an error retrieving your tags.' },
         ];
 
-        // If a restricted user with Linode creation permission goes to create a Linode, do not display the defaultError to them.
-        if (e[0].reason === 'Unauthorized' && this.props.fromLinodeCreateFlow) {
+        // If a restricted user with entity creation permission goes to create/edit an entity, do not display the defaultError to them.
+        if (e[0].reason.match(/unauthorized/i)) {
           return;
         }
 
--- a/packages/manager/src/components/TagsInput/TagsInput.tsx
+++ b/packages/manager/src/components/TagsInput/TagsInput.tsx
@@ -23,6 +23,7 @@ export interface Props {
   hideLabel?: boolean;
   name?: string;
   tagError?: string;
+  fromLinodeCreateFlow?: boolean;
   value: Item[];
   onChange: (selected: Item[]) => void;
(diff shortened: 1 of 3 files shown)


tiffwong (approved):

packages/manager/src/components/TagsInput/TagsInput.tsx
```diff
@@ -64,10 +65,16 @@ class TagsInput extends React.Component<Props, State> {
...
         this.setState({ accountTags });
       })
-      .catch((_) => {
+      .catch((e) => {
         const defaultError = [
           { reason: 'There was an error retrieving your tags.' },
         ];
+
+        // If a restricted user with Linode creation permission goes to create a Linode, do not display the defaultError to them.
+        if (e[0].reason === 'Unauthorized' && this.props.fromLinodeCreateFlow) {
```
Jskobos, line 74: This check is good and should handle all the cases. Maybe we could err on the side of OCD and do `.match(/unauthorized/)` to be more future-proof.

Jskobos (changes requested):
The ticket was made because of the Linode create flow, but a restricted user is going to run into the same issue when creating/editing e.g. a Domain:

Screen Shot 2021-04-14 at 11 30 50 AM

I don't think we need the fromLinodeCreateFlow prop (I don't like `fromWhatever` prop-based stuff in general, so I'm biased). The check you have in the tags input is good and should be fine for all cases.

Have you reached out to Matthew about this one? Hiding the error works and is probably an acceptable workaround, but the UX isn't great--the dropdown says "Type to choose or create a tag" but a restricted user can never choose, the list is empty. If they know the name of an existing tag they can type it themselves, but it's not a great experience.

DevDW:
> Have you reached out to Matthew about this one? Hiding the error works and is probably an acceptable workaround, but the UX isn't great--the dropdown says "Type to choose or create a tag" but a restricted user can never choose, the list is empty. If they know the name of an existing tag they can type it themselves, but it's not a great experience.

I hadn't but I just reached out to get his perspective

Jskobos (approved):
Thanks for the updates, if Matthew has a different approach we can go with that, otherwise this looks good.

I had another (optional) thought though, which is that while intercepting the error works well, we know in advance the request will fail for any restricted user (`const { isRestrictedUser } = useAccountManagement();`), so we could save that round trip by not making the request in the first place.

DevDW:
> I had another (optional) thought though, which is that while intercepting the error works well, we know in advance the request will fail for any restricted user (`const { isRestrictedUser } = useAccountManagement();`), so we could save that round trip by not making the request in the first place.

This sounds good. For now I'll merge this PR but I made M3-5081 for this

PR #7586, opened 2021-04-14
M3-5046 Add: Upload image tab

Jskobos:
## Description

Branched off of #7575, will rebase and open for review once that's in.

## Note to Reviewers

To test: point the app at the dev/alpha API and go to /images/create/upload-image.

I'd sort of like to refactor all of this, but the fact that the other tab is a bunch of old class component code made me hold off for now. Seems like there should be some way to reuse the label/description fields between tabs, and ideally to tie the whole thing together with Formik. Alternatively we could eliminate the shared state altogether and have each tab be a self-contained Formik form. as is we've sort of got the worst of both worlds, but these forms are small enough that it doesn't bother me that much.

## Todo

- [  ] Tests 
- [  ] Rebase


packages/manager/src/store/image/image.requests.ts
```diff
@@ -30,6 +32,10 @@ export const createImage = createRequestThunk(
   ({ diskID, label, description }) => _create(diskID, label, description)
 );
 
+export const uploadImage = createRequestThunk(uploadImageActions, (payload) =>
```
Jskobos, line 35: I'm not happy to be adding more things using the old pattern, but trying to query-ify this on top of everything else is asking for trouble.

tiffwong (approved):

packages/manager/src/features/Images/ImageUpload.tsx
```diff
@@ -1,7 +1,138 @@
...
+  const classes = useStyles();
+  const { push } = useHistory();
+  const [region, setRegion] = React.useState<string>('');
+  const dispatch: Dispatch = useDispatch();
+  const regions = useRegionsQuery().data ?? [];
+  const [submitting, setSubmitting] = React.useState(false);
+  const [errors, setErrors] = React.useState<APIError[] | undefined>();
+
+  const handleSubmit = () => {
+    setSubmitting(true);
```
johnwcallahan, line 56: Should clear errors here too.

packages/manager/src/features/Images/ImageUpload.tsx
```diff
@@ -1,7 +1,138 @@
...
+    description,
+    changeLabel,
+    changeDescription,
+  } = props;
+  const classes = useStyles();
+  const { push } = useHistory();
+  const [region, setRegion] = React.useState<string>('');
+  const dispatch: Dispatch = useDispatch();
+  const regions = useRegionsQuery().data ?? [];
+  const [submitting, setSubmitting] = React.useState(false);
```
johnwcallahan, line 52: Maybe make a comment that we should RQ-ify this later.

packages/manager/src/features/Images/ImageUpload.tsx
```diff
@@ -1,7 +1,138 @@
...
+      .catch((e) => {
+        setSubmitting(false);
+        setErrors(e);
+      });
+  };
+
+  const errorMap = getErrorMap(['label', 'description', 'region'], errors);
+
+  return (
+    <Paper className={classes.container}>
```
johnwcallahan, line 82: The Papers between tabs are different widths. I kind of like the more narrow "Upload Image" paper, but either way we ought to be consistent. @matthewbartholomew @tiffwong  which should we go with here?

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -182,6 +190,22 @@ export const ImagesLanding: React.FC<CombinedProps> = (props) => {
       ? 'Are you sure you want to cancel this Image upload?'
       : 'Are you sure you want to delete this Image?';
 
+  const [successDialogOpen, setSuccessDialogOpen] = React.useState(false);
+  const [uploadURL, setUploadURL] = React.useState<string | undefined>();
+
+  const handleCloseSuccessDialog = () => {
+    setSuccessDialogOpen(false);
+    window.setTimeout(() => setUploadURL(undefined), 500);
```
johnwcallahan, line 198: Is this that trick we use to keep the URL form changing while the modal is closing? Might be good to make this is own function for reusability and ease of reading.

johnwcallahan (changes requested):
Only blocking thing is clearing the error state on submit. I'll try to take another pass once @matthewbartholomew's updates are in.

packages/manager/src/features/Images/ImageUploadSuccessDialog.tsx
```diff
@@ -108,10 +113,10 @@ export const ImageUploadSuccessDialog: React.FC<Props> = (props) => {
             <div>
               <Button
                 buttonType="secondary"
+                className={classes.button}
                 onClick={() => {
                   handleCopy(0, curlExample);
                 }}
-                outline
               >
                 Copy Curl Request
```
johnwcallahan, line 121: I saw @matthewbartholomew mention that the capitalization is still under review; personally I think either works as long as they're consistent.

packages/manager/src/features/Images/ImageUpload.tsx
```diff
@@ -81,13 +83,7 @@ export const ImageUpload: React.FC<Props> = (props) => {
...
     <Paper className={classes.container}>
       {errorMap.none ? <Notice error text={errorMap.none} /> : null}
-      <form
-        style={{ width: '100%' }}
-        onSubmit={(e: React.FormEvent) => {
-          e.preventDefault();
-          handleSubmit();
-        }}
-      >
+      <div style={{ width: '100%' }}>
```
johnwcallahan, line 86: Minor/not exactly related: Kinda seems like a lot of space between the text and this button (same on the "Capture Image" tab)... maybe 16px instead of 32px?

johnwcallahan (approved):

PR #7587, opened 2021-04-15
Feedback: Remove redundant title from Linode Details Analytics tab

tiffwong:
Will address the different units for the IPv4 and IPv6 graphs in a separate PR

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeSummary/LinodeSummary.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeSummary/LinodeSummary.tsx
@@ -65,8 +65,12 @@ const styles = (theme: Theme) =>
       maxWidth: 150,
     },
     graphControls: {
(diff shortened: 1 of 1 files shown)


HanaXu:
On second thought, can we move the dropdown back to the right?

tiffwong:
@HanaXu I can but is there a specific reason why? 🤔

HanaXu:
I feel like left alignment looks a bit more awkward than right alignment, and it seems that right alignment is the norm (in general, and also in various other places in the cloud manager)

DevDW (approved):
I'm fine with either placement (left or right) but I'd still vote the same way as I did in prep for the placement on the right

tiffwong:
@HanaXu @DevDW I'll bring it up in the channel again 😂

HanaXu:
LGTM 🎉

Jskobos (approved):

PR #7588, opened 2021-04-15
Feedback: Machine Images table row updates

tiffwong:
- Always expose actions for pending Images
- Update "Cancel" to "Cancel Upload"
- Remove progress bar for pending images and have the corresponding size be "N/A"

### DIFF
--- a/packages/manager/src/features/Images/ImageRow.tsx
+++ b/packages/manager/src/features/Images/ImageRow.tsx
@@ -47,7 +47,7 @@ const ImageRow: React.FC<CombinedProps> = (props) => {
         <TableCell data-qa-image-date>{formatDate(created)}</TableCell>
       </Hidden>
       <TableCell data-qa-image-size>
-        {isImageUpdating(event) ? 'N/A' : `${size} MB`}
+        {isImageUpdating(event) ? 'Pending' : `${size} MB`}
       </TableCell>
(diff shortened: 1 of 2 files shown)


packages/manager/src/features/Images/ImageRow.tsx
```diff
@@ -65,7 +46,9 @@ const ImageRow: React.FC<CombinedProps> = (props) => {
         ) : null}
         <TableCell data-qa-image-date>{formatDate(created)}</TableCell>
       </Hidden>
-      <TableCell data-qa-image-size>{size} MB</TableCell>
+      <TableCell data-qa-image-size>
+        {isImageUpdating(event) ? 'N/A' : `${size} MB`}
```
Jskobos, line 50: I'd prefer "Pending" here, and it should show if the status is 'pending_upload' or 'creating'.

Jskobos (changes requested):
Removing the progress bar also removed the progress display, which shows the percent complete (we do this on Linodes too). That should be brought back imo.

Also, a "pending_upload" image doesn't trigger the logic in isImageUpdating, which is event-based. So those images still show a size of 0, which should be Pending instead.

Jskobos (approved):

DevDW (approved):

PR #7594, opened 2021-04-19
M3-4225: Fix Typing of renderUnitPrice and Test

codyfinn:
## Description

This PR corrects an incorrect argument type declaration for a function called `renderUnitPrice`. Previously the function took a type of `null | number` but in practice was being given a `null | string`. The function was extracted from the `InvoiceTable` react component and moved into the `billingUtils.ts` file. Tests were added to verify that strings that parse into valid numbers return a formatted string for the unit price. Tests were also added that if a string that cannot be parsed into a Number or null are passed tot he function then the function returns null. 

## How to test

To run the unit tests associated with this PR run `yarn test ./packages/manager/src/features/Billing/billingUtils.test.ts` from the project root directory.

### DIFF
--- a/packages/manager/src/features/Billing/InvoiceDetail/InvoiceTable.tsx
+++ b/packages/manager/src/features/Billing/InvoiceDetail/InvoiceTable.tsx
@@ -11,6 +11,7 @@ import TableCell from 'src/components/TableCell';
 import TableRowEmptyState from 'src/components/TableRowEmptyState';
 import TableRowError from 'src/components/TableRowError';
 import TableRowLoading from 'src/components/TableRowLoading';
+import { renderUnitPrice } from 'src/features/Billing/billingUtils.ts';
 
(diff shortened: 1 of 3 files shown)


johnwcallahan (changes requested):
Nice work, just a few notes.

packages/manager/src/App.tsx
```diff
@@ -84,7 +84,7 @@ export class App extends React.Component<CombinedProps, State> {
      */
     // eslint-disable-next-line
     document.addEventListener('keydown', (event: KeyboardEvent) => {
-      if (event.ctrlKey && event.shiftKey && event.key === 'D') {
+      if (event.ctrlKey && event.shiftKey && event.key === 'L') {
```
Jskobos, line 87: @codyfinn is this the right branch?

johnwcallahan (approved):

PR #7596, opened 2021-04-20
M3-5067: remove mapStateToProps in favor of hooks in StackScriptsDetail

HanaXu:
## Description

Removes old `mapStateToProps` pattern in favor of using hooks

## How to test

Tech story, doesn't seem like the values were actually being used anywhere yet.

### DIFF
--- a/packages/manager/src/features/StackScripts/StackScriptsDetail.tsx
+++ b/packages/manager/src/features/StackScripts/StackScriptsDetail.tsx
@@ -91,7 +91,6 @@ export const StackScriptsDetail: React.FC<CombinedProps> = (props) => {
   );
 
   const username = profile.data?.username;
-  const stackScriptGrants = profile?.data?.grants?.stackscript ?? [];
   const userCannotAddLinodes = _isRestrictedUser && !_hasGrant('add_linodes');
 
   React.useEffect(() => {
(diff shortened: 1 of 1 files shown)


packages/manager/src/features/StackScripts/StackScriptsDetail.tsx
```diff
@@ -99,6 +91,8 @@ export const StackScriptsDetail: React.FC<CombinedProps> = (props) => {
   );
 
   const username = profile.data?.username;
+  const stackScriptGrants = profile?.data?.grants?.stackscript ?? [];
```
DevDW, line 94: Being that this is unused and the grants get checked in `components/StackScript/StackScript.tsx`, I think this can be removed.

packages/manager/src/features/StackScripts/StackScriptsDetail.tsx
```diff
@@ -99,6 +91,8 @@ export const StackScriptsDetail: React.FC<CombinedProps> = (props) => {
...
+  const userCannotAddLinodes = _isRestrictedUser && !_hasGrant('add_linodes');
```
DevDW, line 95: Let's go ahead and use this to determine if the "Deploy New Linode" button should be enabled or disabled. You can add `disabled={userCannotAddLinodes}` to that `<Button />`

packages/manager/src/features/StackScripts/StackScriptsDetail.tsx
ndukatz: I came here to say this

DevDW (approved):

ndukatz (approved):
Looks good! e2e passed

PR #7601, opened 2021-04-21
Fix: image_upload events

Jskobos:
## Description

Adjust our handling of the image_upload event.

In general, if an event has a percent_complete from 0-99 we consider it "in progress." This means that we show a progress bar for it in the notifications drawer, may display related entities differently throughout the app, and (most importantly) we don't throttle our polling interval if there are any in progress events still running. In practice this means we'll request events every 1.5 seconds, instead of ramping down to every 16 seconds.

The image_upload event, generated when you POST to image/uploads, is different, in that it can stay in percent_complete: 0 for up to 24 hours (until the user actually uploads an image to the provided URL). There may be some other long-running events we can consider that would fall into a similarly long pending state, but for now we need to treat this one event differently so that we're not spamming the API with /account/events calls.

Question: we might want to hide the duration for these events, since it's still weird ("completed in 16 hours"). I did this for the drawer, but we could consider doing the same thing for the events landing page.

## How to test

1. Create an Image using the /images/create/upload flow
2. On develop you should see progress for this event in the Notification drawer, here it should be rendered without the bar:

Develop:
Screen Shot 2021-04-21 at 1 28 49 PM

This branch:
Screen Shot 2021-04-21 at 1 37 39 PM

3. Open dev tools to the Network tab and filter on "events". On develop, you should see a call to this endpoint every ~2 seconds, here it should throttle down to once every 16 seconds (unless something else is happening on your account).

4. Run the tests:

`yarn test event.handlers`

### DIFF
--- a/packages/manager/src/store/events/event.helpers.ts
+++ b/packages/manager/src/store/events/event.helpers.ts
@@ -18,7 +18,7 @@ import { EntityEvent, ExtendedEvent } from './event.types';
  * for that period, would be pointless. We treat these events differently
  * when determining whether they are in progress.
  */
-const LONG_PENDING_EVENTS = ['image_upload'];
+export const LONG_PENDING_EVENTS = ['image_upload'];
 
 /** We use the epoch on our initial request to get all of the users events. */
 export const epoch = new Date(`1970-01-01T00:00:00.000`).getTime();
--- a/packages/manager/src/store/events/event.helpers.ts
+++ b/packages/manager/src/store/events/event.helpers.ts
@@ -10,6 +10,16 @@ import updateRight from 'src/utilities/updateRight';
 
 import { EntityEvent, ExtendedEvent } from './event.types';
 
+/**
+ * Some events can be in a pending state, with a percent_complete
+ * of 0 and a status of "scheduled", for hours or days. We don't want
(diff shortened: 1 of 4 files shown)


tiffwong (approved):
Everything looks good to me!

It said there were no tests found when I tried to run `yarn test event.handlers` but all the tests passed when I ran `yarn test event`

packages/manager/src/store/events/event.helpers.ts
```diff
@@ -148,10 +158,16 @@ export const addToEvents = (prevArr: Event[], arr: Event[]) =>
       : [el, ...updatedArray];
   }, prevArr);
 
-export const isInProgressEvent = ({
-  percent_complete,
-}: Pick<Event, 'percent_complete'>) =>
-  percent_complete !== null && percent_complete < 100;
+export const isInProgressEvent = (event: Event) => {
+  const { percent_complete } = event;
+  if (percent_complete === null) {
```
johnwcallahan, line 163: I'm envisioning what this would look like as a chained ternary and it doesn't look pretty. Thanks for the verbosity here. 😅

johnwcallahan (approved):
Looks good; fixes the problem and I didn't find any regressions.

DevDW (approved):
LGTM 🚢

PR #7602, opened 2021-04-21
M3-5086: Filter out creating and pending images from ImageSelect

HanaXu:
## Description
Filters out images with the status of `creating` or `pending_upload` in the ImageSelect component.

## How to test
1. Create a manual image for a Linode in `Images`
2. While the image status is creating or pending upload, go to `Linodes` and click on `Create Linode`
3. In the `Images` tab within Create Linode, click on the `Choose an image` dropdown
4. The image we just created should not display
5. Go back to image status and wait for the status to be `Ready`
6. Check the dropdown again and the new image should now display

### DIFF
--- a/packages/manager/src/components/ImageSelect/ImageSelect.tsx
+++ b/packages/manager/src/components/ImageSelect/ImageSelect.tsx
@@ -96,7 +96,7 @@ export const imagesToGroupedItems = (images: Image[]) => {
               const isDeprecated = thisImage.deprecated;
               const fullLabel =
                 thisImage.label + (isDeprecated ? ' (Deprecated)' : '');
-              const _option = {
+              return {
                 created: thisImage.created,
                 label: fullLabel,
                 value: thisImage.id,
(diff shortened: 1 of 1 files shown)


packages/manager/src/components/ImageSelect/ImageSelect.tsx
```diff
@@ -135,12 +135,15 @@ export const ImageSelect: React.FC<Props> = (props) => {
         .reason
     : undefined;
 
+  const imageIsReady = (image: Image) =>
+    !['creating', 'pending_upload'].includes(image.status);
```
Jskobos, line 139: I think `image.status === 'available'` is just as good here, there's only 3 statuses possible in an /images response

Jskobos (approved):
Looks good 👍  There are a few linter warnings in this file, you might as well clean those up while you're in here.

tiffwong (approved):
Great job!

PR #7603, opened 2021-04-21
M3-5079: More precise helper text on Images Landing

DevDW:
## Description
Revised, more precise helper text developed in collaboration w/ John and Jackley. The linked guide will be updated next week with more Automatic Images information.

### DIFF
--- a/packages/manager/src/features/Images/ImagesLanding.tsx
+++ b/packages/manager/src/features/Images/ImagesLanding.tsx
@@ -497,7 +497,7 @@ export const ImagesLanding: React.FC<CombinedProps> = (props) => {
           <Typography className={classes.imageTableSubheader}>
             {machineImagesEnabled
               ? `These are images you manually uploaded or captured from an existing Linode disk.`
-              : `These are images you captured from an existing disk.`}
+              : `These are images you captured from an existing Linode disk.`}
           </Typography>
         </div>
         <EntityTable
--- a/packages/manager/src/features/Images/ImagesLanding.tsx
+++ b/packages/manager/src/features/Images/ImagesLanding.tsx
@@ -513,11 +513,6 @@ export const ImagesLanding: React.FC<CombinedProps> = (props) => {
           <Typography className={classes.imageTableSubheader}>
             These are images we automatically capture when Linode disks are
             deleted. They will be deleted after the indicated expiration date.
-            See{' '}
-            <Link to="https://www.linode.com/docs/guides/linode-images/">
-              here
-            </Link>{' '}
-            for more details.
           </Typography>
         </div>
         <EntityTable
--- a/packages/manager/src/features/Images/ImagesLanding.tsx
+++ b/packages/manager/src/features/Images/ImagesLanding.tsx
@@ -496,8 +496,8 @@ export const ImagesLanding: React.FC<CombinedProps> = (props) => {
           <Typography variant="h3">Manual Images</Typography>
           <Typography className={classes.imageTableSubheader}>
             {machineImagesEnabled
-              ? `These are images you manually uploaded or captured from an existing disk or Linode. They never expire.`
-              : `These images are created from a Linode's disk.`}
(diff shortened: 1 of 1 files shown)


packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -496,7 +496,7 @@ export const ImagesLanding: React.FC<CombinedProps> = (props) => {
           <Typography variant="h3">Manual Images</Typography>
           <Typography className={classes.imageTableSubheader}>
             {machineImagesEnabled
-              ? `These images are created from a Linode's disk or by uploading your own image.`
+              ? `These are images you manually uploaded or captured from an existing disk or Linode. They never expire.`
```
Jskobos, line 499: Maybe "from an existing Linode disk" is a bit more accurate.

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -511,8 +511,12 @@ export const ImagesLanding: React.FC<CombinedProps> = (props) => {
         <div className={classes.imageTableHeader}>
           <Typography variant="h3">Automatic Images</Typography>
           <Typography className={classes.imageTableSubheader}>
-            These images are created automatically when a Linode is deleted.
-            They will be deleted after the indicated expiration date.
+            These are images we automatically capture when most Linodes or disks
+            are deleted. They automatically expire. See{' '}
+            <Link to="https://www.linode.com/docs/guides/linode-images/">
```
Jskobos, line 516: Just double-checking, they're planning to update this doc with some auto-image info? There's not much helpful in there atm.

packages/manager/src/features/Images/ImagesLanding.tsx
johnwcallahan, line 499: The "They never expire" bit might be confusing because if you create an image for upload but don't upload it in 24 hours, it does "expire".
 
If we remove that part here, maybe keep "They will be deleted after the indicated expiration date." in the automatic images section.

Also, pretty sure we should remove "or Linode" from this section since to create an image this way you need to select a specific disk.

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -496,7 +496,7 @@ export const ImagesLanding: React.FC<CombinedProps> = (props) => {
           <Typography variant="h3">Manual Images</Typography>
           <Typography className={classes.imageTableSubheader}>
             {machineImagesEnabled
-              ? `These images are created from a Linode's disk or by uploading your own image.`
+              ? `These are images you manually uploaded or captured from an existing disk or Linode. They never expire.`
               : `These images are created from a Linode's disk.`}
```
johnwcallahan, line 500: This text needs updating too: "{These are images you captured from from an existing disk.

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -511,8 +511,12 @@ export const ImagesLanding: React.FC<CombinedProps> = (props) => {
         <div className={classes.imageTableHeader}>
           <Typography variant="h3">Automatic Images</Typography>
           <Typography className={classes.imageTableSubheader}>
-            These images are created automatically when a Linode is deleted.
-            They will be deleted after the indicated expiration date.
+            These are images we automatically capture when most Linodes or disks
```
Jskobos, line 514: Same comment, I understand where you're going with this--for most normal users, clicking "delete Linode" is what makes this happen--but "Linode or disk" could be confusing in a different way.

packages/manager/src/features/Images/ImagesLanding.tsx
DevDW: Yes, my understanding is that update will be released next Wednesday

johnwcallahan (approved):
Pushed up one small change to add "Linode" in front of "disk".

Jskobos (approved):

PR #7605, opened 2021-04-22
M3-5062: VLAN Refinements in Linode Create flow

DevDW:
## Description
When VLANs are not available for the selected region, show the "Attach a VLAN" section in the Linode Create flow instead of hiding it but:
- Display notice with regions listed where VLANs are available
- Disable text fields


tiffwong (approved):

packages/manager/src/features/linodes/LinodesCreate/AttachVLAN.tsx
```diff
@@ -77,6 +80,17 @@ const AttachVLAN: React.FC<CombinedProps> = (props) => {
       </Box>
       <Grid container>
         <Grid item xs={12}>
+          {!regionSupportsVLANs ? (
+            <Notice warning>
+              <Typography>
+                <strong>
+                  VLAN instances are not available in the selected region.
```
Jskobos, line 87: We shouldn't hard code this. If a datacenter gets VLAN support, we'd have to release to include it in the list. You already have a list of regions that support VLANs, from where `regionSupportsVLANS` is calculated (might want to move that down into this component). There's also a method in src/utilities to join a list of region IDs into a comma separated list. I've been meaning to add support for semicolons, which you've used here and are a good idea. An optional `separator = ','` argument would be a nice addition to the helper function.

Jskobos (changes requested):
Works well, we need to fix the hard-coding but otherwise good to go. Also, though, we discussed in sprint planning what to do about the case where you haven't selected a region yet. Right now if you scroll straight to the VLAN input you get a message saying "VLANs aren't supported in the selected region", which could be a bit confusing because there is no selected region. Please add a simple `const message = selectedRegionID ? <what you already have> : 'You must select a region before adding a VLAN'` or similar.

packages/manager/src/features/linodes/LinodesCreate/AttachVLAN.tsx
DevDW: Good point about the hard-coding, thank you. Not sure if this was the precise file you had in mind, but the placement seems fair:

https://github.com/linode/manager/blob/e78c98b5bd49c52752eb5d35926b15212f5fe76a/packages/manager/src/utilities/arrayToCommaSeparatedList/arrayToList.ts#L14

packages/manager/src/utilities/arrayToCommaSeparatedList/arrayToList.test.ts
```diff
@@ -19,6 +19,15 @@ describe('Array to comma-separated list', () => {
     );
   });
 
+  it('should return a list with items containing commas as semicolon-separated', () => {
```
Jskobos, line 22: I'm being pedantic here, sorry: "It should use the provided delimiter" etc; it isn't using a semicolon because the items have commas, it's because the test case is passing `;`

packages/manager/src/utilities/doesRegionSupportVLANs.ts
```diff
@@ -12,3 +13,7 @@ export const doesRegionSupportVLANs = (
   }
   return regionMetaData.capabilities.includes('Vlans');
 };
+
+export const regionsWithVLANs = (regionsData: Region[] | ExtendedRegion[]) => {
```
Jskobos, line 17: Appreciate the precision but unless you're using a field specific to ExtendedRegion I don't think you need to include it in the types.

Jskobos: (because it's a superset of Region, and TS generally doesn't care in its checking that the types match precisely, only that everything is compatible)

packages/manager/src/utilities/arrayToCommaSeparatedList/arrayToList.ts
```diff
@@ -10,6 +10,11 @@ export default (input: string[]): string => {
   } else {
     const head = input.slice(0, -1);
     const tail = input[input.length - 1];
+
+    if (separator) {
```
Jskobos, line 14: This would be terser as an optional argument with a default value, `..., separator: string = ',')`

packages/manager/src/features/linodes/LinodesCreate/AttachVLAN.tsx
```diff
@@ -67,9 +65,32 @@ const AttachVLAN: React.FC<CombinedProps> = (props) => {
...
+
+  const regionsThatSupportVLANs = regionsWithVLANs(regions).map(
+    (region: ExtendedRegion) => region.display
+  );
+  const unavailableInRegionMessage = `VLAN instances are not available in the selected region. Currently they are available in ${arrayToList(
+    regionsThatSupportVLANs,
+    ';'
+  )}.`;
+
+  const warningMessage = selectedRegion
```
Jskobos, line 90: What do you think about including the second sentence (the list of regions) no matter what? So this conditional only handles the first sentence. That way, for impulsive users who get to the bottom of the view, they know that they have to select a region and also what regions they can choose from. Otherwise you might have to scroll back and forth twice.

Jskobos (commented):
Thanks for the updates, this is good now pending a few cleanup items.

packages/manager/src/features/linodes/LinodesCreate/AttachVLAN.tsx
DevDW: Agreed; I pushed that change up

packages/manager/src/features/linodes/LinodesCreate/AttachVLAN.tsx
```diff
@@ -77,6 +103,13 @@ const AttachVLAN: React.FC<CombinedProps> = (props) => {
       </Box>
       <Grid container>
         <Grid item xs={12}>
+          {!regionSupportsVLANs ? (
+            <Notice warning>
```
johnwcallahan, line 107: I know this is the pattern we use for GPUs, but this banner is much more visible. It really jumps out and makes you think something is wrong if you've scrolled down without selecting a region, or if you've selected an unsupported region.

Screen Shot 2021-04-27 at 9 03 38 AM


@tiffwong @matthewbartholomew what are your thoughts on making this text-only (instead of a `<Notice />` component? And/or updating the text to be less demanding? ("You **must** select a region...")

At the very least it should be consistent with GPUs, which currently does not have bolded text (not sure if it should be bolded or not).
Screen Shot 2021-04-27 at 9 04 22 AM

PR #7606, opened 2021-04-22
M3-4801: light dark mode shortcut change

codyfinn:
## Description

This PR changes the keyboard shortcuts on Windows and Linux. This is in to preserve `Ctrl + Shift + D` and `Ctrl + Shift + K` on MacOS. The new code detects the OS the browser reports in `navigator.appVersion` and if the OS is not MacOS then it will use the `Alt + Shift` prefix instead.

## How to test

To test this open the manager in Firefox on a Windows machine and test that `Alt + Shift + D` toggles the dark and light themes and that `Alt + Shift + K` brings up the Go To Open menu.


tiffwong (commented):
Can confirm that the MacOS shortcuts work but I couldn't get the commands to work with VMware 😬

Jskobos (commented):
I can't get the Windows shortcuts working using Browserstack (which does work for the Mac shortcuts). It could be a problem with my keyboard or something else random, if you've got an actual Windows machine and it's working then probably everything's fine. I'm interested if anyone else has had any success here.

We should probably add some documentation here, including a link to the references you used. Future devs seeing mystical stuff like 'Î' and '\uf8ff' may get confused, or at least spend some time hunting down keycode charts.

codyfinn:
> I can't get the Windows shortcuts working using Browserstack (which does work for the Mac shortcuts). It could be a problem with my keyboard or something else random, if you've got an actual Windows machine and it's working then probably everything's fine. I'm interested if anyone else has had any success here.
> 
> We should probably add some documentation here, including a link to the references you used. Future devs seeing mystical stuff like 'Î' and '\uf8ff' may get confused, or at least spend some time hunting down keycode charts.

Those special characters are no longer used. Those only show up in MacOS  when using those key chords for some reason. On Linux and Windows the letter comes up as the one on the keyboard. Everything should work now.

DevDW (commented):
Same case for me–the MacOS shortcuts are working, but I wasn't able to get the Windows shortcuts working on Browserstack. I do have a laptop that runs Windows, so I'll try to get Cloud set up locally on there to test and report back.

codyfinn:
Just came across that `navigator.appVersion` is deprecated. Should I change this to use `navigator.userAgent` instead?

Jskobos (commented):
OK now I definitely don't know what's happening (with BrowserStack), I'm never even getting into the `switch` statement and when I log the event it's nothing remotely close to what I'm pressing. Everything looks good (I still think a comment like what you've got for the PR would be helpful to have inline here), and I'm happy to approve if people on actual Windows machines can verify the shortcuts work.

codyfinn:
> OK now I definitely don't know what's happening (with BrowserStack), I'm never even getting into the `switch` statement and when I log the event it's nothing remotely close to what I'm pressing. Everything looks good (I still think a comment like what you've got for the PR would be helpful to have inline here), and I'm happy to approve if people on actual Windows machines can verify the shortcuts work.

This PR no longer uses special characters. Only MacOS seems to have the behavior of pressing `Alt + Shift` and another letter resulting in a character not on English keyboards. On Windows and Linux results in the letter pressed.

Jskobos:
> This PR no longer uses special characters. Only MacOS seems to have the behavior of pressing `Alt + Shift` and another letter resulting in a character not on English keyboards. On Windows and Linux results in the letter pressed.

I should have been more clear and included a screenshot the first time. I'm using BrowserStack to test this, and their docs say that they support all browser keyboard shortcuts. My keyboard has an Alt key, yet when I press Alt + Shift + D nothing happens. When I console.log the keyboard event, it says that event.shiftKey is false and event.key is a keypad number:

Screen Shot 2021-04-27 at 11 29 50 AM

This is confusing because the docs say that this specific situation should work (see [here](https://www.browserstack.com/question/488)); it may be this bug: https://stackoverflow.com/questions/65113392/browserstack-ctrlenter-keyboard-event-not-registering-in-chrome-or-ie.

Again, if it works on an actual Windows machine then I'm fine to approve.

johnwcallahan (approved):

bnussman (commented):
Looks good and works as expected on all the configurations I listed below. 

One thing to note is that on Firefox on Windows, _alt_ **alone** toggles the menu bar, but when doing Alt + Shift + D, the menu bar does not toggle so all seems good. 

As a side question, would it be appropriate to include a hint on `/profile/settings` to tell the user the keyboard shortcut for dark mode or do we want this to remain an easter egg?

Tested on the following configurations:
- macOS - Chrome, Safari
- Windows - Chrome, Edge, Firefox
- Pop_OS! - Chrome

tiffwong (approved):
The Windows shortcuts works on VMware for Chrome and Firefox, great job! 🎉 

(The MacOS shortcuts also work)

PR #7608, opened 2021-04-22
M3-5081: Convert TagsInput.tsx to function component & don't request tags for restricted users

DevDW:
## Description
Following up on https://github.com/linode/manager/pull/7582 – convert file to a function component & change the approach from intercepting the error when a user is restricted to not requesting tags for restricted users per https://github.com/linode/manager/pull/7582#pullrequestreview-636619551.

## How to test
Please ensure the tags functionality is unchanged.

You can have an unrestricted user and a restricted user that has access to create Linodes. In the Create flow for the restricted user, no error should display on the Tags field, and there should be no network request to `/tags`.


Jskobos (approved):
Refactoring looks good. When I was clicking around to test, I found that there are 2 other components (AddTag.tsx and TagsPanel.tsx) that make the request for /tags independently. It's not great to have so much inconsistency, and probably the best long-term solution is to have a tagsQuery where `useTagsQuery({ enabled: !isRestrictedUser })` or something. But for now, we should probably add the logic you've got here to those other two components. They swallow their errors so it's not user-affecting, but seems better overall.

tiffwong (commented):
Everything looks good except when I try to add a tag via the tag drawer, it says "Not found" even when I try to create a new one and the drawer header also says "undefined" 🤔 

Screen Shot 2021-04-26 at 9 14 51 AM

tiffwong (approved):
Hmm this seems to be a problem on develop as well so approving since I don't think it's related to the changes you made

PR #7609, opened 2021-04-23
change status from disk imagize to capturing image

HanaXu:
## Description
Changes status transition text from `Disk Imagize` to `Capturing Image`
## How to test
1. Go to `Images` and create/capture an image.
2. The status in `Images` should be `Creating`
3. Go to `Linodes`, the linode status you captured the image from should be `Capturing image`

### DIFF
--- a/packages/manager/src/features/linodes/transitions.test.ts
+++ b/packages/manager/src/features/linodes/transitions.test.ts
@@ -1,25 +1,17 @@
 import { transitionText } from './transitions';
+import { eventFactory } from 'src/factories/events';
 
 describe('transitionText helper', () => {
   it('should remove "linode" from the event and capitalize it', () => {
-    expect(
-      transitionText('loading', 123, {
-        id: 123,
-        action: 'linode_snapshot',
-        secondary_entity: null,
-        created: '2012-12-12',
-        entity: null,
-        percent_complete: null,
-        rate: null,
-        read: true,
-        seen: true,
-        status: 'started',
-        time_remaining: null,
-        username: 'linode',
-        duration: 0,
-        message: null,
-      })
-    ).toEqual('Snapshot');
+    const mockEvent = eventFactory.build({
+      action: 'linode_snapshot',
+      percent_complete: null,
+      read: true,
+      seen: true,
+    });
+    expect(transitionText('loading', mockEvent.id, mockEvent)).toEqual(
+      'Snapshot'
+    );
   });
 
   it('should use status if an event is missing and capitalize it', () => {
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/linodes/transitions.ts
```diff
@@ -33,6 +33,12 @@ export const transitionAction = [
   'linode_migrate_datacenter',
 ];
 
+const transitionActionMap = {
```
johnwcallahan, line 36: You could provide a little type safety by typing this as `Partial<Record<EventAction, string>>`.

packages/manager/src/features/linodes/transitions.ts
```diff
@@ -71,17 +69,16 @@ export const transitionText = (
     }
   }
 
-  if (recentEvent?.action === 'linode_migrate_datacenter') {
-    return 'Migrating';
+  if (recentEvent && transitionActionMap[recentEvent.action]) {
```
johnwcallahan, line 72: I think there's a good opportunity here for a simple unit test or two in `transitions.test.ts` to verify this new behavior.

packages/manager/src/features/linodes/transitions.ts
```diff
@@ -71,17 +69,16 @@ export const transitionText = (
...
+    return transitionActionMap[recentEvent.action];
   }
 
-  let event;
-  if (recentEvent && transitionAction.includes(recentEvent.action)) {
-    event = recentEvent.action.replace('linode_', '').replace('_', ' ');
-  } else {
-    event = status.replace('_', ' ');
+  if (recentEvent && transitionAction.includes(recentEvent?.action)) {
```
johnwcallahan, line 76: I feel this function is still ripe with opportunity for refactor, and I'm not totally convinced this change makes things much clearer (still have the duplication of `.replace()`, and now `capitalizeAllWords` is duplicated).

Now that we have the mapping of actions to labels, maybe we could expand it to inlude for the actions in `transitionAction` and remove the `transitionAction` logic altogether (i.e. this whole `if` block).

packages/manager/src/features/linodes/transitions.ts
```diff
@@ -22,21 +22,15 @@ export const transitionStatus = [
...
-];
-
-const transitionActionMap = {
+const transitionActionMap: Partial<Record<EventAction, string>> = {
+  backups_restore: 'Backups Restore',
+  linode_snapshot: 'Snapshot',
   linode_mutate: 'Upgrading',
+  linode_clone: 'Cloning',
   linode_migrate_datacenter: 'Migrating',
+  disk_resize: 'Resizing',
```
Jskobos, line 31: This should be "Disk Resizing" or similar, there's a separate plan resize event that can happen.

packages/manager/src/features/linodes/transitions.ts
```diff
@@ -22,21 +22,15 @@ export const transitionStatus = [
...
   disk_imagize: 'Capturing Image',
+  disk_duplicate: 'Duplicating',
```
Jskobos, line 33: Same here, Disk Duplicating so it doesn't get confused with Cloning

packages/manager/src/features/linodes/transitions.test.ts
```diff
@@ -46,4 +46,30 @@ describe('transitionText helper', () => {
       })
     ).toEqual('Optimus');
   });
+
+  it('should use present tense if an event is a transition event and capitalize it', () => {
+    expect(
```
Jskobos, line 51: You could use a factory here, 
```
import { eventFactory } from 'src/factories/events'; 
const mockEvent = linodeFactory.build({ action: 'disk_imagize', ...<whatever specific fields you need>})
```

packages/manager/src/features/linodes/transitions.test.ts
```diff
@@ -46,4 +46,30 @@ describe('transitionText helper', () => {
       })
     ).toEqual('Optimus');
   });
+
+  it('should use present tense if an event is a transition event and capitalize it', () => {
```
Jskobos, line 50: Present continuous tense if we're being pedantic 🙃

Jskobos (changes requested):
I think the refactor looks good. It looks like `buildLinodeCloneTransitionText` isn't used anywhere in the app anymore, it can probably be deleted.

The only required changes are the event renames and the test cleanup.

Jskobos (approved):

packages/manager/src/features/linodes/transitions.test.ts
```diff
@@ -1,49 +1,47 @@
 import { transitionText } from './transitions';
+import { eventFactory } from 'src/factories/events';
 
 describe('transitionText helper', () => {
   it('should remove "linode" from the event and capitalize it', () => {
```
johnwcallahan, line 5: I don't think this first test is providing value anymore since the "linode" string removal logic is no longer present.

packages/manager/src/features/linodes/transitions.test.ts
```diff
@@ -1,49 +1,47 @@
...
-        status: 'started',
-        time_remaining: null,
-        username: 'linode',
-        duration: 0,
-        message: null,
-      })
-    ).toEqual('Snapshot');
+    const mockEvent = eventFactory.build({
+      action: 'linode_snapshot',
+      percent_complete: null,
```
johnwcallahan, line 8: I believe the only field needed to test the functionality here is `action`, so you could probably remove the other fields from all these test events.

johnwcallahan (approved):
Nice job on the refactor 👍 

Left a few minor/optional notes on the tests, but I think this is good to go.

PR #7611, opened 2021-04-26
Change: Refactor schemas out of the JS client

Jskobos:
## Description

Now that the validation package is building, we can move the schemas out of the JS client. To do this, I:

- Made sure the version of each schema in `@linode/validation/src/` matches the version in the JS client
- Delete the latter
- Update imports

The next step will be to add an options object to the JS client request methods. This will be a breaking change.
## How to test

Tests and checks should pass as before, behavior of the app should be unchanged. No schemas or references to them should remain in the folders I went through (account, databases, domains, transfers, firewalls, and images).


DevDW (approved):
LGTM 🚢

packages/validation/src/images.schema.ts
```diff
@@ -1,17 +1,26 @@
...
-    .max(50, 'Length must be 50 characters or less.')
-    .matches(
-      /^[a-zA-Z0-9,.?\-_\s']+$/,
-      'Image labels cannot contain special characters.'
-    ),
-  description: string().notRequired().min(1).max(65000),
+});
+
+export const uploadImageSchema = baseImageSchema.shape({
+  label: labelSchema.required('Label is required.'),
```
johnwcallahan, line 22: Nice solution for making `label` required (or not).

johnwcallahan (approved):

PR #7612, opened 2021-04-26
M3-5057: TextField error message positioning and width

tiffwong:
## Description

This PR refactors `TextField` into a FC and adds a conditional prop that adjusts the position and width of an editable `TextField` when it's part of a breadcrumb. This will prevent the error message from wrapping to a newline when there's space in a form.

ie.
Screen Shot 2021-04-09 at 2 33 21 PM

The original ask for this ticket was to look into replacing `position: absolute` and the fixed width with `display: flex`. I ended up cleaning up the responsiveness and tightening the whitespace instead because there was no clean way to move the error message after the save/cancel buttons since the error message is rendered as part of `TextField`.

## How to test

Please check the error message for `TextField` components throughout the app at different viewports. You can find an editable breadcrumb on most of the product detail pages.

### DIFF
--- a/packages/manager/src/components/EnhancedNumberInput/EnhancedNumberInput.test.tsx
+++ b/packages/manager/src/components/EnhancedNumberInput/EnhancedNumberInput.test.tsx
@@ -1,4 +1,5 @@
-import { fireEvent, render } from '@testing-library/react';
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import * as React from 'react';
 
 import { wrapWithTheme } from 'src/utilities/testHelpers';
@@ -27,7 +28,7 @@ describe('EnhancedNumberInput', () => {
 
     const addButton = getByTestId('increment-button');
 
-    fireEvent.click(addButton);
+    userEvent.click(addButton);
     expect(setValue).toHaveBeenCalledWith(2);
   });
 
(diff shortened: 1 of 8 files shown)


packages/manager/src/components/EnhancedNumberInput/EnhancedNumberInput.test.tsx
```diff
@@ -48,7 +49,7 @@ describe('EnhancedNumberInput', () => {
     );
 
     const input = getByTestId('textfield-input');
-    fireEvent.change(input, { target: { value: '100' } });
+    userEvent.type(input, '100');
```
johnwcallahan, line 52: I believe this is failing because `setValue` never gets called with `100`... it gets called each type a character is typed. Before it was working since the value was set altogether. You could probably check the end result instead of what the function is called with to get it to pass.

johnwcallahan (commented):
Left a comment about the broken EnhancedNumberInput test.

The general approach of passing in the `inBreadcrumb` prop to `TextField` is kind of an anti-pattern... ideally the TextField component should know/care nothing about a "BreadCrumb" component. The way we're trying to control for every width and length also makes me nervous, as there are some cases where it doesn't work out:

(shouldn't wrap here):
Screen Shot 2021-04-29 at 1 47 19 PM

Screen Shot 2021-04-29 at 1 47 27 PM

tiffwong:
Closing this since I'm approaching it from another angle

PR #7613, opened 2021-04-27
For release: Remove beta notices and link for VLANs since they're leaving beta

Jskobos:
## Description

Remove the beta notice and legal text from the VLAN section of the Linode create flow and config modal. This will be released on 4/28 when the feature goes into GA.

## How to test

Everything should work as before; the beta chip and info should not be displayed.


johnwcallahan (approved):
Pushed up a small change.

tiffwong (approved):

PR #7614, opened 2021-04-27
M3-4841 consistent storage units render

codyfinn:
## Description

This PR updates formatting for storage units across the cloud manager application. There are a few places where storage sizes are listed that needed to updated in place but for the Linode plans there is an update to the reducer that should be reflected in all places where the name of the Linode plan are visible.

## How to test

Check to see if there are any places in the app where there is not a space before the storage unit

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
@@ -205,7 +205,7 @@ export class SelectPlanPanel extends React.Component<CombinedProps> {
             <TableCell className={classes.radioCell}>
               {!isSamePlan && (
                 <FormControlLabel
-                  label={type.heading.replace('GB', ' GB')}
+                  label={type.heading}
                   aria-label={type.heading}
                   className={'label-visually-hidden'}
                   control={
--- a/packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
@@ -205,7 +205,7 @@ export class SelectPlanPanel extends React.Component<CombinedProps> {
             <TableCell className={classes.radioCell}>
               {!isSamePlan && (
                 <FormControlLabel
-                  label={type.heading}
+                  label={type.heading.replace('GB', ' GB')}
                   aria-label={type.heading}
                   className={'label-visually-hidden'}
                   control={
--- a/packages/manager/src/features/linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
+++ b/packages/manager/src/features/linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
@@ -211,7 +211,9 @@ export const LinodeRow: React.FC<CombinedProps> = (props) => {
 
       <Hidden xsDown>
         <TableCell className={classes.planCell} data-qa-ips>
-          <div className={classes.planCell}>{type?.label ?? 'Unknown'}</div>
+          <div className={classes.planCell}>
+            {(type?.label ?? 'Unknown').replace('GB', ' GB')}
+          </div>
         </TableCell>
         <TableCell className={classes.ipCell} data-qa-ips>
           <div className={classes.ipCellWrapper}>
--- a/packages/manager/src/features/linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
+++ b/packages/manager/src/features/linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
@@ -211,7 +211,7 @@ export const LinodeRow: React.FC<CombinedProps> = (props) => {
 
       <Hidden xsDown>
         <TableCell className={classes.planCell} data-qa-ips>
-          <div className={classes.planCell}>{label}</div>
+          <div className={classes.planCell}>{type?.label}</div>
         </TableCell>
         <TableCell className={classes.ipCell} data-qa-ips>
           <div className={classes.ipCellWrapper}>
--- a/packages/manager/src/features/linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
+++ b/packages/manager/src/features/linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
@@ -211,9 +211,7 @@ export const LinodeRow: React.FC<CombinedProps> = (props) => {
 
       <Hidden xsDown>
         <TableCell className={classes.planCell} data-qa-ips>
-          <div className={classes.planCell}>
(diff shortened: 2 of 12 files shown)


johnwcallahan (approved):
Great work tracking these down. I audited the app and wasn't able to find any inconsistencies. I left a minor, optional comment but otherwise good to go.

packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
```diff
@@ -205,7 +205,7 @@ export class SelectPlanPanel extends React.Component<CombinedProps> {
             <TableCell className={classes.radioCell}>
               {!isSamePlan && (
                 <FormControlLabel
-                  label={type.heading}
+                  label={type.heading.replace('GB', ' GB')}
```
Jskobos, line 208: I think this is double dipping, though the extra space isn't visible to the user:

Screen Shot 2021-04-29 at 9 22 59 AM

...since this data comes from Redux where you've already done the replacement. As a safety, when/if you refactor this replace logic to a utility, maybe you could use a regex that checks for GB or MB next to a non-whitespace character.

Jskobos (approved):
Looks good & thorough, I can't find any stragglers either. There's a non-user facing cleanup thing, and I made one suggestion for when/if you're addressing John's feedback. Neither is blocking.

packages/manager/src/features/linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
```diff
@@ -211,7 +211,9 @@ export const LinodeRow: React.FC<CombinedProps> = (props) => {
 
       <Hidden xsDown>
         <TableCell className={classes.planCell} data-qa-ips>
-          <div className={classes.planCell}>{type?.label ?? 'Unknown'}</div>
+          <div className={classes.planCell}>
+            {(type?.label ?? 'Unknown').replace('GB', ' GB')}
```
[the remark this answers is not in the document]
codyfinn: the `.replace('GB', ' GB')` has been moved to the reducer. This code is extraneous and will be removed in the next commit. I will make a utility function that will more generically format a string containing a storage unit like Jared suggested that way if we have future strings that are similar that function can be reused.

tiffwong (approved):
Great job!

packages/manager/src/utilities/formatStorageUnits.ts
```diff
@@ -0,0 +1,14 @@
+const storageRegex = /([0-9])([kMGTPEZY]?i?[B])/;
+
+function replaceFunc(match: string, p1: string, p2: string) {
+  if (match.length === 0) return 'unknown';
+  return `${p1} ${p2}`;
+}
+
+function formatStorageUnits(unformattedString: string | null) {
+  const cleanedUnformattedString = unformattedString ?? 'unknown';
+  if (!cleanedUnformattedString.match(storageRegex)) return 'unknown';
```
johnwcallahan, line 10: We have the `curly` eslint rule in place, so you'll need to add curly braces and break these to multiple lines.

Screen Shot 2021-05-03 at 12 29 50 PM

packages/manager/src/utilities/formatStorageUnits.ts
```diff
@@ -0,0 +1,14 @@
+const storageRegex = /([0-9])([kMGTPEZY]?i?[B])/;
+
+function replaceFunc(match: string, p1: string, p2: string) {
+  if (match.length === 0) return 'unknown';
```
johnwcallahan, line 4: I think a better/safer alternative here is the string itself. I believe defaulting to "Unknown" its actually not necessary anymore, since we are requesting even shadow plan types in AuthenticationWrapper (with `ensureAllTypes()`).

packages/manager/src/utilities/formatStorageUnits.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+describe('formatStorageUnits', () => {
+  it('returns "unknown" when formatting a string that does not contain a storage unit', () => {
+    expect(formatStorageUnits('Linode High Memory')).toBe('unknown');
+    expect(formatStorageUnits('Linode GB')).toBe('unknown');
+    expect(formatStorageUnits('')).toBe('unknown');
+    expect(formatStorageUnits(null)).toBe('unknown');
+  });
+
+  it('returns a string that contains the storage units and measurement seperated by a single space character', () => {
```
johnwcallahan, line 11: Spelling of "separated'.

johnwcallahan (changes requested):

packages/manager/src/utilities/formatStorageUnits.ts
codyfinn: I guess it should actually throw an error or return null or something else. This is supposed to be a generic function that doesn't only work on plan names but instead that works where ever we want to ensure consistent spacing with storage units.

johnwcallahan: But if it's run on a string without storage units? E.g. if we introduce a new plan name without storage units in the title, we'd wouldn't want this to throw an error or return null... we'd just want the original string.

packages/manager/src/utilities/formatStorageUnits.ts
```diff
@@ -0,0 +1,14 @@
+const storageRegex = /([0-9])([kMGTPEZY]?i?[B])/;
```
johnwcallahan, line 1: I plugged this into regex101.com I think we should add a `+` after the group of digits: `([0-9]+)` and make the whole thing case insensitive.

packages/manager/src/utilities/formatStorageUnits.ts
codyfinn: Good point!

packages/manager/src/utilities/formatStorageUnits.ts
johnwcallahan: 👍  good points.

packages/manager/src/utilities/formatStorageUnits.ts
```diff
@@ -1,13 +1,14 @@
...
 function replaceFunc(match: string, p1: string, p2: string) {
-  if (match.length === 0) return 'unknown';
   return `${p1} ${p2}`;
 }
 
-function formatStorageUnits(unformattedString: string | null) {
-  const cleanedUnformattedString = unformattedString ?? 'unknown';
-  if (!cleanedUnformattedString.match(storageRegex)) return 'unknown';
+function formatStorageUnits(unformattedString: string) {
+  const cleanedUnformattedString = `${unformattedString}`;
```
johnwcallahan, line 8: Super nitpicky... doesn't seem "cleanedUnformattedString" is necessary anymore.

johnwcallahan (approved):

PR #7616, opened 2021-04-27
WIP: Add BareMetalRescue confirmation dialog inside RescueDialog

Jskobos:
## Description

Bare metal Linodes have a different rescue procedure, since their disks aren't managed through the API.

- Added a separate method to the JS client (`rescueMetalLinode`); this could have used the existing method, since it's the same endpoint, but I felt this was less confusing.
- Added logic to the RescueDialog so that if a Linode is using a bare metal plan a confirmation dialog opens, instead of the full Rescue dialog.

Still in progress:

Seems like there ought to be a way to refactor this to make it more clean/use more shared stuff. I'm working on a RescueContainer, not quite sure how it's going yet.

## How to test


packages/api-v4/src/linodes/actions.ts
```diff
@@ -140,6 +140,26 @@ export const rescueLinode = (
...
+ * Rescue Mode is based on the Finnix recovery distribution, a self-contained
+ * and bootable Linux distribution.
+ *
+ * Bare Metal Linodes have a different rescue flow, since their disks are not
+ * managed through the Linode API. This method hits the same endpoint as rescueLinode,
+ * but does not accept an array of devices.
+ *
+ * @param linodeId { number } The id of the Linode to boot into rescue mode.
+ */
+export const rescueMetalLinode = (linodeId: number): Promise<{}> =>
```
Jskobos, line 157: This could be merged with the above method, feel free to talk me out of my approach here.

packages/manager/src/__data__/types.ts
```diff
@@ -22,6 +22,7 @@ export const types: LinodeType[] = [
     successor: null,
     memory: 2048,
     vcpus: 1,
+    gpus: 0,
```
Jskobos, line 25: I opened a tech debt task to remove usage of this file.

packages/manager/src/components/AddNewLink/AddNewLink_CMR.tsx
```diff
@@ -8,7 +8,6 @@ export interface Props extends Omit<TooltipProps, 'children' | 'title'> {
   disabled?: boolean;
   disabledReason?: string;
   label: string;
-  left?: boolean;
   onClick: (e?: React.MouseEvent<HTMLElement>) => void;
```
Jskobos, line 11: The `left` prop, only used here afaict, seems to have been removed from Button, so it was being passed down to the DOM and triggering a console warning.

packages/manager/src/features/linodes/LinodesCreate/LinodeCreateContainer.tsx
```diff
@@ -595,7 +595,7 @@ class LinodeCreateContainer extends React.PureComponent<CombinedProps, State> {
       type && {
         title: type.label,
         details: `${typeLabelDetails(type.memory, type.disk, type.vcpus)}`,
-        monthly: type.price.monthly,
+        monthly: type.price.monthly ?? 0,
```
Jskobos, line 598: All of this stuff is safe access since price.monthly can be `null` now.

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/BareMetalRescue.tsx
```diff
@@ -0,0 +1,74 @@
...
+      setLoading(false);
+    }
+  }, [isOpen]);
+
+  const handleSubmit = () => {
+    setError(undefined);
+    setLoading(true);
+    rescueMetalLinode(linodeID)
+      .then(() => {
+        setLoading(false);
```
Jskobos, line 35: There's still some repetition here, I thought further refactoring wasn't worth the complexity; please feel free to push back.

packages/api-v4/src/linodes/actions.ts
DevDW: I think having this separate may be clearer

DevDW (approved):
LGTM 👌

tiffwong (approved):
LGTM

PR #7617, opened 2021-04-27
M3-4957: "Last Backup" cell for Bare Metal instances

DevDW:
## Description
Backups are unavailable for Bare Metal instances. This PR adds "N/A" with a tooltip explaining this to the "Last Backup" cell for Bare Metal instances in the Linodes landing table.

### DIFF
--- a/packages/manager/src/components/BackupStatus/BackupStatus.tsx
+++ b/packages/manager/src/components/BackupStatus/BackupStatus.tsx
@@ -14,7 +14,8 @@ import HelpIcon from 'src/components/HelpIcon';
 
 type ClassNames =
   | 'icon'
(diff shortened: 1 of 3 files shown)


Jskobos (approved):
Code looks good. I think it would be nice to add some left padding or whatnot to the icon so that it lines up with the cloud icon:
Screen Shot 2021-04-29 at 8 51 27 AM

Another question for later: do we really want this column to be sortable? In the case above ^^^ toggling the sort doesn't do anything. We could work something out but I don't know how useful that is to begin with.

tiffwong (approved):
Everything looks good but I feel like the help icon is a little big, can we make it a little smaller?

PR #7620, opened 2021-04-28
M3-5058: Show promo service type in billing summary

HanaXu:
## Description
Displays `Applies to: <service_type>` in the promo billing summary if the service type is not "all".

## How to test

1. You will need to update your `.env` file to use dev services
2. Login to a dev test account that has a specific service type promo applied to it. The 4th dev test account currently has the service type `linodes` applied.
3. Click on `Accounts`
4. You should see a promotions box, under the `expires` line, there should be another line that says `Applies to: <service_type`

### DIFF
--- a/packages/api-v4/src/account/types.ts
+++ b/packages/api-v4/src/account/types.ts
@@ -56,7 +56,7 @@ export interface ActivePromotion {
   this_month_credit_remaining: string;
   credit_monthly_cap: string;
   image_url: string;
-  service_type: string;
+  service_type: PromotionServiceType;
 }
 
 export type PromotionServiceType =
--- a/packages/api-v4/src/account/types.ts
+++ b/packages/api-v4/src/account/types.ts
@@ -56,7 +56,23 @@ export interface ActivePromotion {
   this_month_credit_remaining: string;
   credit_monthly_cap: string;
   image_url: string;
+  service_type: string;
 }
+
+export type PromotionServiceType =
+  | 'all'
+  | 'backup'
+  | 'blockstorage'
+  | 'db_mysql'
+  | 'ip_v4'
+  | 'linode'
+  | 'linode_disk'
+  | 'linode_memory'
+  | 'longview'
+  | 'managed'
+  | 'nodebalancer'
+  | 'objectstorage'
+  | 'transfer_tx';
 
 interface CreditCard {
   expiry: string | null;
--- a/packages/manager/src/factories/account.ts
+++ b/packages/manager/src/factories/account.ts
@@ -13,6 +13,7 @@ export const promoFactory = Factory.Sync.makeFactory<ActivePromotion>({
   credit_monthly_cap: '20.00',
   credit_remaining: '20.00',
   this_month_credit_remaining: '20.00',
+  service_type: 'all',
 });
 
 export const accountFactory = Factory.Sync.makeFactory<Account>({
@@ -51,6 +52,7 @@ export const accountFactory = Factory.Sync.makeFactory<Account>({
       credit_monthly_cap: '20.00',
       credit_remaining: '20.00',
       this_month_credit_remaining: '20.00',
+      service_type: 'all',
     },
     {
       image_url: '',
@@ -61,6 +63,7 @@ export const accountFactory = Factory.Sync.makeFactory<Account>({
       credit_monthly_cap: '20.00',
       credit_remaining: '20.00',
       this_month_credit_remaining: '20.00',
+      service_type: 'all',
     },
   ],
   euuid: '278EC57D-7424-4B3A-B35C3CE395787567',
--- a/packages/manager/src/factories/account.ts
+++ b/packages/manager/src/factories/account.ts
@@ -54,17 +54,6 @@ export const accountFactory = Factory.Sync.makeFactory<Account>({
       this_month_credit_remaining: '20.00',
       service_type: 'all',
     },
-    {
-      image_url: '',
-      summary: 'BAD_PROMO',
-      description:
-        'This promotion is invalid since it has an expiration of null',
-      expire_dt: null,
-      credit_monthly_cap: '20.00',
-      credit_remaining: '20.00',
-      this_month_credit_remaining: '20.00',
-      service_type: 'all',
-    },
   ],
   euuid: '278EC57D-7424-4B3A-B35C3CE395787567',
 });
--- a/packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.test.tsx
+++ b/packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.test.tsx
@@ -67,6 +67,22 @@ describe('BillingSummary', () => {
     getByTextWithMarkup('Monthly cap: $20.00');
   });
 
+  it('displays promo service type unless the service type is all', () => {
+    const promotions = [
+      promoFactory.build(),
+      promoFactory.build({ summary: 'MY_PROMO_CODE', service_type: 'linode' }),
+    ];
+    renderWithTheme(
+      <BillingSummary
+        balance={0}
+        balanceUninvoiced={5}
+        promotions={promotions}
+      />
+    );
+    expect(screen.queryByText('Applies to: All')).not.toBeInTheDocument();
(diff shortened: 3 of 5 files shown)


packages/api-v4/src/account/types.ts
```diff
@@ -56,8 +56,24 @@ export interface ActivePromotion {
   this_month_credit_remaining: string;
   credit_monthly_cap: string;
   image_url: string;
+  service_type: string;
```
Jskobos, line 59: This should probably be `PromotionServiceType`

packages/manager/src/factories/account.ts
```diff
@@ -61,6 +63,7 @@ export const accountFactory = Factory.Sync.makeFactory<Account>({
       credit_monthly_cap: '20.00',
       credit_remaining: '20.00',
       this_month_credit_remaining: '20.00',
+      service_type: 'all',
```
Jskobos, line 66: I don't know why we added a "BAD_PROMO" to the factory here but it should be removed. If a test is depending on it it should do the mocking within the test. Then you can delete this repetition and just do `active_promotions: promoFactory.buildList(2)`

Jskobos: Oh it looks like it was me 🙃 ...please delete this and if a test breaks we can patch it.

packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.test.tsx
```diff
@@ -67,6 +67,22 @@ describe('BillingSummary', () => {
     getByTextWithMarkup('Monthly cap: $20.00');
   });
 
+  it('displays promo service type unless the service type is all', () => {
```
Jskobos, line 70: Text is incorrect

Jskobos (changes requested):
Good work on this, the types and mapping etc. are all appropriate. The ticket didn't specify this, but I think we should remove the bit of code that's only displaying a single promotion. Now that these product-specific ones exist, it's possible to have one type = 'all' and one type = 'whatever' on your account; for example the test account we're using has two valid promos on it, but only the Linodes one is displayed. We can open a separate ticket for this if you feel it's out of scope or needs further discussion.

packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
```diff
@@ -57,6 +60,22 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+const serviceTypeMap: Partial<Record<PromotionServiceType, string>> = {
+  all: 'All',
+  backup: 'Backups',
+  blockstorage: 'Volumes',
+  db_mysql: 'DBaaS',
+  ip_v4: 'IPv4',
+  linode: 'Linodes',
+  linode_disk: 'Storage',
+  linode_memory: 'Memory',
+  longview: 'Longviews',
```
johnwcallahan, line 72: Should be "Longview".

packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
```diff
@@ -57,6 +60,22 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+  managed: 'Managed',
+  nodebalancer: 'NodeBalancers',
+  objectstorage: 'Object Storage',
+  transfer_tx: 'Service Transfers',
```
johnwcallahan, line 76: Are you sure this refers to Service Transfers (free service)? My guess is that this refers to transfer overages... I'd check with Nick.

packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
```diff
@@ -57,6 +60,22 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
 }));
 
+const serviceTypeMap: Partial<Record<PromotionServiceType, string>> = {
+  all: 'All',
```
johnwcallahan, line 64: If we're not displaying this, maybe remove it from this mapping.

packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
```diff
@@ -260,12 +280,14 @@ export const PromoDisplay: React.FC<PromoDisplayProps> = React.memo((props) => {
           </Typography>
         ) : null}
       </Box>
-      {/* @todo: Add support for service_type ("Applies to: Linodes"). */}
       {expire_dt ? (
         <Typography>
           Expires: <DateTimeDisplay value={expire_dt} displayTime={false} />
         </Typography>
       ) : null}
+      {service_type !== 'all' ? (
```
johnwcallahan, line 288: We should add one more check here, to make sure the service type is in our mapping before rendering this bit. E.g. with the current logic, if the API adds a new service type without us handling for it, we'll end up displaying `Applies to: `.

johnwcallahan (changes requested):
Great work; requesting a couple of small changes.

johnwcallahan:
> The ticket didn't specify this, but I think we should remove the bit of code that's only displaying a single promotion

@Jskobos we do display multiple promotions correctly ([here](https://github.com/linode/manager/blob/333445e4396732e4c361fe94578e53002c555537/packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx#L194))...  I think the example you're talking about is because the expiry is `null`, and we're filtering those out [here](https://github.com/linode/manager/blob/a989db90290bcee6a8749c830a77c9e1b59334a5/packages/manager/src/store/account/account.requests.ts#L25).

packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
HanaXu: I originally left it out, but then decided to keep it in case we ever needed it in the future and to more closely match the api

packages/api-v4/src/account/types.ts
```diff
@@ -56,8 +56,24 @@ export interface ActivePromotion {
...
+  | 'ip_v4'
+  | 'linode'
+  | 'linode_disk'
+  | 'linode_memory'
+  | 'longview'
+  | 'managed'
+  | 'nodebalancer'
+  | 'objectstorage'
+  | 'transfer_tx';
+
```
codyfinn, line 76: Could this be changed to an Enum instead of a type?
```suggestion
export enum PromotionServiceEnum {
  all = 'All',
  backup = 'Backups',
  blockstorage = 'Volumes',
  dm_mysql = 'DBaas',
  ip_v4 = 'IPv4',
  linode = 'Linodes',
  linode_disk = 'Storage',
  linode_memory = 'Memory',
  longview = 'Longviews',
  managed = 'Managed',
  nodebalancer = 'NodeBalancers',
  objectstorage = 'Object Storage',
  transfer_tx = 'Service Transfers',
};

```
This would then remove the need for the `serviceTypeMap` in `BillingSummary.tsx`

Jskobos: This is a cool idea...is there some way to use enums as union types (quick googling says no but I'm not super familiar with these). I'd like to have type safety/autocomplete for the JS client PromoType. Also, the string values are a bit Cloud-specific, so the Enum should probably live outside of api-v4.

codyfinn: Do you mean union typing two Enums together or do you mean treating the keys/values as union types?

codyfinn: A union type of keys of an enum can be achieved through `keyof typeof PromotionServiceEnum`. the Resulting type can be used to restrict the value assigned to a variable to only the keys of the enum.

HanaXu: This seems a bit out of scope for this story as it's a larger architectural change. We can create a tech story to address this if anyone feels strongly.

Jskobos (approved):

tiffwong (approved):

johnwcallahan (approved):
Great work on this.

PR #7621, opened 2021-04-28
M3-5013: Change OBJ drag and drop instructional text color

DevDW:
## Description
Change the color of the instructional text in the Object Storage drag and drop area to `theme.palette.text.primary` (gray in light mode, white in dark mode).


HanaXu (approved):
LGTM 👍

tiffwong (approved):

PR #7623, opened 2021-04-28
M3-4917: Firewalls Linodes tab refinements

tiffwong:
## Description

- Reduce size of copy to 14px and add spacing around it
- Make button match style used in the Rules tab


Jskobos (approved):

HanaXu (changes requested):
Can we make the margin on the button the same as on the rules tab and fix the padding?
image
image

tiffwong:
@HanaXu great catch for the bottom margin! By the fixing the padding, do you mean the left and right padding on the button?

HanaXu:
Yep!

tiffwong:
@HanaXu you bring up a great point! I brought it up to Matthew and since the interaction that would highlight the padding is relatively fast, we're going to leave it as is since we want to keep the left and right padding the same. I'll create a ticket for looking into removing the color change or an alternate way to handle that state though 😁

HanaXu (approved):

PR #7625, opened 2021-05-03
Change: Validation Schemas (Part 2)

Jskobos:
## Description

Move more schemas from @linode/api-v4 to @linode/validation. I stopped after Linodes because those were complicated (as expected) and the diff is already pretty big here. I deleted some old CMR files and adjusted some tests (Enzyme tests that had been pointed at the old versions of RebuildFromImage etc.).


tiffwong (approved):
Great job cleaning up all these files and all the tests passed for me

DevDW (approved):

PR #7626, opened 2021-05-03
M3-5070: SideMenu scrollbar overlap

tiffwong:
## Description

Refactors the SideMenu component and prevents the scrollbar from overlapping it when the height of the viewport is decreased. 

Since the menu is only in the collapsed view when the width is > 600px, I set the vertical overflow to hidden in order to hide the scrollbar but the scroll is restored on hover. 

## How to test

Set your MacOS preferences to always show the scrollbar and decrease the height of the viewport. The scrollbar should no longer be visible on the collapsed view of the SideMenu until you hover which also expands the menu.

### DIFF
--- a/packages/manager/src/components/SideMenu.tsx
+++ b/packages/manager/src/components/SideMenu.tsx
@@ -1,55 +1,43 @@
 import * as React from 'react';
-import { compose } from 'recompose';
 import Drawer from 'src/components/core/Drawer';
 import Hidden from 'src/components/core/Hidden';
-import {
-  createStyles,
-  Theme,
-  withStyles,
-  WithStyles,
-} from 'src/components/core/styles';
-import withFeatureFlagConsumer, {
-  FeatureFlagConsumerProps,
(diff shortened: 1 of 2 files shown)


packages/manager/src/components/SideMenu.tsx
```diff
@@ -1,108 +1,90 @@
...
+  menuDocked: {
+    height: '100%',
+  },
+  desktopMenu: {
+    transform: 'none',
+  },
+  collapsedDesktopMenu: {
+    width: 52,
+    '&:hover': {
+      overflowY: 'scroll',
```
HanaXu, line 27: Can we change the overflowY from `scroll` to `auto` so that the scroll bar only shows up when it's needed?

image

HanaXu (approved):
LGTM. Just one minor comment

Jskobos (approved):
I don't think this is related, checking to see if it's on the radar (dark theme only): 

Screen Shot 2021-05-04 at 9 19 55 AM

PR #7627, opened 2021-05-03
updating cypress-file-upload for support ticket test

ndukatz:
## Description
This change was needed because of an outdated plugin

**What does this PR do?**
just a small support ticket e2e test update

## How to test
- `yarn up` in one terminal
- `yarn cy:e2e -s ./cypress/integration/helpAndSupport/open-support-ticket.spec.ts` in another

```
REACT_APP_LOGIN_ROOT
REACT_APP_API_ROOT
REACT_APP_APP_ROOT
REACT_APP_LAUNCH_DARKLY_ID
REACT_APP_CLIENT_ID
MANAGER_OAUTH

### CODE REVIEW

File: packages/manager/cypress/tsconfig.json [hunk 1]
```diff
@@ -1,14 +1,17 @@
...
+    "allowJs": true,
+    "checkJs": false,
+    "strict": true,
+    "skipLibCheck": true,
+    "noImplicitAny": false,
+    "target": "es5",
+    "lib": ["es5", "dom"],
+    "baseUrl": "..",
+    "paths": {
+      "@src/*": ["src/*"]
```
Jskobos, line 12: Is this line necessary? I'm fuzzy on this now (and was never super clear to begin with) but I think this is for importing stuff via paths like `import component from '@src/components';` anyway I'm not sure but the tests pass without it

Jskobos (approved):

packages/manager/cypress/tsconfig.json
ndukatz: This will be more useful for the notifications test I think. but I just slipped it in this pr... though it wasn't necessary for this specific test.

johnwcallahan (approved):

PR #7630, opened 2021-05-04
fix Add Disk button styling in RescueDialog

HanaXu:
## Description
Fixes `Add Disk` button styling and alignment in `RescueDialog`

## How to test
In `Linodes`, click on a kebab for a linode and then click `Rescue`

### DIFF
--- a/packages/manager/src/components/Button/Button.tsx
+++ b/packages/manager/src/components/Button/Button.tsx
@@ -51,6 +51,7 @@ const useStyles = makeStyles((theme: Theme) => ({
   superCompact: {
     paddingLeft: 0,
     paddingRight: 0,
+    minWidth: '50px',
   },
   outline: {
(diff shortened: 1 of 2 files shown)


packages/manager/src/components/Button/Button.tsx
```diff
@@ -51,6 +51,7 @@ const useStyles = makeStyles((theme: Theme) => ({
   superCompact: {
     paddingLeft: 0,
     paddingRight: 0,
+    minWidth: '50px',
```
johnwcallahan, line 54: This is a global change, but I spot checked a couple places and this fixed similar bugs there (with no regressions that I could find).

johnwcallahan (approved):
Thanks for jumping on this one, nice work.

DevDW (approved):
🚢

PR #7632, opened 2021-05-05
TPA Feedback

tiffwong:
## Description

Manually add Linode as a provider to reduce repetitive code and make it easier to read. Also conditionally hide text that contains "Google" until TPA is actually released

### DIFF
--- a/packages/manager/src/featureFlags.ts
+++ b/packages/manager/src/featureFlags.ts
@@ -75,10 +75,8 @@ export interface MainContentBanner {
   key: string;
 }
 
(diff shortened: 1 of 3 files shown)


DevDW (approved):
LGTM

This is a contrived case since, at minimum, Github should always be a `tpaProvider`, but when I set `tpaProviders: []` in `useFlags.ts`:

Screen Shot 2021-05-06 at 12 54 15 PM

"...another provider such as Github" even though there are no options aside from Linode.

johnwcallahan (approved):

PR #7634, opened 2021-05-06
Change: Refactor StackScript requests and display logic to use new "mine" field

Jskobos:
API has added a new field to the StackScript envelope, "mine", which is true
if the StackScript is on your account (created by you or a user on your account).
This field is filterable. Using the field and filter we can replace a few
awkward workarounds we'd previously relied on:

1. Getting a list of "Account" stackscripts was difficult, and our approach
relied on (for restricted users) iterating through a list of SS grants and
requesting each individually. For unrestricted users, we requested a list
of users on the account then filtered the full list of scripts by those
usernames. We can now replace these with a single request filtered by "mine",
which should handle community and account lists for both restricted and
unrestricted users.

2. The same problem occurred in StackScripts detail, there wasn't enough
information in a StackScript object to tell us (when loading a details
page directly) "can this user edit this script?" and "when navigating back
to the landing page, which tab should we load?" without some workarounds.
Using "mine" and existing permissions logic, we can make this simpler and more
accurate.

I hoped to do a bigger refactor here, but StackScripts are very complicated and it was too deep a rabbit hole to go down here. Please test with as many combinations as possible (you'll need to point at the alpha API):

1. Unrestricted user: account and community SS tabs, public and private scripts,
2. Restricted user: Same as above, plus at least one script that the user has read_only access to.


DevDW (approved):
LGTM; I tested public and private account StackScripts and community StackScripts in both unrestricted and restricted user cases.

I did observe that the mocks never populated anything for StackScripts when I turned them on, though.

tiffwong (approved):
Woah, this is awesome! I tested the different permission levels for restricted and unrestricted users on account and community StackScripts and they all look great!

PR #7641, opened 2021-05-10
Storybook update pt1

tiffwong:
## Description

Import the font into Storybook (this was lost in the update) and cleanup the Button story ([http://localhost:6006/?path=/story/button--default](http://localhost:6006/?path=/story/button--default))

Next step is to figure out how to use `mdx` files to build out the documentation for our components


packages/manager/.storybook/preview.tsx
```diff
@@ -1,4 +1,36 @@
...
+  (Story) => {
+    const _key = select('theme', ['light', 'dark'], 'light');
+
+    return wrapWithTheme(
+      <ThemeProvider theme={options[_key]}>
+        <CssBaseline />
+        <div
+          style={{
+            margin: '1rem',
+            padding: '4rem',
```
Jskobos, line 27: Maybe we could set the background to be 100% size, not sure what's best.

Jskobos (approved):
Looks good, I don't remember what we used to do for dark theme so I threw on a background and some padding. Please adjust as needed.

DevDW (approved):
Looks good 🚢

PR #7647, opened 2021-05-11
M3-5057: TextField error message positioning and width

tiffwong:
## Description

This PR moves the error message generated from the breadcrumb below the input field for all breakpoints. This should prevent the error message from colliding with the CTAs and from wrapping too soon below input fields in a form.

ie. 
Screen Shot 2021-04-09 at 2 33 21 PM

Another small change is that the username is dropped from the StackScripts breadcrumb since it's redundant and makes the breadcrumb way too long.

## How to test

Please check the error message for TextField components throughout the app at different viewports. You can find an editable breadcrumb on the Linode, NodeBalancer, Firewall, StackScript, Domain, and Kubernete detail pages.


packages/manager/src/features/StackScripts/StackScriptsDetail.tsx
```diff
@@ -165,29 +143,17 @@ export const StackScriptsDetail: React.FC<CombinedProps> = (props) => {
...
-        container
-        className={classes.root}
-        alignItems="center"
-        justify="space-between"
-      >
+      <Grid container className={classes.root} justify="space-between">
         <Grid item className="p0">
           <Breadcrumb
             pathname={props.location.pathname}
-            labelOptions={{ prefixComponent: userNameSlash, noCap: true }}
```
DevDW, line 190: I think we need to keep `noCap: true` here, otherwise we end up with the StackScript breadcrumb having capital letters that aren't present in the StackScript name:

Before–
Screen Shot 2021-05-17 at 12 40 01 PM

After–
Screen Shot 2021-05-17 at 12 40 05 PM

DevDW (approved):
Left a comment for one small fix but LGTM to me overall

bnussman (commented):
Looks good, only small issue I found was minor overlap with Inputs when the error message was large and message was many lines. 
Screen Shot 2021-05-18 at 11 45 03 AM

tiffwong:
@bnussman the input fields will actually be full width for smaller viewports once #7659 is merged which should prevent the error message from overlapping

bnussman (approved):

PR #7652, opened 2021-05-12
M3-5130 added link to documentation on rescue modal

bnussman:
## Description

Added a link to Linode's ['Rescue and Rebuild' documentation](https://www.linode.com/docs/guides/rescue-and-rebuild/) on the Rescue Dialog. 

I modeled this off of the Resize Dialog in `LinodeResize.tsx`.

## How to test

Open the **rescue** modal for any Linode and verify there is a 'Learn more.' link that brings you to the documentation linked above.


Jskobos (changes requested):
This looks good, but we should add the same thing to the bare metal version of this dialog in BareMetalRescue.tsx (this is why rescue helper text is a shared constant). Probably you could add the link into the shared text, if you adjust the typings so that everything knows it's JSX and not a string.

bnussman:
Made changes to add it to Bare Metal. If we prefer shared.tsx over RescueDescription.tsx please let me know!

Jskobos (approved):
Thanks for the change, I think the RescueDescription component works well. An optional cleanup would be to move the `<Typography />` inside the new component so that it's applied consistently in both places.

DevDW (approved):
Nice work 🚢

PR #7654, opened 2021-05-13
M3-5122 create cluster UI updates

bnussman:
## Description

Updates the Create Cluster page to new UI spec and sets the default number of nodes to 3 in the Node Pool Panel. 

Copy and Kubernetes documentation link for less than 3 nodes needs to be updated and finalized. 

## How to test

Visit `/kubernetes/create` to view the UI changes and ensure pool selection still works as expected.


packages/manager/src/components/CheckoutBar/styles.ts
```diff
@@ -35,7 +35,7 @@ export const useStyles = makeStyles((theme: Theme) => ({
     border: 0,
   },
   sidebarTitle: {
-    fontSize: '1.5rem',
+    fontSize: '24px',
```
tiffwong, line 38: Was there a reason why you changed this from `rem` to `px`?

packages/manager/src/components/EnhancedNumberInput/EnhancedNumberInput.tsx
```diff
@@ -8,8 +8,8 @@ import TextField from 'src/components/TextField';
 
 const useStyles = makeStyles(() => ({
   button: {
-    width: 40,
-    height: 40,
+    width: 35,
+    height: 34,
```
tiffwong, line 12: These aren't actually being applied to the plus and minus buttons, they're being overwritten by the dimensions on L52-53

packages/manager/src/features/Kubernetes/KubeCheckoutBar/KubeCheckoutBar.tsx
```diff
@@ -56,7 +67,14 @@ export const KubeCheckoutBar: React.FC<Props> = (props) => {
           />
         ))}
         {showWarning && (
-          <Notice warning important text={nodeWarning} spacingTop={16} />
+          <Notice warning important spacingTop={16}>
+            <Typography className={`${classes.notice} noticeText`}>
+              {nodeWarning}
+              <Link to="https://kubernetes.io/docs/setup/production-environment/">
+                Kubernetes docs and resources.
+              </Link>
```
tiffwong, line 75: This should probably link to a Linode docs but as for which one, I'm not sure

tiffwong (changes requested):
Great job getting this out! 

According to the design, the outlines for the plus and minus buttons should be the same color as the input field (#cccccc)
Screen Shot 2021-05-13 at 2 17 05 PM

The divider and the plan name is looking a little tight, can you also increase the margin to 4px?
Screen Shot 2021-05-13 at 2 18 31 PM

I left a few other cleanup items but otherwise, this is awesome!

packages/manager/src/components/EnhancedNumberInput/EnhancedNumberInput.tsx
```diff
@@ -122,7 +122,8 @@ export const EnhancedNumberInput: React.FC<FinalProps> = (props) => {
       })}
     >
       <Button
-        buttonType="primary"
+        buttonType="secondary"
+        outline={true}
```
Jskobos, line 126: `outline` is fine here, the `={true}` is explicit which is nice but not needed in this case imo.

packages/manager/src/features/Kubernetes/kubeUtils.ts
```diff
@@ -6,7 +6,7 @@ import { LinodeType } from '@linode/api-v4/lib/linodes';
 import { pluralize } from 'src/utilities/pluralize';
 import { ExtendedCluster, ExtendedPoolNode, PoolNodeWithPrice } from './types';
 
-export const nodeWarning = `We recommend at least 3 nodes in each pool. Fewer nodes may affect availability.`;
+export const nodeWarning = `We recommend at least 3 nodes in each pool. Fewer nodes may affect availability. For more info see `;
```
Jskobos, line 9: Having a sentence fragment inside a constant is a bit risky. In this case, the warning is used elsewhere so you end up with:

Screen Shot 2021-05-14 at 9 32 47 AM

...in the Resize Pool drawer in the LKE detail view.

Probably the best thing to do is make another display component like you did for Rescue and include the link in there.

Jskobos (changes requested):
Nice work, I have just a few suggestions in addition to Tiffany's. Clicking "Add" for one of the number inputs resets its count to 0 (this is controlled on line 103 of NodePoolPanel.tsx). This is ok behavior, it might even be what we want, but it's perfectly possible to have multiple node pools of the same type and this might discourage that. Please check with Matthew, and depending on what he says you might want to reset to 3 on submit instead of 0.

tiffwong (approved):

packages/manager/src/features/Kubernetes/KubeNodeWarning.tsx
```diff
@@ -0,0 +1,40 @@
...
+      fontSize: '1rem',
+    },
+  },
+}));
+
+interface Props {
+  spacingTop?: 0 | 8 | 16 | 24;
+}
+
+function KubeNodeWarning(props: Props) {
```
DevDW, line 24: to be consistent with our current patterns, I'd say change this to

```
const KubeNodeWarning: React.FC<Props> = (props) => {
...
```

DevDW (approved):
Looks good, left one minor comment.

This is an existing issue, but I noticed in mobile view that the number input is larger than the plus and minus signs:

Screen Shot 2021-05-17 at 11 13 51 AM

PR #7655, opened 2021-05-13
M3-5083: Dismissible Images pricing banner

DevDW:
## Description
Adds a dismissible banner that explains Images pricing to the Images landing and and Images creation pages. The displaying of the banner is conditional upon the feature flag and whether the user has images or not.

## How to test
The feature flag is on in the testing environment, so please ensure that the banner does not display on Images pages when you have no images, and that the banner does display on Images pages when you do have images.

The banner should not return on its own after being dismissed (if you dismiss it and want to bring it back for further testing, you can hit the `profile/settings?preferenceEditor=true` page and remove the corresponding entry from `dismissed_notifications`).


bnussman (approved):
Looks good to me!

codyfinn (approved):
Looks good to me!

tiffwong (changes requested):
Functionality looks good, only thing I noticed is that the padding between the vertical bar and copy is a bit tight. I attached a screenshot of the banner on a Kubernetes that needs updating which is more consistent with the spacing

Screen Shot 2021-05-14 at 2 45 15 PM
Screen Shot 2021-05-14 at 2 45 22 PM

bnussman (approved):
Correct me if I'm wrong but, I think the only issue I could see is if the feature flag data is bad. 

Ex. If the `text` field does not end with a trailing space then we'll get this. 
example

Otherwise, looks good.

DevDW:
> Correct me if I'm wrong but, I think the only issue I could see is if the feature flag data is bad.
> 
> Ex. If the `text` field does not end with a trailing space then we'll get this.
> example

Yes, I've vacillated once or twice in this PR between including the space within the code or forcing it to be included in the text string in LD. I switched it back to be in the code instead; maybe that's the safer option.

packages/manager/src/features/Images/ImagesPricingBanner.tsx
```diff
@@ -0,0 +1,72 @@
...
+    marginBottom: theme.spacing(),
+  },
+}));
+
+export const ImagesPricingBanner: React.FC<{}> = (_) => {
+  const { images } = useImages('private');
+  const flags = useFlags();
+  const classes = useStyles();
+
+  const bannerHasContents = checkFlagsForImagesPricingBanner(flags);
```
johnwcallahan, line 25: The `checkFlagsForImagesPricingBanner` function does path checking the old school JS way. It'd be more succinct and stylistic to use optional chaining, e.g. `Boolean(flags?.imagesPricingBanner?.key)` (like you do a few lines down). 

In fact, you could use `bannerCoreText` for this instead to make it more streamlined:

```ts
const bannerCoreText = flags.imagesPricingBanner?.text;

if (images.results === 0 || !bannerCoreText) {
  return null;
}
```

packages/manager/src/features/Images/ImagesPricingBanner.tsx
```diff
@@ -0,0 +1,72 @@
...
+    return null;
+  }
+
+  const bannerCoreText = flags.imagesPricingBanner?.text;
+
+  const bannerLede = flags.imagesPricingBanner?.lede ?? '';
+  const bannerLinkUrl = flags.imagesPricingBanner?.link?.url ?? '';
+  const bannerLinkText = flags.imagesPricingBanner?.link?.text ?? '';
+
+  const generateLede = () => {
```
johnwcallahan, line 37: Any particular reason you made these functions? I separating them from this FC's `return` value is good, since it allows us to see the structure of the content, but you could easily do:

```tsx
const Lede = bannerLede ? (
  <span>
    <strong>{bannerLede}</strong>:{' '}
  </span>
) : null;
```

... instead of defining separate functions.

johnwcallahan (changes requested):
The approach looks good; just have a few stylistic things.

HanaXu (approved):
LGTM!

tiffwong (approved):

johnwcallahan (approved):

PR #7656, opened 2021-05-13
Update LinodeDiskActionMenu.tsx

mighteejim:
## Description

This PR prevents disks from being deleted while a Linode is running.

## How to test

1. Create and boot a Linode using any distro.
2. Navigate to the "storage" tab.
3. Attempt to delete one of the disks under the Disks table.
4. Power off your Linode and repeat step 3.

Screenshots are attached.

CleanShot 2021-05-13 at 13 02 49@2x
CleanShot 2021-05-13 at 13 02 43@2x
CleanShot 2021-05-13 at 13 04 01@2x

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeAdvanced/LinodeDiskActionMenu.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeAdvanced/LinodeDiskActionMenu.tsx
@@ -70,8 +70,7 @@ export const DiskActionMenu: React.FC<CombinedProps> = (props) => {
       onClick: () => {
         props.onResize();
       },
-      disabled: linodeStatus !== 'offline' || readOnly,
-      tooltip: linodeStatus !== 'offline' || readOnly ? _tooltip : '',
+      ...disabledProps,
     },
     {
       title: 'Imagize',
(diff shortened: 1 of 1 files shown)


packages/manager/src/features/linodes/LinodesDetail/LinodeAdvanced/LinodeDiskActionMenu.tsx
```diff
@@ -92,7 +92,9 @@ export const DiskActionMenu: React.FC<CombinedProps> = (props) => {
       onClick: () => {
         props.onDelete();
       },
-      ...(readOnly ? disabledProps : {}),
+      
+      disabled: linodeStatus !== 'offline' || readOnly,
+      tooltip: linodeStatus !== 'offline' || readOnly ? _tooltip : '',
```
DevDW, line 97: https://github.com/linode/manager/blob/fcc49952a0c069431abbc526a3169a3280ab4dee/packages/manager/src/features/linodes/LinodesDetail/LinodeAdvanced/LinodeDiskActionMenu.tsx#L44-L57

This code block seems to account for the "if the Linode is off or the user has readOnly permission" logic, so I think replacing these couple of lines (and the same lines for the "Resize" action) with `...disabledProps` would cut back on unnecessary repetition.

DevDW (approved):
Thanks for the PR, Jim!

tiffwong (approved):
Approved pending Dajahi's feedback 🎉

packages/manager/src/features/linodes/LinodesDetail/LinodeAdvanced/LinodeDiskActionMenu.tsx
mighteejim: Done! Thanks for the suggestion, much simpler.

PR #7657, opened 2021-05-14
M3-4799 create cluster button flickering fix

codyfinn:
## Description

This PR fixes the rendering issues on the Kubernetes create cluster screen for chromium based browsers. This is accomplished by removing an unused radio button that was `visually-hidden` which interacted strangely with the `EnhancedNumberInput` component.

## How to test

Open the manager in a chromium based browser and navigate to the create Kubernetes cluster screen. Shrink the browser window height enough so that the linode plan quantity selection table is just out of view. Refresh the page. Scroll down to the plan quantity selection table. All the quantity selectors should be visible and hovering over any of the fields or buttons should not cause any other field/button to disappear.

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/SelectPlanQuantityPanel.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/SelectPlanQuantityPanel.tsx
@@ -5,7 +5,6 @@ import { isEmpty, pathOr } from 'ramda';
 import * as React from 'react';
 import { compose } from 'recompose';
 import Button from 'src/components/Button';
-import FormControlLabel from 'src/components/core/FormControlLabel';
 import Hidden from 'src/components/core/Hidden';
 import {
   createStyles,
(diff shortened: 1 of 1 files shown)


tiffwong (approved):
Woah, this is awesome! Thanks for solving the mystery of the disappearing buttons! 🕵️‍♂️ 🔍 🎉

bnussman (approved):
No more flickering on the create cluster page! 🎉

johnwcallahan (commented):
Getting rid of the visually hidden element likely has an impact on accessibility.

tiffwong (approved):
I ran lighthouse and the score is the same for this branch and develop. I also checked keyboard accessibility using VoiceOver, JAWSs and NVDA which all work as expected so I don't think there should be an issue with removing the radio button

Can you remove the `role="radiogroup"` from [line 249](https://github.com/linode/manager/pull/7657/files#diff-ef9d119d258fb518fe9efe9a6e7201fdfefbde4694efb635e2fa99a174102e7bR249) though? We're not using radio buttons inside that section so I don't think we need that and it's actually lowering our score on lighthouse

johnwcallahan (approved):

PR #7658, opened 2021-05-17
M3-5119: added getting_started.sh script

codyfinn:
## Description

Adds a bash script to install `Homebrew`, `Volta`, `Git`, and `GitHub CLI`.


bnussman (commented):
Worked great when I used it! Also tested it on Pop_OS! 20.10 and it worked flawlessly so Linux users should also have success with the script. 

If we think the script is solid, it would be nice to see it as an option in the [Getting Started docs](https://github.com/linode/manager/blob/develop/docs/GETTING_STARTED.md).

bnussman (approved):

getting_started.sh
```diff
@@ -0,0 +1,74 @@
...
+bash_config=".bash_profile"
+zsh_config=".zshrc"
+echo export VOLTA_HOME='$HOME/.volta' >> "$HOME/$bash_config"
+echo export PATH='$VOLTA_HOME/bin:$PATH' >> "$HOME/$bash_config"
+echo export VOLTA_HOME='$HOME/.volta' >> "$HOME/$zsh_config"
+echo export PATH='$VOLTA_HOME/bin:$PATH' >> "$HOME/$zsh_config"
+
+volta install node@14.15.4
+volta install yarn@1
+if [ "$1" == '--no-fork' ]; then
```
johnwcallahan, line 62: What would you think about the following changes:

- By default (without a flag) we clone `linode/manager`. 
- With a `--fork` flag (or similar) we fork `linode/manager` and clone the fork (like you've got here).

That way you always end up with an install of Cloud, even if you don't want to fork it yourself.

I know I mentioned having the option to install dependencies only, which may still be useful, but I think the default case described above might be the most common use case.

johnwcallahan (commented):
Works great when I run this on my mac.

I agree that `getting_started.md` should be updated, and I'd also suggest adding a few print statements throughout the script letting the user know exactly what's happening.

getting_started.sh
codyfinn: I think we should default to fork and clone. Since this is targeted at people who are not super familiar with the terminal. This way if they just run a curl command that downloads the `getting-started.sh` and runs it they end up with a setup that matches what the current getting started documents suggest. Then if a more experienced user wants to do the forking and clone themselves then they can and then run the script with a `--no-fork` command.

getting_started.sh
```diff
@@ -0,0 +1,74 @@
...
+              ||..  '  .|'     '   ||   ||   ||   ||   ||    ||
+               ''|||.  ||          ||''|'    ||   ||...|'    ||
+             .     '|| '|.      .  ||   |.   ||   ||         ||
+             |'....|'   ''|....'  .||.  '|' .||. .||.       .||.
+
+
+
+EOF
+
+Install homebrew https://brew.sh
```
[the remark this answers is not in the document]
codyfinn: Yeah thanks

docs/GETTING_STARTED.md
```diff
@@ -1,5 +1,14 @@
 # Getting Started
 
+## Using the getting_started.sh script
+1. Open the terminal application
+2. Paste this into the terminal appliation `/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/codyfinn/manager/M3-5119-startup-script/getting_started.sh)"`
```
johnwcallahan, line 5: Even though it won't exist until you merge it, I'd go ahead and change this so it looks at linode/manager/develop instead of your fork.

Last bit of feedback... the getting started script should probably go in the `scripts` directory.

johnwcallahan (approved):

PR #7659, opened 2021-05-17
M3-5123: Update Credit Card Drawer

tiffwong:
## Description

Tighten up spacing, add accepted credit cards, and right-align buttons

screencapture-localhost-3000-account-billing-2021-05-17-14_43_23

## How to test

Navigate to `/account/billing` and click "Edit" in the "Payment Method" panel


bnussman:
Seems like the `contact information` link is not working in the Update Credit Card drawer. This is also true for production... Maybe it an easy fix, but maybe we need another issue for it.

bnussman:
Also looks like the Expiration Date error message will overflow offscreen. This is also an existing problem with manager.
Screen Shot 2021-05-18 at 9 41 50 AM

tiffwong:
@bnussman great catch! I just pushed up a fix for the link and the overflowing error message should be fixed once #7647 is merged 🤞

bnussman (approved):

HanaXu (approved):
LGTM, just one minor comment. We should probably right align the buttons in the contact info drawer to be more consistent with this drawer (and probably swap the cancel and ok buttons as well). Also something to keep in mind for other drawers as part of the UI audit.

PR #7629, opened 2021-05-03
M3-5128: Add docs link in "Last Backup" Bare Metal column tooltips

DevDW:
## Description
- Adds a link to our ["Backing Up Your Data"](https://www.linode.com/docs/guides/backing-up-your-data/) guide in the tooltip
- Refactors file to use `useStyles()`

## How to Test
Use an account that has a Bare Metal Linode and navigate to the Linodes landing table. Check the tooltip in the "Last Backup" cell

PR #7660, opened 2021-05-17
M3-5129:  Added connect with LISH copy to rescue modal for Bare Metal instances

bnussman:
## Description

Adds copy to the Rescue Dialog for Bare Metal instances informing the user that they should use the LISH console to connect to their Linode upon rescue. Also includes a link that directly opens LISH for them. 

screenshot

## How to test

Open the Rescue Dialog on a Bare Metal Linode and ensure copy and links are good.

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDescription.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDescription.tsx
@@ -35,7 +35,7 @@ const RescueDescription: React.FC<Props> = (props) => {
         {rescueDescription.text}{' '}
         <Link to={rescueDescription.link}>Learn more.</Link>
       </Typography>
-      {linodeId && (
+      {linodeId ? (
         <Typography className={classes.copy}>
           {`When your Linode has successfully rebooted into Rescue Mode, use the `}
           <button
@@ -49,7 +49,7 @@ const RescueDescription: React.FC<Props> = (props) => {
             Learn more.
           </Link>
         </Typography>
-      )}
+      ) : null}
     </React.Fragment>
   );
 };
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDescription.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDescription.tsx
@@ -1,8 +1,8 @@
 import * as React from 'react';
-import ExternalLink from 'src/components/ExternalLink';
 import Typography from 'src/components/core/Typography';
 import { lishLaunch } from 'src/features/Lish/lishUtils';
 import { makeStyles, Theme } from 'src/components/core/styles';
+import Link from 'src/components/Link';
 
 const rescueDescription = {
   text: `If you suspect that your primary filesystem is corrupt, use the Linode Manager to boot your Linode into Rescue Mode. This is a safe environment for performing many system recovery and disk management tasks.`,
@@ -10,7 +10,7 @@ const rescueDescription = {
 };
 
 interface Props {
-  linodeId: number;
+  linodeId?: number;
 }
 
 const useStyles = makeStyles((theme: Theme) => ({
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDescription.tsx
```diff
@@ -1,20 +1,56 @@
...
+
+const RescueDescription = (props: Props) => {
+  const { linodeId } = props;
+  const classes = useStyles();
+
+  return (
+    <React.Fragment>
+      <Typography>
+        {rescueDescription.text}{' '}
+        <ExternalLink
```
DevDW, line 31: We can use the `<Link />` component instead here; we're moving in the direction of getting rid of `<ExternalLink />` since `<Link />` does the same thing when the URL starts with "https": https://github.com/linode/manager/pull/7629#discussion_r625779478

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDescription.tsx
```diff
@@ -1,20 +1,56 @@
...
+          fixedIcon
+          text="Learn more."
+          link={rescueDescription.link}
+        />
+      </Typography>
+      <Typography className={classes.copy}>
+        {`When your Linode has successfully rebooted into Rescue Mode, use the `}
+        <button
+          className={classes.lishLink}
+          onClick={() => lishLaunch(linodeId)}
```
DevDW, line 41: When I clicked this, the LISH console that popped up was never able to validate the token. Is this also happening for you?

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDescription.tsx
```diff
@@ -1,20 +1,56 @@
...
+        >
+          LISH Console
+        </button>
+        {` to access it. `}
+        <ExternalLink
+          fixedIcon
+          text="Learn more."
```
DevDW, line 48: I wonder if we should say "Learn more" twice in the same modal? 🤔

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDescription.tsx
```diff
@@ -1,20 +1,56 @@
...
-    {RESCUE_HELPER_TEXT}{' '}
-    <ExternalLink
-      fixedIcon
-      text="Learn more."
-      link="https://www.linode.com/docs/guides/rescue-and-rebuild/"
-    />
-  </Typography>
-);
+interface Props {
+  linodeId: number;
```
DevDW, line 13: Since this was added for the logic launching the LISH console and that copy should only pop up for Bare Metal instances, let's go ahead and make this an optional prop, and remove the passing in of the linodeId to this component in `RescueDialog.tsx`

DevDW (changes requested):

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDescription.tsx
bnussman: LISH is opening successfully for me, I'll look into it.

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDescription.tsx
DevDW: @matthewbartholomew Any particular opinion on this?

DevDW (approved):
LGTM now 👌 

I still haven't been able to get the LISH console up and running for a Bare Metal instance, but that's also happening when I try it from the Linode Details page (so it's not an issue with how it's being invoked here). I'm not sure what's causing this for the instances I'm testing with, but if everyone else is able to successfully launch LISH with theirs, we should be good to go.

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDescription.tsx
```diff
@@ -1,20 +1,57 @@
...
+   * connection with LISH upon reboot into Rescue Mode. This is
+   * intended for the Bare Metal dialog.
+   */
+  return (
+    <React.Fragment>
+      <Typography>
+        {rescueDescription.text}{' '}
+        <Link to={rescueDescription.link}>Learn more.</Link>
+      </Typography>
+      {linodeId && (
```
johnwcallahan, line 38: A while back Jared convinced us that using ternaries is better, since using the logical `&&` operator can sometimes result in sneaky bugs. So this would be `linodeId ? (...) : null`.

johnwcallahan (approved):

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDescription.tsx
bnussman: Good to know! I was wondering why everyone was doing that... I'll make that change.

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDescription.tsx
matthewbartholomew: I think we can drop the second Learn More link. I expect that users who need more guidance will click on the first link even if they need info about using the LISH Console.

PR #7663, opened 2021-05-18
[DO NOT MERGE] add alma distro

acourdavault:
## Description

add alma distro


packages/manager/src/utilities/images.ts
```diff
@@ -15,6 +15,7 @@ import {
 } from 'ramda';
 
 export const distroIcons = {
+  Alma: 'alma',
```
acourdavault, line 18: @johnwcallahan i think in DB it will be `linode/almalinux8`

PR #7668, opened 2021-05-19
added time for scheduled migrations to banner

bobby-palko:
## Description

This adds a timestamp of when a Linode will enter the queue for a scheduled migration. If the time is within the next 24 hours, it also displays the rounded hours remaining. This updates the banner on the Linode page as well as the text in the notification center. 

## How to test

Configure a cold migration for a Linode and verify the time appears


packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
```diff
@@ -152,12 +153,12 @@ const interceptNotification = (
   if (notification.type === 'migration_scheduled') {
     const jsx = (
       <Typography>
-        You have a scheduled migration pending!{' '}
+        You have a migration pending for{' '}
```
HanaXu, line 156: We should keep the `scheduled` wording since `migration_scheduled` and `migration_pending` are different notification types.

i.e. `You have a scheduled migration for`

HanaXu (changes requested):
Looks fine functionally! Left some comments on the wording

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/MigrationNotification.tsx
```diff
@@ -86,11 +91,28 @@ const MigrationNotification: React.FC<Props> = (props) => {
       ? 'enter the migration queue now'
       : 'schedule your migration';
 
+  const migrationText = () => {
+    let returnText = `You have a migration pending, which will automatically execute `;
```
HanaXu, line 95: Same here: `You have a scheduled migration, which will automatically execute `

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/MigrationNotification.tsx
```diff
@@ -86,11 +91,28 @@ const MigrationNotification: React.FC<Props> = (props) => {
...
+    const formattedMigrationTime = formatDate(migrationTime as string);
+
+    const now = DateTime.local();
+    const hourDifference = migrationTimeObject.diff(now, 'hours').as('hours');
+
+    return hourDifference <= 24
+      ? (returnText += `in approximately ${Math.round(
+          hourDifference
+        )} hour(s) (${formattedMigrationTime}).`)
+      : (returnText += `on ${formattedMigrationTime}.`);
```
HanaXu, line 108: `returnText` can be part of the template literal, though we'd probably want a better variable name

packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
bobby-palko: Makes sense to me. Updated!

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/MigrationNotification.tsx
bobby-palko: Same here. Updated.

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/MigrationNotification.tsx
bobby-palko: I honestly don't know why I didn't have it part of the template literal in the first place. It has been renamed to `baseText` and the template now includes the variable.

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/MigrationNotification.tsx
```diff
@@ -86,11 +91,28 @@ const MigrationNotification: React.FC<Props> = (props) => {
       ? 'enter the migration queue now'
       : 'schedule your migration';
 
+  const migrationText = () => {
+    const baseText = `You have a scheduled migration pending, which will automatically execute`;
```
HanaXu, line 95: Same here. We should also update the `migrationActionDescription` variable to  `schedule your migration now` because the action will always be to enter the migration queue

packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
```diff
@@ -152,12 +153,12 @@ const interceptNotification = (
   if (notification.type === 'migration_scheduled') {
     const jsx = (
       <Typography>
-        You have a scheduled migration pending!{' '}
+        You have a scheduled migration pending for{' '}
```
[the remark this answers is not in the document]
bobby-palko: updated

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/MigrationNotification.tsx
bobby-palko: also updated

HanaXu (approved):
LGTM! :tada:

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/MigrationNotification.tsx
```diff
@@ -84,13 +89,30 @@ const MigrationNotification: React.FC<Props> = (props) => {
...
+
+    const formattedMigrationTime = formatDate(migrationTime as string);
+
+    const now = DateTime.local();
+    const hourDifference = migrationTimeObject.diff(now, 'hours').as('hours');
+
+    return hourDifference <= 24
+      ? `${baseText} in approximately ${Math.round(
+          hourDifference
+        )} hour(s) (${formattedMigrationTime}).`
```
DevDW, line 107: Let's pull in `pluralize` here to handle the hour/hours issue more cleanly:

`import { pluralize } from 'src/utilities/pluralize';`

then I'd do something like:

```
    const roundedHourDifference = Math.round(
      migrationTimeObject.diff(now, 'hours').as('hours')
    );
    const approximateTime = pluralize('hour', 'hours', roundedHourDifference);

    return roundedHourDifference <= 24
      ? `${baseText} in approximately ${approximateTime} (${formattedMigrationTime}).`
      : `${baseText} on ${formattedMigrationTime}.`;
  };
```

DevDW (commented):
Thanks for the PR! Left one cleanup comment and I want to do a bit more testing to make sure the type castings hold up under different circumstances; looks good so far though.

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/MigrationNotification.tsx
bobby-palko: brought in `pluralize` for the hour/hours issue. the casting _should_ be fine; it's only casted since `notification.when` may not get passed from the API (in which case this logic won't fire anyway), so the prop is declared as `string | null`. since luxon requires a string, it's only casted to get rid of the `null` type and make TS happy. lmk if you need anything else!

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/MigrationNotification.tsx
```diff
@@ -84,13 +90,31 @@ const MigrationNotification: React.FC<Props> = (props) => {
...
+      ? `${baseText} in approximately ${approximateTime} (${formattedMigrationTime}).`
+      : `${baseText} on ${formattedMigrationTime}.`;
+  };
 
   return (
     <>
       <Notice important warning>
         <Typography>
-          {notificationMessage}
+          {migrationTime ? migrationText() : notificationMessage}
```
DevDW, line 117: I think this may need to be a check for `migration_scheduled` instead. Right now, it seems that `migration_pending` notifications will be handled the same way:

https://github.com/linode/manager/blob/59c7a88a974ff7163f18bc3d8ecdd1c5beef4b65/packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/Notifications.tsx#L46-L60

Therefore if you were to go to a Linode with a pending migration, it'd have a scheduled migration banner.

DevDW (changes requested):
Upon further testing I haven't seen any issues with the type casting, but realized that `migration_pending` and `migration_scheduled` are now showing the same notification banner on the Linode details page. (you can turn on the mocks for testing)

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/MigrationNotification.tsx
bobby-palko: Interesting. I tested this by scheduling all sorts of different migrations for one of my Linodes, and both banners showed as I would suspect. Regardless, I've updated this to check for the `notificationType`, as it's a cleaner approach anyway. I'd be interested in learning how to use the mocks when you have some time (just hit me up on Slack - @rpalko)

DevDW (approved):
For the sake of clarity I'd probably rename `migrationText` to something like `migrationScheduledText`, but the functionality LGTM now 👍  will reach out re: the mocks

PR #7669, opened 2021-05-20
M3-5172: Change Images landing tables to have default sorting by label ascending

DevDW:
## Description
The default sorting for Images landing tables should now be by label, ascending (instead of label, descending).


johnwcallahan (approved):
Works great!

DevDW:
> LGTM! Just one comment relating to how strings are currently sorted.
> 
> Is this intentional?
> image

Yes, this is the current behavior across the app currently. @johnwcallahan Is this behavior that should be revisited in another ticket?

johnwcallahan:
> @johnwcallahan Is this behavior that should be revisited in another ticket?

Good eye, but I think this is OK.

PR #7670, opened 2021-05-20
M3-5113: Fix viewing Object Storage contents

bnussman:
## Description

Addresses potential issue with loading paginated data from the API  when loading objects in an object storage bucket.

Some users experience incomplete results when loading their objects.  It appears that the API sometimes does not respect the page number query parameter resulting in a response that tells the client: "here are a few items (less than the quested page number), but there are still more to load." When the API does that, it also returns a nextMarker that does not make sense. This may be an issue with the API or Object Storage. 

This PR fixes this edge case by removing the check `results >= 100` so our infinite scroll waypoint renders depending on `allObjectsFetched` which is going to be accurate. The actual bug fix is in `BucketDetail.tsx`.

This PR also enables the MSW to mock objects in a bucket. This can be factored out of the PR if needed.   

## How to test

Go to an object storage bucket and make sure infinite scroll (pagination) works as expected. Test on large accounts. 

Also ensure the mock service worker still works as expected and see if it correctly mocks objects in object storage.

### DIFF
--- a/packages/manager/src/factories/objectStorage.ts
+++ b/packages/manager/src/factories/objectStorage.ts
@@ -39,3 +39,14 @@ export const objectStorageObjectFactory = Factory.Sync.makeFactory<ObjectStorage
     name: Factory.each((id) => `example-${id}`),
   }
 );
+
+export const makeObjectsPage = (
+  e: ObjectStorageObject[],
+  override: { is_truncated: boolean; next_marker: string | null }
+) => ({
+  data: e,
+  is_truncated: override.is_truncated || false,
+  next_marker: override.next_marker || null,
+});
+
+export const staticObjects = objectStorageObjectFactory.buildList(250);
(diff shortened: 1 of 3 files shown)


HanaXu:
Hmm this looks like it might be an API issue? Since we are specifying/it is the default to return 100 items per page, but we are getting `is_truncated: true` after 10 items in the root id folder specifically.

If I click on one of the folders in the id folder, we are correctly getting `is_truncated: false` and seem to be loading up to the 100 limit.

Have you tried checking with someone on the API team about this? @johnwcallahan would probably know who to contact

HanaXu (commented):
Hmm this looks like it might be an API issue? Since we are specifying/it is the default to return 100 items per page, but we are getting `is_truncated: true` after 10 items in the root id folder specifically.

If I click on one of the folders in the id folder, we are correctly getting `is_truncated: false` and seem to be loading up to the 100 limit.

Have you tried checking with someone on the API team about this? @johnwcallahan would probably know who to contact

johnwcallahan (approved):
Tested this out several accounts/buckets and fixes the issue, and I haven't found any regressions. 

Follow up with the Storage team to make sure we understand what's happening here, but barring anything unusual I'm in favor of merging. 

Great job with the pagination logic in serverHandlers.

HanaXu (approved):

PR #7673, opened 2021-05-21
M3-5000: Auto fill request description when opening support ticket

HanaXu:
## Description
If additional verification is required, autofill the request description with the plan the user was trying to create.

## How to test
Try to create a Linode on an account that requires additional verification.

### DIFF
--- a/packages/manager/src/request.tsx
+++ b/packages/manager/src/request.tsx
@@ -55,26 +55,28 @@ export const handleError = (error: AxiosError) => {
     },
   } = store.getState();
 
-  const displayTitle = entities.find((e) => e.id === requestData.type)?.label;
+  const planTitle = entities.find((e) => e.id === requestData.type)?.label;
 
-  const requestedLinode = {
-    title: `${
-      requestData?.type?.match(/gpu/i) ? 'GPU' : 'Verification'
-    } Request for a ${displayTitle} Plan`,
-    description:
-      'Hello. I am trying to deploy a ' +
-      displayTitle +
+  const titlePrefix = `${
+    requestData?.type?.match(/gpu/i) ? 'GPU' : 'Verification'
+  } Request`;
+
+  const description = planTitle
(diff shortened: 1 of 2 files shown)


DevDW (approved):
LGTM, works well from my testing 👌

johnwcallahan (commented):
Two things:

- We should check in with someone from Support on what this exact messaging should be (if you haven't already).
- The way we're intercepting errors and replacing with React components is definitely not ideal, but I really don't love sending `type_display_title` with the payload to the API. Have you  thought about doing the id => label conversion in the error interceptor? We could use the cached types data, or potentially a function that access the store directly. (this is dependent on my first point about getting the exact messaging, though).

packages/manager/src/request.tsx
```diff
@@ -42,23 +42,39 @@ export const handleError = (error: AxiosError) => {
...
+  const requestedLinode = {
+    title: `${
+      requestData?.type?.match(/gpu/i) ? 'GPU' : 'Verification'
+    } Request for a ${displayTitle} Plan`,
+    description:
+      'Hello. I am trying to deploy a ' +
+      displayTitle +
+      ' Plan, but I am getting a messaging asking me for additional verification.' +
+      '\n\nMy use case is ___.\n\n' +
+      'Please let me know if there is any other information I can provide. Thank you.',
```
HanaXu, line 69: Ended up using string concatenation bc of formatting issues with template literals

johnwcallahan (changes requested):
We should make sure the plan title actually exists before constructing the message and make sure a better fallback is in place. If you try to receive a service transfer with a low reputation, you also get the "verification request" error, resulting in:

Screen Shot 2021-06-14 at 12 07 05 PM

I think that's why "Verification Request" was the old fallback, which I think would be good to keep for cases like this.

johnwcallahan (approved):

PR #7677, opened 2021-05-25
M3-5061: Create Linode UI Refinements

tiffwong:
## Description

- Add divider between "Root Password" and "SSH Keys"
- Move VLAN into it's own panel
- Rename the add-ons panel and add divider between "Backups" and "Private IP"

## How to test

Go to http://localhost:3000/linodes/create


bnussman (changes requested):
If the feature flag for VLANs or a customer isn't permitted use VLANs for whatever reason, we get a blank box on the Create Linode page.  
Screen Shot 2021-05-26 at 10 05 12 AM

bnussman (approved):

HanaXu:
We should add margin between the SSH Keys and Add-ons if the customer doesn't have access to VLANs

image

HanaXu (commented):
We should add margin between the SSH Keys and Add-ons if the customer doesn't have access to VLANs

image

HanaXu (approved):

PR #7678, opened 2021-05-26
Remove checkout bar from Linode Create

johnwcallahan:
## Description

This removes the checkout bar from the Linode Create bar.

It does not remove the checkout bar from anything else (Volumes, NodeBalancers, Kubernetes).

## How to test

Check all pathways of Linode Creation.


tiffwong (approved):
Was able to successfully create a Linode via all pathways 👍

ndukatz (approved):
Looks good and passes tests with updates that I pushed to this branch. Also tested manually and looks good there too.

PR #7683, opened 2021-06-01
M3-5162: changed warning message to include new language

bnussman:
## Description

Changes the language in the warning recommending users use at least 3 nodes for their LKE cluster. 

## How to test

Prepare to buy a Kubernetes Cluster with less than 3 nodes and ensure the warning says `We recommend a minimum of 3 node in each NodePool to avoid downtime during upgrades and maintenance.`

### DIFF
--- a/packages/manager/src/features/Kubernetes/kubeUtils.ts
+++ b/packages/manager/src/features/Kubernetes/kubeUtils.ts
@@ -6,7 +6,7 @@ import { LinodeType } from '@linode/api-v4/lib/linodes';
 import { pluralize } from 'src/utilities/pluralize';
 import { ExtendedCluster, ExtendedPoolNode, PoolNodeWithPrice } from './types';
 
-export const nodeWarning = `We recommend at least 3 nodes in each pool. Fewer nodes may affect availability.`;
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/Kubernetes/kubeUtils.ts
```diff
@@ -6,7 +6,7 @@ import { LinodeType } from '@linode/api-v4/lib/linodes';
 import { pluralize } from 'src/utilities/pluralize';
 import { ExtendedCluster, ExtendedPoolNode, PoolNodeWithPrice } from './types';
 
-export const nodeWarning = `We recommend at least 3 nodes in each pool. Fewer nodes may affect availability.`;
+export const nodeWarning = `We recommend a minimum of 3 nodes in each NodePool to avoid downtime during upgrades and maintenance.`;
```
tiffwong, line 9: In the header and docs, "Node Pools" is spelled with a space so we should make it consistent here

tiffwong (approved):
Left one comment but otherwise it looks good!

HanaXu (approved):

DevDW (approved):
🚢

PR #7685, opened 2021-06-01
M3-5150: updated event text for domain record creation and update

bnussman:
## Description

Updates the logic for displaying an event when the API sends us an optional message. This change allows us to use the API's message in our own message we generate. 

Domain record creation and update event messages now use our own message combined with what the API gives as a message. 

## How to test

View your events in the drawer or on the Events landing page and verify that the message for domain record creation and updates are more descriptive and grammatically correct.

### DIFF
--- a/packages/manager/src/eventMessageGenerator.ts
+++ b/packages/manager/src/eventMessageGenerator.ts
@@ -156,12 +156,10 @@ export const eventMessageCreators: { [index: string]: CreatorsForStatus } = {
     notification: (e) => `Domain ${e.entity!.label} has been deleted.`,
   },
   domain_record_create: {
-    notification: (e) =>
-      `A domain record has been created for ${e.entity!.label}`,
+    notification: (e) => `${e.message} added to ${e.entity!.label}`,
   },
   domain_record_update: {
-    notification: (e) =>
-      `A domain record has been updated for ${e.entity!.label}`,
+    notification: (e) => `${e.message} updated for ${e.entity!.label}`,
   },
   domain_record_delete: {
     notification: (e) =>
@@ -692,12 +690,6 @@ export default (e: Event): string => {
     eventMessageCreators
   );
 
-  if (e.message) {
-    // If the API has specified a message for this event, rely on that instead of
-    // our custom logic.
-    return formatEventWithAPIMessage(e);
-  }
-
   /** we couldn't find the event in our list above */
   if (!fn) {
     /** log unknown events to the console */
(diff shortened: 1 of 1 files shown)


packages/manager/src/eventMessageGenerator.ts
```diff
@@ -692,12 +690,6 @@ export default (e: Event): string => {
     eventMessageCreators
   );
 
-  if (e.message) {
-    // If the API has specified a message for this event, rely on that instead of
-    // our custom logic.
-    return formatEventWithAPIMessage(e);
-  }
-
```
DevDW, line 700: I need to do some more testing to see if this is a valid concern, but it seems like moving this would be a wide-reaching change for events messages (i.e., the way we show other event notifications might be inadvertently affected by this).

Perhaps in the `if` condition we should also be checking to see if `e.action` is not included in a list of actions (which `domain_record_create` and `domain_record_update` would be in) for which we do not want to use the formatted API message 🤔

bnussman: I agree, this is a a big change for event messages. Might be misunderstanding, but  e.action `domain_record_create` and `domain_record_update` are included in the list since we want to take the API's message and append to it to make it grammatically correct. 

For events that come where `e.action` is not included in our list, we fallback to just the just API's message or the action and label if no message is provided.

tiffwong (approved):
Outside of Dajahi's comment, everything else looks good to me!

Unrelated to the work you did, is there a reason why we're using ticks instead of quotes?

packages/manager/src/eventMessageGenerator.ts
DevDW: OK, I think this works–I didn't see any instances of events messages being adversely impacted 👍

DevDW (approved):

bnussman:
> Outside of Dajahi's comment, everything else looks good to me!
> 
> Unrelated to the work you did, is there a reason why we're using ticks instead of quotes?

Not sure exactly what you mean. Maybe give me an example?

tiffwong:
> > Unrelated to the work you did, is there a reason why we're using ticks instead of quotes?
> 
> @tiffwong Not sure exactly what you mean. Maybe give me an example?

@bnussman I'm referring to the ticks surrounding the domain name and IP address
Screen Shot 2021-06-03 at 4 39 47 PM

bnussman:
@tiffwong I noticed that too. I guess the API decided that was best. We could replace it with `'`'s if we think that would look better...

tiffwong:
@bnussman if there's no reason for those, I say change them to quotes but maybe double check with API first

PR #6653, opened 2020-07-16
M3-4334 API Pagination for Domains

Jskobos:
## Description

Add some convoluted handling for API pagination when using EntityTable. The idea, still being refined, is something like this:

1. You arrive at a landing page
2. A thunk request is made for whatever the entity is (Domains for now)
3. If the number of Domains is within the API's request limit (500 atm), we're fine. Use the returned data, paginate it client-side, proceed as normal.
4. If the number is higher than one request, hand off to Pagey. In practice, this is enabled by passing a JS client method to the EntityTable `row` object.
5. Pagey will make its normal request, starting with one for the minimum page size. This means the first, Thunked request is thrown away (technically stored in Redux but not used elsewhere). It may be possible to do some clever workarounds to save the results of the first request, but for now this is already an improvement for very large accounts; 2 requests to a viewable page as opposed to dozens or hundreds.

## Note: 

- steps 3-5 above aren't fully implemented yet, since the store-account-size-in-preferences work is still pending. For now, Domains will always take the Pagey path; it makes a getAllDomains request, then makes a getDomains(page: 1, page_size: 25 or whatever) that actually populates the table.

- Use an account with an enormous amount of Domains to test this PR. Reach out if you need help mocking or locating an appropriate test account.

- The group by tag toggle is still displayed, but it has no effect (at all now on Domains, eventually just for large accounts). Grouping by tag was the original reason we stopped using API pagination, and there's still no way around it.

- Unfortunately everything has been affected at this point, please test:
   - group by tag
   - Entity table list view (Firewalls has this)
   - Entity table with API pagination (Domains)

And then all of the above with the cmr flag off.

PR #7687, opened 2021-06-02
M3-5208: Fix Domains table header for large accounts

tiffwong:
## Description

Tables using the `APIPaginatedTable` component did not have the right styling and are missing the last column in the header since the props weren't being passed down.

## How to test

Login to `prod-test-001` and go to `/domains`

### DIFF
--- a/packages/manager/src/components/EntityTable/EntityTableHeader.tsx
+++ b/packages/manager/src/components/EntityTable/EntityTableHeader.tsx
@@ -190,6 +190,8 @@ export const GroupByTagToggle: React.FC<GroupByTagToggleProps> = React.memo(
             onClick={toggleGroupByTag}
             disableRipple
             className={classes.toggleButton}
+            // Group by Tag is only available when you have less than 500 Domains
+            // See https://github.com/linode/manager/pull/6653 for more details
             disabled={isLargeAccount}
           >
             <GroupByTag />
--- a/packages/manager/src/components/EntityTable/EntityTableHeader.tsx
+++ b/packages/manager/src/components/EntityTable/EntityTableHeader.tsx
@@ -14,25 +14,14 @@ import { HeaderCell } from './types';
 const useStyles = makeStyles((theme: Theme) => ({
   hiddenHeaderCell: theme.visually.hidden,
   sortCell: {
-    fontFamily: theme.font.bold,
-    padding: '10px 15px',
     '& .MuiTableSortLabel-active:hover': {
       color: theme.palette.primary.main,
     },
   },
-  thead: {
-    '& p': {
-      fontFamily: theme.font.bold,
-      fontWeight: 500,
-    },
-  },
-  '& .MuiTableCell-head': {
-    borderBottom: 0,
-  },
   groupByTagCell: {
-    textAlign: 'right',
     backgroundColor: theme.cmrBGColors.bgTableHeader,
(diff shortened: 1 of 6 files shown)


HanaXu (commented):
Not sure how big of an effort it would be, but can we also standardize the domains header styling to match the other headers?

image

image

image

image

image

DevDW (commented):
In addition to the styling tweaks for dark mode, the grouping by tag functionality doesn't seem to be working here

tiffwong:
@HanaXu the table should match now, lmk if I missed anything!

@DevDW we discussed this but for posterity's sake, turns out Group by Tag is disabled for large accounts. See [this PR](https://github.com/linode/manager/pull/6653#pullrequestreview-451921762) for more details on API pagination

HanaXu (commented):
Looking a lot better! Just have some comments

The Domains Last Modified Header is looking a bit wide and I don't think the date format will be longer than it is currently?
image

And for tables in general, the table header hover states are kind of inconsistent. (you don't have to address now, just something to note).

In Linodes and StackScripts, the whole header is clickable and the hover state is a dark blue:

image

In Volumes, NodeBalancers, Firewalls, Images and Domains, you need to hover over the actual word to be able to filter.

In Kubernetes and Object Storage, the whole header is clickable but the hover state is a different shade of blue:

image

I do think we should have the whole header clickable instead of just the text tho. We should also figure out which blue hover state we want to stick with.

packages/manager/src/components/EntityTable/EntityTableHeader.tsx
```diff
@@ -205,6 +196,7 @@ export const GroupByTagToggle: React.FC<GroupByTagToggleProps> = React.memo(
             onClick={toggleGroupByTag}
             disableRipple
             className={classes.toggleButton}
+            disabled={isLargeAccount}
```
DevDW, line 199: might be worth adding a comment explaining why this is the case

DevDW (approved):

packages/manager/src/features/Domains/DomainsLanding.tsx
```diff
@@ -126,67 +75,92 @@ interface Props {
...
-    removeDialogLoading: false,
-    selectedDomainLabel: '',
-    disableDialogOpen: false,
-  };
-
-  static docs: Linode.Doc[] = [Domains];
+const getHeaders = (
+  matchesXsDown: boolean,
+  matchesSmDown: boolean,
+  matchesMdDown: boolean
```
tiffwong, line 88: Ugly but couldn't figure out a better way to do this

packages/manager/src/features/Domains/DomainsLanding.tsx
```diff
@@ -42,79 +39,31 @@ import {
...
-    },
-  });
-
-interface State {
-  importDrawerOpen: boolean;
-  importDrawerSubmitting: false;
-  importDrawerErrors?: APIError[];
-  remote_nameserver: string;
-  createDrawerOpen: boolean;
-  createDrawerMode: 'clone' | 'create';
```
tiffwong, line 107: Removed these props since they're no longer being used

tiffwong:
@HanaXu great catch on the header styles! I'll address those in a follow up pr 😂

HanaXu (approved):
LGTM!

PR #7676, opened 2021-05-24
Add GooglePay logic, make a one time payment

HanaXu:
## Description

- Adds Braintree SDK
- Adds Braintree logic to handle Google Pay payments
- Adds new hook `useScript` and `useLazyScript`
- Adds a Google Pay Button for one-time payments (will go in the `Make a Payment` drawer)
- Adds Google Pay button for adding Google pay as a payment method (commented out for now, will be handled in a follow up PR by @bnussman )
- All Google Pay related changes are behind a feature flag
- Fix existing formatting bug in make a payment drawer where clicking on the arrows in the amount input changes the formatting from 2 decimal to none. (e.g. 5.00 -> 5)

## How to test
Use dev services, In Account -> Make a payment click on the GPay button. The google pay flow modal should open, go through the flow and click Pay (you won't be charged in dev). There should be a success notification at the top of the drawer.

PR #7693, opened 2021-06-04
M3-5126: Update make payment drawer UI

HanaXu:
## Description

This PR only focuses on the UI, logic for clicking on the gpay button will be handled in https://github.com/linode/manager/pull/7676

- Updated make a payment drawer UI (hidden behind the google pay feature flag)
- Added Card and Third Party Payment types
- Fixed minor regression in Storybook for the Payment Method Row Component
- Refactored and cleaned up code

## How to test
1. Ensure the Launch Darkly google pay flag is on in the test env.
2. Go to `Accounts` and click on `Make a Payment`
3. You should be able to see the new UI changes
4. Turn off the feature flag and the UI should default back to the old UI


ndukatz (approved):
Looks good! The space between these two buttons are a little weird to me but that could just be me. 
Screen Shot 2021-06-08 at 4 25 45 PM

packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/CreditCard.tsx
```diff
@@ -0,0 +1,95 @@
...
+        <Typography className={classes.cardInfo} data-qa-contact-cc>
+          {lastFour
+            ? `${type} ****${lastFour}`
+            : 'No payment method has been specified for this account.'}
+        </Typography>
+        <Typography data-qa-contact-cc-exp-date>
+          {isCardExpired ? (
+            <span className={classes.expired}>{`Expired ${expiry}`}</span>
+          ) : (
+            <span>{`Expires ${expiry}`}</span>
```
DevDW, line 87: Screen Shot 2021-06-08 at 3 58 49 PM

but when I click the "Make a Payment" button:

Screen Shot 2021-06-08 at 3 59 04 PM

It appears that we're currently not properly handling in full the case where a user has no payment methods on file (even in prod). Seems to be tied to this:

https://github.com/linode/manager/blob/fe65c798487ed77763840009667fcf53bfab6add/packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/PaymentDrawer.tsx#L273

bnussman (approved):
I think well end up cleaning this up and updating types as Google Pay comes together, but looks good!

packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/CreditCardPayment.tsx
```diff
@@ -97,63 +114,99 @@ export const CreditCard: React.FC<Props> = (props) => {
...
-            <Grid container direction="row" wrap="nowrap" alignItems="center">
-              <Grid item className={classes.cardSection}>
-                <Typography className={classes.cardText}>
-                  Card ending in {lastFour}
-                </Typography>
-                {Boolean(expiry) && (
+        {showGooglePay ? (
+          <>
+            <Grid item className={classes.cardSectionNew}>
+              <CreditCard type={'Visa'} expiry={expiry} lastFour={lastFour} />
```
DevDW, line 128: `type={type}`

DevDW (approved):
LGTM after one more change below

packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/PaymentDrawer.test.tsx
```diff
@@ -65,23 +65,23 @@ describe('Make a Payment Panel', () => {
...
-      userEvent.click(screen.getByText(/pay now/i));
+      userEvent.click(screen.getByText(/pay via credit card/i));
       userEvent.click(screen.getByTestId('credit-card-submit'));
       expect(
         await screen.findByText(/your payment has been processed but/i)
       ).toBeInTheDocument();
     });
 
-    it('should not display a warning for a normal successful payment', async () => {
+    it.skip('should not display a warning for a normal successful payment', async () => {
```
HanaXu, line 77: Skipping these tests for now.

After setting `lastFour` to default to `''`, the UI is fixed, but `PaymentDrawer.test.tsx` ends up failing bc there is no pay button (which I updated to be hidden if there is no credit card on file, which is determined by if there is a lastFour).

The value `lastFour: '9999'` defined in the test doesn’t seem to be actually used. Instead, PaymentDrawer seems to be accessing the store to get the value, so the props being passed into the component are ignored.

tiffwong (changes requested):
Seems like the credit card and PayPal payment buttons are disabled for amounts under $5 but the GPay one isn't. Also the cursor on hover should also reflect a disabled state

The white dividers also stand out a little too much on dark mode imo

Screen Shot 2021-06-16 at 2 02 06 PM

HanaXu:
@tiffwong Good catch, the button states are handled in https://github.com/linode/manager/pull/7676, didn't want to rewrite the same code since it's all going to get merged.

tiffwong (approved):

PR #7695, opened 2021-06-07
M3-5027: Dedicated Maintenance Table with Routing

bnussman:
## Description

- Adds dedicated page at `/account/maintenance` to view maintenance events
- Adds table that is API paginated and sorted for maintenance events 
- Enables the mock service worker to mock maintenance events to test this api paginated and sorted table. (Might refactor or remove this ugly logic before merge)
- Adds Link to Maintenance Page in User dropdown in the navigation bar

Screen Shot 2021-06-28 at 10 12 36 AM

## How to test

Visit `/account/maintenance`


DevDW (commented):
The routing is working well 👍 left several comments

packages/manager/src/features/Account/maintenance/MaintenanceTable.tsx
```diff
@@ -0,0 +1,155 @@
...
+import { AccountMaintenance } from '@linode/api-v4/lib/account/types';
+import capitalize from 'src/utilities/capitalize';
+import usePagination from 'src/hooks/usePagination';
+import TableSortCell from 'src/components/TableSortCell/TableSortCell_CMR';
+import sync from 'css-animation-sync';
+import TableRowEmptyState from 'src/components/TableRowEmptyState';
+
+interface Props {
+  // we will add more types when the endpoint supports then
+  type: 'Linode';
```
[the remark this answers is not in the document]
bnussman: will reassess later but keeping `Linode` because the API returns type starting with a capital letter

bnussman:
Waiting on API for sorting...

packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx
```diff
@@ -0,0 +1,235 @@
...
+import TableRowError from 'src/components/TableRowError';
+import StatusIcon from 'src/components/StatusIcon';
+import { AccountMaintenance } from '@linode/api-v4/lib/account/types';
+import capitalize from 'src/utilities/capitalize';
+import usePagination from 'src/hooks/usePagination';
+import TableSortCell from 'src/components/TableSortCell/TableSortCell_CMR';
+import sync from 'css-animation-sync';
+import TableRowEmptyState from 'src/components/TableRowEmptyState';
+import Link from 'src/components/Link';
+import { CSVLink } from 'react-csv';
```
johnwcallahan, line 19: This should be imported from DownloadCSV instead, which does some cleanup to prevent CSV injection.

packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx
```diff
@@ -0,0 +1,235 @@
...
+  };
+
+  const handleOrderChange = (key: string, order: 'asc' | 'desc') => {
+    setOrderBy(key);
+    setOrder(order);
+  };
+
+  const downloadCSV = async () => {
+    await fetch();
+    // @ts-expect-error everything's fine
```
johnwcallahan, line 156: I tried to find an alternative solution for this, didn't come up with much. You could pass an `any` type param to `useRef` but I don't think that's much better.

packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx
```diff
@@ -0,0 +1,235 @@
...
+        <Grid item className={classes.CSVlinkContainer}>
+          <CSVLink
+            ref={csvRef}
+            headers={headers}
+            filename={`maintenance-${formatDate(DateTime.local().toISO())}.csv`}
+            data={csv || []}
+          />
+          <a
+            className={`${classes.CSVlink} ${classes.CSVlink}`}
+            onClick={downloadCSV}
```
johnwcallahan, line 224: I'd add a comment block somewhere in this component explaining what happens here, and why (re: async data loading).

packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx
```diff
@@ -0,0 +1,235 @@
...
+  const renderTableRow = (item: AccountMaintenance) => {
+    return (
+      <TableRow key={item.entity.id}>
+        <TableCell>
+          <Link to={`/${item.entity.type}s/${item.entity.id}`} tabIndex={0}>
+            {item.entity.label}
+          </Link>
+        </TableCell>
+        <TableCell>{item.when}</TableCell>
+        <TableCell>{item.type}</TableCell>
```
johnwcallahan, line 118: I think we should format these types better, e.g. "Cold Migration". We could just add a function that replaces underscores with spaces and uppercases each word.

packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx
```diff
@@ -0,0 +1,235 @@
...
+const headers = [
+  { label: 'label', key: 'entity.label' },
+  { label: 'id', key: 'entity.id' },
+  { label: 'type', key: 'entity.type' },
+  { label: 'url', key: 'entity.url' },
+  { label: 'when', key: 'when' },
+  { label: 'reason', key: 'reason' },
+  { label: 'status', key: 'status' },
+  { label: 'type', key: 'type' },
+];
```
johnwcallahan, line 43: For the CSV data, I don't think all of these fields necessary. I'd advocate for removing:

- ID
- URL

And adding "Entity Type" ("Linodes", "Volumes", etc.)

Also, the columns should be in the same order as they are presented in the UI.

packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx
```diff
@@ -0,0 +1,235 @@
...
+        page={pagination.page}
+        pageSize={pagination.pageSize}
+        eventCategory={`${type} Maintenance Table`}
+      />
+      <Grid container className={classes.CSVwrapper} justify="flex-end">
+        <Grid item className={classes.CSVlinkContainer}>
+          <CSVLink
+            ref={csvRef}
+            headers={headers}
+            filename={`maintenance-${formatDate(DateTime.local().toISO())}.csv`}
```
johnwcallahan, line 219: This format results in a space in the filename. Maybe we don't need the time?

packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx
```diff
@@ -0,0 +1,235 @@
...
+          oneLine
+          numberOfColumns={5}
+          colSpan={5}
+          widths={[15, 15, 15, 15, 40]}
+        />
+      );
+    } else if (error) {
+      return <TableRowError colSpan={5} message={error[0].reason} />;
+    } else if (data?.results == 0) {
+      return <TableRowEmptyState colSpan={5} />;
```
johnwcallahan, line 141: I think we need a better empty state here. Maybe just adding a custom message to the Table Row Empty State.. "No pending maintenance" or something, and remove the CSV link.

johnwcallahan (commented):
I know this is WIP but I wanted to go ahead and leave some thoughts.

packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx
bnussman: Updated message. I will also check in with Matthew at some point on this and the whole PR in general.

packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx
bnussman: I did have this originally but it caused probems with the ref even if we modify DownloadCSV use `..rest` props. Opting to import cleanCSVData and use it with the direct `react-csv` component.

matthewbartholomew:
John and I discussed adding an entity header to the table. It's useful now but will be needed as we add more entity types in the future. It's going to be the same lock up as Firewalls but I did a quick mockup.

bnussman:
@matthewbartholomew (cc @johnwcallahan) got it, thanks! I will go with that for now. When we support entities other than Linodes, I propose we use accordions for each entity. If an entity has no maintenance events we can default it to be collapsed. 

Screen Shot 2021-06-23 at 10 04 54 AM

packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx
```diff
@@ -0,0 +1,265 @@
...
+  capitalize: {
+    textTransform: 'capitalize',
+  },
+  heading: {
+    marginBottom: theme.spacing(),
+  },
+}));
+
+const MaintenanceTable: React.FC<Props> = (props) => {
+  const prefrenceKey = 'account-maintenance';
```
johnwcallahan, line 77: Spelling of "preference".

packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx
```diff
@@ -0,0 +1,265 @@
...
+
+    return null;
+  };
+
+  const downloadCSV = async () => {
+    await getCSVData();
+    csvRef.current.link.click();
+  };
+
+  React.useEffect(() => {
```
johnwcallahan, line 176: Actually, I don't think this is necessary and currently preventing the React Query caching from working (also see note below about adding filters to the cache key). The query will automatically be run again when the params or filter change, and relying on that instead of manually `refetching()` means the underlying caching may be utilized.

packages/manager/src/queries/accountMaintenance.ts
```diff
@@ -1,21 +1,42 @@
...
-    getAllAccountMaintenance,
-    queryPresets.longLived
+    () => getAllAccountMaintenance(params, filter),
+    { ...queryPresets.longLived, enabled }
+  );
+};
+
+export const useAccountMaintenanceQuery = (params: any, filter: any) => {
+  return useQuery<ResourcePage<AccountMaintenance>, APIError[]>(
+    [queryKey, params.page, params.page_size],
```
johnwcallahan, line 35: I believe `filter` should be included here too, and I think everything will be fine simply giving `params` instead of the individual fields, since the object is serialized by RQ.

johnwcallahan (commented):
The adjustments you made look great. 

We should get the mobile version in a better spot before merging. Typically we start dropping columns as the viewport gets smaller (I'd suggest dropping "Reason" and "Type" in that order, but feel free to explore other options.)

Also let's get a few tests written before merging. I'd love to say the basic table states tested as well as the new `useOrder` hook. 

Speaking of `useOrder`... one more thing: could you provide some basic documentation in the form of code comments for the new hook?

ndukatz (approved):
Looks good. I think grey when pending and flashing orange when started works well for now. All tests passed.

DevDW:
going to look at this in more depth tomorrow morning but maybe we should add a very small bit of right padding or something on the text in the "Reason" column to avoid it running right up against the edge:

Screen Shot 2021-06-24 at 6 07 11 PM

packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx
```diff
@@ -0,0 +1,265 @@
...
+  // we will add more types when the endpoint supports them
+  type: 'Linode';
+}
+
+// Headers for CSV download.
+const headers = [
+  { label: 'Entity Label', key: 'entity.label' },
+  { label: 'Entity Type', key: 'entity.type' },
+  { label: 'Entity ID', key: 'entity.id' },
+  { label: 'When', key: 'when' },
```
DevDW, line 40: Are we still renaming this column to `Date` as discussed in prep?

packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx
```diff
@@ -0,0 +1,293 @@
...
+  { label: 'Status', key: 'status' },
+  { label: 'Reason', key: 'reason' },
+];
+
+const useStyles = makeStyles((theme: Theme) => ({
+  CSVlink: {
+    [theme.breakpoints.down('sm')]: {
+      marginRight: theme.spacing(),
+    },
+    color: '#888F91',
```
tiffwong, line 55: You can use `theme.cmrTextColors.tableHeader` instead of the hex code

DevDW (approved):
I think this is good to go after a few wrap-up changes 👌 

Screen Shot 2021-06-28 at 10 33 03 AM

In mobile view, we need to add a bit of left padding on the table title to match others throughout the app

tiffwong (approved):
Looks good! 

It doesn't have to be in this PR but the width of the "Type", "Status" and "When" columns  can be decreased to give the label column some more space when it's < 1280px. Same thing for the smaller viewports as well
Screen Shot 2021-06-28 at 10 53 34 AM

PR #7697, opened 2021-06-07
Cleanup Divider Component

tiffwong:
## Description

This PR moves the basic styles to the component itself and also added props to easily adjust the spacing above and below the divider without having to create classes for the styles every time.


packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworkingIPTransferPanel.tsx
```diff
@@ -1,660 +0,0 @@
-import { assignAddresses } from '@linode/api-v4/lib/networking';
```
tiffwong, line 1: Realized this file wasn't being used anywhere so I deleted it

packages/manager/src/components/core/Divider.tsx
```diff
@@ -5,28 +5,29 @@ import * as React from 'react';
...
   },
 }));
 
 /* eslint-disable-next-line */
 export interface DividerProps extends _DividerProps {}
 
 interface Props extends _DividerProps {
-  type?: 'landingHeader' | 'other';
+  spacingTop?: number;
+  spacingBottom?: number;
```
HanaXu, line 20: Can we rename spacing to margin?

HanaXu (commented):
Left some comments on dark mode!

packages/manager/src/components/core/Divider.tsx
tiffwong: We use `spacingTop` as a custom prop for a couple other components so I'm going to leave it as is so it's consistent

DevDW (approved):
Went through the app and everything I observed looked good 👍

HanaXu (approved):
LGTM! 👍

PR #7703, opened 2021-06-15
M3-5146: image upload success fail

codyfinn:
## Description

This PR handles the failure and success of uploading custom Images. When an Image is successfully uploaded and decompressed by the server a Toast notification is shown to the user. When an Image fails to decompress on the server a persistent Toast notification is shown to the user. In addition to the persistent Toast notification a user will also see that the status of the Image on the Image Landing table will be changed to 'Failed'.

## How to test

Upload a new image that you expect will successfully be decompressed by the server. You should see the Toast message that informs you of the success. Upload a new image that you expect to fail decompression. You should see the persistent Toast notification informing you of the failure and the status of the Image in the table should be "Failed".


bnussman:
When an upload is successful, could we refresh the images table so it shows the updated state? It's confusing when there is a toast saying the upload was successful but the table's state is not updated

codyfinn:
> When an upload is successful, could we refresh the images table so it shows the updated state? It's confusing when there is a toast saying the upload was successful but the table's state is not updated

Good catch. We don't need to refresh the screen. We just need to listen for events that have the finished status in a similar way that we listen to ones that have a failed status. The table will then automatically update like it does for failures.

HanaXu (commented):
Idk if this is happening for anyone else, but it's really inconsistent when I get the error toast after uploading an invalid image. Sometimes, it just won't show up at all. Sometimes, it'll display on another Cloud Manager tab I have open, but not on the tab where I uploaded the image.

When I do get an error toast, in the notification drawer, it says it failed after 23 hours 57 minutes when I know it was only a minute or two.

image

image

If an upload failed, in the table we should change the Size from `pending` to something like `N/A`, and there shouldn't be a `cancel upload` button displayed.

image

What it looks like after clicking on the cancel upload on a failed image. Semi-related, we should display which image we are trying to cancel the upload for.

image

And specify which image was deleted as well
image

The table doesn't seem to be updating even though a toast says the image upload was successful:
image
image

bnussman:
> change the Size from `pending` to something like `N/A`
 
I agree with this, noticed it as well.

DevDW (changes requested):

packages/manager/src/features/ToastNotifications/ToastNotifications.tsx
```diff
@@ -25,15 +25,23 @@ export const toastSuccessAndFailure = curry(
     enqueueSnackbar: WithSnackbarProps['enqueueSnackbar'],
     eventStatus: EventStatus,
     successMessage: string | undefined,
-    failureMessage: string | undefined
+    failureMessage: string | undefined,
+    persistSuccess: boolean,
+    persistFailure: boolean
```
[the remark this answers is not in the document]
codyfinn: I could use enqueNotification directly and not use the premade _toast function. That way everything else would be left alone.

packages/manager/src/features/ToastNotifications/ToastNotifications.tsx
```diff
@@ -81,6 +89,16 @@ class ToastNotifications extends React.PureComponent<WithSnackbarProps, {}> {
               `Image ${secondaryLabel} created successfully.`,
               `Error creating Image ${secondaryLabel}.`
             );
+          case 'image_upload':
+            const isDeletion = event.message === 'Upload cancelled.';
+            return _toast(
+              `Image ${label} has been uploaded.`,
+              isDeletion
+                ? undefined
+                : `There was a problem uploading image ${label}. ${event.message}`,
```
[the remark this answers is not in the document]
johnwcallahan: Also, maybe consider replacing "uploading" with "processing", to avoid this slightly awkward toast stack:
Screen Shot 2021-06-15 at 3 29 23 PM

packages/manager/src/store/events/event.helpers.ts
```diff
@@ -18,7 +18,7 @@ import { EntityEvent, ExtendedEvent } from './event.types';
  * for that period, would be pointless. We treat these events differently
  * when determining whether they are in progress.
  */
-export const LONG_PENDING_EVENTS = ['image_upload'];
+export const LONG_PENDING_EVENTS = [''];
```
johnwcallahan, line 21: This array is used in RenderEvent.tsx... maybe we should update that to reference `isLongPendingEvent` and remove this array altogether.

packages/manager/src/store/events/event.helpers.ts
```diff
@@ -158,14 +158,21 @@ export const addToEvents = (prevArr: Event[], arr: Event[]) =>
...
+  if (percent_complete === null || isLongPendingEvent(event)) {
     return false;
-  } else if (LONG_PENDING_EVENTS.includes(event.action)) {
-    return percent_complete > 0 && percent_complete < 100;
   } else {
-    return percent_complete !== null && percent_complete < 100;
+    return (
+      percent_complete !== null &&
+      percent_complete < 100 &&
+      !isLongPendingEvent(event)
```
johnwcallahan, line 174: I think this is already covered on L168, right?

packages/manager/src/store/selectors/imageEvents.ts
```diff
@@ -14,8 +17,7 @@ export default createSelector<ApplicationState['events'], Event[], Event[]>(
   (events) =>
     events.filter(
       (thisEvent: Event) =>
-        thisEvent.action === 'disk_imagize' &&
-        !!thisEvent.secondary_entity &&
-        isInProgressEvent(thisEvent)
+        isEventInProgressDiskImagize(thisEvent) ||
+        isEventFailedImageUpload(thisEvent)
```
johnwcallahan, line 21: Instead of adding this condition to this events selector, have you thought about using `useEvents` in ImagesLanding instead?

packages/manager/src/features/ToastNotifications/ToastNotifications.tsx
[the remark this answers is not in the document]
DevDW: After thinking about this more, the fix I mentioned doesn't seem as bad–it builds in the option of persisting toasts if we happen to want to do that in the future in other cases 🧐 one can argue that having to throw all those booleans in throughout the file would muddy it up a bit, but I'm not sure if that's entirely valid. Interested in hearing @johnwcallahan's thoughts on this though.

Because this is curried, I don't think that default params are an option here. So I think our choices are what I described in my previous comment, or not using `_toast` and enqueueing independently for the `image_upload` case as Cody said.

packages/manager/src/store/events/event.helpers.ts
```diff
@@ -185,6 +178,18 @@ export const isCompletedEvent = ({
...
 
+export const isEventInProgressDiskImagize = (event: Event): boolean => {
+  return (
+    event.action === 'disk_imagize' &&
+    Boolean(event.secondary_entity) &&
+    isInProgressEvent(event)
+  );
+};
+
+export const isEventFailedImageUpload = (event: Event): boolean => {
```
DevDW, line 189: This variable is only checking for the event's action type and doesn't have a check for whether it failed or not, so maybe rename this variable and add a comment explaining its purpose/use case?

DevDW (approved):
I think all of the major issues have been addressed so this is in a good state to get it merged. Left one suggestion about a variable renaming.

For a follow-up PR I think some refactoring such as what John mentioned about possibly using `useEvents` in `ImagesLanding.tsx` would be good

packages/manager/src/features/ToastNotifications/ToastNotifications.tsx
```diff
@@ -52,7 +60,16 @@ class ToastNotifications extends React.PureComponent<WithSnackbarProps, {}> {
...
+        const _toastWithPersist = toastSuccessAndFailure(
+          enqueueSnackbar,
+          event.status
+        );
+        const _toast = toastSuccessAndFailure(
+          enqueueSnackbar,
+          event.status,
+          false,
+          false
+        );
```
DevDW, line 72: This is a good middle ground between the approaches discussed yesterday IMO 👍

packages/manager/src/store/events/event.helpers.ts
codyfinn: I knew I forgot to change something. It used to only check for failed upload messages but we still need to check for successful ones. I can change that real quick.

johnwcallahan (approved):

PR #7711, opened 2021-06-17
M3-5133: Fix bug that prevented /dev/sdh Block Device from showing up 

DevDW:
## Description
When editing a Linode's default config or adding a new one, clicking the "Add a Device" button should be possible until you reach `/dev/sdh`. Users were being blocked from getting there because the condition disabling the button was set 1 lower than it should've been. 

## How to test
Edit default configs and add new ones from a Linode's "Configurations" tab. Confirm that clicking the "Add a Device" button until it is disabled takes you up to `/dev/sdh`. Save Changes/Add Configuration and confirm that the proper disks/volumes are tied to the Block Devices you assigned them to.

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
@@ -414,6 +414,8 @@ const LinodeConfigDialog: React.FC<CombinedProps> = (props) => {
     volumes: props.volumes,
   };
 
+  const deviceSlots = ['sda', 'sdb', 'sdc', 'sdd', 'sde', 'sdf', 'sdg', 'sdh'];
+
   /**
    * Form change handlers
    * (where formik.handleChange is insufficient)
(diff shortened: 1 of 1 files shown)


packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
```diff
@@ -694,7 +694,7 @@ const LinodeConfigDialog: React.FC<CombinedProps> = (props) => {
                 buttonType="secondary"
                 superCompact
                 onClick={() => setDeviceCounter((counter) => counter + 1)}
-                disabled={readOnly || deviceCounter >= 6}
+                disabled={readOnly || deviceCounter >= 7}
```
johnwcallahan, line 697: To avoid a magic number, you could define the `slots` (['sda', 'sdb', etc.]) elsewhere in the component and reference the length of that array (minus 1) instead of `7`.

johnwcallahan (approved):
This looks good, seems like there are two tangentially related bugs we should fix:

- If you have a device on sda and sdh, you only see sda, sdb, and sdc when you open the edit config modal.
- We don't surface error messages if there's a problem with device assignment.

tiffwong (approved):

DevDW:
> This looks good, seems like there are two tangentially related bugs we should fix:
> 
> * If you have a device on sda and sdh, you only see sda, sdb, and sdc when you open the edit config modal.
> * We don't surface error messages if there's a problem with device assignment.

Going to address these issues in a follow-up PR but will merge this one now

PR #7715, opened 2021-06-21
Fix console warning for Divider custom props

tiffwong:
## Description

Fixes the `unknown-prop` warning that shows up in the console as a result of the custom props (`spacingTop`, `spacingBottom`) that I made for Divider. More information on that [here](https://reactjs.org/warnings/unknown-prop.html).

Passing the remaining properties as `...rest` should prevent the DOM from complaining

## How to test

Check the console to make sure that there's no longer an error


DevDW (approved):
🚢

bnussman (approved):

HanaXu (approved):

PR #7716, opened 2021-06-21
M3-5179: Use React Query for account users endpoint

bnussman:
## Description

- Uses React Query to populate and paginate `/account/users`.
- The pagination logic and the entire `UsersLanding.tsx` was refactored.
 
## How to test

Ensure `/account/users` in Cloud Manager still functions as expected. Try to test cases where the table would need to paginate (more than 25 users) even though this is rare.

### DIFF
--- a/packages/manager/src/features/Users/UsersLanding.tsx
+++ b/packages/manager/src/features/Users/UsersLanding.tsx
@@ -1,11 +1,6 @@
-import { map as mapPromise } from 'bluebird';
-import { deleteUser, getUsers, User } from '@linode/api-v4/lib/account';
-import { APIError } from '@linode/api-v4/lib/types';
-import * as memoize from 'memoizee';
-import { withSnackbar, WithSnackbarProps } from 'notistack';
+import { deleteUser, User } from '@linode/api-v4/lib/account';
+import { useSnackbar } from 'notistack';
 import * as React from 'react';
-import { RouteComponentProps, withRouter } from 'react-router-dom';
-import { compose } from 'recompose';
 import UserIcon from 'src/assets/icons/user.svg';
 import AddNewLink from 'src/components/AddNewLink';
 import {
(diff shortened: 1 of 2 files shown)


packages/manager/src/features/Users/UsersLanding.tsx
```diff
@@ -138,12 +128,8 @@ const UsersLanding: React.FC<CombinedProps> = (props) => {
   const theme = useTheme<Theme>();
   const matchesSmDown = useMediaQuery(theme.breakpoints.down('sm'));
 
-  React.useEffect(() => {
-    request();
-  }, [request]);
-
   const addUser = () => {
-    request();
+    refetch();
```
DevDW, line 132: If you go to add a restricted user, you get redirected to the "User Permissions" tab for that new user without a RQ refetch occurring. The newly-added user doesn't appear in the "Users" table as a result unless a manual page refresh is done. It's because of this piece in `CreateUserDrawer.tsx`:

https://github.com/linode/manager/blob/9861c468ce84ec6fd7fbb61c25063b58efcb6834/packages/manager/src/features/Users/CreateUserDrawer.tsx#L61-L69

if we move `addUser(user)` outside the conditional and just do the `push(...)` if `user.restricted` is true, everything should work out

DevDW (changes requested):
Looks good overall 👍 ; found one bug when adding restricted users.

packages/manager/src/features/Users/UsersLanding.tsx
bnussman: Ah, thanks for catching this! Fixed!

packages/manager/src/features/Users/UsersLanding.tsx
```diff
@@ -102,21 +97,16 @@ interface Props {
...
-    request,
-    onDelete,
-    enqueueSnackbar,
-    data: users,
-    error,
-    loading,
-  } = props;
-
+const UsersLanding: React.FC<Props> = (props) => {
+  const pagination = usePagination(1);
```
DevDW, line 101: Although it's pretty uncommon for accounts to have >25 users, it's arguably worth adding a `preferenceKey` here so that if a user changes their page size, the size they chose will be persisted for this table.

DevDW (approved):
Left one other optional comment, but this is good to go 🚢

packages/manager/src/features/Users/UsersLanding.tsx
bnussman: This is a very good point and something I might make a new hook for. With all of the React Query paginated tables I've been working on, none of them are aware of query params (like `?order=asc&orderBy=label`) and previous order and page size preference like you are saying. I'll be looking into adding this in a clean way

tiffwong (approved):

PR #7717, opened 2021-06-23
M3-5256: Match Image Upload file types more broadly

johnwcallahan:
## Description

**What does this PR do?**
Currently we are determining if an image is the correct file type by making sure the mime-type is "application/x-gzip". This PR makes this check more broad so that "application/gzip" (which is valid) also passes the test.

## How to test

Image uploads should continue to work as before. I was unable to reproduce the original bug report, and am unclear about the mechanics of the mime-type here. Even if a file as a mime-type of `application/gzip` on my computer, when I drag it to the upload area the browser reads it as `application/x-gzip`. **So this is more of a theoretical fix that shouldn't break anything.**

### DIFF
--- a/packages/manager/src/components/FileUploader/FileUploader.tsx
+++ b/packages/manager/src/components/FileUploader/FileUploader.tsx
@@ -185,7 +185,7 @@ const FileUploader: React.FC<CombinedProps> = (props) => {
 
   // This function will be called when the user drops non-.gz files, more than one file at a time, or files that are over the max size.
   const onDropRejected = (files: FileRejection[]) => {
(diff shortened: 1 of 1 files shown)


DevDW (approved):
LGTM -- I was also unable to reproduce the original bug, but functionality appears unchanged (tested in Chrome and Firefox).

It seems that mime-type is handled differently depending on configuration, browser, etc. but x-gzip and gzip are [functional equivalents](https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.5).

bnussman (approved):
Found no issues with accepting a more broad scope of gziped files

PR #7718, opened 2021-06-23
M3-5254: Improve presentation of info & fix error-surfacing bug in LinodeConfigDialog

DevDW:
## Description
- [x] If devices are assigned non-sequentially (e.g. `sda` and `sdb` are assigned, then no others until `sdh`), display up to the last assigned device to the user instead of having the user click "Add a Device" until reaching that slot
- [x] Surface the errors when a user assigns a volume or a disk that is already assigned on a different slot

## How to Test
Try to add or edit a Linode's configuration and assign the same disk or volume for more than one Block Device–the error should be surfaced for each.

If you edit an existing configuration, you should see up to the last assigned slot when the modal opens, i.e. if no slots after `/dev/sdd` are assigned, you should see `/dev/sda`, `/dev/sdb`, `/dev/sdc`, and `/dev/sdd`.


packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
```diff
@@ -401,6 +416,7 @@ const LinodeConfigDialog: React.FC<CombinedProps> = (props) => {
         // Create mode; make sure loading/error states are cleared.
         resetForm({ values: defaultFieldsValues });
         setUseCustomRoot(false);
+        setDeviceCounter(1);
```
DevDW, line 419: This is being set explicitly because if a user went to edit a config first and then went to add a configuration, the modal to add a config would display the same number of device slots as was in the edit modal

tiffwong (approved):
LGTM

bnussman (approved):
Good work 👍🏼

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/DeviceSelection.test.ts
```diff
@@ -0,0 +1,19 @@
...
+  'disk ID 2005 is referenced more than once',
+  'volume ID 3005 is referenced more than once',
+  'disk ID 4005 is referenced more than once',
+  'volume ID 5005 is referenced more than once',
+];
+
+describe('Device selection component', () => {
+  describe('Grabbing the disk_id or volume_id from the error reason', () => {
+    it('should create a string out of the disk_id or volume_id in the format {type}-{id}', () => {
+      expect(extractDiskOrVolumeId(errorReasons[0])).toBe('disk-2005');
```
johnwcallahan, line 13: Super minor. You could choose to make this dynamic:


```ts
const tests = [
  { input: "disk ID 2005 is referenced more than once", expected: 'disk-2005' },
  // ... etc.
]
```

Then you just iterate through the tests:
```ts
  tests.forEach(thisTest => {
    expect(extractDiskOrVolumeId(thisTest.reason).toBe(thisTest.expected)
  }
```

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
```diff
@@ -373,8 +389,18 @@ const LinodeConfigDialog: React.FC<CombinedProps> = (props) => {
        */
       if (config) {
         const devices = createStringsFromDevices(config.devices);
-        const initialCounter = Object.keys(devices).length;
-        setDeviceCounter(initialCounter);
+
+        // If device slots are populated out of sequential order (e.g. sda and sdb are assigned but no others are until sdf), ascertain the last assigned slot to determine how many device slots to display initially.
```
johnwcallahan, line 393: Minor: this is quite a long line... I'd recommend adding a newline or two.

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
johnwcallahan: Might be clearer to set a `deviceCounterDefault` value somewhere and use it here as well as the original `useState` call.

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/DeviceSelection.test.ts
```diff
@@ -0,0 +1,19 @@
...
+      expect(extractDiskOrVolumeId(errorReasons[1])).toBe('volume-3005');
+      expect(extractDiskOrVolumeId(errorReasons[2])).toBe('disk-4005');
+      expect(extractDiskOrVolumeId(errorReasons[3])).toBe('volume-5005');
+    });
```
johnwcallahan, line 17: What happens when an ID isn't included in the error reason? (looks like it returns "volume-NaN"). A test should be added for this case, and a better value should be returned as well (maybe `null`).

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/DeviceSelection.tsx
```diff
@@ -99,3 +117,10 @@ const DeviceSelection: React.FC<CombinedProps> = (props) => {
 };
 
 export default DeviceSelection as React.ComponentType<Props>;
+
+const blockDeviceRegex = /[0-9][0-9]+/g;
```
johnwcallahan, line 121: Why two digit minimum?

johnwcallahan (commented):
In the error message, I'd like to see us substituting the device ID with the device label (e.g. `Debian 10 Disk already in use`).

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/DeviceSelection.tsx
DevDW: I couldn't think of a reason to keep a two digit minimum so I changed this accordingly

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/DeviceSelection.test.ts
```diff
@@ -1,19 +1,39 @@
...
+  { input: 'disk ID 2005 is referenced more than once', expected: 'disk-2005' },
+  {
+    input: 'volume ID 3005 is referenced more than once',
+    expected: 'volume-3005',
+  },
+  { input: 'disk ID 4005 is referenced more than once', expected: 'disk-4005' },
+  {
+    input: 'volume ID 5005 is referenced more than once',
+    expected: 'volume-5005',
+  },
```
johnwcallahan, line 13: What about these test cases are different from the first two? Seems like you could cut them.

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/DeviceSelection.tsx
```diff
@@ -89,7 +89,7 @@ const DeviceSelection: React.FC<CombinedProps> = (props) => {
               label={`/dev/${slot}`}
               errorText={
                 selectedDevice?.value === diskOrVolumeInErrReason
-                  ? errorText
+                  ? adjustedErrorText(errorText!, selectedDevice.label)
```
johnwcallahan, line 92: I don't think we should be using the assertion operator (it's too easy to break it later). Either check that `errorText` is defined or handle for it being undefined in the helper function.

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/DeviceSelection.tsx
```diff
@@ -118,9 +118,24 @@ const DeviceSelection: React.FC<CombinedProps> = (props) => {
...
   const type = errorText.includes('disk') ? 'disk' : 'volume';
-  return `${type}-${Number(errorText.match(blockDeviceRegex)?.[0])}`; // return a string structured in a way that it is easily compared to selectedDevice.value
+
+  const blockDeviceRegexMatch = errorText.match(blockDeviceRegex)?.[0];
+
+  if (!blockDeviceRegexMatch) {
+    return null;
+  }
+
+  return `${type}-${Number(blockDeviceRegexMatch)}`; // return a string structured in a way that it is easily compared to selectedDevice.value
```
johnwcallahan, line 132: More idiomatic would be to return a ternary expression.

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/DeviceSelection.tsx
```diff
@@ -118,9 +118,24 @@ const DeviceSelection: React.FC<CombinedProps> = (props) => {
...
+};
+
+export const adjustedErrorText = (errorText: string, deviceLabel?: string) => {
+  if (!deviceLabel) {
+    return errorText;
+  }
+
+  return `${deviceLabel} is already in use.`;
```
johnwcallahan, line 140: (same comment about ternary expressions)

johnwcallahan (changes requested):
The disk label substitution works well. This seems a bit odd though... what are your thoughts on only displaying the error on the last instance of the field error?
Screen Shot 2021-07-13 at 8 23 19 AM

DevDW:
> The disk label substitution works well. This seems a bit odd though... what are your thoughts on only displaying the error on the last instance of the field error?
> Screen Shot 2021-07-13 at 8 23 19 AM

It seems it might be a bit arbitrary to display it on the last instance only since Block Devices don't have to be assigned sequentially. For example, if `/sdg` has VolumeA on it already and a user tries to assign VolumeA to `/sdb` as well, we would encounter a situation where the error would get placed only on the slot that's already assigned (`/sdg` in this example).

johnwcallahan:
> It seems it might be a bit arbitrary to display it on the last instance only since Block Devices don't have to be assigned sequentially. For example, if `/sdg` has VolumeA on it already and a user tries to assign VolumeA to `/sdb` as well, we would encounter a situation where the error would get placed only on the slot that's already assigned (`/sdg` in this example).

Great point, let's keep it as is.

johnwcallahan (approved):

PR #7719, opened 2021-06-24
Updates Package.json

martinmckenna:
## Description

1. Adds `files` key to validation package (which determines what gets published to NPM)

2. Changes `main` key in `package.json` (which dictates where things are imported from).
    * As of right now `import { blah } from '@linode/api-v4'` by default imports things from `lib/index.js` which is unminified code, unlike `./index.js` which is minified.
    * In other words, importing from `./index.js` is impossible right now
    * Didn't update `types` key because it's fine as is
  
## How to test

1. Run all projects locally
2. Make changes to `api-v4`
3. Ensure manager still hot reloads with changes

-------

1. Also ensure imports from `@linode/api-v4/lib/whatever` still works

-------

1. Run `cd ./packages/validation && npm publish --dry-run` to see what files are getting published to NPM. Comparing this branch to dev, you'll see all the `src` code along with other unnecessary files are getting published under the `validation` package and now it should just be the relevant ones

Before:

Screen Shot 2021-06-24 at 12 36 31 PM

## Possible Issues

When I installed the SDK, I noticed `./index.js` and `./index.d.ts` weren't actually published to NPM. Obviously not sure if that was a cause from a bad publish or it's a genuine issue that needs to be fixed prior to merging this PR

Screen Shot 2021-06-24 at 12 31 16 PM

### DIFF
--- a/packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/UpdateContactInformationForm/UpdateContactInformationForm.tsx
+++ b/packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/UpdateContactInformationForm/UpdateContactInformationForm.tsx
@@ -1,5 +1,4 @@
-import { updateAccountInfo } from '@linode/api-v4';
-import { Account } from '@linode/api-v4/lib/account';
+import { Account, updateAccountInfo } from '@linode/api-v4/lib/account';
 import { APIError } from '@linode/api-v4/lib/types';
 import countryData from 'country-region-data';
 import { defaultTo, lensPath, pick, set } from 'ramda';
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,9 +17,10 @@ jobs:
       - run: yarn --frozen-lockfile
       - run: yarn workspace @linode/validation run build
       - uses: actions/upload-artifact@v2
-        with:
           name: packages-validation-lib
-          path: packages/validation/lib
+          path: |
+            packages/validation/index.js
+            packages/validation/lib
 
   publish-validation:
     runs-on: ubuntu-latest
@@ -31,7 +32,7 @@ jobs:
       - uses: actions/download-artifact@v2
         with:
           name: packages-validation-lib
-          path: packages/validation/lib
+          path: packages/validation
       - uses: JS-DevTools/npm-publish@v1
         id: npm-publish
         with:
@@ -66,7 +67,7 @@ jobs:
       - uses: actions/download-artifact@v2
         with:
           name: packages-validation-lib
-          path: packages/validation/lib
+          path: packages/validation
       - run: yarn workspace @linode/api-v4 run test
 
   build-sdk:
@@ -86,14 +87,17 @@ jobs:
       - uses: actions/download-artifact@v2
         with:
           name: packages-validation-lib
-          path: packages/validation/lib
+          path: packages/validation
       - run: yarn --frozen-lockfile
       - run: yarn workspace @linode/api-v4 run build
       - run: yarn workspace @linode/api-v4 run build:node
       - uses: actions/upload-artifact@v2
         with:
           name: packages-api-v4-lib
-          path: packages/api-v4/lib
+          path: |
+            packages/api-v4/index.js
+            packages/api-v4/index.node.js
+            packages/api-v4/lib
 
   test-manager:
     runs-on: ubuntu-latest
@@ -112,11 +116,11 @@ jobs:
       - uses: actions/download-artifact@v2
         with:
           name: packages-validation-lib
-          path: packages/validation/lib
+          path: packages/validation
       - uses: actions/download-artifact@v2
         with:
           name: packages-api-v4-lib
-          path: packages/api-v4/lib
+          path: packages/api-v4
       - run: yarn --frozen-lockfile
       - run: yarn workspace linode-manager run test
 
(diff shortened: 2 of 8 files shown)


DevDW:
Thanks for the PR @martinmckenna!

> 1. Run `cd ./packages/api-v4 && npm publish --dry-run` to see what files are getting published to NPM. Comparing this branch to dev, you'll see all the `src` code along with other unnecessary files are getting published under the `validation` package and now it should just be the relevant ones

Should this command be `cd ./packages/validation && npm publish --dry-run`? Running that, I see 78 total files on `develop` vs. 51 total files on this branch, but 233 total files on each branch with `cd ./packages/api-v4 && npm publish --dry-run`.

martinmckenna:
@DevDW yep sorry about that you're right I'll update the descrip

DevDW (commented):
> When I installed the SDK, I noticed `./index.js` and `./index.d.ts` weren't actually published to NPM. Obviously not sure if that was a cause from a bad publish or it's a genuine issue that needs to be fixed prior to merging this PR

I think they're being excluded because of

https://github.com/linode/manager/blob/f609a139454adcce3dcef88ce981f95eb2c74d01/.gitignore#L5-L8

If we want/need to make it so that those files get published to NPM for the JS client and /validation package, we probably need to make an [`.npmignore` file](https://npm.github.io/publishing-pkgs-docs/publishing/the-npmignore-file.html) that doesn't include those files 🧐

martinmckenna:
interesting. do you think it'd be worth just creating an empty file and adding it to the PR @DevDW 

from that article:

> If you want to include something that is excluded by your .gitignore file, you can create an empty .npmignore file to override it

DevDW (commented):
> interesting. do you think it'd be worth just creating an empty file and adding it to the PR @DevDW
> 
> from that article:
> 
> > If you want to include something that is excluded by your .gitignore file, you can create an empty .npmignore file to override it

If I understand correctly how it'd work, if there's a `.npmignore` file present, npm would ignore the `.gitignore` file entirely; if `.npmignore` is blank, then npm would end up not excluding any files... so the `.npmignore` file would have to include everything `.gitignore` does, minus the stuff we _don't_ want to be ignored by npm. 🧐 

But, after reading the last section of [this](https://awsm.page/npm/npmignore-vs-gitignore-which-one-to-use/), I think the better approach is to do what you did with the `files` key for /validation for /api-v4 as well. That avoids the overhead of having a new file that would have to be maintained every time there's a change to `.gitignore`.

martinmckenna:
@DevDW updated

tiffwong (commented):
After looking into this, I agree with the whitelisting approach using `package.json` via `files` rather than `.npmignore' since that way would require less work and is less prone to error in the case of someone forgetting to update both files

johnwcallahan (commented):
Thanks for the PR @martinmckenna!

I noticed that importing from `@linode/api-v4` doesn't work well with Node. I'm seeing "window is not defined" errors (which I think is coming from minified Axios code).

To reproduce:

1. Checkout this branch
2. `cd packages/api-v4`
3. `yarn run build`
4. `node index.js`
5. Observe: "ReferenceError: window is not defined".

If I add `target: node` to the webpack config and rebuild, the error does not happen.

I wonder if we'll need to have a bundle config for both the browser and Node.

martinmckenna:
@johnwcallahan you think it might be worth extending the webpack config and run something like `yarn build:node --config webpack.node.config.js` which pumps out an `index.node.js` or something like that and run that on `prePublishOnly`?

johnwcallahan:
> @johnwcallahan you think it might be worth extending the webpack config and run something like `yarn build:node --config webpack.node.config.js` which pumps out an `index.node.js` or something like that and run that on `prePublishOnly`?

@martinmckenna yeah seems like that could work. Then in our package.json we'd point `main` to `index.node.js` and `browser` to the existing `./index.js`?

martinmckenna:
i'll knock this out sometime today hopefully sorry lol

martinmckenna:
@johnwcallahan updated! I was thinking about just creating a base webpack config and extending from that rather than copy/pasting, but webpack doesn't do that out-of-the-box and I just wanted to get this out instead of installing another package.

`cd ./packages/api-v4 && yarn build:node && node ./index.node.js` should do the trick

packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/UpdateContactInformationForm/UpdateContactInformationForm.tsx
```diff
@@ -1,5 +1,4 @@
-import { updateAccountInfo } from '@linode/api-v4';
-import { Account } from '@linode/api-v4/lib/account';
+import { Account, updateAccountInfo } from '@linode/api-v4/lib/account';
```
martinmckenna, line 1: now that `@linode/api-v4` is being imported from `./index.js` by default now, it's important that methods imported from the SDK in the consuming app are imported from the same place the `baseRequest` that has the Axios interceptors applied to it is.

johnwcallahan (approved):
Sorry for the delay! This looks great, thanks @martinmckenna. 

I tried to push an update but was unable to push to your fork. There's a new-ish import for `deletePaymentMethod` in PaymentInformation.tsx that needs updating re: your comment about consistent imports and interceptors.

All else looks good; approved pending that small change.

DevDW:
@martinmckenna I'm going to prioritize taking a look at this on Monday morning; hopefully we can get it into the next release

DevDW (approved):

PR #7721, opened 2021-06-24
M3-5259: Referral page changes

tiffwong:
## Description

Changes to reflect the upcoming referral threshold:
- If a user has not spent $25, display message and hide referral link
- If a user has spent $25 but has not referred anyone yet, show referral link but hide results
- If a user has spent $25 and has referred people, show referral link and results

Created a local variable that currently represent if users are allowed to make referrals

**Remaining tasks:**
- [x] Replace local variable with actual variable from API
- [x] Remove code for testing purposes

## How to test

- Go to `/profile/referrals`
- Make sure each state matches the corresponding designs including when the flag is turned off
- Uncomment the corresponding code to see the different states if your test account doesn't have any referrals


HanaXu (changes requested):
Will look at this more tomorrow but we should display the images on separate rows for mobile

image

tiffwong:
@HanaXu I totally forgot about responsiveness 🤦‍♀️

johnwcallahan (commented):
Will look more tomorrow, but I believe Matthew has another screenshot showing a "Pending referrals" line if `referrals.pending > 0`. 

How are we feeling about dark mode? Contrast is a little much, maybe:
Screen Shot 2021-06-24 at 5 48 12 PM

DevDW (commented):
This is looking good to me so far 👌 

The middle image seems like it might need a closer cut by the person's arm since a little bit of the background runs along it (this is only visible in dark mode) 🤔 

John mentioned the "Pending referrals" line which is right, that's in the screenshots shared on Tuesday

tiffwong:
@johnwcallahan Matthew brought up some other dark mode color changes so I'll address the contrast in a follow up PR since it would be a global change

@DevDW those are apparently the OG image files from the marketing site so leaving those as is for now

packages/manager/src/components/CopyableTextField/CopyableTextField.tsx
```diff
@@ -1,62 +1,45 @@
...
-    copyIcon: {
-      marginRight: theme.spacing(1),
+const useStyles = makeStyles((theme: Theme) => ({
+  input: {
+    backgroundColor: theme.bg.offWhite,
+  },
+  copyIcon: {
+    marginRight: theme.spacing(0.5),
+    '& svg': {
+      height: 14,
```
tiffwong, line 13: Was requested to reduce the height for the copy icon globally so I refactored this component while I was here

packages/manager/src/assets/icons/referrals/step-1.svg
```diff
@@ -0,0 +1 @@
....4 8.4-6.4 22.8 3.7 28.6 4.7-4.7 4-3.6 3.9-8.3-4.1-3-5-8.7-2-12.8.3-.4.6-.7.9-1l15.2-15.2c3.6-3.6 9.3-3.6 12.9 0 8 8-2.1 15-10.1 23 .3 1.6 2 5.4 1.1 11.7l15.4-15.4c17.1-17.1-8.7-42.9-25.7-25.8zm-2.8 22.1c-4.7 4.7-4 3.6-3.9 8.3 4.1 3 5 8.7 2 12.8-.3.4-.6.7-.9 1l-16.4 16.4c-8.5 8.5-21.4-4.4-12.9-12.9 18.7-18.7 7.7-5.9 10.1-23-1.2 1.2 4.8-4.9-16.6 16.6-17 17 8.7 42.9 25.8 25.8l16.4-16.4c8.4-8.3 6.6-22.7-3.6-28.6z" fill="#32363b"/><g fill="#00f279"><path d="m66.9 42.8c-1.2-1.7-.9-4.3-2.4-3.2s-4.3.9-3.2 2.5.9 4.3 2.5 3.2 4.2-1 3.1-2.5z"/><path d="m61.8 21.1c-3.8-2.7-3.6-7.6-5.2-5.3-2.7 3.8-7.6 3.6-5.3 5.2 3.8 2.7 3.6 7.6 5.2 5.3 2.7-3.8 7.6-3.6 5.3-5.2z"/><path d="m90.6 19.9c-1.9-3.8-2-8.1-.4-12 .3-.4.1-1-.3-1.3-.3-.2-.7-.2-1 0-3.8 1.9-8.1 2-12 .4-.5-.2-1.1.2-1.2.7-.1.2-.1.4 0 .6 1.9 3.8 2 8.2.4 12-.2.5 0 1 .5 1.2.3.1.5.1.8 0 3.7-1.9 8.1-2 12-.4.5.2 1.1-.1 1.3-.7 0-.1 0-.3-.1-.5z"/><path d="m244.9 121c1.9-.1 3.7-1.9 3.8-.1s1.9 3.7.1 3.8-3.7 1.9-3.8.1-1.9-3.6-.1-3.8zm3 22.2c3.5 2.7 3 7.4 4.7 5.3 2.8-3.5 7.4-3.1 5.3-4.7-3.5-2.7-3-7.4-4.7-5.3-2.8 3.5-7.4 3.1-5.3 4.7zm-27.4-.3c3.5 8.1-2.8 14.3.9 12.7 8.1-3.5 14.3 2.8 12.7-.9-3.5-8.1 2.8-14.3-.9-12.7-8.2 3.5-14.3-2.8-12.7.9z"/></g></g></svg>
```
HanaXu, line 1: These are more-so images than icons 🤔 . We should move them to `/src/assets/referrals/step-1.svg` etc instead

packages/manager/src/components/CopyableTextField/CopyableTextField.tsx
HanaXu: The copy icon isn't exactly centered

image
image

packages/manager/src/features/Profile/Referrals/Referrals.tsx
```diff
@@ -69,47 +149,151 @@ class Referrals extends React.Component<CombinedProps, {}> {
...
-    );
-  }
-}
-
-const styled = withStyles(styles);
+              ) : null}
+            </Grid>
+            {flags.referral ? (
+              <>
+                <Grid item>
```
HanaXu, line 220: We should hide this for state 2

image

packages/manager/src/features/Profile/Referrals/Referrals.tsx
```diff
@@ -69,47 +149,151 @@ class Referrals extends React.Component<CombinedProps, {}> {
...
+                        <Grid item>Completed referrals</Grid>
+                        <Grid item>{completed}</Grid>
+                      </Grid>
+                      <Grid
+                        container
+                        justify="space-between"
+                        className={classes.earned}
+                      >
+                        <Grid item>Credit earned</Grid>
+                        <Grid item>${credit}</Grid>
```
HanaXu, line 247: The credit amount should be green to be consistent with how we currently display credits

image

HanaXu (changes requested):
Left some feedback on consistency!

packages/manager/src/features/Profile/Referrals/Referrals.tsx
tiffwong: Are you referring to the images?

HanaXu: The empty Grid item which is causing extra padding between the referral link and image

packages/manager/src/features/Profile/Referrals/Referrals.tsx
```diff
@@ -69,47 +151,151 @@ class Referrals extends React.Component<CombinedProps, {}> {
...
+              <>
+                {allowReferral && total !== undefined && total > 0 ? (
+                  <div className={classes.resultsWrapper}>
+                    {pending != undefined && pending > 0 ? (
+                      <Grid
+                        container
+                        justify="space-between"
+                        className={classes.referrals}
+                      >
+                        <Grid item>Pending referrals</Grid>
```
HanaXu, line 230: Maybe add a colon after each of these lines

packages/manager/src/features/Profile/Referrals/Referrals.tsx
```diff
@@ -2,64 +2,146 @@ import { path } from 'ramda';
...
+  },
+  limitNotice: {
+    marginLeft: theme.spacing(),
+    '& p': {
+      color: `${theme.cmrTextColors.tableStatic} !important`,
+      fontSize: '0.875rem',
+    },
+  },
+  images: {
+    marginTop: theme.spacing(5),
```
HanaXu, line 66: 5 is a bit much, what about 4?

HanaXu (approved):
Left two more minor comments but LGTM :money_with_wings: !

packages/manager/src/features/Profile/Referrals/Referrals.tsx
```diff
@@ -69,47 +151,151 @@ class Referrals extends React.Component<CombinedProps, {}> {
...
-}
-
-const styled = withStyles(styles);
+              ) : null}
+            </Grid>
+            {flags.referral ? (
+              <>
+                {allowReferral && total !== undefined && total > 0 ? (
+                  <div className={classes.resultsWrapper}>
+                    {pending != undefined && pending > 0 ? (
```
DevDW, line 224: `pending !== undefined...`

DevDW (approved):
LGTM 👍

ndukatz (approved):

PR #7723, opened 2021-06-28
M3-5269: Fix Object Storage landing table UI bugs

DevDW:
## Description
Fix UI bugs (row border misalignments and column header mismatching) in Object Storage landing table at smaller viewports

### DIFF
--- a/packages/manager/src/features/ObjectStorage/BucketLanding/BucketTableRow.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketLanding/BucketTableRow.tsx
@@ -33,6 +33,15 @@ const useStyles = makeStyles((theme: Theme) => ({
       textDecoration: 'underline',
     },
   },
+  bucketRegion: {
+    width: '20%',
+  },
+  bucketSize: {
+    width: '15%',
+  },
+  bucketObjects: {
+    width: '10%',
+  },
 }));
 
 interface BucketTableRowProps extends ObjectStorageBucket {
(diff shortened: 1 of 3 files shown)


tiffwong (approved):

packages/manager/src/features/ObjectStorage/BucketLanding/BucketTableRow.tsx
```diff
@@ -86,14 +81,18 @@ export const BucketTableRow: React.FC<CombinedProps> = (props) => {
...
-      <TableCell parentColumn="Region">
-        <Typography variant="body1" data-qa-region>
-          {formatObjectStorageCluster(cluster) || cluster}
-        </Typography>
-      </TableCell>
-      <TableCell parentColumn="Created">
-        <DateTimeDisplay value={created} data-qa-created />
-      </TableCell>
+      <Hidden xsDown>
+        <TableCell parentColumn="Region">
```
tiffwong, line 85: Non-blocking but I don't think we need the `parentColumn` prop anymore

bnussman:
My thoughts on this were to make some changes in the process of fixing this:

| Column  | Hide   |
|---------|--------|
| Objects | xsDown |
| Region  | xsDown |
| Created | smDown |

DevDW:
> My thoughts on this were to make some changes in the process of fixing this. It makes dropping the columns a bit more progressive as you shrink the viewport
> 
> Column	Hide
> Objects	xsDown
> Region	xsDown
> Created	smDown

Pushed some changes, I think it works a bit better if `Created` gets hidden at `mdDown` but let me know if you agree/disagree

bnussman (approved):

PR #7731, opened 2021-06-30
M3-5263: Integrate Algolia Search Results in the Cloud Manager search bar 

DevDW:
## Description
Verify that Algolia search results can be integrated into the top menu search bar (using both client-side and API search) and print results to console.

## How to test
With your console open, type a fragment of one of the entity labels on your account into the top menu search bar. For example, if you have a series of Linodes with the word "test" in them, type in "test." If there are 20 or fewer results for that search term, results should be logged to the console including any matches from the Algolia search. You can tell which ones are Algolia because they will have a nested `source` property equal to either "Linode Documentation" or "Linode Community Site".


johnwcallahan (commented):
The basic approach here looks great, nice work. Let's leave this open for reference until we are clear to work on the implementation.

johnwcallahan:
Closing this out since we are not moving forward with this feature at this time.

PR #7733, opened 2021-07-01
E2e create linode test update

ndukatz:
## Description
This change was needed because of some recent ui changes to this page

**What does this PR do?**
small create-linode e2e test update to catch the case of not updating to running without reloading.
Also a cypress eslint update and a small update for cleanup

## How to test
- `yarn up` in one terminal
- `yarn cy:e2e -s ./cypress/integration/linodes/create-linode.spec.ts` in another

```
REACT_APP_LOGIN_ROOT
REACT_APP_API_ROOT
REACT_APP_APP_ROOT
REACT_APP_LAUNCH_DARKLY_ID
REACT_APP_CLIENT_ID
MANAGER_OAUTH
```

### DIFF
--- a/packages/manager/cypress/integration/linodes/create-linode.spec.ts
+++ b/packages/manager/cypress/integration/linodes/create-linode.spec.ts
@@ -4,22 +4,16 @@ import {
   deleteAllTestLinodes,
 } from '../../support/api/linodes';
 import { assertToast } from '../../support/ui/events';
-import {
-  containsClick,
-  containsVisible,
-  getClick,
-} from '../../support/helpers';
+import { containsClick, fbtVisible, getClick } from '../../support/helpers';
 import { selectRegionString } from '../../support/ui/constants';
 
 describe('create linode', () => {
-  beforeEach(() => {
-    cy.visitWithLogin('/linodes/create');
-    cy.get('[data-qa-deploy-linode]');
-  });
   it('creates a nanode', () => {
     const rootpass = strings.randomPass();
     const linodeLabel = makeLinodeLabel();
     // intercept request
+    cy.visitWithLogin('/linodes/create');
+    cy.get('[data-qa-deploy-linode]');
     cy.intercept('POST', '*/linode/instances').as('linodeCreated');
     cy.get('[data-qa-header="Create"]').should('have.text', 'Create');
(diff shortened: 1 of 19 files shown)


packages/manager/cypress/integration/linodes/create-linode.spec.ts
```diff
@@ -29,7 +23,8 @@ describe('create linode', () => {
     getClick('[data-qa-deploy-linode]');
     cy.wait('@linodeCreated').its('response.statusCode').should('eq', 200);
     assertToast(`Your Linode ${linodeLabel} is being created.`);
-    containsVisible('PROVISIONING');
```
patthiel, line 32: I think it's ok to keep this assertion as well, since it _should_ end up in this state at first everytime

packages/manager/cypress/support/index.js
```diff
@@ -15,31 +15,18 @@
...
-  deleteAllTestNodeBalancers();
-  deleteAllTestVolumes();
-  deleteAllTestImages();
-  deleteAllTestClients();
-  deleteAllTestAccessKeys();
-  deleteAllTestBuckets();
-  deleteAllTestFirewalls();
-  deleteAllTestStackscripts();
+
+before(() => {
```
patthiel, line 26: is it too expensive to run these in beforeEach / afterEach? Are there tests that rely on carried over state? Asking because, if we change this to beforeEach / afterEach, we can basically remove all the cleanup methods from being called within tests (at the beginning or end)

ndukatz: Most of the cleanup within tests occurs at the end - and even now we could probably remove those. it does add some time to do it before each on every test for sure. The before will actually run at the beginning of the run for each container so I am primarily trying to optimize for ci runs. I'm not necessarily set on this long term. I did want to see if it would work well and save time on ci runs though first.

patthiel: do these calls happen synchronously or async? 

If they're async, it won't add much time. Although i do remember having to build a delay between removing Linodes and then removing volumes (because, if you remove a linode that has a volume attached, it takes up to like 30 seconds for the API to display the Volume as not being attached to a linode. Otherwise it fails to remove it saying "you can't delete volumes attached to linodes").

tiffwong (approved):

PR #7734, opened 2021-07-01
Moved Toast / Snackbar to the top right

bnussman:
## Description

Moves Toast (aka Snackbar) to the top right of the screen for all toast notifications. 

Screen Shot 2021-07-01 at 2 40 51 PM


DevDW (commented):
My vote would be for keeping the snackbar in the bottom right; one drawback of moving it to the top right is that it obscures several top menu icons

bnussman:
> My vote would be for keeping the snackbar in the bottom right; one drawback of moving it to the top right is that it obscures several top menu icons

I think we are in agreement bottom right is our best option

PR #7735, opened 2021-07-01
M3 5205 retry image upload

codyfinn:
## Description

Allows users to retry a failed Image upload. If during the processing of an image on the server side an error is encountered then the Image in the ImagesLanding table will change it's status to 'failed' and two options will be available to the user via the ImageActionMenu on the right side of the row. First will be Remove which will remove the Image entry from the table. The Image will already have been removed on the server side since the Image processing failed. The other option will be to Retry the Image upload. This will bring the user back to the Image upload form where the Label and Description will be pre-filled based on the failed Image they are retrying. 

## How to test

Attempt to upload an image known to cause an error during processing. Wait for the processing to fail. Check that the image has two actions that can be taken upon it. Remove and Retry should both work as described above.

### DIFF
--- a/packages/manager/src/features/Images/ImagesActionMenu.tsx
+++ b/packages/manager/src/features/Images/ImagesActionMenu.tsx
@@ -36,22 +36,23 @@ export const ImagesActionMenu: React.FC<CombinedProps> = (props) => {
     id,
     label,
     status,
+    event,
     onRestore,
     onDeploy,
     onEdit,
     onDelete,
+    onRetry,
   } = props;
 
   const actions: Action[] = React.useMemo(() => {
-    // @todo remove first half of this conditional when Machine Images is GA
     const isDisabled = status && status !== 'available';
-    return status === 'pending_upload'
+    const isFailed = event?.status === 'failed';
+    return isFailed
       ? [
-          // Cancelling a pending upload is functionally equivalent to deleting it
           {
-            title: 'Cancel Upload',
+            title: 'Retry',
             onClick: () => {
-              onDelete(label, id, status);
+              onRetry(id, label, description);
             },
           },
         ]
@@ -89,7 +90,17 @@ export const ImagesActionMenu: React.FC<CombinedProps> = (props) => {
             },
           },
         ];
-  }, [status, description, id, label, onDelete, onRestore, onDeploy, onEdit]);
+  }, [
+    status,
+    description,
+    id,
+    label,
+    onDelete,
+    onRestore,
+    onDeploy,
+    onEdit,
+    onRetry,
+  ]);
 
   /**
    * Moving all actions to the dropdown menu to prevent visual mismatches
--- a/packages/manager/src/features/Images/ImagesActionMenu.tsx
+++ b/packages/manager/src/features/Images/ImagesActionMenu.tsx
@@ -49,6 +49,7 @@ export const ImagesActionMenu: React.FC<CombinedProps> = (props) => {
 
   const actions: Action[] = React.useMemo(() => {
     const isDisabled = status && status !== 'available';
+    const isAvailable = !isDisabled;
     const isFailed = event?.status === 'failed';
     return isFailed
       ? [
@@ -97,7 +98,7 @@ export const ImagesActionMenu: React.FC<CombinedProps> = (props) => {
             },
           },
           {
-            title: 'Delete',
+            title: isAvailable ? 'Delete' : 'Cancel Upload',
             onClick: () => {
               onDelete(label, id, status);
             },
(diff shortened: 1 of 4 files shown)


DevDW (commented):
Looks good, came across some edge cases:

1. If a soon-to-fail upload is in processing and you edit it and save the changes, it seems that no further updates on it come through once it fails (no failure toast, no event, no update in the table)

2. Since the failure toast is persistent, when you click "Retry" and arrive on the "Upload Image" tab to try again, the toast remains (if you haven't dismissed it)–maybe we can look into clearing it when a user is retrying the upload?

codyfinn:
> Looks good, came across some edge cases:
> 
>     1. If a soon-to-fail upload is in processing and you edit it and save the changes, it seems that no further updates on it come through once it fails (no failure toast, no event, no update in the table)
> 
>     2. Since the failure toast is persistent, when you click "Retry" and arrive on the "Upload Image" tab to try again, the toast remains (if you haven't dismissed it)–maybe we can look into clearing it when a user is retrying the upload?

I disabled the edit action for images that are not available for use. This should prevent the first issue from happening. Looking into auto dismissing the toast persistent toast notification users get when they encounter an error while the image is processing I could not find a good way to make the notification auto close. Some refactoring of how we handle toast notifications would need to be done to allow something like that and I think it is outside the scope of this story.

packages/manager/src/features/Images/ImagesActionMenu.tsx
```diff
@@ -36,28 +36,33 @@ export const ImagesActionMenu: React.FC<CombinedProps> = (props) => {
...
     onEdit,
     onDelete,
+    onRetry,
   } = props;
 
   const actions: Action[] = React.useMemo(() => {
-    // @todo remove first half of this conditional when Machine Images is GA
     const isDisabled = status && status !== 'available';
-    return status === 'pending_upload'
+    const isFailed = event?.status === 'failed';
```
DevDW, line 49: We should probably add `event` to the dependency array here

packages/manager/src/features/Images/ImagesActionMenu.tsx
```diff
@@ -36,28 +36,33 @@ export const ImagesActionMenu: React.FC<CombinedProps> = (props) => {
...
-              onDelete(label, id, status);
+              onRetry(id, label, description);
             },
           },
         ]
       : [
           {
             title: 'Edit',
+            disabled: isDisabled,
+            tooltip: isDisabled
```
DevDW, line 63: Let's add `tooltip={action.tooltip}` where <InlineMenuAction ... /> gets returned, so that the tooltip gets surfaced for when this button is disabled

DevDW (approved):
Disabling the "Edit" action is fair. Left a couple of cleanup items, then this is good to go to me

bnussman (approved):
I know we have talked about these items and they may be out of the scope of this specific PR but...
- ~~Can we autofill the same region when the user clicks **Retry**?~~ Nevermind
- Should we have a cancel button beside the **Retry** button in the case a user does not want to retry? 
- It would be nice to dismiss the error toast when the user retries
- If there is a validation error, the browse files button goes away and does not come back when the error is fixed by the user
Screen Shot 2021-07-07 at 11 15 20 AM

Overall, the retry experience was really good and went smoothly! It would be awesome if we could fix the issue with the browse files button going away before merge.

HanaXu (approved):
I think we should change the `Delete` to `Cancel` in the dropdown to match the Modal, but otherwise LGTM!

Screen Shot 2021-07-07 at 2 52 55 PM
Screen Shot 2021-07-07 at 2 53 18 PM

codyfinn:
> I know we have talked about these items and they may be out of the scope of this specific PR but...
> 
>     * ~Can we autofill the same region when the user clicks **Retry**?~ Nevermind
> 
>     * Should we have a cancel button beside the **Retry** button in the case a user does not want to retry?
> 
>     * It would be nice to dismiss the error toast when the user retries
> 
>     * If there is a validation error, the browse files button goes away and does not come back when the error is fixed by the user
> 
> 
> Screen Shot 2021-07-07 at 11 15 20 AM
> 
> Overall, the retry experience was really good and went smoothly! It would be awesome if we could fix the issue with the browse files button going away before merge.

I added a cancel button on images that are processing. The dismissal of the toast notification is possible but I think it would require work that is outside the scope of this story. Same with the validation error. There is a bug with the underlying component that would require more work than this story covers.

bnussman:
Understood! 👍🏼🚢

codyfinn:
> I think we should change the `Delete` to `Cancel` in the dropdown to match the Modal, but otherwise LGTM!
> 
> Screen Shot 2021-07-07 at 2 52 55 PM
> 
> Screen Shot 2021-07-07 at 2 53 18 PM

Should be addressed in my most recent commit.

PR #7737, opened 2021-07-02
M3-5175: Use React Query for Account

bnussman:
## Description

Uses React Query for fetching and state management for a user's account information.

## How to test

Test literally everything 😳


patthiel:
@ndukatz we should put this PR through it's paces with the cypress tests

packages/manager/src/IdentifyUser.tsx
```diff
@@ -45,10 +35,10 @@ export const IdentifyUser: React.FC<Props> = (props) => {
 
   // Configure user for GTM once we have the info we need.
   React.useEffect(() => {
-    if (euuid) {
-      initGTMUser(euuid);
+    if (account?.euuid) {
+      initGTMUser(account.euuid);
     }
-  }, [euuid]);
+  }, [account]);
```
johnwcallahan, line 41: Would this be made better by assigning `account.euuid` to a variable and checking that instead? (since that'd be a flat value instead of an object)

johnwcallahan (commented):
I'm about halfway looking though the code changes. So far, so good.

packages/manager/src/queries/account.ts
```diff
@@ -0,0 +1,23 @@
...
+import { mutationHandlers, queryPresets } from './base';
+
+const queryKey = 'account';
+
+export const useAccount = () =>
+  useQuery<Account, APIError[]>(
+    queryKey,
+    getAccountInfo,
+    queryPresets.oneTimeFetch
+  );
```
codyfinn, line 17:
```suggestion
export const useAccount = () => {
    const accountQuery = useQuery<Account, APIError[]>(
        queryKey,
        getAccountInfo,
        queryPresets.oneTimeFetch
    );
    const { data: account, ...rest } = accountQuery;
    return { account, ...rest };
}
```

codyfinn: Just an idea to prevent having to rename the property every time we use the account hook.

packages/manager/src/IdentifyUser.tsx
```diff
@@ -96,7 +88,7 @@ export const IdentifyUser: React.FC<Props> = (props) => {
           .catch(() => setFeatureFlagsLoaded());
       }
     }
-  }, [client, userID, accountCountry, username, taxID]);
+  }, [client, userID, username, account]);
```
HanaXu, line 91: Adding on to John's comment, maybe assign accountCountry and taxID to a variable as well

packages/manager/src/features/Billing/BillingDetail.tsx
```diff
@@ -31,31 +31,23 @@ const useStyles = makeStyles((theme: Theme) => ({
 type CombinedProps = SetDocsProps & RouteComponentProps<{}>;
 
 export const BillingDetail: React.FC<CombinedProps> = (props) => {
-  const { account, requestAccount } = useAccount();
+  const { data: account, error, isLoading } = useAccount();
```
HanaXu, line 34: accountError and accountLoading to be consistent

packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/PaymentDrawer.tsx
```diff
@@ -61,11 +49,14 @@ export const getMinimumPayment = (balance: number | false) => {
 
 const AsyncPaypal = makeAsyncScriptLoader(paypalScriptSrc())(PayPal);
 
-export const PaymentDrawer: React.FC<CombinedProps> = (props) => {
-  const { accountLoading, balance, expiry, lastFour, open, onClose } = props;
+export const PaymentDrawer: React.FC<Props> = (props) => {
+  const { open, onClose } = props;
+  const { data: account, isLoading: accountLoading, refetch } = useAccount();
```
HanaXu, line 54: Let's alias refetch too

packages/manager/src/features/GlobalNotifications/EmailBounce.test.tsx
```diff
@@ -1,49 +1,48 @@
...
-            account: {
-              data: { email: 'account@example.com' } as any,
-            },
-          },
-        },
-      }
-    );
-    // Notice is visible.
-    getByTestId('billing_email_bounce');
+  // it('renders an account email bounce notice', async () => {
```
HanaXu, line 12: You can skip this test instead of commenting it out: `it.skip(...)`

tiffwong (approved):
Everything looks good to me!

packages/manager/src/queries/account.ts
bnussman: I like this idea but my only concern is that is kind of breaks our React Query convention we have going so far. I'd like to hear others thoughts on making this change

ndukatz (approved):
Looks good as far as I can tell. All tests passed.

HanaXu (approved):

PR #7738, opened 2021-07-02
Use new status page url for system status

bnussman:
## Description

Uses new env variable `LINODE_STATUS_PAGE_URL` instead of  `LINODE_STATUS_PAGE_ID` to populate system status. This also improves defaults because you no longer need an env variable for the fetch to work. 

## How to test

Ensure Manager can still fetch for system status.

### DIFF
--- a/packages/manager/.env.example
+++ b/packages/manager/.env.example
@@ -31,8 +31,8 @@ REACT_APP_APP_ROOT='http://localhost:3000'
 # Max page size requested of the Linode API:
 # REACT_APP_API_MAX_PAGE_SIZE=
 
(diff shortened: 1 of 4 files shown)


johnwcallahan (changes requested):
Looks good, but you'll need to fix the failing test by updating the glob in serverHandlers.ts, currently:

```
*statuspage.io/api/v2/incidents*
```

johnwcallahan (approved):

codyfinn (approved):
Looks good to me!

tiffwong (approved):

PR #7740, opened 2021-07-06
M3-5285: Update the Edit Credit Card drawer to use new API endpoint

bnussman:
## Description

- Use `POST /account/payment-methods` since `POST /account/credit-card` is deprecated.
- Refetch payment methods after the edit occurs
- Adds a success toast 
- Updated the credit card icon map to correctly map based on the API's card types

## How to test

- Point Cloud Manager to dev environment 
- Try editing your credit card by using the `Edit` button on your credit card's action menu. 
  - You can find a list of test credit cards [here](https://developer.paypal.com/braintree/docs/guides/credit-cards/testing-go-live/php)
- Ensure you credit card gets updated, you see a success toast, and your payment methods show the new card.

### DIFF
--- a/packages/manager/src/features/Billing/BillingPanels/PaymentInfoPanel/UpdateCreditCardDrawer/UpdateCreditCardDrawer.tsx
+++ b/packages/manager/src/features/Billing/BillingPanels/PaymentInfoPanel/UpdateCreditCardDrawer/UpdateCreditCardDrawer.tsx
@@ -1,8 +1,7 @@
-import { saveCreditCard } from '@linode/api-v4/lib/account';
 import { APIError } from '@linode/api-v4/lib/types';
 // eslint-disable-next-line no-restricted-imports
 import { InputBaseComponentProps } from '@material-ui/core';
-import { take, takeLast } from 'ramda';
+import { take } from 'ramda';
 import * as React from 'react';
 import NumberFormat, { NumberFormatProps } from 'react-number-format';
 import { Link } from 'react-router-dom';
@@ -27,6 +26,9 @@ import accountContainer, {
 import { cleanCVV } from 'src/features/Billing/billingUtils';
 import useFlags from 'src/hooks/useFlags';
 import { getAPIErrorOrDefault, getErrorMap } from 'src/utilities/errorUtils';
+import { addPaymentMethod } from '@linode/api-v4/lib/account/payments';
+import { queryClient } from 'src/queries/base';
+import { useSnackbar } from 'notistack';
 
 const useStyles = makeStyles((theme: Theme) => ({
   root: {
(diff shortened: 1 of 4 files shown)


packages/manager/src/features/Billing/BillingPanels/PaymentInfoPanel/UpdateCreditCardDrawer/UpdateCreditCardDrawer.tsx
```diff
@@ -112,21 +115,21 @@ export const UpdateCreditCardDrawer: React.FC<CombinedProps> = (props) => {
...
     })
-      .then(() => {
-        const credit_card = {
-          last_four: takeLast(4, cardNumber),
-          expiry: `${String(expMonth).padStart(2, '0')}/${expYear}`,
-          cvv,
-        };
-        // Update Redux store so subscribed components will display updated
-        // information.
-        props.saveCreditCard(credit_card);
```
HanaXu, line 129: We can probably remove the Redux code associated with this too and remove this reference from `PaymentDrawer.test.ts` cc @johnwcallahan : https://github.com/linode/manager/pull/6020/files

bnussman: If we end up merging https://github.com/linode/manager/pull/7737, that will remove all of the Redux stuff you are referring to.

packages/manager/src/features/Billing/BillingPanels/PaymentInfoPanel/UpdateCreditCardDrawer/UpdateCreditCardDrawer.tsx
```diff
@@ -112,21 +115,21 @@ export const UpdateCreditCardDrawer: React.FC<CombinedProps> = (props) => {
...
+      .then(async () => {
+        await queryClient.refetchQueries(['account-payment-methods-all']);
```
bnussman, line 129: Should I refetch payment methods or manually mutate the react query cache?

HanaXu: Good point, we seem to prefer `invalidateQueries`. Also related, I don't think payment methods will change often, so maybe we should update the queryPresets from `longLived` to `oneTimeFetch`? 🤔

HanaXu (approved):
LGTM! Just noticed the credit card in the Make a Payment drawer doesn't update w/o refresh so I can make a ticket and work on that today.

bnussman:
> LGTM! Just noticed the credit card in the Make a Payment drawer doesn't update w/o refresh so I can make a ticket and work on that today.

Good point. That drawer will probably change a bit when users can have multiple cards too

PR #7741, opened 2021-07-06
M3-5281: Add loading spinner while GPay loads

HanaXu:
## Description
- Adds loading spinner to the GPay Chip & Button components while Google Pay is still loading
- If there is an error loading or initializing Google Pay, the spinner is replaced by an error message
- Error messages while making a payment in the Make a Payment drawer are now toasts to match the Add Payment Method drawer
- Minor refactoring

## How to test
In `account/billing` on dev:
- Refresh the page and click on `Add Payment Method`. In the drawer, you should briefly see a loading spinner before it's replaced by the Google Pay Chip
- Refresh the page again and this time click on `Make a Payment`. In that drawer, you should briefly see a loading spinner near the bottom before it's replaced by the Google Pay Button
- Adding Gpay as a Payment Method & making a payment should still work as before
- Switch to Prod and repeat the above. Both spinners should be replaced by an error message

### DIFF
--- a/packages/manager/src/components/HelpIcon/HelpIcon.tsx
+++ b/packages/manager/src/components/HelpIcon/HelpIcon.tsx
@@ -1,5 +1,6 @@
-import HelpOutline from '@material-ui/icons/HelpOutline';
 import * as React from 'react';
+import HelpOutline from '@material-ui/icons/HelpOutline';
+import ErrorOutline from '@material-ui/icons/ErrorOutline';
 import IconButton from 'src/components/core/IconButton';
 import Tooltip, { TooltipProps } from 'src/components/core/Tooltip';
 
@@ -8,6 +9,8 @@ interface Props
   text: string | JSX.Element;
   className?: string;
   interactive?: boolean;
+  isError?: boolean;
+  size?: number;
   classes?: any;
   leaveDelay?: boolean;
   tooltipPosition?:
@@ -33,6 +36,8 @@ const HelpIcon: React.FC<CombinedProps> = (props) => {
     className,
     tooltipPosition,
     interactive,
+    isError,
+    size = 24,
     leaveDelay,
     classes,
   } = props;
@@ -49,7 +54,15 @@ const HelpIcon: React.FC<CombinedProps> = (props) => {
       classes={classes}
     >
       <IconButton className={className} data-qa-help-button>
-        <HelpOutline />
+        {isError ? (
+          <ErrorOutline
+            style={{
+              fontSize: size,
+            }}
+          />
+        ) : (
+          <HelpOutline />
+        )}
       </IconButton>
     </Tooltip>
   );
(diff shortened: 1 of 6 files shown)


packages/manager/src/features/Billing/BillingPanels/PaymentInfoPanel/AddPaymentMethodDrawer/AddPaymentMethodDrawer.tsx
```diff
@@ -20,25 +21,26 @@ const useStyles = makeStyles(() => ({
...
-            <GooglePayChip onAdd={onClose} />
-          </Grid>
+      <Grid className={classes.root} container>
+        <Grid item direction="column" xs={8} md={9}>
+          <Typography variant="h3">Google Pay</Typography>
+          <Typography>
+            You&apos;ll be taken to Google Pay to complete sign up.
+          </Typography>
+        </Grid>
+        <Grid item xs={4} md={3}>
```
bnussman, line 42: I think it looks better if we center the loading spinner. Here is what I came up with

```typescript
<Grid
  item
  xs={4}
  md={3}
  container
  justify="center"
  alignContent="center"
>
```

packages/manager/src/features/Billing/BillingPanels/PaymentInfoPanel/GooglePayChip.tsx
```diff
@@ -46,21 +50,19 @@ export const GooglePayChip: React.FC<Props> = (props) => {
...
 
-  const doToast = (message: string, variant: VariantType) =>
-    enqueueSnackbar(message, {
-      variant,
-    });
+  const handleMessage = (message: string, variant: VariantType) => {
+    makeToast(message, variant);
+    if (variant === 'success') {
+      onClose();
+    }
```
bnussman, line 64: nice job, this is clever

packages/manager/src/features/Billing/BillingPanels/PaymentInfoPanel/GooglePayChip.tsx
```diff
@@ -70,20 +72,24 @@ export const GooglePayChip: React.FC<Props> = (props) => {
...
+      handleMessage
     );
   };
 
+  if (status === 'error' || clientTokenError) {
+    return <Notice error text="Error loading Google Pay." />;
+  }
+
+  if (initializationError) {
+    return <Notice error text="Error initializing Google Pay." />;
```
bnussman, line 84: Maybe we should change the error text to just error so the error state looks less weird. Just a thought.

Screen Shot 2021-07-07 at 3 15 51 PM

bnussman (approved):
I like the error icon! Looks good!

My thoughts are
- We should make sure using a toast for these error states is ok
- We might want to make the error icon its own component

packages/manager/src/components/HelpIcon/HelpIcon.tsx
```diff
@@ -33,6 +36,8 @@ const HelpIcon: React.FC<CombinedProps> = (props) => {
     className,
     tooltipPosition,
     interactive,
+    isError,
```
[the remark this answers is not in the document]
HanaXu: It would have a lot of similar code as the HelpIcon so I was thinking about refactoring this component later to be more generic

PR #7743, opened 2021-07-07
M3-4721: Changes to how account balances are communicated in app

DevDW:
## Description
For `payment_due` notifications where a user is outside of their grace period, the API will now be returning a severity of `critical` instead of `major`. We will key off of `critical` to govern the intensity of our messaging regarding account balance, on /account/billing and in the notification drawer.


DevDW:
`git rebase` is not being cooperative here when I try to drop those unrelated commits so I'm going to close this and open a new PR

PR #7744, opened 2021-07-07
M3-4721: Changes to how account balances are communicated in app

DevDW:
## Description
For `payment_due` notifications where a user is outside of their grace period, the API will now be returning a severity of `critical` instead of `major`. We will key off of `critical` to govern the intensity of our messaging regarding account balance, on /account/billing and in the notification drawer.

## How to Test
Check two accounts: one with a past due balance, and one with an outstanding balance that is not yet past due (reach out for a couple of accounts that meet this condition already).

The account with a past due balance should have an Account Balance panel and a drawer notification that look as follows:

Screen Shot 2021-07-08 at 12 18 38 PM

Screen Shot 2021-07-08 at 12 18 53 PM

The account with an outstanding balance that is not yet past due should have an Account Balance panel and a drawer notification that look as follows:

Screen Shot 2021-07-08 at 12 20 20 PM

Screen Shot 2021-07-08 at 12 20 35 PM

Other balance cases (no balance or a credit) should be unchanged from their current handling.

### DIFF
--- a/packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
+++ b/packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
@@ -87,12 +87,10 @@ export const BillingSummary: React.FC<BillingSummaryProps> = (props) => {
   const notifications = useNotifications();
 
   // If a user has a payment_due notification with a severity of critical, it indicates that they are outside of any grace period they may have and payment is due immediately.
-  const isBalanceOutsideGracePeriod = Boolean(
-    notifications.find(
-      (notification) =>
-        notification.type === 'payment_due' &&
-        notification.severity === 'critical'
-    )
+  const isBalanceOutsideGracePeriod = notifications.some(
+    (notification) =>
+      notification.type === 'payment_due' &&
+      notification.severity === 'critical'
   );
 
   const { promotions, balanceUninvoiced, balance } = props;
--- a/packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
+++ b/packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
@@ -150,34 +150,17 @@ export const BillingSummary: React.FC<BillingSummaryProps> = (props) => {
     promotions && promotions.length > 0 ? { xs: 12, md: 4 } : { xs: 12, sm: 6 };
 
   const generateBalanceJSX = (balance: number, pastDueBalance: boolean) => {
-    if (pastDueBalance) {
-      return (
-        <Typography style={{ marginTop: 16 }}>
-          <button
-            className={classes.makeAPaymentButton}
-            onClick={() => replace(routeForMakePayment)}
-          >
-            Make a payment immediately
-          </button>{' '}
-          to avoid service disruption.
-        </Typography>
-      );
-    }
-
-    if (balance > 0) {
-      return (
-        <Typography style={{ marginTop: 16 }}>
-          <button
-            className={classes.makeAPaymentButton}
-            onClick={() => replace(routeForMakePayment)}
-          >
-            Make a payment.
-          </button>
-        </Typography>
-      );
-    }
-
-    return null;
+    return balance > 0 ? (
+      <Typography style={{ marginTop: 16 }}>
+        <button
+          className={classes.makeAPaymentButton}
+          onClick={() => replace(routeForMakePayment)}
+        >
+          {pastDueBalance ? 'Make a payment immediately' : 'Make a payment.'}
+        </button>
+        {pastDueBalance ? `${' '} to avoid service disruption.` : null}
+      </Typography>
+    ) : null;
   };
 
   return (
--- a/packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
+++ b/packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
@@ -16,6 +16,7 @@ import Typography from 'src/components/core/Typography';
 import Currency from 'src/components/Currency';
 import DateTimeDisplay from 'src/components/DateTimeDisplay';
 import HelpIcon from 'src/components/HelpIcon';
+import useNotifications from 'src/hooks/useNotifications';
 import PaymentDrawer from './PaymentDrawer';
 
 const useStyles = makeStyles((theme: Theme) => ({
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
```diff
@@ -83,6 +84,16 @@ interface BillingSummaryProps {
 
 export const BillingSummary: React.FC<BillingSummaryProps> = (props) => {
   const classes = useStyles();
+  const notifications = useNotifications();
+
+  // If a user has a payment_due notification with a severity of critical, it indicates that they are outside of any grace period they may have and payment is due immediately.
+  const isBalanceOutsideGracePeriod = Boolean(
+    notifications.find(
```
codyfinn, line 91: Could the `.includes` method be used here instead of `.find`? This would return a boolean by default and would remove the requirement to typecast the result.

DevDW: `.includes()` is more for straight element value checks I believe, but I used `.some()` instead of `.find()` to eliminate the need for typecasting 👍

packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
```diff
@@ -120,25 +128,58 @@ export const BillingSummary: React.FC<BillingSummaryProps> = (props) => {
...
+    ? 'Balance'
+    : balance < 0
+    ? 'Credit'
+    : 'You have no balance at this time.';
 
   const accountBalanceClassnames = classnames({
-    [classes.noBalance]: balance === 0,
-    [classes.positiveBalance]: balance > 0,
+    [classes.noBalanceOrNotDue]:
+      balance === 0 || (balance > 0 && !isBalanceOutsideGracePeriod),
```
HanaXu, line 143: What about `balance === 0 || !pastDueBalance` instead?

packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
```diff
@@ -167,17 +208,7 @@ export const BillingSummary: React.FC<BillingSummaryProps> = (props) => {
...
-                <button
-                  className={classes.makeAPaymentButton}
-                  onClick={() => replace('/account/billing/make-payment')}
-                >
-                  Make a payment immediately
-                </button>{' '}
-                to avoid service disruption.
-              </Typography>
-            ) : null}
+            {generateBalanceJSX(balance, pastDueBalance)}
```
HanaXu, line 211: What about this instead? (less code duplication)
``` 
{balance > 0 ? (
  <Typography style={{ marginTop: 16 }}>
    <button
      className={classes.makeAPaymentButton}
      onClick={() => replace('/account/billing/make-payment')}
    >
      {pastDueBalance
        ? 'Make a payment immediately'
        : 'Make a Payment'}
    </button>
    {pastDueBalance ? `${' '}to avoid service disruption.` : null}
  </Typography>
) : null}
```

packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
DevDW: For `pastDueBalance` to be true, the balance has to be greater than 0 and outside the user's grace period; using `!pastDueBalance` here instead would make this class kick in at times when we don't want, e.g. when the user has a credit (balance < 0, therefore `pastDueBalance` would be false).

tiffwong (commented):
Was able to test an account with outstanding balance that's not due yet, will test one that's overdue tomorrow

packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
```diff
@@ -120,25 +128,41 @@ export const BillingSummary: React.FC<BillingSummaryProps> = (props) => {
...
     promotions && promotions.length > 0 ? { xs: 12, md: 4 } : { xs: 12, sm: 6 };
 
+  const generateBalanceJSX = (balance: number, pastDueBalance: boolean) => {
+    return balance > 0 ? (
+      <Typography style={{ marginTop: 16 }}>
+        <button
+          className={classes.makeAPaymentButton}
+          onClick={() => replace(routeForMakePayment)}
+        >
+          {pastDueBalance ? 'Make a payment immediately' : 'Make a payment.'}
```
johnwcallahan, line 159: IMO "Make a payment" doesn't need a period, but that's kind of subjective.

packages/manager/src/features/Billing/BillingPanels/BillingSummary/BillingSummary.tsx
```diff
@@ -120,25 +128,41 @@ export const BillingSummary: React.FC<BillingSummaryProps> = (props) => {
...
+      balance === 0 || (balance > 0 && !isBalanceOutsideGracePeriod),
+    [classes.pastDueBalance]: pastDueBalance,
     [classes.credit]: balance < 0,
   });
 
   // The layout changes if there are promotions.
   const gridDimensions: Partial<Record<Breakpoint, GridSize>> =
     promotions && promotions.length > 0 ? { xs: 12, md: 4 } : { xs: 12, sm: 6 };
 
+  const generateBalanceJSX = (balance: number, pastDueBalance: boolean) => {
```
johnwcallahan, line 152: `generateBalanceJSX` doesn't need to be a function... `balance` and `pastDueBalance` are already in scope. You could do something like:

```ts
const BalanceDisplay =
  balance > 0 ? (
    <Typography style={{ marginTop: 16 }}>
   // etc.
```

packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
```diff
@@ -227,6 +227,14 @@ const interceptNotification = (
     return notification;
   }
 
+  if (
+    notification.type === 'payment_due' &&
+    notification.severity !== 'critical'
+  ) {
+    // A critical severity indicates the user's balance is past due; temper the messaging a bit outside of that case by replacing the exclamation point.
+    notification.message = notification.message.replace('!', '.');
```
johnwcallahan, line 235: Nice touch.

johnwcallahan (approved):
The approach and UX look good to me, just a few minor comments.

HanaXu (approved):

PR #7749, opened 2021-07-09
M3-5772: Cleanup Button story

tiffwong:
## Description

Tasks:

- [x] Remove examples of "Cancel" and "Remove" buttons (also remove related code from the actual component)
- [x] Remove "Loading with Text" story
- [x] Remove references to "compact" and replace with the code for "superCompact"
- [x] Add example of a compact button for secondary buttons only
- [x] Document use cases for the different variants
- [x] Add details to `ArgsTable`

Future tasks to be addressed in a separate PR:

- Figure out why the Component API Button's generated code isn't accurate

## How to test

- Check the buttons throughout the app
- Run `yarn storybook` to see the documentation for Buttons (it's nested inside the UI Elements folder)


packages/manager/src/features/Support/AttachFileListItem.tsx
```diff
@@ -76,54 +72,52 @@ export const AttachFileListItem: React.FC<CombinedProps> = (props) => {
     file.errors && file.errors.length ? file.errors[0].reason : undefined;
 
   return (
-    <React.Fragment>
```
tiffwong, line 79: Remove unnecessary `<React.Fragment>`

DevDW (approved):
Looks good! The Button docs in Storybook :plusplus:

packages/manager/src/components/Button/button.stories.mdx
```diff
@@ -1,16 +1,162 @@
...
+<!-- TODO: Look into why the code generated from binding the component is not accurate -->
+
+export const Template = (args) => (
+  <Button buttonType="primary" {...args}>
+    I'm a button, style me using the table below
+  </Button>
+);
+
+<Canvas>
+  <Story name="Default">{Template.bind({})}</Story>
```
HanaXu, line 132: If we set default values for the args here, the extra `Set ...` click goes away:

```
<Story
  name="Default" 
  args={{
    compact: false,
    disabled: false,
    disableRipple: false,
    fullWidth: false,
    loading: false,
    tooltipText: '',
  }}
>
  {Template.bind({})}
</Story>
```

HanaXu:
Can we address these console warnings and errors?

image

image

This one's probably a bit out of scope and can probably be looked into later, but I don't think Storybook should be trying to make calls to the API 🤔 : 
image

HanaXu (approved):
Looks great! 😎

PR #7751, opened 2021-07-09
Add new minimum TTL values for domains

rmcintosh:
## Description

Now that APIv4 accepts a couple of lower values for TTLs, let's allow it via the frontend as well.


tiffwong (approved):

HanaXu (changes requested):
We'll also want to add this to the `msToReadable` function:
```
30: '30 seconds',
120: '2 minutes',
```

otherwise the TTL fields display blank:
image

PR #7732, opened 2021-07-01
Hotfix/referral link

tiffwong:
## Description

Fix referral link showing for customers who have not made a payment/met the $25 minimum threshold

## How to test

Check `/profile/referrals` for accounts that should and shouldn't be showing the referral link

PR #7752, opened 2021-07-09
Cleanup Referrals

tiffwong:
## Description

Removal conditional logic since the referral changes have been released to GA

## How to test

Make sure the referral states look like they're supposed to, message me if you need help seeing the different states


packages/manager/src/featureFlags.ts
```diff
@@ -22,7 +22,6 @@ export interface Flags {
   firewalls: boolean;
   oneClickApps: OneClickApp;
   promotionalOffers: PromotionalOffer[];
-  thirdPartyAuth: boolean;
```
tiffwong, line 25: This flag is also no longer used so I removed it

packages/manager/src/features/Profile/Referrals/Referrals.tsx
```diff
@@ -103,183 +91,111 @@ type CombinedProps = StateProps;
...
-    code,
-    url,
-    total,
-    completed,
-    pending,
-    credit,
-  } = props;
+  const { profileLoading, url, total, completed, pending, credit } = props;
 
   const allowReferral = url !== undefined;
```
bnussman, line 97: Getting an empty field. Do we need this change again?
```ts
const allowReferral = Boolean(url);
```

bnussman (changes requested):

packages/manager/src/features/Profile/Referrals/Referrals.tsx
DevDW: I'm getting the same thing -- I think that when the hotfix went out for this, `staging` wasn't merged into `develop` so this change didn't get in

tiffwong: @bnussman @DevDW that's exactly what happened...thanks for catching that! 😓

DevDW (approved):
LGTM now, can you add the other changes from https://github.com/linode/manager/pull/7732/files so that `develop` gets updated when you merge this?

bnussman (approved):

PR #7753, opened 2021-07-12
M3-5163: Fix Nodebalancer Table console error children with the same key

bnussman:
## Description

- Fixes a console error telling us we need to have children with difficulty keys. 
- Fixed by using the Nodebalacer's config id when rendering the configs instead of duplicating the Nodebalancer's id. 

## How to test

- Make sure you have at least one Nodebalancer and that Nodebalancer has more than one config on it.
- Go to `/nodebalancers`
- Check your console for errors regarding duplicate keys.


tiffwong (approved):

HanaXu (approved):
LGTM! Could we also fix the same key error when selecting a region in `nodebalancers/create`?

image

PR #7754, opened 2021-07-12
M3-5280: Expand all Linode Settings accordions by default

bnussman:
## Description

- Expands all Linode Settings accordions by default
- Forces fetch for disks when password section is mounted

## How to test

- Visit `/linodes/{id}/settings`
- Make sure all accordions are expanded by default and all forms function as expected

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeSettingsPasswordPanel.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeSettingsPasswordPanel.tsx
@@ -59,6 +59,10 @@ class LinodeSettingsPasswordPanel extends React.Component<
     disks: [],
   };
 
+  componentDidMount() {
+    this.searchDisks();
+  }
+
   handleSubmit = () => {
     const { diskId, value } = this.state;
     const { linodeId, isBareMetalInstance } = this.props;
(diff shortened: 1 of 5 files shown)


tiffwong (approved):

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeSettingsPasswordPanel.tsx
```diff
@@ -59,6 +59,10 @@ class LinodeSettingsPasswordPanel extends React.Component<
     disks: [],
   };
 
+  componentDidMount() {
+    this.searchDisks();
```
DevDW, line 63: With this and the expanded state being the default, I wonder how necessary calling `searchDisks()` upon every panel expansion (through `handlePanelChange()`) is now

DevDW (approved):
Looks good 👍

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeSettingsPasswordPanel.tsx
bnussman: I think I will opt to leave it as is. Like you said, handlePanelChange only calls `searchDisks()` when the drawer changes to `open` so I don't think it hurts to leave it

PR #7755, opened 2021-07-12
M3-5072: Google Pay

HanaXu:
## Description
Adds Google Pay

## How to test
In `account/billing`:
- Test the Make a Payment flow
- Test adding Google Pay as a Payment Method
- Test making a different payment method the default
- Going to `/account/billing/add-payment-method` should open the drawer
- The "Google Pay is now available..." notice should display in the Payment Methods drawer if Google Pay has not been added as a payment method yet. The notice should be dismissible and saved to the user's preferences.
- Ensure adding a credit card to an account with no payment methods work
- Ensure updating a credit card works
- Ensure the Google Pay loading/success/error states work as expected
- Billing & Payment History table updates after making a payment and works as expected for both drawers
- Turn the feature flag off and the Make a Payment drawer should default to the old UI w/o GPay
    - The Payment Methods panel will show the updated UI, but that's fine


bnussman:
Should we do something about `'nonce' is declared but its value is never read.` before we merge? I tried to disable linting on that line but couldn't get it to stop

HanaXu:
I added `@ts-expect-error will be used later` with the merge to line 88 in `GooglePay.ts`, so the error shouldn't be showing up anymore 🤔

bnussman:
> I added `@ts-expect-error will be used later` to line 88 in `GooglePay.ts`, so the error shouldn't be showing up anymore 🤔

You're right, looks good!

packages/manager/src/features/Billing/Providers/GooglePay.ts
```diff
@@ -0,0 +1,129 @@
...
+    return setMessage('Unable to open Google Pay.', 'error');
+  }
+
+  let paymentDataRequest;
+
+  try {
+    paymentDataRequest = await googlePaymentInstance.createPaymentDataRequest({
+      merchantInfo: {
+        merchantId: GPAY_MERCHANT_ID || '',
+      },
```
bnussman, line 54: We might want to add `merchantName: "Linode"` or `merchantName: "Linode, LLC"` or something. I'm not sure what will happen in prod without this

packages/manager/src/features/Billing/Providers/GooglePay.ts
```diff
@@ -0,0 +1,129 @@
...
+    });
+  } catch (error) {
+    return setMessage('Unable to open Google Pay.', 'error');
+  }
+
+  const googlePayClient = new google.payments.api.PaymentsClient({
+    environment: GPAY_CLIENT_ENV as google.payments.api.Environment,
+    merchantInfo: {
+      merchantId: GPAY_MERCHANT_ID || '',
+    },
```
bnussman, line 67: And maybe make the `merchantInfo` a variable because we put it in two places

packages/manager/src/features/Billing/Providers/GooglePay.ts
HanaXu: I assumed the merchant name would be associated with the `merchantId` 🤔 

On that note, do we want to do anything with the commented out `merchantId` on line 35?

bnussman: Yep, will update in my commit

tiffwong (approved):
Everything looks good to me, great job! 🚢

ndukatz (approved):
Looks good to me. Great job folks. All e2e passed as well. 

The only thing I noticed is when editing/adding a new card, the forms don't support autofilling with 1pass, and the gpay one doesn't even support copying and pasting. This may be by design so that's probably fine. But that is often how I add my cc info on sites.

DevDW (commented):
From my testing, things are looking good so far... in the case where an account has no existing payment methods though (e.g., Google Pay never added and no credit card on file), are they not able to add a credit card at this time?

HanaXu:
@DevDW Good point. Are users currently able to add a credit card if the account has no existing payment methods?

DevDW:
> @DevDW Good point. Are users currently able to add a credit card if the account has no existing payment methods (with the feature flag off)?

Yes, it looks like this:

Screen Shot 2021-07-13 at 10 48 30 AM

(user clicks "Edit"...)

Screen Shot 2021-07-13 at 10 48 42 AM

packages/manager/src/factories/accountPayment.ts
```diff
@@ -0,0 +1,23 @@
...
+      'Visa',
+      'MasterCard',
+      'Discover',
+      'American Express',
+      'JCB',
+    ]),
+  })),
+  created: '2021-05-21T14:27:51',
+  type: Factory.each(() =>
+    pickRandom(['credit_card', 'credit_card', 'google_pay'])
```
DevDW, line 20: Do we need to skew towards credit cards by including it twice here?

DevDW (approved):
Functionality looks good to me after another round of testing, great work! Left a few minor comments

packages/manager/src/factories/accountPayment.ts
bnussman: Not necessary whatsoever, I just assumed we would see more credit cards than 3ed party payment methods for now

bnussman:
Since users can only have 1 Google Pay payment method right now, should we add some sort of indication that adding another Google Pay will actually replace their current Google Pay?

codyfinn (approved):
Looks good to me

ndukatz (approved):

johnwcallahan (approved):
I've looked through a lot of the code at tested functionality, so far it's working well. I'm approved this so we can go ahead and get it to staging to conduct further testing.

bnussman (approved):

PR #7756, opened 2021-07-12
M3-5177: Use React Query for Account Settings

bnussman:
## Description

- Remove `/v4/account/settings` data from Redux
- Uses **React Query**'s  _Query_ and _Mutation_ for `/v4/account/settings` 

## How to test

- Test all general use of Cloud Manager. Things that are most likely to break include `backups`, `object storage`, `managed`, `longview` and `linode network helper`.
- Ensure all everything at [`/account/settings`](http://localhost:3000/account/settings) still work as expected


packages/manager/src/features/Account/EnableManaged.tsx
```diff
@@ -88,8 +87,8 @@ export const EnableManaged: React.FC<CombinedProps> = (props) => {
     enableManaged()
       .then(() => {
         handleClose();
-        update({ managed: true });
-        //push('/managed');
+        updateAccountSettingsData({ managed: true });
+        // push('/managed');
```
codyfinn, line 91: Can you remove this comment if it is not used?

codyfinn (approved):
I left a nitpicking comment but other than that this looks great! Nice job moving us closer to complete react query usage!

packages/manager/src/store/backupDrawer/index.ts
```diff
@@ -293,23 +296,20 @@ export const enableAutoEnroll: EnableAutoEnrollThunk = () => (
...
   const state = getState();
   const { backups } = state;
-  const hasBackupsEnabled = pathOr(
-    false,
-    ['__resources', 'accountSettings', 'data', 'backups_enabled'],
-    state
-  );
   const shouldEnableBackups = Boolean(backups.autoEnroll);
+  const { data: accountSettings } = useAccountSettings();
+  const { mutateAsync: updateAccountSettings } = useMutateAccountSettings();
```
bnussman, line 301: OH NO :0 I made an invalid hook call without even noticing this.

ndukatz (approved):
Looks good! all E2E passed. all the areas you specified looked good in manual checks as well. Also tested managed.

packages/manager/src/features/Account/GlobalSettings.tsx
```diff
@@ -103,38 +81,53 @@ const GlobalSettings: React.FC<CombinedProps> = (props) => {
...
+
+  const {
+    backups_enabled,
+    managed,
+    network_helper,
+    object_storage,
+  } = accountSettings;
+
+  const toggleAutomaticBackups = () => {
+    updateAccount({ backups_enabled: !backups_enabled }).catch(displayError);
```
HanaXu, line 107: Do we want to return here as well?

HanaXu (approved):
LGTM!

packages/manager/src/features/Account/GlobalSettings.tsx
bnussman: Good catch.  Going to remove the returns, they are not needed.

PR #7757, opened 2021-07-13
M3-5220 & 5213: Firewall Details table header and Longview Plan chip

tiffwong:
## Description

This PR defines the table header as 46px so it'll be consistent throughout the app and also adjusts the Longview Plan chip style to match the one used on the Kubernetes Landing table (I also tightened up the spacing while I was in here)

## How to test

**M3-5220:**
Go to any Firewall Detail page and on the Rules tab, check the height for the table header and row. The table header should be 46px and the row should be 40px.

**M3-5213:**
Make sure the account you're logged into is **not** managed  and go to `/longview/plan-details` 

Before:
Screen Shot 2021-07-13 at 2 06 35 PM

After:
Screen Shot 2021-07-13 at 2 04 45 PM


packages/manager/src/features/Longview/LongviewLanding/LongviewPlans.tsx
```diff
@@ -35,43 +35,26 @@ import { MapState } from 'src/store/types';
 import { getAPIErrorOrDefault } from 'src/utilities/errorUtils';
 
 const useStyles = makeStyles((theme: Theme) => {
-  const border = `1px solid ${theme.bg.main}`;
-
   return {
```
DevDW, line 38: Can we get rid of this `return` and do `const useStyles = makeStyles((theme: Theme) => ({ ... })` like we do elsewhere?

DevDW (approved):
LGTM -- I am seeing 41px for the row height though, not 40px:

Screen Shot 2021-07-15 at 11 04 36 AM

HanaXu (approved):
Changes LGTM, but do we need this outer paper? Looks a bit awkward imo

image

tiffwong:
@HanaXu good point, I removed the `Paper` when it's only the table that's visible. I also hid the table for restricted users so lmk what you think!

HanaXu:
@tiffwong Good catch on the restricted users, changes LGTM! Another thought is to toast the success/error messages similar to `longview/clients`, but that can be handled in another story since it's a bit out of scope.

tiffwong:
@HanaXu turns out that users with `read` or `read-write` access can see the current plan but users with no access don't so I opted to only hide the table for users with no access 😅

packages/manager/src/features/Longview/LongviewLanding/LongviewPlans.tsx
```diff
@@ -250,6 +228,13 @@ export const LongviewPlans: React.FC<CombinedProps> = (props) => {
 
   const isManaged = accountSettings ? accountSettings.managed : false;
 
+  // Hide table if current plan is not being displayed
+  // ie. Users with no access to Longview
+  const isTableDisplayed =
```
tiffwong, line 233: Not the most elegant way but I couldn't figure out how to access map through the Longview permissions

HanaXu:
@tiffwong If a user has no access, should we be displaying `Unauthorized` (similar to `/account/billing` etc) instead of `You don't have permissions to change the Longview plan...`?

tiffwong:
@HanaXu hmm that's a good point. I just left it as it currently is in production but this might be a good thing to bring up to Matthew. In the meantime, I might merge it and address that in a future PR if everything else looks good

PR #7705, opened 2021-06-15
M3 5250: Clean up Tables CMR

tiffwong:
## Description

Replace old Table-related components with the new CMR version and rename the file wherever it's being imported

## How to test

Make sure all the components that were updated all look good

PR #7761, opened 2021-07-13
M3-5260: Fix staticContext console warning

tiffwong:
## Description

Fix the `staticContext` console warning caused by #7705 

## How to test

Go to any page with a table on it and the console should no longer be displaying a warning


bnussman (approved):
Looks like `Tag.tsx` does the same thing by declaring and no using staticContent

DevDW (approved):

PR #7763, opened 2021-07-13
M3-5226: LKE Detail page UI tweaks

DevDW:
## Description
- Linkify Linode labels and decrease font size 
- Move status icons into "Status" column
- Remove icons in left-hand section of LKE Summary Details panel
- Delete old, unused files that were used for alternate status indicator implementations


packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeSummaryPanel.tsx
```diff
@@ -8,13 +8,7 @@ import * as React from 'react';
 import { useDispatch } from 'react-redux';
 import { useHistory } from 'react-router-dom';
 import DetailsIcon from 'src/assets/icons/code-file.svg';
-import CPUIcon from 'src/assets/icons/cpu-icon.svg';
-import DiskIcon from 'src/assets/icons/disk.svg';
 import DownloadIcon from 'src/assets/icons/lke-download.svg';
-import MapPin from 'src/assets/icons/map-pin-icon.svg';
-import MiniKube from 'src/assets/icons/mini-kube.svg';
-import PriceIcon from 'src/assets/icons/price-icon.svg';
```
tiffwong, line 16: These three icons aren't being used anywhere else so they can probably be deleted

tiffwong (approved):
Can you add a border around table similar to the `SelectPlanPanel` in the Linode Create flow? Left one other cleanup items but this looks great otherwise!

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/NodePoolsDisplay/NodeTable.tsx
```diff
@@ -195,9 +205,11 @@ export const NodeRow: React.FC<NodeRowProps> = React.memo((props) => {
...
   const recentEvent = useRecentEventForLinode(instanceId ?? -1);
 
-  const rowLink = instanceId ? `/linodes/${instanceId}` : undefined;
-  const statusIndicator =
-    nodeStatus === 'ready' && instanceStatus === 'running' ? 'active' : 'other';
+  const linodeLink = instanceId ? `/linodes/${instanceId}` : undefined;
+  const iconStatus =
+    nodeStatus === 'ready' && instanceStatus === 'running'
+      ? 'active'
+      : 'inactive';
```
codyfinn, line 212: This feels like a code smell to me. Upon looking into it we do not currently have a type defined for the status of a Node. The current type for a `PoolNodeResponse` looks like this `export interface PoolNodeResponse {
    id: string;
    instance_id: number | null;
    status: string;
}` We are only using a basic string type when we should be using a type that defines what valid strings are for Node's status.

DevDW: I agree 👍  As discussed on Slack, this will be handled in a new JIRA ticket

codyfinn (approved):
Looks good to me!

PR #7765, opened 2021-07-14
Remove remaining CMR flag dependency and clean up Accordion

tiffwong:
## Description

This PR was originally meant to remove the remaining CMR flag dependency but I got sidetracked from cleaning up `Accordion`...

For the original scope, there should no longer be any usages of `flags.CMR` used throughout the app.

As for cleaning up `Accordion`, I replaced the old file with the new version, refactored it into a FC, and moved the styles over to the theme file.

## How to test

Make sure nothing is broken.

The `Accordion` component should look a little different now (ie. the caret icon on the right instead of the left, curious to hear everyone's feedback).


packages/manager/src/components/Accordion/Accordion.tsx
```diff
@@ -86,70 +41,73 @@ export interface Props extends AccordionProps {
...
+    summaryProps,
+    detailProps,
+    headingProps,
+    actions,
+    success,
+    warning,
+    error,
+    headingNumberCount,
+    ...accordionProps
+  } = props;
```
DevDW, line 70: Maybe move this to the top of the function, destructure `defaultExpanded` here as well and initialize `open` to just `defaultExpanded` instead of `props.defaultExpanded`

packages/manager/src/components/Accordion/Accordion.tsx
```diff
@@ -86,70 +41,73 @@ export interface Props extends AccordionProps {
...
+      <AccordionSummary
+        onClick={handleClick}
+        expandIcon={<KeyboardArrowDown className="caret" />}
+        {...summaryProps}
+        data-qa-panel-summary={props.heading}
+      >
+        <Typography {...headingProps} variant="h3" data-qa-panel-subheading>
+          {props.heading}
+        </Typography>
+        {headingNumberCount && headingNumberCount > 0 && (
```
DevDW, line 85: `headingNumberCount && headingNumberCount > 0 ? ( ... ) : null`

packages/manager/src/features/Longview/LongviewDetail/DetailTabs/Apache/Apache.tsx
```diff
@@ -104,10 +102,7 @@ export const Apache: React.FC<Props> = (props) => {
           alignItems="center"
         >
           <div>
-            <Typography
-              className={flags.cmr ? classes.cmrSpacing : ''}
-              variant="h2"
-            >
+            <Typography className={classes.cmrSpacing} variant="h2">
```
DevDW, line 105: Perhaps as a follow-up ticket, since this PR marks (some of) the last CMR cleanup, `cmrSpacing` can be renamed throughout the app to something better/more accurate now

DevDW (approved):
LGTM, left a few optional comments. It could just be because I'm used to it, but I like the caret on the left better

codyfinn (approved):
Looks good to me!

packages/manager/src/features/Longview/LongviewDetail/DetailTabs/Apache/Apache.tsx
tiffwong: Good idea! There's also a few more CMR versions floating around that I was going to address once this is merged so I'll add this todo to that future PR

PR #7766, opened 2021-07-14
M3-5247: Remove "Other Entities" from Monthly Network Transfer section of Network tab

DevDW:
## Description
The "Other Entities" data point has been a recurring source of confusion and opaque fluctuations. This PR removes it since Monthly Network Transfer Pool was added back to the Linodes landing page a while ago.

## How to test
Checking the "Network" tab of a Linode detail page, "Other Entities" should no longer be listed in the "Monthly Network Transfer" section. Confirm that no other data or graphs are thrown off by this removal.

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/NetworkingSummaryPanel/NetworkTransfer.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/NetworkingSummaryPanel/NetworkTransfer.tsx
@@ -126,15 +126,11 @@ const TransferContent: React.FC<ContentProps> = (props) => {
   const classes = useStyles();
 
   /**
-   * In this component we display four pieces of information:
+   * In this component we display three pieces of information:
    *
    * 1. Account-level transfer quota for this month
    * 2. The amount of transfer THIS Linode has used this month
-   * 3. The amount of transfer OTHER things on your account have used this month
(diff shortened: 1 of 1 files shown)


packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/NetworkingSummaryPanel/NetworkTransfer.tsx
```diff
@@ -185,11 +180,6 @@ const TransferContent: React.FC<ContentProps> = (props) => {
       <Typography className={`${classes.legendItem} ${classes.darkGreen}`}>
         {linodeLabel} ({linodeUsedInGB} GB)
       </Typography>
-      {otherEntitiesUsedInGB > 0 && (
-        <Typography className={`${classes.legendItem} ${classes.lightGreen}`}>
```
tiffwong, line 189: The `lightGreen` styles can be removed since it's not being used anywhere else in this component

tiffwong (approved):
Left one cleanup item, looks good otherwise!

bnussman (approved):
Looks good but did we consider only removing this for accounts where they only have a single Linode?

DevDW:
> Looks good but did we consider only removing this for accounts where they only have a single Linode?

It wasn't considered explicitly, but since this has come up as a point of confusion several times before (even in demos), removing it across the board is likely the best bet

PR #7770, opened 2021-07-16
Improve Payment Methods Mocking

bnussman:
## Description

- Improves Payment Methods mocking
- ~~Makes Payment Methods on `/account/billing` scrollable~~
  - We will not make it scrollable because native browsers hide the scroll bar when user is not scrolling and it unintentionally  hides payment methods from the user

## How to test

- Test MSW on `/account/billing`
- ~~Make sure you are able to scroll in the payment methods panel~~

### DIFF
--- a/packages/manager/src/factories/accountPayment.ts
+++ b/packages/manager/src/factories/accountPayment.ts
@@ -1,14 +1,23 @@
 import * as Factory from 'factory.ts';
 import { PaymentMethod } from '@linode/api-v4';
+import { pickRandom } from 'src/utilities/random';
 
 export const paymentMethodFactory = Factory.Sync.makeFactory<PaymentMethod>({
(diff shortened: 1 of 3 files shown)


packages/manager/src/factories/accountPayment.ts
```diff
@@ -1,13 +1,22 @@
...
+      'MasterCard',
+      'Discover',
+      'American Express',
+      'JCB',
+    ]),
+  })),
   created: '2021-05-21T14:27:51',
-  type: 'credit_card',
+  type: Factory.each(() =>
+    pickRandom(['credit_card', 'credit_card', 'google_pay'])
```
bnussman, line 19: Repeated `credit_card` twice intentionally because I would expect credit cards to be more common payment methods than `google_pay`

HanaXu (approved):

PR #7771, opened 2021-07-16
Allow dismissing of Google Pay available notice in Payment Methods paper

HanaXu:
## Description
Allows dismissing of the `Google Pay is now available...` notice and saves it to the user's preferences

## How to test
In `account/billing` on an account without Google Pay added as a payment method, you should see the `Google Pay is now available...` notice in the Payment Method paper along with an `x` button. Clicking the `x` button should dismiss the notice and the notice should no longer appear after refresh.


johnwcallahan (approved):
Functionality works well. To my eye the "X" button should be closer to the copy (instead of right-aligned) but that's subjective.
Screen Shot 2021-07-16 at 1 51 23 PM

bnussman (approved):
Looks good! I kind of agree with @johnwcallahan  but if the `X` was beside the `Add Google Pay` link, I think two clickable things close to each other may not be great either.

PR #7772, opened 2021-07-16
M3-5276 added Mui-disabled rules for MuiButton

codyfinn:
## Description

Changed the colors of disabled buttons for the dark theme.

## How to test

Change to the dark theme. Make sure that when buttons are disabled they match the new spec of `#454b54` for the background and `#5c6470` for the text.


packages/manager/src/themes.ts
```diff
@@ -305,8 +305,12 @@ const darkThemeOptions = {
...
         },
         '&$disabled': {
-          color: '#888',
-          opacity: 0.5,
+          backgroundColor: '#454b54',
+          color: '#5c6470',
+        },
+        '&Miu-disabled': {
+          backgroundColor: '#454b54',
+          color: '#5c6470',
```
tiffwong, line 313: These don't do anything since they're missing the `.` to signify that it's a class but they're also not needed since the `$disabled` basically converts the styles to that when it's rendered

packages/manager/src/themes.ts
```diff
@@ -324,9 +328,13 @@ const darkThemeOptions = {
         '&:active': {
           backgroundColor: primaryColors.dark,
         },
+        '&Miu-disabled': {
+          backgroundColor: '#454b54',
+          color: '#5c6470',
+        },
```
tiffwong, line 334: Same thing here

tiffwong (changes requested):
Changes look good but there's duplicate styles that aren't doing anything so will approve once those are cleaned up 👍

bnussman (approved):
Disabled buttons in dark mode look much better now 🙂

tiffwong (approved):

PR #7773, opened 2021-07-16
E2e create linode from image test

ndukatz:
## Description
This change was needed because of some recent ui changes to this page

**What does this PR do?**
This is a test addition as well as a cypress update and a few other small things that I noticed needed changes after the update. 

The first test mocks the image and the linode then goes through the flow of creating a linode from the images tab, and tests that an image shows up for selection. The second test uses the same mocks but navigates directly to the image detail to create the linode. 

## How to test
- `yarn up` in one terminal
- `yarn cy:e2e -s ./cypress/integration/images/create-linode-from-image.spec.ts` in another

```
REACT_APP_LOGIN_ROOT
REACT_APP_API_ROOT
REACT_APP_APP_ROOT
REACT_APP_LAUNCH_DARKLY_ID
REACT_APP_CLIENT_ID
MANAGER_OAUTH
```


patthiel:
we should also add a test for navigating to the create page with the query params for the specific image.  

https://cloud.linode.com/linodes/create/?type=Images&imageID=private/$IMAGE_ID

This flow is tripped when selecting the action menu "Deploy new Linode" from an image row item on the `/images` page

packages/manager/cypress/integration/images/create-linode-from-image.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    cy.intercept('GET', `*/linode/instances/${linodeId}`, (req) => {
+      req.reply(mockLinode);
+    }).as('mockLinodeResponse');
+
+    cy.visitWithLogin('/linodes/create?type=Images');
+    cy.wait('@mockImage');
+    cy.get('[data-qa-enhanced-select="Choose an image"]').within(() => {
+      containsClick('Choose an image');
+    });
+    fbtClick(imageLabel);
```
patthiel, line 48: Assert that the visual next to the select appears as well? 

Screen Shot 2021-07-20 at 3 22 23 PM

patthiel:
I know it seems like this test is limited, but it does test a critical path linode create flow within the manager. 

Right now we don't have a specific end-to-end test in the API to Create a Linode from a Machine Image, which we should prioritize adding (we do have tests for uploading machine images and post-processing them). However, I think it would be of greater value to test this at the API layer than at the Cloud Manager layer.

tiffwong (commented):
One of the tests failed for me
Screen Shot 2021-07-27 at 4 36 03 PM

tiffwong (approved):
Increasing the timeout fixed it, great job!

packages/manager/cypress/integration/images/smoke-create-image.spec.ts
```diff
@@ -21,7 +21,7 @@ describe('create image', () => {
     cy.intercept('POST', '*/images', (req) => {
       req.reply(200);
     }).as('postImages');
-    createLinode().then((linode) => {
+    createLinode().then((linode: any) => {
```
DevDW, line 24: What are your thoughts on using more precise types in these sorts of tests? Or does it not offer many (or any) benefits?

DevDW (approved):
LGTM, didn't observe any breaks from the package upgrades and the tests passed

packages/manager/cypress/integration/images/smoke-create-image.spec.ts
ndukatz: You're probably right. I set to any to avoid some errors when grabbing the label or id. but the Linode type would probably work better, right?

PR #7775, opened 2021-07-19
Log Google Pay Errors to Sentry

HanaXu:
## Description
Logs errors relating to Google Pay to Sentry.


bnussman:
I'm not exactly sure about this, but I wonder if passing `error` as the first parameter would benefit the sentry view of an error. 

so maybe this format for capturing exceptions
```ts
reportException(error, {
  message: errorMsg
});
```

DevDW (approved):
LGTM -- as a somewhat tangential question, maybe I overlooked it in a different file but is the `warning` variant handled for anywhere in the logic?

https://github.com/linode/manager/blob/2bde018b394184ea65617cf7e1b470c62d8f4102/packages/manager/src/features/Billing/Providers/GooglePay.ts#L84

bnussman (approved):
Looks good! When we deploy, we should watch the Sentry logs to make sure this does not flood Senty with too many errors at any point.

DevDW (approved):
LGTM, let's see how any errors look when they come in and assess if we want to keep this order of arguments for `reportException` or revert them back.

PR #7777, opened 2021-07-20
M3-5319: Make sure Bare Metal plans appear for accounts w/ appropriate tag

DevDW:
## Description
When pointed at the prod API, Cloud is not displaying Bare Metal plans, even for accounts properly tagged for Bare Metal. This is occurring because of caching logic in `linodeType.requests.ts`.

This PR removes that caching logic. I think this path is preferable to requesting types upon hitting the Linode Create page because `requestTypes()` only runs once as a secondary request in `AuthenticationWrapper.tsx`.

## How to test
Point at the prod API; accounts not tagged for Bare Metal should not see Bare Metal plans in the "Linode Plan" panel of the Create flow. Accounts tagged for Bare Metal should see the plans.


johnwcallahan (commented):
The changes to `linodeTypes.requests.ts` look good.

We'll also want to remove the linode types endpoint from buildRequests.js and the types.json file (which means we'll have to adjust a few places around the app that use it).

HanaXu (approved):
LGTM!

DevDW:
@HanaXu that review dismissal was due to an automatic setting for merging into `staging` 😕

HanaXu:
@DevDW Oof, same for my PR. I just re-requested your review on that 👀

HanaXu (approved):

tiffwong (approved):
LGTM

PR #7778, opened 2021-07-20
M3-5191: Use React Query for Profile and Grants

bnussman:
## Description
- [x] Uses React Query for all profile related API fetching and store
- [x] Use a wrapper FC for using profile in class components instead of fetching data in `mapStateToProps`
- This PR required a lot of changes in general
  - I refactored some class components to functional components where I saw it would not be too difficult
  - This PR can have a big effect on users that are restricted and have limited grants

## How to test

- Test everything in Cloud Manager ☁️


DevDW (commented):
I have quite a bit of testing left to do across the app, but I noticed that if a restricted user has just `read_only` access for a Linode, on this branch the action buttons for that Linode aren't disabled on the Linodes landing page

HanaXu:
There were a lot of issues, but I think I finally merged develop in correctly after the 5th try :crossed_fingers:. This should be ready for review now

HanaXu (changes requested):
I'm also seeing the buttons not being disabled for restricted users with `read_only` access for a Linode. Other than that, things look fine to me! Just mainly have cleanup comments

This branch:
image
Remote dev:
image

The dark mode button text style for `images/create` seem to be off on this branch:

image

The following can be handled in a separate PR; We don't currently handle this consistently in Volumes, NodeBalancers, Firewalls, Images, Domains either:

image

When we disable the buttons, we should display an error notice similar to `images/create` so users know why the buttons are disabled.

image

We should do the same in `images/upload` and disable the input fields as well. And display a similar Notice error in `account/billing` and disable the buttons there for users with `read_only` access

DevDW (commented):
Going to revisit this in-depth tomorrow, but on a restricted account I observed calls to `/account` and `/settings` being fired off every few seconds. 🤔  We'll want to get to the bottom of that

bnussman:
ImagesDrawer.tsx needs some attention. Unable to edit a description of an Image.

DevDW (commented):
This happens on prod as well, but when a StackScript is created from a restricted account, it can't be edited, etc. until the page is refreshed.

Prod behavior for a restricted user navigating to `/images/create` is that they can't even get there; this branch improves upon by that by allowing navigation to it and disabling interaction with the contents of the "Capture Image" tab. We should give the "Upload Image" tab similar treatment by disabling all of the fields.

bnussman:
@HanaXu @DevDW  Addressed yall's feedback. I definitely would like to work on a follow up PR down the road that improves consistency for restricted users / users with only some grants. For now, trying to make sure this PR has no regressions in comparison to prod.

johnwcallahan (commented):
I've reviewed the bottom half of the files so far.

In general I'm a little hesitant by the pattern of getting query data in `mapStateToProps`. It feels anti-pattern-y to me, though it seems to work to the extent that the function is run with the query is re-fetched. One option would be to build a wrapper FC that uses the useProfile hook and injects the data into class components.

DevDW (commented):
Gave this another lookover and it's looking good to me. If you're looking to do that wrapper component in this PR, I'll hold off the approval until that's in so I can test that

tiffwong (approved):
Everything works the same as prod for me 👍

packages/manager/src/queries/accountSettings.ts
```diff
@@ -25,13 +25,11 @@ export const useMutateAccountSettings = () => {
   );
 };
 
-export const isManaged = Boolean(
-  queryClient.getQueryData<AccountSettings>(queryKey)?.managed
-);
+export const isManaged = () =>
+  Boolean(queryClient.getQueryData<AccountSettings>(queryKey)?.managed);
```
codyfinn, line 29: Is there a reason this was changed to a function?

packages/manager/src/queries/accountSettings.ts
```diff
@@ -25,13 +25,11 @@ export const useMutateAccountSettings = () => {
...
 
-export const accountBackupsEnabled = Boolean(
-  queryClient.getQueryData<AccountSettings>(queryKey)?.backups_enabled
-);
+export const accountBackupsEnabled = () =>
+  Boolean(queryClient.getQueryData<AccountSettings>(queryKey)?.backups_enabled);
```
codyfinn, line 32: Is there a reason this was changed to a function?

codyfinn (approved):
Looks good to me! Nice job!

HanaXu (approved):
LGTM! 👍

DevDW (approved):
Great work! I didn't encounter any new issues in this round of testing.

I investigated the repeated API calls for restricted users in greater depth. The cause of what is bothering me there seems to be the `refetchOnWindowFocus` setting. It is true by default. So for restricted users, since the queries go stale after 3 attempts, the refetching happens when the window is brought back into focus.

Maybe it is worth exploring in a follow-up the extent to which we can set `refetchOnWindowFocus` to false if it is a restricted user? Endpoints of particular note from what I saw are `/account`, `/settings`, and `/payment-methods`. If it is explored further, we might be looking at something like the below, for example (I just did this to see if the general approach would work; refinements can be made):

Screen Shot 2021-08-10 at 5 15 59 PM

PR #7779, opened 2021-07-20
M3-5301: Replace sw-precache & sw-precache-webpack-plugin with workbox-webpack-plugin

DevDW:
## Description
Replace [sw-precache](https://github.com/GoogleChromeLabs/sw-precache) and [sw-precache-webpack-plugin](https://github.com/goldhand/sw-precache-webpack-plugin) due to a deep dependency and both packages being either deprecated or no longer maintained.

Info on migrating to Workbox from those packages: https://developers.google.com/web/tools/workbox/guides/migrations/migrate-from-sw

From the above source,

> `workbox-webpack-plugin` integrates directly with the webpack build process and "knows" about all of the assets generated by a given compilation. This means that, for many use cases, you can rely on the default behavior of `workbox-webpack-plugin` without additional configuration, and get an equivalent service worker to what `sw-precache-webpack-plugin` provides.

For info on config options for `GenerateSW()` from `Workbox`: https://developers.google.com/web/tools/workbox/reference-docs/latest/module-workbox-webpack-plugin.GenerateSW

## How to test
Ensure that nothing is broken in the app and its ability to compile.

To those ends, `yarn up` should be successful, and running `npm run build` after that should also be successful.

If you turn on the Mock Service Worker and go to the "Application" tab in Chrome DevTools, you should see an entry for the MSW.


ndukatz (approved):
Looks good and passed all E2E

johnwcallahan (commented):
Looks good pending the registration of the service worker which we talked about offline. Nice work.

tiffwong (approved):
Everything works as normal for me! 👍

bnussman (approved):
Looks good!

PR #7782, opened 2021-07-21
M3-5329: Refresh Users Table on Permission Change

bnussman:
## Description

Minor change that improves user experience. If a user grants full permissions to another user on their account, we need to refetch the users list so it us up to date. 

## How to test

- Go to `/account/users`
- Find a user that is not you and click `User Permissions`
- Toggle `Full Account Access` and make sure changes are reflected when you navigate back to `/account/users`


tiffwong (commented):
It might be related to my internet speed but `/users` page isn't refreshing for me after I change the permissions in another tab 🤔

bnussman:
@tiffwong We can't really do anything about updating programmatically across different tabs unfortunately. This only works within the same tab.

DevDW (approved):

ndukatz (approved):
works well

tiffwong (approved):

PR #7790, opened 2021-07-23
[Demo feedback] M3-5332: Billing UI changes

DevDW:
## Description
- [x] Add "Make a Payment" button to the Billing Summary
Screen Shot 2021-07-23 at 4 54 47 PM
<hr>

- [x] Add an instructional, second sentence to notification in the drawer indicating to users what happens when the notification is clicked
Screen Shot 2021-07-23 at 4 55 00 PM

Open to suggested revisions for the language of that second sentence.

### DIFF
--- a/packages/manager/src/components/LandingHeader/LandingHeader.tsx
+++ b/packages/manager/src/components/LandingHeader/LandingHeader.tsx
@@ -25,6 +25,7 @@ export interface Props extends Omit<HeaderProps, 'actions'> {
   breadcrumbProps?: BreadcrumbProps;
 }
 
+export const defaultCreateButtonWidth = 152;
 /**
  * This component is essentially a variant of the more abstract EntityHeader
  * component, included as its own component because it will be used in
@@ -44,8 +45,6 @@ export const LandingHeader: React.FC<Props> = (props) => {
     loading,
     breadcrumbProps,
   } = props;
-
-  const defaultCreateButtonWidth = 152;
 
   const actions = React.useMemo(
     () => (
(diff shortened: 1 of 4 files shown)


HanaXu (commented):
Not sure how I feel about the blue `Make a Payment` button and its placement.

For the notification drawer, I was under the impression that we wanted two links. One would be the red link: `You have an outstanding balance` linking to the Account page. And next to it, a blue link: "`Make a Payment` now to avoid service disruption." linking to the Make a Payment drawer

Would like to hear Matthew's thoughts on how we should display this too

tiffwong (commented):
I think this is what it shows when a payment is not overdue:
Screen Shot 2021-07-26 at 1 53 06 PM

We should probably change the text below the button to something different from the button and if we do decide to keep the button, I think aligned to the bottom right will be more consistent with where we're headed for the rest of the site

johnwcallahan (changes requested):
> Not sure how I feel about the blue `Make a Payment` button and its placement.
> 
> For the notification drawer, I was under the impression that we wanted two links. One would be the red link: `You have an outstanding balance` linking to the Account page. And next to it, a blue link: "`Make a Payment` now to avoid service disruption." linking to the Make a Payment drawer
> 
> Would like to hear Matthew's thoughts on how we should display this too

This is correct; the feedback was to include a call to action in the Notification Drawer. The Billing Summary panel can remain unchanged, as there is already a call to action (the "Make a payment immediately to avoid service disruption" button).

Originally a "Make Payment" button in the Notification drawer was suggested, but I agree with @HanaXu that a link might be more appropriate. Please sync with Matthew to come to a decision and update this PR accordingly.

DevDW:
Closed in favor of https://github.com/linode/manager/pull/7798.

PR #7791, opened 2021-07-26
Update Google Pay notice icon

HanaXu:
## Description
Updates the Google Pay Notice Icon to use the mark per the guidelines

## How to test
Check the Payment Methods paper on an account that has not dismissed the notification and does not have Google Pay added as a payment method


bnussman (dismissed):
Looks good. 

Only minor things is that the Icon is not a static size so on a small device, the icon shrinks to a point where it seems a little too small. 

Screen Shot 2021-07-26 at 11 19 35 AM

DevDW (approved):
I agree with Banks re: the sizing. Also, as the viewport shrinks, we hit a couple of situations where the text gets pretty close to the "X":

Screen Shot 2021-07-26 at 11 37 57 AM

Maybe some styling can be applied to the `<Link />` that "Add Google Pay" is in so that it'll be on its own line at certain breakpoints.

packages/manager/src/features/Billing/BillingPanels/PaymentInfoPanel/PaymentInformation.tsx
```diff
@@ -173,9 +182,9 @@ const PaymentInformation: React.FC<Props> = (props) => {
             className={classes.googlePayNoticeContainer}
             preferenceKey="google-pay-available-notification"
           >
-            <Box display="flex" alignItems="center">
-              <GooglePay width={16} height={16} />
-              <Typography className={classes.googlePayNotice}>
+            <Box display="flex" alignItems="flex-start">
```
bnussman, line 185: I think `alignItems="center"` should be kept so we get this

Screen Shot 2021-07-26 at 12 34 03 PM

instead of
Screen Shot 2021-07-26 at 12 34 16 PM

bnussman (approved):

DevDW (approved):
LGTM! 🚢

PR #7795, opened 2021-07-26
M3-5277: My Profile / Account dropdown changes

tiffwong:
## Description

Remove the hover state for the button in the top nav and adjust the link styles once the menu is expanded.

## How to test

- The background for My Profile / Account in the top nav should no longer change.
- The links inside the menu should be 14px and no longer have a hover state as well unless the user is navigating through it via a keyboard.


johnwcallahan (approved):
Looks great, I like the smaller font size.

codyfinn (approved):
Yeah I agree with John the smaller font size looks a lot better

PR #7796, opened 2021-07-27
M3-5244: Standardize secret token modals

HanaXu:
## Description
Standardize secret token modals to one design (the tokens in the screenshot were deleted dw)

image
image
image

## How to test
The new dialog should display a warning that the secret token will only be shown once. The keys/secret token should be in a copyable text field. The only way to dismiss the dialog is via the `x` button on the top right (clicking on the backdrop or pressing `esc` should not dismiss the dialog).
- Create a Personal Access Token in `profile/tokens`
- Create an OAuth App in `profile/clients`
- Create an Object Storage access key in `object-storage/access-keys`


tiffwong (commented):
Can you reduce the whitespace between the dialog title and the warning message?
Screen Shot 2021-07-27 at 4 40 51 PM

Also should the secret token names (ie. "Client Secret", "Personal Access Token", "Secret Key") be lowercase?

codyfinn (approved):

packages/manager/src/features/Profile/SecretTokenDialog/SecretTokenDialog.tsx
```diff
@@ -0,0 +1,84 @@
...
+      disableEscapeKeyDown
+      maxWidth="sm"
+    >
+      <Notice
+        spacingTop={8}
+        warning
+        className={classes.noticeText}
+        text={`For security purposes, we can only display your ${
+          objectStorageKey ? 'secret key' : title.toLowerCase()
+        } once, after which it can't be recovered. Be sure to keep it in a safe place.`}
```
DevDW, line 49: Totally optional, but I'd store this entire string in a variable up around line 33 and then pass it here

packages/manager/src/features/Profile/OAuthClients/Modals.tsx
```diff
@@ -23,44 +22,19 @@ interface Props {
...
-          {`Here is your client secret! Store it securely, as it won't be shown again.`}
-        </Typography>
-        <Notice
-          warning
-          text={props.secret}
-          spacingTop={24}
-          className={classes.key}
-        />
-      </ConfirmationDialog>
+        value={props.secret}
```
DevDW, line 36: I think we can just destructure this (and more generally destructure more of these props throughout the file, like `closeDialogs` is destructured already but the variable isn't used consistently)

DevDW (approved):
Looks good, just a couple of minor comments

PR #7800, opened 2021-07-27
M3-5168: Fix "Unknown Plan" and "Unknown Image" in Search

bnussman:
## Description

- Fixes `Unknown Image` and `Unknown Type` in Cloud Manager's search.
  - Fixes plan names and image names showing for large customers
  - Fixed `Unknown Image` for LKE Linodes

Before(s):
Screen Shot 2021-07-27 at 4 21 29 PM
Screen Shot 2021-07-27 at 4 22 06 PM
Screen Shot 2021-07-27 at 4 31 02 PM

After(s):

Screen Shot 2021-07-27 at 4 21 46 PM
Screen Shot 2021-07-27 at 4 21 57 PM

## How to test

- Test searching on accounts
  - Test on Large and Small accounts
  - Try searching for different types of entities
  - Ensure no functionality has been lost as compared to `cloud.linode.com`


codyfinn (approved):

packages/manager/src/features/Search/useAPISearch.tsx
```diff
@@ -39,21 +36,22 @@ export const useAPISearch = (): Search => {
           combinedResults: [],
         });
       }
+      const resources = store.getState().__resources;
+
+      const types = resources.types.entities as LinodeType[];
+      const images = resources.images.itemsById;
```
DevDW, line 42: From testing so far I haven't observed any issues from not filtering these for only public images like it was before, but I'm not entirely clear on the original intent behind that filter

DevDW (commented):
Looks like the fix works 👌

packages/manager/src/features/Search/useAPISearch.tsx
bnussman: Looked more into this. I wanted to include private images in hopes to display the private image on Linodes that were created from custom images, but I found out that the `image` field on a Linode created from a custom image is `null` so Including private images seems to be useless since we can't link them up

tiffwong (approved):

PR #7801, opened 2021-07-27
Handle API warning message when making a payment

HanaXu:
## Description
Some API warning messages do not have a `detail` field, which was causing the app to crash when trying to make a credit card payment. This fixes that and also adds warning handling for Google Pay payments

## How to test
See slack

### DIFF
--- a/packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/PaymentDrawer.tsx
+++ b/packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/PaymentDrawer.tsx
@@ -133,7 +133,7 @@ export const PaymentDrawer: React.FC<Props> = (props) => {
     paymentWasMade = false,
     warnings = undefined
   ) => {
-    if (paymentWasMade) {
+    if (paymentWasMade && !warnings) {
       enqueueSnackbar(message, {
         variant: 'success',
       });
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/PaymentDrawer.tsx
```diff
@@ -277,7 +277,7 @@ const Warning: React.FC<WarningProps> = (props) => {
   /** The most common API warning includes "please open a Support ticket",
    * which we'd like to be a link.
    */
-  const ticketLink = warning.detail.match(/open a support ticket\./i) ? (
+  const ticketLink = warning.detail?.match(/open a support ticket\./i) ? (
```
bnussman, line 280: Should we update the match to also match what Nick sent? User's might want to take action on the example warning he sent

HanaXu: I looked into doing that but the code started getting a bit ugly because of inconsistencies between the wording `open a support ticket.` vs `contact support ...` and where the wording appears (title vs detail). Talked to Nick about possibly standardizing this and they'll look into it when they have the time (ARB-2650)

bnussman: Awesome, thanks for doing all that

DevDW (commented):
This is working; I understand this is literally a warning from the API, but since it indicates an attempted payment didn't go through as expected by the customer, I wonder if the `error` variant of the `<Notice />` to get the red left border instead of the yellow one would be better since it gives a stronger indicator the payment didn't go through?

HanaXu:
@DevDW It's possible for a payment to successfully go through, but to have a warning where "resources fail to unsuspend" so I think we should just handle warnings as warnings

bnussman (approved):
Nice job. I wonder if we need to keep the `paymentWasMade` parameter around? Does that value ever get set to false?

DevDW (approved):
LGTM, I agree that `paymentWasMade` can be removed if it is never being set to false anywhere

PR #7802, opened 2021-07-27
M3-5248: Change look of Selectable Linode Plan to fix wrapping

bnussman:
## Description

- Refactors `SelectionCard.tsx` to be a functional component
- Removes the check icon from the selected plan and changes selected state
  - The goal here is to make things like `2GB RAM` not wrap so it is more clear and readable

**The new selected state:**

https://user-images.githubusercontent.com/6440455/127504305-022e998a-fd4b-454b-bd82-6ad2c61f2fcd.mov

## How to test

- Test selecting a Linode plan at `/linodes/create` on a small screen size

### DIFF
--- a/packages/manager/src/components/SelectionCard/SelectionCard.tsx
+++ b/packages/manager/src/components/SelectionCard/SelectionCard.tsx
@@ -217,19 +217,6 @@ const SelectionCard: React.FC<CombinedProps> = (props) => {
     }
   };
 
-  const content = () => {
-    return (
-      <CardBase
-        heading={heading}
-        renderIcon={renderIcon}
-        subheadings={subheadings}
-        fullWidth={variant === 'quantityCheck' || variant === 'selectable'}
-      >
-        {renderVariant()}
-      </CardBase>
-    );
-  };
-
   const renderVariant = () => {
     switch (variant) {
       case 'info':
@@ -262,14 +249,11 @@ const SelectionCard: React.FC<CombinedProps> = (props) => {
             )}
           </Grid>
         );
-      case 'selectable':
-        return null;
       /**
        * The vast majority of these components use the check variant, so
        * keep that as the default case.
        */
       case 'check':
-      default:
         return (
           <Grid
             item
@@ -284,10 +268,24 @@ const SelectionCard: React.FC<CombinedProps> = (props) => {
             )}
           </Grid>
         );
+      case 'selectable':
+      default:
+        return null;
     }
   };
 
-  const cardGrid = () => (
+  const content = (
+    <CardBase
+      heading={heading}
+      renderIcon={renderIcon}
+      subheadings={subheadings}
+      fullWidth={variant === 'quantityCheck' || variant === 'selectable'}
+    >
+      {renderVariant()}
+    </CardBase>
+  );
+
+  const cardGrid = (
     <Grid
       item
       xs={12}
@@ -298,8 +296,8 @@ const SelectionCard: React.FC<CombinedProps> = (props) => {
       className={classNames(
         {
           [classes.root]: true,
-          checked: checked === true,
-          [classes.disabled]: disabled === true,
+          checked,
+          [classes.disabled]: disabled,
           [classes.showCursor]: onClick && !disabled,
           selectionCard: true,
         },
@@ -309,17 +307,19 @@ const SelectionCard: React.FC<CombinedProps> = (props) => {
       onKeyPress={handleKeyPress}
       data-qa-selection-card
     >
-      {content()}
+      {content}
     </Grid>
   );
 
-  return tooltip ? (
-    <Tooltip title={tooltip} placement={'top'}>
-      {cardGrid()}
-    </Tooltip>
-  ) : (
-    cardGrid()
-  );
+  if (tooltip) {
+    return (
+      <Tooltip title={tooltip} placement="top">
+        {cardGrid}
+      </Tooltip>
+    );
+  }
+
+  return cardGrid;
 };
 
 export default React.memo(SelectionCard);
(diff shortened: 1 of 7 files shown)


packages/manager/src/components/SelectionCard/SelectionCard.tsx
```diff
@@ -3,170 +3,153 @@ import * as classNames from 'classnames';
...
+    '& svg, & span': {
+      fontSize: '32px',
+      color: '#939598',
     },
-    subheading: {
-      fontSize: '0.875rem',
-      color: theme.palette.text.primary,
+    '& img': {
+      maxHeight: '32px',
+      maxWidth: '32px',
```
tiffwong, line 57: Instead of `'32px'`, you can actually just write these as `32`

packages/manager/src/components/SelectionCard/SelectionCard.tsx
```diff
@@ -3,170 +3,153 @@ import * as classNames from 'classnames';
...
-      border: '1px solid ' + `${theme.bg.main}`,
-      margin: 0,
-      padding: '0 !important',
-      transition: `
+  },
+  checked: {
+    display: 'flex',
+    animation: '$fadeIn 225ms ease-in-out forwards 10ms',
+    '& svg': {
+      borderRadius: '16px',
```
tiffwong, line 64: Same thing regarding `'16px'` here

packages/manager/src/components/SelectionCard/SelectionCard.tsx
```diff
@@ -3,170 +3,153 @@ import * as classNames from 'classnames';
...
+      border: '1px solid',
+      borderColor: theme.palette.primary.main,
```
tiffwong, line 66: This can be rewritten as "border: `1px solid ${theme.palette.primary.main}`"

tiffwong: (but make sure the line is surrounded by `)

packages/manager/src/components/SelectionCard/SelectionCard.tsx
```diff
@@ -3,170 +3,153 @@ import * as classNames from 'classnames';
...
+  subheading: {
+    fontSize: '0.875rem',
+    color: theme.palette.text.primary,
+  },
+  innerGrid: {
+    width: '100%',
+    height: '100%',
+    minHeight: 70,
+    backgroundColor: theme.bg.offWhiteDT,
+    border: '1px solid ' + `${theme.bg.main}`,
```
tiffwong, line 98: This can also be wrapped entirely with ticks instead of concatenating them

tiffwong (changes requested):
Looks great! Left a few cleanup items and the only other thing that needs to be changed is dark mode
Screen Shot 2021-07-28 at 6 19 25 PM

bnussman:
Changed the selected colors, take a look at the video in the PR description

packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
```diff
@@ -271,7 +271,7 @@ export class SelectPlanPanel extends React.Component<CombinedProps> {
             subheadings={type.subHeadings}
             disabled={planTooSmall || isSamePlan || disabled || isDisabledClass}
             tooltip={tooltip}
-            variant={'check'}
+            variant="selectable"
```
bnussman, line 274: Should I make this change on other selectable places like `/linodes/create?type=Clone%20Linode`?

tiffwong: That would be great!

tiffwong (approved):
Looks much better now! 🎉

packages/manager/src/components/SelectionCard/SelectionCard.tsx
```diff
@@ -298,52 +287,39 @@ class SelectionCard extends React.PureComponent<CombinedProps, {}> {
...
+      xs={12}
+      sm={6}
+      lg={4}
+      xl={3}
+      tabIndex={0}
+      className={classNames(
+        {
+          [classes.root]: true,
+          checked: checked === true,
+          [classes.disabled]: disabled === true,
```
HanaXu, line 302: This can be
```
checked,
[classes.disabled]: disabled
```

packages/manager/src/components/SelectionCard/SelectionCard.tsx
```diff
@@ -298,52 +287,39 @@ class SelectionCard extends React.PureComponent<CombinedProps, {}> {
...
+          [classes.showCursor]: onClick && !disabled,
+          selectionCard: true,
```
HanaXu, line 304: Is `selectionCard: true` needed?

packages/manager/src/components/SelectionCard/SelectionCard.tsx
```diff
@@ -298,52 +287,39 @@ class SelectionCard extends React.PureComponent<CombinedProps, {}> {
...
+        },
+        className
+      )}
+      onClick={handleClick}
+      onKeyPress={handleKeyPress}
+      data-qa-selection-card
+    >
+      {content()}
```
HanaXu, line 312: I don't think `content()`, `renderVariant()` or `cardGrid()` need to be functions. What about turning `renderVariant` into an object that returns jsx, replace `renderVariant()` in `content()` with a ternary, etc?

HanaXu (commented):
Changes look good! Left some cleanup comments

DevDW (approved):
I agree with @HanaXu that those handful of functions can be refactored to not be functions, but everything else looks good 👍

packages/manager/src/components/SelectionCard/SelectionCard.tsx
```diff
@@ -280,7 +254,6 @@ class SelectionCard extends React.PureComponent<CombinedProps, {}> {
        * keep that as the default case.
        */
       case 'check':
-      default:
```
HanaXu, line 283: Let's remove the comment above the `check` case since it's no longer being used as the default case

HanaXu (approved):
LGTM!

PR #7805, opened 2021-07-29
M3-5349  E2E update test to reflect copy update

ndukatz:
## Description
This change was needed because of some recent copy changes to this page

**What does this PR do?**
just a small access-keys e2e test update

## How to test
- `yarn up` in one terminal
- `yarn cy:e2e -s ./cypress/integration/objectStorage/access-keys.spec.ts` in another

## Make sure your the mock service worker is off and your .env contains:
```
REACT_APP_LOGIN_ROOT
REACT_APP_API_ROOT
REACT_APP_APP_ROOT
REACT_APP_LAUNCH_DARKLY_ID
REACT_APP_CLIENT_ID
MANAGER_OAUTH
```

### DIFF
--- a/packages/manager/cypress/integration/objectStorage/access-keys.spec.ts
+++ b/packages/manager/cypress/integration/objectStorage/access-keys.spec.ts
@@ -1,9 +1,5 @@
 import { makeTestLabel } from '../../support/api/common';
-import {
-  createBucket,
-  deleteAllTestAccessKeys,
-  deleteBucketByLabel,
-} from '../../support/api/objectStorage';
+import { createBucket } from '../../support/api/objectStorage';
 import {
   containsVisible,
   fbtClick,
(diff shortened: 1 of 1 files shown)


packages/manager/cypress/integration/objectStorage/access-keys.spec.ts
```diff
@@ -33,9 +29,9 @@ describe('access keys', () => {
       fbtClick('Submit');
       cy.wait('@createAccessKey');
       fbtVisible(
-        'Your keys have been generated. For security purposes, we can only display your Secret Key once, after which it can’t be recovered.'
+        "For security purposes, we can only display your secret key once, after which it can't be recovered. Be sure to keep it in a safe place."
```
patthiel, line 32: I don't think we should be this specific about text assertions, it will lead you down the path of having to update the text any time we update copy. Is it possible to change this reference to be dynamic and get the expected text from the code? https://github.com/linode/manager/blob/5e9440e621c8a44073b7af851060ef7c82ea16c1/packages/manager/src/features/Profile/SecretTokenDialog/SecretTokenDialog.tsx#L47-L49

ndukatz:
For now we can keep the copy hard coded. Later we can work on getting text like this more dynamically.

tiffwong (approved):

DevDW (approved):
When I first ran the test I got this:

Screen Shot 2021-08-02 at 1 49 52 PM

After `yarn up`'ing again, though, I ran the tests a few more times and didn't see the error again.

PR #7812, opened 2021-07-30
M3-5355: Fix users with active promotions unable unable to edit their account

bnussman:
## Description

- Fix users unable to edit their accounts (specify users with active promotions)
- This was caused by #7737
  - When the POST goes through to update a users account, I passed all information returned by `/account`. We only need to send data editable in the Update Account drawer. I used `pick()` in this to ensure we exclude other account data we POST to the api

## How to test
- Test updating account information (accounts with and without active promotions)


HanaXu (approved):
Fix works for me on an account with and without promotions 👍

packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/UpdateContactInformationForm/UpdateContactInformationForm.tsx
```diff
@@ -89,13 +89,26 @@ class UpdateContactInformationForm extends React.Component<
   async componentDidMount() {
     const account = queryClient.getQueryData<Account>('account');
 
-    const accountWithoutCard: Partial<Account> = {
-      ...account,
-      credit_card: undefined,
-    };
+    const cleanAccount = pick(
```
DevDW, line 92: Perhaps this variable can be renamed to something clearer? These are essentially the editable billing contact info fields, so maybe `editableBillingContactFields` or something similar? Maybe even a comment explaining why this has to be done would be helpful for the future.

DevDW (approved):
Works for me on accounts with and without promos applied. Left one comment

PR #7814, opened 2021-08-02
M3-5331: Remove redundant headers

tiffwong:
## Description

Remove redundant headers from the Linode Detail tab pages and the settings related pages.

After removing the headers, there are several pages (ie. `/account/users`, `/profile/keys`, and `/linodes/id/configurations`) that are left with only a button in the header region. We decided to generally leave the buttons as they are except we removed the white background and aligned the button to the right. 

## How to test

Check the page headers and make sure the ones remaining add some context or information rather than repeating the tab label.

### DIFF
--- a/packages/manager/src/features/Profile/LishSettings/LishSettings.tsx
+++ b/packages/manager/src/features/Profile/LishSettings/LishSettings.tsx
@@ -28,35 +28,19 @@ import getAPIErrorFor from 'src/utilities/getAPIErrorFor';
 import scrollErrorIntoView from 'src/utilities/scrollErrorIntoView';
 
 type ClassNames =
-  | 'root'
-  | 'title'
   | 'intro'
   | 'modeControl'
   | 'sshWrap'
   | 'keyTextarea'
-  | 'image'
-  | 'addNew'
-  | 'remove'
-  | 'button';
+  | 'addNew';
 
 const styles = (theme: Theme) =>
   createStyles({
-    root: {
-      padding: theme.spacing(3),
-      paddingBottom: theme.spacing(3),
-    },
-    title: {
-      marginBottom: theme.spacing(2),
-    },
     intro: {
       marginBottom: theme.spacing(2),
     },
     modeControl: {
       display: 'flex',
-    },
-    image: {
-      display: 'flex',
-      flexWrap: 'wrap',
     },
     addNew: {
       ...theme.applyLinkStyles,
@@ -69,13 +53,6 @@ const styles = (theme: Theme) =>
       [theme.breakpoints.up('md')]: {
         minWidth: 415,
       },
-    },
-    remove: {
-      ...theme.applyLinkStyles,
-    },
-    button: {
-      margin: 0,
-      padding: 0,
     },
   });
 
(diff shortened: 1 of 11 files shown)


packages/manager/src/features/Users/UsersLanding.tsx
```diff
@@ -234,45 +229,38 @@ const UsersLanding: React.FC<Props> = (props) => {
           text={`Error when deleting user, please try again later`}
         />
       )}
-      <div className={classes.root}>
```
tiffwong, line 237: The only change here is removing the surrounding `<div>` since the `root` class is no longer necessary.

packages/manager/src/features/linodes/LinodesDetail/LinodeAdvanced/LinodeConfigs.tsx
```diff
@@ -529,10 +513,6 @@ class LinodeConfigs extends React.Component<CombinedProps, State> {
 
 const styled = withStyles(styles);
 
-const errorBoundary = PanelErrorBoundary({
-  heading: 'Advanced Configurations',
-});
-
```
tiffwong, line 535: This seems to function as a backup header and was displayed when I removed the original header. Not sure if there's any other purpose for this though 🤔

DevDW: It seems like `PanelErrorBoundary` is used throughout the codebase to return an accordion with an error if something wrong happens with a regular component and its contents, but I'm not convinced it is serving an important role in Cloud Manager currently 🧐

DevDW (approved):
LGTM, what do you think about making those now-headerless buttons vertically centered between the bottom border of the tabs and the top border of the table?

example of that case:

Screen Shot 2021-08-02 at 3 51 36 PM

bnussman:
Would `Settings` be something we would also want to remove?
Screen Shot 2021-08-04 at 10 21 03 AM

HanaXu (approved):
LGTM, we'll probably want to also remove the redundant headers from `profile/clients` and `profile/lish`

image

image

packages/manager/src/features/Profile/LishSettings/LishSettings.tsx
```diff
@@ -199,47 +175,48 @@ class LishSettings extends React.Component<CombinedProps, State> {
...
+                      buttonType="secondary"
+                      compact
                       onClick={this.onPublicKeyRemove(idx)}
-                      className={classes.remove}
                       data-qa-remove
                     >
                       Remove
-                    </button>
-                  )}
+                    </Button>
```
tiffwong, line 187: Used the `<Button>` component instead since it's more consistent with our usages of buttons vs links

packages/manager/src/features/Profile/LishSettings/LishSettings.tsx
```diff
@@ -199,47 +175,48 @@ class LishSettings extends React.Component<CombinedProps, State> {
...
+              <Button
+                buttonType="secondary"
+                compact
                 onClick={this.addSSHPublicKeyField}
-                className={classes.addNew}
               >
                 Add SSH Public Key
-              </button>
-            </React.Fragment>
+              </Button>
```
tiffwong, line 201: Same case here regarding `<Button>`

PR #7816, opened 2021-08-02
M3-5346: GooglePay display in Safari

tiffwong:
## Description

Fixes the vertical stretching and missing icon occurring in Payment Methods for GooglePay in Safari and iOS Chrome.

Screen Shot 2021-08-02 at 4 49 03 PM

Turns out Safari's default setting for `alignItems` is `stretch` so defining it as `flex-start` fixes the issue. A similar thing is happening to the GooglePay icon where Safari needs the height or width of the svg in order to render it.

## How to test

Go to `accounts/billing` on different browsers and check the display for different resolutions.


DevDW (approved):
Looks good to me 🚢  -- tested in Safari, Firefox, and Chrome at different viewports

HanaXu (approved):
LGTM! 👌

PR #7817, opened 2021-08-03
M3-5353: Volume Create - remove checkout sidebar

tiffwong:
## Description

This PR removes the sidebar from the Volume Create page, moves the create button to below the form, and moves the pricing below the size textfield. The new storage options banner inside the Volume Create form is also changed to green.

Also fixed a minor bug in `Breadcrumb.tsx` where classes aren't actually being applied.

## How to test

- Make sure you can create a Volume via `/volumes/create` and from a Linode
- Check for any regressions in resizing a Volume


packages/manager/src/features/Volumes/VolumeDrawer/SizeField.tsx
```diff
@@ -47,7 +53,11 @@ const SizeField: React.FC<CombinedProps> = ({
         value={value}
         {...rest}
       />
-      <FormHelperText>The size of the new volume in GiB</FormHelperText>
+      <FormHelperText>
+        {resize || isFromLinode
+          ? 'The size of the new volume in GiB.'
+          : `${price}/month`}
```
DevDW, line 59: We should prepend a dollar sign here. It also feels like the pricing text is buried and easy to miss now compared to the previous placement. Maybe we should consider placing it right above the "Create Volume" button?

bnussman: I have the same feedback. `$` is needed and I was thinking price to the left of the `Create Volume` button.

tiffwong: Added the `$` but I discussed the issue with the pricing text location with Matthew and we decided to punt it for later since we'll have to be dealing with this issue when we remove the sidebar from the other create flows

DevDW (approved):
🚢

ndukatz:
Looks good! It still passes the standard volumes create flow automation without any changes to the test. 

I did notice that I was unable to add a tag when creating a volume from the linode page. This error pops up when clicking Create Volume:
Screen Shot 2021-08-05 at 5 30 47 PM

If there was no tag present, I got this message when trying to create volume from an Atlanta linode:
Screen Shot 2021-08-05 at 5 31 03 PM

tiffwong:
@ndukatz that seems to be happening for me in develop as well, I'll make a ticket for it

ndukatz (approved):

PR #7787, opened 2021-07-23
M3-5300 Update several dependencies

johnwcallahan:
## Description

This PR upgrades several dependencies to pick up security updates:

**prismjs**
- forced resolution to 1.24.1
- minor version jump

**glob-parent**
- forced resolution to 5.1.2
- ⚠️ major version jump (not a direct dependency)

**browserslist**
- forced resolution to 14.16.5
- minor version jump

**lodash**
- forced resolution to 4.17.19
- minor version jump

**immer**
- forced resolution to 8.0.4
- ⚠️ major version jump, code changes made accordingly

**trim**
- forced resolution to 0.0.3
- patch version jump

**ws**
- forced resolution to 7.5.3
- ⚠️ major version jump (not a direct dependency)

**merge**
- upgraded to 2.1.0 via npm-dts-webpack-plug upgrade
- ⚠️  major version jump (not a direct dependency)

**property-expr**
- upgrade to 2.0.4 via yup upgrade
- code changes made to fix breaking changes in yup

## How to test

Build, run, test the app. Try using the sdk/validation library separately. Run e2e tests and unit tests.

PR #7822, opened 2021-08-03
Fix: unable to add tags to Volumes during creation.

johnwcallahan:
## Description

This fixes a bug in develop where creating a new volume with a tag results in a field error: "Unable to tag Volume." 

This appears to have been introduced in the recent [Yup upgrade](https://github.com/linode/manager/pull/7787). I don't exactly know why, though. We've always been validating tags in the CreateVolumeSchema as `array().of(string())`, but we've been sending an array of `_Tag`, which are objects. Maybe yup wasn't validating properly beforehand. In any case, we should dig more into this issue because it may also be present elsewhere.

To fix this, I extended the schema in CreateVolumeForm.

## How to test

Create a Volume from the Create Volume page with a tag. Check that other validation still works as expected.


ndukatz (approved):
This resolves the issue. Thanks for the quick fix! The volumes test that found this is now passing as well.

johnwcallahan:
> * Domains, Firewalls, and maybe other entities don't allow you to add a tag until after the entity is created (should we allow this at some point?)

That's a good question. Some of these (like Domains) used to support tags at creation, but we took them out to streamline the creation flow.

PR #7825, opened 2021-08-04
M3-5347: Handle errors thrown when payment popup is closed or timesout

codyfinn:
## Description

This PR adds better error handling with respect to GooglePay errors. In particular on non-chrome browsers if a user closed the GooglePay pop-up window without authorizing a payment an error is thrown that was then logged into Sentry and caused a Toast notification to pop up that informed the user "Unable to add payment method". A similar error ocurred if the GooglePay pop-up was left up for more than 5 minutes. This should be handled in the same way. 

## How to test

Attempt to add a new payment method using GooglePay on a non-chrome browser. Close the pop-up window without authorizing payment. This should silently fail without informing the user. There should be no Toast notification or error logged to sentry. The same should be true if the pop-up is left open for more than 5 minutes.

### DIFF
--- a/packages/manager/src/features/Billing/Providers/GooglePay.ts
+++ b/packages/manager/src/features/Billing/Providers/GooglePay.ts
@@ -152,9 +152,6 @@ export const gPay = async (
     setProcessing(false);
   } catch (error) {
     setProcessing(false);
-    if (error.message && (error.message as string).includes('User closed')) {
-      return;
-    }
 
     if (error.statusCode === 'CANCELED') {
       return;
--- a/packages/manager/src/features/Billing/Providers/GooglePay.ts
+++ b/packages/manager/src/features/Billing/Providers/GooglePay.ts
@@ -156,6 +156,10 @@ export const gPay = async (
       return;
     }
 
+    if (error.statusCode === 'CANCELED') {
+      return;
+    }
+
     const errorMsg = isOneTimePayment
       ? 'Unable to complete Google Pay payment'
       : 'Unable to add payment method';


bnussman:
I think we also want to cover `DOMException: The "paymentrequest" event timed out after 5 minutes.` 

That gets thrown when a user idles in the Google popup for > 5 minutes and returns to Cloud Manager. We should not log this to Sentry and up to you if the error toast should happen.

packages/manager/src/features/Billing/Providers/GooglePay.ts
```diff
@@ -156,6 +156,10 @@ export const gPay = async (
       return;
```
bnussman, line 156: Up to you, but you could combine the two if statements

```ts
if (error.message && (error.message as string).includes('User closed') || error.statusCode === 'CANCELED') {
      return;
}
```

bnussman (approved):
Works for the cases we discussed and is super simple! 🚀

packages/manager/src/features/Billing/Providers/GooglePay.ts
codyfinn: I think ideally I would want to refactor this so that we don't return when we hit these types of errors, but was having a hard time keeping the code clean when doing that. I think keeping the errors separate and returning makes it easier to read for now.

HanaXu: We can also cover the time out here

```javascript
if (error.message && /user closed|timed out/i.test(error.message) || error.statusCode === 'CANCELED') {
      return;
}
```

HanaXu (approved):
I agree with @bnussman, we should address the time out before merging but otherwise LGTM 👍

packages/manager/src/features/Billing/Providers/GooglePay.ts
bnussman: Cool with me @codyfinn

bnussman:
> I agree with @bnussman, we should address the time out before merging but otherwise LGTM 👍

His code does account for the 5 minute time out, we verified it! Should be good to go!

HanaXu:
@bnussman Ah, you're right! Do we still need to check the error message in the if statement then since both closing and timeout have a statusCode of `CANCELED`? 🤔

PR #7826, opened 2021-08-04
M3 5339: Improve firewall error message for unsupported hosts

HanaXu:
## Description
Improve firewall error message when trying to add a firewall to a Linode on an unsupported host

- Remove the Id
- Make the Linode label a link to that Linode's detail page
- Remove unnecessary `open a Support ticket` text
    - Users can initiate inter-DC migration on their own. And in the event that this was a migration unrelated to one they queued up, there should already be a ticket open for it

Before:

image

After:

image

## How to test
Try to add a firewall to a Linode that is on an unsupported host (reach out for exact steps)

### DIFF
--- a/packages/manager/src/components/Notice/Notice.tsx
+++ b/packages/manager/src/components/Notice/Notice.tsx
@@ -14,7 +14,7 @@ import {
 import Typography, { TypographyProps } from 'src/components/core/Typography';
 import Grid, { GridProps } from 'src/components/Grid';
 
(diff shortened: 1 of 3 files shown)


ndukatz (approved):
Looks good! still passes firewall tests. Was able to see the message on prod-test-004 pointing to staging api

bnussman (approved):
Looks good to me **as long as we are 100% sure Linode on unsupported hosts have a pending migration.**

PR #7828, opened 2021-08-05
M3-5118: Linode Create flow filtering for Bare Metal plans

DevDW:
## Description
Bare Metal plans are not available across all regions presently. Like with GPU Linodes, this should be indicated in the Linode Create flow when a user goes to the Bare Metal tab for a region that those plans are not available in:

Screen Shot 2021-08-05 at 12 41 05 PM

Also added `doesRegionSupportFeature.ts`, which is more flexible than the separate files that check whether specific individual features are supported by a particular region or not, and eliminates the need for those files.

## How to test
Pointing to dev, begin the Linode Create flow. Select a region other than Newark and you should see the view above; select Newark and you should be able to proceed normally (full creation dependent on back-end considerations).

Confirm that the "Attach VLAN" section has not been adversely impacted by the small refactors that touched it.

`yarn test doesRegionSupportFeature`


codyfinn (approved):
🚢

bnussman (approved):
Looks good!

PR #7833, opened 2021-08-06
M3-5373: Remove workbox-webpack-plugin

DevDW:
## Description
This is essentially a rollback of https://github.com/linode/manager/pull/7779.

Although the app was working with Workbox, we want to do a bit more analysis and testing to see what exactly it is doing within the context of the app before releasing it into production. One risk with Workbox is that it can make debugging deployments more difficult.

Workbox will be explored further with a possible follow-up PR re-adding it.

## How to test
The app should compile and function properly, and no service workers should be registered.


ndukatz (approved):
Looks good. Ran all tests and all passed. As well as some other manual checks I did around the app. Should be good to go.

PR #7834, opened 2021-08-06
M3-5379 E2E Cypress Upgrade to (hopefully) address CI performance issues

ndukatz:
There's not a lot you need to do to test this. The best way to see that it worked is to get it merged. I've verified that it still runs tests and doesn't break the app. You can run one or a few tests if you'd like to verify that as well.


bnussman (approved):
All for updating dependencies when there are no negative side effects

tiffwong (approved):

PR #7836, opened 2021-08-09
M3-5385: Handling for entity_transfer_accept_recipient events

DevDW:
## Description
Add handling for new `entity_transfer_accept_recipient` event; fix link strings for Service Transfer events so that clicking on those notifications brings you to the "Service Transfers" tab. 

## How to test
Initiate and complete a service transfer; for `entity_transfer_accept_recipient` events, the recipient account should see a notification in the drawer that reads, "You have accepted a service transfer," and the account that sent the transfer should just have one that says, corresponding to the `entity_transfer_accept` event, "A service transfer has been accepted." Clicking both should bring you to the "Service Transfers" tab (`/account/service-transfers`).

### DIFF
--- a/packages/manager/src/eventMessageGenerator.ts
+++ b/packages/manager/src/eventMessageGenerator.ts
@@ -171,6 +171,9 @@ export const eventMessageCreators: { [index: string]: CreatorsForStatus } = {
   entity_transfer_accept: {
     notification: (_) => `A service transfer has been accepted.`,
   },
+  entity_transfer_accept_recipient: {
+    notification: (_) => `You have accepted a service transfer.`,
+  },
   entity_transfer_cancel: {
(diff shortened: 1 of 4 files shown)


HanaXu (commented):
On the recipient account, I'm seeing a bit more text than just `You have accepted a service transfer`. I'm also seeing `by current_account` instead of the account that sent the transfer

image

Semi-related, should we make the success toast a little more specific like `Service transfer successfully accepted`?:

image

bnussman:
Screen Shot 2021-08-09 at 2 40 58 PM
 Works well and linking is also good. Only suggestion would be to change up the wording... Maybe something like 
```
A service transfer has been accepted by banks.
```

bnussman (approved):

HanaXu (approved):

PR #7838, opened 2021-08-09
M3-5321: Add restricted user support for Firewalls

HanaXu:
## Description
Adds restricted user support for Firewalls

## How to test
In `/account/users/<username_here>/permissions` for a limited user:

- The ability to add Firewalls should now be an option under `Global Permissions`
- You should be able to see Firewalls, similar to other entities, under `Specific Permissions`

In `/firewalls`:

- If a user has no permissions for a specific Firewall, the firewall row should not display at all
- If a user has `read_only` permissions for a specific Firewall, the buttons for that firewall should be disabled and hovering over it should display tooltip text explaining why
- If a user has `read_write` permissions for a specific Firewall, the buttons for that firewall should not be disabled and the user should be able to modify the Firewall
- If a user does not have the ability to add Firewalls, the Create Firewall drawer should display an error notice at the top and the inputs should be disabled

In `/firewalls/id`:
- If a user has `read_only` permissions, there should be an error notice under the Rules/Linodes tab and the inputs should be disabled
    - In the Linodes tab, I specifically left the Linode label clickable, let me know if I should disable that as well

To add a Firewall to a Linode, you need read/write permissions on the Firewall and on the Linode. If a user has permissions for a Firewall but does not have any Linode permissions, the Linode list will come back empty, even if there are actually Linodes behind the firewall.


packages/manager/src/features/Firewalls/FirewallDetail/Devices/FirewallLinodesLanding.tsx
```diff
@@ -134,6 +136,15 @@ const FirewallLinodesLanding: React.FC<CombinedProps> = (props) => {
 
   return (
     <>
+      {disabled ? (
+        <Notice
+          text={
+            "You don't have permissions to modify this Firewall. Please contact an account administrator for details."
+          }
+          error={true}
```
bnussman, line 144: just use `error` shorthand

bnussman (changes requested):
Looks like we need to check for the add_firewall grant in `packages/manager/src/features/Firewalls/FirewallLanding/AddFirewallDrawer.tsx`. Right now it only checks if a user is restricted, not the actual add permission for Firewalls. 

Screen Shot 2021-08-10 at 10 26 20 AM
Screen Shot 2021-08-10 at 10 19 07 AM

bnussman:
One thing I am noticing (API related) is that a restricted user can create a Firewall, but then by default, that restricted user **does not** have read or write perms on the firewall they created. This behavior seems confusing and I think we should propose that if a restricted user creates a Firewall, they should have default access to read and write on it.

HanaXu:
@bnussman Good catch, that will be addressed in ARB-2671

bnussman (approved):

DevDW (approved):
Nice job! 🚢

PR #7840, opened 2021-08-10
M3-5381: Fix disabled action buttons in dark mode

tiffwong:
## Description

This fixes the background color for disabled buttons and the Upload Image button in dark mode.

Example:
Screen Shot 2021-08-06 at 4 14 30 PM

Screen Shot 2021-08-09 at 4 53 33 PM

## How to test
- Turn dark mode on
- Go to `/linodes` on a restricted account (or turn off one or more Linodes on any account)
- The background of the disabled inline actions should be the same color as the table background
- Go to `/images/create/upload` 
- The "Browse File" font and background color should be different


DevDW (approved):
🚢

DevDW:
it sounds like a hotfix will be going out tomorrow; I think it'd be great to include this in with that

PR #7843, opened 2021-08-10
M3-5173: Remove now-unneeded fakeAtlanta logic

DevDW:
## Description
Now that the new Atlanta facility is up and running, the `fakeAtlanta` logic that was added last year to show the region as disabled in situations where the API didn't return Atlanta can be removed.

## How to test
Ensure all Create and edit flows function as normal and there are no regressions in Region dropdowns.


codyfinn (approved):
🚢

bnussman (approved):

PR #7844, opened 2021-08-10
E2E removing after hook to avoid conflicts in ci runs

ndukatz:
There is not too much to test with this. This change is primarily for the benefit of ci runs. I made a small change to the username test but it's minimal. This should help to avoid some api conflicts when trying to delete right after a test runs. Now we will only do cleanup before each test block so things should go a bit more smoothly.


bnussman:
I don't exactly know the implications of this, but willing to try. Just revisit if there are any disadvantages of not cleaning up after a run

bnussman (approved):

HanaXu (approved):

DevDW (approved):

PR #7858, opened 2021-08-12
M3-5393: Workbox in Cloud

DevDW:
## Description
Following up on https://github.com/linode/manager/pull/7779 and https://github.com/linode/manager/pull/7833, this is an exploration of using Workbox in Cloud.

**Reference materials**
* Service workers: https://developers.google.com/web/fundamentals/primers/service-workers
* Workbox: https://developers.google.com/web/tools/workbox
* Workbox in Create React App: https://web.dev/precache-with-workbox-react/, https://create-react-app.dev/docs/making-a-progressive-web-app/
* Github Issue about decision to include Workbox in Create React App: https://github.com/facebook/create-react-app/issues/2398


DevDW:
Since debugging deployments can be made more difficult and the applicability of Workbox's strengths for our use case in Cloud is questionable, we will not be utilizing Workbox at this time.

PR #7862, opened 2021-08-13
M3-5394: Fix Kubernetes Debian images showing up in the Rebuild Dialog

bnussman:
## Description

- Fixes Kubernetes Debian images showing up in the Rebuild Dialog
- Caused by #7800 
- Fixed by filtering out Kube images when ImageSelection variant is `all`

Please check my filtering logic. I want to make sure Linode's Kubernetes images get filtered out but I want to make sure that if a user has a script with Kube in the name, it still gets shown.

## How to test

- Check all Image selections in Cloud Manager
- Specifically Check the Linode Rebuild from Image Dialog

### DIFF
--- a/packages/manager/src/components/ImageSelect/ImageSelect.tsx
+++ b/packages/manager/src/components/ImageSelect/ImageSelect.tsx
@@ -143,6 +143,8 @@ export const ImageSelect: React.FC<Props> = (props) => {
       case 'private':
         return !thisImage.is_public && thisImage.status === 'available';
       case 'all':
+        // Even if all images are requested, we don't want to expose the Kubernetes images
+        return !thisImage.label.match(/kube/i);
       default:
         return true;
     }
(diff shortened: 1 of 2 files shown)


packages/manager/src/components/ImageSelect/ImageSelect.tsx
```diff
@@ -143,6 +143,13 @@ export const ImageSelect: React.FC<Props> = (props) => {
       case 'private':
         return !thisImage.is_public && thisImage.status === 'available';
       case 'all':
+        /*
+         * Even if all images are requested, we don't want to expose the Kubernetes images
+         * We keep images that don't have kube in the name that are created by Linode
```
DevDW, line 148: `"We don't show images with 'kube' in the label that are created by Linode"` ?

DevDW (approved):

tiffwong (approved):

PR #7863, opened 2021-08-13
M3-5395: Fix users unable to set a label and description when capturing an Image

bnussman:
## Description

- Fixed a random but somewhat important bug I found where on `/images/create/disk`, the **label** and **description** fields were not getting passed to the API and not being used. This made creating an Image from a disk really confusing. 

## How to test

- Test creating an Image from a disk and also test uploading an Image.
- Make sure label and description get passed to the API in the dev tools network tab (previously it those values were not being sent)


DevDW (approved):
Huh, this is weird. I wonder how long this bug has been there? https://github.com/linode/manager/pull/7575 (the PR that added this component) was tested pretty thoroughly and this has been rolled out now for a number of months; it would be surprising if it flew under the radar for so long.

Regardless, great catch! It appears to be functioning as intended now 👍

tiffwong (approved):

PR #7864, opened 2021-08-16
M3-5204: Add Cost Estimator for Custom Images

codyfinn:
## Description

Adds price estimation for creating custom images from existing Linode Disks. The price estimation is based on the size of the disk and the current unit cost of $0.10 / GB of Disk storage. Included is also utility function to convert between storage units easily and another utility function to calculate total prices given a unit price and a given quantity.

## How to test

Navigate to the CreateCustomImageTab component and select a Linode and a Disk. The cost estimator should show the correct price for the size of the disk.

`yarn test unitPriceCalculator`
`yarn test unitConversions --t convertStorageUnit`


johnwcallahan (commented):
UI question for Matthew and the team: should we consider adding "Estimated: " before the price? (since you could e.g. resize the disk).

packages/manager/src/utilities/unitConversions.ts
```diff
@@ -189,3 +189,34 @@ export const convertBytesToTarget = (unit: StorageSymbol, value: number) => {
...
+  TB = 4,
+}
+
+type StorageUnitExponentKey = keyof typeof StorageUnitExponents;
+
+export const convertStorageUnit = (
+  sourceUnit: StorageUnitExponentKey,
+  sourceQuantity: number | undefined,
+  targetUnit: StorageUnitExponentKey
+) => {
```
bnussman, line 207: Can you add a tsdoc style comment for this function?

packages/manager/src/utilities/unitPriceCalculator.ts
```diff
@@ -0,0 +1,6 @@
+export default function unitPriceCalculator(
+  unitPrice: number,
+  quantity: number
+) {
+  return `$${Math.max(unitPrice * quantity, 0).toFixed(2)}`;
+}
```
bnussman, line 6: And maybe this one if you feel like its would be useful to readers

DevDW (commented):
Looking good to me so far

packages/manager/src/utilities/unitConversions.ts
```diff
@@ -176,16 +176,48 @@ const determineDecimalPlaces = (
 export const convertBytesToTarget = (unit: StorageSymbol, value: number) => {
   switch (unit) {
     case 'byte':
-      return value;
     case 'bytes':
+    case 'B':
```
DevDW, line 180: Seeing this TS error:
Screen Shot 2021-08-19 at 11 10 08 AM

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
```diff
@@ -162,6 +164,15 @@ export const CreateImageTab: React.FC<Props & ImagesDispatch> = (props) => {
 
   const requirementsMet = checkRequirements();
 
+  const selectedDiskData: Disk | undefined = disks.find(
+    (d) => `${d.id}` === selectedDisk
+  );
+  const selectedDiskSizeInMB = convertStorageUnit(
```
DevDW, line 170: I think this variable should be renamed to `selectedDiskSizeInGB`

DevDW (approved):
Nice work! Left a couple of small comments

bnussman (approved):
🚀

PR #7865, opened 2021-08-17
M3-5368: Fix support ticket input with different font sizes

bnussman:
## Description

- FIxes weird text overflow issues in the Create Ticket dialog Text Field or Ticket Reply Text Field when user changes their browsers font size (chrome://settings/?search=Font)
- This issues was happening because the TextField component takes `rows` as a prop which dictates the size, but we also had a height css property on it, these were most likely conflicting.
- This fix is to remove the static CSS heigh we assigned and rely on the `rows` prop for size

Before
Screen Shot 2021-08-17 at 10 42 37 AM

After
Screen Shot 2021-08-17 at 10 43 07 AM

## How to test

- Go to Chrome's settings and change the browser's font size to **large** or anything bigger
- Test the **Open a Support Ticket** dialog
- Make sure the reply Text Field for a ticket thread still looks right
- Make sure the size of ticket related dialogs look correct, because this PR changes the size a bit

### DIFF
--- a/packages/manager/src/features/Support/SupportTicketDetail/TabbedReply/TicketReply.tsx
+++ b/packages/manager/src/features/Support/SupportTicketDetail/TabbedReply/TicketReply.tsx
@@ -11,7 +11,6 @@ type ClassNames = 'replyField';
 const styles = () =>
   createStyles({
     replyField: {
-      height: 320,
       marginTop: 0,
       '& > div': {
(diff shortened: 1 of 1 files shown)


ndukatz (approved):
Looks good, works well with various font and screen sizes. Still passes support e2e.

tiffwong (approved):

PR #7866, opened 2021-08-17
M3-5384: Add support for multiple payment methods in make payment drawer

HanaXu:
## Description
- Adds support for displaying multiple payment methods in the Make a Payment drawer
- If a user is making a payment through the action dropdown, auto-select that payment in the Make a Payment drawer
- Refactor CardBase into a functional component
- Refactor variant logic out of SelectionCard; accept a renderVariant function instead, remove duplicate css
    - I went ahead and deleted the check mark since it was only being used in one other place and selectable does the job

## How to test
The Make a Payment drawer should now display multiple payment methods. The default payment method is auto-selected unless the user opened the drawer through the payment method dropdown menu, then that payment method is selected. Making a payment should charge the selected card (check the network tab).

Check places in the app where the SelectionCard component is used, ensure it looks good on mobile and that there are no regressions.


packages/manager/src/components/PaymentMethodRow/ThirdPartyPayment.tsx
```diff
@@ -20,17 +20,13 @@ const useStyles = makeStyles((theme: Theme) => ({
...
-    alignItems: 'flex-start',
+    alignItems: 'center',
     justifyContent: 'center',
     paddingLeft: 6,
     paddingRight: 6,
     width: 45,
-    '& svg': {
-      // Safari needs the height/width defined in order to render the image
-      width: 33,
-    },
```
HanaXu, line 33: Added width to the svg itself

packages/manager/src/assets/icons/payment/googlePay.svg
```diff
@@ -1,4 +1,4 @@
-<svg id="G_Pay_Acceptance_Mark" enable-background="new 0 0 1094 742" version="1.1" viewBox="0 0 752 400" xml:space="preserve" xmlns="http://www.w3.org/2000/svg">
+<svg id="G_Pay_Acceptance_Mark" width="33" enable-background="new 0 0 1094 742" version="1.1" viewBox="0 0 752 400" xml:space="preserve" xmlns="http://www.w3.org/2000/svg">
```
bnussman, line 1: Is this causing a regression in the Add Payment Method drawer?

packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/PaymentDrawer.tsx
```diff
@@ -203,21 +329,129 @@ export const PaymentDrawer: React.FC<Props> = (props) => {
...
+                        ? getTPPIcon(paymentMethod.type)
+                        : getCreditCardIcon(paymentMethod.data.card_type);
+                    return <Icon />;
+                  };
+
+                  const renderVariant = () => {
+                    return paymentMethod.is_default ? (
+                      <Grid item className={`${classes.chip}`} xs={3}>
+                        <Chip label="DEFAULT" component="span" />
+                      </Grid>
```
bnussman, line 365: Screen Shot 2021-08-17 at 12 07 14 PM
I find that the placement of the `DEFAULT` being different to be a bit odd looking, but I personally don't mind it.

packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/PaymentDrawer.tsx
```diff
@@ -203,21 +329,129 @@ export const PaymentDrawer: React.FC<Props> = (props) => {
...
+          {showGooglePay ? (
+            <Grid container direction="column">
+              <Grid item>
+                <Typography
+                  variant="h3"
+                  className={classes.header}
+                  style={{ marginBottom: 16 }}
+                >
+                  <strong>Payment Methods</strong>
+                </Typography>
```
bnussman, line 341: We should be consistent with the `:`'s

Screen Shot 2021-08-17 at 12 21 00 PM

packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/PaymentDrawer.tsx
```diff
@@ -128,6 +200,63 @@ export const PaymentDrawer: React.FC<Props> = (props) => {
...
+          payment_method_id: paymentMethodId,
+        }
+      : {
+          usd: (+usd).toFixed(2),
+          cvv,
+        };
+
+    makePayment({
+      ...makePaymentData,
+    })
```
bnussman, line 238: Use this, no need to spread here
```ts
makePayment(makePaymentData)
```

packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/PaymentDrawer.tsx
```diff
@@ -203,21 +329,129 @@ export const PaymentDrawer: React.FC<Props> = (props) => {
...
+                  };
+
+                  const renderVariant = () => {
+                    return paymentMethod.is_default ? (
+                      <Grid item className={`${classes.chip}`} xs={3}>
+                        <Chip label="DEFAULT" component="span" />
+                      </Grid>
+                    ) : (
+                      <Grid />
+                    );
```
bnussman, line 368: Can we return `null` instead of `<Grid />`?

packages/manager/src/assets/icons/payment/googlePay.svg
bnussman: The icon got tiny

Screen Shot 2021-08-17 at 12 31 18 PM

bnussman (commented):
Would it be possible to line up the right side of the **Pay Now** button and the selectable cards?

Screen Shot 2021-08-17 at 3 20 34 PM

HanaXu:
@bnussman good catch! 🔍

bnussman (changes requested):
When a user clicks a different payment method in the drawer, can we update the URL in the address bar with the new ID? This isn't super important, but could be useful for some users and we do it many other places in Cloud Manager.

HanaXu:
@bnussman I decided to pass the id using the state instead of in the url since most users won't know what their payment method id is, and this makes the code less complicated

tiffwong (approved):
Couldn't find any regressions, great job!

bnussman (approved):
Awesome job!

packages/manager/src/components/SelectionCard/SelectionCard.tsx
```diff
@@ -80,93 +44,18 @@ const useStyles = makeStyles((theme: Theme) => ({
...
-  variant?: 'check' | 'info' | 'quantityCheck' | 'selectable';
   className?: string;
-  inputValue?: number;
-  setInputValue?: (value: number) => void;
-  submitForm?: (e: any) => void;
-  buttonDisabled?: boolean;
-  displayButton?: boolean;
+  onClick?: (e: React.SyntheticEvent<HTMLElement>) => void;
+  renderIcon?: () => JSX.Element;
+  renderVariant?: () => any;
```
DevDW, line 58: `renderVariant?: () => JSX.Element;`

DevDW (approved):
LGTM! ➕➕

ndukatz (approved):
Looks good. I didn't find any regressions and it still passes google pay tests with the update I pushed up to this branch. I will add a bit more testing around this later.

packages/manager/src/components/SelectionCard/SelectionCard.tsx
HanaXu: I made this any since the `renderVariant` function in `PaymentDrawer.tsx` can return null if the payment method is not the default and TS complains about that since null is not a JSX element

DevDW: I see, how about `renderVariant?: () => JSX.Element | null;`?

HanaXu: Oh yeah 😅

PR #7867, opened 2021-08-18
M3-3874: Tighten spacing for Support Ticket Details

tiffwong:
## Description

This is a fairly older ticket which described the "Add Update" button going off the screen on mobile if there's a lot of text. I was still able to access the button when I tried to replicate this so it might've gotten fixed. 

I did reduce the vertical spacing on that page, collapsed the "Formatting Tips" by default aligned the "Add Update" button to the right so it's consistent with our style guide.

Also refactored the test for `AttachFileListItem` and fixed a bug where the `defaultExpanded` prop wasn't being applied to the `Accordion` component.

## How to test

- Go to `/support/tickets/[insert id]` 
- Either create a bunch of replies on the ticket or message me for a test account that already has content in it
- Make sure you can access the "Add Update" button on mobile

### DIFF
--- a/packages/manager/src/components/Accordion/Accordion.tsx
+++ b/packages/manager/src/components/Accordion/Accordion.tsx
@@ -71,7 +71,11 @@ export const Accordion: React.FC<CombinedProps> = (props) => {
   const notice = success || warning || error || null;
 
   return (
-    <_Accordion defaultExpanded {...accordionProps} data-qa-panel>
+    <_Accordion
+      defaultExpanded={defaultExpanded}
+      {...accordionProps}
+      data-qa-panel
+    >
       <AccordionSummary
         onClick={handleClick}
         expandIcon={<KeyboardArrowDown className="caret" />}
--- a/packages/manager/src/features/Support/AttachFileListItem.tsx
+++ b/packages/manager/src/features/Support/AttachFileListItem.tsx
@@ -4,51 +4,36 @@ import * as React from 'react';
 import { compose, withHandlers } from 'recompose';
 import Button from 'src/components/Button';
 import InputAdornment from 'src/components/core/InputAdornment';
-import {
-  createStyles,
-  Theme,
-  withStyles,
-  WithStyles,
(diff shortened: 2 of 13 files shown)


packages/manager/src/components/Accordion/Accordion.tsx
```diff
@@ -71,7 +71,11 @@ export const Accordion: React.FC<CombinedProps> = (props) => {
   const notice = success || warning || error || null;
 
   return (
-    <_Accordion defaultExpanded {...accordionProps} data-qa-panel>
+    <_Accordion
+      defaultExpanded={defaultExpanded}
```
tiffwong, line 75: Fix for prop not being applied to the `Accordion` component

packages/manager/src/features/Support/AttachFileListItem.tsx
```diff
@@ -98,15 +86,19 @@ export const AttachFileListItem: React.FC<CombinedProps> = (props) => {
...
         <Grid item>
           <Button
-            buttonType="primary"
+            buttonType="outlined"
             onClick={onClick}
+            data-testid="delete-button"
             data-qa-delete-button
-          />
+          >
+            Delete
```
tiffwong, line 100: This button originally didn't have any text for some reason

Screen Shot 2021-08-18 at 7 59 24 PM

bnussman:
Do we need to fix the button placement for this modal? 
Screen Shot 2021-08-19 at 12 40 32 PM

bnussman (commented):
I'm noticing that when creating a ticket, the `X` icon is used to delete an attachment, but in the reply section, we are using a `Delete` button. Should we make these the same? 

Screen Shot 2021-08-20 at 9 39 27 AM

Screen Shot 2021-08-20 at 9 39 13 AM

DevDW (approved):
To @bnussman's question, I think the modal and the detail view should be made more consistent with each other, but this looks good otherwise!

bnussman (approved):

PR #7868, opened 2021-08-18
M3-5358: Allow more room for image names

bcmertz:
## Description

This PR increases the width of the linode image name width

## How to test

Clone branch and view images under /images

N/A No relevant unit tests - small styling change only

## Notes

ImagesLanding.tsx: "widthPercent" before only added up to 100 in the case of tableType=='automatic' and flagEnabled==true headers being passed in. 

This PR has the side effect of making the width percents not add up to 100 in this case but the way components/EntityTable/EntityTableHeaders.tsx presents the width to the TableCell doesn't seem to have a 100% sum requirement. Perhaps it's OK to ignore this since we do in most other instances, and making the ratios change to sum to 100% is kinda unwieldily?

### DIFF
--- a/packages/manager/src/features/Images/ImagesLanding.tsx
+++ b/packages/manager/src/features/Images/ImagesLanding.tsx
@@ -56,9 +56,6 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
 }));
 
-/**
- * @todo remove conditional cell sizes when we're no longer relying on a flag here.
- */
 const getHeaders = (
   tableType: 'automatic' | 'manual',
   flagEnabled: boolean = true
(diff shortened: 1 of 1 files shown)


DevDW (commented):
Nice, this works but I think we can do a bit more to tighten things up in this file and simplify the styling logic.

Machine Images is now in General Availability (GA), so we can address those `@todo`'s in the file and remove the second parameter from `getHeaders`.

Because the action menu column is always collapsed with no inline action outside of one case (if I'm not mistaken), we don't need it taking up 35% of the total row width; decreasing it to maybe 15% seems fine. And then maybe 20% each for "Created" and "Expires".

tiffwong (approved):
Looks good! Approving based on the changes Dajahi recommended

bcmertz:
Thanks for the feedback @tiffwong  and @DevDW ! I made the changes, apologies for the delay :)

DevDW (approved):
🎉  🚀

PR #7869, opened 2021-08-18
M3-5364 E2E Linode Landing Tests

ndukatz:
## Description
first test of ticket for linode landing. checks basic elements and menu items

## How to test
- `yarn up` in one terminal
- `yarn cy:e2e -s ./cypress/integration/linodes/smoke-linode-landing-table.spec.ts` in another

```
REACT_APP_LOGIN_ROOT
REACT_APP_API_ROOT
REACT_APP_APP_ROOT
REACT_APP_LAUNCH_DARKLY_ID
REACT_APP_CLIENT_ID
MANAGER_OAUTH
```

### DIFF
--- a/packages/manager/cypress/integration/linodes/smoke-linode-landing-table.spec.ts
+++ b/packages/manager/cypress/integration/linodes/smoke-linode-landing-table.spec.ts
@@ -1,6 +1,10 @@
 import { contains } from 'cypress/types/jquery';
 import { createLinode } from '../../support/api/linodes';
-import { getClick, getVisible } from '../../support/helpers';
+import { fbtVisible, getClick, getVisible } from '../../support/helpers';
+import { linodeFactory } from '@src/factories/linodes';
+import { makeResourcePage } from '@src/mocks/serverHandlers';
+
+const appRoot = Cypress.env('REACT_APP_APP_ROOT');
 
 const deleteLinodeFromActionMenu = (linodeLabel) => {
   getClick(`[aria-label="Action menu for Linode ${linodeLabel}"]`);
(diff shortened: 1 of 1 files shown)


packages/manager/cypress/integration/linodes/smoke-linode-landing-table.spec.ts
```diff
@@ -11,23 +15,80 @@ const deleteLinodeFromActionMenu = (linodeLabel) => {
   cy.wait('@deleteLinode').its('response.statusCode').should('eq', 200);
 };
 
+const mockLinodes = makeResourcePage(linodeFactory.buildList(5));
+
 describe('linode landing', () => {
+  it('checks the landng page menu items, labels, and buttons', () => {
```
patthiel, line 21: this is testing a lot in a single test function. Since these tests aren't taking any actions i think it's fine for you to break them into several small `it()` tests. 

perhaps use a `beforeEach()` or `before()` to load up the mocks then visit the page. this way you keep it DRY but also make the tests narrow in scope and easier to debug.

packages/manager/cypress/integration/linodes/smoke-linode-landing-table.spec.ts
```diff
@@ -11,23 +15,80 @@ const deleteLinodeFromActionMenu = (linodeLabel) => {
...
+    cy.intercept('GET', '*/linode/instances/*', (req) => {
+      req.reply(mockLinodes);
+    }).as('getLinodes');
+    cy.visitWithLogin('/');
+    cy.wait('@getLinodes');
+    cy.wait('@getProfile').then((xhr) => {
+      const username = xhr.response?.body.username;
+      cy.url().should('eq', `${appRoot}/linodes`);
+
+      // side menu
```
patthiel, line 32: these sort of tests are why i liked the page object pattern. If one of these data attributes changes then you have a bunch of references in the framework that have to be updated. Have you thought about using UI maps or page object model at all to make that maintenance easier?

tiffwong (approved):
The test failed when I had Managed on but passed when I turned it off so that might be something to factor in when you're testing it
Screen Shot 2021-08-23 at 1 08 16 PM

packages/manager/cypress/integration/linodes/smoke-linode-landing-table.spec.ts
```diff
@@ -11,23 +15,80 @@ const deleteLinodeFromActionMenu = (linodeLabel) => {
...
 
+const mockLinodes = makeResourcePage(linodeFactory.buildList(5));
+
 describe('linode landing', () => {
+  it('checks the landng page menu items, labels, and buttons', () => {
+    cy.intercept('GET', '*/profile').as('getProfile');
+    cy.intercept('GET', '*/linode/instances/*', (req) => {
+      req.reply(mockLinodes);
+    }).as('getLinodes');
+    cy.visitWithLogin('/');
```
bnussman, line 26: @tiffwong brought up a good point. Testing on a Managed account does not work with this test. We can fix this here by just doing `cy.visitWithLogin('/linodes');` 

Another way we can fix it could be by mocking `/account/settings` by doing something like

```ts
const mockAccountSettings = accountSettingsFactory.build({ managed: false });

cy.intercept('GET', '*/account/settings', (req) => {
  req.reply(mockAccountSettings);
}).as('getAccountSettings');
```

ndukatz:
@bnussman that is a good call. For now I'll set managed to false, but will add tests for managed in future PR.

bnussman (approved):

PR #7870, opened 2021-08-19
M3-5375: Handle API maintenance mode

DevDW:
## Description
If the API is in maintenance mode, users would not be able to do anything in Cloud because successful API requests wouldn't be possible.

To avoid that, this PR creates a maintenance view that will display if the API is in maintenance mode:

Screen Shot 2021-08-19 at 4 32 47 PM

## How to test
To just test the view itself, you can make the condition on line 261 of `MainContext.tsx` true.

To test the logic for the response headers & errors, you'll need to use devenv. Please reach out if you need details on putting the API into maintenance mode in that environment.

### DIFF
--- a/packages/manager/src/request.tsx
+++ b/packages/manager/src/request.tsx
@@ -16,7 +16,6 @@ const handleSuccess: <T extends AxiosResponse<any>>(response: T) => T | T = (
   response
 ) => {
   if (!!response.headers['x-maintenance-mode']) {
-    store.dispatch(handleLogout());
     Promise.reject(response);
   }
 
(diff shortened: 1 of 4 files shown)


codyfinn (approved):
🚢

packages/manager/src/request.tsx
```diff
@@ -16,31 +16,42 @@ const handleSuccess: <T extends AxiosResponse<any>>(response: T) => T | T = (
...
   const url = config.url ?? '';
   const status: number = error.response?.status ?? 0;
   const errors: APIError[] = error.response?.data?.errors ?? [
     { reason: DEFAULT_ERROR_MESSAGE },
   ];
 
+  const apiInMaintenanceMode =
+    !!error.config?.headers['x-maintenance-mode'] ||
+    errors.some((err) => {
+      return Boolean(err.reason.match(/maintenance mode/i));
```
johnwcallahan, line 44: I'm inclined to think we should rely solely on the `x-maintenance-mode` header for this. My reasoning is that matching on the presence of `maintenance mode` anywhere in the error text could result in false positives, and the consequences are severe if this condition is met (i.e. the app is unusable).

johnwcallahan (commented):
Nice job using the existing account activation error patterns. The overall approach and UI looks good, have one comment about triggering this condition.

packages/manager/src/request.tsx
DevDW: That's fair and a good point -- I changed it to rely solely on that header 👍

HanaXu (commented):
I think the view would look cleaner if we centered the screen and reduced the logo size a bit 🤔

johnwcallahan (approved):

PR #7871, opened 2021-08-20
M3-5399: Create Agreement Confirmation Checkbox Component

bnussman:
## Description

- Creates a checkbox component containing our new agreement copy

Screen Shot 2021-08-20 at 10 26 11 AM

## How to test

- Look at the Linode's create sidebar, I put one there to temporarily test this component

### DIFF
--- a/packages/manager/src/features/Account/Agreements/AgreementCheckbox.tsx
+++ b/packages/manager/src/features/Account/Agreements/AgreementCheckbox.tsx
@@ -1,0 +1,47 @@
+import * as React from 'react';
+import CheckBox from 'src/components/CheckBox';
+import Box from 'src/components/core/Box';
+import Typography from 'src/components/core/Typography';
+import Link from 'src/components/Link';
+
+interface Props {
+  checked: boolean;
+  onChange: (
+    event: React.ChangeEvent<HTMLInputElement>,
+    checked: boolean
+  ) => void;
+}
+
+const AgreementCheckbox: React.FC<Props> = (props) => {
+  const { checked, onChange } = props;
+
+  return (
+    <Box display="flex" flexDirection="row">
+      <Box>
+        <CheckBox
+          checked={checked}
+          onChange={onChange}
+          style={{ paddingLeft: 0 }}
+        />
+      </Box>
+      <Box>
+        <Typography>
+          <React.Fragment>
+            I have read and agree to the{' '}
+            <Link to="https://www.linode.com/legal-privacy/">
+              Linode Privacy Policy
+            </Link>{' '}
+            and{' '}
+            <Link to="https://www.linode.com/eu-model/">
+              EU Standard Contractual Clauses
+            </Link>
+            , which govern the cross border transfer of data relating to the
+            European Economic Area. click here.
+          </React.Fragment>
+        </Typography>
+      </Box>
+    </Box>
+  );
+};
+
+export default AgreementCheckbox;
--- a/packages/manager/src/features/Account/Agreements/AgreementCheckbox.tsx
+++ b/packages/manager/src/features/Account/Agreements/AgreementCheckbox.tsx
@@ -26,18 +26,16 @@ const AgreementCheckbox: React.FC<Props> = (props) => {
       </Box>
       <Box>
         <Typography>
-          <React.Fragment>
-            I have read and agree to the{' '}
-            <Link to="https://www.linode.com/legal-privacy/">
-              Linode Privacy Policy
-            </Link>{' '}
-            and{' '}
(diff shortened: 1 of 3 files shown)


bnussman:
Should the font size be `1rem` or should we leave as is?

packages/manager/src/features/Account/Agreements/AgreementCheckbox.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+interface Props {
+  checked: boolean;
+  onChange: (
+    event: React.ChangeEvent<HTMLInputElement>,
+    checked: boolean
+  ) => void;
+}
+
+const AgreementCheckbox: React.FC<Props> = (props) => {
```
johnwcallahan, line 15: minor: maybe make the component name more specific? ("EUAgreementCheckbox" or something..)

packages/manager/src/features/Account/Agreements/EUAgreementCheckbox.tsx
```diff
@@ -0,0 +1,45 @@
...
+          I have read and agree to the{' '}
+          <Link to="https://www.linode.com/legal-privacy/">
+            Linode Privacy Policy
+          </Link>{' '}
+          and{' '}
+          <Link to="https://www.linode.com/eu-model/">
+            EU Standard Contractual Clauses
+          </Link>
+          , which govern the cross border transfer of data relating to the
+          European Economic Area. click here.
```
DevDW, line 38: "click here" is stray text I believe

packages/manager/src/features/Account/Agreements/EUAgreementCheckbox.tsx
```diff
@@ -0,0 +1,45 @@
...
+        <Typography>
+          I have read and agree to the{' '}
+          <Link to="https://www.linode.com/legal-privacy/">
+            Linode Privacy Policy
+          </Link>{' '}
+          and{' '}
+          <Link to="https://www.linode.com/eu-model/">
+            EU Standard Contractual Clauses
+          </Link>
+          , which govern the cross border transfer of data relating to the
```
DevDW, line 37: We should hyphenate "cross-border"

DevDW:
> Should the font size be `1rem` or should we leave as is?

I'm leaning towards leaving as is. I don't mind the larger font size when the checkout bar is positioned on the right. In smaller viewports, when the checkout bar gets shifted to the bottom, the larger font size doesn't look as good to me though.

DevDW (approved):

HanaXu (approved):
LGTM, would it be better tho to just define the onChange behavior in this component since it seems that checking behavior will be the same across the app? 🤔

johnwcallahan:
> would it be better tho to just define the onChange behavior in this component since it seems that checking behavior will be the same across the app?

@HanaXu I was wondering about that, but I do think this component needs to be controlled because throughout the UI the parent(s) of this component will need to keep track of the state of the checkbox (whether it's checked or not). 

Maybe once we get it into the real world a better pattern will be revealed.

johnwcallahan (approved):

packages/manager/src/features/Account/Agreements/EUAgreementCheckbox.tsx
```diff
@@ -0,0 +1,45 @@
...
+const EUAgreementCheckbox: React.FC<Props> = (props) => {
+  const { checked, onChange } = props;
+
+  return (
+    <Box display="flex" flexDirection="row">
+      <Box>
+        <CheckBox
+          checked={checked}
+          onChange={onChange}
+          style={{ paddingLeft: 0 }}
```
tiffwong, line 24: Instead of changing the padding, what about setting it as `marginLeft: -8px`? Otherwise the hover state/circle looks odd

Screen Shot 2021-08-24 at 1 15 02 PM

packages/manager/src/features/Account/Agreements/EUAgreementCheckbox.tsx
```diff
@@ -0,0 +1,45 @@
...
+    event: React.ChangeEvent<HTMLInputElement>,
+    checked: boolean
+  ) => void;
+}
+
+const EUAgreementCheckbox: React.FC<Props> = (props) => {
+  const { checked, onChange } = props;
+
+  return (
+    <Box display="flex" flexDirection="row">
```
tiffwong, line 19: If you add `alignItems="flex-start"` here, you won't need the `<Box>`'s around the `<Checkbox />` and `<Typography>`

tiffwong (approved):
Left a few cleanup items but looks great!

Issue #7750, opened 2021-07-09
Missing validation on firewall rule label


andysh-uk:
When adding a firewall rule that has spaces, there is no error displayed in the UI, but the API has returned a validation error.

The UI just looks like it does nothing - no error is presented to the user.

To reproduce:

- Create a new firewall rule
- Enter a name with a space (mine was: "MariaDB Downloads")
- Click Add Changes, and then Save

Expected behaviour:

Warning message advising that the firewall label contains spaces on clicking Add Changes.

Actual behaviour:

The UI does nothing; the new row remains blue like it is unsaved. In the browser tools, a validation warning has been returned from the request:

```
PUT  https://cloud.linode.com/api/v4beta/networking/firewalls/xxx/rules
```

```json
{"errors": [{"reason": "Must include only ASCII letters, numbers, underscores, periods, and dashes", "field": "rules.outbound[4].label"}]}
```

HanaXu:
@andysh-uk The error seems to be displayed in the UI after clicking save. Can you confirm if this is still an issue?

![image](https://user-images.githubusercontent.com/14323019/128921725-8e1e3915-ed23-4a47-9f5c-ebb883546511.png)

andysh-uk:
Hi @HanaXu 

Yes, I can confirm that message is there.

However I have 2 issues with the implementation of this:

- on an iPad (not tried on a desktop browser yet) on clicking Save Changes, the screen jumps slightly as if it is trying to scroll up, but the validation error still isn’t visible. It took me a few seconds to realise it had appeared at the top of the screen, out of view
- If I’ve added a few rules, how do I know which one caused the validation error? I feel that message should be displayed when you click to add the rule, in the same way as you get “Must be a valid IPv4 or IPv6 range” when entering an invalid IP address. It should highlight the label field and tell me then it should be alpha

PR #7874, opened 2021-08-20
Add additional client-side Firewall rule label validation

HanaXu:
## Description
Re issue https://github.com/linode/manager/issues/7750: Adds additional client-side Firewall rule label validation to match the API

I looked into validating using `FirewallRuleTypeSchema` but the current structure of the code made it difficult and it was a bit out of scope, so I just added onto the `validateForm` function

- Label is required
    - There is inconsistency between actual API behavior and the docs. The [docs](https://www.linode.com/docs/api/networking/#firewall-create__request-body-schema) says label is required, however you can create a Firewall rule without a label. ARB-2686 has been created to address this
- Label must begin with a letter
- Label must only include ASCII letters, numbers, underscores, periods, and dashes

## How to test

- Go to `/firewalls/<firewall_id>`
- In the Rules tab, edit an existing rule or add a new inbound/outbound rule
- In the add/edit form, edit the label with valid/invalid text, then click the blue Add button
    - If the text is invalid, there should be a red error message under the label input and the drawer should not close
    - If the text is valid, the drawer should close and the Rule Table should be updated with the new rule

`yarn test FirewallRuleDrawer`

### DIFF
--- a/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.test.tsx
+++ b/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.test.tsx
@@ -96,8 +96,54 @@ describe('utilities', () => {
       );
       expect(validateForm('TCP', 'invalid-port')).toHaveProperty('ports');
     });
+    it('validates label', () => {
+      const expectedResults = [
+        {
+          value: 'test',
+          result: {},
+        },
+        {
+          value: 'accept-inbound-HTTP',
+          result: {},
+        },
+        {
+          value: '',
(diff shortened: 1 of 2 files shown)


ndukatz (approved):
Looks good, works well. Tested all scenarios with inbound and outbound rules. Unit test covers most cases and passed for me, though you may want to add another case with special characters that are invalid but with a valid length and start. example: (b&a$e#c). E2E around firewalls also passed and I've added a ticket to write a few more e2e tests around this specific fix.

bnussman (approved):

codyfinn (approved):
Looks good to me!

PR #7875, opened 2021-08-23
M3-5278: Update buttons for Domains

tiffwong:
## Description

Make add buttons on the Domain Detail page primary and remove "Delete Domain" from the edit drawer

## How to test

- Go to `/domains`
- Click on "Edit" and the "Delete Domain" button should no longer show up in the drawer
- Click on the label for any primary domain
- The add buttons on the right of each table should now be in the primary style

### DIFF
--- a/packages/manager/src/features/Domains/DeleteDomain.tsx
+++ b/packages/manager/src/features/Domains/DeleteDomain.tsx
@@ -2,7 +2,7 @@ import { withSnackbar, WithSnackbarProps } from 'notistack';
 import * as React from 'react';
 import { compose } from 'recompose';
 import Button from 'src/components/Button';
-import { makeStyles } from 'src/components/core/styles';
+import { makeStyles, Theme } from 'src/components/core/styles';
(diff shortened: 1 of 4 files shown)


bnussman:
Did we address the button placement for this drawer in one of our PR's? I know mine didn't, so if yours also didn't go ahead and fix the button placement in this PR

Screen Shot 2021-08-23 at 11 52 26 AM

bnussman (approved):
Looks good, but I feel like the **Delete** Button in the bottom right looks very out of place. 

Also on a small screen size, the delete button show in the screenshot does not have any right padding while the primary buttons do 

Screen Shot 2021-08-23 at 11 57 14 AM

DevDW (approved):
LGTM aside from the padding issue in mobile view that @bnussman mentioned 👌 

On a side note, I think "Add a NS Record" should be "Add **an** NS Record," and "Add a SRV Record" should be "Add **an** SRV Record" (the [third section](https://www.grammar.com/a-vs-an-when-to-use/) touches on this)

PR #7876, opened 2021-08-23
Make Payment Drawer UI fixes

HanaXu:
## Description
- Hide the `Pay Now` button if there are no payment methods on file
- Fix displaying of expired cards
- Show tooltip if a card is expired and display the tooltip on the left side instead of the right

## How to test
- On an account with no payment methods, the `Pay Now` button should not display in the make payment drawer
- If a payment method is expired:
    - The text in the SelectionCard should be red and read `expired`
    - The `Pay Now` button should be disabled and a tooltip should appear on the left side of the button

### DIFF
--- a/packages/manager/src/components/PaymentMethodRow/PaymentMethodRow.tsx
+++ b/packages/manager/src/components/PaymentMethodRow/PaymentMethodRow.tsx
@@ -65,7 +65,7 @@ const PaymentMethodRow: React.FC<Props> = (props) => {
       onClick: () => {
         history.push({
           pathname: '/account/billing/make-payment/',
-          state: { id: paymentMethod.id },
+          state: { paymentMethod },
         });
       },
     },
--- a/packages/manager/src/components/SelectionCard/CardBase.tsx
+++ b/packages/manager/src/components/SelectionCard/CardBase.tsx
@@ -106,7 +106,7 @@ const CardBase: React.FC<CombinedProps> = (props) => {
           return (
             <div
               key={idx}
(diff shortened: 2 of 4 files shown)


DevDW (approved):

bnussman (changes requested):
Just caught this:

Seems like we need to handle expired payment methods better in the drawer. I don't see the red text saying `Expired`. We also should add a Tooltip to the `Pay Now` button explaining that the card they select has expired. 

Use the MSW and edit `packages/manager/src/factories/accountPayment.ts` to simulate expired payment methods. 

If it is ok with you, we should address this asap (in this PR)

Screen Shot 2021-08-23 at 3 49 41 PM

bnussman:
Changes work well. Only issue I'm seeing now is the 3ed party payment method buttons are stacking. 
Screen Shot 2021-08-24 at 9 37 12 AM

HanaXu:
@bnussman they're not stacking for me, what resolution are you on? 🤔

tiffwong (commented):
Nice job fixing the expired styles! 

The buttons are also stacking for me and I'm at 1240px. If you remove the right padding on the Paypal button, it doesn't wrap anymore but that doesn't solve the buttons being different widths 🤔

bnussman (approved):

HanaXu:
@bnussman @tiffwong I still haven't been able to reproduce the stacking even with the MSW, so I just created a ticket (M3-5419) if one of yall want to address this in a future PR

PR #7877, opened 2021-08-23
M3-5418 Swap shared and dedicated cpu tabs

ndukatz:
Swaps the order of the “Shared CPU” and “Dedicated CPU” tabs. I was initially going to make dedicated the standard, but api still has shared as standard. 

When you go to Create Linode, Resize Linode, Clone Linode, or Create LKE Cluster “Dedicated CPU” should be the first tab, selected by default. Test creating, resizing, cloning, creating LKE cluster flows as well just to be safe.

I will be adding some test updates to this pr as well.

### DIFF
--- a/packages/manager/cypress/integration/linodes/clone-linode.spec.ts
+++ b/packages/manager/cypress/integration/linodes/clone-linode.spec.ts
@@ -6,6 +6,7 @@ import {
 import {
   containsClick,
   containsVisible,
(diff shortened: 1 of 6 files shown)


DevDW (approved):
🚢

tiffwong (approved):
🚀

johnwcallahan (approved):
Nice work.

PR #7881, opened 2021-08-24
M3-5408: Add EU Contract Component to creation flows

bnussman:
## Description

- Adds the `<EUAgreementCheckbox />` component to the following creation flows
  - Linode Create
  - LKE Cluster Create
  - NodeBalancer Create
  - Volume Create
  - OBJ Bucket Create
  - Image Create
  - Linode Migrate
  - Linode Disk/Config Clone
- The  `<EUAgreementCheckbox />` component should only be rendered when a user chooses an EU region as the targeted region. 
- Adds `isEURegion(region: string): boolean` function to help conditionally render component

## How to test

- For now, just check for styling and rendering when choosing an EU region, functionality is to come

### DIFF
--- a/packages/manager/src/components/CheckoutBar/CheckoutBar.tsx
+++ b/packages/manager/src/components/CheckoutBar/CheckoutBar.tsx
@@ -14,6 +14,7 @@ interface Props {
   submitText?: string;
   children?: JSX.Element;
   footer?: JSX.Element;
+  agreement?: JSX.Element;
 }
 
 type CombinedProps = Props;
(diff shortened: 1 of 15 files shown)


tiffwong (commented):
Great job getting this up!

Couple things I found:

For Create Volumes and Linode Migrations, the create button gets pushed to the left when it wraps at < 960px. The button should stay right aligned after wrapping to make it consistent.

Screen Shot 2021-08-25 at 3 49 24 PM

Not related to this PR but on a similar note, we should right align the button for cloning a disk.
Screen Shot 2021-08-25 at 3 51 36 PM

For Creating an Image, the text below the EU agreement wraps at 80% so we should make the EU agreement text also do the same thing.
Screen Shot 2021-08-25 at 3 52 32 PM

HanaXu (commented):
We should align the text with the checkbox on desktop and mobile:

Screen Shot 2021-08-25 at 11 54 04 AM

Screen Shot 2021-08-25 at 11 58 06 AM

PR #7886, opened 2021-08-26
M3-5427: Vertically center icons and texts in dropdown

tiffwong:
## Description

Remove spacing related styles that were throwing off the vertical alignment

## How to test

Check dropdown components throughout the app

### DIFF
--- a/packages/manager/src/components/EnhancedSelect/Select.styles.ts
+++ b/packages/manager/src/components/EnhancedSelect/Select.styles.ts
@@ -111,6 +111,9 @@ export const styles = (theme: Theme) =>
         cursor: 'pointer',
         padding: '10px',
         fontSize: '0.9rem',
+        '& svg': {
+          marginTop: 2,
+        },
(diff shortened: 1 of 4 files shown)


HanaXu (approved):

codyfinn:
I can't seem to remember how to get the exact measurements but here is what I am seeing on my end.
Screen Shot 2021-08-30 at 12 11 18 PM
I think the spacing difference between the top and the bottom of the flag icon is pretty noticeable.

DevDW (commented):
I'm observing the same thing as Cody–probably has something to do with the specific styling on the flags? The Images dropdown for choosing a distribution, which is similar in terms of the icon + text pairing, looks good to me

tiffwong:
@codyfinn @DevDW the recent changes should fix the spacing in Firefox, lmk if it looks good to you

DevDW (approved):
Looks good! 🚢

PR #7888, opened 2021-08-27
M3-4279: Remove duplicate text truncation helpers

tiffwong:
## Description

Combine both text truncation utilities into one file. The original intent was to get rid of `truncateText.ts` but I decided to merge the two instead since it breaks the text at the end of the word whereas the functions in `truncate.ts` breaks the text mid-word

## How to test

Check everywhere that `truncateText` is be called to make sure there's no regression


packages/manager/src/utilities/truncate.ts
```diff
@@ -1,3 +1,31 @@
...
+    // the string there
+    for (const letter of charsAfterMax) {
+      // A space means we're at the end of the word so break out of
+      // this loop
+      if (letter.match(/\W/)) {
+        break;
+      }
+      result.push(letter);
+    }
+    return `${result.join('')}...`;
```
DevDW, line 23: There are varying ideas on the specifics of how to use ellipses (spaces between each ellipsis mark vs. no spaces, for example), which we don't need to worry about here, but at a minimum we'll want to keep a space before the ellipsis to keep things grammatically correct.

packages/manager/src/utilities/truncate.ts
```diff
@@ -1,3 +1,31 @@
+/**
+ * Truncate a string and add an ellipsis at the end but ensures the
+ * text ends on a word rather than mid-word.
+ */
+export const truncate = (str: string, maxLength: number) => {
+  if (str.length > maxLength + 3) {
```
DevDW, line 6: We should keep the `+ 4` from the original function (see next comment)

DevDW (commented):
Good idea to consolidate these related utilities 👍

bnussman (approved):
Did not find any regressions!

DevDW (approved):
LGTM, but where did the `package-lock.json` change come from?

tiffwong:
@DevDW no idea but I just removed it 😂

PR #7889, opened 2021-08-27
M3-5317: Preserve Ticket Reply in LocalStorage

codyfinn:
## Description

When a user starts to write a reply to a support ticket the ticket ID and the user's response is stored in LocalStorage. This mimics the behavior of opening a support ticket. Unlike opening a support ticket a user may have multiple ticket replies to edit. When a user opens a ticket the application checks to see if `LocalStorage` has an entry in the `ticketReply` entry that has a `ticketId` that matches the support ticket the user is replying to. If the `ticketId` matches then it will prefill the reply text box with what is stored in `LocalStorage`. If the `ticketId` does not match what the user types will overwrite what is currently stored in `LocalStorage`.

## How to test

There are 4 key functionalities to test

- Navigate to a support ticket and start to type a reply then leaving before submitting. Navigate to that same support ticket the text box should be prefilled with the text that was written before you navigated away.
- Navigate to a support ticket and start to type a reply then leaving before submitting. Navigate to a different support ticket. The text box should not be prefilled with any text from the previous support ticket.
- Navigate to a support ticket and start to type a reply then without submitting the reply allow the session to expire. Start a new session and navigate to the same support ticket. The text box should be prefilled with the same text that was there before the session expired.
- Navigate to a support ticket and start to type a reply. Then without submitting the reply manually log out. Log back in and navigate to the same support ticket. The text box should not contain any of the text that was previously there before you logged out.

### DIFF
--- a/packages/manager/src/features/Support/SupportTicketDetail/TabbedReply/ReplyContainer.tsx
+++ b/packages/manager/src/features/Support/SupportTicketDetail/TabbedReply/ReplyContainer.tsx
@@ -17,6 +17,8 @@ import { FileAttachment } from '../../index';
 import Reference from './MarkdownReference';
 import ReplyActions from './ReplyActions';
 import TabbedReply from './TabbedReply';
+import { storage } from 'src/utilities/storage';
+import { debounce } from 'throttle-debounce';
 
 const useStyles = makeStyles((theme: Theme) => ({
   replyContainer: {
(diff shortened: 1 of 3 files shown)


bnussman (approved):
All functionality worked well and as expected! 🚀 Only downfall is that it only persists the reply for the last edited ticket like you explained, but I think this is fine. Most users won't be responding to many tickets simultaneously.

DevDW (approved):
LGTM! 👍

PR #7856, opened 2021-08-12
M3-5273: Update Button Placements and Styles (includes Migration modal fixes)

tiffwong:
## Description

Wait for #7846 to be merged first in case of conflicts

Right align button for:
- Linode Resize modal
- Linode Rebuild modal
- Linode Rescue modal
- Linode Migrate modal (also included some styling changes to better show the hierarchy) 

Switch order and right align buttons for:
- Linode Detail → Network tab → IP Transfer
- Linode Detail → Network tab → IP Sharing
- Linode Detail → Network tab → Add an IP Address
- Linode Detail → Storage tab → Create Volume
- Domains → Edit

## How to test

Test the modals and drawers mentioned above.

PR #7846, opened 2021-08-10
M3-5274 & M3-5275: Updated Button Placement and Styles

bnussman:
## Notice

This PR will right align all Action Buttons in drawers, but I did not update the order in some drawers. @tiffwong will have a PR that fixes other drawers button placement. 

## Description

- [x] My Profile → SSH Keys → Add SSH Key
  - Switch order of buttons and right align
- [x] My Profile → LISH Console Settings
  - Make ‘Add SSH Public Key’ Secondary button style
- [x] My Profile → API Tokens → Add Personal Access Token
  - Switch order of buttons and right align
- [x] My Profile → API Tokens → View Scopes drawer
  - Remove ‘Done’ button
- [x] My Profile → API Tokens → Edit Personal Access Token drawer
  - Switch order of buttons and right align
- [x] My Profile → OAuth Apps → Create OAuth App, Edit OAuth App
  - Switch order of buttons and right align
- [x] Managed → Monitors → Add Monitor drawer
  - Switch order of buttons and right align
- [x] Managed → Credentials → Add Credential drawer
  - Switch order of buttons and right align
- [x] Managed → SSH Access → Edit SSH Access drawer
  - [x] Fix button right padding since there is only one child
- [x] Managed → Contacts → Add Contact drawer
  - Right align button
- [x] Volumes → Edit Volume Drawer
  - Switch order of buttons and right align
- [x] Volumes landing → Resize, Clone, Attach
  - Switch order of buttons and right align
- [x] Volumes → Configuration drawer
  - Remove close button
- [x] Firewalls detail → Add Rule, Edit Rule, Create Firewall
  - Switch order of buttons and right align
- [x] Firewalls detail → Add Inbound Rule, Add Outbound Rule buttons
  - Should be Primary Buttons
- [x] StackScripts → Edit
  - Switch order of buttons and right align
- [x] Users & Grants → Add a User
  - Switch order of buttons and right align
- [x] Account → Settings → Object Storage, Close Account
  - Make both buttons Secondary with Outline
- [x] Account → User Profile
  - Change ‘Delete’ to Secondary with Outline

## How to test

- Test all affected components, all drawers, and all Dialogs

PR #7897, opened 2021-08-30
M3-5434: Fix remaining Button position issues in Drawers and Modals

bnussman:
## Description

- Fixes Numerous Bugs for `/account/users`
  - If you started changing your own user's username, the Delete button would become not diabled, which is bad becuase a user should not be able to delete themselves here 
  - Fixed changing a user's username. This use to cause an error upon save.
- #7856 and #7846 right aligned all buttons in drawers and modals and updated the order accordingly. Some drawers were missed in those PRs so this PR intends to fix any drawers or modals we missed

Button Alignment Fixes:
- `packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/EditRDNSDrawer.tsx`
  - Buttons were right aligned but order was wrong
- `packages/manager/src/features/Images/ImagesDrawer.tsx`
  - Buttons were right aligned but order was wrong
- `packages/manager/src/features/ObjectStorage/BucketDetail/ObjectDetailsDrawer.tsx`
  - Right aligned save button for updating a specific object's access level
- `packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx`
  - Right align and change order of `Cancel` and `Save Changed`/`Add Configuration` modal action buttons

## How to test

- Check the Components listed above

### DIFF
--- a/packages/manager/src/components/ActionsPanel/ActionsPanel.tsx
+++ b/packages/manager/src/components/ActionsPanel/ActionsPanel.tsx
@@ -9,7 +9,7 @@ import {
 } from 'src/components/core/styles';
 import RenderGuard from 'src/components/RenderGuard';
 
-type ClassNames = 'root';
+type ClassNames = 'root' | 'rightAlign';
 
 const styles = (theme: Theme) =>
   createStyles({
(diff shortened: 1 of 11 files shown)


tiffwong (approved):
Great job! 🎉

HanaXu (commented):
Can we change the button alignment for `/account/users/{username}/permissions` back to the left? It's not a drawer nor modal

HanaXu (approved):
👌 Maybe swap the order back as well for `/permissions`

bnussman:
@HanaXu Looking back on the feedback, in a follow up PR we should actually right align and change the order. Although this is not a modal or drawer, this is now standard to right align buttons like these. Here is an example. Just for the sake of merging this, I reverted the changes for `/account/users/{useranme}/permissions`, but I think we should make the update eventually. 

Screen Shot 2021-09-15 at 5 40 07 PM

PR #7898, opened 2021-08-30
Feature flag image price estimation

codyfinn:
## Description

Un-comment the image price estimation and put it behind the `imagesPriceInfo` feature flag.

## How to test

If the `imagesPriceInfo` flag is overridden then the price estimator for copying a disk image should be shown on the `ImageCreateTab` component.

### DIFF
--- a/packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
+++ b/packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
@@ -212,8 +212,8 @@ export const CreateImageTab: React.FC<Props & ImagesDispatch> = (props) => {
         handleChange={(linode) => handleLinodeChange(linode.id)}
         filterCondition={(linode) =>
           availableLinodesToImagize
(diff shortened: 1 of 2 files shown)


packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
```diff
@@ -231,6 +234,14 @@ export const CreateImageTab: React.FC<Props & ImagesDispatch> = (props) => {
...
           data-qa-disk-select
         />
+        {isImagePricingEnabled ? (
+          <Typography className={classes.helperText} variant="body1">
+            {`Estimated: ${calculateCostFromUnitPrice(
+              0.1,
+              selectedDiskSizeInGB
+            )}/month`}
+          </Typography>
+        ) : undefined}
```
DevDW, line 244: I don't see it making a difference, but we normally go with `null` here

DevDW (approved):

johnwcallahan (approved):

PR #7900, opened 2021-08-31
M3-5402: Functionality for EU Contract on NodeBalancer Create

bnussman:
## Preview

Screen Shot 2021-08-31 at 12 32 27 AM

## Description

- Functionality for EU Contract on NodeBalancer Create
  - The component should only be visible if the user has not signed the EU MC contract and is not a restricted user (restricted users have no access to view agreements)
  - The "Create NodeBalancer" button should be disabled unless the the EU MC confirmation is checked.
  - After the EU MC confirmation has been checked, when the user clicks "Create NodeBalancer", a POST to /account/agreements signing the eu model contract should be made. 

### TODO
  - [x] Agreements Factory
  - [x] Mutate React Query Cache when agreement is signed

## How to test

- Go to `/nodebalancers/create`
- Try to create a Nodebalancer in a non-EU region
  - You should not see an Agreement Checkbox and you should be able to create a Nodebalancer as per usual
- Try to create a Nodebalancer in an EU region
  - Try creating a Nodebalancer in an EU region as a restricted user
    - You should be able to create as normal but you should not see the agreement checkbox. Restricted users will not have to agree to this agreement

- MSW code for mocking an API error when signing (put in serverHandlers.ts)
```ts
  rest.post('*/account/agreements', (req, res, ctx) => {
    return res(ctx.status(500), ctx.json({ reason: 'Unknown error' }));
  }),
```

### DIFF
--- a/packages/manager/src/features/NodeBalancers/NodeBalancerCreate.tsx
+++ b/packages/manager/src/features/NodeBalancers/NodeBalancerCreate.tsx
@@ -305,7 +305,7 @@ class NodeBalancerCreate extends React.Component<CombinedProps, State> {
     });
   };
 
-  createNodeBalancer = async () => {
+  createNodeBalancer = () => {
     const {
       nodeBalancerActions: { createNodeBalancer },
     } = this.props;
--- a/packages/manager/src/features/NodeBalancers/NodeBalancerCreate.tsx
+++ b/packages/manager/src/features/NodeBalancers/NodeBalancerCreate.tsx
@@ -65,6 +65,7 @@ import {
   queryKey,
   updateAccountAgreementsData,
 } from 'src/queries/accountAgreements';
+import { withSnackbar, WithSnackbarProps } from 'notistack';
 
 type ClassNames = 'title' | 'sidebar';
 
@@ -91,7 +92,8 @@ type CombinedProps = WithNodeBalancerActions &
   WithRegions &
   RouteComponentProps<{}> &
   WithStyles<ClassNames> &
-  AgreementsProps;
+  AgreementsProps &
+  WithSnackbarProps;
 
 interface NodeBalancerFieldsState {
   label?: string;
@@ -306,8 +308,9 @@ class NodeBalancerCreate extends React.Component<CombinedProps, State> {
     });
   };
 
-  createNodeBalancer = () => {
+  createNodeBalancer = async () => {
     const {
+      enqueueSnackbar,
       nodeBalancerActions: { createNodeBalancer },
     } = this.props;
     const { nodeBalancerFields } = this.state;
@@ -325,12 +328,20 @@ class NodeBalancerCreate extends React.Component<CombinedProps, State> {
     this.setState({ submitting: true, errors: undefined });
 
     if (this.state.agree) {
-      queryClient.executeMutation({
-        variables: { eu_model: true, privacy_policy: true },
-        mutationFn: signAgreement,
-        mutationKey: queryKey,
-        onSuccess: updateAccountAgreementsData,
-      });
+      try {
+        await queryClient.executeMutation({
+          variables: { eu_model: true, privacy_policy: true },
+          mutationFn: signAgreement,
+          mutationKey: queryKey,
+          onSuccess: updateAccountAgreementsData,
+        });
+      } catch (error) {
+        this.setState({ submitting: false });
+        enqueueSnackbar('Error signing agreement.', {
+          variant: 'error',
+        });
+        return;
+      }
     }
 
     createNodeBalancer(nodeBalancerRequestData)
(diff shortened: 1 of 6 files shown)


packages/manager/src/features/NodeBalancers/NodeBalancerCreate.tsx
```diff
@@ -705,11 +732,20 @@ class NodeBalancerCreate extends React.Component<CombinedProps, State> {
               } Summary`}
               onDeploy={this.createNodeBalancer}
               calculatedPrice={10}
-              disabled={this.state.submitting || this.disabled}
+              disabled={
+                this.state.submitting ||
+                this.disabled ||
+                (showAgreement &&
+                  !this.state.agree &&
+                  !profile.data?.restricted)
```
bnussman, line 740: Should be able to remove 

```ts
!profile.data?.restricted
```

packages/manager/src/features/NodeBalancers/NodeBalancerCreate.tsx
```diff
@@ -474,11 +495,17 @@ class NodeBalancerCreate extends React.Component<CombinedProps, State> {
...
-    const { classes, regionsData } = this.props;
+    const { classes, regionsData, agreements, profile } = this.props;
     const { nodeBalancerFields } = this.state;
     const hasErrorFor = getAPIErrorFor(errorResources, this.state.errors);
     const generalError = hasErrorFor('none');
 
+    const showAgreement = Boolean(
+      isEURegion(nodeBalancerFields.region) &&
+        !profile.data?.restricted &&
+        agreements.data?.eu_model === false
```
HanaXu, line 506: !agreements.data?.eu_model

packages/manager/src/features/NodeBalancers/NodeBalancerCreate.tsx
```diff
@@ -312,6 +324,15 @@ class NodeBalancerCreate extends React.Component<CombinedProps, State> {
     /* Clear config errors */
     this.setState({ submitting: true, errors: undefined });
 
+    if (this.state.agree) {
+      queryClient.executeMutation({
+        variables: { eu_model: true, privacy_policy: true },
+        mutationFn: signAgreement,
+        mutationKey: queryKey,
+        onSuccess: updateAccountAgreementsData,
```
HanaXu, line 332: We should handle errors too

HanaXu (approved):
LGTM! Left some minor comments ☑️

packages/manager/src/queries/accountAgreements.ts
```diff
@@ -11,3 +15,28 @@ export const useAccountAgreements = () =>
...
+      onSuccess: updateAccountAgreementsData,
+    }
+  );
+};
+
+/**
+ * To be used as an onSuccess function for a React Query
+ * Mutation to update the React Query Cache
+ * @param newData data returned by the API - in this case the API always returns {}
+ * @param variables varibles passed to the mutation to be sent to the API
```
johnwcallahan, line 32: Spelling of "variables".

packages/manager/src/queries/accountAgreements.ts
```diff
@@ -11,3 +15,28 @@ export const useAccountAgreements = () =>
...
+ */
+export const updateAccountAgreementsData = (
```
johnwcallahan, line 34: This could be made generic and added to `queries/base.ts` (it's similar to `creationHandlers()`.

packages/manager/src/features/NodeBalancers/NodeBalancerCreate.tsx
```diff
@@ -312,6 +327,23 @@ class NodeBalancerCreate extends React.Component<CombinedProps, State> {
...
+      try {
+        await queryClient.executeMutation({
+          variables: { eu_model: true, privacy_policy: true },
+          mutationFn: signAgreement,
+          mutationKey: queryKey,
+          onSuccess: updateAccountAgreementsData,
+        });
+      } catch (error) {
+        this.setState({ submitting: false });
+        enqueueSnackbar('Error signing agreement.', {
```
johnwcallahan, line 340: I think we should keep these errors generic ("There was an error creating your NodeBalancer.") since to the user's POV this is just one action.

packages/manager/src/features/NodeBalancers/NodeBalancerCreate.tsx
```diff
@@ -91,6 +103,7 @@ interface NodeBalancerFieldsState {
 }
 
 interface State {
+  agree: boolean;
```
johnwcallahan, line 106: Minor: this variable name could be more specific.

packages/manager/src/features/NodeBalancers/NodeBalancerCreate.tsx
```diff
@@ -324,22 +321,21 @@ class NodeBalancerCreate extends React.Component<CombinedProps, State> {
...
-
     createNodeBalancer(nodeBalancerRequestData)
       .then((nodeBalancer) => {
         this.props.history.push(`/nodebalancers/${nodeBalancer.id}/summary`);
         // GA Event
         sendCreateNodeBalancerEvent(
           `${nodeBalancer.label}: ${nodeBalancer.region}`
         );
+        if (signedAgreement) {
+          queryClient.executeMutation({
```
johnwcallahan, line 332: We may want to `catch` and report errors here... though we don't have to address that now.

johnwcallahan (approved):

PR #7901, opened 2021-08-31
M3-5400: EU contract functionality for Linode create

HanaXu:
## Description
Add functionality for the EU Contract on Linode Create.

- The component should only be visible if the user has not signed the EU MC contract and is not a restricted user (restricted users have no access to view agreements)
- The "Create Linode" button should be disabled unless the the EU MC confirmation is checked.
- After the EU MC confirmation has been checked, when the user clicks "Create Linode", a POST to /account/agreements signing the eu model contract should be made. 

TODO:
- [x] Mutate React Query Cache when agreement is signed (wait for [7900](https://github.com/linode/manager/pull/7900) to get merged)

## How to test
On an account that hasn't signed the eu agreement:
- Go to `/linodes/create`
- Select a non-EU region:
    - You should not see the EU agreement checkbox in the checkout bar
- Select an EU region:
   - You should see the EU agreement checkbox and the `Create Linode` button should be disabled until it is checked
- Check the checkbox and click `Create Linode`
   - You should see a POST request made to `/account/agreements`
   - If there was an error POSTing to the agreements endpoint, you should see an error toast
   - If there was an error POSTing to `/linode/instances` after a successful POST to agreements (e.g. on dev bc there is no availability in the EU region), the EU checkbox should no longer display

Restricted users should not be able to see the EU agreement checkbox regardless of whether the parent account has signed or not signed the agreement and should be able to create a Linode as normal.

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/LinodeCreateContainer.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/LinodeCreateContainer.tsx
@@ -476,6 +476,7 @@ class LinodeCreateContainer extends React.PureComponent<CombinedProps, State> {
       try {
         await signAgreement({
           eu_model: true,
+          privacy_policy: true,
         });
         queryClient.invalidateQueries(queryKey);
         this.setState({
--- a/packages/manager/src/features/linodes/LinodesCreate/LinodeCreateContainer.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/LinodeCreateContainer.tsx
@@ -47,8 +47,6 @@ import {
   getOneClickApps,
 } from 'src/features/StackScripts/stackScriptUtils';
 import { accountBackupsEnabled } from 'src/queries/accountSettings';
-import { queryClient } from 'src/queries/base';
-import { queryKey } from 'src/queries/accountAgreements';
 import { CreateTypes } from 'src/store/linodeCreate/linodeCreate.actions';
 import {
   LinodeActionsProps,
@@ -477,29 +475,6 @@ class LinodeCreateContainer extends React.PureComponent<CombinedProps, State> {
 
     this.setState({ formIsSubmitting: true, errors: [] });
 
-    if (signedAgreement) {
-      try {
-        await signAgreement({
-          eu_model: true,
-          privacy_policy: true,
-        });
-        queryClient.invalidateQueries(queryKey);
-        this.setState({
-          formIsSubmitting: false,
-          showAgreement: false,
-        });
-      } catch (err) {
-        return this.setState({
-          formIsSubmitting: false,
-          errors: [
-            {
-              reason: 'Unable to create Linode.',
-            },
-          ],
-        });
-      }
-    }
-
     const payload = { ..._payload };
 
     /**
@@ -607,6 +582,14 @@ class LinodeCreateContainer extends React.PureComponent<CombinedProps, State> {
     return request()
       .then((response: Linode) => {
         this.setState({ formIsSubmitting: false });
+
+        if (signedAgreement) {
+          // @TODO Use React Query Mutation
+          signAgreement({
+            eu_model: true,
+            privacy_policy: true,
+          });
+        }
 
         /** if cloning a Linode, upsert Linode in redux */
         if (createType === 'fromLinode') {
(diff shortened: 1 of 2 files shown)


bnussman (approved):
Looks good to me. I'm going to take some things you did such as
- await agreement signing before making the api request to make the entity
- block entity creation if signing the agreement fails
- give user a visual error if the agreement api call fails

and apply them to my PR.

One minor thing we both didn't cover was un-checking the agreement if the user switched to another region and then back to an EU region, but I don't think that is necessary.

Like you said, use the React Query Cache Mutation so we don't need to refetch the endpoint once that work is merged and available

packages/manager/src/features/linodes/LinodesCreate/LinodeCreateContainer.tsx
```diff
@@ -439,7 +471,24 @@ class LinodeCreateContainer extends React.PureComponent<CombinedProps, State> {
...
+        queryClient.invalidateQueries(queryKey);
+        this.setState({
+          showAgreement: false,
+        });
+      } catch (err) {
+        return this.props.enqueueSnackbar('Error signing agreement.', {
+          variant: 'error',
+        });
+      }
+    }
```
bnussman, line 490: Oops! Forgot something. Should we do
```ts
this.setState({ formIsSubmitting: true });
```
so we have some loading state for this API call. Remember to
```ts
this.setState({ formIsSubmitting: false });
```
if the API gives an error for POSTing the agreement data.

bnussman: Helper code for testing this 

```ts
  rest.post('*/account/agreements', async (req, res, ctx) => {
    const delay = (ms: number) => new Promise(res => setTimeout(res, ms));
    await delay(2000);
    return res(ctx.status(500), ctx.json({ reason: 'Unknown error' }));
  }),
  rest.get('*/account/agreements', (req, res, ctx) =>
    res(ctx.json({ eu_modal: false, privacy_policy: true }))
  ),
```

packages/manager/src/features/linodes/LinodesCreate/LinodeCreateContainer.tsx
```diff
@@ -439,7 +471,24 @@ class LinodeCreateContainer extends React.PureComponent<CombinedProps, State> {
     return getLabel(arg1, arg2, arg3);
   };
 
-  submitForm: HandleSubmit = (_payload, linodeID?: number) => {
+  submitForm: HandleSubmit = async (_payload, linodeID?: number) => {
+    if (this.state.showAgreement) {
```
bnussman, line 475: Logicly, does it make sense to make the API call if the Checkbox is showing, or if they checked the box? I chose to say `if(signedAgreement)` but I think we are accomplishing the same thing

bnussman: Made this change and changed default `signedAgreement` state value to false (was the boolean value from the API)

packages/manager/src/features/linodes/LinodesCreate/LinodeCreateContainer.tsx
bnussman: Added proper loading state

packages/manager/src/features/linodes/LinodesCreate/LinodeCreateContainer.tsx
```diff
@@ -457,14 +489,19 @@ class LinodeCreateContainer extends React.PureComponent<CombinedProps, State> {
...
+            errors: [
+              {
+                field: 'root_pass',
+                reason: passwordError,
+              },
+            ],
+          },
+          () => {
+            scrollErrorIntoView();
+          }
```
bnussman, line 503: Added this as a minor UX fix, make sure this is ok!

DevDW (commented):
Looks good, the only issue I observed was that with the mocks on and code in the MSW so that POSTs to `/account/agreements` fail, I never saw an error toast

bnussman:
@DevDW Now that we are POSTing after successful entity creation I think we will just silently fail if the agreement request fails.

DevDW (approved):

PR #7904, opened 2021-08-31
M3-5430: Banner for upcoming API maintenance

DevDW:
## Description
Adds a site-wide dismissible banner in Cloud for upcoming API maintenance events returned from Statuspage.

- [X] Fix and extend mocking for scheduled maintenances
- [X] Add `<APIMaintenanceBanner />` component
- [X] Add `apiMaintenance` feature flag & support for it
- [X] Include `<APIMaintenanceBanner />` in GlobalNotifications.tsx

To-Do:
- [x] Ensure dismissibility functions as intended in this case
- [ ] Add test (may opt to add this in a follow-up PR)
- [X] Cleanup

## How to test
This has to be tested in two separate situations:

a) Testing mocked Maintenances from Statuspage
1. Turn on your MSW & have the LD feature flag serve the tests
2. Observe: one site-wide banner for Linode Database Infrastructure
3. Uncomment the temp code in `useFlags.ts` to set the `apiMaintenance` flag to off/the state where there are no maintenance IDs we want to display banners for
4. Observe: the site-wide banner should not be rendered

b) Test persisted dismissibility of banner
1. Turn your MSW off & have the LD feature flag serve "Newark Maintenance"
2. Observe: a single site-wide banner for scheduled network maintenance (this is unrelated to API maintenance, but is intended as a proof-of-concept to pull an actual incident/maintenance from Statuspage into the new component and have it be dismissible)
3. Dismiss the banner. Navigating elsewhere in the app, refreshing the page, logging out and back in, etc. should not bring the banner back.

### DIFF
--- a/packages/manager/src/features/GlobalNotifications/GlobalNotifications.tsx
+++ b/packages/manager/src/features/GlobalNotifications/GlobalNotifications.tsx
@@ -1,10 +1,15 @@
 import * as React from 'react';
 import AbuseTicketBanner from 'src/components/AbuseTicketBanner';
+import useFlags from 'src/hooks/useFlags';
 import { useRegionsQuery } from 'src/queries/regions';
-import RegionStatusBanner from './RegionStatusBanner';
+import { APIMaintenanceBanner } from './APIMaintenanceBanner';
 import { EmailBounceNotificationSection } from './EmailBounce';
+import RegionStatusBanner from './RegionStatusBanner';
 
 const GlobalNotifications: React.FC<{}> = () => {
+  const flags = useFlags();
+  const apiMaintenanceIds = flags.apiMaintenance ?? [];
+
   const regions = useRegionsQuery().data ?? [];
 
   return (
@@ -12,6 +17,7 @@ const GlobalNotifications: React.FC<{}> = () => {
       <EmailBounceNotificationSection />
       <RegionStatusBanner regions={regions} />
       <AbuseTicketBanner />
+      <APIMaintenanceBanner apiMaintenanceIds={apiMaintenanceIds} />
     </>
   );
 };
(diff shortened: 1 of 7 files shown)


packages/manager/src/hooks/useFlags.ts
```diff
@@ -30,6 +30,7 @@ export const useFlags = () => {
     // !!! Override the CMR flag so that it's officially "released" !!!!
     // @todo: clean up all conditional logic based on this flag.
     cmr: true,
+    // apiMaintenance: [],
```
DevDW, line 33: will be removed prior to merging -- only present for ease of testing

DevDW:
Going ahead and opening this up for review; may add a test to this PR or in a follow-up

packages/manager/src/features/GlobalNotifications/GlobalNotifications.tsx
```diff
@@ -1,17 +1,23 @@
...
+  const apiMaintenanceIds = flags.apiMaintenance ?? [];
+
   const regions = useRegionsQuery().data ?? [];
 
   return (
     <>
       <EmailBounceNotificationSection />
       <RegionStatusBanner regions={regions} />
       <AbuseTicketBanner />
+      <APIMaintenanceBanner apiMaintenanceIds={apiMaintenanceIds} />
```
bnussman, line 20: Thoughts on conditionally rendering this? 

```ts
{apiMaintenanceIds.length > 0 ? <APIMaintenanceBanner apiMaintenanceIds={apiMaintenanceIds} /> : null}
```

One thing I'm finding interesting is the fact that this is the first component in Cloud to use `useMaintenanceQuery`. If we conditionally render this, then we can prevent a call to `status.linode.com` on initialization of Cloud Manager.

DevDW: That's a good point, I went ahead and made that change 🚀

HanaXu (approved):
LGTM 👌

For test A, I only saw one banner because the other one was commented out (line 109-118 in `statusPage.ts`). We might want to update the description for that

DevDW:
> LGTM 👌
> 
> For test A, I only saw one banner because the other one was commented out (line 109-118 in `statusPage.ts`). We might want to update the description for that

I commented that out for demo purposes on Friday but hadn't updated the PR description. That code is basically just there to demonstrate the logic can handle multiple IDs being passed in to generate the correct number of banners

bnussman (approved):

PR #7906, opened 2021-09-01
Fix Linode Maintenance Query

bnussman:
## Description

- Enables the getAllAccountMaintenace query by default
- Adds padding to the `Reason` column in the Maintenance Table (`/account/maintenance`)
- Parses any possible markdown returned in the reason field for the Maintenance Table (`/account/maintenance`)

## How to test

- Make sure all maintenance events are populated on the Linodes landing page
- Make sure `/account/maintenance` still works as expected (still is API paginated and sorted)
- Make sure a notice banner appears for a Linode having maintenance on the Linode Detail Page

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 ### Changed:
 - Labels’ “(required)” substring adjusted to normal weight
 
+### Fixed:
+- Populating all Linode maintenance events
+
 ## [2021-08-26] - v1.47.1
 
 ### Added:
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 ### Changed:
 - Labels’ “(required)” substring adjusted to normal weight
 
+### Fixed:
+- Populating all Linode maintenance events
```
DevDW, line 17: The phrasing here strikes me as if it's saying that populating all Linode maintenance events was a bug -- maybe we could say something like, "Not all Linode maintenance events being populated" since that was the bug?

DevDW (dismissed):
LGTM -- didn't observe any breaks, and I checked the specific user account referenced in the chat and the Maintenance page and Linodes landing table both look correct

codyfinn (approved):
🚢

DevDW (approved):

PR #7911, opened 2021-09-02
M3-5405: Functionality for EU Contract on Image Create

bnussman:
## Preview

Screen Shot 2021-09-02 at 12 57 00 PM

## Description

- Functionality for EU Contract on Image Create
  - The component should only be visible if
    - An EU region is selected
    - The user has not signed the EU MC contract
    - The user is not a restricted user (restricted users have no access to view agreements)
  - The "Create Image" button should be disabled unless the the EU MC confirmation is checked.
    - The dropzone and `Browse Files` button should be disabled if the checkbox is not checked 
  - After the EU MC confirmation has been checked, when the user clicks "Create Image", a POST to /account/agreements signing the eu model contract should be made. 

## How to test

- Go to `/images/create/disk` and `/images/create/upload`
- Try to create a Bucket in a non-EU region
  - You should not see an Agreement Checkbox and you should be able to create a Bucket as per usual
- Try to create a Image in an EU region
  - Try creating a Image in an EU region as a restricted user
    - You should be able to create as normal but you should not see the agreement checkbox. Restricted users will not have to agree to this agreement

### DIFF
--- a/packages/manager/src/features/Images/ImageUpload.tsx
+++ b/packages/manager/src/features/Images/ImageUpload.tsx
@@ -17,6 +17,10 @@ import Prompt from 'src/components/Prompt';
 import TextField from 'src/components/TextField';
 import { Dispatch } from 'src/hooks/types';
 import { useCurrentToken } from 'src/hooks/useAuthentication';
+import {
+  useAccountAgreements,
+  useMutateAccountAgreements,
+} from 'src/queries/accountAgreements';
 import { useGrants, useProfile } from 'src/queries/profile';
 import { useRegionsQuery } from 'src/queries/regions';
 import { redirectToLogin } from 'src/session';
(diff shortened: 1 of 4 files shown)


HanaXu (approved):

packages/manager/src/features/Images/ImageUpload.tsx
```diff
@@ -61,19 +65,28 @@ export interface Props {
 
 export const ImageUpload: React.FC<Props> = (props) => {
   const { label, description, changeLabel, changeDescription } = props;
+
   const { data: profile } = useProfile();
   const { data: grants } = useGrants();
+  const { data: agreements } = useAccountAgreements();
+  const { mutate: signAgreement } = useMutateAccountAgreements();
```
johnwcallahan, line 72: I mentioned this in your migrate PR but I think we'll want to use mutateAsync here.

packages/manager/src/features/Images/ImageUpload.tsx
```diff
@@ -113,7 +126,14 @@ export const ImageUpload: React.FC<Props> = (props) => {
...
   };
 
-  const uploadingDisabled = !label || !region || !canCreateImage;
+  const onSuccess = () => {
+    if (signedAgreement) {
+      signAgreement({ eu_model: true, privacy_policy: true });
+    }
+  };
+
+  const uploadingDisabled =
```
johnwcallahan, line 135: The part of FileUploader that determines the placeholder text should be looking at this value, I think (L406-407). Currently it's possible to be in a half-and-half state:

Screen Shot 2021-09-03 at 10 15 52 PM

johnwcallahan (commented):
Looks good, one non-blocking comment about the disabled state.

Similar to your earlier point about cloning configs/disks, I don't think we actually need this on the "Capture Image" tab, since selecting a Linode from an EU region means you've already been prompted earlier to sign the agreement.

tiffwong (approved):

DevDW (approved):
LGTM -- I agree the EU agreement checkbox can be removed from the "Capture Image" tab, though.

PR #7912, opened 2021-09-02
M3-5403: Functionality for EU Contract on Volumes Create

johnwcallahan:
## Description

**What does this PR do?**

Adds functionality for the GDPR agreement on the Volumes Create page.

## How to test

Things to test:
- Create a Volume normally without having signing the agreement
- Create a Volume singing the agreement
- Create a Volume singing the agreement with either request failing
- Other things

Use the MSW to mock an account needing to sign the agreement. Uncomment the /agreements POST in serverHandlers to simulate a failure.

### DIFF
--- a/packages/manager/src/features/Volumes/VolumeCreate/CreateVolumeForm.tsx
+++ b/packages/manager/src/features/Volumes/VolumeCreate/CreateVolumeForm.tsx
@@ -22,11 +22,21 @@ import { dcDisplayNames, MAX_VOLUME_SIZE } from 'src/constants';
 import withVolumesRequests, {
   VolumesRequests,
 } from 'src/containers/volumesRequests.container';
+import EUAgreementCheckbox from 'src/features/Account/Agreements/EUAgreementCheckbox';
+import LinodeSelect from 'src/features/linodes/LinodeSelect';
 import { hasGrant } from 'src/features/Profile/permissionsHelpers';
+import useFlags from 'src/hooks/useFlags';
+import {
+  reportAgreementSigningError,
+  useAccountAgreements,
+  useMutateAccountAgreements,
+} from 'src/queries/accountAgreements';
+import { useGrants, useProfile } from 'src/queries/profile';
 import { ApplicationState } from 'src/store';
 import { MapState } from 'src/store/types';
 import { Origin as VolumeDrawerOrigin } from 'src/store/volumeForm';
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/Volumes/VolumeCreate/CreateVolumeForm.tsx
```diff
@@ -206,11 +222,20 @@ const CreateVolumeForm: React.FC<CombinedProps> = (props) => {
           ? errors.config_id
           : undefined;
 
+        const showAgreement =
+          isEURegion(values.region) &&
+          !profile?.restricted &&
+          accountAgreements?.eu_model === false;
```
johnwcallahan, line 228: @bnussman I realized this needs to explicitly check for `false`, otherwise if `accountAgreements` isn't available (if the request fails for some reason) this condition will be true.

bnussman: Good catch. I'll make this change in the other PRs.

packages/manager/src/features/Volumes/VolumeCreate/CreateVolumeForm.tsx
```diff
@@ -158,6 +168,12 @@ const CreateVolumeForm: React.FC<CombinedProps> = (props) => {
           tags: tags.map((v) => v.value),
         })
           .then(({ filesystem_path, label: volumeLabel }) => {
+            if (hasSignedAgreement) {
+              updateAccountAgreements({ eu_model: true }).catch((err) => {
```
bnussman, line 172: I think we should send `{ eu_model: true, privacy_policy: true }` because the agreement copy includes both.

bnussman (approved):
Functionality is good, variable names are really good (we should make this consistent when we clean up), and error handling looks good.

tiffwong (approved):
I wasn't able to simulate a failure when I uncommented the code in `serverHandlers.ts` but everything else looks good to me!

packages/manager/src/features/Volumes/VolumeCreate/CreateVolumeForm.tsx
johnwcallahan: Sounds good, I'm going to merge as-is but let's discuss and make it consistent when we smooth everything out.

PR #7915, opened 2021-09-02
M3-5406: Functionality for EU Contract on Linode Migrate

bnussman:
## Description

- Functionality for EU Contract on Migrate Linode Modal
  - The component should only be visible if
    - An EU region is selected
    - The user has not signed the EU MC contract
    - The user is not a restricted user (restricted users have no access to view agreements)
  - The "Enter Migration Queue" button should be disabled unless the the EU MC confirmation is checked.
  - After the EU MC confirmation has been checked, when the user clicks "Enter Migration Queue", a POST to /account/agreements signing the eu model contract should be made. 

## How to test

- Go to any Linode and open the Migrate modal
- Try to migrate to  a non-EU region
  - You should not see an Agreement Checkbox and you should be able to migrate as per usual
- Try to migrate to a EU region

### DIFF
--- a/packages/manager/src/features/linodes/MigrateLanding/MigrateLinode.tsx
+++ b/packages/manager/src/features/linodes/MigrateLanding/MigrateLinode.tsx
@@ -20,6 +20,11 @@ import { displayType } from 'src/features/linodes/presentation';
 import useExtendedLinode from 'src/hooks/useExtendedLinode';
 import { useImages } from 'src/hooks/useImages';
 import { useTypes } from 'src/hooks/useTypes';
+import {
+  useAccountAgreements,
+  useMutateAccountAgreements,
+} from 'src/queries/accountAgreements';
+import { useProfile } from 'src/queries/profile';
 import { useRegionsQuery } from 'src/queries/regions';
 import { ApplicationState } from 'src/store';
 import { formatDate } from 'src/utilities/formatDate';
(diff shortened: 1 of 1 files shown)


HanaXu (approved):

johnwcallahan (approved):

packages/manager/src/features/linodes/MigrateLanding/MigrateLinode.tsx
```diff
@@ -80,6 +85,9 @@ const MigrateLanding: React.FC<CombinedProps> = (props) => {
   const { types } = useTypes();
   const linode = useExtendedLinode(linodeID);
   const { images } = useImages();
+  const { data: profile } = useProfile();
+  const { data: agreements } = useAccountAgreements();
+  const { mutate: signAgreement } = useMutateAccountAgreements();
```
johnwcallahan, line 90: I think we'll want to use `mutateAsync` so we can use promises and `catch` and report errors. What are your thoughts using `mutate`?

tiffwong (approved):

packages/manager/src/features/linodes/MigrateLanding/MigrateLinode.tsx
bnussman: Yes, good catch. I saw your error handling code and this should be `mutateAsync` so we can follow your pattern.

PR #7918, opened 2021-09-02
M3-5431 EU Model Contract agreement banner + notification 

johnwcallahan:
## Description

This adds two things:
1.  Banner indicating the user must accept the EU Model Contract agreement
2. Handling for the EU Model Contract agreement notification

The new modal component sits in the GlobalNotifications wrapper component, and context is used in order to control the component from across the app. This pattern is debatable and honestly I didn’t think about it as much as I would like since we’re moving quickly. If you have other suggestions I’d love to consider them.

I also did a slight refactor of the DismissibleBanner component to be able to use the functionality from another component.

**TODO**: ~Listen for the correct API notification (I just mocked it in for now)~

## How to test

I included a mock notification in serverHandlers so if you use the MSW you should see the banner/notification.

To test dismissibility you'll need to turn the MSW off and remove `!hasComplianceNotification` from L23 of ComplianceBanner.tsx.

To test end-to-end you'll need to use devenv... (haven't done this myself yet).

### DIFF
--- a/packages/manager/src/features/GlobalNotifications/ComplianceBanner.tsx
+++ b/packages/manager/src/features/GlobalNotifications/ComplianceBanner.tsx
@@ -34,8 +34,9 @@ const ComplianceBanner: React.FC<{}> = () => {
       >
         <Typography>
           Please review the compliance update for guidance regarding the EU
-          Standard Contractual Clauses and its application to user deployments
-          in Linode’s London and Frankfurt data centers.
+          Standard Contractual Clauses and its application to users located in
+          Europe as well as deployments in Linode’s London and Frankfurt data
+          centers
         </Typography>
         <Button
           buttonType="primary"
--- a/packages/manager/src/features/GlobalNotifications/ComplianceBanner.tsx
+++ b/packages/manager/src/features/GlobalNotifications/ComplianceBanner.tsx
@@ -13,7 +13,7 @@ const ComplianceBanner: React.FC<{}> = () => {
 
   const hasComplianceNotification =
     notifications.filter((thisNotification) => {
-      return thisNotification.message.match(/compliance update/gi);
+      return thisNotification.message.match(/eu-model/gi);
     }).length > 0;
 
   const { hasDismissedBanner, handleDismiss } = useDismissibleBanner(
--- a/packages/manager/src/features/GlobalNotifications/ComplianceBanner.tsx
+++ b/packages/manager/src/features/GlobalNotifications/ComplianceBanner.tsx
@@ -1,46 +1,51 @@
 import * as React from 'react';
-
-import DismissibleBanner from 'src/components/DismissibleBanner';
-import Grid from 'src/components/core/Grid';
-import Typography from 'src/components/core/Typography';
 import Button from 'src/components/Button';
-import ComplianceUpdateModal from './ComplianceUpdateModal';
+import Box from 'src/components/core/Box';
+import Typography from 'src/components/core/Typography';
+import { useDismissibleBanner } from 'src/components/DismissibleBanner/DismissibleBanner';
+import Notice from 'src/components/Notice';
+import { complianceUpdateContext } from 'src/context/complianceUpdateContext';
+import useNotifications from 'src/hooks/useNotifications';
 
 const ComplianceBanner: React.FC<{}> = () => {
-  const [dialogOpen, setDialogOpen] = React.useState(false);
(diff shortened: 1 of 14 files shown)


packages/manager/src/features/GlobalNotifications/ComplianceBanner.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { complianceUpdateContext } from 'src/context/complianceUpdateContext';
+import useNotifications from 'src/hooks/useNotifications';
+
+const ComplianceBanner: React.FC<{}> = () => {
+  const context = React.useContext(complianceUpdateContext);
+  const notifications = useNotifications();
+
+  const hasComplianceNotification =
+    notifications.filter((thisNotification) => {
+      return thisNotification.message.match(/compliance update/gi);
```
DevDW, line 16: The mocks didn't work for me for the global banner out-of-the-box because this substring is not in the `message` on the mocked notification, only in the JSX for the intercepted notification.

In `useFormattedNotifications.tsx` you do the match on "eu-model", which should work here as well

bnussman: Same feedback here, we need to know a matching that will work in both places.

packages/manager/src/features/GlobalNotifications/ComplianceUpdateModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const [error, setError] = React.useState('');
+  const [checked, setChecked] = React.useState(false);
+  const dispatch: Dispatch = useDispatch();
+
+  const {
+    mutateAsync: updateAccountAgreements,
+    isLoading,
+  } = useMutateAccountAgreements();
+
+  const handleClick = () => {
```
DevDW, line 27: Maybe rename to `handleAgree` or something similar?

packages/manager/src/features/GlobalNotifications/ComplianceUpdateModal.tsx
```diff
@@ -0,0 +1,92 @@
...
+    setError('');
+    updateAccountAgreements({ eu_model: true })
```
DevDW, line 29: In the other PRs it seemed that we were also passing in `privacy_policy: true`, do we need to do that here as well?

johnwcallahan:
Thanks for the feedback @DevDW and @bnussman! I consolidated the check for the notification and added a unit test.

packages/manager/src/features/GlobalNotifications/GlobalNotifications.tsx
```diff
@@ -3,15 +3,24 @@ import AbuseTicketBanner from 'src/components/AbuseTicketBanner';
...
   return (
     <>
       <EmailBounceNotificationSection />
       <RegionStatusBanner regions={regions} />
       <AbuseTicketBanner />
+      <ComplianceBanner />
+      <ComplianceUpdateModal
+        open={complianceModelContext.isOpen}
+        onClose={complianceModelContext.close}
+      />
```
HanaXu, line 23: Should we use context inside this component instead of passing it in as a prop?

packages/manager/src/features/GlobalNotifications/ComplianceUpdateModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+      <Typography>
+        Recent legal changes now require all users with deployments in Linode’s
+        London or Frankfurt data centers to be covered under the EU Standard
+        Contractual Clauses.
+      </Typography>
+      <Typography style={{ marginTop: 16 }}>
+        After your review, please click the box below if the agreement is
+        acceptable, or{' '}
+        <SupportLink
+          text="contact us"
```
HanaXu, line 82: "contact us" is a bit ambiguous. We seem to generally follow the CTA pattern "open a Support ticket"

packages/manager/src/features/GlobalNotifications/GlobalNotifications.tsx
johnwcallahan: Good idea!

bnussman (approved):
Looks good, found not issues. 

I don't know exactly what will be in the notification from the api, but before megere just make sure the condition will catch it!

```ts
notification.type === 'notice' && /eu-model/gi.test(notification.message)
```

DevDW (approved):
LGTM 👍

PR #7919, opened 2021-09-03
M3-5231: Reduce bottom padding for grouped display tags

bcmertz:
## Description

This PR reduces the padding between grouped display tags. Matthew and I settled on simply reducing the padding after taking a look at other stylistic changes. Note: MUI themes specify a spacing value of 8 * n (where n is the index)

Verified with Matthew 2px is what is wanted here

## How to test

Create a few Linodes, tag them in the details view, return to the Linodes screen, and group by tag

### DIFF
--- a/packages/manager/src/features/linodes/LinodesLanding/DisplayGroupedLinodes.tsx
+++ b/packages/manager/src/features/linodes/LinodesLanding/DisplayGroupedLinodes.tsx
@@ -34,7 +34,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     height: 'auto',
     '& td': {
       // This is maintaining the spacing between groups because of how tables handle margin/padding. Adjust with care!
-      padding: `${theme.spacing(2) + 4}px 0 ${theme.spacing(1) + 2}px`,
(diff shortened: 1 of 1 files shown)


johnwcallahan (changes requested):
The change looks great on smaller viewports. You'll need to update L45 as well since it applies padding for md and above breakpoints.

packages/manager/src/features/linodes/LinodesLanding/DisplayGroupedLinodes.tsx
```diff
@@ -34,7 +34,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     height: 'auto',
     '& td': {
       // This is maintaining the spacing between groups because of how tables handle margin/padding. Adjust with care!
-      padding: `${theme.spacing(2) + 4}px 0 ${theme.spacing(1) + 2}px`,
+      padding: `${theme.spacing(2) + 4}px 0 ${theme.spacing(0) + 2}px`,
```
tiffwong, line 37: If the spacing isn't a multiple of 8, you can just write it as the number since this is a remnant of when we used to have a compact toggle (ie. `padding: 20px 2px`)

PR #7922, opened 2021-09-03
M3-5446: Simplify copy on API Maintenance screen

DevDW:
## Description
Follow-up to https://github.com/linode/manager/pull/7870 based on feedback received during demo today to simplify the copy.

### DIFF
--- a/packages/manager/src/components/MaintenanceScreen.tsx
+++ b/packages/manager/src/components/MaintenanceScreen.tsx
@@ -38,15 +38,15 @@ export const MaintenanceScreen: React.FC<{}> = () => {
 
   const maintenanceCopy = (
     <Typography variant="h2" className={classes.errorHeading}>
-      We are currently undergoing scheduled maintenance.
+      We are undergoing scheduled maintenance.
     </Typography>
   );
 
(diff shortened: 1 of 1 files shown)


packages/manager/src/components/MaintenanceScreen.tsx
```diff
@@ -38,15 +38,15 @@ export const MaintenanceScreen: React.FC<{}> = () => {
...
   );
 
   const statusPageCopy = (
     <Typography className={classes.subheading}>
-      To stay updated on the status of the Linode Cloud Manager and API, you may
-      visit{' '}
-      <Link to="https://status.linode.com/">https://status.linode.com</Link>.
+      Visit{' '}
+      <Link to="https://status.linode.com/">https://status.linode.com</Link> for
+      status updates on the Linode Cloud Manager and API.
```
johnwcallahan, line 49: Minor: I think you could remove "status" here (certainly subjective and non-blocking).

johnwcallahan (approved):

tiffwong (approved):
Looks great! I also don't think "status" is needed in the copy

PR #7926, opened 2021-09-07
M3-5445: Change font size for Credentials and Contacts copy

tiffwong:
## Description

Change the font size of the copy to 14px and removed the `subtitle1` variant since it was applying additional styles that we don't need

## How to test

Go to `/managed` and click on the "Credentials" and "Contacts" tab to see the copy


bnussman (approved):
Looks more consistent now with the SSH Key access tab. The text on these pages have always looked a bit out of place to me, but this helps!

codyfinn (approved):
🚢

PR #7928, opened 2021-09-07
M30-5401: EU Contract Cluster Create

codyfinn:
## Description

Add's the ability to sign the EU contract agreement for the Create Kubernetes Cluster page.

## How to test

If a user who has not accepted the GDPR agreement tries to create a cluster in an EU region then:

- In the checkout bar there should be a check-box component that allows the customer to agree to the terms.
- When the check-box is not checked then the checkout button should be disabled.
- When the check-box is checked then the checkout button should be enabled.
- If the user has agreed to the GDPR agreement then the checkbox should not be visible.
- If the user selects a region that is not in the EU then the checkbox should not be visible.

### DIFF
--- a/packages/manager/src/features/Kubernetes/CreateCluster/CreateCluster.tsx
+++ b/packages/manager/src/features/Kubernetes/CreateCluster/CreateCluster.tsx
@@ -28,6 +28,7 @@ import { getMonthlyPrice } from '.././kubeUtils';
 import { PoolNodeWithPrice } from '.././types';
 import KubeCheckoutBar from '../KubeCheckoutBar';
 import NodePoolPanel from './NodePoolPanel';
+import { signAgreement } from '@linode/api-v4/lib/account';
 
 const useStyles = makeStyles((theme: Theme) => ({
   root: {
@@ -138,6 +139,7 @@ export const CreateCluster: React.FC<CombinedProps> = (props) => {
   const [version, setVersion] = React.useState<Item<string> | undefined>();
   const [errors, setErrors] = React.useState<APIError[] | undefined>();
   const [submitting, setSubmitting] = React.useState<boolean>(false);
+  const [hasAgreed, setAgreed] = React.useState<boolean>(false);
   const {
     data: versionData,
     isError: versionLoadError,
@@ -179,11 +181,18 @@ export const CreateCluster: React.FC<CombinedProps> = (props) => {
 
     createKubernetesCluster(payload)
       .then((cluster) => push(`/kubernetes/clusters/${cluster.id}`))
+      .then(() => {
+        signAgreement({ eu_model: hasAgreed });
+      })
       .catch((err) => {
         setErrors(getAPIErrorOrDefault(err, 'Error creating your cluster'));
         setSubmitting(false);
         scrollErrorIntoView();
       });
+  };
+
+  const toggleHasAgreed = () => {
+    setAgreed(!hasAgreed);
   };
 
   const addPool = (pool: PoolNodeWithPrice) => {
@@ -351,6 +360,8 @@ export const CreateCluster: React.FC<CombinedProps> = (props) => {
             createCluster,
             classes,
           ]}
+          hasAgreed={hasAgreed}
+          toggleHasAgreed={toggleHasAgreed}
         />
       </Grid>
     </Grid>
--- a/packages/manager/src/features/Kubernetes/CreateCluster/CreateCluster.tsx
+++ b/packages/manager/src/features/Kubernetes/CreateCluster/CreateCluster.tsx
@@ -200,9 +200,7 @@ export const CreateCluster: React.FC<CombinedProps> = (props) => {
       });
   };
 
-  const toggleHasAgreed = () => {
-    setAgreed(!hasAgreed);
-  };
+  const toggleHasAgreed = () => setAgreed((prevHasAgreed) => !prevHasAgreed);
 
   const addPool = (pool: PoolNodeWithPrice) => {
     setNodePools([...nodePools, pool]);
--- a/packages/manager/src/features/Kubernetes/CreateCluster/CreateCluster.tsx
+++ b/packages/manager/src/features/Kubernetes/CreateCluster/CreateCluster.tsx
@@ -28,7 +28,10 @@ import { getMonthlyPrice } from '.././kubeUtils';
 import { PoolNodeWithPrice } from '.././types';
 import KubeCheckoutBar from '../KubeCheckoutBar';
 import NodePoolPanel from './NodePoolPanel';
-import { signAgreement } from '@linode/api-v4/lib/account';
+import {
+  reportAgreementSigningError,
+  useMutateAccountAgreements,
+} from 'src/queries/accountAgreements';
 
 const useStyles = makeStyles((theme: Theme) => ({
   root: {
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/Kubernetes/KubeCheckoutBar/KubeCheckoutBar.tsx
```diff
@@ -28,23 +31,27 @@ export const KubeCheckoutBar: React.FC<Props> = (props) => {
...
       calculatedPrice={getTotalClusterPrice(pools)}
       isMakingRequest={submitting}
-      disabled={pools.length < 1}
+      disabled={pools.length < 1 || !hasAgreed}
       onDeploy={createCluster}
       submitText={'Create Cluster'}
       agreement={
-        isEURegion(region) ? (
-          <EUAgreementCheckbox checked={false} onChange={() => null} />
+        isEURegion(region) && !isRestrictedUser ? (
```
bnussman, line 53: We also don't want this to render if the user has signed the agreement already. So we only render the agreement when `agreements?.eu_model === false`.
```ts
!profile?.restricted && agreements?.eu_model === false && isEURegion(selectedRegion)
```

packages/manager/src/features/Kubernetes/KubeCheckoutBar/KubeCheckoutBar.tsx
```diff
@@ -28,23 +31,27 @@ export const KubeCheckoutBar: React.FC<Props> = (props) => {
...
     updatePool,
     region,
+    hasAgreed,
+    toggleHasAgreed,
   } = props;
 
   // Show a warning if any of the pools have fewer than 3 nodes
   const showWarning = pools.some((thisPool) => thisPool.count < 3);
 
+  const { _isRestrictedUser: isRestrictedUser } = useAccountManagement();
```
bnussman, line 41: Consider just using `useProfile`. `useAccountManagement` may have some unnecessary overhead.

packages/manager/src/features/Kubernetes/CreateCluster/CreateCluster.tsx
```diff
@@ -179,13 +181,20 @@ export const CreateCluster: React.FC<CombinedProps> = (props) => {
 
     createKubernetesCluster(payload)
       .then((cluster) => push(`/kubernetes/clusters/${cluster.id}`))
+      .then(() => {
+        signAgreement({ eu_model: hasAgreed });
+      })
```
bnussman, line 186: This may just be me, but the two `.then()`'s are sketching me out. Can we consolidate the `.then()`'s into one function?

Also, you can pull in new changes into this PR from `gdpr-compliance` and update the API call to use our React Query Mutation with this logic around it. 

```ts
import {
  reportAgreementSigningError,
  useAccountAgreements,
  useMutateAccountAgreements,
} from 'src/queries/accountAgreements';

...

const { mutateAsync: updateAccountAgreements } = useMutateAccountAgreements();

...

if (hasAgreed) {
      updateAccountAgreements({
        eu_model: true,
        privacy_policy: true,
      }).catch(reportAgreementSigningError);
}
```

bnussman (changes requested):

packages/manager/src/features/Kubernetes/CreateCluster/CreateCluster.tsx
codyfinn: The `.then()` is needed to ensure that the EU agreement happens after and only if the cluster creation was successful.

johnwcallahan: I believe the two `then()`s that @bnussman is referring to are:

1. the first `then()` which uses history.push to navigate the user to the cluster details page
2. the second `then()` which calls `signAgreement`

These can be consolidated to something like:

```ts
createKubernetesCluster(payload).then(cluster => {
  push(`/kubernetes/clusters/${cluster.id}
  if (hasAgreed) {
    updatedAccountAgreements({...}).catch(reportAgreementSigningError);
  }
}).catch( // handle cluster creation error

File: packages/manager/src/features/Kubernetes/KubeCheckoutBar/KubeCheckoutBar.tsx [hunk 4]
```diff
@@ -28,23 +32,30 @@ export const KubeCheckoutBar: React.FC<Props> = (props) => {
...
+    isEURegion(region) && !profile?.restricted && !agreements?.eu_model;
+
   return (
     <CheckoutBar
       data-qa-checkout-bar
       heading="Cluster Summary"
       calculatedPrice={getTotalClusterPrice(pools)}
       isMakingRequest={submitting}
-      disabled={pools.length < 1}
+      disabled={pools.length < 1 || !hasAgreed}
```
johnwcallahan, line 53: This needs to take `showGDPRCheckbox` into account as well. (Currently the button is disabled if you're a customer that doesn't need to sign the agreement)

johnwcallahan (changes requested):
One quick blocking comment, everything else lgtm.

packages/manager/src/features/Kubernetes/KubeCheckoutBar/KubeCheckoutBar.tsx
```diff
@@ -28,23 +32,30 @@ export const KubeCheckoutBar: React.FC<Props> = (props) => {
...
+    toggleHasAgreed,
   } = props;
 
   // Show a warning if any of the pools have fewer than 3 nodes
   const showWarning = pools.some((thisPool) => thisPool.count < 3);
 
+  const { data: profile } = useProfile();
+  const { data: agreements } = useAccountAgreements();
+  const showGDPRCheckbox =
+    isEURegion(region) && !profile?.restricted && !agreements?.eu_model;
```
bnussman, line 45: Can we change `!agreements?.eu_model` to `agreements?.eu_model === false`?

@johnwcallahan mentioned that this would be prefered so in the case `agreements` is undefined, the checkbox would show.

bnussman (approved):

bnussman (approved):

packages/manager/src/features/Kubernetes/CreateCluster/CreateCluster.tsx
```diff
@@ -178,14 +184,26 @@ export const CreateCluster: React.FC<CombinedProps> = (props) => {
...
       .catch((err) => {
         setErrors(getAPIErrorOrDefault(err, 'Error creating your cluster'));
         setSubmitting(false);
         scrollErrorIntoView();
       });
   };
 
+  const toggleHasAgreed = () => {
+    setAgreed(!hasAgreed);
```
HanaXu, line 204: Should we use the function updater pattern here?
```typescript
setAgreed((hasAgreed) => !hasAgreed)
```

packages/manager/src/features/Kubernetes/KubeCheckoutBar/KubeCheckoutBar.tsx
```diff
@@ -28,23 +32,37 @@ export const KubeCheckoutBar: React.FC<Props> = (props) => {
...
+  const { data: profile } = useProfile();
+  const { data: agreements } = useAccountAgreements();
+  const showGDPRCheckbox =
+    isEURegion(region) &&
+    !profile?.restricted &&
+    agreements?.eu_model === false;
+
+  const needsAPool = pools.length < 1;
+  const disableCheckout = Boolean(
+    needsAPool || (!hasAgreed && showGDPRCheckbox)
```
HanaXu, line 51: I don't think `needsAPool` needs to be a variable as it's kind of self explanatory

HanaXu (approved):
LGTM, just left two minor non-blocking comments! ☑️

johnwcallahan (approved):
Agreed with Hana's minor comments, looks good though.

packages/manager/src/features/Kubernetes/KubeCheckoutBar/KubeCheckoutBar.tsx
codyfinn: I think while it is very easy to understand the meaning of `pools.length < 1` putting it in a variable does improve the readability of the value of `disableCheckout` as well as shortening what is already a long variable declaration.

PR #7930, opened 2021-09-08
M3-5437: Fix validation saying "expiring too far in the future" when entering credit card expiration date

bnussman:
## Description

- Allows user to enter a **two digit** or **four digit** expiry date
- Users were getting `Expiry too far in the future.` when entering a four digit year (MM/YYYY).
- Added `parseExpiryYear = (expiryYear: string | undefined): string | undefined` function to parse user input (year only)
  - Added some unit tests for this

## How to test

- Test the validation for the Credit Card form in the **Add a Payment Method** Drawer
  - Test the **Expiry** field specifically
  - Test with a 4 digit expiry year
  - Test with a 2 digit expiry year
- `yarn test creditCard`

### DIFF
--- a/packages/manager/src/utilities/creditCard.test.ts
+++ b/packages/manager/src/utilities/creditCard.test.ts
@@ -1,10 +1,14 @@
+import { CreditCardSchema } from '@linode/validation';
 import { DateTime } from 'luxon';
-import { take } from 'ramda';
+import { take, takeLast } from 'ramda';
 import {
   formatExpiry,
   hasExpirationPassedFor,
   parseExpiryYear,
 } from './creditCard';
+
+const currentYear = new Date().getFullYear();
+const currentYearFirstTwoDigits = take(2, String(currentYear));
 
 describe('isCreditCardExpired', () => {
   describe('give today is 01/01/2019', () => {
@@ -63,8 +67,6 @@ describe('formatExpiry', () => {
 });
 
 describe('parseExpiryYear', () => {
-  const currentYearFirstTwoDigits = take(2, String(new Date().getFullYear()));
-
   [
     [undefined, undefined],
     ['2024', '2024'],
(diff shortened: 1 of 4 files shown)


packages/manager/src/utilities/creditCard.test.ts
```diff
@@ -56,3 +65,107 @@ describe('formatExpiry', () => {
...
+        card_number: '1111111111111111',
+        expiry: `09/${takeLast(2, String(currentYear + 1))}`,
+        cvv: '123',
+      },
+      result: true,
+    },
+    {
+      data: {
+        card_number: '1111111111111111',
+        expiry: `09/${takeLast(2, String(currentYear + 21))}`,
```
DevDW, line 107: Is 20 years in the future the API max for the expiry date? If so, might be good to add a comment explaining that & the choice of the numbers being added to the current year

packages/manager/src/utilities/creditCard.test.ts
```diff
@@ -56,3 +65,107 @@ describe('formatExpiry', () => {
...
+      result: 'Expiration year must not be in the past.',
+    },
+    {
+      data: {
+        card_number: '1111111111111111',
+        expiry: '00/24',
+        cvv: '123',
+      },
+      result: 'Expiration month must be a number from 1 to 12.',
+    },
```
DevDW, line 143: I think it's worth adding a case covering a 3-digit expiry year:

```
      data: {
        card_number: '1111111111111111',
        expiry: '04/999',
        cvv: '123',
      },
```

DevDW (approved):
Functionality LGTM 👌 Left a couple of comments on the test

packages/manager/src/utilities/creditCard.test.ts
```diff
@@ -56,3 +65,127 @@ describe('formatExpiry', () => {
...
+      },
+      result: true,
+    },
+    {
+      data: {
+        card_number: '1111111111111111',
+        // Credit Card expiry years can't be more than 20 years in the future.
+        // We also use currentYear to make sure this test does not fail in many
+        // years down the road.
+        // Using takeLast to similate a user entering the year in a 2 digit format.
```
DevDW, line 110: `simulate`

johnwcallahan (commented):
Were you ever able to reproduce the bug? 

Otherwise seems OK. We could fine tune the error messaging a little more, for example I think this should say "Expiration required" or something similar.

Screen Shot 2021-09-16 at 2 07 08 PM

bnussman:
@johnwcallahan I was not able to reproduce the bug we discussed.

I am looking into making the error more general ("Expiration required") without convoluting the Yup scheme since the month and year validation are done separately.

johnwcallahan (approved):

PR #7932, opened 2021-09-09
M3-5441: Fix Create Access Key button order

tiffwong:
## Description

Remove the cancel button, change the primary button label depending on the mode, and disable the primary button if no change has been detected in edit mode. Also includes **M3-5459** which involves renaming the drawer title from "Create an Access Key" to "Create Access Key".

## How to test

Go to `/object-storage/access-keys` and click "Create Access Key"
- The drawer title should be "Create Access Key"

Close the drawer and click "Edit Label" on any access key 
- The button should say "Save Changes"
- The button should also be disabled unless a change is detected

### DIFF
--- a/packages/manager/src/features/ObjectStorage/AccessKeyLanding/AccessKeyDrawer.tsx
+++ b/packages/manager/src/features/ObjectStorage/AccessKeyLanding/AccessKeyDrawer.tsx
@@ -76,6 +76,7 @@ export const AccessKeyDrawer: React.FC<CombinedProps> = (props) => {
   const { objectStorageBuckets: buckets } = useBuckets();
   const { data: accountSettings } = useAccountSettings();
   const hidePermissionsTable = buckets.data.length === 0;
+  const createMode = mode === 'creating';
 
   const [dialogOpen, setDialogOpen] = React.useState<boolean>(false);
   // This is for local display management only, not part of the payload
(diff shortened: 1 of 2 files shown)


bnussman (approved):
Awesome improvements 🚀

HanaXu (commented):
Is there a reason why we're removing the secondary button? 🤔

Secondary buttons are still present in the `Create OAuth App` drawer and `Add Personal Access Token` drawer

tiffwong:
@HanaXu good point, tbh I don't remember why we decided to remove it but I just added it back in 😅

HanaXu (approved):

packages/manager/src/features/ObjectStorage/AccessKeyLanding/AccessKeyDrawer.tsx
```diff
@@ -217,22 +217,25 @@ export const AccessKeyDrawer: React.FC<CombinedProps> = (props) => {
...
+                    disabled={
+                      isRestrictedUser ||
+                      (mode !== 'creating' &&
+                        values.label === initialLabelValue)
+                    }
+                    loading={isSubmitting}
+                    onClick={beforeSubmit}
+                    data-qa-submit
+                  >
+                    {mode === 'creating' ? 'Create Access Key' : 'Save Changes'}
```
DevDW, line 237: Since `mode === 'creating'` is checked multiple times, it'd be good to make a variable for it (e.g. `const creatingMode`)

DevDW (approved):

PR #7933, opened 2021-09-09
M3-5442: Linode Details: remove /128 prefix length from IPv6 address

codyfinn:
## Description

Removes the prefix length for IPv6 on the Linode Details Page

## How to test

Navigate to a Linode that has an IPv6 address. In the details header the IPv6 address should not include the length prefix `/128`

### DIFF
--- a/packages/manager/src/features/linodes/LinodeEntityDetail.tsx
+++ b/packages/manager/src/features/linodes/LinodeEntityDetail.tsx
@@ -117,9 +117,7 @@ const LinodeEntityDetail: React.FC<CombinedProps> = (props) => {
     progress = getProgressOrDefault(recentEvent);
     transitionText = _transitionText(linode.status, linode.id, recentEvent);
   }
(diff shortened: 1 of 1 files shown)


bnussman (approved):

packages/manager/src/features/linodes/LinodeEntityDetail.tsx
```diff
@@ -117,6 +117,9 @@ const LinodeEntityDetail: React.FC<CombinedProps> = (props) => {
     progress = getProgressOrDefault(recentEvent);
     transitionText = _transitionText(linode.status, linode.id, recentEvent);
   }
+  const trimmedIPv6 = linode.ipv6
+    ? linode.ipv6.replace('/128', '')
+    : linode.ipv6;
```
DevDW, line 122: Perhaps this can be simplified to 

`const trimmedIPv6 = linode.ipv6?.replace('/128', '') || null;` 

?

DevDW (approved):

PR #7935, opened 2021-09-09
M3-5448: Global changes to help icon

tiffwong:
## Description

Make help icons consistent throughout the app and remove styles that are no longer needed. 

Help icons should generally be:
- Gray (#888f91) by default and blue (#3683dc) on hover
- 20x20px

**Note:** Found a "(required)" regression in `MultipleIPInput` so added a required prop there to make it consistent.

## How to test

Check the help icons throughout the app.

### DIFF
--- a/packages/manager/src/components/ActionMenu/ActionMenu.tsx
+++ b/packages/manager/src/components/ActionMenu/ActionMenu.tsx
@@ -96,7 +96,6 @@ const useStyles = makeStyles((theme: Theme) => ({
     color: '#fff',
     '& :hover': {
       color: '#4d99f1',
-      backgroundColor: 'transparent',
     },
     padding: '0 0 0 8px',
     '& svg': {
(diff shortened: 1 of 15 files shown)


bnussman (approved):
Changes look good to me, my only thought is that now that the help icon is smaller, in some cases it looks like it is far from the related tooltip.

Screen Shot 2021-09-21 at 2 10 51 PM

DevDW (approved):
🚢

HanaXu (changes requested):
The "HelpIcon" is also used as an error icon for Google Pay but there is a regression here:
(M3-5457 was created to address making the component more generic)

Before:

image

After:

image

The icon looks a bit far in `profile/display` on a restricted user:

image


And looks a bit off in Linode -> Resize:

image

HanaXu (approved):
🚀

PR #7942, opened 2021-09-10
Change: raise threshold for large account

johnwcallahan:
## Description

This raises the large account threshold so search filters work for more accounts. See PDI-2071 for more details, and let's discuss offline.


bnussman (approved):
Scarry, but seemed to do fine for large accounts I tested on!

ndukatz (approved):
Seems to work well. I was able to test with over 500 with MSW, but I could not verify that it works with 1500 or that it does not work with 1500+. Either the mock service worker or the site could not handle that many linodes. It kept crashing when I tried to test with that many.

PR #7945, opened 2021-09-13
Disable Glish on Bare Metal Machines

bcmertz:
## Description

This disables glish for bare metal machines by checking if the current Linode is bare metal

## How to test

First you must authorize yourself for bare metal then create a bare metal linode and a non bare metal linode. Verify that the bare metal one has no glish and the non bare metal one does

Note: Not important but I went down a bunch of dead ends trying to use src/hooks/useExtendedLinode but because Lish is not a FC that didn't work. Then I thought about passing in the data from store.getState() in src/index.tsx because I misguidedly thought that perhaps Lish being a separate popup couldn't access the store and needed it passed in without realizing api calls work. TLDR - read documentation 🙃

### DIFF
--- a/packages/manager/src/features/Lish/Lish.tsx
+++ b/packages/manager/src/features/Lish/Lish.tsx
@@ -254,7 +254,7 @@ class Lish extends React.Component<CombinedProps, State> {
                 />
               </SafeTabPanel>
             )}
-            {linode && token && isBareMetal && (
+            {linode && token && !isBareMetal && (
               <SafeTabPanel index={1} data-qa-tab="Glish">
                 <Glish
                   token={token}
--- a/packages/manager/src/features/Lish/Lish.tsx
+++ b/packages/manager/src/features/Lish/Lish.tsx
@@ -107,6 +107,7 @@ class Lish extends React.Component<CombinedProps, State> {
         if (!this.mounted) {
           return;
         }
+
         this.setState({
           linode,
           loading: false,
@@ -196,10 +197,6 @@ class Lish extends React.Component<CombinedProps, State> {
       title: 'Weblish',
       routeName: `${this.props.match.url}/weblish`,
     },
-    {
-      title: 'Glish',
-      routeName: `${this.props.match.url}/glish`,
-    },
   ];
 
   matches = (p: string) =>
@@ -208,6 +205,22 @@ class Lish extends React.Component<CombinedProps, State> {
   render() {
     const { classes } = this.props;
     const { authenticated, loading, linode, token } = this.state;
+
+    const isBareMetal = linode && linode.type && linode.type.includes('metal');
+
+    if (!isBareMetal) {
+      this.tabs = [
+        /* NB: These must correspond to the routes inside the Switch */
+        {
+          title: 'Weblish',
+          routeName: `${this.props.match.url}/weblish`,
+        },
+        {
+          title: 'Glish',
+          routeName: `${this.props.match.url}/glish`,
+        },
+      ];
+    }
 
     const navToURL = (index: number) => {
       this.props.history.push(this.tabs[index].routeName);
@@ -241,7 +254,7 @@ class Lish extends React.Component<CombinedProps, State> {
                 />
               </SafeTabPanel>
             )}
-            {linode && token && (
+            {linode && token && isBareMetal && (
               <SafeTabPanel index={1} data-qa-tab="Glish">
                 <Glish
                   token={token}
--- a/packages/manager/src/features/Lish/Lish.tsx
+++ b/packages/manager/src/features/Lish/Lish.tsx
@@ -19,6 +19,7 @@ import TabLinkList from 'src/components/TabLinkList';
 import Typography from 'src/components/core/Typography';
 import ErrorState from 'src/components/ErrorState';
 import NotFound from 'src/components/NotFound';
+import { Tab } from 'src/components/TabLinkList/TabLinkList';
 import Glish from './Glish';
 import Weblish from './Weblish';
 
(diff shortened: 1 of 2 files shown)


packages/manager/src/features/Lish/Lish.tsx
```diff
@@ -241,7 +254,7 @@ class Lish extends React.Component<CombinedProps, State> {
                 />
               </SafeTabPanel>
             )}
-            {linode && token && (
+            {linode && token && isBareMetal && (
```
bnussman, line 257: I think this should be `!isBareMetal`:

```ts
{linode && token && !isBareMetal && (
```

packages/manager/src/features/Lish/Lish.tsx
```diff
@@ -209,6 +206,22 @@ class Lish extends React.Component<CombinedProps, State> {
...
 
+    const isBareMetal = linode && linode.type && linode.type.includes('metal');
+
+    if (!isBareMetal) {
+      this.tabs = [
+        /* NB: These must correspond to the routes inside the Switch */
+        {
+          title: 'Weblish',
+          routeName: `${this.props.match.url}/weblish`,
+        },
```
bnussman, line 217: Instead of repeating this code for the weblish tab, we could potentially use a spread operator. Maybe something like:

```ts
let tabs = [
  {
    title: 'Weblish',
    routeName: `${this.props.match.url}/weblish`,
  },
];

if (!isBareMetal) {
  tabs = [
    ...tabs,
    {
      title: 'Glish',
      routeName: `${this.props.match.url}/glish`,
    },
  ];
}
```

bnussman (changes requested):
I'm still seeing Glish for my Bare Metal Linode 😧. I'm not sure exactly why that is but I'm thinking it may have something to do with tabs being a variable and not part of the components state. I had success with making the tabs variable a variable that is local to the render method.  I'm sure there are numerous way to go about this. 

As a side note for testing PR's like this, we could consider adding a Bare Metal Linode to the Mock Service Worker

DevDW (commented):
I'm observing the same thing as @bnussman and agree with most of his suggestions. Another option for the `tabs` variable after moving it into the `render()` method is:

```
    const tabs = [
      /* NB: These must correspond to the routes inside the Switch */
      {
        title: 'Weblish',
        routeName: `${this.props.match.url}/weblish`,
      },
      !isBareMetal
        ? {
            title: 'Glish',
            routeName: `${this.props.match.url}/glish`,
          }
        : null,
    ].filter(Boolean) as Tab[];
```

where the `Tab` used for the type is imported from `TabLinkList.tsx`.

Perhaps some styling tweaks are also in order when the Glish tab is hidden (bottom white border)?

Screen Shot 2021-09-13 at 12 47 28 PM

packages/manager/src/features/Lish/Lish.tsx
bcmertz: good catch, must've mixed up my linodes while testing

bcmertz:
> As a side note for testing PR's like this, we could consider adding a Bare Metal Linode to the Mock Service Worker

Great idea, I'll add that to this PR!

packages/manager/src/features/Lish/Lish.tsx
bcmertz: thanks! I think it might be nice to keep it typed like @DevDW suggested

bcmertz:
> I'm observing the same thing as @bnussman and agree with most of his suggestions. Another option for the `tabs` variable after moving it into the `render()` method is:
> 
> ```
>     const tabs = [
>       /* NB: These must correspond to the routes inside the Switch */
>       {
>         title: 'Weblish',
>         routeName: `${this.props.match.url}/weblish`,
>       },
>       !isBareMetal
>         ? {
>             title: 'Glish',
>             routeName: `${this.props.match.url}/glish`,
>           }
>         : null,
>     ].filter(Boolean) as Tab[];
> ```
> 
> where the `Tab` used for the type is imported from `TabLinkList.tsx`.
> 
> Perhaps some styling tweaks are also in order when the Glish tab is hidden (bottom white border)?
> 
> Screen Shot 2021-09-13 at 12 47 28 PM

Done!

bnussman (approved):
Nice job addressing feedback! Looks good 💻

bcmertz:
> Nice job addressing feedback! Looks good 💻

thanks! I added a bare metal mock Linode, would you mind checking it? It appears the reason glish appears for the mocked bare metal is because api-v4/lib/linodes/linodes.js/getLinode 404s. Is this expected?

DevDW (approved):
Nice work! 🚢

bnussman:
That 404 would be expected. You'd have to add something like this below, where the id of the Linode `26` matches the id of the Bare Metal Linoide returned in the `*/linode/instances` mock.   

```ts
rest.get('*/instances/26', async (req, res, ctx) => {
  return res(
    ctx.json(
      linodeFactory.build({
        type: 'metal',
      })
    )
  );
}),
```

packages/manager/src/features/Lish/Lish.tsx
```diff
@@ -227,35 +233,40 @@ class Lish extends React.Component<CombinedProps, State> {
...
+      return <CircleProgress noInner className={classes.progress} />;
+    }
+
+    // Only show 404 component if we are missing _both_ linode and token
+    if (!loading && !linode && !token) {
+      return <NotFound className={classes.notFound} />;
+    }
+
     return (
       <React.Fragment>
```
DevDW, line 247: can add

`// eslint-disable-next-line react/jsx-no-useless-fragment`

above this line to make the linter happy:

Screen Shot 2021-09-16 at 5 00 30 PM

PR #7947, opened 2021-09-13
M3-5438: Fix bug preventing Linode Configuration changes

DevDW:
## Description
Currently, there's a bug preventing users from editing their Linode Configurations successfully without a workaround. Clicking "Save Changes" would look unresponsive.

Basically, non-VLAN network interfaces should be able to have `null` label fields in the schema (the appearance of a lack of response was because this schema-related error does not get surfaced in the UI -- you can edit `handleError` in `LinodeConfigDialog.tsx` to log errors to the console to observe the error prior to the fix in this PR). Making this adjustment makes the Config dialog function as intended.

## How to test
Try to:
- edit a Linode Configuration's IPAM address by itself and save the changes
- edit anything else in the config aside from the IPAM address and save the changes
- edit the IPAM address + other stuff in the dialog and save the changes

All should now work without incident.

### DIFF
--- a/packages/validation/src/linodes.schema.ts
+++ b/packages/validation/src/linodes.schema.ts
@@ -36,7 +36,7 @@ export const linodeInterfaceSchema = array()
               /[a-zA-Z0-9-]+/,
               'Must include only ASCII letters, numbers, and dashes'
             ),
-          otherwise: string().notRequired(),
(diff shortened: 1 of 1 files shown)


bnussman (approved):
Fixed 👍🏼🚀

HanaXu (approved):
LGTM! 👌

PR #7948, opened 2021-09-14
Upgrade Sentry SDK

bnussman:
## Description

- Updates `@sentry/browser` to the latest version `6.12.0`
- I'm not exactly sure why the lock file changed a ton but I think it is ok

## How to test

- Define `REACT_APP_SENTRY_URL` in your local env
  - To get the Sentry URL ask me or find it in Sentry (Cloud Manager -> Project Settings -> DSNs)
- Add some code in Cloud Manager anywhere that will call Sentry's captureException function to test that exceptions get sent to Sentry
```ts
import { captureException } from '@sentry/browser';

captureException("Test Sentry");
```
- Go to Sentry and verify that exceptions are captured as expected

### DIFF
--- a/packages/manager/package.json
+++ b/packages/manager/package.json
@@ -27,7 +27,7 @@
     "@reach/popover": "^0.10.5",
     "@reach/router": "^1.3.3",
     "@reach/tabs": "^0.10.5",
-    "@sentry/browser": "~5.20.1",
+    "@sentry/browser": "^6.12.0",
     "@types/markdown-it": "^10.0.2",
     "@types/react-beautiful-dnd": "^13.0.0",
     "@types/react-router-hash-link": "^1.2.1",
--- a/packages/manager/src/initSentry.ts
+++ b/packages/manager/src/initSentry.ts
@@ -9,6 +9,7 @@ export const initSentry = () => {
       dsn: SENTRY_URL,
       release: process.env.VERSION,
       environment: process.env.NODE_ENV,
+      autoSessionTracking: false,
       beforeSend,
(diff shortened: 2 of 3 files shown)


DevDW (commented):
I keep getting errors on the POST requests for some reason. When I check DevTools, the "Status" column on the Network tab says "blocked:other" for them, and in the console I see entries that say `net::ERR_BLOCKED_BY_CLIENT`

johnwcallahan (changes requested):
We'll want to disable auto session tracking, since v6 does this by default now: https://github.com/getsentry/sentry-javascript/blob/master/CHANGELOG.md#600

Also, I believe the excessive lockfile changes can be fixed by manually updating package.json then running `yarn install` from the root directory.

DevDW (commented):
I'm still seeing what I saw back in September, but interestingly, I also encountered an app crash on this branch when logging into a dev account that has Managed enabled 🤔 

Screen Shot 2021-11-17 at 11 58 33 AM

safe accessing the `[0]` and `.reason` seemed to fix it

edit: I was able to use `captureException` successfully but had to use an Incognito Chrome window to circumvent the blocks I was encountering. Ready to approve this once we figure out the crash issue

HanaXu (approved):
POSTing works for me. I am also seeing the same Managed error as Dajahi, approving pending the safe access change

bnussman:
I want to merge, but that Object()()[] error should not be possible. Will investigate further

bnussman:
Closing due to really odd dependency issues after bumping Sentry version

PR #7949, opened 2021-09-14
Removed Unused Dependencies

bnussman:
## Description

- Removes unused packages from `package.json`'s
- Updates `useAPIRequest.test.ts` to play better with hook testing

## How to test

 - Test everything including Unit Tests, Cypress, and Storybook


patthiel:
@bnussman we also have tests for components via selenium that are the files ending in `.spec.js` in each component directory that can be removed. @ndukatz @bnussman have you thought about testing components with storybook via cypress at all?

bnussman:
I'm so sorry 😢 @patthiel. I'm not dead set on merging this PR, just wanted to put it up as a proof of concept for cleaning up the repo. I think we have the ability to test storybook via cypress, but we just don't right now.

ndukatz (approved):
Looks Great! Thank you for doing this. I've been meaning to get to it for a while. All UI and unit tests pass.

packages/manager/package.json
```diff
@@ -249,19 +224,16 @@
     "postcss-loader": "^3.0.0",
     "prettier": "~2.2.1",
     "react-dev-utils": "^7.0.3",
-    "react-test-renderer": "~16.12.0",
     "redux-mock-store": "^1.5.3",
     "reselect-tools": "^0.0.7",
     "selenium-standalone": "^6.13.0",
-    "set-value": "^3.0.1",
```
bnussman, line 256: @johnwcallahan If we decide to merge this, this removes set-value as a dependency, will this cause problems with #7953 ?

johnwcallahan: I don't think it'd cause issues, (it would make "set-value" in `resolutions` obsolete).

PR #7952, opened 2021-09-15
Upgrade Axios again to fix app crashing bug

johnwcallahan:
## Description

There was a bug introduced in axios v0.21.2 that handled interceptors incorrectly. 

We upgraded to 0.21.2 here: https://github.com/linode/manager/pull/7937

The result was that the app would crash when the API returned an error, since we rely on an interceptor to transform Axios errors into `APIError[]`. 

See the axios changelog: https://github.com/axios/axios/blob/master/CHANGELOG.md

## How to test
Generally test that things are working throughout the app (happy paths and error paths).

### DIFF
--- a/packages/api-v4/package.json
+++ b/packages/api-v4/package.json
@@ -23,7 +23,7 @@
   "dependencies": {
     "@linode/validation": "0.4.0",
     "@types/yup": "^0.29.13",
-    "axios": "~0.21.2",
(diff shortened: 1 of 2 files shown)


HanaXu (approved):
LGTM, nice job! 🔍  🚀

ndukatz (approved):
This fixes the problem. All tests passed here

PR #7954, opened 2021-09-16
M3-5452: Add support for Autoscaling Node Pools

HanaXu:
## Description
Adds support for Autoscaling a Node Pool on a Kubernetes Cluster's summary page

## How to test
Go to `kubernetes/clusters/<cluser_id>/summary`, ensure the `autoscaler` feature flag is on

- You should see an `Autoscale Pool` button before the Resize Pool button
    - If the Autoscaler is on, you should see min/max values next to the autoscale pool button
    - If the Autoscaler is off, you should not see min/max values next to the autoscale pool button
- Clicking on Autoscale Pool should display a dialog
- The Save Changes button should be disabled until there has been a change in the form and there are no errors
- If Autoscaler is off, the Min/Max input fields should be disabled
- If the Minimum value is not between 1-99 or greater than the Maximum value, there should be an error displayed under the input field
- If the Maximum value is not between 1-100, there should be an error displayed under the input field
- If the Maximum value is less than the current Max, then a warning notice should display
    - Clicking on `resize` in the warning should close the modal and open the resize drawer for that pool
- Clicking Save Changes should make a request to `lke/clusters/<cluster_id>/pools/<pool_id>`
    - Any API errors should be displayed as a Notice error in the dialog
    - A successful request should close the dialog, pop a success toast and update the Min/Max values next to the Autoscale Pool button (might take a few seconds to update)

### DIFF
--- a/packages/manager/src/features/Kubernetes/KubernetesClusterDetail/NodePoolsDisplay/AutoscalePoolDialog.tsx
+++ b/packages/manager/src/features/Kubernetes/KubernetesClusterDetail/NodePoolsDisplay/AutoscalePoolDialog.tsx
@@ -3,16 +3,17 @@ import { AutoscaleNodePoolSchema } from '@linode/validation/lib/kubernetes.schem
 import { useFormik } from 'formik';
 import * as React from 'react';
 import * as classNames from 'classnames';
+import FormControlLabel from 'src/components/core/FormControlLabel';
+import { makeStyles, Theme } from 'src/components/core/styles';
+import Typography from 'src/components/core/Typography';
 import ActionsPanel from 'src/components/ActionsPanel';
 import Button from 'src/components/Button';
 import ConfirmationDialog from 'src/components/ConfirmationDialog';
-import FormControlLabel from 'src/components/core/FormControlLabel';
-import Typography from 'src/components/core/Typography';
 import Grid from 'src/components/Grid';
+import Link from 'src/components/Link';
 import Notice from 'src/components/Notice';
 import TextField from 'src/components/TextField';
 import Toggle from 'src/components/Toggle';
-import { makeStyles, Theme } from 'src/components/core/styles';
 
 interface Props {
   poolID: number;
@@ -173,9 +174,12 @@ const AutoscalePoolDialog: React.FC<Props> = (props) => {
         </Notice>
       ) : null}
       <Typography>
-        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Duis augue
-        arcu, semper id diam vitae, ultrices aliquet est. Morbi mauris risus,
-        mattis non sodales at, ultrices et quam.
+        Set minimum and maximum node pool constraints for LKE to resize your
+        cluster based on demand and control cluster resource costs. Autoscaler
+        maximum limit is 100.{' '}
+        <Link to={'https://www.linode.com/docs/products/compute/kubernetes/'}>
+          Learn More.
+        </Link>
       </Typography>
       <FormControlLabel
         label="Autoscaler"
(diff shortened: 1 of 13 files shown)


matthewbartholomew:
A few small UI adjustments to Pool table and the modal.

matthewbartholomew:
A few small adjustments to the UI. Let me know if you would like to discuss options or challenges. 
Autoscale button and values
Autoscale modal

matthewbartholomew:
**LKE detail page**
(2) If a pool is not autoscaled no min/max value should show
By default—no values
If autoscaling is turned off values are removed from table
(1) Close space between Autoscale Pool button and values to 8px space (-3px)

**Modal**
(3) Open space above toggle (add 12px)
(4) Decrease space above Min/Max fields (-3px)
(5) Make min/max entry fields narrower (70px) (issues with the grid?)
Toggle does not have a label indicating state—that’s a good change from the original designs

packages/validation/src/kubernetes.schema.ts
```diff
@@ -1,10 +1,39 @@
...
 export const nodePoolSchema = object().shape({
   type: string(),
   count: number(),
 });
 
+export const AutoscaleNodePoolSchema = object({
+  enabled: boolean(),
+  min: number().when('enabled', {
+    is: true,
+    then: number().test(
```
DevDW, line 12: Not sure if it's necessary, but should `number()` have `.required()` chained onto it here and also for the `max`?

DevDW (commented):
Still need to review a couple of more files, but looks good so far. Left a comment/question regarding the schema

bnussman (commented):
One thing I noticed is that when I change the autoscaler min / max value on a pool, it takes a few seconds for the `(Min 4 / Max 10)` values beside the button up update. Can we perform an immediate refetch or something to make the experience a bit better?

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/NodePoolsDisplay/AutoscalePoolDialog.tsx
```diff
@@ -0,0 +1,258 @@
...
+    onSubmit: submitForm,
+  });
+
+  return (
+    <ConfirmationDialog
+      open={open}
+      title="Autoscale Pool"
+      onClose={handleClose}
+      actions={() => (
+        <form onSubmit={handleSubmit}>
```
bnussman, line 131: Is there any benefit to wrapping this in a `<form />` over putting the submit function in the `onClick` of the `<Button />`?

HanaXu: Good point, not that I know of

HanaXu:
> One thing I noticed is that when I change the autoscaler min / max value on a pool, it takes a few seconds for the `(Min 4 / Max 10)` values beside the button up update. Can we perform an immediate refetch or something to make the experience a bit better?

Not sure how to do that 🤔

tiffwong (commented):
Everything looks great!

The only question I have is regarding using the touched hook on the inputs to control the disabled state of the Save Changes button. If I make changes to the dialog and reopen it, the button actually stays enabled. 

Also based on the other places where we disable the button if no changes have been detected, I think the button should be disabled based on if the current value matches the initial values instead of depending on if the user has interacted with the input field. 🤔

DevDW (approved):
LGTM, nice job! 🚀

tiffwong (approved):
Great job! 🚢

bnussman (approved):

PR #7955, opened 2021-09-16
E2e run workflow on staging and PRs with e2e label

ndukatz:
This PR changes the UI automation workflow to run on pushes to staging, as well as PR's when the `e2e` label is added. For now, all tests will run. Soon I will create more workflows for running only select tests based on what is changed in the PR.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -8,12 +8,16 @@ env:
 on:
   schedule:
     - cron: "0 13 * * 1-5"
+  pull_request_target:
+    types: [labeled]
   push:
     branches:
       - develop
+      - staging
 
 jobs:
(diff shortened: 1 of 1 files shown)


bnussman (approved):
This is awesome! 🚀 


Can you explain this bit to me?
```yaml
 with:
  ref: ${{ github.event.pull_request.head.sha }}
```

ndukatz:
@bnussman since this will be running on forked PR's and pointing towards a public repo (which is not the way they intended Actions to be used), this line just makes sure to grab the latest commit from the PR and run against that. Without this, it would end up just running against the latest develop commit.

johnwcallahan (approved):

patthiel (approved):
awesome! :shipit:

PR #7960, opened 2021-09-20
M3-5481: Remove "File System Path" column from Volumes landing table

DevDW:
## Description
Removes the "File System Path" column from the Volumes landing table

## How to test
Check the Volumes landing table and ensure the "File System Path" column has been removed and that the table looks good across viewport sizes.

Also check the Volumes table on the Linode Detail --> Storage tab to make sure it has not been impacted.

### DIFF
--- a/packages/manager/src/features/Volumes/VolumesLanding.tsx
+++ b/packages/manager/src/features/Volumes/VolumesLanding.tsx
@@ -107,13 +107,13 @@ const volumeHeaders = [
     label: 'Label',
     dataColumn: 'label',
     sortable: true,
-    widthPercent: 25,
+    widthPercent: 40,
   },
   {
     label: 'Region',
     dataColumn: 'region',
     sortable: true,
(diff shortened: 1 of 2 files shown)


packages/manager/src/features/Volumes/VolumesLanding.tsx
```diff
@@ -107,32 +107,25 @@ const volumeHeaders = [
...
     sortable: true,
-    widthPercent: 25,
+    widthPercent: 40,
   },
   {
     label: 'Region',
     dataColumn: 'region',
     sortable: true,
-    widthPercent: 15,
+    widthPercent: 10,
```
tiffwong, line 116: I think the region column is a little tight so might be nice to keep it at 15%

tiffwong (approved):
Left one optional feedback but everything else looks good!

HanaXu (approved):

PR #7964, opened 2021-09-22
M3-5482: Add NVMe chip for NVMe volumes

DevDW:
## Description
- Add `hardware_type` to `Volume` interface
- Update mocks, fix TS errors
- Display an NVMe chip on the Volumes landing page to the right of the labels of volumes that have a `hardware_type` of `nvme`

Screen Shot 2021-09-22 at 4 40 50 PM

## How to test
I updated the mocks, so you can turn on the MSW and see two mocked volumes with the NVMe chip.

### DIFF
--- a/packages/api-v4/src/volumes/types.ts
+++ b/packages/api-v4/src/volumes/types.ts
@@ -9,6 +9,7 @@ export interface Volume {
   updated: string;
   filesystem_path: string;
   tags: string[];
+  hardware_type: string;
 }
 
 export type VolumeStatus =
--- a/packages/api-v4/src/volumes/types.ts
+++ b/packages/api-v4/src/volumes/types.ts
@@ -9,7 +9,7 @@ export interface Volume {
   updated: string;
   filesystem_path: string;
   tags: string[];
-  hardware_type: string;
+  hardware_type: VolumeHardwareType;
 }
 
 export type VolumeStatus =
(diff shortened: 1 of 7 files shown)


bnussman:
Looks good! This is a very extreme screen size, but can we center the Chip vertically?

Screen Shot 2021-09-22 at 7 49 34 PM

packages/manager/src/features/Volumes/VolumeTableRow.tsx
```diff
@@ -93,10 +109,32 @@ export const VolumeTableRow: React.FC<CombinedProps> = (props) => {
...
+          wrap="nowrap"
+          justify="space-between"
+          alignItems="flex-end"
+        >
+          {isVolumesLanding ? (
+            <>
+              <Grid item>
+                <div>{label}</div>
+              </Grid>
+              <Grid item>
```
HanaXu, line 123: Maybe something like `style={{ alignSelf: 'center' }}` here

HanaXu (commented):
Seconded on aligning the chip vertically, it looks more exaggerated the longer the volume name is:

image

packages/api-v4/src/volumes/types.ts
```diff
@@ -9,6 +9,7 @@ export interface Volume {
   updated: string;
   filesystem_path: string;
   tags: string[];
+  hardware_type: string;
```
bnussman, line 12: Thoughts on adding a type for `hardware_type`? Something like:

```ts
type HardwareType = 'hdd' | 'nvme';
```

bnussman (approved):

HanaXu (approved):

PR #7965, opened 2021-09-22
M3-5322: Fix formatting of long strings in Domains TXT Record

HanaXu:
## Description
Truncate TXT Record value and fix overflowing bug

Before:

image

After:

image

## How to test
- Go to `domains/<domain_id>`
- Add/Edit a TXT Record value to be at least 439 characters
- The value should be truncated with an ellipsis and should not overflow into the next cell


DevDW (approved):
🚢

tiffwong (approved):

PR #7967, opened 2021-09-24
M3-5486: Add Upgrade Volumes Dialog

bnussman:
## Description

- Adds a modal for upgrading block storage volumes to NVMe
  - Modal is accessible by clicking the button in the upgrade volumes banner #7969
  - Modal can also be accessed by going to `/linodes/{linode_id}/upgrade`
  - If you click the `Upgrade to NVMe` chip on the Volumes landing page, it will redirect you to `/linodes/{linode_id}/upgrade`, automatically opening the upgrade dialog #7968 

## Todo
- [x] Find a way to deep link from the Volumes Landing page directly to a Linode Detail page **with this new dialog open.**
  - Maybe url matching like we do for `https://cloud.linode.com/account/billing/add-payment-method`
- [x] Pull in changes from the feature branch and hook up linking from buttons from banners and volumes page

## How to test

- Currently investigating a way to fully mock and test

Screen Shot 2021-09-24 at 9 14 17 AM

PR #7968, opened 2021-09-24
M3-5483: Add "Upgrade to NVMe" chip for volumes pending migration

tiffwong:
## Description

- Add `volume_migration_scheduled` and `volume_migration_imminent` to the `NotificationType` interface
- Add "Upgrade to NVMe" chip to volumes that have a "volume_migration_pending" notification
- Clicking the chip will redirect the page to the attached linode's detail page
- Update mocks

## How to test

Turn on MSW and go to `/volumes` and you should see a volume with the "Upgrade to NVMe" chip


packages/manager/src/features/Volumes/VolumeTableRow.tsx
```diff
@@ -85,13 +88,29 @@ export const VolumeTableRow: React.FC<CombinedProps> = (props) => {
...
 
   const isNVMe = hardwareType === 'nvme';
 
+  const scheduledVolumeMigrationNotifications = notifications.filter(
+    (notification) =>
+      notification.type === ('volume_migration_scheduled' as NotificationType)
+  );
+  const matchedNotification = scheduledVolumeMigrationNotifications.find(
+    (notification) =>
+      notification.entity?.id === id && notification.body?.includes(region)
```
DevDW, line 106: I think we'll need to make this an OR condition. Reason being the case of unattached volumes in the data center/region undergoing upgrades -- those should still have a chip

packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -401,14 +402,15 @@ export const handlers = [
     return res(ctx.json(record));
   }),
   rest.get('*/volumes', (req, res, ctx) => {
-    const hddVolumes = volumeFactory.buildList(2, {
+    const hddVolume = volumeFactory.build({
```
DevDW, line 405: Let's do something like

```
  rest.get('*/volumes', (req, res, ctx) => {
    const hddVolumeAttached = volumeFactory.build({
      region: 'us-southeast',
      linode_id: 10,
    });
    const hddVolumeUnattached = volumeFactory.build({
      region: 'us-southeast',
    });

    const nvmeVolumes = volumeFactory.buildList(2, {
      hardware_type: 'nvme',
    });

    const volumes = [...nvmeVolumes, hddVolumeAttached, hddVolumeUnattached];
    return res(ctx.json(makeResourcePage(volumes)));
  }),
```
for this so we have the case of an unattached volume in a data center undergoing upgrades accounted for

DevDW (commented):
Looks good so far; left a few comments. We'll need to add some more logic to handle to handle for the case of unattached volumes in a data center undergoing NVMe upgrades, plus the bit I commented on the ticket with re: using the events.

packages/manager/src/features/Volumes/VolumeTableRow.tsx
```diff
@@ -85,13 +89,35 @@ export const VolumeTableRow: React.FC<CombinedProps> = (props) => {
...
+      notification.type === ('volume_migration_scheduled' as NotificationType)
+  );
+  const matchedNotification = scheduledVolumeMigrationNotifications.find(
+    (notification) =>
+      notification.entity?.id === id || notification.body?.includes(region)
+  );
+  const eligibleForUpgradeToNVMe = Boolean(matchedNotification);
+
+  const goToAttachedLinode = () => {
+    history.push(`/linodes/${linodeId}`);
```
bnussman, line 113: Can you update this to
```ts
history.push(`/linodes/${linodeId}/upgrade`);
```

I am going to make a change in #7967 that will open the dialog when user visits `.../${linodeId}/upgrade`

packages/manager/src/features/Volumes/VolumeTableRow.tsx
```diff
@@ -85,13 +93,36 @@ export const VolumeTableRow: React.FC<CombinedProps> = (props) => {
...
 
+  const scheduledVolumeMigrationNotifications = notifications.filter(
+    (notification) =>
+      notification.type === ('volume_migration_scheduled' as NotificationType)
+  );
+  const matchedNotification = scheduledVolumeMigrationNotifications.find(
+    (notification) =>
+      notification.entity?.id === id || notification.body?.includes(region)
+  );
+  const eligibleForUpgradeToNVMe = Boolean(matchedNotification);
```
HanaXu, line 114: This can be consolidated to:
```typescript
const eligibleForUpgradeToNVMe = notifications.some(
  (notification) =>
    notification.type ===
      ('volume_migration_scheduled' as NotificationType) &&
    (notification.entity?.id === id || notification.body?.includes(region))
);
```

packages/manager/src/features/Volumes/VolumeTableRow.tsx
```diff
@@ -85,13 +93,36 @@ export const VolumeTableRow: React.FC<CombinedProps> = (props) => {
...
+
+  const goToAttachedLinode = () => {
+    history.push(`/linodes/${linodeId}/upgrade`);
+  };
```
HanaXu, line 118: We should check if the volume is attached before pushing, otherwise it pushes you to `/linodes/null/upgrade`

packages/manager/src/features/Volumes/VolumeTableRow.tsx
```diff
@@ -85,13 +93,36 @@ export const VolumeTableRow: React.FC<CombinedProps> = (props) => {
...
+        ('volume_migration_scheduled' as NotificationType) &&
+      (notification.entity?.id === id || notification.body?.includes(region))
+  );
+
+  const goToAttachedLinode = () => {
+    // Make sure Volume is attached before redirect
+    if (linodeId) {
+      history.push(`/linodes/${linodeId}/upgrade`);
+    }
+  };
```
HanaXu, line 118: On second thought, should we be displaying the chip at all if the volume is unattached? Since clicking the Chip causes nothing to happen. The story description also says to only show the chip for attached volumes. 🤔 

If we're not showing the Chip for unattached volumes, we could just inline this function in the Chip onClick

packages/manager/src/features/Volumes/VolumeTableRow.tsx
```diff
@@ -126,11 +157,20 @@ export const VolumeTableRow: React.FC<CombinedProps> = (props) => {
               {isNVMe ? (
                 <Grid item className={classes.chipWrapper}>
                   <Chip
-                    className={classes.chip}
+                    className={`${classes.chip} ${classes.nvmeChip}`}
                     label="NVMe"
                     data-testid="nvme-chip"
                   />
                 </Grid>
+              ) : eligibleForUpgradeToNVMe && !nvmeUpgradeScheduled ? (
```
bnussman, line 165: So users can't force an upgrade immediately even if is is scheduled? 😢 I thought all upgrades to NVMe were technically scheduled? The meaning of these notification types confuse me a little.

DevDW: The notification type names and event type names are rather similar so it's easy to get them mixed up, and maybe we can think about renaming this specific `nvmeUpgradeScheduled` variable, but the logic behind that variable is that the user went through the upgrade flow successfully, at which point an event is generated (`volume_migrate_scheduled`) indicating the volume has been queued up

bnussman: Thank you, I understand the flow much better now

packages/manager/src/features/Volumes/VolumeTableRow.tsx
```diff
@@ -85,13 +93,29 @@ export const VolumeTableRow: React.FC<CombinedProps> = (props) => {
...
   const isNVMe = hardwareType === 'nvme';
 
+  const eligibleForUpgradeToNVMe = notifications.some(
+    (notification) =>
+      notification.type ===
+        ('volume_migration_scheduled' as NotificationType) &&
+      (notification.entity?.id === id || notification.body?.includes(region))
+  );
+
+  const nvmeUpgradeScheduled = events.some(
```
DevDW, line 113: Perhaps we can rename this to `nvmeUpgradeScheduledByUser` to shed some more light on the flow and clarify things a bit

DevDW (approved):
I think we've covered the scenarios under which we want to show the chip and the scenarios under which we do not, so this should be good to go now 🚢

HanaXu (approved):

PR #7969, opened 2021-09-27
M3-5485: Upgrade volumes banner on Linode details

HanaXu:
## Description
Display upgrade volumes banner on a Linode's details page if upgradeable volumes are attached

## How to test
Set `numUpgradeableVolumes` to 1 or 1+, turn on the MSW, and go to a Linode's details page:

image

image

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
@@ -266,9 +266,6 @@ const LinodeDetailHeader: React.FC<CombinedProps> = (props) => {
     return <CircleProgress />;
   }
 
-  const getVolumesByLinode = (linodeId: number) =>
-    getVolumesForLinode(volumes.itemsById, linodeId).length;
-
   const volumesForLinode = getVolumesForLinode(volumes.itemsById, linode.id);
   const numAttachedVolumes = volumesForLinode.length;
 
@@ -403,7 +400,7 @@ const LinodeDetailHeader: React.FC<CombinedProps> = (props) => {
         variant="details"
         id={linode.id}
         linode={linode}
-        numVolumes={getVolumesByLinode(linode.id)}
+        numVolumes={numAttachedVolumes}
         username={profile?.username}
         linodeConfigs={linodeConfigs}
         backups={linode.backups}
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
@@ -18,6 +18,7 @@ import { DialogType } from 'src/features/linodes/types';
 import { notificationContext as _notificationContext } from 'src/features/NotificationCenter/NotificationContext';
 import useFlags from 'src/hooks/useFlags';
 import useLinodeActions from 'src/hooks/useLinodeActions';
+import useNotifications from 'src/hooks/useNotifications';
 import useReduxLoad from 'src/hooks/useReduxLoad';
 import useVolumes from 'src/hooks/useVolumes';
 import { getVolumesForLinode } from 'src/store/volume/volume.selector';
@@ -83,6 +84,16 @@ const LinodeDetailHeader: React.FC<CombinedProps> = (props) => {
   });
 
   const matchedLinodeId = Number(match?.params?.linodeId ?? 0);
+
+  const notifications = useNotifications();
+
+  const showUpgradeVolumesBanner = notifications.some(
+    (notification) =>
+      notification?.entity?.id === matchedLinodeId &&
+      ['volume_migration_scheduled', 'volume_migration_imminent'].includes(
+        notification.type
+      )
+  );
 
   const notificationContext = React.useContext(_notificationContext);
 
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
```diff
@@ -351,6 +362,39 @@ const LinodeDetailHeader: React.FC<CombinedProps> = (props) => {
...
+              </Typography>
+            </Grid>
+            <Grid item>
+              <Button buttonType="primary" onClick={() => null}>
+                Upgrade Volume
+              </Button>
+            </Grid>
+          </Grid>
+        </DismissibleBanner>
+      ) : null}
```
bnussman, line 397: Could turn into a compoonent in another file.

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
```diff
@@ -84,6 +85,16 @@ const LinodeDetailHeader: React.FC<CombinedProps> = (props) => {
 
   const matchedLinodeId = Number(match?.params?.linodeId ?? 0);
 
+  const notifications = useNotifications();
+
+  const showUpgradeVolumesBanner = notifications.some(
+    (notification) =>
+      notification?.entity?.id === matchedLinodeId &&
```
bnussman, line 92: After looking at the API spec, I believe `notification?.entity?.id` will be the ID of a Volume, not a Linode. Is that what you see, and are we going to need to check against all attached volumes if that's the case?

HanaXu: Good catch!

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
HanaXu: I think it's fine leaving it here as it's temp code afaik

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
```diff
@@ -277,6 +281,16 @@ const LinodeDetailHeader: React.FC<CombinedProps> = (props) => {
     linode.region === region &&
     numAttachedVolumes === 0;
 
+  const showUpgradeVolumesBanner = notifications.some(
+    (notification) =>
+      ['volume_migration_scheduled', 'volume_migration_imminent'].includes(
+        notification.type
+      ) &&
+      getVolumesForLinode(volumes.itemsById, linode.id).find(
```
bnussman, line 289: I believe `.some()` could also be used here instead of `.find()` but I believe functionality will be the same

bnussman (approved):

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
```diff
@@ -351,6 +365,39 @@ const LinodeDetailHeader: React.FC<CombinedProps> = (props) => {
...
+        >
+          <Grid
+            container
+            direction="row"
+            alignItems="center"
+            justify="space-between"
+          >
+            <Grid item>
+              <Typography>
+                Volume(s) attached to this Linode are eligible for a{' '}
```
DevDW, line 381: I know the copy isn't finalized, but we should use something (if it exists in the codebase already) or write something (if nothing exists in the codebase already) similar to the `pluralize` utility so we can avoid having that "(s)" to handle plural cases

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
```diff
@@ -277,6 +281,16 @@ const LinodeDetailHeader: React.FC<CombinedProps> = (props) => {
     linode.region === region &&
     numAttachedVolumes === 0;
 
+  const showUpgradeVolumesBanner = notifications.some(
+    (notification) =>
+      ['volume_migration_scheduled', 'volume_migration_imminent'].includes(
```
DevDW, line 286: I am unsure if `volume_migration_imminent` should be included here. Taking a look at STOR-1204, that notification type is described as being "returned when a volume has a [migration] row scheduled for a past date" -- I think that implies there is no further action the user should or can take

DevDW (commented):
Looks good, left a couple of comments

DevDW (approved):
Looks good to me aside from a refactor that can consolidate some logic 👍

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
```diff
@@ -264,7 +269,8 @@ const LinodeDetailHeader: React.FC<CombinedProps> = (props) => {
   const getVolumesByLinode = (linodeId: number) =>
     getVolumesForLinode(volumes.itemsById, linodeId).length;
 
-  const numAttachedVolumes = getVolumesByLinode(linode.id);
+  const volumesForLinode = getVolumesForLinode(volumes.itemsById, linode.id);
```
DevDW, line 272: I think it would be better to refactor `getVolumesByLinode()` so that it returns the array of volumes instead of the length (and make adjustments accordingly in the file where it is currently used); that way, we won't need this extra variable and can consolidate this related logic a bit.

PR #7971, opened 2021-09-30
M3-5499: Cleanup TextField story

tiffwong:
## Description

- Remove props that are no longer in use or have no effect (ie. `affirmative`, `small`, and `tiny`).
- Convert to a function component (required in order to import the props to the `ArgsTable` in the docs)
- Synchronize the naming convention for the `Button` docs and add default values to the `ArgsTable`

## How to test

- Check the text fields throughout the app
- Run `yarn storybook` to see the documentation for TextFields (it's nested inside the `Components`/`Forms` folder)


DevDW (changes requested):
I'm not sure exactly what broke in `TextField.tsx` to cause the issue, but the number inputs in the Add a Node Pool drawer aren't updating properly

packages/manager/src/components/TextField.tsx
```diff
@@ -150,34 +98,59 @@ interface TextFieldPropsOverrides extends TextFieldProps {
...
+    theme,
+    tooltipText,
+    type,
+    value,
+    ...textFieldProps
+  } = props;
 
-  handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-    const { type, min, max, onChange } = this.props;
+  const [_value, setValue] = React.useState<string | number>(
```
DevDW, line 140: I looked into this some more; if you initialize this to `''` and add a `useEffect` like

```
  React.useEffect(() => {
    if (typeof value === 'string' || typeof value === 'number') {
      setValue(value);
    }
  }, [value]);
```

elsewhere, instances of `EnhancedNumberInput` start behaving as expected again

tiffwong: Hmm the `EnhancedNumberInput` test still fails and when I press the up arrow, the value doesn't increase like it does in prod 🤔

packages/manager/src/components/TextField.tsx
```diff
@@ -7,140 +7,88 @@ import CircleProgress from 'src/components/CircleProgress';
...
   /**
-   * number amounts allowed in textfield
-   * "type" prop must also be set to "number"
+   * The number amounts allowed in TextField and
+   * the "type" prop must also be set to "number"
    */
-  min?: number;
   max?: number;
-  dataAttrs?: Record<string, any>;
+  min?: number;
```
HanaXu, line 87: Are the min and max props used anywhere other than in `EnhancedNumberInput`? It might make more sense to refactor this out of `TextField` and into `EnhancedNumberInput`

packages/manager/src/components/TextField.tsx
tiffwong: @bnussman are you referring to the most recent changes?

PR #7974, opened 2021-10-05
M3-5489: Handle language for Volume migration events

HanaXu:
## Description
Display friendly messages and progress indicators for volume migration event types.

## How to test
Turn on the MSW and open the notifications drawer

image

### DIFF
--- a/packages/manager/src/eventMessageGenerator.ts
+++ b/packages/manager/src/eventMessageGenerator.ts
@@ -663,6 +663,15 @@ export const eventMessageCreators: { [index: string]: CreatorsForStatus } = {
     finished: (e) => `Volume ${e.entity!.label} has been detached.`,
     notification: (e) => `Volume ${e.entity!.label} has been detached.`,
   },
+  volume_migrate: {
+    started: (e) => `Volume ${e.entity!.label} is being upgraded to NVMe.`,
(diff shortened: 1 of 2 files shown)


packages/manager/src/eventMessageGenerator.ts
```diff
@@ -663,6 +663,15 @@ export const eventMessageCreators: { [index: string]: CreatorsForStatus } = {
...
   },
+  volume_migrate: {
+    started: (e) => `Volume ${e.entity!.label} is being upgraded to NVMe.`,
+    failed: (e) => `Volume ${e.entity!.label} failed to upgrade to NVMe.`,
+    finished: (e) => `Volume ${e.entity!.label} has been upgraded to NVMe.`,
+  },
+  volume_migrate_scheduled: {
+    scheduled: (e) =>
+      `Volume ${e.entity!.label} has been scheduled for an upgrade to NVMe.`,
+  },
```
HanaXu, line 674: Not sure if this is correct

DevDW:
> @DevDW I just undid the changes, I couldn't figure out a way to get rid of the event actions in the mocks without going too out of scope

One option, maybe less than ideal, would be to make an exception in `formatEventWithAPIMessage()` for it by adding

```
  if (e.action === 'volume_migrate') {
    return `${e.message}`;
  }
```

at the start of that function

PR #7975, opened 2021-10-06
M3-5369: Consistent spelling of "canceled"

tiffwong:
## Description

Change all instances of "cancelled" to "canceled".

## How to test

Make sure nothing broke.

### DIFF
--- a/packages/manager/src/features/ToastNotifications/ToastNotifications.tsx
+++ b/packages/manager/src/features/ToastNotifications/ToastNotifications.tsx
@@ -106,11 +106,10 @@ class ToastNotifications extends React.PureComponent<WithSnackbarProps, {}> {
               `Image ${label} is now available.`,
               isDeletion
                 ? undefined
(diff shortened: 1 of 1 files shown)


packages/api-v4/src/entity-transfers/types.ts
```diff
@@ -1,7 +1,7 @@
 export type EntityTransferStatus =
   | 'pending'
   | 'accepted'
-  | 'cancelled'
+  | 'canceled'
```
bnussman, line 4: Revert this to `cancelled` since the API will return still `canceled`.

bnussman (changes requested):

packages/api-v4/src/entity-transfers/types.ts
DevDW, line 4: I think we should keep "cancelled" here to match what the API returns.

DevDW (commented):
Functionality looks unchanged, have a couple of comments/questions

PR #7977, opened 2021-10-07
[DNM] M3-5460: Remove copy warning about deleting non empty buckets

bcmertz:
## Description

Removes warning for deleting non empty Object Storage buckets

## How to test

Create Object Storage bucket, fill with some files, then delete. Verify there is no warning about deleting a non empty bucket. Note: until ARB-2708 is merged this should still yield an API error mentioning a non empty bucket.


bcmertz:
ready to merge once ARB-270 is merged

PR #7991, opened 2021-10-14
Updates to Block Storage Migrations

bnussman:
## Description

- Upgrade Volumes Dialog
  - Changed “All Volumes attached to Linode” to “All Volumes attached to this Linode ([insert name of Linode here])”
  - Un-bolded the number of Linodes in the migration queue
- Upgrade Volumes Banner
  - Changed "Learn More" to "Learn more"

## How to test

- Check for changes in the components above


matthewbartholomew:
Change text to “All Volumes attached to Linode [insert name of Linode here]”
Change Button text to Enter Upgrade Queue

bnussman:
@HanaXu  wrapped in Typography... Font looks kind of small. Should I increase the font size or do you think it is ok. I looked at other Dialogs and it seems like smaller font is normal... 

With text wrapped in topography:
Screen Shot 2021-10-14 at 8 51 40 PM

PR #8002, opened 2021-10-21
M3-5543: Add "Upgrade Pending" chip to Volumes

tiffwong:
## Description

Add a "Upgrade Pending" chip to Volumes that have been queued for upgrade to NVMe by the user.

## How to test

1. Switch to a dev environment and go to `/volumes`
2. If you have Volumes that were created before NVMe was available, skip the next step
3. Otherwise, use `curld` to create an attached Volume that uses HDD
4. Schedule a migration in alpha
5. The Volume should now have the "Upgrade to NVMe" chip next to it
6. Click the chip and go through the flow on the attached Linodes Detail page
7. If you go back to `/volumes`, the chip should now say "Upgrade Pending"
8. The page should refresh once it finishes migrating which might take several minutes

Feel free to message me for help in particular for creating an eligible Volume and scheduling a migration


HanaXu (commented):
Thoughts on adding the chips to `/linodes/<linode_id>/storage` as well? 🤔 

image

tiffwong:
@HanaXu good catch, I'll make a ticket for this but it probably won't go out in the initial release since the Linode Volume's are rendered differently so it might take a bit more time to get that part working

HanaXu (approved):
LGTM 🚀

DevDW (approved):
Looks great!

I'm not sure why the mocked events still show that prepended "volume_migrate" piece since the change in `RenderProgressEvent.tsx` should've fixed that. The Notifications drawer looks good with real data though. 

Screen Shot 2021-11-08 at 3 56 33 PM

bnussman (approved):

PR #8010, opened 2021-10-27
M3-5529: Update "Add an IP Address" drawer and provide for ability to add IPv6 ranges in it

DevDW:
## Description
Updates the types and requests to support the new IPv6 functionality, updates the mocks accordingly, and makes a few changes in the "Add an IP Address" drawer:

- Changes IPv4 dropdown to radio buttons
- Adds IPv6 radio buttons

## To-Do
- [x] Clean up & refactor `<AddIPDrawer />`
- [x] Final description copy

## How to test
From the "Network" tab of a Linode Detail page, click "Add an IP Address" and confirm that no functionality has been lost for allocating IPv4 addresses and that you can now allocate IPv6 addresses successfully.

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/AddIPDrawer.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/AddIPDrawer.tsx
@@ -1,13 +1,18 @@
 import { allocateIPAddress } from '@linode/api-v4/lib/linodes';
+import { createIPv6Range, IPv6Prefix } from '@linode/api-v4/lib/networking';
+import { APIError } from '@linode/api-v4/lib/types';
 import * as React from 'react';
 import { Link } from 'react-router-dom';
 import ActionsPanel from 'src/components/ActionsPanel';
 import Button from 'src/components/Button';
+import FormControlLabel from 'src/components/core/FormControlLabel';
+import Radio from 'src/components/core/Radio';
+import RadioGroup from 'src/components/core/RadioGroup';
 import { makeStyles, Theme } from 'src/components/core/styles';
 import Tooltip from 'src/components/core/Tooltip';
 import Typography from 'src/components/core/Typography';
 import Drawer from 'src/components/Drawer';
-import EnhancedSelect, { Item } from 'src/components/EnhancedSelect/Select';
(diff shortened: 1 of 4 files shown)


packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/AddIPDrawer.tsx
```diff
@@ -122,29 +193,43 @@ const AddIPDrawer: React.FC<CombinedProps> = (props) => {
   );
 
   const _tooltipCopy =
-    disabled && selected
+    disabledIPv4 && selectedIPv4
       ? readOnly
         ? 'You do not have permission to modify this Linode.'
-        : tooltipCopy[selected]
+        : tooltipCopy[selectedIPv4]
       : null;
```
codyfinn, line 200: Is it possible to un-nest this logic?

DevDW: Do you think breaking it up into some `if` statements would be easier to follow?

johnwcallahan (commented):
I know this is still WIP, but I had a quick look and the general approach looks good to me.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/AddIPDrawer.tsx
codyfinn: That would probably help make it more readable. I'm more focused on the nested logic in general though.

HanaXu (commented):
Thoughts on adding a success toast after allocation? 🤔

DevDW:
> Thoughts on adding a success toast after allocation? 🤔

I think that would be fine but I'm not sure how necessary it is since the allocation happens quickly and the user remains on the same page where the new IP shows up

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/AddIPDrawer.tsx
DevDW: Going to opt to keep it as is for now unless anyone feels strongly about it; I don't think the nesting is too bad here

HanaXu (approved):
🚀

PR #8011, opened 2021-10-28
M3-5553: Revisions to Account Maintenance Page for Volumes

bnussman:
## Description

Screen Shot 2021-10-28 at 2 18 26 PM

- Ensure the Status indicator and text are properly aligned vertically
  - Same issue as #7983
- Hide Pagination controls if there are not >25 results
- Move the `Download CSV` link outside of the table to the bottom right below it
  - Still hides link when accordion is collapsed 

## How to test

- Go to `/account/maintenance` in Cloud Manager
- Test the three changes listed above
- Double check functionality of tables and accordians

### DIFF
--- a/packages/manager/src/components/PaginationFooter/PaginationFooter.tsx
+++ b/packages/manager/src/components/PaginationFooter/PaginationFooter.tsx
@@ -49,7 +49,6 @@ export interface PaginationProps {
   eventCategory: string;
   showAll?: boolean;
   fixedSize?: boolean;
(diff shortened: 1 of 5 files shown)


tiffwong (approved):

bnussman:
I was able to address @DevDW's feedback. The solution is not super elegant, but works as expected. Can I get one more review or a re-review from someone? cc @tiffwong

tiffwong (approved):

PR #8017, opened 2021-11-01
M3-5559: Button alignment in the Enable All Backups drawer

bnussman:
## Description

- Fixes button order in the **_Enable All Backups_** drawer

### Before

Screen Shot 2021-11-01 at 6 51 47 PM

### After

Screen Shot 2021-11-01 at 7 10 38 PM

## How to test

- Go to `/account/settings` on a non-managed account
- Click `enable now` to open the **_Enable All Backups_** drawer
- Make sure buttons are right aligned and the primary action button is on the right side

### DIFF
--- a/packages/manager/src/features/Backups/BackupDrawer.tsx
+++ b/packages/manager/src/features/Backups/BackupDrawer.tsx
@@ -179,15 +179,6 @@ export class BackupDrawer extends React.Component<CombinedProps, {}> {
           <Grid item>
             <ActionsPanel style={{ padding: 0, margin: 0 }}>
               <Button
-                onClick={this.handleSubmit}
-                loading={loading || enabling || enrolling}
-                buttonType="primary"
-                data-qa-submit
-                data-testid={'submit'}
-              >
-                Confirm
-              </Button>
-              <Button
                 onClick={close}
                 buttonType="secondary"
                 className="cancel"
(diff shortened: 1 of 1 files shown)


packages/manager/src/features/Backups/BackupDrawer.tsx
```diff
@@ -178,15 +178,6 @@ export class BackupDrawer extends React.Component<CombinedProps, {}> {
           </Grid>
           <Grid item>
             <ActionsPanel style={{ padding: 0, margin: 0 }}>
```
codyfinn, line 180: Would it make sense to have ActionPanel sort it's children based on the type of buttons they are? Where it will always put the primary buttons last.

codyfinn (approved):
Left a suggestion but the PR is good IMO

packages/manager/src/features/Backups/BackupDrawer.tsx
tiffwong: I think all (or at least the ones we accounted for) should now be in the right order so it might not be necessary anymore but definitely a good idea 😂

tiffwong (approved):

packages/manager/src/features/Backups/BackupDrawer.tsx
bnussman: What @tiffwong said, haha

PR #8020, opened 2021-11-02
M3-5542 & 5544: Upgrade Volume banner and modal copy

tiffwong:
## Description

Update the copy for the banner and modal.

**Note:** Ignore the branch name, I got my features mixed up 😅 

## How to test

1. Switch to a dev environment and go to `/volumes`
2. If you have Volumes that were created before NVMe was available, skip the next step
3. Otherwise, use `curld` to create an attached Volume that uses HDD
4. Schedule a migration in alpha
5. The Volume should now have the "Upgrade to NVMe" chip next to it
6. Click the chip which should redirect you to the attached Linode Detail page with a modal open
7. Review the copy on the modal and click "Cancel"
8. Review the banner (if you confirm the previous modal, the banner will automatically dismiss)

Feel free to message me for help in particular for creating an eligible Volume and scheduling a migration

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
@@ -463,9 +463,9 @@ const LinodeDetailHeader: React.FC<CombinedProps> = (props) => {
                 {numUpgradeableVolumes === 1
                   ? 'A Volume attached to this Linode is '
                   : 'Volumes attached to this Linode are '}
-                eligible for a <b>free upgrade</b> to high performance NVMe
-                Block Storage.{' '}
-                <Link to="https://www.linode.com/blog/cloud-storage/nvme-block-storage-now-available/">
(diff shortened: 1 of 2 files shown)


packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
```diff
@@ -463,9 +463,9 @@ const LinodeDetailHeader: React.FC<CombinedProps> = (props) => {
                 {numUpgradeableVolumes === 1
                   ? 'A Volume attached to this Linode is '
                   : 'Volumes attached to this Linode are '}
-                eligible for a <b>free upgrade</b> to high performance NVMe
-                Block Storage.{' '}
-                <Link to="https://www.linode.com/blog/cloud-storage/nvme-block-storage-now-available/">
+                eligible for a <b>free upgrade</b> to Linode&rsquo;s
```
codyfinn, line 466: I think there is a simpler way to deal with the apostrophes where we use the replace method on string children of typography. For the purposes of fixing the copy though the PR is good.

HanaXu (approved):

PR #8022, opened 2021-11-02
M3-5468: Fix Linode Networking Graph Overlapping with DNS Resolvers

codyfinn:
## Description

Prevents the Linode Networking graph from flowing outside of it's container and into the DNS Resolvers table. This was an issue where since the graph is rendered using a Canvas element it cannot be sized dynamically easily. The docs for chartjs specify that in order to dynamically resize the chart specifying a relative position and a relative width on the graph would be necessary. https://www.chartjs.org/docs/latest/configuration/responsive.html#important-note

## How to test

1. Open a Linode details page.
2. Ensure your browser window is relatively wide.
2. Switch to the Networking tab.
3. Slowly shrink the horizontal size of your browser window.
4. Observe if at any point the DNS Resolvers table overlays the Networking graph.

### DIFF
--- a/packages/manager/src/components/LineGraph/LineGraph.tsx
+++ b/packages/manager/src/components/LineGraph/LineGraph.tsx
@@ -58,6 +58,10 @@ type CombinedProps = Props;
 
 const useStyles = makeStyles((theme: Theme) => ({
   ...MetricDisplayStyles(theme),
+  canvasContainer: {
+    position: 'relative',
+    width: '80%',
+  },
 }));
 
 const lineOptions: ChartDataSets = {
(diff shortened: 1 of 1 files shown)


packages/manager/src/components/LineGraph/LineGraph.tsx
```diff
@@ -58,6 +58,10 @@ type CombinedProps = Props;
 
 const useStyles = makeStyles((theme: Theme) => ({
   ...MetricDisplayStyles(theme),
+  canvasContainer: {
+    position: 'relative',
+    width: '80%',
```
codyfinn, line 63: I picked this width arbitrarily there might be a relative width that makes more sense in the given scenario.

tiffwong (approved):
Great job! 🚀

HanaXu (approved):
🚢

PR #8032, opened 2021-11-10
Feature/paypal

HanaXu:


packages/api-v4/src/account/types.ts
```diff
@@ -396,19 +390,38 @@ export interface AccountMaintenance {
...
+      data: CreditCardData;
+    }
+  | {
+      id: number;
+      type: 'paypal';
+      is_default: boolean;
+      created: string;
+      data: PayPalData;
+    };
+
```
codyfinn, line 419: Could these be stored in their own types? That way we can use these types for typechecking more easily

DevDW (approved):
Things look to be working well 👌 

For the `@todo`'s about removing some lines of code when PayPal + Braintree is done, should those lines be removed at this time? Or will they be removed as part of a cleanup PR after this is released?

packages/manager/src/features/Billing/BillingPanels/PaymentInfoPanel/AddPaymentMethodDrawer/AddCreditCardForm.tsx
```diff
@@ -28,15 +27,13 @@ const useStyles = makeStyles((theme: Theme) => ({
...
 
 interface Values {
   card_number: string;
   expiry_month: string;
   expiry_year: string;
   cvv: string;
-  is_default: boolean;
   address: string;
   address2: string;
   city: string;
```
codyfinn, line 39: moving ahead interface's who's properties are of general string types could use string templates to have greater assurance that the values we have are what we expect. Typescript is just making sure that wherever we use this type has a string in it. It could be a string of any pattern which might not be what we expect.

codyfinn (approved):
I left a couple of non-blocking comments on this PR. Looks great nice work on this new feature!

HanaXu:
@DevDW They're going to be removed sometime after the feature is released and is deemed stable

PR #8037, opened 2021-11-11
M3-5530: Add IPv6 ranges to IP Transfer modal

HanaXu:
## Description
Add functionality to move/swap IPv6 Ranges in the IP Transfer Modal

## How to test
- On the dev env, go to `linodes/<linode_id>/networking`
    - Pick a Linode that already has an IPv6 Range, or add an IPv6 Range to a Linode (reach out for more details)
    - Pick a second Linode that has more than one public IPv4 address, or add an additional IPv4 address to a Linode (reach out for more details)
- Click on the `IP Transfer` button to open up the transfer modal
- You should see the IPv6 Range as a row in the modal and should be able to move or swap it to another Linode
    - To swap it with an IPv4 address, the Linode you're swapping with must have at least one public IP after assignment.
    - To swap it with another IPv6 range, your account needs to be able to have more than 1 max allowed IPv6 range, otherwise you'll get an error. This error is expected for now
- IPv4 moving/swapping should work as normal

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPTransfer.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPTransfer.tsx
@@ -24,7 +24,10 @@ import Select, { Item } from 'src/components/EnhancedSelect/Select';
 import Grid from 'src/components/Grid';
 import Notice from 'src/components/Notice';
 import usePrevious from 'src/hooks/usePrevious';
+import { queryKey } from 'src/queries/networking';
+import { queryClient } from 'src/queries/base';
 import { useLinodesQuery } from 'src/queries/linodes';
+import { useIpv6RangesQuery } from 'src/queries/networking';
 import { getAPIErrorOrDefault } from 'src/utilities/errorUtils';
 import { debounce } from 'throttle-debounce';
 
@@ -128,7 +131,6 @@ const LinodeNetworkingIPTransferPanel: React.FC<CombinedProps> = (props) => {
     readOnly,
   } = props;
   const classes = useStyles();
-  console.log(ipAddresses, props)
   const [ips, setIPs] = React.useState<IPRowState>(
     props.ipAddresses.reduce(
       (acc, ip) => ({
@@ -157,6 +159,8 @@ const LinodeNetworkingIPTransferPanel: React.FC<CombinedProps> = (props) => {
     },
     open // only run the query if the modal is open
   );
+
+  const { data: ipv6RangesData } = useIpv6RangesQuery();
 
   const linodes = Object.values(data?.linodes ?? []).filter(
     (l) => l.id !== linodeID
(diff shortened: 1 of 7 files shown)


packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPTransfer.tsx
```diff
@@ -403,6 +425,7 @@ const LinodeNetworkingIPTransferPanel: React.FC<CombinedProps> = (props) => {
             setSubmitting(false);
             setError(undefined);
             setSuccessMessage('IP transferred successfully.');
+            queryClient.invalidateQueries(queryKey);
```
johnwcallahan, line 428: Minor suggestion: 

I'd recommend importing `queryKey as ipv6RangeQueryKey` or something similar, and adding a comment before this line to explain why we're doing this.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPTransfer.tsx
```diff
@@ -193,13 +198,29 @@ const LinodeNetworkingIPTransferPanel: React.FC<CombinedProps> = (props) => {
...
+            const linodeIpv6Ranges = getLinodeIpv6Ranges(firstLinode.ipv6);
+            return [...firstLinode.ipv4, ...linodeIpv6Ranges];
+          })
         )
       )
     );
     setIPs((currentState) => newState(currentState));
   };
 
+  const getLinodeIpv6Ranges = (ipv6: string | null) => {
```
johnwcallahan, line 211: Great thinking putting this in a separate function! 

I'd recommend going one step further and moving this function outside of the component, give`ipv6RangesData` as a second argument , and then add a few unit tests. Let me know if you want me to provide more clarity here.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPTransfer.tsx
```diff
@@ -157,6 +160,8 @@ const LinodeNetworkingIPTransferPanel: React.FC<CombinedProps> = (props) => {
     open // only run the query if the modal is open
   );
 
+  const { data: ipv6RangesData } = useIpv6RangesQuery();
```
johnwcallahan, line 163: I think we could use an error state if IPv6 ranges fail to load (currently if you block requests to `*ranges*` there's a silent failure ... IPv6 ranges are missing from the appropriate dropdowns).

We probably want the error indicator to go where the `<CircleProgress />` component is rendered while loading. 

Speaking of loading... the loading state should also take into account whether or not ipv6 ranges are loading as well (from this query).

johnwcallahan (changes requested):
Awesome work with this, especially navigating these old components with ancient patterns!! 

Just two things I think should be addressed before moving forward:

1. Error/loading state
2. Extracting the range mapping function and adding a few unit tests. 

The other comment is a minor suggestion.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPTransfer.tsx
DevDW, line 211: would be good to add a unit test for this given the multiple layers of operation on strings

DevDW (commented):
Nice work, left a couple of comments in addition to what John mentioned re: the error state

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPTransfer.tsx
HanaXu: I didn't want the IPv6 error state to block IPv4 stuff so I decided not to replace the `<CircleProgress />` component with the error notice.

image

johnwcallahan: @HanaXu good idea!

johnwcallahan (approved):
Great work @HanaXu! This LGTM.

DevDW (approved):
🚢

PR #8041, opened 2021-11-12
M3-5576: Fix row height and extra spacing around ActionMenu in Safari

tiffwong:
## Description

The `InlineMenuAction` and `ActionMenu` components weren't fully extending to the entire row height in Safari for certain entities (ie. Volumes and Images). This seems to be related to having an extra `div` inside the `TableCell` and applying the styling on that `div`. 

In order to make things more consistent and prevent having to copy and paste the styles for each `ActionMenu`, I created a `TableCell` variant that cleans up the code.

## How to test

Check the `ActionMenu`'s throughout the app on Chrome **and** Safari.

### DIFF
--- a/packages/manager/src/components/PaymentMethodRow/PaymentMethodRow.tsx
+++ b/packages/manager/src/components/PaymentMethodRow/PaymentMethodRow.tsx
@@ -23,6 +23,9 @@ export const useStyles = makeStyles((theme: Theme) => ({
   },
   actions: {
     marginLeft: 'auto',
+    '& button': {
+      margin: 0,
+    },
   },
   item: {
     display: 'flex',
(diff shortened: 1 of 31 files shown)


codyfinn (approved):
Awesome work finding the solution to this!

bnussman (changes requested):
Looks great but I'm still seeing something like 2px of padding in the PaymentMethodRow component in Safari, can we get this fixed too? 

Screen Shot 2021-11-15 at 9 32 59 AM

bcmertz (commented):
Awesome job! Could SortableTableHead.tsx also be refactored?

bnussman (approved):
This is great! Makes using Cloud Manager in Safari so much better and cleans up a lot of styles. Awesome work

PR #8042, opened 2021-11-12
M3-5535: Handle IPv6 events language

DevDW:
## Description
Before:
Screen Shot 2021-11-12 at 2 03 17 PM

After:
Screen Shot 2021-11-12 at 1 54 31 PM

Open to changing the language, perhaps removing the "on your account" phrase.

## How to test
- Add an IPv6 range
- Delete it
- Check the Notifications drawer


johnwcallahan (approved):
I vote for removing the "on your account" part, but I think it works either way 🤷 . 

Nice job getting this out so quickly 👍

HanaXu (approved):
Agreed on removing "on your account"

PR #8045, opened 2021-11-16
M3-4983: Transferred column on NodeBalancer page not showing the correct amount

codyfinn:
## Description

Removes use of `Math.ceiling` from `convertMegabytesTo` function. This use of `Math.ceiling` causing only increases in the magnitude on the order of the nearest whole storage unit to be shown to the user. Adds tests for `convertMegabytesTo` function.

## How to test

Ensure that when a Node Balancer has transferred a quantity of bytes that should be represented with a string containing a number with two decimal places of precision that the representation is as accurate as possible. For example if a Node Balancer has transfered `1.432342 GB` of data then that should be represented as `1.43 GB` and not as `1.00 GB` or `2.00 GB`.

yarn test packages/manager/src/utilities/unitConversions.test.ts

### DIFF
--- a/packages/manager/src/utilities/unitConversions.ts
+++ b/packages/manager/src/utilities/unitConversions.ts
@@ -8,20 +8,16 @@ export const convertMegabytesTo = (data: number, removeDecimals?: boolean) => {
   if (totalToBytes >= gb) {
     // convert bytes to GB
     return removeDecimals
-      ? `${Math.max(Math.ceil(totalToBytes / gb))} GB`
-      : `${Math.max(Math.ceil(totalToBytes / gb)).toFixed(2)} GB`;
+      ? `${Math.max(totalToBytes / gb)} GB`
+      : `${Math.max(totalToBytes / gb).toFixed(2)} GB`;
   }
   if (totalToBytes >= mb) {
     // convert bytes to MB
-    return `${Math.max(Math.ceil((totalToBytes / mb) * 100) / 100).toFixed(
-      2
-    )} MB`;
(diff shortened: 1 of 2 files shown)


packages/manager/src/utilities/unitConversions.ts
```diff
@@ -8,20 +8,16 @@ export const convertMegabytesTo = (data: number, removeDecimals?: boolean) => {
   if (totalToBytes >= gb) {
     // convert bytes to GB
     return removeDecimals
-      ? `${Math.max(Math.ceil(totalToBytes / gb))} GB`
-      : `${Math.max(Math.ceil(totalToBytes / gb)).toFixed(2)} GB`;
+      ? `${Math.max(totalToBytes / gb)} GB`
```
DevDW, line 11: Do we need the `Math.max()` method as it is used on this and the following lines, being that only one argument is being passed in? It doesn't seem to be serving a purpose -- I removed them and the test still passed, the `Transferred` amount still looked OK, etc.

DevDW (approved):
Looks good 👍  left one comment re: `Math.max()`

packages/manager/src/utilities/unitConversions.ts
codyfinn: good point! I think the reason it is there is in case data is `null`. If `Math.max` is passed `null` as its first argument it will return `0`. I do not think that null can be passed to `convertMegabytesTo` in theory as long as the typechecker is doing its job right.

bnussman (approved):
Looks good! 🚀

tiffwong (approved):

PR #8047, opened 2021-11-17
M3-5567: Add warning when booting into Rescue Mode that Cloud Firewall will not be enabled

DevDW:
## Description
Linodes booted into Rescue Mode no longer have firewall rules enabled.

This PR adds a warning notice in the Rescue dialog that says, `Cloud Firewall rules are not enabled when booting into Rescue Mode.` if the Linode is assigned to a firewall.

To facilitate this, a new method `getLinodeFirewalls` has been added to the JS Client, along with a new RQ query `useLinodeFirewalls`.

## How to Test
There are 3 cases. Only case #1 should display the notice; the other two cases should look the same as they do in prod.

1. A Linode assigned to a firewall
Screen Shot 2021-11-22 at 2 21 41 PM

2. A Linode not assigned to a firewall
Screen Shot 2021-11-22 at 2 21 20 PM

3. A Bare Metal instance (this PR modifies the logic for `BareMetalRescue.tsx` and `RescueDescription.tsx`)
Screen Shot 2021-11-22 at 2 21 00 PM

Please also check React Query to confirm that the data is being stored and handled as expected.

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDescription.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDescription.tsx
@@ -7,6 +7,8 @@ import Link from 'src/components/Link';
 const rescueDescription = {
   text: `If you suspect that your primary filesystem is corrupt, use the Linode Manager to boot your Linode into Rescue Mode. This is a safe environment for performing many system recovery and disk management tasks.`,
   link: 'https://www.linode.com/docs/guides/rescue-and-rebuild/',
+  firewallWarning:
+    'Cloud Firewall rules are not enabled when booting into Rescue Mode.',
 };
 
 interface Props {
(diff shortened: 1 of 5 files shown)


packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDescription.tsx
```diff
@@ -35,6 +37,9 @@ const RescueDescription: React.FC<Props> = (props) => {
         {rescueDescription.text}{' '}
         <Link to={rescueDescription.link}>Learn more.</Link>
       </Typography>
+      <Typography className={classes.copy}>
+        {rescueDescription.firewallWarning}
+      </Typography>
```
codyfinn, line 42: Should this be a `Notice` instead? It kind of blends in with the text above it.

HanaXu (commented):
Agreed with Cody on making the warning a Notice. Conditionally showing the warning would be nice

tiffwong (approved):
Approving but I also agree with Cody and Hana that a conditional `Notice` would be ideal

HanaXu (approved):
🚀

bcmertz (approved):

PR #8048, opened 2021-11-18
M3-5059: Persisting when editing stackscripts on separate computers

codyfinn:
## Description

When a StackScript is edited without being saved the data stored in local storage now includes the `updated` field. Before the data from local storage is loaded its `updated` is compared to the `updated` field from the response. If response has been updated since the edits in the local storage then the local storage data is not loaded.

## How to test

1. Open a StackScript in two different browsers.
2. In browser A edit the StackScript without saving the changes.
3. In browser A navigate away from the StackScript.
4. In browser B edit the StackScript in a different way and save the changes.
5. In browser A navigate to the same StackScript.
6. In browser A click the `edit` button for the StackScript.
7. The prefilled text in the StackScript should reflect the changes from browser B.

### DIFF
--- a/packages/manager/src/utilities/storage.ts
+++ b/packages/manager/src/utilities/storage.ts
@@ -70,9 +70,6 @@ interface TicketReply {
 
 interface StackScriptData extends StackScriptPayload {
   id: number | string;
-}
-
-interface StackScriptDataWithUpdated extends StackScriptData {
   updated: string;
 }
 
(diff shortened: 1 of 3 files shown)


codyfinn:
I reverted back to using local storage since using session storage alone would not fix the issues in the ticket.

DevDW (commented):
I tested this a few times with browser A being Chrome and browser B being Firefox. Each time, what I observed is that step 5 (StackScript detail page) would have the changes from browser B, but step 7 (StackScript edit page) would have the changes from browser A.

packages/manager/src/utilities/storage.ts
```diff
@@ -72,6 +72,10 @@ interface StackScriptData extends StackScriptPayload {
   id: number | string;
 }
 
+interface StackScriptDataWithUpdated extends StackScriptData {
```
DevDW, line 75: is `StackScriptData` being used anywhere now? If not, can these interfaces be combined?

DevDW (approved):
LGTM now 🚢

HanaXu (approved):
LGTM! I agree with Dajahi on combining the interfaces

PR #8049, opened 2021-11-19
M3-5516: Add url page number handling for linodes

HanaXu:
## Description
- You can now load a specific page in `/linodes` through the url with a page query parameter (e.g. `/linodes?page=2`).
- Clicking on a page number in `/linodes` adds a page query param to the url.
- Added back `PaginationControl` to storybook

## How to test
1. Turn on the MSW and go to `/linodes`
2. Click on the page 2 button in the table, the url should update with the page number
3. Click on a Linode to go to the Linode's details page
4. Click the browser's back button, you should be redirected to the previous page you were on (in this case, page 2)
5. Update the url to `linodes?page=3`, the third page should load
6. Run storybook (`yarn storybook`), you should see `PaginationControl` under components

### DIFF
--- a/packages/manager/src/components/PrimaryNav/PrimaryNav.tsx
+++ b/packages/manager/src/components/PrimaryNav/PrimaryNav.tsx
@@ -131,7 +131,7 @@ export const PrimaryNav: React.FC<Props> = (props) => {
       [
         {
           display: 'Linodes',
-          href: '/linodes',
+          href: '/linodes?page=1',
           activeLinks: ['/linodes', '/linodes/create'],
           icon: <Linode />,
         },
--- a/packages/manager/src/features/linodes/LinodesLanding/DisplayLinodes.tsx
+++ b/packages/manager/src/features/linodes/LinodesLanding/DisplayLinodes.tsx
@@ -16,6 +16,8 @@ import IconButton from 'src/components/core/IconButton';
 import Tooltip from 'src/components/core/Tooltip';
 import GroupByTag from 'src/assets/icons/group-by-tag.svg';
 import TableView from 'src/assets/icons/table-view.svg';
+import queryString from 'query-string';
+import { useLocation } from 'react-router-dom';
 
 const useStyles = makeStyles((theme: Theme) => ({
   controlHeader: {
@@ -53,6 +55,7 @@ interface Props {
   toggleGroupLinodes: () => boolean;
   linodeViewPreference: 'grid' | 'list';
   linodesAreGrouped: boolean;
+  updateUrl: (page: number) => void;
 }
 
 type CombinedProps = Props & OrderByProps;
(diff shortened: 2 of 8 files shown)


tiffwong (approved):
Everything works as expected, great job! 🎉

bnussman (changes requested):
Found two things

- When using the back button in the browser, the table data does not update when I go back a page. 
- When comparing to production, changing page's feels a bit more delayed.
  - Haven't looked much at the implementation yet. Not exactly sure what can be done about this

codyfinn (commented):
When I click on a page link the linodes displayed change, but when I use the back and forward buttons the linodes displayed do not change.

packages/manager/src/features/linodes/LinodesLanding/DisplayLinodes.tsx
```diff
@@ -70,29 +75,36 @@ const DisplayLinodes: React.FC<CombinedProps> = (props) => {
...
     if (thisLinode.maintenance) {
       acc++;
     }
     return acc;
   }, 0);
+  const pageSize =
+    numberOfLinodesWithMaintenance > infinitePageSize
+      ? getMinimumPageSizeForNumberOfItems(numberOfLinodesWithMaintenance)
+      : infinitePageSize;
+  const maxPageSize = Math.ceil(count / pageSize);
```
DevDW, line 93: This variable refers to the greatest possible page number given the number of results, right? So if there were 12 results and the user had the page size set to 25, it'd be `12/25`  and get rounded up to 1?

If that is the case, I think this variable should be renamed to something a bit clearer, like `maxPageNumber`

DevDW (approved):
Looks to be working well 👍 left a comment about potentially renaming a variable for the sake of clarity

bnussman (changes requested):
When I am at `/linodes` and hit the next arrow on the pagination buttons, I get sent to the route `/linodes?page=NaN`

bnussman:
Wow sorry, I keep finding some weird things. Seems like changing the page size from `Show 25` to `Show 100` does not update the table data, only the pagination controls.

HanaXu:
@bnussman thanks for catching these issues early! 🕵️ 🔎

packages/manager/src/components/PrimaryNav/PrimaryNav.tsx
```diff
@@ -131,7 +131,7 @@ export const PrimaryNav: React.FC<Props> = (props) => {
       [
         {
           display: 'Linodes',
-          href: '/linodes',
+          href: '/linodes?page=1',
```
bnussman, line 134: I'd be in favor of reverting this to keep the url clean. I'd assume most customers don't have many pages anyways.

bnussman (approved):
Nice work! 🚀

PR #8050, opened 2021-11-23
M3-5492 & M3-5547: Axis label UI refinements

tiffwong:
## Description

This PR makes the line graph's axis label color match the one used for table headers and prevents the y-axis labels from being cutoff if there are more than 3 digits.

## How to test

1. Go to any Linode Detail view
2. Go to the "Analytics" tab if it's not on there already
3. Scroll down and inspect the font color
4. The "Disk IO" graph will usually have 3 digits in the y-axis but in case there isn't, make a new Linode


codyfinn (approved):
🚀

bnussman (approved):
Looks good! As an unrelated change, I think we should increase the contrast of this line in dark mode.
Screen Shot 2021-11-23 at 3 21 08 PM

PR #8051, opened 2021-11-23
M3-5549: Backup Auto Enrollment – Remove redundant head

codyfinn:
## Description

Removes the redundant header on the account settings backup auto enrollment section. There was also a tooltip that would show up to inform managed users that they are already auto enrolled in backups. I removed the tooltip and replaced it with a success notice.

## How to test

1. Open the Account Settings page as managed user.
2. There should be no redundant header in the back up auto enrollment section.
3. There should be a success notice informing you that you are already enrolled in auto backups.
4. Open the Account Settings page as an unmanaged user.
5. There should be no redundant header in the back up auto enrollment section.

### DIFF
--- a/packages/manager/src/components/Notice/Notice.tsx
+++ b/packages/manager/src/components/Notice/Notice.tsx
@@ -128,7 +128,7 @@ interface Props extends GridProps {
   flag?: boolean;
   notificationList?: boolean;
   spacingTop?: 0 | 8 | 16 | 24;
-  spacingBottom?: 0 | 8 | 16 | 24;
+  spacingBottom?: 0 | 8 | 16 | 20 | 24;
   breakWords?: boolean;
   onClick?: (e: React.MouseEvent<HTMLElement>) => void;
   // Dismissible Props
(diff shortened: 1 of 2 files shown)


bnussman (commented):
Looks solid, but I have two minor things
- The external link icon looks really big
- I feel like the spacing below the Managed notice could be a little bit less

Screen Shot 2021-11-23 at 3 09 30 PM

HanaXu (changes requested):
Agreed with Banks on removing the extra margin in the Notice

image

tiffwong (changes requested):
The changes looks great other than the spacing that's been mentioned already. 

Unrelated to the changes made but the link hover state is including the space.
Screen Shot 2021-12-01 at 4 58 34 PM

Also `createStyles` can be replaced by `useStyles` here instead since it's a FC and there's several classes that can be removed (including `classes.link`)

tiffwong (approved):

HanaXu (approved):

PR #8052, opened 2021-11-23
M3-5521: Contextual help for Linode Create

bnussman:
## Description

This PR logs a Google Analytics event when a user does any of the following on `/linodes/create`
- Clicks the `Use our speedtest page` in the region section
- Clicks the SMTP guide link in the checkout sidebar
- Views the IPAM tooltip in the VLANs section

## How to test

- View the dev console and make sure the debugging Google Analytics logs an event when any of those 3 actions listed above are performed. 
Screen Shot 2021-11-23 at 3 01 57 PM

### DIFF
--- a/packages/manager/src/components/TextField.tsx
+++ b/packages/manager/src/components/TextField.tsx
@@ -89,6 +89,7 @@ interface BaseProps {
   optional?: boolean;
   required?: boolean;
   tooltipText?: string;
+  tooltipOnMouseEnter?: React.MouseEventHandler<HTMLDivElement> | undefined;
   value?: Value;
 }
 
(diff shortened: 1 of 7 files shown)


packages/manager/src/components/TextField.tsx
```diff
@@ -89,6 +89,7 @@ interface BaseProps {
   optional?: boolean;
   required?: boolean;
   tooltipText?: string;
+  tooltipOnMouseEnter?: React.MouseEventHandler<HTMLDivElement> | undefined;
```
HanaXu, line 92: Can we rename this to something like `handleOnMouseEnter`?

HanaXu (approved):
LGTM! Just left one non-blocking comment on renaming a variable 📈

packages/manager/src/components/TextField.tsx
bnussman: I think this would be misleading because that prop is for the TextField component.

codyfinn (approved):
🚀

PR #8054, opened 2021-11-23
M3-5551: Refactor logic for displaying NVMe-related chips

DevDW:
## Description
In `VolumeTableRow.tsx`, the approach of using `notifications.some()` comes with some performance drawbacks since the array is looped through for every row, on every render.

This PR lifts that logic to `VolumesLanding.tsx` and uses a `.reduce()` to add booleans on the volume objects.

## To Do:
- [x] Testing for edge cases and ensuring there have been no regressions
- [x] Cleanup (removing comments, fixing types, etc.)

## How to test
Create a few HDD volumes (attached and unattached) and have them go through the full NVMe migration flow. The chips should show up just as they have been with no regressions.

### DIFF
--- a/packages/manager/src/features/Volumes/VolumesLanding.tsx
+++ b/packages/manager/src/features/Volumes/VolumesLanding.tsx
@@ -470,7 +470,7 @@ const filterVolumeEvents = (event: Event): boolean => {
   );
 };
 
-const includeNVMeBooleans = (
+const addNVMeBooleansToVolume = (
   volumes: Volume[],
   notifications: Notification[],
   events: Event[]
@@ -529,7 +529,7 @@ export default compose<CombinedProps, Props>(
       volumesResults,
       volumesError
     ) => {
-      const volumesWithNVMeBooleans = includeNVMeBooleans(
+      const volumesWithNVMeBooleans = addNVMeBooleansToVolume(
         volumesData,
         ownProps.notifications,
         ownProps.eventsData
--- a/packages/manager/src/features/Volumes/VolumesLanding.tsx
+++ b/packages/manager/src/features/Volumes/VolumesLanding.tsx
@@ -1,9 +1,4 @@
-import {
-  Event,
-  EventAction,
-  Notification,
-  NotificationType,
-} from '@linode/api-v4/lib/account';
+import { Event, Notification } from '@linode/api-v4/lib/account';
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/Volumes/VolumesLanding.tsx
```diff
@@ -461,13 +470,55 @@ const filterVolumeEvents = (event: Event): boolean => {
...
+  notifications: Notification[],
+  events: Event[]
+) => {
+  return volumes.reduce((acc: any, eachVolume) => {
+    const { id } = eachVolume;
+
+    const eligibleForUpgradeToNVMe = notifications.some(
+      (notification) =>
+        notification.type ===
+          ('volume_migration_scheduled' as NotificationType) &&
```
johnwcallahan, line 484: Why do you need to cast this this to `NotificationType`?

packages/manager/src/features/Volumes/VolumesLanding.tsx
```diff
@@ -461,13 +470,55 @@ const filterVolumeEvents = (event: Event): boolean => {
   );
 };
 
+const addNVMeBooleansToVolume = (
+  volumes: Volume[],
+  notifications: Notification[],
+  events: Event[]
+) => {
+  return volumes.reduce((acc: any, eachVolume) => {
```
johnwcallahan, line 478: I'd avoid using `any`. For this type of thing you could define an `ExtendedVolume` interface or similar, which is a regular `Volume` with the additional upgrade booleans.

johnwcallahan:
Nice job working out how to do this in VolumesLanding! The approach looks good; left a few minor-ish comments.

packages/manager/src/features/Volumes/VolumesLanding.tsx
DevDW: the casting was needed when this was in `VolumeTableRow.tsx`, but it doesn't need the casting here so I removed them -- thanks for pointing it out

packages/manager/src/features/Volumes/VolumesLanding.tsx
DevDW: Meant to circle back to changing the `any` to `ExtendedVolume[]` since I had updated the type previously to have the NVMe-related booleans, thanks for catching this

johnwcallahan (approved):

HanaXu (approved):
🚢

PR #8055, opened 2021-11-24
Fix Console Warnings

bnussman:
## Description

- Fixes two **massive** console errors in hopes to improve usability of dev tools
- One fix required a useless fragment which is unfortunate, but usability and styles should remain unchanged

This first error shows when you are on `/volumes` and you have >= 1 volume
Screen Shot 2021-11-23 at 10 12 01 PM

The second error shows on `/account/billing` and many other places in Cloud Manager
Screen Shot 2021-11-23 at 9 59 15 PM

## How to test

- Ensure all `InlineMenuAction` work as expected and have no visual regressions

### DIFF
--- a/packages/manager/src/components/InlineMenuAction/InlineMenuAction.tsx
+++ b/packages/manager/src/components/InlineMenuAction/InlineMenuAction.tsx
@@ -1,6 +1,7 @@
+/* eslint-disable react/jsx-no-useless-fragment */
 import * as React from 'react';
 import { Link } from 'react-router-dom';
-import Button from 'src/components/Button/Button.tsx';
+import Button from 'src/components/Button/Button';
 import { makeStyles, Theme } from 'src/components/core/styles';
 import Tooltip from 'src/components/core/Tooltip';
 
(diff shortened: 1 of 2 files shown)


packages/manager/src/components/InlineMenuAction/InlineMenuAction.tsx
```diff
@@ -79,14 +80,16 @@ const InlineMenuAction: React.FC<CombinedProps> = (props) => {
...
       >
-        <Button
-          className={`${className} ${classes.btnRoot}`}
-          onClick={onClick}
-          disabled={disabled}
-          loading={loading}
-        >
-          {actionText}
-        </Button>
+        <>
```
codyfinn, line 83: What does the fragment do here?

bnussman: I believe i'm preventing the Tooltip from forwarding a ref to the Button

DevDW: Another option seems to be using `forwardRef` in `Button.tsx`, like:

```
export const Button: React.FC<CombinedProps> = React.forwardRef(
  (props, ref) => { ... }
```

but I'm not sure that change is worth doing over the fragment

DevDW (approved):
🚢

HanaXu (approved):

tiffwong (approved):
🚀

PR #8056, opened 2021-11-29
M3-5522: Object Storage Landing contextual help

bnussman:
## Description

- Logs an event when the user clicks
  - The `Docs` button in the Object Storage Landing header
  - The `Learn more about storage options...` docs link in the Object Storage empty state

ObjectStorage_EmptyLanding_ContextualHelp

## How to test

Click one of the buttons described above and check the console to make sure Google Analytics logs the event. (You may not see this is the console if you have an ad blocker or use [Pi-hole](https://pi-hole.net). Google Analytics may be blocked preventing this from loading)

Screen Shot 2021-11-29 at 11 52 35 AM

### DIFF
--- a/packages/manager/src/utilities/ga.ts
+++ b/packages/manager/src/utilities/ga.ts
@@ -370,6 +370,10 @@ export const sendEntityTransferCopyDraftEmailEvent = (): void => {
 };
 
 export const sendHelpButtonClickEvent = (url: string, from?: string) => {
+  if (from === 'Object Storage Landing') {
+    sendObjectStorageDocsEvent('Docs');
+  }
+
   sendEvent({
     category: 'Help Button',
     action: url,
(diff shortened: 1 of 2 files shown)


packages/manager/src/utilities/ga.ts
```diff
@@ -370,6 +370,10 @@ export const sendEntityTransferCopyDraftEmailEvent = (): void => {
 };
 
 export const sendHelpButtonClickEvent = (url: string, from?: string) => {
+  if (from === 'Object Storage Landing') {
+    sendObjectStorageDocsEvent('Docs');
```
DevDW, line 374: I noticed that there's a separate, existing command that comes through from clicking this "Docs" button:

Screen Shot 2021-11-29 at 4 16 07 PM
 
it seems to accomplish the same goal as what we'd want this new one for -- do we need both of these?

DevDW (approved):
Looks good, left one comment regarding the necessity of the `Docs` one.

Also, in case anyone else has difficulty getting the commands logged in the console -- for me, I had to uncomment the Google Analytics credentials in my `.env` file and disable uBlock Origin.

packages/manager/src/utilities/ga.ts
bnussman: I agree with you on this but I close to log this event in two ways for the viewing pleasure in Google Analytics. Does anyone else have an opinion on this?

HanaXu (approved):
I can't seem to access the GA dashboard anymore but both events logged in the console 👍

PR #8057, opened 2021-11-29
Styling: border-radius of 1px for all button variants

johnwcallahan:
## Description

Previously "secondary" Buttons had a border-radius of 4px (coming from MUI) since the border-radius was not overwritten for that variant. 

This moves the `borderRadius: 1` specification to the `root` of the MUIButton class so it applies to all variants.

## How to test

**To test:** 
- look through the app to make sure there are visual regressions with buttons
- find a secondary button (e.g. "Add Linode to Firewall" drawer) and compare the border-radius to production (seen when you click and hold the button).

### DIFF
--- a/packages/manager/src/themeFactory.ts
+++ b/packages/manager/src/themeFactory.ts
@@ -527,6 +527,7 @@ const themeDefaults: ThemeDefaults = () => {
       MuiButton: {
         root: {
           border: 'none',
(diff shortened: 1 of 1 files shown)


tiffwong (approved):
Lgtm!

HanaXu (approved):
🕵️

PR #8060, opened 2021-11-30
M3-5609 PasswordInput component story

johnwcallahan:
## Description

- Convert PasswordInput story to MDX
- Remove unused styles in PasswordInput component

## How to test

Run Storybook and visit http://localhost:6006/?path=/docs/ui-elements-password-input--password-input. Run the app and ensure there are no regressions with the PasswordInput component.


bnussman:
This isn't urgent, but do we want the value field in the args table to update the actual component? 

https://user-images.githubusercontent.com/6440455/144460500-276aa9e4-1557-4e01-8f0f-acae4376e464.mov

DevDW (approved):
Aside from possibly having the input update based on what's typed into the "Control" field for `value` as Banks pointed out, this LGTM 🚢

packages/manager/src/components/PasswordInput/PasswordInput.stories.mdx
```diff
@@ -0,0 +1,99 @@
...
+      onChange: () => null,
+      required: false,
+      disabledReason: '',
+      hideStrengthLabel: false,
+      hideValidation: false,
+    }}
+    argTypes={{
+      value: {
+        description: 'The current value of the underlying TextField',
+      },
```
codyfinn, line 50: By adding a `control` field to the value object and setting a `type` field in there to null you will display the field and not allow the user to control the value.

```suggestion
      value: {
        description: 'The current value of the underlying TextField',
        control: {
          type: null,
        },
      },
```

codyfinn (changes requested):
Everything looks good. Just left a request for change to address the point that Banks brought up.

johnwcallahan:
Thanks everyone. I made a change that will let you control the component with the ArgsTable control. This results in a "controlled component" console error in the Story, but I think that's OK in this case.

codyfinn:
> Thanks everyone. I made a change that will let you control the component with the ArgsTable control. This results in a "controlled component" console error in the Story, but I think that's OK in this case.

I think it makes more sense to not allow the user to edit it in the ArgsTable since the point of the component isn't really to display the value so much as it is to capture it.

bnussman:
@codyfinn While I agree the point of the component isn't to display the value so much as it is to capture it, there is a benefit in allowing the ArgsTable to dictate the `value` for purposes of documenting our component library and the props components utilize.

bnussman (approved):

codyfinn (approved):

PR #8061, opened 2021-11-30
Cloud Manager v1.56.0 Staging --> Master

DevDW:
## Description
Merge `staging` into `master` for Cloud v1.56.0


HanaXu (approved):

bnussman (dismissed):

packages/manager/src/features/linodes/CloneLanding/utilities.test.ts
```diff
@@ -321,7 +321,7 @@ describe('utilities', () => {
     });
 
     it('handles edge cases', () => {
-      expect(getEstimatedCloneTime(0, 'sameDatacenter')).toBe('0 seconds');
+      expect(getEstimatedCloneTime(0, 'sameDatacenter')).toBe('0 seconds ago');
```
bnussman, line 324: This has actually been happening a lot on PRs. I usually just re-run the test for anyone whenever it happens. This is some weird edge case that randomly started happening a few weeks ago (I think). Can we update the test to allow both `0 seconds` **and**  `0 seconds ago`. 

With this change, were going to have the same issue. The function sometimes return `0 seconds` **and**  sometimes returns `0 seconds ago` too.

codyfinn: I think allowing either is just pushing the problem under the rug. We would be better served actually solving the cause.

bnussman: Agreed, it definitely should be solved on our end, but I believe (not 100% sure) that this is internal to Luxon

codyfinn: Yeah I think it is. Maybe the solve (for another time) would be to remove Luxon here at least. Could probably get the same results with the builtin Date class.

bnussman (approved):

HanaXu (approved):

PR #8063, opened 2021-11-30
M3-5588 Drawer component story

johnwcallahan:
## Description

This adds stories for the Drawer and NotificationDrawer components. I combined the stories on one page -- definitely open to feedback here.

There are also a few other things going on:

- Add support for MSW in Storybook. 
	- This helps the issue that @HanaXu pointed out which is that some components make API requests.
	- We’ll need to figure out how to address this long term when we publish the component library.
- Add support for an optional parameter in the hooks that drive the NotificationDrawer.
	- This is so I could provide mock notifications and events.
- Updated the eventFactory to include better mock data.

## How to test

Run Storybook and visit http://localhost:6006/?path=/docs/components-drawer--default-story.


packages/manager/src/features/NotificationCenter/NotificationData/useEventNotifications.tsx
```diff
@@ -18,8 +18,8 @@ const unwantedEvents: EventAction[] = [
   'volume_update',
 ];
 
-export const useEventNotifications = () => {
-  const { events } = useEvents();
+export const useEventNotifications = (givenEvents?: ExtendedEvent[]) => {
+  const events = givenEvents ?? useEvents().events;
```
codyfinn, line 22: Would changing the `useEvents` hook make more sense in this case? We could pass the `givenEvents` as an optional parameter to the hook. If the hook would otherwise return a falsey value it would instead return the `givenEvents`;

packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
```diff
@@ -26,14 +26,16 @@ export interface ExtendedNotification extends Notification {
...
+  givenNotifications?: Notification[]
+): NotificationItem[] => {
   const context = React.useContext(notificationContext);
   const {
     dismissNotifications,
     hasDismissedNotifications,
   } = useDismissibleNotifications();
 
-  const notifications = useNotifications();
+  const notifications = givenNotifications ?? useNotifications();
```
codyfinn, line 38: Same question as my comment here https://github.com/linode/manager/pull/8063/files#r761195212 but for the `useNotifications` hook.

packages/manager/src/features/NotificationCenter/NotificationData/useEventNotifications.tsx
codyfinn: I realized that `useEventNotifications` exemplifies the changes I am suggesting.

packages/manager/src/features/NotificationCenter/NotificationData/useFormattedNotifications.tsx
codyfinn: I realized that `useEventNotifications` exemplifies the changes I am suggesting.

packages/manager/src/features/NotificationCenter/NotificationData/useEventNotifications.tsx
johnwcallahan: Great feedback. My thinking was that if you already have event data you want to use (`givenEvents`), there would be no need to use the `useEvents` hook (same thing for your other comment about useNotifications).

`useEventNotifications` and `useFormattedNotifications` both return something other than raw events/notifications, so to me it made sense to optionally receive data. It would make unit testing these hooks easier as well.

HanaXu (changes requested):
The `show code` output is kind of unreadable 😅 

image

packages/manager/src/components/Drawer/Drawer.stories.mdx
```diff
@@ -0,0 +1,110 @@
...
+#### Overview
+
+- Drawers are essentially modal dialogs that appear on the right of the screen rather than the center.
+- Like traditional modals, they block interaction with the page content.
+- They are elevated above the app’s UI and don’t affect the screen’s layout grid.
+
+#### Behavior
+
+- Clicking a button on the screen opens the drawer.
+- Clicking off of the drawer closes it and data will not be saved.
```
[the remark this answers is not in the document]
johnwcallahan: True! Removed.

packages/manager/src/components/Drawer/Drawer.stories.mdx
```diff
@@ -0,0 +1,110 @@
...
+    decorators={decorators}
+    args={{
+      open: false,
+      onClose: () => null,
+      title: 'My Drawer',
+      wide: false,
+    }}
+    argTypes={{
+      open: {
+        description: 'Whether or not the Drawer is currently open',
```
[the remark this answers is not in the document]
johnwcallahan: I tried a few things to get around this, but wasn't able to find a good way to address it.

johnwcallahan:
> The `show code` output is kind of unreadable 😅

Good call; addressed.

packages/manager/src/components/Drawer/Drawer.stories.mdx
```diff
@@ -0,0 +1,149 @@
...
+          label="Input Some Text"
+          placeholder="This is a placeholder"
+        />
+        <ActionsPanel>
+          <Button buttonType="primary" onClick={close}>
+            Save
+          </Button>
+          <Button buttonType="secondary" onClick={close}>
+            Cancel
+          </Button>
```
bnussman, line 54: Update button order so action button is on the right

```suggestion
          <Button buttonType="secondary" onClick={close}>
            Cancel
          </Button>
          <Button buttonType="primary" onClick={close}>
            Save
          </Button>
```

packages/manager/src/components/Drawer/Drawer.stories.mdx
```diff
@@ -0,0 +1,149 @@
...
+#### Overview
+
+- Drawers are essentially modal dialogs that appear on the right of the screen rather than the center.
+- Like traditional modals, they block interaction with the page content.
+- They are elevated above the app’s UI and don’t affect the screen’s layout grid.
+
+#### Behavior
+
+- Clicking a button on the screen opens the drawer.
+- Drawers can also be closed by clicking the “X” icon or the “Close” button.
```
bnussman, line 28: Include `esc` key for closing drawers

```suggestion
- Drawers can be closed by pressing the `esc` key, clicking the “X” icon, or clicking the “Cancel” button.
```

packages/manager/src/components/Drawer/Drawer.stories.mdx
```diff
@@ -0,0 +1,149 @@
...
+    label="Input Some Text"
+    placeholder="This is a placeholder"
+  />
+  <ActionsPanel>
+    <Button buttonType="primary" onClick={close}>
+      Save
+    </Button>
+    <Button buttonType="secondary" onClick={close}>
+      Cancel
+    </Button>
```
bnussman, line 146: Update button order so action button is on the right here also

```suggestion
    <Button buttonType="secondary" onClick={close}>
      Cancel
    </Button>
    <Button buttonType="primary" onClick={close}>
      Save
    </Button>
```

bnussman (changes requested):
Non-blocking, but it would be nice if the `Notification Drawer` Story was nested under the `Drawer` section in Storybook's sidebar.

johnwcallahan:
@bnussman great feedback; addressed!

packages/manager/src/components/Drawer/Drawer.stories.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+<Canvas>
+  <Story
+    name="Default"
+    decorators={decorators}
+    parameters={{ docs: { source: { code: DrawerSourceCode } } }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
```
HanaXu, line 69: What about removing this and renaming the `Component API` story to `Default` since they basically seem to be the same? This would also cleanup the sidebar a bit (specifically in Canvas view, there is no difference between `Default` and `Component API`)

HanaXu (changes requested):

packages/manager/src/components/Drawer/Drawer.stories.mdx
tiffwong: It seems like the code generated from the ArgsTable never updates regardless of the changes made under the `Component API` example so I'm okay with moving it up. In that case, would we move the other examples below the ArgsTable (ie. for Buttons: primary, ArgsTable, secondary, outlined)?

tiffwong (approved):
I arbitrarily picked `###` to dictate "Overview" and other subheadings to be `h3` but we should probably decide on some standard so that everything looks cohesive. Otherwise everything looks good and amazing job getting MSW linked to Storybook!

bnussman (approved):
Looks good and great job with the MSW

packages/manager/src/components/Drawer/Drawer.stories.mdx
johnwcallahan: You're right that the code in the ArgsTable doesn't change, but the component does change (e.g. if you toggle "Wide" in the ArgsTable, the drawer will be updated accordingly), which is why I think a separate component specifically for the ArgsTable makes sense. It's less clear in this example with the Drawer since you have to click a button in order to see the drawer.

(Also, I'd love to eventually find a way to update the code when values in the ArgsTable changes!)

HanaXu (approved):

PR #8065, opened 2021-12-02
M3-5631: Add beta badge to HA Upgrade and HA Checkbox

codyfinn:
## Description

Add a beta badge for the HA Upgrade dialog and to the HA Checkbox in the Kubernetes Checkout Bar.

## How to test

1. Navigate to Kubernetes page.
2. Open Kubernetes creation.
3. Ensure that next to the `Enable HA Cluster Plane` label there is a `BETA` badge.
4. Navigate back to Kubernetes Landing page.
5. Open an existing non-HA enabled Cluster.
6. Click the `Upgrade to HA` button.
7. Ensure that next to the `Enable HA Cluster Plane` label there is a `BETA` badge.

### DIFF
--- a/packages/manager/src/components/PrimaryNav/PrimaryNav.tsx
+++ b/packages/manager/src/components/PrimaryNav/PrimaryNav.tsx
@@ -181,7 +181,6 @@ export const PrimaryNav: React.FC<Props> = (props) => {
           href: '/kubernetes/clusters',
           activeLinks: ['/kubernetes/create'],
           icon: <Kubernetes />,
-          isBeta: true,
         },
         {
           hide: !showDatabases,
--- a/packages/manager/src/components/PrimaryNav/PrimaryNav.tsx
+++ b/packages/manager/src/components/PrimaryNav/PrimaryNav.tsx
@@ -181,6 +181,7 @@ export const PrimaryNav: React.FC<Props> = (props) => {
           href: '/kubernetes/clusters',
           activeLinks: ['/kubernetes/create'],
           icon: <Kubernetes />,
+          isBeta: true,
         },
         {
           hide: !showDatabases,
--- a/packages/manager/src/features/Kubernetes/KubeCheckoutBar/HACheckbox.tsx
+++ b/packages/manager/src/features/Kubernetes/KubeCheckoutBar/HACheckbox.tsx
@@ -36,6 +36,7 @@ const useStyles = makeStyles((theme: Theme) => ({
   chip: {
     fontWeight: theme.typography.fontWeightRegular,
     margin: 0,
+    marginTop: -1,
   },
 }));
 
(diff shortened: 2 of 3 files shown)


tiffwong (changes requested):
Can we reduce the space between the text and the chip?
Screen Shot 2021-12-07 at 5 21 24 PM

Also it looks kinda odd having the chip on a newline by itself
Screen Shot 2021-12-07 at 4 05 00 PM

For smaller resolutions, can we increase the space between the text and the chip?
Screen Shot 2021-12-07 at 5 50 30 PM

bnussman (commented):
Would be nice to have the chip in a less awkward position here, but I don't really have any good solutions in mind 🤔

Screen Shot 2021-12-07 at 11 51 25 AM

HanaXu (commented):
What about removing the word `enable` to make room for the chip? It's kind of redundant anyway 🤔

codyfinn:
I think we are going to keep the Enable in the label. The text wrap looks ok. The spacing issue has been addressed.

tiffwong (changes requested):
The dialog looks good but can you add some more spacing between the text and the chip for the create flow? The chip also doesn't look vertically aligned with the text 
Screen Shot 2021-12-09 at 2 01 50 PM

HanaXu (changes requested):
Usually font-sizes for Chips are smaller, can we reduce the font-size to something like `0.65rem`?

Remove font-bold in the Chip and reduce font-size
image


image


Reduce font-size
image

image

DevDW (commented):
In the dialog, the alignment for the checkbox pieces seem a bit off. I think if "Enable HA Control Pane" is nudged down to the red line, it'll look balanced.

Screen Shot 2021-12-13 at 3 26 20 PM

Looks like the font weight of the chip in the Checkout bar needs to be reduced to match that of the chip in the dialog

tiffwong (changes requested):
The checkout bar chip still seems a little off, setting `margin: 0` seems to fix it
Screen Shot 2021-12-15 at 4 09 56 PM

DevDW (approved):
If you shrink the viewport on the "Create Cluster" page down enough that the summary shifts to the bottom of the page, the "Enable ..." text seems slightly misaligned with the "Beta" text:

Screen Shot 2021-12-17 at 5 32 18 PM

it looks like it's not exactly centered relative to the checkbox either, so maybe doing that will be enough to fix the alignment relative to the "Beta" text also?

Looks good everywhere else though so approving pending the tweak above 👍

tiffwong (approved):
Based on Dajahi's feedback, I'd say add `margin-top: -1` to the chip class.

Approving pending the adjustment mentioned above.

HanaXu (approved):

bcmertz (approved):
LGTM pending requested changes

tiffwong (approved):

PR #8067, opened 2021-12-03
Cypress Stability Improvements

bnussman:
## Description

- Adds more timeout to some requests because in these cases, Cloud Manager had not fully loaded yet. More timeout will allow Cloud Manager to load. 

- Uses a random Firewall name to ensure the API does not return `400` error for conflicting Firewall names
  - Cypress will delete the firewall after the test, but that was not enough (probably due to multiple e2es running at the same time)

## How to test

- Run test that changed locally, or just check the GitHub Actions of this PR

### DIFF
--- a/packages/manager/cypress.json
+++ b/packages/manager/cypress.json
@@ -7,7 +7,7 @@
   "viewportWidth": 1440,
   "viewportHeight": 900,
   "projectId": "5rhsif",
-  "requestTimeout": 10000,
+  "requestTimeout": 20000,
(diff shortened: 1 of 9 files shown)


HanaXu (approved):
LGTM! Can we up the `randomTitle` count 10 just in case 😆

bnussman:
Considering changing in `cypress.json`: 
```json
"requestTimeout": 10000,
```
```json
"requestTimeout": 20000,
```

PR #8068, opened 2021-12-03
M3-5591: Add Story for Paper

codyfinn:
## Description

Adds a story in storybook mdx format for the Paper component

## How to test

Ensure the Paper component has a story that renders correctly in storybook.

### DIFF
--- a/packages/manager/src/components/Paper.stories.mdx
+++ b/packages/manager/src/components/Paper.stories.mdx
@@ -11,6 +11,14 @@ export const PaperTemplate = (args, context) => (
 
 # Papers
 
+## Overview
+Papers are surfaces that display content and actions on a single topic.
+They should be easy to scan for relevant and actionable information. Elements like text and images should be placed on them in a way that clearly indicates hierarchy.
+
+## Usage
+- Papers are used for grouping information.
+- Papers allow for flexible layouts.
+
 <hr />
 
 <Canvas>
--- a/packages/manager/src/components/Paper.stories.mdx
+++ b/packages/manager/src/components/Paper.stories.mdx
@@ -25,12 +25,12 @@ They should be easy to scan for relevant and actionable information. Elements li
     <Story
       name="Paper"
       args={{
-        error: undefined
+        error: '',
       }}
       argTypes={{
           variant: {
             description: 'The variant to use when rendering thepaper.',
-            options: ['elevation', 'outlined', undefined],
+            options: ['outlined', undefined],
             control: {
               type: 'radio',
             },
--- a/packages/manager/src/components/Paper.stories.mdx
+++ b/packages/manager/src/components/Paper.stories.mdx
@@ -19,8 +19,6 @@ They should be easy to scan for relevant and actionable information. Elements li
 - Papers are used for grouping information.
 - Papers allow for flexible layouts.
 
-<hr />
-
 <Canvas>
     <Story
       name="Paper"
--- a/packages/manager/src/components/Paper.stories.mdx
+++ b/packages/manager/src/components/Paper.stories.mdx
@@ -3,4 +3,37 @@ import Paper from '/src/components/core/Paper';
 
 <Meta title="Components/Papers" component={Paper} />
 
+export const PaperTemplate = (args, context) => (
+    <div id='paper-story'>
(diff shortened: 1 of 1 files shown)


packages/manager/src/components/Paper.stories.mdx
```diff
@@ -3,4 +3,37 @@ import Paper from '/src/components/core/Paper';
 
 <Meta title="Components/Papers" component={Paper} />
```
tiffwong, line 4: Can you rename this from `Papers` to `Paper`? This way it own't create another level before you get to the docs

packages/manager/src/components/Paper.stories.mdx
```diff
@@ -3,4 +3,37 @@ import Paper from '/src/components/core/Paper';
...
+
 # Papers
+
+<hr />
+
+<Canvas>
+    <Story
+      name="Paper"
+      args={{
+        error: undefined
```
tiffwong, line 22: We never use the `error` prop from what I can tell so can we remove the related logic and styles from the doc and component?

tiffwong: We also have an outlined variant which should be added to the `ArgsTable`

tiffwong (changes requested):
Left some cleanup items

packages/manager/src/components/Paper.stories.mdx
HanaXu: Looks like we use `error` in the `PreviewReply` component. But since that's the only place it's used, I agree with Tiff on removing the `error` prop. We can just display the error in the `PreviewReply` component

bnussman (changes requested):
No need for the wrapping `<div />` here

packages/manager/src/components/Paper.stories.mdx
DevDW: What would be the benefit of removing the `error` prop here? I'm inclined to say we should keep it, unless we're going to be deprecating the `error` prop for the component and removing it from the `PreviewReply` component.

DevDW: @HanaXu Ah okay. Here's what it looks like in the app with an error:

Screen Shot 2021-12-13 at 3 41 35 PM

I would agree with removing the prop and cleaning up the `PreviewReply` component to use a `<Notice />` instead.

The way it is currently doesn't work too well; there's an error indicated but it still shows a preview anyway.

So if there is an error, I think `<PreviewReply />` should render an error notice instead of the paper with markdown.

codyfinn: I can't seem to get the error to show up. I tried to upload a giant file and that gave me an error but that showed up as a notice in the reply page. Not sure what the error does in the `PreviewReply` case.

bnussman:
Looks solid! Last thing to add would be the proposed copy from Jira.

bnussman (approved):

packages/manager/src/components/Paper.stories.mdx
```diff
@@ -1,6 +1,49 @@
...
+- Papers are used for grouping information.
+- Papers allow for flexible layouts.
+
+<hr />
+
+<Canvas>
+    <Story
+      name="Paper"
+      args={{
+        error: undefined
```
HanaXu, line 28: We can make a tech debt ticket to address removing the error prop from the Paper component. For now, can we set the error to be an empty string so that there's one less click to update the string?

image

HanaXu (changes requested):

packages/manager/src/components/Paper.stories.mdx
```diff
@@ -1,6 +1,49 @@
...
+
+## Overview
+Papers are surfaces that display content and actions on a single topic.
+They should be easy to scan for relevant and actionable information. Elements like text and images should be placed on them in a way that clearly indicates hierarchy.
+
+## Usage
+- Papers are used for grouping information.
+- Papers allow for flexible layouts.
+
+<hr />
```
tiffwong, line 22: Can we remove this divider?

tiffwong (approved):
Left one optional cleanup item but otherwise lgtm!

HanaXu (approved):

PR #8069, opened 2021-12-03
M3-5585: Add Story for Navigation

tiffwong:
## Description

This PR adds stories for all things menu related (ie. `TopMenu`, `SideMenu`, and `UserMenu`) along with mock profile data.

**Update:** Moved the controls for the `SideMenu` to the `ArgsTable` and also imported the Storybook Controls add-on so the props can also be adjusted under the Canvas tab. This was implemented to account for the `SideMenu` leaking out of the container for viewports less than 960px wide. 

The latest version of the Storybook Controls add-on causes react-router-dom to break so we're using an older version.
[Storybook v6.4 is breaking users using react-router v5](https://github.com/storybookjs/storybook/issues/16837)

## How to test

- Run `yarn storybook` and go to [http://localhost:6006/?path=/docs/components-navigation-top-menu--top-menu](http://localhost:6006/?path=/docs/components-navigation-top-menu--top-menu)
- See the stories for `TopMenu`, `SideMenu`, and `UserMenu`
- I added a conditional for undefined `gravatarURL`s so check for any regressions related to that


codyfinn (commented):
The buttons are a bit strange given how most other things are driven by the args tables. If the buttons are necessary it would be nice if they were styled like our normal buttons.

bnussman (commented):
Looks really good, but for the `Side Menu` story, it would be really awesome if we could use an args table and/or Sorybook's [controls](https://storybook.js.org/docs/react/essentials/controls) to control it instead of the buttons in the middle of the view. Would this be possible?

bnussman (commented):
I've also run into those Storybook dependency issues you listed above. Hopefully that's something we can easily address soon. 

In the Side Menu Story, on a small screen size and `open` is `true`, it takes focus over the entire view. Is there any solution for that?

Screen Shot 2021-12-13 at 11 14 35 AM

codyfinn (approved):
Definitely prefer the controls to the buttons on the canvas before! Non-blocking but would adding the [viewport plugin](https://storybook.js.org/docs/react/essentials/viewport) be worth it for stories like these? It would allow the user to play with the smaller screen sizes without having to shrink their browser's window.

tiffwong:
@bnussman I noticed that too but I wasn't able to figure out a fix for it.

@codyfinn the viewport add-on is already installed, it's the rightmost button
Screen Shot 2021-12-14 at 5 35 30 PM

bnussman (approved):

PR #8071, opened 2021-12-06
M3-5610: Storybook for Radios

bnussman:
## Description

- Updates Storybook for the Radio component
- Now uses mdx
- Fixed padding on radio causing ripple to be a non-square size
- Use makeStyles() in the`<Radio />` component
- Removed the `error` and `warning` variant

## How to test

- Check for regressions with `<Radio />`
- View all Radio stories in Storybook (`yarn storybook`)

## 🔎 **Preview**
[https://m3-5610-update-story-for-rad.manager-ec2.pages.dev](https://m3-5610-update-story-for-rad.manager-ec2.pages.dev/?path=/docs/ui-elements-radio--radio#radio)

## Args Notes

It would be nice if Storybook automatically knew **all** of the types we generally extend in our own components (in this case all ButtonProps) so we don't have to manually populate the args. Seems like we're not the only ones having this issue. Typescript isn't inferring all prop types, so our args table is not completely populated with all possible props form MUI that can be used on the component 😕

[This fix](https://storybook.js.org/docs/ember/configure/typescript#overriding-the-configuration-to-infer-additional-props) did not seem to work on my initial attempt 

https://github.com/storybookjs/storybook/issues/12558
https://github.com/reactjs/react-docgen/issues/416
https://github.com/storybookjs/storybook/tree/next/addons/docs/react#typescript-props-with-react-docgen


packages/manager/src/themeFactory.ts
```diff
@@ -1078,6 +1078,12 @@ const themeDefaults: ThemeDefaults = () => {
           color: primaryColors.main,
           '&$checked': {
             color: primaryColors.main,
+            '&:hover': {
+              backgroundColor: 'rgba(36, 83, 233, 0.04)',
+            },
+          },
+          '&:hover': {
+            backgroundColor: 'rgba(36, 83, 233, 0.04)',
```
bnussman, line 1086: Any idea on how I could clean this up without using an arbitrary color?

packages/manager/src/components/Radio/Radio.stories.mdx
```diff
@@ -0,0 +1,125 @@
...
+- Expose all available options
+- Select a single option from a list
+ 
+### Guidelines
+- If there are 3 or fewer items to select, use radio buttons rather than drop-down menus.
+- If possible, offer a default selection.
+- Because radio buttons allow only one choice, make sure that the options are both comprehensive and distinct.
+- Let users select an option by clicking on either the button itself or its label to provide as big a target area as possible.
+ 
+### Reasons for A Default Selection
```
tiffwong, line 31: Just saw that the "A" is also capitalized in the ticket but can you make it lowercase?

packages/manager/src/components/Radio/Radio.stories.mdx
```diff
@@ -0,0 +1,125 @@
...
+- Expedite tasks
+- Give people control and align with their expectations
+
+## Radio
```
tiffwong, line 35: Since there's the header above, we probably don't need this one since it's a duplicate

packages/manager/src/components/Radio/Radio.stories.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+<Canvas>
+  <Story name="Radio">
+   <Radio />
+  </Story>
+</Canvas>
+
+## Controled Radio
```
tiffwong, line 43: I think "controlled" is spelled with two `l`'s

packages/manager/src/components/Radio/Radio.stories.mdx
```diff
@@ -0,0 +1,125 @@
...
+
+## Component API
+
+<Canvas>
+  <Story
+    name="Component API"
+    args={{
+      checked: false,
+      disabled: false,
+      color: "primary",
```
tiffwong, line 73: I don't think we ever use the `color` prop for `Radio` so we can remove it from the code and the docs

tiffwong: Actually I see that we use it inside the `Radio` component itself so we could just hide it from the `ArgsTable` since the prop is always `primary`

packages/manager/src/themeFactory.ts
tiffwong: I'm not seeing these styles be applied 🤔 

Also `colorSecondary` seems to apply to `Radio`'s that have `color="secondary"` so I don't think anything of those styles are ever used on Cloud so we can probably remove that whole section

tiffwong (changes requested):
Left some items that can be cleaned up a bit. 

Also I think we can exclude the following props from the `ArgsTable`:
- `color`
- `disableRipple`
- `inputProps`
- `inputRef`
- `size` (this has no effect from what I can see)
- `disableFocusRipple`
- `centerRipple`
- `disableTouchRipple
- `focusRipple`
- `focusVisibleClassName`
- `onFocusVisible`
- `TouchRippleProps`

packages/manager/src/themeFactory.ts
bnussman: I made this change to fix this pink background around some Radios

Here's an example:
Screen Shot 2021-12-09 at 12 02 28 PM

Should we fix this in the styles the way I did, or should we address this in the file the component is used in?

To fix it without modifying our theme, I would have to replace 
```ts
control={<Radio />}
``` 
with
```ts
control={<Radio color="primary" />}
``` 
on line `242` and `283` of `packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/AddIPDrawer.tsx`

What do you think @tiffwong

packages/manager/src/components/Radio/Radio.stories.mdx
```diff
@@ -121,5 +119,18 @@ export const Template = (args) => {
   exclude={[
     'ref',
     'action',
+    'edge',
```
HanaXu, line 122: Can we remove `edge` from exclude since it does change the UI? And add `required` to exclude since there's no visual change

https://user-images.githubusercontent.com/14323019/145465159-b1465f37-75fb-4598-a9e4-9d360c300605.mov

HanaXu (changes requested):
Can we fix the Controlled Radio and Radio Group being cut off in Canvas view?

image

image

packages/manager/src/components/Radio/Radio.stories.mdx
bnussman: I was thinking the `required` prop may have some accessibility implications, but I went ahead and removed it

packages/manager/src/themeFactory.ts
tiffwong: In that case, let's go with the way you originally had it. The only way to not use an arbitrary code is to create a variable but unless we plan to reuse the color for something else, the way you did it is fine

packages/manager/src/components/Radio/Radio.stories.mdx
tiffwong: Hmm might be nice to keep required just so we know it's an option

tiffwong (approved):
Great work on this! 🎉

bnussman:
## 🔎 **Preview**
[https://m3-5610-update-story-for-radio.manager-ec2.pages.dev/](https://m3-5610-update-story-for-radio.manager-ec2.pages.dev/)

HanaXu (approved):
🚀

PR #8073, opened 2021-12-08
Override Domain type display in search results

johnwcallahan:
## Description

I noticed we were displaying the raw `domain.type` in search results (in the search bar and on the search landing page). This PR overrides that to ensure we display `"primary" | "secondary"` instead of `"master" | "slave"` which is a change we’ve made everywhere else in the app.

## How to test

Have a primary and a secondary Domain on your account and search for them in the search bar. Production will display the type as being either 'master' or 'slave', while in this PR they should be displayed as 'primary' or 'secondary'.


bnussman (approved):
Good catch, now shows `primary` and `secondary`.

HanaXu (approved):
🔍

PR #8077, opened 2021-12-14
DBAAS1-73: Route / component scaffolding

tiffwong:
## Description

Add routes and placeholder pages/components for
- cloud.linode.com/databases
- cloud.linode.com/databses/create
- cloud.linode.com/databases/{id} (redirects to /summary)
- cloud.linode.com/databases/{id}/summary
- cloud.linode.com/databases/{id}/backups
- cloud.linode.com/databases/{id}/settings

## How to test

Test the urls while toggling the flag.


HanaXu (commented):
For `/databases/{id}` I see the redirect to summary
image

But trying to access `databases/{id}/summary` directly, I see this instead of the summary. I'm also seeing this error for `databases/{id}/backups` and `databases/{id}/settings` 🤔 
image

packages/manager/src/features/Databases/DatabaseDetail/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+// import { DocumentTitleSegment } from 'src/components/DocumentTitle';
+import SafeTabPanel from 'src/components/SafeTabPanel';
+import TabLinkList from 'src/components/TabLinkList';
+import useFlags from 'src/hooks/useFlags';
+
+const DatabaseSummary = React.lazy(() => import('./DatabaseSummary'));
+const DatabaseBackups = React.lazy(() => import('./DatabaseBackups'));
+const DatabaseSettings = React.lazy(() => import('./DatabaseSettings'));
+
+type CombinedProps = RouteComponentProps<{ id: string }>;
```
bnussman, line 14: Can we use a hook inside this component instead of `RouteComponentProps`? I think `useLocation` is what we'd use.

packages/manager/src/features/Databases/index.tsx
```diff
@@ -1,11 +1,13 @@
...
 import { DocumentTitleSegment } from 'src/components/DocumentTitle';
 import SuspenseLoader from 'src/components/SuspenseLoader';
+import useFlags from 'src/hooks/useFlags';
 import useReduxLoad from 'src/hooks/useReduxLoad';
 
 const DatabaseLanding = React.lazy(() => import('./DatabaseLanding'));
 const DatabaseDetail = React.lazy(() => import('./DatabaseDetail'));
+const DatabaseCreate = React.lazy(() => import('./DatabaseCreate'));
 
 type Props = RouteComponentProps<{}>;
```
bnussman, line 12: Maybe use a hook here instead I think this one would be `useRouteMatch`.

packages/manager/src/features/Databases/index.tsx
```diff
@@ -1,28 +1,28 @@
...
+  }
 
   return (
     <React.Suspense fallback={<SuspenseLoader />}>
       <DocumentTitleSegment segment="Databases" />
-      <Route component={DatabaseLanding} exact path={path} />
-      <Route component={DatabaseDetail} path={`${path}/:id`} />
+      <Switch>
+        <Route component={DatabaseCreate} path={`/databases/create`} />
+        <Route component={DatabaseDetail} path={`/databases/:databaseId`} />
```
bnussman, line 23:
```suggestion
        <Route component={DatabaseCreate} path="/databases/create" />
        <Route component={DatabaseDetail} path="/databases/:databaseId" />
```

packages/manager/src/features/Databases/DatabaseDetail/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+interface Props {}
+
+type CombinedProps = Props;
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const DatabaseDetail: React.FC<CombinedProps> = (props) => {
+  const history = useHistory();
+  const params = useParams<{ databaseId: string }>();
+  const databaseId = params.databaseId;
```
bnussman, line 23:
```suggestion
  const { databaseId } = useParams<{ databaseId: string }>();
```

bnussman (commented):
Added a few cleanup items. 

Can we make the `Add a Database` button link to `/databases/create` instead of opening the old Create Dialog? 

Screen Shot 2021-12-14 at 1 06 11 PM

packages/manager/src/features/Databases/DatabaseDetail/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const matches = (p: string) => {
+    return Boolean(matchPath(p, { path: location.pathname }));
+  };
+
+  const clampTabChoice = () => {
+    const tabChoice = tabs.findIndex((tab) => matches(tab.routeName));
+    return tabChoice < 0 ? 0 : tabChoice;
+  };
+
+  const navToURL = (index: number) => {
```
HanaXu, line 48: Can we rename this to something like `handleTabChange`?

packages/manager/src/features/Databases/DatabaseDetail/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+      title: 'Backups',
+      routeName: `/databases/${databaseId}/backups`,
+    },
+    {
+      title: 'Settings',
+      routeName: `/databases/${databaseId}/settings`,
+    },
+  ];
+
+  const matches = (p: string) => {
```
HanaXu, line 39: Since this is only used in one place, I don't think `matches` needs to be a function

HanaXu (approved):
LGTM, left some optional cleanup comments 🚢

DevDW (approved):
The navigation is working well 👍 

One thing I noticed is that in the case where a user is on the Database Detail page and goes to an invalid route (say, `/databases/100/testing`), they get rerouted to the Summary tab, but the URL doesn't update accordingly.

I agree with the couple of variable renamings that Hana suggested as well

tiffwong:
@DevDW I just checked a bunch of tabs and it seems like that's a problem for all of the components that use `<SafeTabPanel />` vs `<NavTabs />` so I'll create a separate ticket to address this

PR #8080, opened 2021-12-14
DBAAS1-74: JS Client & Types scaffolding for DBaaS

DevDW:
## Description
Adds DBaaS-related JS Client methods and types based on spec doc.

## Note to Reviewers
The app will compile successfully, but there are numerous type checking errors in old DBaaS files. These will go away as those files get deleted or cleaned up in other PRs and those PRs get merged.


packages/api-v4/src/databases/databases.ts
```diff
@@ -127,13 +158,76 @@ export const deleteDatabase = (databaseID: number) =>
   );
 
 /**
- * resetDatabasePassword
+ * getDatabaseBackups
+ *
+ * Return backups information for a database
+ *
+ */
+export const getDatabaseBackups = (databaseID: number) =>
```
bnussman, line 166: To match functions like `getMySQLDatabases`, should this be `getMySQLDatabaseBackups`? I'm not sure how we're planning to handle this when we support many types of DBs... Same applies to a few functions

DevDW: Just confirmed in a meeting with API that each engine will have its own namespace, so I'm going to rename several of these to be MySQL-specific where needed (basically anything that is `/databases/mysql ...` will have its method name reflect MySQL in it)

bnussman: Sounds good! Thanks for checking on that.

In the future we may need a pattern something like this though 🤔
```ts
export const getDatabaseBackups = (type: 'mysql' | 'postgresql' | 'redis', databaseID: number) =>
  Request<Page<DatabaseBackup>>(
    setURL(`${API_ROOT}/databases/${type}/instances/${databaseID}/backups`),
    setMethod('GET')
  );
```

DevDW: It's probably a good idea to structure it similar to that now instead of having to refactor it down the line actually

HanaXu: Agreed on restructuring it now

packages/api-v4/src/databases/types.ts
```diff
@@ -6,84 +6,113 @@ export interface DatabaseType {
...
+export type DatabaseStatus =
+  | 'creating'
   | 'running'
-  | 'starting'
   | 'failed'
-  | 'succeeded'
-  | 'unknown';
+  | 'degraded'
+  | 'updating'
+  | 'deleted';
```
bnussman, line 38: I don't think we'll want `deleted` included because it will never be returned publicly

bnussman (approved):
Looks good to me! Left one minor change to consider.

tiffwong (approved):

PR #8081, opened 2021-12-14
DBAAS1-75: React Query for DBaaS

bnussman:
## Description

- Adds React Query hooks for DBaaS
- Fully API paginate and filter the Databases landing table

## How to test

- Can't really test the queries till they are used in components.
- Once the MSW has handlers for Databases, we can test this more.


packages/manager/src/queries/databases.ts
```diff
@@ -0,0 +1,145 @@
...
+            return { ...oldEntity, ...data };
+          }
+        );
+      },
+    }
+  );
+
+export const useCreateDatabaseMutation = () =>
+  useMutation<CreateDatabaseResponse, APIError[], CreateDatabasePayload>(
+    (data) => createDatabase(data.engine || 'mysql', data),
```
codyfinn, line 71: Could this be a default parameter instead of an OR?

packages/api-v4/src/databases/databases.ts
```diff
@@ -169,10 +169,17 @@ export const deleteDatabase = (engine: Engine, databaseID: number) =>
  * Return backups information for a database
  *
  */
-export const getDatabaseBackups = (engine: Engine, databaseID: number) =>
+export const getDatabaseBackups = (
+  engine: Engine,
+  databaseID: number,
+  params?: any,
+  filters?: any
```
codyfinn, line 176: Are params and filters best described as type `any`? Maybe a better type would be `unknown`. It would have the impact that the compiler would not allow us to manipulate the the data passed to those arguments without first checking that it has the shape you anticipate.

packages/manager/src/queries/databases.ts
DevDW: I think the default parameter would be good as well

DevDW (approved):
Looks good -- I tested it with my mocks and the Database landing table was populated, and I used `useDatabaseQuery` in `DatabaseSummary.tsx` and logged a database object to the console successfully. The corresponding RQ queries looked as expected as well

packages/api-v4/src/databases/databases.ts
bnussman: I agree with this completely. I used `any` because it seems to be the go-to for api-v4 but in the future we should look into establishing types for `params` and `filter`. Going to leave as `any` right now to keep scope focused.

codyfinn (approved):

PR #8082, opened 2021-12-15
[M3-5534, M3-5692, M3-5709, M3-5760]: IPv6 sharing support

bcmertz:
### Description
This PR adds the ability to share IPv6 ranges and feature flags it. Removes range deletion option for /116's. Reworks how we get IPv6 range data to use proper IPv6 endpoints, as /linode/instances/${id}/ips surfaces ipv6 information that is not useful here despite appearing nearly identical to the information it surfaces for ipv4

### How to test
Test against staging! 🎉

### Double check
- `refreshIPs`: `Promise<void>[]` return type for  works as expected everywhere it is consumed. Asynchronous reducer and nested requests work as expected

- `sharedRanges`, `availableRanges`, `staticRanges` constructs are good additions and appropriately used

- Feature flag properly toggles IPv6 sharing functionality but doesn't break general IPv6 functionality. IPv4 remains working the same

- IPv6 loading and error states are handled correctly and don't interfere with preexisting loading and error states

- Sharing statically routed v6 ranges correctly applies the share to the Linode statically routed to, AND THEN the Linode being shared to explicitly by the user

<del>

- `CancellableRequest` construct is used properly, as this is the first usecase of it in our codebase

- `GetLinodeIPsInstantiation` type is defined in the correct place to match our code conventions

- `IPSharing.tsx` useEffect lifecycle doesn't introduce memory leaks (from uncanceled requests after unmounting) or cause race conditions (from concurrent requests). These should both be handled by cancelling previous requests in the useEffect cleanup function

- `getIPChoicesAndLabels` has moved to being an asynchronous function that reduces on awaited promises. It's a tricky pattern and should be paid special attention
</del>

### Future Notes
We can't determine if a DC supports IPv6 sharing until API adds IPv6 Sharing to the capabilities array in the /regions endpoint. So for now we let requests in DCs that don't support IPv6 sharing fail and present the relevant error message that "This Linode does not support IPv6 Sharing at this time. Please contact support."


packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -105,20 +86,79 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+    true // always make the request so we have the data preloaded and don't need to wait to make further requests
   );
 
+  let isLoading = resp.isLoading;
+  const data = resp.data;
+
+  const getIPChoicesAndLabels = async (linodeID: number, linodes: Linode[]) => {
+    const choiceLabels = {};
+    const promisedChoices = linodes
+      .filter((thisLinode: Linode) => {
```
codyfinn, line 103: can this `.filter.map` be replaced by a `.reduce` instead? That way we don't go through the array twice.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -105,20 +86,79 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+      })
+      .map(async (thisLinode: Linode) => {
+        // side-effect of this mapping is saving the labels
+        let ips: string[] = [];
+        const ipv6Ranges: string[] = [];
+        if (flags.ipv6Sharing) {
+          // this fails for non datacenters that don't support sharing IPv6 addresses
+          // for now we just default to the given error message rather than preventing
+          // this from running
+          await getLinodeIPs(thisLinode.id).then((ips) => {
```
codyfinn, line 115: Rather than mixing promise chains with async/await could this be done using only async/await?

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
```diff
@@ -471,6 +468,13 @@ class LinodeNetworking extends React.Component<CombinedProps, State> {
     const sharedIPs = uniq<string>(
       pathOr([], ['ipv4', 'shared'], linodeIPs).map((i: IPAddress) => i.address)
     );
+    const sharedIPv6 = uniq<string>(
+      pathOr([], ['ipv6', 'global'], linodeIPs)
+        .filter((i: IPRange) => {
```
codyfinn, line 473: Another possible simplification to use `reduce` instead of `.filter.map`

codyfinn (changes requested):
Just a couple of small changes but overall looks good

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
bcmertz: done

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
bcmertz: great idea, done!

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
bcmertz: nice, done!

bcmertz:
@codyfinn any idea why we previously didn't show IP Sharing choices when there was 1 available IP address? Is there any reason to keep that around?

codyfinn:
> @codyfinn any idea why we previously didn't show IP Sharing choices when there was 1 available IP address? Is there any reason to keep that around?

I do not know the reason. Maybe because if you only have one IP to share then you only have one linode? @DevDW might be a better person to ask.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -105,20 +86,81 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+        let ips: string[] = [];
+        const ipv6Ranges: string[] = [];
+        if (flags.ipv6Sharing) {
+          // this fails for non datacenters that don't support sharing IPv6 addresses
+          // for now we just default to the given error message rather than preventing
+          // this from running
+          const linodeIPs = await getLinodeIPs(thisLinode.id);
+          const ranges = pathOr([], ['ipv6', 'global'], linodeIPs);
+          ranges &&
+            ranges.forEach((i: IPRange) => {
```
codyfinn, line 119: I think both of these could be replaced with optional chaining.
`const ranges = linodeIPs?.ipv6?.global`

`ranges?.forEach((i: IPRange) => {...`

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -105,20 +86,81 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+
+        // side-effect of this mapping is saving the labels
+        let ips: string[] = [];
+        const ipv6Ranges: string[] = [];
+        if (flags.ipv6Sharing) {
+          // this fails for non datacenters that don't support sharing IPv6 addresses
+          // for now we just default to the given error message rather than preventing
+          // this from running
+          const linodeIPs = await getLinodeIPs(thisLinode.id);
+          const ranges = pathOr([], ['ipv6', 'global'], linodeIPs);
```
codyfinn, line 117: I think both of this could be replaced with optional chaining.
`const ranges = linodeIPs?.ipv6?.global ?? []`

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -105,20 +86,81 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+          ranges &&
+            ranges.forEach((i: IPRange) => {
+              ipv6Ranges.push(`${i.range}/${i.prefix}`);
+            });
```
codyfinn, line 121: This could be replaced with the `array.map` function assigning it's evaluation to the ipv6Ranges variable.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -105,20 +86,81 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+      Promise.resolve([])
+    );
+
+    return Promise.all(promisedChoices).then((ips) => {
+      return {
+        ipChoices: flatten<string>(ips),
+        ipChoiceLabels: choiceLabels,
+      };
+    });
+  };
```
codyfinn, line 140: This is not a required change at all but I think you could leverage the reduce function a bit more to build up the object that is returned at the end of the function as opposed to creating two arrays.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
```diff
@@ -471,6 +468,17 @@ class LinodeNetworking extends React.Component<CombinedProps, State> {
     const sharedIPs = uniq<string>(
       pathOr([], ['ipv4', 'shared'], linodeIPs).map((i: IPAddress) => i.address)
     );
+    const sharedIPv6 = uniq<string>(
+      pathOr([], ['ipv6', 'global'], linodeIPs).reduce(
```
codyfinn, line 472: Another place I think the optional chaining operator could clean the code up a bit

DevDW:
> > @codyfinn any idea why we previously didn't show IP Sharing choices when there was 1 available IP address? Is there any reason to keep that around?
> 
> I do not know the reason. Maybe because if you only have one IP to share then you only have one linode? @DevDW might be a better person to ask.

Hmm, I am not sure on the reasoning behind not showing it. That line of code came in with https://github.com/linode/manager/pull/7520, but I don't see any discussion surrounding that bit specifically. I'd check with API in case there's a backend reason for it, but my initial thought is that there's no reason why we shouldn't show the available IP address.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
bcmertz: done

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
```diff
@@ -471,10 +468,17 @@ class LinodeNetworking extends React.Component<CombinedProps, State> {
     const sharedIPs = uniq<string>(
       pathOr([], ['ipv4', 'shared'], linodeIPs).map((i: IPAddress) => i.address)
     );
+    const ranges = linodeIPs?.ipv6?.global ?? [];
+    const sharedIPv6 = uniq<string>(
+      ranges.reduce((acc: string[], i: IPRange) => {
+        if (i.route_target !== null) {
+          return acc;
+        }
+        return acc.concat(`${i.range}/${i.prefix}`);
```
codyfinn, line 477: Could this be made more concise with a ternary something like:
 ```suggestion
        if (i.route_target !== null) {
          return acc;
        }
        return acc.concat(`${i.range}/${i.prefix}`);
        return i.route_target ? acc.concat(`${i.range}/${i.prefix}`) : acc;
```

**Review by codyfinn — APPROVED** [review]
I left a small suggestion but the code looks good with or without it. Nice job!

**Comment by DevDW** [comment]
Hoping to take a look on Monday -- devenv refused to run properly for me today 😔

File: packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx [hunk 10]
```diff
@@ -105,20 +87,81 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
     linodeSharedIPs,
   } = props;
 
-  const { data, isLoading } = useAllLinodesQuery(
+  // will later be set if we make getLinodeIPsRequest
+  // to figure out the shareable ipv6 ranges
+  // we check against this on unmount to cancel any outstanding requests
```
bcmertz, line 92: test

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -103,22 +84,72 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+  const availableRangesMap = {};
+  availableRanges.forEach((range) => {
+    if (availableRangesMap.hasOwnProperty(range.linodes[0])) {
+      availableRangesMap[range.linodes[0]].push(
+        `${range.range}/${range.prefix}`
+      );
+    } else {
+      availableRangesMap[range.linodes[0]] = [`${range.range}/${range.prefix}`];
+    }
+  });
```
codyfinn, line 99: Can this be pulled out into a function that is not inside a component? That way the component logic is simpler and the functionality can be tested more easily. Also I think this might be another good place to use the `.reduce` function.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -103,22 +84,72 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+          });
+      }
+    });
+    linodeSharedIPs.forEach((range) => {
+      if (!choiceLabels.hasOwnProperty(range)) {
+        choiceLabels[range] = '';
+      }
+    });
+    return choiceLabels;
+  };
```
codyfinn, line 137: Can this be pulled out into a function that is not inside a component? That way the component logic is simpler and the functionality can be tested more easily. Also I think this might be another good place to use the `.reduce` function.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -103,22 +86,42 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+    availableRanges
+  );
+
+  const resp = useAllLinodesQuery(
     { page_size: API_MAX_PAGE_SIZE },
     {
       region: linodeRegion,
     },
-    open // Only run the query if the modal is open
+    true // always make the request so we have the data preloaded and don't need to wait to make further requests
```
bnussman, line 101: If functionality is not compromised, we should still lazy load this.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
```diff
@@ -229,10 +244,69 @@ class LinodeNetworking extends React.Component<CombinedProps, State> {
...
+            let slaac;
+            if (!resp.is_bgp) {
+              await getLinodeIPs(this.props.linode.id).then((ips) => {
+                slaac = ips?.ipv6?.slaac.address;
+              });
+            }
+
+            // any range that is shared to this linode
+            if (resp.is_bgp && resp.linodes.includes(this.props.linode.id)) {
+              this.props.flags.ipv6Sharing && sharedRanges.push(range);
```
bnussman, line 272: I don't know how this should work, but to me, its confusing that when a Linode that "owns" an IPv6 range, if the IP is shared, in the Share IP modal, the IP shows up there... Should the shared range only show on the Linode's modal that doesn't "own the IP"... 

The code below explains what I mean, but needs modification in `ipResponseToDisplayRows` to work as I'm explainig

```suggestion
            if (
              resp.is_bgp &&
              resp.linodes.includes(this.props.linode.id) &&
              !this.state.linodeIPs?.ipv6?.global.some(
                (thisRange) => thisRange.range === range.range
              )
            ) {
              this.props.flags.ipv6Sharing && sharedRanges.push(range);
```

bnussman (commented):
Functionality is looking solid against my local env but I have some questions and comments.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
```diff
@@ -229,10 +244,69 @@ class LinodeNetworking extends React.Component<CombinedProps, State> {
...
+          if (range.region !== this.props.linode.region) {
+            return acc;
+          }
+
+          await getIPv6RangeInfo(range.range).then(async (resp) => {
+            let slaac;
+            if (!resp.is_bgp) {
+              await getLinodeIPs(this.props.linode.id).then((ips) => {
+                slaac = ips?.ipv6?.slaac.address;
+              });
```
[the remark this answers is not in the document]
bcmertz: good point! we could do it that way, however if the refreshIPv4 promise doesn't resolve first we won't have the slaac address in state. It appears that promises resolve in order but underneath they can resolve out of order so I'm hesitant to make this change

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -103,22 +86,42 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+    const ipChoices = getIPChoicesAndLabels(
+      linodeID,
+      linodes,
+      flags.ipv6Sharing,
+      availableRangesMap,
+      linodeSharedIPs
+    );
+    setipChoices(ipChoices);
+    isLoading = false;
+  }, [linodeID, data, availableRanges]);
```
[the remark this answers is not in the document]
bcmertz: Hmm afaict we need the useEffect to update ipChoices when we get the data back from our request on line 96. However, to your point we could make ipChoices a var that the rest of the code could access to avoid keeping it in state.

However, I don't know how worthwhile it is only making that one suggested change and introducing a global ipChoices variable instead of ipChoices state

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
bcmertz: It's a little strange (took me a long time to get), but this is the intended behavior. Showing it allows us to unshare that Linode from that range.

We want ranges not shared to a Linode in availableRanges, and ranges that are shared to a Linode in shared ranges

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -103,22 +87,82 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+          availableRangesMap[currentValue.id].forEach((range: string) => {
+            previousValue[range] = currentValue.label;
+            if (ipToLinodeID.hasOwnProperty(range)) {
+              updateipToLinodeID({
+                [range]: [ipToLinodeID[range]].concat([currentValue.id]),
+              });
+            } else {
+              updateipToLinodeID({ [range]: [currentValue.id] });
+            }
+          });
```
codyfinn, line 130:
```suggestion
       if (flags.ipv6Sharing) {
        availableRangesMap?.[currentValue.id].forEach((range: string) => {
          previousValue[range] = currentValue.label;
          updateipToLinodeID({
            [range]: [...(ipToLinodeID?.[range] ?? []), currentValue.id],
          });
        });
      }

      return previousValue;
    }, {});
```

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -103,22 +88,77 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+  const resp = useAllLinodesQuery(
     { page_size: API_MAX_PAGE_SIZE },
     {
       region: linodeRegion,
     },
     open // Only run the query if the modal is open
   );
 
+  let isLoading = resp.isLoading;
+  const data = resp.data;
```
HanaXu, line 107: This can be simplified to

```suggestion
  { data: linodesData, isLoading } = useAllLinodesQuery(
    { page_size: API_MAX_PAGE_SIZE },
    {
      region: linodeRegion,
    },
    open // Only run the query if the modal is open
  );
```

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -103,22 +88,77 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+  React.useEffect(() => {
+    // don't try anything until we've finished the request for the Linodes data
+    if (isLoading) {
+      return;
+    }
+    isLoading = true;
+    const ipChoices = getIPChoicesAndLabels(linodeID, linodes);
+    setipChoices(ipChoices);
+    isLoading = false;
+  }, [linodeID, data, availableRanges]);
```
HanaXu, line 161: We can remove the `linodes` variable and React Query handles the loading states for us so this can be simplified to

```suggestion
  React.useEffect(() => {
    // don't try anything until we've finished the request for the Linodes data
    if (!linodesData) {
      return;
    }
    const ipChoices = getIPChoicesAndLabels(linodeID, linodesData);
    setipChoices(ipChoices);
  }, [linodeID, linodesData, availableRanges]);
```

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
HanaXu: ^ oops, forgot the `const`

bcmertz: just remembered I didn't do this because isLoading can't be a const

bcmertz: unless theres a way around that

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -154,36 +193,110 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
-
+    const groupedUnsharedRanges = {};
+    const finalIPs: string[] = uniq(
+      ipsToShare.reduce((previousValue, currentValue) => {
+        if (currentValue === undefined || currentValue === null) {
+          return previousValue;
+        }
+        const strippedIP: string = currentValue.includes('/')
+          ? currentValue.substr(0, currentValue.indexOf('/'))
+          : currentValue;
```
codyfinn, line 205:
```suggestion
        const strippedIP: string = currentValue.split('/')[0];
```

I think this is equivalent.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -154,36 +193,110 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+
+        // Filter out v4s and shared v6 ranges as only v6s and unshared ips will be added
+        const isStaticv6 =
+          ipToLinodeID.hasOwnProperty(currentValue) &&
+          ipToLinodeID[currentValue].length === 1;
```
codyfinn, line 210:
```suggestion
        const isStaticv6 = ipToLinodeID?.[currentValue]?.length === 1;
```

I think this is equivalent.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
bcmertz: nice

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -154,36 +187,106 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+      }, [])
+    );
+
+    // use local variable and state because useState won't update state right away
+    // and useEffect won't play nicely here
+    let localErrors: APIError[] | undefined = undefined;
     setErrors(undefined);
-    setSubmitting(true);
+    let localSubmitting = true;
+    setSubmitting(localSubmitting);
```
HanaXu, line 230: Why do we need to use an additional local variable? Isn't that what useState is essentially?

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -154,36 +187,106 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+      (arg0: { linode_id: number; ips: string[] }): Promise<any>;
+      (payload: IPSharingPayload): Promise<{}>;
+      (payload: IPSharingPayload): Promise<{}>;
+    };
+    const promises: Promise<void | {}>[] = [];
+    if (flags.ipv6Sharing) {
+      share = shareAddresses;
+
+      // share unshared ranges first to their staticly routed Linode, then later we can share to the current Linode
+      ipsToShare.forEach((ip) => {
```
HanaXu, line 243: Do we need this outer `forEach` if `ip` is never referenced?

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
bcmertz: unfortunately we do unless there's a workaround I can't find, useState is asynchronous (sorta), and can't be depended on to update state synchronously. Useeffect could be a work around but can't be used here. here's a stackoverflow with more info

https://stackoverflow.com/a/54069332

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -291,6 +407,27 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+  } = availableRanges.reduce((previousValue, currentValue) => {
+    if (previousValue.hasOwnProperty(currentValue.linodes[0])) {
+      previousValue[currentValue.linodes[0]].push(
+        `${currentValue.range}/${currentValue.prefix}`
+      );
+    } else {
+      previousValue[currentValue.linodes[0]] = [
+        `${currentValue.range}/${currentValue.prefix}`,
+      ];
+    }
```
codyfinn, line 424:
```suggestion
    previousValue[currentValue.linodes[0]] = [...(previousValue?.[currentValue.linodes[0]]) ?? [], `${currentValue.range}/${currentValue.prefix}`];
```

codyfinn: I think that is equivalent

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -213,6 +313,20 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
           <Grid container>
             <Grid item sm={12} lg={8} xl={6}>
+              {flags.ipv6Sharing && (
+                <Notice warning>
+                  <Typography style={{ fontSize: '0.875rem' }}>
+                    <strong>Warning:</strong> Converting a statically routed
+                    IPv6 range to a shared range will break existing IPv6
+                    connectivity unless each Linode that shares the range has
+                    BGP setup to advertise that range. Follow{' '}
+                    <Link to="https://deploy-preview-5342--nostalgic-ptolemy-b01ab8.netlify.app/docs/guides/ip-failover/">
```
DevDW, line 323: Can we use the finalized link here?

bcmertz: definitely once it's merged :)

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -213,6 +313,20 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+                      this guide
+                    </Link>{' '}
+                    to setup BGP on a Linode.
```
DevDW, line 326: `set up` (since it's a verb here)

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
```diff
@@ -229,10 +244,69 @@ class LinodeNetworking extends React.Component<CombinedProps, State> {
...
+          errorResponse,
+          'There was an error retrieving your IPv6 network information.'
+        );
+        this.setState({
+          ipv6Error: errors[0].reason,
+          ipv6Loading: false,
+        });
+      });
+
+    return [refreshIPv4, refreshIPv6];
```
codyfinn, line 305: Ideally I would like to see the combination of async/await and promise chains consolidated to one or the other. Also the nested promise chains are another thing I would want to see addressed.

bcmertz: I'll break this cleanup task into another ticket

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
bcmertz: I think this should instead be:
```
previousValue[currentValue.linodes[0]] = [...(previousValue?.[currentValue.linodes[0]] ?? []), `${currentValue.range}/${currentValue.prefix}`];
```

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -103,22 +88,70 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+  const { data: linodes, isLoading } = useAllLinodesQuery(
     { page_size: API_MAX_PAGE_SIZE },
     {
       region: linodeRegion,
     },
     open // Only run the query if the modal is open
   );
 
-  const linodes = data ?? [];
+  const getIPChoicesAndLabels = (linodeID: number, linodes: Linode[]) => {
```
DevDW, line 106: For a follow-up: I think this function can be moved back outside the FC with the flag boolean added as another parameter. Ideally we'd be able to add a unit test for it to ensure it is working as intended.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -103,22 +88,70 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+      }
+    });
+
+    return choiceLabels;
+  };
+
+  const [ipChoices, setipChoices] = React.useState({});
+  const [ipToLinodeID, setipToLinodeID] = React.useState({});
+
+  const updateipToLinodeID = (newData: Record<string, number[]>) => {
```
DevDW, line 141: `updateIPToLinodeID`

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -103,22 +88,70 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+    linodeSharedIPs.forEach((range) => {
+      if (!choiceLabels.hasOwnProperty(range)) {
+        choiceLabels[range] = '';
+      }
+    });
+
+    return choiceLabels;
+  };
+
+  const [ipChoices, setipChoices] = React.useState({});
```
DevDW, line 138: `setIPChoices`

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -103,22 +88,70 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+  const [ipToLinodeID, setipToLinodeID] = React.useState({});
```
DevDW, line 139: `setIPToLinodeID`

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
bcmertz: trying to see if we can get it ahead of time before its released

bcmertz: done

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
bcmertz: we'd also have to pass availableRangesMap, updateIPToLinodeID, ipToLinodeID, and linodeSharedIPs. Still move?

DevDW: might still be worth doing for the sake of the test, but otherwise I'd say no

bcmertz: follow up PR

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
```diff
@@ -229,10 +244,69 @@ class LinodeNetworking extends React.Component<CombinedProps, State> {
           initialLoading: false,
         });
       });
+
+    const refreshIPv6 = getAllIPv6Ranges({})
```
DevDW, line 248: I don't think the `{}` is needed

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
```diff
@@ -229,10 +244,69 @@ class LinodeNetworking extends React.Component<CombinedProps, State> {
...
+        const availableRanges: IPRangeInformation[] = [];
+        const rangeConstruction = await ranges.reduce(async (acc, range) => {
+          await acc;
+          // filter user ranges outside dc
+          if (range.region !== this.props.linode.region) {
+            return acc;
+          }
+
+          await getIPv6RangeInfo(range.range).then(async (resp) => {
+            let slaac;
```
DevDW, line 263: an explanatory comment here could be helpful

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworkingActionMenu.tsx
```diff
@@ -45,7 +45,10 @@ export const LinodeNetworkingActionMenu: React.FC<CombinedProps> = (props) => {
   const deletableIPTypes = ['IPv4 – Public', 'IPv6 – Range'];
 
   const actions = [
-    onRemove && ipAddress && deletableIPTypes.includes(ipType)
+    onRemove &&
+    ipAddress &&
+    ipAddress.prefix !== 116 &&
```
DevDW, line 50: maybe we could put the `ipAddress.prefix !== 116` piece in a variable along with an explanatory comment to add a bit of clarity about the logic behind that in the context of this file

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
```diff
@@ -229,10 +244,69 @@ class LinodeNetworking extends React.Component<CombinedProps, State> {
...
+            if (resp.is_bgp && resp.linodes.includes(this.props.linode.id)) {
+              // any range that is shared to this linode
+              this.props.flags.ipv6Sharing && sharedRanges.push(range);
+            } else if (!resp.is_bgp && range.route_target === slaac) {
+              // any range that is statically routed to this linode
+              staticRanges.push(range);
+            } else {
+              // any range that is not shared to this linode or static on this linode
+              this.props.flags.ipv6Sharing && availableRanges.push(resp);
+            }
```
HanaXu, line 279: Can we move the flag checking to be the first thing checked for short circuiting?

```suggestion
            if (
              this.props.flags.ipv6Sharing &&
              resp.is_bgp &&
              resp.linodes.includes(this.props.linode.id)
            ) {
              // any range that is shared to this linode
              sharedRanges.push(range);
            } else if (!resp.is_bgp && range.route_target === slaac) {
              // any range that is statically routed to this linode
              staticRanges.push(range);
            } else if (this.props.flags.ipv6Sharing) {
              // any range that is not shared to this linode or static on this linode
              availableRanges.push(resp);
            }
```

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
```diff
@@ -419,11 +493,18 @@ class LinodeNetworking extends React.Component<CombinedProps, State> {
...
+    if (IPRequestError) {
+      errorText =
+        IPRequestError ||
+        'There was an error retrieving your networking information.';
+    } else if (ipv6Error) {
+      errorText =
+        ipv6Error ||
+        'There was an error retrieving your networking information.';
+    }
+    return errorText ? <ErrorState errorText={errorText} /> : null;
```
HanaXu, line 507: This function is only called when there is an error so we can simplify this to
```suggestion
    const errorText =
      IPRequestError ||
      ipv6Error ||
      'There was an error retrieving your networking information.';
    return <ErrorState errorText={errorText} />;
```

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -154,36 +187,104 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+          })
         );
-
-        setErrors(errors);
-        setSubmitting(false);
-        setSuccessMessage(undefined);
       });
+    } else {
+      share = shareAddressesv4;
+    }
```
HanaXu, line 263: We can define the share value with a ternary
```typescript
    const share = flags.ipv6Sharing ? shareAddresses : shareAddressesv4;
    const promises: Promise<void | {}>[] = [];

    if (flags.ipv6Sharing) {
      // share unshared ranges first to their staticly routed Linode, then later we can share to the current Linode
      Object.keys(groupedUnsharedRanges).forEach((linode_id) => {
        promises.push(
          shareAddresses({
            linode_id: parseInt(linode_id, 10),
            ips: groupedUnsharedRanges[linode_id],
          }).catch((errorResponse) => {
            const errors = getAPIErrorOrDefault(
              errorResponse,
              'Unable to complete request at this time.'
            );
            localErrors = errors;
            setErrors(errors);
            localSubmitting = false;
            setSubmitting(false);
            setSuccessMessage(undefined);
          })
        );
      });
    }
```

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
```diff
@@ -229,10 +244,74 @@ class LinodeNetworking extends React.Component<CombinedProps, State> {
...
+              // any range that is shared to this linode
+              sharedRanges.push(range);
+            } else if (!resp.is_bgp && range.route_target === slaac) {
+              // any range that is statically routed to this linode
+              staticRanges.push(range);
+            } else if (this.props.flags.ipv6Sharing) {
+              // any range that is not shared to this linode or static on this linode
+              availableRanges.push(resp);
+            }
+          });
```
HanaXu, line 285: Let's unnest this a bit
```suggestion
          // get info on an IPv6 range; if its shared check if its shared to our Linode;
          // if its not shared (!is_bgp) figure out if its a slaac address or a statically routed range
          const resp = await getIPv6RangeInfo(range.range);
          let slaac;
          if (!resp.is_bgp) {
            await getLinodeIPs(this.props.linode.id).then((ips) => {
              slaac = ips?.ipv6?.slaac.address;
            });
          }

          if (
            this.props.flags.ipv6Sharing &&
            resp.is_bgp &&
            resp.linodes.includes(this.props.linode.id)
          ) {
            // any range that is shared to this linode
            sharedRanges.push(range);
          } else if (!resp.is_bgp && range.route_target === slaac) {
            // any range that is statically routed to this linode
            staticRanges.push(range);
          } else if (this.props.flags.ipv6Sharing) {
            // any range that is not shared to this linode or static on this linode
            availableRanges.push(resp);
          }
```

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworkingActionMenu.tsx
```diff
@@ -44,8 +44,11 @@ export const LinodeNetworkingActionMenu: React.FC<CombinedProps> = (props) => {
 
   const deletableIPTypes = ['IPv4 – Public', 'IPv6 – Range'];
 
+  // if we have a 116 we don't want to give the option to remove it
+  const is116 = ipAddress?.prefix === 116;
```
HanaXu, line 48: Can we rename this variable to be a bit more specific?

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -154,36 +186,97 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
...
+        const isStaticv6 = ipToLinodeID?.[currentValue]?.length === 1;
+        // For any IP in finalIPs that isn't shared (length of linode_ids === 1)
+        // make note in groupedUnsharedRanges so that we can first share that IP to
+        // the Linode it is statically routed to, then to the current Linode
+        if (isStaticv6) {
+          const linode_id = ipToLinodeID[currentValue][0];
+          if (groupedUnsharedRanges.hasOwnProperty(linode_id)) {
+            groupedUnsharedRanges[linode_id] = [
+              groupedUnsharedRanges[linode_id],
+            ].concat([strippedIP]);
```
DevDW, line 214: can we do this instead of the array concatenation?

```suggestion
            groupedUnsharedRanges[linode_id] = [...groupedUnsharedRanges[linode_id], [strippedIP]];
```

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
DevDW: I would like us to revisit this in the follow-up PR and see if there's a better way for us to achieve what we're looking for here

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -213,6 +305,20 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
           )}
           <Grid container>
             <Grid item sm={12} lg={8} xl={6}>
+              {flags.ipv6Sharing && (
```
DevDW, line 308: let's go with a ternary here instead of the `&&`

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
bcmertz: good suggestion, this was actually a bug if there were more than 1 unshared range being shared

HanaXu (approved):

PR #8091, opened 2021-12-21
M3-5632: Fixed Customer Unable to Edit Cloud Firewall Rules

bnussman:
## 🐛 Bug Description

This PR fixes some customers unable to edit Firewall rules

### What would happen
- User was able to click `Save Changes` but no change occurred when editing a firewall rule.
- An API validation error occurred but was not surfaced to the frontend (only viewable in devtools)

### The cause
- This was caused by what we suppose to be a very old firewall rule that had **many** rules with an empty IPv6 array that the API no longer accepts when you make a PUT

### The Fix
- Ensure we don't pass any empty IPv4 or IPv6 arrays in the PUT request

##  🧪 How to test

- I was unable to reproduce on my personal account. Refer to the internal ticket to see which Firewall and Account had this issue
- Test editing a firewall rule in various ways
- Test creating a firewall rule

### DIFF
--- a/packages/manager/src/features/Firewalls/FirewallDetail/Rules/firewallRuleEditor.ts
+++ b/packages/manager/src/features/Firewalls/FirewallDetail/Rules/firewallRuleEditor.ts
@@ -251,7 +251,7 @@ export const removeICMPPort = (
     return thisRule;
   });
 
-const removeEmptyAdderessArrays = (rules: ExtendedFirewallRule[]) => {
+const removeEmptyAddressArrays = (rules: ExtendedFirewallRule[]) => {
   return rules.map((rule) => {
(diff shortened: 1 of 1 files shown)


packages/manager/src/features/Firewalls/FirewallDetail/Rules/firewallRuleEditor.ts
```diff
@@ -251,12 +251,28 @@ export const removeICMPPort = (
     return thisRule;
   });
 
+const removeEmptyAdderessArrays = (rules: ExtendedFirewallRule[]) => {
```
HanaXu, line 254: Address

HanaXu (approved):

DevDW (approved):

PR #8092, opened 2021-12-21
Update URL after redirecting from a URL that doesn't exist

tiffwong:
## Description

If a user tries to navigate to a URL that doesn't exist which uses the `SafeTabPanel`, the expected behavior is to redirect them to the content on the default tab. This fixes the issue of the URL not being updated when this happens for DBaaS.

Will release another PR that fixes this for other pages.

## How to test

- Go to `/databases/[id]`
- Navigate to a different tab
- Change the URL to something else (ie. `/databases/[id]/test`)
  - The URL should now be updated to `/databases/[id]`


bcmertz (approved):
LGTM and works as expected!

bnussman (approved):

HanaXu (approved):

PR #8093, opened 2021-12-21
M3-5638: Fix Editing a Secondary Domain Freezing Cloud Manager

bnussman:
## 🐛 Bug Description

When searching for a secondary domain and clicking on it, Cloud Manager would freeze and call the useEffect many many times in an infinite loop

https://user-images.githubusercontent.com/6440455/146985697-0613b6f5-00fc-47b3-be43-66832c9801ff.mov

The fix was removing a redundant `useEffect` that called same function as the other `useEffect`

I also fixed the button order in the `Import a Zone` drawer

## How to test

- Test editing a primary and secondary domain
- Test editing from search and from the landing table
- Make sure that what happens in the video above does not happen anymore

### DIFF
--- a/packages/manager/src/features/Domains/DomainsLanding.tsx
+++ b/packages/manager/src/features/Domains/DomainsLanding.tsx
@@ -178,16 +178,6 @@ export const DomainsLanding: React.FC<CombinedProps> = (props) => {
     openForEditing,
   ]);
 
-  React.useEffect(() => {
-    // Open the "Edit Domain" drawer if so specified by this component's props.
-    if (domainForEditing) {
-      props.openForEditing(
-        domainForEditing.domainLabel,
-        domainForEditing.domainId
-      );
-    }
-  }, [domainForEditing, props]);
-
   const navigateToCreate = () => {
     props.history.push(DOMAIN_CREATE_ROUTE);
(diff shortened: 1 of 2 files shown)


packages/manager/src/features/Domains/DomainsLanding.tsx
```diff
@@ -178,16 +178,6 @@ export const DomainsLanding: React.FC<CombinedProps> = (props) => {
...
-  React.useEffect(() => {
-    // Open the "Edit Domain" drawer if so specified by this component's props.
-    if (domainForEditing) {
-      props.openForEditing(
-        domainForEditing.domainLabel,
-        domainForEditing.domainId
-      );
-    }
-  }, [domainForEditing, props]);
-
```
DevDW, line 190: I was curious about how this ended up being duplicated in this file; I thought it may have been from a rebase or merge commit and it slipped through the cracks, but I found that this file was refactored from a class component to a function component earlier this year and some logic was lost in the process of turning `componentDidUpdate` into a `useEffect`

DevDW (approved):
LGTM, the freezing bug is no longer happening 🚀

HanaXu (approved):

PR #8094, opened 2021-12-21
DBAAS1-106: Database Landing Table

bnussman:
## Preview
Screen Shot 2021-12-28 at 12 36 40 PM

## Description

- Makes DBaaS landing table according to spec

## How to test

- Enable the MSW
- Go to `/databases` and make sure the landing table matches the proposed spec in the internal ticket description
- Check responsiveness


bcmertz:
The "Label" sort direction looks like it should match the styling of "Created" sort arrows according to the designs

bcmertz:
Nice job! Two small things:

- "Label" should be "Cluster Label"
- "Created" should have the same sort arrow as "Label"

according to DBaaS MVP designs in "Cloud Manager Final" project (there are outdated designs posted elsewhere)

bcmertz (changes requested):
Nice job! Two small things:

- "Label" should be "Cluster Label"
- "Created" should have the same sort arrow as "Label"

According to DBaaS MVP designs in "Cloud Manager Final" project (there are outdated designs posted elsewhere)

bnussman:
I changed `Label` to `Cluster Label` but I'm unsure of any action to do about the sort arrows. It is standard that you can only sort by one column at once. The default sort I chose is by `label` `descencing`.

bcmertz:
Good point, I suppose its a little error in the designs, thanks!

packages/manager/src/features/Databases/DatabaseLanding/DatabaseRow.tsx
```diff
@@ -1,24 +1,107 @@
...
+  creating: 'other',
+  running: 'active',
+  failed: 'error',
+  degraded: 'inactive',
+  updating: 'other',
+};
+
+export const databaseEngineMap: Record<Engine, string> = {
+  mysql: 'MySQL',
+};
```
bcmertz, line 31: The nimbus DBaaS PR describes the following engines:
```class DatabaseEngineType(Enum):
    mysql = "mysql"
    postrgresql = "postgresql"
    mongodb = "mongodb"
    redis = "redis"```

It might be worth adding support here and to update `Engine` type in `api-v4/lib/databases/types.d.ts` to reflect we expect one of those strings, not just mysql

**Review by bcmertz — CHANGES_REQUESTED** [review]

File: packages/manager/src/features/Databases/DatabaseLanding/DatabaseRow.tsx (hunk 1 above)
Comment (bnussman) [reply packages/manager/src/features/Databases/DatabaseLanding/DatabaseRow.tsx:31]: I think we may want to hold off since MySQL will be the only option for now. Any thoughts @DevDW @johnwcallahan ?

Comment (johnwcallahan) [reply packages/manager/src/features/Databases/DatabaseLanding/DatabaseRow.tsx:31]: I don't see any harm in going ahead and adding them to the map. One less thing to update once we support more engine types.

**Review by bcmertz — APPROVED** [review]
🎉

File: packages/manager/src/features/Databases/DatabaseLanding/DatabaseRow.tsx [hunk 2]
```diff
@@ -1,24 +1,110 @@
...
+        </div>
+      </TableCell>
+      <Hidden xsDown>
+        <TableCell>{configuration}</TableCell>
+      </Hidden>
+      <TableCell>
+        {`${databaseEngineMap[engine]} v${getDatabaseVersionNumber(version)}`}
+      </TableCell>
+      <Hidden smDown>
+        <TableCell>{dcDisplayNames[region]}</TableCell>
```
DevDW, line 99: in case the region isn't in the array, we should probably have `region` or possibly `"Unknown Region"` as a backup (we do this elsewhere, `KubeSummaryPanel.tsx` for example)

DevDW (approved):
LGTM, left a couple of minor comments 👍

packages/manager/src/features/Databases/DatabaseLanding/DatabaseRow.tsx
```diff
@@ -1,24 +1,110 @@
...
+  } = database;
+
+  const configuration =
+    failover_count === 0 ? (
+      'Primary'
+    ) : (
+      <>
+        {`Primary +${failover_count}`}
+        <Chip
+          className={`${chipClasses.chip} ${chipClasses.nvmeChip}`}
```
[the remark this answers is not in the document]
bnussman: To keep scope focused, can I make these changes in a follow up ticket?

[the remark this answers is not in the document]
DevDW: definitely, this isn't blocking by any means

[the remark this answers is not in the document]
bnussman: Thank you!

PR #8097, opened 2021-12-22
Small types & factories fixes/updates

DevDW:
## Description
Fix TypeScript error from `id` field in `databaseBackupFactory` and update `databaseType` interface and `databaseTypeFactory` per API specs

### DIFF
--- a/packages/api-v4/src/databases/types.ts
+++ b/packages/api-v4/src/databases/types.ts
@@ -55,7 +55,8 @@ export interface SSLFields {
   certificate: string | null;
 }
 
-export interface Database {
+// DatabaseInstance is the interface for the shape of data returned by the /databases/instances endpoint.
+export interface DatabaseInstance {
   id: number;
   label: string;
   engine: Engine;
@@ -63,6 +64,7 @@ export interface Database {
   region: string;
   version: string;
   status: DatabaseStatus;
+  failover_count: number;
   updated: string;
   created: string;
   instance_uri: string;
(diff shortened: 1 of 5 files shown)


bnussman (changes requested):
I few things I found we should change: 
- The `DatabaseFactory` should now return `failover_count`
- `getEngineDatabase` should return `CreateDatabaseResponse` type instead of `Database`
- It would be nice to have two factories for database instances since `getEngineDatabase` and `getDatabases` return different schemes for database instances

packages/api-v4/src/databases/types.ts
```diff
@@ -87,17 +89,18 @@ interface ConnectionStrings {
...
   label: string;
   region: string;
   status: DatabaseStatus;
-  type: string;
-  standby_count: StandbyCount;
+  type: DatabaseType;
+  failover_count: FailoverCount;
   engine: Engine;
   encrypted: boolean;
   ipv4_public: string[];
```
HanaXu, line 103: is `ipv4_public` supposed to be a string or string array?

bnussman: `string` I believe

DevDW: looks like a string -- just fixed it, thanks for catching

bnussman (approved):
Changes looks good. I like the decision to use `DatabaseInstance` and `Database` 🚀

HanaXu (approved):
🚀

bcmertz (approved):
LGTM!

PR #8099, opened 2021-12-23
DBaaS1-84: Access Control in Database Details

DevDW:
## Description
Adds the Access Control section to the Summary tab of the Database Detail page.

## Note to Reviewers
I'll go ahead and open this up to review to solicit more feedback and testing to make sure things are functioning smoothly. Still in the midst of adding some tests (if you have anything specific you think should be tested, feel free to suggest it), doing some cleanup, and improving styling, but the major pieces are all in place.

To test, go to a Database detail page. On the Summary tab, at the bottom, you should see an Access Controls section. Ensure that removing IPs via the table works, as well as adding and removing IPs via the drawer.

## To-Do:
- [x] Add "Remove" button per mockups
- [x] Full functionality
- [x] Improve empty state
- [x] Improve styling (fix "Add Access Controls" button)
- [x] Code cleanup
- [x] Add some tests


packages/manager/src/features/Databases/DatabaseDetail/DatabaseSummary/AccessControls.tsx
```diff
@@ -0,0 +1,75 @@
...
+        />
+      );
+    }
+
+    return (
+      <Table className={classes.table}>
+        <TableBody>
+          {accessControlsList.map((accessControl, idx) => (
+            <TableRow key={idx}>
+              <TableCell key={idx} className={classes.cell}>
```
codyfinn, line 51: I think we generally want to avoid using the index in an array as the key.

packages/manager/src/queries/databases.ts
```diff
@@ -46,7 +46,7 @@ export const useDatabaseMutation = (engine: Engine, id: number) =>
     {
       onSuccess: (data) => {
         queryClient.setQueryData<Database | undefined>(
-          `${queryKey}-${id}`,
+          [queryKey, Number(id)],
```
bnussman, line 49: Thanks :)

packages/manager/src/features/Databases/DatabaseDetail/AddAccessControlDrawer.tsx
```diff
@@ -42,33 +44,42 @@ const AddAccessControlDrawer: React.FC<CombinedProps> = (props) => {
...
     // readOnly,
   } = props;
 
+  // console.log(allowList);
+
   const classes = useStyles();
 
   const [error, setError] = React.useState<string | undefined>('');
-  const [loading, setLoading] = React.useState<boolean>(false);
+  const [submitting, setSubmitting] = React.useState<boolean>(false);
```
codyfinn, line 59: When you provide `useState` with a default value it should infer the typing. In the case above for the `error` variable it is needed since undefined is being unioned with the string type. Unless you want to keep the typing for consistency.

packages/manager/src/features/Databases/DatabaseDetail/AddAccessControlDrawer.tsx
```diff
@@ -42,33 +44,42 @@ const AddAccessControlDrawer: React.FC<CombinedProps> = (props) => {
...
   }, [open]);
 
   const handleAllowInboundSourcesClick = () => {
-    setLoading(true);
+    setSubmitting(true);
+
+    // Get the IP address strings out of the objects and filter empty strings out.
+    const allowListRetracted = allowList
+      .map((extendedIP) => extendedIP.address)
+      .filter((ip) => ip !== '');
```
codyfinn, line 74: I would expect the number of allowed IP addresses to be pretty short and if it is then the `.map.filter` is probably fine. Otherwise I would suggest `.reduce` to avoid looping through the array twice.

packages/manager/src/features/Databases/DatabaseDetail/AddAccessControlDrawer.tsx
```diff
@@ -42,33 +44,42 @@ const AddAccessControlDrawer: React.FC<CombinedProps> = (props) => {
...
-        setLoading(false);
+        setSubmitting(false);
         onClose();
       })
       .catch((e: APIError[]) => {
-        setLoading(false);
+        setSubmitting(false);
         setError(getErrorStringOrDefault(e));
       });
   };
```
codyfinn, line 85: Do we want to continue to use the promise chain pattern or should we start migrating towards async await?

packages/manager/src/features/Databases/DatabaseDetail/DatabaseSummary/AccessControls.tsx
DevDW: Hmm, I did a search in the codebase and saw `key={idx}` quite a bit. Do you think it should be an interpolated string instead (e.g. `access-control-${idx}`)?

packages/manager/src/features/Databases/DatabaseDetail/AddAccessControlDrawer.tsx
DevDW: removed this bit since I ended up using Formik

packages/validation/src/databases.schema.ts
```diff
@@ -1,4 +1,5 @@
 import { array, object, string } from 'yup';
+import { IP_ERROR_MESSAGE, validateIP } from '../lib/firewalls.schema';
```
HanaXu, line 2: For some reason, importing lib schemas inside a schema causes a compile error so we should import from src instead
image

```suggestion
import { IP_ERROR_MESSAGE, validateIP } from './firewalls.schema';
```

HanaXu (commented):
I'm also seeing the "1" issue in this PR 🤔

packages/manager/src/queries/databases.ts
```diff
@@ -47,15 +47,15 @@ export const useDatabaseMutation = (engine: Engine, id: number) =>
     {
       onSuccess: (data) => {
         queryClient.setQueryData<Database | undefined>(
-          [queryKey, id],
+          [queryKey, Number(id)],
```
DevDW, line 50: This may prove to be temporary, but the mutation doesn't work properly without this coercion -- it treats the `id` as a string otherwise. Will look into this further at a later time.

bcmertz (commented):
Remove IP address modal has an overly large font

packages/manager/src/features/Databases/DatabaseDetail/AddAccessControlDrawer.tsx
```diff
@@ -0,0 +1,199 @@
...
+        {error ? <Notice error text={error} /> : null}
+        <Typography variant="body1" className={classes.instructions}>
+          To restrict connections to trusted sources, add at least one inbound
+          source below.{' '}
+          <ExternalLink to="https://www.linode.com/docs">
+            Learn more about securing your cluster.
+          </ExternalLink>
+        </Typography>
+        <form onSubmit={handleSubmit}>
+          <MultipleIPInput
```
bcmertz, line 137: In this component, when an IP is in error, the "X" button becomes misaligned

packages/manager/src/features/Databases/DatabaseDetail/DatabaseSummary/AccessControls.tsx
codyfinn: The bigger concern is that it is supposed to be a unique identifier for that element. So if the underlying array changes it's sorting react can still track changes in a performant manner. React links to this article for why it is a bad practice: https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318

packages/manager/src/features/Databases/DatabaseDetail/DatabaseSummary/AccessControls.tsx
```diff
@@ -0,0 +1,235 @@
...
+  const ipTable = (accessControlsList: string[]) => {
+    if (accessControlsList.length === 0) {
+      return (
+        <Notice warning className={classes.restrictWarning}>
+          <Typography
+            className={classes.restrictWarningText}
+            data-testid="notice-no-access-controls"
+          >
+            Warning: your cluster is open to all incoming connections. Secure
+            this database cluster by restricting access.{' '}
```
bnussman, line 142: I think the Database won't be accessible if there are no access controls specified. Is my understanding correct?

DevDW: That is my current understanding as well. This draft of the warning copy is an earlier version prior to that being clarified; I need to confirm whether there is an updated version yet

packages/manager/src/features/Databases/DatabaseDetail/AddAccessControlDrawer.tsx
DevDW: I'll create a separate ticket for this issue

HanaXu (commented):
Removing an IP via the table works but adding/removing from the drawer isn't working for me 🤔 

https://user-images.githubusercontent.com/14323019/150017127-6b8572fe-404d-4174-8c49-eac1e2a07288.mov

DevDW:
> Removing an IP via the table works but adding/removing from the drawer isn't working for me 🤔
> 
>  Screen.Recording.2022-01-18.at.3.50.18.PM.mov

Odd, I just tried it 🧐 

https://user-images.githubusercontent.com/32860776/150019244-90d37e4e-c366-4d59-ad7a-3a9f0ac6bde9.mov

Did you have the latest version of this branch pulled? What browser were you using?

HanaXu:
@DevDW nvm, I just did a clean `yarn up` and it's working now 😅

HanaXu (changes requested):
The remove button padding seems uneven
image


Can we improve the mobile view? The primary button is taking up a lot of space

image

packages/manager/src/features/Databases/DatabaseDetail/AccessControls.tsx
```diff
@@ -0,0 +1,225 @@
...
+    },
+  },
+  sectionTitle: {
+    marginBottom: '0.25rem',
+  },
+  sectionText: {
+    marginBottom: '1rem',
+    width: '65%',
+    marginRight: 0,
+  },
```
HanaXu, line 42:
```suggestion
  sectionText: {
    marginBottom: '1rem',
    width: '65%',
    marginRight: 0,
    [theme.breakpoints.down('xs')]: {
      width: '100%',
    },
  },
```

packages/manager/src/features/Databases/DatabaseDetail/AccessControls.tsx
```diff
@@ -0,0 +1,225 @@
...
+      marginBottom: '1rem',
+    },
+  },
+  table: {
+    width: '50%',
+    border: `solid 1px ${theme.cmrBorderColors.divider}`,
+    '&:last-child': {
+      borderBottom: 'none',
+    },
+  },
```
HanaXu, line 56:
```suggestion
  table: {
    width: '50%',
    border: `solid 1px ${theme.cmrBorderColors.divider}`,
    '&:last-child': {
      borderBottom: 'none',
    },
    [theme.breakpoints.down('xs')]: {
      width: '100%',
    },
  },
```

HanaXu (changes requested):
Can we add an `xs` breakpoint for phones?

image


image

tiffwong (approved):
Everything looks good! 

Non-blocking but it seems weird to me that the button says "Add Inbound Sources" even if you're removing one. Is it possible to make it dynamic or change it something more generic (ie. "Save Changes")?

HanaXu (approved):

DevDW:
> Everything looks good!
> 
> Non-blocking but it seems weird to me that the button says "Add Inbound Sources" even if you're removing one. Is it possible to make it dynamic or change it something more generic (ie. "Save Changes")?

will raise this point in demo prep tomorrow to see if we want to tweak this

PR #8108, opened 2022-01-07
M3-5649: Add Engine to Database Details URL

bnussman:
## Description

- Adds a databases engine to the Cloud Manager URL
  - **Example:** `https://cloud.linode.com/databases/5` will now be `https://cloud.linode.com/databases/mysql/5`
- The motivation behind this change is that the API requires **both** `id` and `engine` to lookup the Databases's full details 
  - Credit to @codyfinn

## How to test

- Test all DBaaS routing
- Test all existing DBaaS features to ensure this change caused no breaking changes to the work we have done so far


packages/manager/src/features/Databases/DatabaseDetail/index.tsx
```diff
@@ -69,7 +70,7 @@ export const DatabaseDetail: React.FC = () => {
 
     // Redirect to the landing page if the path does not exist
     if (tabChoice < 0) {
-      history.push(`/databases/${databaseId}`);
+      history.push(`/databases/${engine}/${id}`);
```
DevDW, line 73: Technically it should redirect to the Summary tab, right? 

I'm also not sure this is fully working as intended. If you tried to hit `/databases/mysql/2/test` for example, it does strip the non-existent path (`/test`) from the URL, but doesn't redirect to the Summary tab:

Screen Shot 2022-01-10 at 11 09 36 AM

DevDW (approved):
LGTM aside from a concern re: redirects in the case of non-existent paths

packages/manager/src/features/Databases/DatabaseDetail/index.tsx
bnussman: I made a minor change, should be fixed now. @DevDW

HanaXu (approved):

bcmertz (approved):

PR #8111, opened 2022-01-10
M3-5645: State/Province dropdown

tiffwong:
## Description

This PR will make the state/province a dropdown field if the US or Canada is selected as the country. If the previous input does not match any of the options, keep that field as is until the user goes to modify their state/province. Additionally, the country and state/province fields are now required and the form layout has been changed to better match the sign up form. 

**Remaining tasks:**
- [x] Display `Virgin Islands` as `Virgin Islands, U.S.`
- [x] Remove `Micronesia`, `Marshall Islands`, `Palau`, `Armed Forces Americas`, `Armed Forces Europe, Canada, Africa and Middle East`, and `Armed Forces Pacific` from the US states dropdown
- [x] Add `United States Minor Outlying Islands` (short code is `UM`)

## How to test

Switch between countries as you input valid and invalid values for the US/Canada and other countries.

### DIFF
--- a/packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/UpdateContactInformationForm/UpdateContactInformationForm.tsx
+++ b/packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/UpdateContactInformationForm/UpdateContactInformationForm.tsx
@@ -399,20 +399,12 @@ class UpdateContactInformationForm extends React.Component<
                 options={regionResults}
                 placeholder="Select region"
                 required
-                // value={
-                //   regionResults.find(({ value }) =>
-                //     fields.state
-                //       ? value === fields.state
-                //       : value === account.state
-                //   ) || account.state
-                // }
                 value={
-                  fields.state
-                    ? {
-                        label: fields.state,
-                        value: fields.state,
-                      }
-                    : ''
+                  regionResults.find(({ value }) =>
+                    fields.state
+                      ? value === fields.state
+                      : value === account.state
+                  ) ?? ''
                 }
                 textFieldProps={{
                   dataAttrs: {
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/UpdateContactInformationForm/UpdateContactInformationForm.tsx
```diff
@@ -173,24 +188,55 @@ class UpdateContactInformationForm extends React.Component<
...
+        };
+      }
+
+      return {
+        value: region.shortCode,
+        label: region.name,
+      };
+    });
+
+    const excludedUSRegions = [
```
DevDW, line 217: can move this variable outside the class component definition

packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/UpdateContactInformationForm/UpdateContactInformationForm.tsx
```diff
@@ -497,6 +532,7 @@ class UpdateContactInformationForm extends React.Component<
         </Button>
         <Button
           buttonType="primary"
+          disabled={flags.regionDropdown && !this.state.isValid}
```
DevDW, line 535: If the flag is off and you empty out the `State / Province` field, it is still possible to submit the form even though the field has "(required)" next to it

DevDW (commented):
Looks good, just a few questions/comments:

1. By `"If the previous input does not match any of the options, keep that field as is until the user goes to modify their state/province"` in the description, do you mean that, for example, if you have US for the country and a state selected, and then change the country, the state/province should remain as-is until you touch that field? If so, I'm not observing that behavior right now -- changing the country clears the state/province field.

2. `Virgin Islands, U.S.` shows up in the Country dropdown and shows up as a State when the United States is selected for the Country also. Is that duplication intended/desirable?

tiffwong:
@DevDW 

1. For example if the user had the US selected as their country and had "fakestate" as the state previously, the state would still be "fakestate" until they go in to modify their state. If they changed the country though, the state/province field would reset. 
2. It's intended afaik but I'm waiting for that to be confirmed.

DevDW (approved):

HanaXu (approved):

bcmertz (approved):
Awesome job, looks and works great!

PR #8119, opened 2022-01-14
Update Cypress Test for Editing Account Information

bnussman:
## Description

- Updates a Cypress test to match a change made by #8111 
- This test is kind of weird right now because it runs the same test but Cloud Manager has two different region / country selections right now that are toggled with a feature flag. I updated the test so it should pass if the flag is both on or off.

## How to test

- Is the change from `ZIP / Postal Code` to `Postal Code` desired? (I have no problem with it, just asking the question to be 100% sure)
- If the e2e GitHub Actions passed ✅, we should be good!
- If you want to run the test locally, run `yarn cy:debug` and run the test from Cypress's GUI

### DIFF
--- a/packages/manager/cypress/integration/account/billing-contact.spec.ts
+++ b/packages/manager/cypress/integration/account/billing-contact.spec.ts
@@ -126,7 +126,7 @@ describe('Billing Contact', () => {
       .clear()
       .type(newAccountData['phone']);
     getClick('[data-qa-contact-country]').type('France{enter}');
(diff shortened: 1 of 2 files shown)


DevDW (approved):
LGTM 👍

tiffwong (approved):
Great job and thanks for catching this! 💯 

Regarding your question about the change from `ZIP / Postal Code` to `Postal Code`, this is so it matches the label used in the signup process.

PR #8121, opened 2022-01-18
M3-5650 - Remove old DBAAS code

bcmertz:
## Description

Remove old DBAAS code, fix type errors, match API calls

## How to test

Verify no more typescript errors, and that all needed functionality has been preserved

PR #8123, opened 2022-01-19
DBaaS Fixes and Updates

bnussman:
## Description

- Adds `version` to the `Database` type
- Uses the same `queryClient` for all `DatabaseCreate` unit tests
- Makes the Database factory use the same type for every database (fixes the no cluster config issue)

## How to test

- Make sure the type change I made is accurate
- make sure unit tests pass
- Make sure the change to the database factory type makes logical sense and fixes the empty cluster config

### DIFF
--- a/packages/api-v4/src/databases/types.ts
+++ b/packages/api-v4/src/databases/types.ts
@@ -97,7 +97,6 @@ export interface Database {
   region: string;
   status: DatabaseStatus;
   type: string;
-  version: string;
   failover_count: FailoverCount;
   engine: Engine;
   encrypted: boolean;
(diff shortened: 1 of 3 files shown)


HanaXu (approved):

packages/api-v4/src/databases/types.ts
```diff
@@ -97,6 +97,7 @@ export interface Database {
   region: string;
   status: DatabaseStatus;
   type: string;
+  version: string;
```
bcmertz, line 100: This is also fixed in #8121 so might be worth removing this

bcmertz (approved):

PR #8131, opened 2022-01-20
M3-5672: Events and Polling for Databases

bnussman:
## Description

- Adds polling to update the React Query Cache for Databases
- Adds event messages for Database events

## How to test

- Spin up a Database and check the events drawer for a relevant message
- Spin up a Database and ensure it goes from `provisioning` to `active` without refreshing (your going to have to wait a long time for it to provision)
- Create a Database from the API while Cloud Manager is open. Ensure the database you create from the API appears in the Landing table without needed to refresh


packages/manager/src/App.tsx
```diff
@@ -102,6 +103,13 @@ export class App extends React.Component<CombinedProps, State> {
       }
     });
 
+    /*
+     * Send any Database events to the Database events handler in the queries file
+     */
+    events$
+      .filter((event) => event.action.startsWith('database'))
+      .subscribe(databaseEventsHandler);
+
```
codyfinn, line 112: If we add anymore of these types of subscriptions to events we should look at refactoring the pattern. Every time we call `filter` we are traversing the array of events which could be long. Might be a good place for `.reduce` in the future.

packages/manager/src/queries/databases.ts
```diff
@@ -136,6 +135,94 @@ export const useRestoreFromBackupMutation = (
...
+    case 'database_create':
+      switch (status) {
+        case 'failed':
+          updateStoreForDatabase(entity!.id, { status: 'failed' });
+        case 'finished':
+          updateStoreForDatabase(entity!.id, { status: 'active' });
+        case 'notification':
+        // A user just made the initial POST to create a DB, we should use this event to ensure
+        // Cloud Manager has the DB in the store. This allows Cloud Manager to be up to date even if
+        // they have multiple instances of Manager open.
```
codyfinn, line 161: We should keep an eye on this: https://react-query.tanstack.com/plugins/persistQueryClient as it moves forward. Being able to put certain information in localstorage/sessionstorage could be very helpful. In this case if this data was in sessionstorage it would be available between tabs.

bnussman: That's an awesome feature and we definitely could leverage it in some way in the future.

packages/manager/src/queries/databases.ts
```diff
@@ -32,15 +32,17 @@ import { Event } from '@linode/api-v4/lib/account/types';
 export const queryKey = 'databases';
 
 export const useDatabaseQuery = (engine: Engine, id: number) =>
-  useQuery<Database, APIError[]>([queryKey, id], () =>
-    getEngineDatabase(engine, id)
+  useQuery<Database, APIError[]>(
+    [queryKey, id],
+    () => getEngineDatabase(engine, id),
+    { refetchInterval: 20000 }
```
bnussman, line 38: Polling is gross but every 20 seconds should be light on the API and this polling will only happen when the user is on the **Database Landing** table. 

While our **events** polling is powerful, it lacks some information to keep our **store** fully in sync right now so lets poll to ensure users are seeing accurate information. 

For example, we get an event when a database restore **starts**, but we don't get any event update when it **succeeds** or **fails**. Therefore, with events, we can transition the database status from `active` to `restoring`, but we cant use events to transition `restoring` to `active`. Polling will make up for this for now.

packages/manager/src/queries/databases.ts
```diff
@@ -32,15 +32,17 @@ import { Event } from '@linode/api-v4/lib/account/types';
...
   );
 
 export const useDatabasesQuery = (params: any, filter: any) =>
   useQuery<ResourcePage<DatabaseInstance>, APIError[]>(
     [`${queryKey}-list`, params, filter],
     () => getDatabases(params, filter),
-    { keepPreviousData: true }
+    { keepPreviousData: true, refetchInterval: 20000 }
```
bnussman, line 45: Like I said, polling is gross but every 20 seconds should be light on the API and this polling will only happen when the user is on the **Database Details** page.  I know it's terrible, but It should make the user experience a fair bit better in some circumstances.

packages/manager/src/utilities/getEventsActionLink.ts
```diff
@@ -102,6 +102,11 @@ export default (
           return `/nodebalancers/${id}/summary`;
       }
 
+    case 'database':
+      const engine =
+        entity?.url.split('/v4/databases/')[1].split('/')[0] || 'mysql';
```
bnussman, line 107: This is questionable. Any suggestions?

packages/manager/src/queries/databases.ts
HanaXu: Can we add a comment explaining why there's a refetch interval? And if the API going to add these events, a TODO to eventually remove refetchInterval

packages/manager/src/utilities/getEventsActionLink.ts
HanaXu: What about using a regex?

```suggestion
        entity?.url.match(/(mysql|postgresql|mongodb|redis)/i)[0];
```

bnussman: Looks nicer, but requires us to update this regex every time we add a new database engine. May still use it because it will probably be a long time until we add anything other than mysql, postgres, mongo, and redis.

HanaXu: Good point, what about

```typescript
entity?.url.match(/databases\/(\w*)\/instances/i)?.[1]
```

packages/manager/src/queries/databases.ts
codyfinn: Are you sure the API doesn't update the events in place. There might be a `restoring` event that gets it's progress or status updated. Similar to image upload.

packages/manager/src/store/events/event.request.ts
```diff
@@ -28,6 +28,17 @@ export const getEvents: ThunkActionCreator<Promise<Event[]>> = () => (
...
+  // We need to keep polling the event for any database that is still creating.
+  // The same event will change its status from `notification` to `finished`.
+  const databaseEventIds = _events
+    .filter(
+      (event) =>
+        event.action === 'database_create' && event.status === 'notification'
+    )
+    .map((event) => event.id);
+
+  const includeIds = [...inIds, ...databaseEventIds];
```
codyfinn, line 40: Could be simplified as a `.reduce` call to minimize going through the array

packages/manager/src/queries/databases.ts
bnussman: I will double check this, but in my initial findings, a new event was never created **and** the event was never updated in place.

packages/manager/src/store/events/event.request.ts
bnussman: I never write reduces, does this look correct? Can't test against API right now.

```ts
  const databaseEventIds = _events.reduce(
    (acc: number[], event: ExtendedEvent) => {
      if (
        event.action === 'database_create' &&
        event.status === 'notification'
      ) {
        return [...acc, event.id];
      }

      return acc;
    },
    []
  );

  const includeIds = [...inIds, ...databaseEventIds];
```

packages/manager/src/queries/databases.ts
bnussman: Can confirm that status never updates in place and no new success event is created. There is a backend ticket for this already

packages/manager/src/store/events/event.request.ts
codyfinn:
```suggestion
  const databaseEventIds = _events.reduce(
    () => {
      acc.push(...event.id);
      return acc;
    }, [])
  
  const includeIds = [...inIds, ...databaseEventIds];
```

packages/manager/src/utilities/getEventsActionLink.ts
```diff
@@ -102,6 +102,12 @@ export default (
           return `/nodebalancers/${id}/summary`;
       }
 
+    case 'database':
+      const engine = entity?.url.match(
+        /(mysql|postgresql|mongodb|redis)/i
+      )?.[0];
+      return `/databases/${engine}/${id}/summary`;
```
DevDW, line 109: I think we'll need similar code in `getEventsActionLinkStrings.ts` -- right now, the events in the Notifications drawer do not link to the corresponding Database's detail page

DevDW (commented):
Looks good so far overall -- left a couple of comments

packages/manager/src/features/Databases/DatabaseDetail/index.tsx
```diff
@@ -29,6 +29,7 @@ export const DatabaseDetail: React.FC = () => {
   const id = Number(databaseId);
 
   const { data: database, isLoading, error } = useDatabaseQuery(engine, id);
+  const { isLoading: isTypesLoading } = useDatabaseTypesQuery();
```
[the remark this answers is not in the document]
bnussman: Thats odd considering React Query should stop after a while per [their defaults ](https://react-query.tanstack.com/guides/query-retries#_top)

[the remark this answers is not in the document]
bnussman: Maybe has something to do with blocking rather then the API serving an error status code?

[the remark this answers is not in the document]
DevDW: Hmm, that's possible. Let's keep an eye on it in the coming days/weeks to make sure

DevDW (approved):

HanaXu (approved):

PR #8132, opened 2022-01-21
M3 5665 - Create page refinement

bcmertz:
## Description

Addresses Create Database feedback

## Done:
1. At mobile screen size breadcrumb needs left margin
4. Add space between top of helper text and baseline of headers Number of Nodes and Access Controls to 12 px (currently 8 px)
5. Move tooltip about limits of selecting Linode 1 GB plan to inline with the text block
6. Reduce size of pricing text to 12 px (currently 14 px)
7. Add (required) to field label
8. Button to be active by default
9. Remove references to IP Ranges from "Add Access Control" copy
10. Enable SSL mode by default
11. Add nanode to DatabaseTypeClass type
12. Update is1GbPlan variable logic so that selecting the 1 GB nanode plan disables the 3 Node option (logic is currently based on memory but we're getting null back from the api rn)
15. remove 'DBAAS MYSQL - ' from plan labels
16. Make shape of tooltip more horizontal (need to double check with Matthew about break points)

## TODO:
1. Reduce spacing between radio button and plan name to 26 px to match Linode plan table (_This seems to get a little messy because of how SelectPlanPanel dynamically expands to fit available width_)
2. [**Not required**] Change the hover state engine icon color to white. It currently does not change on hover and remains gray (see the gray dolphin in the attached screenshot)

### DIFF
--- a/packages/api-v4/src/databases/types.ts
+++ b/packages/api-v4/src/databases/types.ts
@@ -1,6 +1,6 @@
 import { BaseType } from '../linodes/types';
 
-export type DatabaseTypeClass = 'standard' | 'dedicated';
+export type DatabaseTypeClass = 'standard' | 'dedicated' | 'nanode';
 
 interface DatabasePriceObject {
   monthly: number;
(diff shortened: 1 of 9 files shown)


bnussman:
Now that we surface errors more often with the action button enabled can we make these changes: 

Move the plan error below its heading
Screen Shot 2022-01-20 at 11 26 12 PM


Make the `failover_count` error look better (probably by updating `packages/validation/src/databases.schema.ts` with a custom message)
Screen Shot 2022-01-20 at 11 27 51 PM

bnussman:
We'll also want to fix the failing unit test.

HanaXu (approved):

bnussman (approved):

PR #8135, opened 2022-01-22
Type to Confirm Toggle

bcmertz:
## Description

- Add type to confirm preference and toggle
- Removes outdated local storage queries from fetching preferences
- Add TypeToConfirm component to reduce code duplication
- Use TypeToConfirm inside DeletionDialog
- Add warning notice where applicable


MBartholomew-design:
Confirming that the check box to turn off type to confirm is unchecked by default

packages/manager/src/components/TypeToConfirm/TypeToConfirm.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+const useStyles = makeStyles((theme: Theme) => ({
+  description: {
+    marginTop: theme.spacing(),
+  },
+}));
+
+type CombinedProps = Props;
+
+const TypeToConfirm: React.FC<CombinedProps> = (props) => {
```
bnussman, line 34: I know this is an outstanding pattern we have, but I think its ok to do this

```suggestion
const TypeToConfirm: React.FC<Props> = (props) => {
```

packages/manager/src/components/TypeToConfirm/TypeToConfirm.tsx
```diff
@@ -0,0 +1,106 @@
...
+import FormControlLabel from 'src/components/core/FormControlLabel';
+
+export interface Props {
+  confirmationText?: JSX.Element | string;
+  onChange: (str: string) => void;
+  label: string;
+  hideDisable?: boolean;
+  hideLabel?: boolean;
+  textFieldStyle?: Record<string, any>;
+  typographyStyle?: Record<string, any>;
```
bnussman, line 18: I think this would more accurate

```suggestion
  textFieldStyle?: React.CSSProperties;
  typographyStyle?: React.CSSProperties;
```

bnussman (commented):
Looking solid. Could we have a loading state for the type to confirm checkbox? It feels a off due to the small delay...

Now that I think about it. A checkbox doesn't feel like the right component to use here since it goes away right after you check it...

tiffwong (approved):

HanaXu (approved):

PR #8143, opened 2022-01-25
M3-5676: Database Backups Refinements

bnussman:
## Description

- Added **relative date** column to the backups table
- Updated **copy** to explain that **backups are retained for 7 days**
- When user successfully initializes a restore, we toast a success message, and redirect them to the summary page
  - The user will see the status `Restoring` immediately they start a restore successfully. This may not be immediately updating until #8131 is merged. 
 
## How to test

- Try restoring a database form a backup. Ensure the flow as described above is accurate
- Check copy change for spelling and grammar
- Check that the relative date in the table works and looks good

### DIFF
--- a/packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DatabaseBackups.tsx
+++ b/packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DatabaseBackups.tsx
@@ -68,17 +68,17 @@ export const DatabaseBackups: React.FC = () => {
 
   const renderTableBody = () => {
     if (databaseError) {
-      return <TableRowError message={databaseError[0].reason} colSpan={2} />;
+      return <TableRowError message={databaseError[0].reason} colSpan={3} />;
     }
     if (backupsError) {
-      return <TableRowError message={backupsError[0].reason} colSpan={2} />;
+      return <TableRowError message={backupsError[0].reason} colSpan={3} />;
     }
     if (isDatabaseLoading || isBackupsLoading) {
-      return <TableRowLoading oneLine numberOfColumns={2} colSpan={2} />;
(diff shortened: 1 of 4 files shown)


tiffwong (approved):

HanaXu (changes requested):
It seems a bit odd to me that the relative date is in the middle of the table instead of next to the absolute date. Can we make them closer? Also, when restoring from a backup, the url changes to `/summary` but the view stays on the backups tab

image

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DatabaseBackups.tsx
```diff
@@ -68,17 +68,29 @@ export const DatabaseBackups: React.FC = () => {
...
+          <TableCell>
+            <Skeleton />
+          </TableCell>
+          <TableCell>
+            <Skeleton />
+          </TableCell>
+          <TableCell>
+            <Skeleton style={{ maxWidth: '85%' }} />
+          </TableCell>
+        </TableRow>
```
bnussman, line 88: Okay, I know this is kind of bad. But it is **finally** a loading state that matches the correct column widths. I can't stand when the loading skeletons don't match the Table head column widths

I want to make this a component but I'll want to do that in a follow up ticket. 


## This is what I'm referring to...

**Loading state with incorrect column size**


https://user-images.githubusercontent.com/6440455/151399964-45daaaa6-2363-4c81-a06a-b100c37bd84f.mov



**The better loading state I made with so much less code**

https://user-images.githubusercontent.com/6440455/151399928-d0869fe7-16db-4a75-8410-2ff78fd9551b.mov

DevDW (approved):
LGTM -- I did have to use the mocks to test, though, as the request to `/backups` was failing with an error of `Could not retrieve backups at this time`.

HanaXu (approved):

PR #8145, opened 2022-01-25
Do not mask DB username

johnwcallahan:
## Description

**What does this PR do?**
This uses a constant in lieu of the response from the `/credentials` endpoint, meaning we no longer need to mask the username.

Screen Shot 2022-01-25 at 6 01 16 PM

## How to test

Test success/loading/error scenarios and check the placement of the button and loading spinner.


tiffwong (approved):
This is great!

bnussman (approved):
Functionality is great! 

Only comment would be that the loading spinner looks a little far away from the text, but this isn't a big deal. 
Screen Shot 2022-01-26 at 11 54 10 AM

johnwcallahan:
Thanks @tiffwong and @bnussman! I added the left margin so the loading spinner would be in the same place as the "Show" button, but I can see the argument both ways. For the sake of time I'm going to merge this; if we get additional feedback we can take another look.

PR #8150, opened 2022-01-26
M3-5689: Fix storage unit for DBaaS

tiffwong:
## Description

The API returns the disk in MB so use a util function to convert the value to GB.

## How to test

Navigate to a database cluster that uses the `Nanode 1 GB` plan and the storage should be `25 GB`.


HanaXu (approved):

DevDW (approved):
🚢

PR #8153, opened 2022-01-27
M3-5662 Remove check for E2E label when running tests against PRs

jdamore-linode:
## Description

**What does this PR do?**
Removes the check for the `e2e` label on PRs before running the Cypress end-to-end tests. Workflows already require approval to run for PRs authored by outside contributors, so the thought is that the check for the `e2e` label isn't really necessary from a security perspective.

## How to test

We're limited in our ability to test this before merging. If the code is merged, I'll open two PRs -- one with this account, and one with my personal account -- to make sure that:

* PRs opened by team members have their E2E tests run automatically
* PRs opened by outside contributors require approval before their E2E tests run

I'll revert these changes if we do not observe the above behavior.

## Other notes/thoughts
* :white_check_mark: This change has been cleared by the security team
* :question: Would there be any value in checking for a `skip-e2e` or similar label for PRs where tests aren't necessary? (documentation and changelog updates, etc.)

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -23,14 +23,14 @@ If the continuous integration workflow was triggered via a push to the `master` 
 ## End-to-End Tests
 Automatic end-to-end testing of Linode Cloud Manager via [Cypress](https://www.cypress.io/) is handled by the `Run Tests On PR` and `Run Tests On Push/Schedule` workflows. These two workflows are very similar, and differ primarily in how they are triggered.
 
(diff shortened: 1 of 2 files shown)


bnussman (approved):
Assuming we don't incur any extra costs for more e2e runs, this looks good to me!

jdamore-linode:
@bnussman That's a good point -- I have a meeting with Pat later this afternoon so I'll run this by him and report back here

DevDW (approved):

PR #8154, opened 2022-01-27
Database Create Copy updates

HanaXu:
## Description
M3-5684: Add notice that beta is free
M3-5685: Update Add Access Controls language
M3-5686: Move the provisioning helper text next to create button

## How to test
In `databases/create` check for the changes above and also check mobile view

`yarn test DatabaseCreate`

### DIFF
--- a/packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
+++ b/packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
@@ -6,6 +6,7 @@ import {
   FailoverCount,
   ReplicationType,
 } from '@linode/api-v4/lib/databases/types';
+import { APIError } from '@linode/api-v4/lib/types';
 import { createDatabaseSchema } from '@linode/validation/lib/databases.schema';
 import { useFormik } from 'formik';
 import { groupBy } from 'ramda';
@@ -13,6 +14,7 @@ import * as React from 'react';
 import { useHistory } from 'react-router-dom';
 import MySQLIcon from 'src/assets/icons/mysql.svg';
 import { makeStyles, Theme } from 'src/components/core/styles';
+import Chip from 'src/components/core/Chip';
 import Divider from 'src/components/core/Divider';
 import FormControl from 'src/components/core/FormControl';
 import FormControlLabel from 'src/components/core/FormControlLabel';
@@ -49,8 +51,7 @@ import getSelectedOptionFromGroupedOptions from 'src/utilities/getSelectedOption
 import { handleAPIErrors } from 'src/utilities/formikErrorUtils';
 import { validateIPs } from 'src/utilities/ipUtils';
 import scrollErrorIntoView from 'src/utilities/scrollErrorIntoView';
-import Chip from 'src/components/core/Chip';
-import { APIError } from '@linode/api-v4/lib/types';
+import { ipFieldPlaceholder } from 'src/utilities/ipUtils';
 
 const useStyles = makeStyles((theme: Theme) => ({
   formControlLabel: {
@@ -68,7 +69,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     },
   },
   createBtn: {
-    [theme.breakpoints.down('xs')]: {
+    [theme.breakpoints.down('sm')]: {
       marginRight: theme.spacing(),
     },
   },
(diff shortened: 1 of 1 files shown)


jdamore-linode (approved):

packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
```diff
@@ -538,25 +566,28 @@ const DatabaseCreate: React.FC<{}> = () => {
                 ))
               : null}
             <MultipleIPInput
-              title="Inbound Sources"
-              placeholder="Add IP Address or range"
+              title="Allowed IP Address(s) or Range(s)"
```
DevDW, line 569: `Address(es)`

packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
```diff
@@ -538,25 +566,28 @@ const DatabaseCreate: React.FC<{}> = () => {
...
+              placeholder="Add IP Address or Range"
```
DevDW, line 570: Since these have to be IPv4 addresses/ranges, we should display the example hint as the placeholder like we now do in the Access Controls IP fields.

You can do

`import { ipFieldPlaceholder } from 'src/utilities/ipUtils';`

and pass `ipFieldPlaceholder` here

DevDW (commented):
Looks good; we can clean up the tablet view a bit (>600 px width) though when it comes to the padding so these aren't flush to the page edges:

Screen Shot 2022-01-28 at 3 27 56 PM

Screen Shot 2022-01-28 at 3 28 11 PM

tiffwong (commented):
I seem to be getting an error when I try to check this out including after running `yarn clean`. Is anyone else getting this?

Screen Shot 2022-01-31 at 9 49 40 AM

bnussman:
I'm getting the same issue as @tiffwong. I assume it's because the API response changed for something?

packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
```diff
@@ -59,7 +60,25 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+      alignItems: 'flex-end',
+      marginTop: theme.spacing(),
+    },
+  },
+  createBtn: {
+    [theme.breakpoints.down('sm')]: {
+      marginRight: theme.spacing(),
+    },
+  },
+  createText: {
```
tiffwong, line 76: Can you add `marginLeft` here to prevent the text from being flush to the window when it's around 600-700px?

tiffwong (changes requested):
Left a couple cleanup items for small-medium resolutions

packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
```diff
@@ -515,19 +536,29 @@ const DatabaseCreate: React.FC<{}> = () => {
             </RadioGroup>
             <FormHelperText>{errors.failover_count}</FormHelperText>
           </FormControl>
+          <Grid item xs={12} md={8}>
+            <Notice success className={classes.notice}>
+              <strong>
+                Notice: There is no charge for database clusters during beta.
```
DevDW, line 542: One thought on this copy is that a single node is not technically a cluster, but single nodes are not being charged for as well during beta. Perhaps we can look at using "instances" instead of "clusters" 🤔

packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
```diff
@@ -538,25 +569,28 @@ const DatabaseCreate: React.FC<{}> = () => {
                 ))
               : null}
             <MultipleIPInput
```
DevDW, line 571: I was thinking that it would make sense for us to enforce the IP masking here as well. This can be achieved relatively painlessly with the `onBlur` prop for this component and logic similar to https://github.com/DevDW/manager/blob/167f7f91c10e420fb41d1c271f0d2c08cdd19e46/packages/manager/src/features/Databases/DatabaseDetail/AddAccessControlDrawer.tsx#L55-L59

(a bit of scope creep, so it's fine to create a separate ticket for this or do a follow-up PR if you want)

DevDW (approved):
Looks good 👍 

left a bit of feedback on a potential copy change and something that may be out of scope for these tickets

packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
HanaXu: Good point, I'm going to leave the copy as is for now and we can revisit this if we get any feedback about the copy

tiffwong (approved):

PR #8155, opened 2022-01-28
M3-5694: Changes to types, schema, and Database components to reflect cluster_size

DevDW:
## Description
The API has revised the shapes of some data points:

1. The database types will no longer have `addons` or `price` properties; they are being replaced by `cluster_size` which is an array here consisting of a `quantity` property and a `price` property.
2. `failover_count` is being replaced by `cluster_size` (different from the above), which is a number (either 1 or 3) here. 

## To-Do:
- [x] Update the logic in the Create flow to reflect these changes

## How to test
`yarn test DatabaseCreate`

- Test things mentioned in the description
- The database create page should no longer crash
- All the pricing comes from the API now, check that the prices displayed match the API and check the POST request
    - 3 Node option for Nanodes are no longer disabled

### DIFF
--- a/packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
+++ b/packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
@@ -356,9 +356,12 @@ const DatabaseCreate: React.FC<{}> = () => {
         ?.price,
       multi: type.cluster_size.find((cluster) => cluster.quantity === 3)?.price,
     });
-    setFieldValue('cluster_size', 3);
+    setFieldValue(
+      'cluster_size',
+      values.cluster_size < 1 ? 3 : values.cluster_size
+    );
     setFieldValue('replication_type', 'semi_synch');
-  }, [dbtypes, setFieldValue, values.type]);
+  }, [dbtypes, setFieldValue, values.cluster_size, values.type]);
 
   if (regionsLoading || !regionsData || versionsLoading || typesLoading) {
     return <CircleProgress />;
--- a/packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
+++ b/packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
@@ -21,7 +21,6 @@ import Chip from 'src/components/core/Chip';
 import Divider from 'src/components/core/Divider';
 import FormControl from 'src/components/core/FormControl';
 import FormControlLabel from 'src/components/core/FormControlLabel';
-import FormHelperText from 'src/components/core/FormHelperText';
 import Paper from 'src/components/core/Paper';
 import RadioGroup from 'src/components/core/RadioGroup';
 import { makeStyles, Theme } from 'src/components/core/styles';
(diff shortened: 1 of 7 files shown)


bnussman:
If I remember correctly, before this, the 3 node option was disabled for nanodes (otr some config like that). Is it supposed to still do that?
Screen Shot 2022-01-31 at 1 22 57 PM
?

packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
```diff
@@ -368,18 +351,14 @@ const DatabaseCreate: React.FC<{}> = () => {
...
-      ).toFixed(2),
+    setNodePricing({
+      single: type.cluster_size.find((cluster) => cluster.quantity === 1)
+        ?.price,
+      multi: type.cluster_size.find((cluster) => cluster.quantity === 3)?.price,
     });
-    setFieldValue('failover_count', is1GbPlan ? 0 : 2);
-    setFieldValue('replication_type', is1GbPlan ? 'none' : 'semi_synch');
-  }, [dbtypes, is1GbPlan, setFieldValue, values.type]);
+    setFieldValue('cluster_size', 3);
```
bnussman, line 359: Changing the plan changes the number of nodes, can we make it so the node count does not change if the plan is changed?

https://user-images.githubusercontent.com/6440455/151851952-df19eff3-8e95-49d9-bfff-70f2dbdc1f8f.mov

bnussman (changes requested):
When I try to create a Database, nothing happens. I'm assuming it has something to do with IP validation... I am entering `192.168.1.1` as the IP. 

Also, do we need to use something other than `Inbound Sources` for the input title so it makes more sense to users? 

Screen Shot 2022-01-31 at 1 32 52 PM

HanaXu:
@bnussman 3 Node options for Nanodes are no longer disabled. Re: nothing happening when creating a database, I just tried creating a database cluster with `192.168.1.1` and it sent the request 🤔 . Have you tried rebuilding validation? The Inbound Sources copy and etc will be updated in https://github.com/linode/manager/pull/8154

packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
```diff
@@ -368,18 +351,17 @@ const DatabaseCreate: React.FC<{}> = () => {
...
+      multi: type.cluster_size.find((cluster) => cluster.quantity === 3)?.price,
     });
-    setFieldValue('failover_count', is1GbPlan ? 0 : 2);
-    setFieldValue('replication_type', is1GbPlan ? 'none' : 'semi_synch');
-  }, [dbtypes, is1GbPlan, setFieldValue, values.type]);
+    setFieldValue(
+      'cluster_size',
+      values.cluster_size < 1 ? 3 : values.cluster_size
+    );
+    setFieldValue('replication_type', 'semi_synch');
```
DevDW, line 363: `semi_synch` is the default here because 3 is the default `cluster_size`?

DevDW (commented):
@HanaXu the Database Create piece looks good to me, the only thing we may want to address is the error that pops up for Nodes when you try creating a DB without a plan selected:

Screen Shot 2022-01-31 at 4 00 24 PM

packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
HanaXu: Oops, good catch! 🔍

HanaXu:
@DevDW How did you get to that state? If a plan isn't selected the Node price should be $0 🤔

DevDW:
> @DevDW How did you get to that state? If a plan isn't selected the Node price should be $0 🤔

whoops, I took that screenshot after I chose a plan after first submitting with no plan selected (so the node pricing and whatnot updated).

but to get the "Nodes are required" error, just try to create a DB without a plan selected. We should probably put that error in a `<Notice />` at the top of that section.

bnussman (approved):

HanaXu (approved):

PR #8159, opened 2022-01-31
Display country in the Billing Contact panel

tiffwong:
## Description

Add the users's country to the panel and fix the alignment for Tax ID (M3-5640).

**Before**
Screen Shot 2022-01-31 at 5 20 36 PM

**After**
Screen Shot 2022-01-31 at 5 20 15 PM

## How to test

Go to `/account/billing`

### DIFF
--- a/packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/ContactInformation.tsx
+++ b/packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/ContactInformation.tsx
@@ -200,7 +200,9 @@ const ContactInformation: React.FC<CombinedProps> = (props) => {
                 {city}
                 {city && state && ','} {state} {zip}
               </Typography>
-              <Typography className={classes.section}>{countryName}</Typography>
+              <Typography className={classes.section}>
+                {countryName?.countryName}
+              </Typography>
             </Grid>
           )}
 
--- a/packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/ContactInformation.tsx
+++ b/packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/ContactInformation.tsx
@@ -14,7 +14,6 @@ import Typography from 'src/components/core/Typography';
 import Grid from 'src/components/Grid';
 import styled from 'src/containers/SummaryPanels.styles';
 import BillingContactDrawer from './EditBillingContactDrawer';
-import { Country } from './UpdateContactInformationForm/types';
 
 const useStyles = makeStyles((theme: Theme) => ({
   ...styled(theme),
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/ContactInformation.tsx
```diff
@@ -130,6 +130,14 @@ const ContactInformation: React.FC<CombinedProps> = (props) => {
...
   }, [editContactDrawerOpen, history.location.state]);
 
+  const countryName = countryData.map((_country: Country) => {
+    if (country === _country.countryShortCode) {
+      return _country.countryName;
+    }
+
+    return null;
+  });
+
```
codyfinn, line 140: This will return an array right? Don't we want a single string? In that case would `.find` be a more appropriate here?

codyfinn (dismissed):

DevDW (commented):
The app crashes for me when I try to go to `/account/billing`:

Screen Shot 2022-02-02 at 5 22 52 PM

looks like it's being caused by 

https://github.com/linode/manager/blob/85638f4902da75d7c15b3c074ddb531b8379fddd/packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/ContactInformation.tsx#L203

packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/ContactInformation.tsx
```diff
@@ -130,6 +129,11 @@ const ContactInformation: React.FC<CombinedProps> = (props) => {
     }
   }, [editContactDrawerOpen, history.location.state]);
 
+  // Finding the country from the countryData JSON
+  const countryName = countryData?.find(
+    (_country) => _country.countryShortCode === country ?? null
+  );
+
```
HanaXu, line 136: Variable name is more accurate if we do

```typescript
const countryName = countryData?.find(
    (_country) => _country.countryShortCode === country
)?.countryName;
```

packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/ContactInformation.tsx
```diff
@@ -208,18 +214,24 @@ const ContactInformation: React.FC<CombinedProps> = (props) => {
...
+              </Typography>
             )}
-
             {taxId && (
-              <div className={classes.section}>{'Tax ID ' + taxId}</div>
+              <Typography
+                className={classes.section}
+                style={{ marginTop: 'auto' }}
+              >
+                <span className={classes.bold}>Tax ID</span> {taxId}
```
HanaXu, line 233: We don't need the bold class if we use the strong tag
```suggestion
                <strong>Tax ID</strong> {taxId}
```

HanaXu (approved):

DevDW (approved):

PR #8163, opened 2022-02-01
Observe account.capabilities to enable DBaaS

johnwcallahan:
## Description

**What does this PR do?**
This PR changes the `flags.databases` checks to make use of the `isFeatureEnabled` function, taking into account the user's capabilities. **When we go to beta the LD flag should be off.** (Otherwise DBaaS would be shown to everyone, not just beta participants).

## How to test

To test, check all combos and make sure that DBaaS works (or doesn't work) correctly:

- LD flag on, account capability off **(DBaaS should work)**
- LD flag on, account capability on **(DBaaS should work)**
- LD flag off, account capability off **(DBaaS should NOT work)**
- LD flag off, account capability on **(DBaaS should work – this is the situation for beta)**

### DIFF
--- a/packages/api-v4/src/account/types.ts
+++ b/packages/api-v4/src/account/types.ts
@@ -39,7 +39,8 @@ export type AccountCapability =
   | 'Cloud Firewall'
   | 'Vlans'
   | 'Machine Images'
-  | 'LKE HA Control Planes';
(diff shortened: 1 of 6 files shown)


bnussman (approved):
All 4 cases worked as expected!

codyfinn (approved):

PR #8165, opened 2022-02-02
M3-5701: Disable show button if database cluster is provisioning

HanaXu:
## Description
Disables the show button in a database cluster's connection details if the database is still provisioning since it will just result in an error

## How to test
Go to the summary tab of a database cluster. To see the disabled button state, turn on the MSW

### DIFF
--- a/packages/manager/src/features/Databases/DatabaseDetail/DatabaseSummary/DatabaseSummaryConnectionDetails.tsx
+++ b/packages/manager/src/features/Databases/DatabaseDetail/DatabaseSummary/DatabaseSummaryConnectionDetails.tsx
@@ -144,8 +144,13 @@ export const DatabaseSummaryConnectionDetails: React.FC<Props> = (props) => {
   };
 
   const ssl = database.ssl_connection ? 'ENABLED' : 'DISABLED';
-  const disableShowBtn = database.status === 'provisioning';
+  const disableShowBtn = ['provisioning', 'failed'].includes(database.status);
+  const disableToolTipText =
+    database.status === 'provisioning'
+      ? 'Your Database Cluster is currently provisioning.'
+      : 'You cannot view your root password when your Database Cluster has failed.';
   // const connectionDetailsCopy = `username = ${credentials?.username}\npassword = ${credentials?.password}\nhost = ${database.host}\nport = ${database.port}\ssl = ${ssl}`;
+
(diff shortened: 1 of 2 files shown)


DevDW (dismissed):
LGTM -- should we consider doing something similar if the DB has a status of "Failed"? (disable the button and have an explanatory tooltip) If you click "Show" on a failed DB, the request always fails so no password can be retrieved for it.

codyfinn (dismissed):
Works for me as well. I agree with Dajahi about the failed state.

packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -165,6 +165,7 @@ const databases = [
       label: `database-${req.params.id}`,
       engine: req.params.engine,
       ssl_connection: true,
+      status: 'provisioning',
```
bnussman, line 168: Revert before merge

packages/manager/src/features/Databases/DatabaseDetail/DatabaseSummary/DatabaseSummaryConnectionDetails.tsx
```diff
@@ -143,22 +144,22 @@ export const DatabaseSummaryConnectionDetails: React.FC<Props> = (props) => {
   };
 
   const ssl = database.ssl_connection ? 'ENABLED' : 'DISABLED';
+  const disableShowBtn = ['provisioning', 'failed'].includes(database.status);
+  const disableToolTipText =
+    database.status === 'provisioning'
+      ? 'Your Database Cluster is currently provisioning.'
+      : 'You cannot view your root password when your Database Cluster has failed.';
```
bnussman, line 151: I wonder if it should be more like this because it lets the user know we're not just hiding it from them to be annoying as it a real technical limitation

```suggestion
      : 'Your root password is unavailable when your Database Cluster has failed.';
```

DevDW: I agree with this change

DevDW (approved):
Works well, approving pending the reversion of the change in `serverHandlers.ts` and the change to the `failed` tooltip copy

DevDW (approved):

bnussman (approved):

PR #8169, opened 2022-02-03
DBaaS React Query Minor Change

bnussman:
## Description

- When a Database goes from `provisioning` to `success`, the host fields also get populated. We need to refetch to pick those up

## How to test

- Test event polling transitioning a Database from `provisioning` to `success`


DevDW (approved):

HanaXu (approved):
Approving to get this merged

PR #8173, opened 2022-02-03
Fixes for the tax collection banner

tiffwong:
## Description

Use LaunchDarkly's flag to control the params instead of variables so changes won't be dependent on opening a PR.

## How to test

Change the billing country to Japan and Norway.

### DIFF
--- a/packages/manager/src/dev-tools/FeatureFlagTool.tsx
+++ b/packages/manager/src/dev-tools/FeatureFlagTool.tsx
@@ -9,9 +9,6 @@ import Grid from 'src/components/core/Grid';
 
 const options: { label: string; flag: keyof Flags }[] = [
   { label: 'Databases', flag: 'databases' },
-  { label: 'Region Dropdown', flag: 'regionDropdown' },
-  { label: 'Tax Collection Pt1', flag: 'taxCollectionPart1' },
-  { label: 'Tax Collection Pt2', flag: 'taxCollectionPart2' },
 ];
 
 const FeatureFlagTool: React.FC<{}> = () => {
--- a/packages/manager/src/features/GlobalNotifications/GlobalNotifications.tsx
+++ b/packages/manager/src/features/GlobalNotifications/GlobalNotifications.tsx
@@ -39,7 +39,10 @@ const GlobalNotifications: React.FC<{}> = () => {
       {!isEmpty(suppliedMaintenances) && !hasDismissedMaintenances ? (
         <APIMaintenanceBanner suppliedMaintenances={suppliedMaintenances} />
       ) : null}
-      {flags.taxCollectionPart1 ? <TaxCollectionBanner /> : null}
+      {flags.taxCollectionBanner &&
+      !!Object.keys(flags.taxCollectionBanner).length ? (
+        <TaxCollectionBanner />
+      ) : null}
     </>
   );
 };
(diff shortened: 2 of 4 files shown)


packages/manager/src/dev-tools/FeatureFlagTool.tsx
```diff
@@ -9,9 +9,6 @@ import Grid from 'src/components/core/Grid';
 
 const options: { label: string; flag: keyof Flags }[] = [
   { label: 'Databases', flag: 'databases' },
-  { label: 'Region Dropdown', flag: 'regionDropdown' },
```
tiffwong, line 12: Removed this as well since it's now in production

packages/manager/src/features/GlobalNotifications/GlobalNotifications.tsx
```diff
@@ -39,7 +39,10 @@ const GlobalNotifications: React.FC<{}> = () => {
       {!isEmpty(suppliedMaintenances) && !hasDismissedMaintenances ? (
         <APIMaintenanceBanner suppliedMaintenances={suppliedMaintenances} />
       ) : null}
-      {flags.taxCollectionPart1 ? <TaxCollectionBanner /> : null}
+      {flags.taxCollectionBanner &&
+      !!Object.keys(flags.taxCollectionBanner).length ? (
+        <TaxCollectionBanner />
+      ) : null}
```
HanaXu, line 45:
```suggestion
      {flags.taxCollectionBanner &&
      Object.keys(flags.taxCollectionBanner).length > 0 ? (
        <TaxCollectionBanner />
      ) : null}
```

codyfinn: I would prefer the `!!` or wrapping in a boolean conversion so `Boolean(Object.keys(flags.tsxCollectionBanner).length)`

codyfinn (dismissed):
Aside from a code style comment looks great!

HanaXu (approved):

packages/manager/src/features/GlobalNotifications/GlobalNotifications.tsx
```diff
@@ -39,7 +39,10 @@ const GlobalNotifications: React.FC<{}> = () => {
       {!isEmpty(suppliedMaintenances) && !hasDismissedMaintenances ? (
         <APIMaintenanceBanner suppliedMaintenances={suppliedMaintenances} />
       ) : null}
-      {flags.taxCollectionPart1 ? <TaxCollectionBanner /> : null}
+      {flags.taxCollectionBanner &&
+      Boolean(Object.keys(flags.taxCollectionBanner).length) ? (
```
DevDW, line 43: What's the reasoning behind the type coercion to a boolean instead of a `> 0` check here?

packages/manager/src/features/GlobalNotifications/TaxCollectionBanner.tsx
```diff
@@ -6,65 +6,23 @@ import Typography from 'src/components/core/Typography';
...
-    //         return '[TEXAS DATE]';
-    //       case 'WA':
-    //         return '[WASHINGTON DATE]';
-    //     }
-    //   }
-    // }
-
-    return null;
-  };
+  const { date, action } = flags.taxCollectionBanner!;
```
DevDW, line 25: I think we should do a better job of handling for when the flag might not exist or these properties might not be populated.

bnussman (approved):

DevDW (approved):
Approving given the time crunch but I think we should have a follow-up PR to address a few things

DevDW:
would it be possible to improve the tablet/mobile formatting at all?

Screen Shot 2022-02-04 at 10 15 14 AM
Screen Shot 2022-02-04 at 10 14 39 AM
Screen Shot 2022-02-04 at 10 14 06 AM

PR #8176, opened 2022-02-04
Clean up for tax collection banner

tiffwong:
## Description

Clean up code and fix mobile/tablet styling for the tax collection banner.

## How to test

In order to see the banner, make sure your country is set to Japan or Norway and that it hasn't previously been dismissed.

### DIFF
--- a/packages/manager/src/features/GlobalNotifications/TaxCollectionBanner.tsx
+++ b/packages/manager/src/features/GlobalNotifications/TaxCollectionBanner.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { useHistory } from 'react-router-dom';
 import Button from 'src/components/Button';
 import Box from 'src/components/core/Box';
+import { makeStyles, Theme } from 'src/components/core/styles';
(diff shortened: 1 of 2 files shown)


packages/manager/src/features/GlobalNotifications/TaxCollectionBanner.tsx
```diff
@@ -33,18 +44,18 @@ const TaxCollectionBanner: React.FC<{}> = () => {
         justifyContent="space-between"
       >
         <Typography>
-          Starting {date}, tax may be applied to your Linode services. For more
-          information, please see the{' '}
+          Starting {bannerDate}, tax may be applied to your Linode services. For
```
DevDW, line 47: If `bannerDate` happened to be an empty string, would we want to render a banner at all?

If not, we can add a check a few lines up to return `null` if `!bannerDate`.

HanaXu (approved):

jdamore-linode (approved):

DevDW (approved):
🚀

PR #8178, opened 2022-02-07
M3-5664: Prevent previous saved support text from loading into a reply

tiffwong:
## Description

This PR prevents the saved text from the initial support ticket creation from autofilling the reply textfield.

The next step is to only autofill the reply textfield if the text is related to that reply which will be addressed in a future ticket.

## How to test

1. Open a new ticket
2. Respond to that ticket
3. Reload the page and the textfield should no longer be autofilled

### DIFF
--- a/packages/manager/src/features/Support/SupportTicketDetail/TabbedReply/ReplyContainer.tsx
+++ b/packages/manager/src/features/Support/SupportTicketDetail/TabbedReply/ReplyContainer.tsx
@@ -69,26 +69,33 @@ const ReplyContainer: React.FC<CombinedProps> = (props) => {
   const textFromStorage = storage.ticketReply.get();
   const isTextFromStorageForCurrentTicket =
     textFromStorage.ticketId === props.ticketId;
+  const isTextFromReply = textFromStorage.replyId !== null;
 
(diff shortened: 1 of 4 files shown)


packages/manager/src/features/Support/SupportTicketDetail/TabbedReply/ReplyContainer.tsx
```diff
@@ -69,26 +69,33 @@ const ReplyContainer: React.FC<CombinedProps> = (props) => {
   const textFromStorage = storage.ticketReply.get();
   const isTextFromStorageForCurrentTicket =
     textFromStorage.ticketId === props.ticketId;
+  const isTextFromReply = textFromStorage.replyId !== null;
```
DevDW, line 72: Is there a circumstance under the current implementation where `replyId` would not be `null`? Perhaps this variable should be renamed -- my expectation was that each of my replies to the ticket would have a `replyId`, but that is not the case

DevDW (commented):
The trade-off here seems to be that we're losing the reply text if a user types in a reply and refreshes or navigates away from the page. Is that desirable?

DevDW (approved):
This is working well 🚀

PR #8180, opened 2022-02-08
Fix icon showing after IP address

tiffwong:
## Description

This PR removes the icon that is being displayed after the IP address in the Linode Landing table on Chrome. It's being caused by the `webkit-scrollbar-thumb` selector on the `xScroll` class that is only referenced in `IPAddress.tsx`. I did some digging and I don't think we need that class anymore.

While I was in there, I cleaned up the component a bit.
- Removed `OverflowIPs.tsx`
- Removed the `copyRight` prop and made the copy icon always be displayed on the right
- Fixed the vertical alignment of the copy icon
- Cleaned up a few unused classes

Here's an article on `webkit-scrollbar-thumb`: https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar

## How to test

- Go to the Linode Landing table on Chrome


packages/manager/src/features/linodes/LinodesLanding/IPAddress.test.tsx
```diff
@@ -46,15 +44,13 @@ describe('IPAddress', () => {
     expect(showmore.prop('items')).toEqual(['8.8.4.4']);
   });
 
-  it('should show the copy icon if copyRight is true', () => {
-    expect(component.find('[data-qa-copy-ip]')).toHaveLength(0);
-    component.setProps({ copyRight: true });
+  it('should show the copy icon', () => {
```
HanaXu, line 47: We can just remove this test since it's accounted for in the test below

HanaXu (approved):
🧹 🧹 🚀

jdamore-linode (approved):
Looks good!

PR #8181, opened 2022-02-08
Fix StackScript Broken Cypress Test

jdamore-linode:
## Description

**What does this PR do?**
This fixes the failing StackScript Cypress test. The test was failing because it attempts to use a target image (Alpine 3.10) which no longer exists. I just updated to Alpine 3.15, which appears to be the newest version available, and I'm now seeing the test succeed.

## How to test

You can reproduce the issue by checking out `develop` and running `yarn up` followed by `yarn cy:debug`. Once the Cypress window opens, select `stackscripts.spec.ts` -- you should observe a failure when it attempts to select Alpine 3.10 as the target image for the Stack Script.

Check out this PR and run the same steps to observe that the test no longer fails.

### DIFF
--- a/packages/manager/cypress/integration/stackscripts/stackscripts.spec.ts
+++ b/packages/manager/cypress/integration/stackscripts/stackscripts.spec.ts
@@ -19,7 +19,7 @@ const createLinode = () => {
 
 describe('stackscripts', () => {
   it('create stackscript, use it to deploy linode', () => {
-    const disk = 'Alpine 3.10';
+    const disk = 'Alpine 3.15';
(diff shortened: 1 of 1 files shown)


bnussman (approved):
🚀

DevDW (approved):
🚢

HanaXu (approved):
Is there a way to update the test to pick the first dropdown value or something? That way, the test wouldn't fail the next time the name changes 🤔

jdamore-linode:
> Is there a way to update the test to pick the first dropdown value or something? That way, the test wouldn't fail the next time the name changes 🤔

That's a good idea -- I'll look into that!

PR #8183, opened 2022-02-08
M3-4635: Disable "Make a Payment" button if restricted user has read_only access

DevDW:
## Description
If a user has `read_only` access for the `account_access` grant, the "Make a Payment" button on the /account/billing page should be disabled.

Additionally, if an account has a balance, there's a link in the "Account Balance" section that says `Make a payment.` below "Balance". Clicking it would open the same payment drawer as clicking the "Make a Payment" button does. I went ahead and made the display of that link conditional on whether the user has `read_only` access or not.

## How to test
On a single account, look at the `/account/billing` page as an unrestricted user and as a restricted user. 

The unrestricted user should see an enabled "Make a Payment" button and the `Make a payment.` text in the "Account Balance" section.

The restricted user should see a disabled "Make a Payment" button and no `Make a payment.` link in the "Account Balance" section.

### DIFF
--- a/packages/manager/src/components/LandingHeader/LandingHeader.tsx
+++ b/packages/manager/src/components/LandingHeader/LandingHeader.tsx
@@ -23,6 +23,7 @@ export interface Props extends Omit<HeaderProps, 'actions'> {
   createButtonText?: string;
   loading?: boolean;
   breadcrumbProps?: BreadcrumbProps;
+  disabledCreateButton?: boolean;
 }
 
 /**
@@ -43,6 +44,7 @@ export const LandingHeader: React.FC<Props> = (props) => {
     createButtonText,
     loading,
     breadcrumbProps,
+    disabledCreateButton,
   } = props;
 
   const defaultCreateButtonWidth = 152;
@@ -59,6 +61,7 @@ export const LandingHeader: React.FC<Props> = (props) => {
             loading={loading}
             onClick={onAddNew}
             style={{ width: createButtonWidth ?? defaultCreateButtonWidth }}
+            disabled={disabledCreateButton}
           >
             {createButtonText ?? `Create ${entity}`}
           </Button>
@@ -73,6 +76,7 @@ export const LandingHeader: React.FC<Props> = (props) => {
       createButtonWidth,
       createButtonText,
       entity,
+      disabledCreateButton,
     ]
   );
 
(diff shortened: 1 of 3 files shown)


jdamore-linode (approved):
Looks good! Confirmed that the "Make a Payment" button is enabled on unrestricted accounts but is disabled for accounts with "read_only" access. Likewise, confirmed that the "Make a Payment" link in the Account Balance panel is shown for unrestricted users but hidden for those with "read_only" access.

Only thing catching my eye is that when Mock Service Worker is enabled, the mocked user only has "read_only" access (because that's the [default for the grants factory](https://github.com/linode/manager/blob/develop/packages/manager/src/factories/grants.ts#L13)) and the Make a Payment button will now always be disabled as a result. I'm not sure if this'll interfere with any tests or otherwise inconvenience us, just calling it out in case.

(I'm running E2E tests for this branch locally right now so I'll report back if it does impact any Cypress tests)

packages/manager/src/components/LandingHeader/LandingHeader.tsx
```diff
@@ -23,6 +23,7 @@ export interface Props extends Omit<HeaderProps, 'actions'> {
   createButtonText?: string;
   loading?: boolean;
   breadcrumbProps?: BreadcrumbProps;
+  disabledCreateButton?: boolean;
```
tiffwong, line 26: Can we change the variable name to `disabledButton` instead since it might not always be referencing a create button like in this instance?

tiffwong (approved):
Functionality looks great!

Left a comment on the variable name but in retrospect, we should probably cleanup the other variable names in a different PR so it encapsulates using the component for other than just create.

Also not related to the changes made in this PR, but can we prevent the `.` from being underlined?
Screen Shot 2022-02-09 at 11 41 45 AM

DevDW:
> Only thing catching my eye is that when Mock Service Worker is enabled, the mocked user only has "read_only" access (because that's the [default for the grants factory](https://github.com/linode/manager/blob/develop/packages/manager/src/factories/grants.ts#L13)) and the Make a Payment button will now always be disabled as a result. I'm not sure if this'll interfere with any tests or otherwise inconvenience us, just calling it out in case.

good catch -- I went ahead and changed that to `read_write`. Since the mocks are for testing purposes I think it is better for us to have them default to being more permissive rather than restrictive. If there's a specific case someone is trying to test that requires `read_only` there, it can easily be changed locally for their purposes

PR #8187, opened 2022-02-10
M3-5660: DRY up E2E tests

jdamore-linode:
## Description

**What does this PR do?**
This replaces the two E2E test workflows, `e2e_pr.yml` and `e2e_schedule_and_push`, with a single workflow that is triggered upon PR creation, push to `develop`, `staging`, and `master`, and at its normal scheduled time (8:00 AM ET).

## Post-Merge Considerations
The new workflow will need to be enabled after this is merged, and no E2E tests will run in the meantime. I seem to have the privileges needed to disable workflows, so I'm assuming that I'll be able to enable this workflow as well.


jdamore-linode:
Temporarily added the `Do Not Merge` and `Work In Progress` labels because this PR will need to be updated to disable Cypress.io recordings on PR E2E tests. See PR #8259 for more info.

jdamore-linode:
Opted to restore the existing workflow files so that we can disable them and enable the new workflow at our leisure. Also, if the new workflow ends up having any issues, we can easily disable it and re-enable these workflows until the issue is fixed.

Once we're satisfied with the state of the new workflow, I'll circle back and remove the old workflows (and I'll write a Jira ticket so that task isn't forgotten)

HanaXu (approved):

Issue #7896, opened 2021-08-30
Linode Networking tab copy


tsmgeek:
Can we have click-2-copy button on the linode networking tab for IP addresses. It exists on summary lists, just not within the node itself.

tiffwong:
@tsmgeek this will be implemented in our next release :)

PR #8188, opened 2022-02-10
Add copy tooltip to IP addresses on the Linode Networking tab

tiffwong:
## Description

We usually have a copy tooltip associated with IP addresses and there's an open issue (https://github.com/linode/manager/issues/7896) asking for it.

## How to test

- Go to the Networking tab on a Linode Detail page
- The copy tooltip should show when the user is hovering over a row
- The copied text should match up with the associated IP address


bnussman (commented):
Would there be any way to improve the awkward wrapping? Maybe disable wrapping? 

Screen Shot 2022-02-11 at 5 24 47 PM

bnussman (approved):

PR #8191, opened 2022-02-11
M3-5593: Convert Search Story to MDX

HanaXu:
## Description
Updates the `DebouncedSearchTextField` story from `tsx` to `mdx`

## How to test
- `yarn storybook`
- Go to http://localhost:6006/?path=/docs/components-search and check all the stories


jdamore-linode (approved):
Looks good to me!

tiffwong (commented):
Not sure if we need to keep the Select Field example.

If we do, can you extend the canvas height to prevent the dropdown options from being cut off?

Screen Shot 2022-02-14 at 4 23 54 PM

PR #8192, opened 2022-02-11
M3-5589: Added Storybook Story for File Uploads

bnussman:
## Description

Adds a Storybook Story for our File Uploader

Because we don't have a generic file uploader component (😔), there are two variants of the File Uploader in this story
- Image Uploader
- Object Uploader

## How to test

- `yarn storybook`
- Find the `Components`/`File Uploader` story and test it!

### DIFF
--- a/packages/manager/src/components/FileUploader/FileUploader.stories.mdx
+++ b/packages/manager/src/components/FileUploader/FileUploader.stories.mdx
@@ -64,18 +64,22 @@ This component enables users to attach and upload custom Images.
       apiError: {
         description: "An error to display if an upload error occoured",
         type: { summary: 'string | undefined', name: 'string', required: true },
+        control: { type: null },
       },
       setErrors: {
         description: "A function that allows you to set an error value in the parent component",
         type: { summary: 'React.Dispatch<React.SetStateAction<APIError[] | undefined>>', required: true },
+        control: { type: null },
       },
       setCancelFn: {
         description: "Allows you to set a cancel upload function in the parent component",
         type: { summary: 'React.Dispatch<React.SetStateAction<(() => void) | null>>', required: true },
+        control: { type: null },
       },
       onSuccess: {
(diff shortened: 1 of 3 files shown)


jdamore-linode (approved):
Looks good!

packages/manager/src/components/FileUploader/FileUploader.stories.mdx
```diff
@@ -0,0 +1,131 @@
+import { Canvas, Meta, Story, ArgsTable } from '@storybook/addon-docs';
+import FileUploader from './FileUploader';
+import ObjectUploader from '/src/features/ObjectStorage/ObjectUploader/ObjectUploader';
+
+<Meta title="Components/File Uploads" />
```
tiffwong, line 5: The documentation says "Uploader" but the side menu has these stories listed says "Uploads" so I think we should make them consistent.

Screen Shot 2022-02-14 at 4 08 52 PM

packages/manager/src/components/FileUploader/FileUploader.stories.mdx
```diff
@@ -0,0 +1,131 @@
...
+      description: "My Ubuntu Image for Production",
+      dropzoneDisabled: false,
+      apiError: undefined,
+      setErrors: undefined,
+      setCancelFn: undefined,
+      onSuccess: undefined,
+    }}
+    argTypes={{
+      label: {
+        description: "The label of the upload that will be sent to the Linode API (used for Image uploads)",
```
tiffwong, line 41: Also all the props here look like they're required other than `onSuccess` 🤔

HanaXu (changes requested):
Setting `apiError` doesn't seem to do anything 🤔 . If we're keeping that field, can we initialize an empty string so that there's no extra `Set...` button click? And maybe set the controls for the function fields to null as well?

image

HanaXu (approved):

PR #8198, opened 2022-02-14
M3-5595: Storybook for Accordions

bnussman:
## Description

- Cleans up and converts the Accordions story to `mdx`
- Removes the Asynchronous Content example
- Adds copy
- Adds a Component API section with and Args Table
- Update Show More Expansion Story
- Convert Show More Expansion to a FC

## How to test

- `yarn storybook`
- Check all Accordion stories ([Link](http://localhost:6006/?path=/docs/ui-elements-accordion--default-story))

### DIFF
--- a/packages/manager/src/components/Accordion/Accordion.stories.mdx
+++ b/packages/manager/src/components/Accordion/Accordion.stories.mdx
@@ -4,7 +4,7 @@ import Accordion from './Accordion';
 import { Canvas, Meta, Story, ArgsTable } from '@storybook/addon-docs';
 
 <Meta
-  title="UI Elements/Accordion"
+  title="UI Elements"
 />
 
 # Accordion
@@ -20,14 +20,6 @@ Accordions are better suited for when people need to focus on content that is ke
   - “People treat clicks like currency and they don't spend it frivolously” - NNG [Citation](https://www.nngroup.com/articles/clickable-elements/)
 - Increases interaction cost
 - Accordions collapsed by default diminishes people's awareness of content
-
-<Canvas>
-  <Story name="Default" >
-    <Accordion heading="This an Accordion">
-      <p>The children of the Accordion go here!</p>
-    </Accordion>
-  </Story>
-</Canvas>
 
 export const renderActions = () => {
   return (
@@ -48,13 +40,11 @@ export const Template = (args) => (
   </Accordion>
 );
 
-## Component API
-
 The Accordion component's props extends the Material UI Accordion props
 
 <Canvas>
   <Story
-    name="Component API"
+    name="Accordion"
     args={{
       heading: "This is an Accordion",
       defaultExpanded: undefined,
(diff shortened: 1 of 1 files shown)


packages/manager/src/components/ShowMoreExpansion/ShowMoreExpansion.tsx
```diff
@@ -1,106 +1,78 @@
...
   };
 
-  componentDidUpdate(prevProps: Props, prevState: State) {
-    if (
-      prevState.open !== this.props.defaultExpanded &&
-      prevProps.defaultExpanded !== this.props.defaultExpanded
-    ) {
-      this.setState({ open: this.props.defaultExpanded });
-    }
-  }
```
bnussman, line 72: Please check behind me on removing this

packages/manager/src/components/Accordion/Accordion.stories.mdx
```diff
@@ -0,0 +1,159 @@
...
+# Accordion
+
+Accordions are better suited for when people need to focus on content that is key for decision making and other supplementary content can be hidden. Accordions should be avoided when users need most if not all of the content on a page. Scrolling isn't as big of an issue as once thought.
+
+### Pros
+- Puts focus on the content the user is interested in 
+- Can make pages seem less complex
+ 
+### Cons
+- Can be cumbersome which adds to cognitive load. “People treat clicks like currency and they don't spend it frivolously.” - NNG [Citation](https://www.nngroup.com/articles/clickable-elements/)
```
tiffwong, line 19: Can the quote be a sub-bullet?

packages/manager/src/components/Accordion/Accordion.stories.mdx
```diff
@@ -0,0 +1,159 @@
...
+- Increases interaction cost
+- Accordions collapsed by default diminish people's awareness of content
```
tiffwong, line 21: "diminishes"

tiffwong (changes requested):
We can remove the `success`/`warning`/`error` examples since they can be generated in the ArgsTable.

Can we initialize the values for the following props to prevent the user from having to click the `Set ...` button to change the value?
- `defaultExpanded`
- `expanded`
- `headingNumberCount`
- `actions`
- `success`
- `warning`
- `error`

And then can we disable the controls for the remaining props?

Also since we're moving the `Show More Expansion` story to its own section, can we reference the use case for it over `Accordion` in `Accordion` and vice versa?

bnussman:
@tiff going to make most of these changes. In the args table, I leave many values uninitialized because they are optional props and imo the `Set boolean` buttons help the reader understand they are setting an optional prop

Additionally, I must leave `expanded` undefined with a `Set boolean` button so `defaultExpanded` won't get overwritten.

HanaXu (commented):
Can we rename the component API to Accordion and remove the default since they're basically the same and to unnest the sidemenu?

tiffwong (approved):

HanaXu (approved):

PR #8205, opened 2022-02-16
M3-5622: Story for Spacing

DevDW:
## Description
Adds a story for Spacing

## How to test
In the `Core` section, you should now see a story titled "Spacing".

### DIFF
--- a/packages/manager/src/components/core/Spacing.stories.mdx
+++ b/packages/manager/src/components/core/Spacing.stories.mdx
@@ -1,0 +1,5 @@
+<Meta title="Core/Spacing" />
+
+# Spacing
+### Overview
+Spacing between elements is typically in multiples of 8 (e.g., 8px, 16px, 24px, etc.).


packages/manager/src/components/core/Spacing.stories.mdx
```diff
@@ -0,0 +1,5 @@
+<Meta title="Core/Spacing" />
+
+# Spacing
+### Overview
+Spacing between elements is typically in multiples of 8 (e.g., 8px, 16px, 24px, etc.).
```
tiffwong, line 5: If this is the only content for now, what do you think of getting rid of "Overview"?

tiffwong (approved):

packages/manager/src/components/core/Spacing.stories.mdx
DevDW: Fair change 👌

jdamore-linode (approved):

PR #8207, opened 2022-02-16
M3-5606: Storybook for Links

bnussman:
## Description

- Makes all link stories `mdx` instead of `tsx`
- Adds copy to each link
- Adds args table for each variant of link
- Made [ExternalLink.tsx](https://github.com/linode/manager/compare/develop...bnussman:M3-5606-storybook-for-links?expand=1#diff-0f4a61ee9add8127a60665c499b8d62a3c7ffc23afb84f39c4d412521e6ea4a2) a FC

## How to test

- `yarn storybook`
- Check all 3 link stories
  - Find these under `UI Elements/Links`


jdamore-linode (approved):
Looks good!

The only thing that caught my eye is that the `href` prop (or equivalent -- they're each named differently) behaves similarly for each component but has a different description for each:

- Docs Link `href` prop: _The destination URL for the link_
- External Link `link` prop: _The destination for the link to bring you to_
- Link `to` prop: _The link's destination. If the value contains http or mailto, it will be considered an external link and open in a new window._

Is there any room to improve consistency in these descriptions? Perhaps we could set the first two descriptions to _The link's destination_ to better match the `<Link />` component?

This can definitely be cleaned up later, no need to delay the merge for this!

HanaXu (approved):
LGTM, could we exclude `className` from the argstable?

PR #8208, opened 2022-02-17
TypeScript Docgen for Storybook

bnussman:
## Description

- Allows Storybook to better infer args from Typescript types
- The diff shows some changes that should be made for this concept to work better for a component (Dialog in this case)

```ts
<Meta title="Components/Paper" component={Paper} />
```

## How to test

- Go look at some stories!
  - Paper
  - Dialog

https://user-images.githubusercontent.com/6440455/154551144-88201c3c-c971-4911-b08f-06efbf0b3fd6.mov

### DIFF
--- a/packages/manager/src/GoTo.tsx
+++ b/packages/manager/src/GoTo.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { makeStyles, Theme } from 'src/components/core/styles';
 import EnhancedSelect, { Item } from 'src/components/EnhancedSelect/Select';
-import MUIDialog from 'src/components/core/Dialog';
+import Dialog from '@material-ui/core/Dialog';
 import { useHistory } from 'react-router-dom';
 import useFlags from './hooks/useFlags';
 import { isFeatureEnabled } from './utilities/accountCapabilities';
@@ -168,7 +168,7 @@ const GoTo: React.FC<CombinedProps> = (props) => {
   ]);
 
   return (
-    <MUIDialog
+    <Dialog
       classes={dialogClasses}
       title="Go To..."
       open={props.open}
@@ -195,7 +195,7 @@ const GoTo: React.FC<CombinedProps> = (props) => {
           menuIsOpen={true}
         />
       </div>
-    </MUIDialog>
+    </Dialog>
   );
 };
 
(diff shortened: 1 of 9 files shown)


tiffwong:
I find I adding `export` in the FC declaration also helps import the props

From https://github.com/storybookjs/storybook/issues/13304#issuecomment-745360896
> e.g
> `export const Widget: React.FC<WidgetProps> = ({`
> vs
> `const Widget: React.FC<WidgetProps> = ({`

bnussman:
The last commit before this comment is a big hot take. It shows what changes need to be made for the `Dialog` type inferences to work in Storybook.

packages/manager/src/GoTo.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { makeStyles, Theme } from 'src/components/core/styles';
 import EnhancedSelect, { Item } from 'src/components/EnhancedSelect/Select';
-import MUIDialog from 'src/components/core/Dialog';
+import Dialog from '@material-ui/core/Dialog';
```
bnussman, line 4: You may be thinking, "We shouldn't be importing from Material UI directly". I would agree, but the old import (`src/components/core/Dialog`) was a **useless** wrapper of the MUI dialog. Therefore, I'd argue the author of this code should have probably imported from `src/components/Dialog`. If the author specifically needed the direct MUI dialog component, an import from MUI directly should be allowed.

packages/manager/src/components/Dialog/Dialog.stories.mdx
```diff
@@ -8,6 +8,7 @@ import { Canvas, Meta, Story, ArgsTable } from '@storybook/addon-docs';
 
 <Meta
   title="Components/Dialogs"
+  component={Dialog}
```
bnussman, line 11: This must be here for Storybook to pull types

packages/manager/src/GoTo.tsx
codyfinn: I agree for the most part. Though I can see how us using wrappers around MUI has advantages. I know this is a huge if but if we move away from MUI then we control the typing and API of our wrappers. So refactoring or migration away from MUI would be easier if we wrap stuff even if it is a direct pull from MUI.

bnussman: I see what you're saying. I guess my argument is that the `src/components/core/` files are not needed and all of the wrapping should be done in `src/components/Dialog/Dialog.tsx`. With our current directory structure, it seems were wrapping twice. I'm thinking wrapping once directly in `src/components/Dialog/Dialog.tsx` is sufficient.

PR #8216, opened 2022-02-21
React Query for Managed

bnussman:
## Description

- Use React Query for all Managed features in Cloud Manager 
- Not part of the active sprint but I had some free time last night and did most of this

## How to test

> ⚠️ I recommend testing this PR against a non-production environment because Linode Support may be alerted when you create Managed Monitors and it may disrupt their work

-  Test all Managed functionality in Cloud Manager
    - View the Testing Checklist in the comments 📄☑️


bnussman:
# Testing Checklist

There is a list of items to check for each tab in the Managed UI

### Summary

- [ ] Stats Graph (check functionality and refetch behavior)
  - [ ] CPU
  - [ ] Network
  - [ ] Disk I/O
- [ ] Open Support Tickets
  - [ ] TODO: this is harder to test 
- [ ] All Monitors Status
  - [ ] TODO: this is harder to test 

### Monitors

- [ ] Add Monitor
  - [ ] Add a monitor with no contact group and no credentials
  - [ ] Add a monitor with a contact group (ensure contacts are populated correctly in the form)
  - [ ] Add a monitor with credentials (ensure credentials are populated correctly in the form)

- [ ] Edit a Monitor
  - [ ] You should be able to edit a monitor in any way and have the React Query cache update accordingly
  - [ ] You should be able to enabled / disable a Monitor and have the cache update accordingly

- [ ] Delete a Monitor
  - [ ] You should be able to delete a monitor and see the cache update accordingly

### SSH Access

> There should be a SSH Access item for every Linode on your account

- [ ] Edit a SSH Key Access
  - [ ] You should be able to edit a ssh key access and see the cache update accordingly
  - [ ] You should be able to enable / disable a SSH Key Access item and see the cache update accordingly

As an edge case, a load SSH Access data, create or delete a Linode, then navigate back to the SSH Access tab. We want the React Query cache to invalidate the SSH Access data so the table is up to date without making the user refresh. This extra logic was not needed previously because the SSH Access data would be loaded every time the user accessed the tab 

### Credentials

- [ ] Add Credential
  - [ ] You should be able to add a credential and see the cache update accordingly 

- [ ] Edit Credential
  - [ ] You should be able to update a credential's label and see the cache update accordingly
  - [ ] You should be able to update just the password of a credential (no UI update expected)
  - [ ] You should be able to update the username **and** password of a credential (no UI update expected)

> A UI update may occur if the credential has been previously decoded by Linode Support and now the username or password changed. Consider testing this. 

- [ ] Delete Credential
  - [ ] You should be able to delete a credential and see the cache update accordingly 

### Contacts

packages/manager/src/queries/managed/helpers.ts
```diff
@@ -23,7 +23,7 @@ export const extendIssues = async (issues: ManagedIssue[]) => {
      * does not include the status or date closed.
      */
     return (
-      getTicket(thisIssue.id)
+      getTicket(thisIssue.entity.id)
```
bnussman, line 26: This existing code I reused had a major bug. The Managed Issue `id` is **not** the ticket `id`. We needed to be using `thisIssue.entity.id` to get the support ticket

packages/manager/src/features/Managed/Monitors/IssueCalendar.tsx
```diff
@@ -40,17 +40,18 @@ export const generateCalendar = (timezone: string, issues: ManagedIssue[]) => {
   const days: CalendarDay[] = [];
 
   // Start with today, since it will be at the top of our list.
-  const day = DateTime.local().setZone(timezone);
   for (let i = 0; i < TOTAL_DAYS; i++) {
+    const day = DateTime.local().minus({ days: i }).setZone(timezone);
     /**
```
bnussman, line 45: This was also a major bug. `day` should have been calculated for every iteration of this loop so that `createdOnTargetDay` would return true

DevDW (commented):
Going to start combing through this. In the meantime I merged in the latest `develop` and resolved the conflicts, please take a second look at those resolutions in case something was removed that shouldn't have been

DevDW (commented):
Still doing some testing and have a few more files to comb through, but leaving some comments

packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -788,6 +790,13 @@ export const handlers = [
     const tickets = supportTicketFactory.buildList(15, { status: 'open' });
     return res(ctx.json(makeResourcePage(tickets)));
   }),
+  rest.get('*/support/tickets/999', (req, res, ctx) => {
```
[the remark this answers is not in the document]
bnussman: I just mocked that specific support ticket with id `999` so it is linked here
Screen Shot 2022-06-15 at 3 45 16 PM

DevDW (approved):
From my testing and comb through of the code, this LGTM -- didn't observe any bugs. Nice work 👍

jdamore-linode (approved):
Nice work Banks! Went through all your test instructions and everything worked really well -- didn't see any issues to bring up. The factories and mocks are really nice additions too, and the factories will be especially handy for the Cypress tests I'm working on

PR #8218, opened 2022-02-22
Add Cypress E2E tests for DBaaS cluster creation

jdamore-linode:
## Description

**What does this PR do?**
* Adds Cypress tests for DBaaS cluster creation using mocked data and API responses
* Invalidates database list query upon database finished or failed event

## How to test

* To verify E2E tests, run the following from the repo root and confirm that tests succeed:
  * `yarn cy:run -s '**/create-database.spec.ts'`
* To verify React Query changes, navigate to DBaaS landing page, create a DB cluster, and confirm that Manager automatically and immediately makes a request to `/databases/instances` after receiving an event with action `database_create` and status `finished`
* Perform additional manual testing to ensure that React Query changes do not cause any unintended side effects


packages/manager/src/queries/databases.ts
```diff
@@ -182,6 +182,7 @@ export const databaseEventsHandler = (event: Event) => {
           // Database status will change from `provisioning` to `active` (or `failed`) and
           // the host fields will populate. We need to refetch to get the hostnames.
           queryClient.invalidateQueries([queryKey, entity!.id]);
+          queryClient.invalidateQueries(`${queryKey}-list`);
```
jdamore-linode, line 185: I'm still trying to figure out React Query, so let me know if this is bad or unnecessary!

I don't think this is strictly necessary for the E2E tests, but I noticed that when getting the `database_create` `finished` event, the landing page statuses did not immediately refresh and instead took a few seconds to reflect the changes (I assume this corresponds to the `refetchInterval` specified in line 51 of this file). This delay occasionally causes the test to time out and I assumed this was not the intended behavior.

packages/manager/cypress/integration/databases/create-database.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+  clusterSize: ClusterSize;
+  dbType: Engine;
+  regionTypeahead: string;
+  region: string;
+  engine: string;
+  version: string;
+}
+
+// Array of database cluster configurations for which to test creation.
+const databaseConfigurations: databaseClusterConfiguration[] = [
```
patthiel, line 21: Nice. If we add other test suites for things like managing access, etc. I would suggest having these mocks live in their own file. But that can be done in follow up tests should it be needed

packages/manager/cypress/integration/databases/create-database.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+    region: 'us-southeast',
+    engine: 'MySQL',
+    version: '5.7.30',
+  },
+];
+
+describe('create a database cluster, mocked data', () => {
+  databaseConfigurations.forEach(
+    (configuration: databaseClusterConfiguration) => {
+      it(`creates a ${configuration.linodeType} ${configuration.engine} v${configuration.version} ${configuration.clusterSize}-node cluster at ${configuration.region}`, () => {
```
patthiel, line 47: Looks good! I think adding a TODO for this test and associated ticket for pricing (post-beta) will be good, as it's the only piece that i see missing

patthiel (commented):
LGTM. Ran the tests and they are all sound. Thx for addressing my feedback!

My review/approval is not required. I would defer to cloud team members for approval.

packages/manager/src/queries/databases.ts
bnussman: I'm good with that change!

Alternatively we could say this below to avoid hitting the API again, but refetching is nice to ensure we have accurate data. Since the event returns a status, I think this approach is pretty safe. You can make the final decision 🙂

```suggestion
          updatePaginatedDatabaseStore(entity!.id, {
            status: status === 'failed' ? 'failed' : 'active',
          });
```

bnussman (approved):

DevDW (approved):
LGTM 🚢

PR #8219, opened 2022-02-22
M3-5621: Update stories for Typography

codyfinn:
## Description

Adds stories for Typography components using the MDX format

## How to test

Ensure the stories for the Typography components render correctly.


packages/manager/src/components/Typography.stories.mdx
```diff
@@ -1,80 +1,93 @@
...
-};
+# Typography
 
-export const Headings = () => (
-  <React.Fragment>
+## Headings
+
+
+### Primary heading (h1)
+Implemented using the `h1` variant.
```
HanaXu, line 14: Can we remove the `Implemented using the <variant> variant` lines as it's kind of redundant?

packages/manager/src/components/Typography.stories.mdx
```diff
@@ -1,80 +1,93 @@
...
+      <Divider />
     <Typography variant="button" gutterBottom>
       Button variant
     </Typography>
-    <Divider />
-  </React.Fragment>
-);
+      <Divider />
+  </Story>
+</Canvas>
```
HanaXu, line 91: This looks pretty cluttered, can we add an args table for Typography instead?

image

packages/manager/src/components/Typography.stories.mdx
```diff
@@ -1,80 +1,93 @@
...
+<Canvas>
+  <Story name="Placeholder">
     <Placeholder title="Placeholder/Null Heading">
       This is a separate component utilized for null state headings.
     </Placeholder>
-    <Divider />
-  </React.Fragment>
-);
+  </Story>
+</Canvas>
```
HanaXu, line 62: Not sure if we need to include this here. There's a separate `Placeholders.stories.mdx` file

HanaXu (changes requested):

packages/manager/src/components/Currency/Currency.stories.mdx
```diff
@@ -0,0 +1,29 @@
...
+)
+
+# Currency
+
+<Canvas>
+  <Story 
+    name="Currency"
+    args={{
+      quantity: 4.51,
+      decimalPlaces: 2,
```
DevDW, line 17: Maybe the default quantity should be `4` so that we can toggle the `decimalPlaces` more flexibly. Right now if I set the decimal places to `0` or `1`, it still shows `4.51`.

packages/manager/src/components/Currency/Currency.stories.mdx
```diff
@@ -0,0 +1,29 @@
...
+      wrapInParentheses: false,
```
DevDW, line 18: for the description of this, it might be good to clarify how we use it. From the current uses I see in the codebase, it only gets used to indicate negative amounts, but the actual logic in `Currency.tsx` is a tad ambiguous as to whether that _has_ to be the case.

packages/manager/src/components/Typography.stories.mdx
codyfinn: I agree that it is cluttered but one thing I was hoping to do was have the different variants be easily compared.

bnussman: I'd like to see an args table too if that's possible

packages/manager/src/components/DateTimeDisplay/DateTimeDisplay.stories.mdx
```diff
@@ -0,0 +1,48 @@
...
+      }
+    },
+    displayTime: {
+      description: 'If true displays time component of the date and time provided.',
+    },
+    format: {
+      description: 'String that specifies a luxon compatible format to use.',
+    },
+    humanizeCutoff: {
+      description: 'If the date and time provided is within the designated time frame then the date is displayed as a relative date.'
```
tiffwong, line 35: All the options display the same result for me in the ArgsTable 🤔

packages/manager/src/components/Typography.stories.mdx
tiffwong: Agreed

packages/manager/src/components/Typography.stories.mdx
tiffwong: I agree with having an ArgsTable

Might be cleaner if we remove `<Canvas>` from these examples and label each variant so we can still be able to compare the different styles

HanaXu (approved):

packages/manager/src/components/Typography.stories.mdx
```diff
@@ -0,0 +1,89 @@
...
+import Typography from 'src/components/core/Typography';
+
+<Meta title="Core/Typography/Text" component={Typography} />
+
+# Typography
+
+## Headings
+
+### Primary heading (h1)
+Utilized for
```
bnussman, line 12: I'm not sure it's worth having `Utilized for` below every heading

bnussman (approved):
Looks good, added one optional comment

packages/manager/src/components/DateTimeDisplay/DateTimeDisplay.stories.mdx
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }}>
+    {DateTimeDisplayTemplate.bind({})}
+  </Story>
+</Canvas>
+
+<ArgsTable
+  story="Date Time Display"
```
DevDW, line 43: Looks like something wonky happened with the `humanizeCutoff` options -- they're not showing up

Screen Shot 2022-03-01 at 5 00 54 PM
s

DevDW (commented):
Looks good overall, but there's one issue with the Date Time Display story

PR #8223, opened 2022-02-23
M3-5732: Improve styling consistency for backup auto enrollment

jdamore-linode:
## Description

**What does this PR do?**
Improves the consistency of the styling for the "enable now" link that is found in the "Backup Auto Enrollment" section of the Account Settings page.

## How to test

Navigate to `/account/settings` and confirm that the "enable now" link style is consistent with other links throughout Cloud Manager.

### DIFF
--- a/packages/manager/src/features/Account/AutoBackups.tsx
+++ b/packages/manager/src/features/Account/AutoBackups.tsx
@@ -1,13 +1,12 @@
-import OpenInNew from '@material-ui/icons/OpenInNew';
+import { makeStyles, Theme } from 'src/components/core/styles';
 import * as React from 'react';
 import Accordion from 'src/components/Accordion';
-import Button from 'src/components/Button';
 import FormControlLabel from 'src/components/core/FormControlLabel';
-import { makeStyles, Theme } from 'src/components/core/styles';
-import Typography from 'src/components/core/Typography';
 import Grid from 'src/components/Grid';
 import Notice from 'src/components/Notice';
+import OpenInNew from '@material-ui/icons/OpenInNew';
 import Toggle from 'src/components/Toggle';
+import Typography from 'src/components/core/Typography';
 
 const useStyles = makeStyles((theme: Theme) => ({
   footnote: {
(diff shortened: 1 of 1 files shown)


packages/manager/src/features/Account/AutoBackups.tsx
```diff
@@ -19,6 +18,10 @@ const useStyles = makeStyles((theme: Theme) => ({
     fontSize: '0.8em',
     marginLeft: theme.spacing(1) / 3,
   },
+  enableBackupsButton: {
+    ...theme.applyLinkStyles,
+    fontSize: 14,
```
DevDW, line 23: let's go with `0.875rem` here -- equivalent sizes but our conventional best practice is to use `rem` over `px`and `em`

packages/manager/src/features/Account/AutoBackups.tsx
```diff
@@ -91,15 +94,14 @@ const AutoBackups: React.FC<Props> = (props) => {
           <Grid item>
             <Typography variant="body1" className={classes.footnote}>
               For existing Linodes without backups,&nbsp;
-              <Button
-                compact
-                buttonType="secondary"
+              <button
                 data-qa-backup-existing
                 onClick={openBackupsDrawer}
                 title="enable now"
```
DevDW, line 100: thoughts on removing the `title`? Right now if you hover over the button, you'll see that little pop-up that says "enable now". If we remove the `title`, that will go away.

DevDW (approved):
LGTM, left a couple of small comments/tweaks 👍

packages/manager/src/features/Account/AutoBackups.tsx
jdamore-linode: Agreed, I think it's pretty redundant as-is. For screen reader users, do you think it'd be preferable to remove the title, or to change the title to something more useful or complete, like "Enable Backup Auto Enrollment"?

DevDW: Good question -- I'm leaning towards the latter, but we are inconsistent in the codebase about having the `title` attribute. @tiffwong what are your thoughts on this?

tiffwong: I did some research and the `title` attribute is not helpful and gets missed a lot on screen readers.

tiffwong (approved):
Great job on this!

PR #8224, opened 2022-02-23
M3-5618: Update story for colors

tiffwong:
## Description

- Clean up the colors used throughout the app and remove `cmr` from the naming convention
- Apply colors to the component itself if they're only used once instead of cluttering the theme colors
- Reorganize the layout to prevent the text overlapping

## How to test

- Run `yarn storybook`
- Go to `http://localhost:6006/?path=/docs/core-color-palette--color-palette`
- Make sure none of the color changes regressed on the app in light and dark mode


bnussman:
As a somewhat related topic, what would it take to get type safety on things like this? Would this be in the scope of your changes? 
Screen Shot 2022-02-23 at 5 30 09 PM

packages/manager/src/components/ColorPalette/ColorPalette.tsx
```diff
@@ -13,30 +13,29 @@ const useStyles = makeStyles((theme: Theme) => ({
...
   },
   swatch: {
-    content: '',
-    height: 64,
-    width: 64,
-    margin: theme.spacing(2),
+    border: '1px solid #888f91',
+    borderRadius: 3,
+    height: theme.spacing(4),
+    width: theme.spacing(4),
```
HanaXu, line 23: Can we make the swatches a little bit bigger?

```suggestion
    height: theme.spacing(5),
    width: theme.spacing(5),
```

packages/manager/src/components/ColorPalette/ColorPalette.tsx
```diff
@@ -136,9 +210,38 @@ export const ColorPalette: React.FC<CombinedProps> = (props) => {
...
+    );
+  };
+
   return (
     <Grid container className={classes.root}>
-      {props.displayBackgrounds ? renderBackgrounds() : renderColors()}
+      {renderColors()}
+      {renderBackgrounds()}
+      {renderTextColors()}
+      {renderBorderColors()}
```
HanaXu, line 244: We can deduplicate all the render functions by doing something like


```typescript

interface Color {
  color: string;
  alias: string;
}

const renderColor = (heading: string, colors: Color[]) => {
    return (
      <>
        <Grid item xs={12}>
          <Typography variant="h2">{heading}</Typography>
        </Grid>
        {colors.map((color) => createSwatch(color.color, color.alias))}
      </>
    );
};

return (
    <Grid container className={classes.root}>
      {renderColor('Primary Colors', primaryColors)}
      {renderColor('Etc.', etc)}
      {renderColor('Background Colors', bgColors)}
      {renderColor('Typography Colors', textColors)}
      {renderColor('Border Colors', borderColors)}
    </Grid>
);
```

In general, we should avoid using index as a key bc it's not very unique. https://stackoverflow.com/a/46735689

In this case, the arrays never update so using index as a key is technically fine, but using alias as a key would be better

```typescript
const createSwatch = (color: string, alias: string) => {
    return (
      <Grid
        item
        className={classes.swatchWrapper}
        key={alias}
        xs={12}
        sm={6}
        md={4}
      >
        <div
          className={classes.swatch}
          style={{ backgroundColor: color }}
        ></div>
        <Typography variant="body1">
          <span className={classes.alias}>{alias}</span>
          <br />
          <span className={classes.color}>{color}</span>
        </Typography>
      </Grid>
    );
  };
```

HanaXu (approved):
The color palette looks amazing, great job cleaning things up! Left some non-blocking clean up comments. 🚀 🌔 

Something else to look into for the future is a single source of truth for the colors. Right now we have colors defined in the theme files and also in `ColorPalette`.

tiffwong:
@bnussman I would have to look into how to do that but I would prefer to address it in a separate PR since this is quite a large PR as it is and there's some more work I want to address in the future

packages/manager/src/components/ColorPalette/ColorPalette.tsx
tiffwong: I compromised at `theme.spacing(4.5)` lol

tiffwong:
@HanaXu agreed, definitely would prefer if the component was able to pull in the colors from the theme without this but that'll take some research. There's definitely more cleanup I'd like to do in regards to the color but this is a good stopping point before deploying Storybook haha

bnussman (approved):
Clean up looks great and Storybook story looks awesome! 🎉 Found no issues with the colors in Cloud Manager

PR #8226, opened 2022-02-24
hotfix: Add logic to the Tax Collection Banner to check for regions/states

tiffwong:
## Description

This adds logic to display the banner conditionally if a list of regions is provided in LaunchDarkly.

## How to test

- Go to `/account/billing`
- Change the country to "Japan" and the date should say "1 April 2022"
- Change the country to "Norway" and the date should say "1 May 2022"
- Change the country to "Canada" and the date should say "May 1, 2022"
- Change the country to "United States", the state to either "Arizona", "Hawaii", "Pennsylvania", "Texas", or "Washington" and the date should say "May 1, 2022"
  - The banner should not show for any other states
- The banner should not display for any countries not mentioned above


packages/manager/src/features/GlobalNotifications/TaxCollectionBanner.tsx
```diff
@@ -30,12 +30,27 @@ const TaxCollectionBanner: React.FC<{}> = () => {
...
+   * If bannerRegions is empty, display the banner for everyone in the country
+   * since everyone will be taxed the same.
+   */
+  const isEntireCountryTaxable = bannerRegions.length === 0;
+
+  /**
+   * If bannerRegions is not empty, only display the banner for customers
+   * whose region is included in the list.
+   */
+  const isCertainRegionsOfCountryTaxable =
```
tiffwong, line 50: Open to suggestions for better variable names

DevDW: I would go with something like `isUserInTaxableRegion` since that's what the logic is really checking for here, not just whether certain regions within the country are taxable.

DevDW (approved):
LGTM, tested all of the cases laid out and saw the banner only in the cases I expected to

HanaXu (commented):
Should we branch from and point to staging?

tiffwong:
Closing this to open a new clean PR pointing from staging

PR #8229, opened 2022-02-25
M3-5739: Apply branding to Storybook

tiffwong:
## Description

This PR uses the `@storybook/theming` package to apply custom theming to Storybook in order for it be more on brand with [cloud.linode.com](cloud.linode.com) and some additional changes include:
- Adding an intro/welcome page
- Switching the order of the canvas and docs tabs

## How to test

Run `yarn storybook` and go to http://localhost:6006/


packages/manager/.storybook/YourTheme.js
```diff
@@ -0,0 +1,36 @@
+// .storybook/YourTheme.js
+import { create } from '@storybook/theming';
+import Logo from '../src/assets/logo/logo.svg';
+
+const brand = {
+  blue: '#3683dc',
+  grey1: '#888f91',
+  grey2: '#3a3f46',
```
HanaXu, line 8: Can we give these better names?

packages/manager/src/components/intro.stories.mdx
```diff
@@ -0,0 +1,5 @@
+import { Meta } from '@storybook/addon-docs';
+
+<Meta title="Welcome" />
```
HanaXu, line 3: Can we rename this to `Intro/Welcome` and add intro as the first element to the storySort order so that welcome is the default page when going to `localhost:6006`?

https://github.com/linode/manager/blob/develop/packages/manager/.storybook/preview.tsx#L48
https://storybook.js.org/docs/react/writing-stories/naming-components-and-hierarchy#sorting-stories

HanaXu (approved):
LGTM, left some minor non-blocking comments 🚢

packages/manager/.storybook/YourTheme.js
tiffwong: @HanaXu I couldn't come up with a better naming convention, do you have any suggestions?

HanaXu: @tiffwong Maybe lightGrey and darkGrey?

tiffwong: @HanaXu my original thought was to leave it open-ended in case we wanted to add more shades of grey but that's something we can address when it comes up 😅

PR #8239, opened 2022-03-01
Prevent preferences call from storybook

codyfinn:
## Description

Prevents storybook from making a call to the preferences endpoint.

## How to test

Run `yarn storybook`
Once it is available to view in your browser open the network tab in dev tools and make sure there are no calls to the preferences endpoint while browsing components.


DevDW (approved):
- [x] No calls to `/preferences` in Storybook
- [x] App & Storybook still look fine, and the theme switching still works
- [ ] Tests passed

HanaXu (approved):
🚢

PR #8240, opened 2022-03-01
Hide broken stories and fix paginated table example

tiffwong:
## Description

Very rough review of the Storybook entries.
Changes include:
- Hide `EntityDetail` and `NotificationDrawer` stories since they're breaking
- Prevent stories from nesting if it's the only story for the component
- Use `LinodeRow` for the paginated table story instead to prevent the content from wrapping

## How to test

- Run `yarn storybook`
- Check the relevant stories


packages/manager/src/components/Table/Table.stories.mdx
```diff
@@ -85,139 +76,77 @@ export const linodes = linodeFactory.buildList(10);
...
+                          backups={linode.backups}
+                          displayStatus={capitalize(linode.status)}
+                          region={linode.region}
+                          status={linode.status}
+                          disk={linode.specs.disk}
+                          memory={linode.specs.memory}
+                          mostRecentBackup={linode.backups.last_successful}
+                          tags={linode.tags}
+                          openTagDrawer={() => null}
+                          type={linode.type}
```
HanaXu, line 127: The plan is showing up as empty, can we hardcode a value?

image

```suggestion
                          type={{ label: 'Linode 2 GB' }}
```

HanaXu (approved):
LGTM, left one non-blocking comment

DevDW (commented):
since the `Drag and Drop` story has no content at the moment, we should hide it

bnussman (approved):
Should `packages/manager/src/components/Placeholder/Placeholder.stories.mdx` be removed? It is empty...

Awesome job updating the Paginationed Table story 🚀

DevDW (approved):
🚢

PR #8241, opened 2022-03-02
Reorganize Storybook structure

tiffwong:
## Description

This PR reorganizes the structure by adding a new section, `Features`, which will contain components that are specific to Linode and can exist on their own.
- `Components` are normally part of a larger feature and are compositions of `Elements` with added logic
- `Elements` has been renamed from `UI Elements` and are essentially our styled version of basic HTML tags
- The previous `Core` section has been removed and divided between the `Intro` and `Elements` section

Also I renamed the first story in a folder to "Default" instead of duplicating the name of the folder.

The next step would be to move the components to the corresponding location in the code library itself which will be addressed in a future PR. For example, `TopMenu`/`SideMenu`/`UserMenu` are in the `/components` folder but `ToastNotifications` is `/features`.

## How to test

Run `yarn storybook`


HanaXu (approved):
🚢

bnussman (approved):

DevDW (approved):

PR #8040, opened 2021-11-11
M3-5473: Webpack 5

bnussman:
## Description

- Use Webpack 5
  - Manager's compile time is now about 20 seconds (was 42 seconds) thanks to `esbuild-loader` instead of `ts-loader`
- Note that Storybook still uses Webpack 4 under the hood. Storybook's support for Webpack 5 is experimental has causes issues when trying to use it.  ([Webpack 5 is considered experimental for Storybook](https://gist.github.com/shilman/8856ea1786dcd247139b47b270912324#storybook-experimental-webpack-5-support))
- Stopped Webpack, Babel, and tsc from annoyingly clearing the console during the `yarn up` process
- Converted to default imports like `import * as [name] from '...';` to `import [name] from '...';`
  - Had to do this to make app compile and had to update some unit test mocking

## Preview

New compile time is now 16 seconds (was 42 seconds) 🚀
Screen Shot 2021-11-11 at 11 26 58 PM

New overlay for errors includes (Typescript error and js compile errors) 🤤
Screen Shot 2021-11-11 at 11 27 56 PM

## How to test

- Test the web pack dev server `yarn up`
- Test a production build

## What broke?
- Nothing 🎉

PR #8243, opened 2022-03-03
Fixed Highlight.js Theme Issues caused by Webpack 5

bnussman:
## Description

Screen Shot 2022-03-03 at 12 57 53 AM

- Highlight.js's theme was stuck in dark mode
  - This resulted in the `View Kubeconfig` drawer displaying the yaml in the wrong theme
- Bug was caused by #8040

## The Fix

- Check against the actual CSS values instead of an inline comment to disable each Highlight.js theme
  - We use to check for a comment in the CSS, but now that we use esbuild, comments are removed and we can't change that

### Before Webpack 5
Screen Shot 2022-03-02 at 11 43 04 PM

### After Webpack 5
Screen Shot 2022-03-02 at 11 46 03 PM

## How to test

- Go to (or create) a Kubernetes Cluster
- View the Kube Config
- Ensure the yaml's theme matches and respects Cloud Manager's theme (Toggle themes with `control` `shift` `d`)
  - A slight delay in the yaml's theme is expected

### DIFF
--- a/packages/manager/src/LinodeThemeWrapper.tsx
+++ b/packages/manager/src/LinodeThemeWrapper.tsx
@@ -6,6 +6,7 @@ import withPreferences, {
   PreferencesActionsProps,
 } from 'src/containers/preferences.container';
 import { dark, light } from 'src/themes';
+import { isProductionBuild } from './constants';
 
 export type ThemeChoice = 'light' | 'dark';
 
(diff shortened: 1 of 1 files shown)


packages/manager/src/LinodeThemeWrapper.tsx
```diff
@@ -31,12 +31,28 @@ const setActiveHighlightTheme = (value: ThemeChoice) => {
...
   links.forEach((thisLink: any) => {
-    const content = thisLink?.textContent ?? '';
-    // We're matching a comment in the style tag's text contents :groan:
-    thisLink.disabled = content.match(inactiveTheme);
+    // Get the inner content of style tag as a text string
+    const content: string = thisLink?.textContent ?? '';
+
+    // If the CSS string contains .hljs and background: #2b2b2b; we can safely assume
+    // we are currently using a11y-dark.css
+    if (content.match('.hljs') && content.match('background: #2b2b2b;')) {
```
bnussman, line 45: The regex matching is pretty loose, but the a11y Highlight.js themes rarely change so this should be reasonable

HanaXu (approved):
🕵️

tiffwong (approved):
Great job getting a fix out for this!

bnussman:
Just realize that there's another problem... This fix only works in local dev because the CSS is not minified in dev. Will work on a fix for production builds too

PR #8244, opened 2022-03-03
M3-5734: Fix status icon alignment in tables

tiffwong:
## Description

This PR fixes the alignment of the status icon with the status text in the `TableCell` component. I added a `statusCell` variant to the component to prevent having to apply the styles every time with the added benefit of cleaning up the styles.

## How to test

Check the alignment for status icons and labels on tables

### DIFF
--- a/packages/manager/src/features/Firewalls/FirewallLanding/FirewallRow.tsx
+++ b/packages/manager/src/features/Firewalls/FirewallLanding/FirewallRow.tsx
@@ -10,7 +10,7 @@ import StatusIcon from 'src/components/StatusIcon';
 import TableCell from 'src/components/TableCell';
 import TableRow from 'src/components/TableRow';
 import useFirewallDevices from 'src/hooks/useFirewallDevices';
-import { capitalize } from 'src/utilities/capitalize';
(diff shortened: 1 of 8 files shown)


bnussman (approved):

packages/manager/src/features/Firewalls/FirewallLanding/FirewallRow.tsx
```diff
@@ -10,7 +10,7 @@ import StatusIcon from 'src/components/StatusIcon';
 import TableCell from 'src/components/TableCell';
 import TableRow from 'src/components/TableRow';
 import useFirewallDevices from 'src/hooks/useFirewallDevices';
-import { capitalize } from 'src/utilities/capitalize';
+import capitalize from 'src/utilities/capitalize';
```
DevDW, line 13: Looks like we do both named and default imports throughout the codebase for this `capitalize` function, which would be nice to standardize at some point (not in-scope for this PR, I may make a separate ticket for it and I'm leaning towards preferring the named import) but this line is a default import vs. in `DatabaseRow.tsx` it is a named import

DevDW (approved):
🚀

PR #8250, opened 2022-03-04
Run Publish Storybook against develop

bnussman:
## Description

Trying to publish storybook from develop

## How to test

Verify this will run against develop

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: ci
 on:
   push:
     branches:
-      - master
+      - develop
   pull_request:
 
 jobs:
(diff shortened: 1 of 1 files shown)


jdamore-linode (approved):
Looks good! This will publish Storybook from `develop` the next time code is pushed to develop (including this PR).

I don't see anything in here that would be problematic if we ran it upon push to `develop` rather than `master` (all the publish steps only run when `master` is the affected branch, etc.)

The only potential caveat here is that we'll need to revert line 6 before our next release otherwise the workflow won't run, but I can't imagine that would be a problem.

DevDW (approved):

PR #8255, opened 2022-03-04
Fix Storybook for Safari and Fix Favicon

bnussman:
## Description

- Removes the attempt to import the service worker at all. This was breaking production storybook for Safari
- Fixes favicon

## How to test

- This will work because I did a production build with Cloudflare Pages and it works now
  - https://b50bbf70.manager-ec2.pages.dev

### DIFF
--- a/packages/manager/.storybook/main.js
+++ b/packages/manager/.storybook/main.js
@@ -13,6 +13,7 @@ module.exports = {
     '@storybook/addon-controls',
     '@storybook/addon-viewport',
   ],
+  staticDirs: ['../public'],
   webpackFinal: (config) => {
     /**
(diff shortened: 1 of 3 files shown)


packages/manager/.storybook/main.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
     '@storybook/addon-controls',
     '@storybook/addon-viewport',
   ],
+  staticDirs: ['../public'],
```
HanaXu, line 16: I'm getting an error when trying to run `yarn storybook` 🤔 

image

DevDW (approved):
LGTM 👍

PR #8256, opened 2022-03-07
M3-5376 Volume create, delete, attach, clone, and resize flows

jdamore-linode:
## Description

**What does this PR do?**
Adds Cypress E2E tests for volume creation, deletion, attach/detach, clone, and resize flows using real data.

There are still some flows that aren't covered (for example, clone and resize via Linode details pages) but I'm hoping this at least puts us in better shape, and that we can improve coverage for other flows with time.

This PR also includes some reorganization of the existing Cypress utility functions, and includes some new utilities to facilitate the use of the `api-v4` package for making API requests. In future PRs I'll begin removing some of the API functions in `cypress/support/api` in favor of their `api-v4` counterparts.

## How to test
The following command can run the new/modified E2E tests to ensure they work as expected:
```bash
yarn cy:e2e --spec "cypress/integration/volumes/*"
```

If you want to confirm that no existing E2E tests were broken by these changes, you can run them all:
```bash
yarn cy:e2e
```


packages/manager/cypress/integration/volumes/create-volume.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+      size: `${randomNumber(10, 250)}`,
+      region: regionId,
+      regionLabel: regionsMap[regionId],
+    };
+
+    cy.defer(createLinode(linodeRequest)).then((linode) => {
+      cy.intercept('POST', '*/volumes').as('createVolume');
+      cy.visitWithLogin('/volumes/create');
+
+      // Fill out and submit volume create form.
```
patthiel, line 95: some of this general form filling stuff is repeated and could be consolidated into a helper/utility function, for easier maintainability

packages/manager/cypress/support/index.js
```diff
@@ -15,16 +15,19 @@
...
 import chaiString from 'chai-string';
+import { authenticate } from './api/authentication';
 import { deleteAllTestData } from './ui/common';
 import 'cypress-real-events/support';
 
 // chai is a global exposed by Cypress which means
 // we can just simply extend it
 chai.use(chaiString);
 
+authenticate();
```
patthiel, line 26: i noticed this in test files as well. Can this not be moved into the before hook? 

```
before(() => {
    authenticate();
    deleteAllTestData();
}
```

then in each test, in theory, won't need a `authenicate();` call

packages/manager/cypress/support/constants/regions.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * @file Constants related to Linode regions.
+ */
+
+/**
+ * Map of region IDs and their corresponding human-readable names.
+ */
+export const regionsMap = {
```
patthiel, line 8: i would source and build this map from https://github.com/linode/manager/blob/3f410d980c966a4c94cd1fe6d2696fb5b7f71f85/packages/manager/src/constants.ts#L131-L153

or just update the references to point to the constants laid out in that file

packages/manager/cypress/support/api/events.ts
```diff
@@ -31,3 +31,24 @@ export const stubLinodeEvent = (linodeData = undefined) => {
   );
   cy.wait('@event');
 };
+
+export const waitForEvent = (eventFilter, attempts = 10) => {
```
patthiel, line 35: this looks neat, have you used this anywhere yet?

packages/manager/cypress/integration/volumes/create-volume.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+      .closest('tr')
+      .within(() => {
+        cy.findByText(volume.label).should('be.visible');
+        cy.findByText(`${volume.size} GB`).should('be.visible');
+        cy.findByText(volume.regionLabel).should('be.visible');
+        cy.findByText('Unattached');
+        fbtClick('Edit');
+      });
+
+    // Confirm that volume tags are set.
```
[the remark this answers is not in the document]
jdamore-linode: Good call!

packages/manager/cypress/support/index.js
jdamore-linode: Unfortunately not -- it doesn't seem that Cypress plays nicely with [Axios interceptors](https://axios-http.com/docs/interceptors) if they're not applied in the spec file where the API calls are made.

packages/manager/cypress/support/api/events.ts
jdamore-linode: Sadly not, and I actually meant to remove this! I was using this in an earlier iteration of one of the tests, and it technically worked, but I found that limiting it by attempts rather than a timeout was kind of trouble prone, but I do think there are places where a function like this could come in handy.

HanaXu (commented):
Can we merge develop in and address conflicts? We'll probably also need to update `create-volume.spec.ts` since tags were removed from the Volume Create flow

PR #8257, opened 2022-03-07
M3-5644: Fix mismatching Marketplace drawers

HanaXu:
## Description
Fix LiteSpeed cPanel drawer opening the cPanel drawer & OpenLiteSpeed Django drawer opening the Django drawer instead of the LiteSpeed versions.

This was due to the `cPanel` and `Django` apps being matched first and returning true before we get to their LiteSpeed versions. So, I updated the find function to find exact matches instead.

## How to test
Go to Marketplace and check that:
- LiteSpeed cPanel opens the LiteSpeed cPanel drawer
- OpenLiteSpeed Django opens the OpenLiteSpeed Django drawer
- Apps with special characters still open their drawers (Minecraft: Java Edition, MySQL/MariaDB, Percona (PPM), Rocket.Chat, WireGuard®)

### DIFF
--- a/packages/manager/src/features/OneClickApps/AppDetailDrawer.tsx
+++ b/packages/manager/src/features/OneClickApps/AppDetailDrawer.tsx
@@ -61,7 +61,7 @@ const styles = (theme: Theme) =>
   });
 
 interface Props {
-  stackscriptID: string;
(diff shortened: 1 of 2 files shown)


jdamore-linode (approved):
Looks good!

PR #8261, opened 2022-03-08
M3-5633: Prevent referral link from being modified

tiffwong:
## Description

This PR disables the referral link textfield to prevent it from being modified.

## How to test

Go to `/profile/referrals` and check the referral link on light and dark mode.


DevDW (approved):
LGTM, in dark mode should the text be white instead of the current gray color though?

Screen Shot 2022-03-08 at 7 49 54 PM

HanaXu (approved):
Just realized that everywhere else that uses the `CopyableTextField` component is also editable, which we probably don't want. Should this be a global change? 🤔

jdamore-linode (approved):

PR #8265, opened 2022-03-09
Patch: restricted users without billing access cannot use the app

johnwcallahan:
## Description

**What does this PR do?**
This adds `accountError` to the dependency array of the effect that "identifies" a user with LD and sets feature flags to "loaded" (thus letting users past the splash screen).

This caused an issue in production where restricted users without billing access (i.e. access to `/account`) could not load the app.

It seems to have been introduced in the Webpack 5 PR, but it is entirely unclear why to me why that is the case. In fact, it's unclear to me why this _ever_ worked (without `accountError` in the dependency array.)

I also changed "retry: false" on the account query so the page would load faster for restricted users. This _may_ have other effects that we should look into.

## How to test

Run the app as a full user and as a restricted user.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 - Early Adopter Program SLA banners to Database Create and empty state Database landing pages for beta
 
+### Fixed:
+
+- Fixed issue where limited users without billing access could not use the app
+
 ## [2022-03-07] - v1.61.0
 
 ### Added:
@@ -20,7 +24,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 ### Changed:
 
 - Drupal Marketplace app logo
-- Feedback link 
+- Feedback link
 - Improve styling consistency for backup auto enrollment
 - Replaced copy icon and added copy tooltip on hover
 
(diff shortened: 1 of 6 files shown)


jdamore-linode (approved):
Looks good! Confirmed that sign in and basic app features work with a normal account and with a restricted account that does not have billing access

DevDW (dismissed):
This fix looks to be working well–the bug is no longer present for a restricted user with no billing access and functionality for an unrestricted user still looks good 🚀

tiffwong (approved):
Logging in as an unrestricted and restricted user both work for me

HanaXu (commented):
Fix LGTM, I don't think we should disable retries across the board for the useAccount query though.

We should retry based on whether the user is restricted (and probably do that for other queries in the future)

So adding a retry parameter to useAccount:
```typescript
export const useAccount = (retry: boolean = true) =>
  useQuery<Account, APIError[]>(queryKey, getAccountInfo, {
    ...queryPresets.oneTimeFetch,
    retry,
  });
```

And updating the useAccountManagement function to:
```typescript
export const useAccountManagement = () => {

  const _isLargeAccount = useSelector(
    (state: ApplicationState) =>
      state.__resources.accountManagement.isLargeAccount
  );

  const _isRestrictedUser = profile?.restricted ?? false;
  const retryAccountQuery = !_isRestrictedUser;

  const { data: accountSettings } = useAccountSettings();
  const { data: account } = useAccount(retryAccountQuery);
...
```

bnussman (commented):
I think whatever change we make to the `/account` endpoint query, we should also make to the `/account/settings` endpoint. They both `403` for restricted users

bnussman (approved):

HanaXu (approved):

DevDW (approved):

PR #8259, opened 2022-03-08
Disable Cypress recording for PR end-to-end tests

jdamore-linode:
## Description

**What does this PR do?**
This disables the recording of Cypress runs for end-to-end tests run on PRs. End-to-end tests that run upon pushing to `develop`, `staging`, and `master`, as well as scheduled tests that run each morning, will continue to be recorded.

We're approaching our monthly limit for test recordings on Cypress.io and thought this would be a workable compromise. If anybody relies on the Cypress.io test recordings for PRs, please let me know!

PR #8268, opened 2022-03-09
Disable Cypress parallelization on PRs while recording is disabled

jdamore-linode:
## Description

**What does this PR do?**
This disables Cypress parallelization for PR E2E tests. This is a temporary measure while Cypress recording is disabled as parallelization is only supported while recording is enabled. In our current state we're getting test failures ([example](https://github.com/linode/manager/runs/5486158088?check_suite_focus=true)).

This will slow down Cypress tests on PRs. Hopefully we'll upgrade our Cypress account soon so that we can reverse this and related changes.

See also PR #8259

### DIFF
--- a/.github/workflows/e2e_pr.yml
+++ b/.github/workflows/e2e_pr.yml
@@ -15,7 +15,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        user: ["USER_1", "USER_2", "USER_3", "USER_4"]
+        # user: ["USER_1", "USER_2", "USER_3", "USER_4"]
+        user: ["USER_1"]
     steps:
       - uses: actions/checkout@v2
         with:
--- a/.github/workflows/e2e_pr.yml
+++ b/.github/workflows/e2e_pr.yml
@@ -59,4 +59,4 @@ jobs:
           browser: chrome
           headless: true
           record: false
-          parallel: true
+          parallel: false


bnussman (approved):
How long will e2es take now? If it's gonna take a long time, I wonder if `fail-fast: true` would be beneficial temporarily assuming all tests are stable (`linode-storage.spec.ts` 🥴)

DevDW (approved):

jdamore-linode:
> How long will e2es take now? If it's gonna take a long time, I wonder if `fail-fast: true` would be beneficial temporarily assuming all tests are stable (`linode-storage.spec.ts` 🥴)

Judging by the speeds I'm seeing on my fork, I'd estimate that tests will take about 40-45 minutes to run without parallelization.

I've also been wondering about the `fail-fast` option! I think it would be helpful to enable it (assuming we can improve the stability of our tests, like you said), but I'll have to do a little more playing around with that to make sure it works as expected. The [Cypress GHA ReadMe](https://github.com/cypress-io/github-action) indicates that enabling `fail-fast` may cause problems with Cypress Dashboard :/

jdamore-linode:
I'm going to close this PR since our Cypress billing cycle resets on Monday. All of these changes (as well as the change made in #8259) will need to be undone, and hopefully in the coming month we'll upgrade our plan so that this doesn't become an issue again.

In the meantime I'll open a new PR to re-enable Cypress recording so that we can get back on track as soon as our billing cycle resets.

PR #8089, opened 2021-12-20
M3-5614: Update stories for Tags

tiffwong:
## Description

Cleanup the stories for Tags and convert the components to FC

## How to test

- Run `yarn storybook` and go to http://localhost:6006/?path=/docs/ui-elements-tags-tag--tag
- Check the app for regressions

PR #8269, opened 2022-03-09
M3-5752: Fixed Tags Panel causing extra PUT requests

bnussman:
## Description

- Fixes Tags panel causing a PUT request on mount
  - Caused by #8089
- Flex Tags panel not allowing you to create a new tag

## How to test

- Text Tags panel functionality against production
- Check to make sure the extra PUT request on mount has stopped occurring


DevDW (approved):
LGTM from the side of the `PUT` request and the issue with creating tags being fixed.

One thing I noticed is that the logic for displaying errors related to tag length and duplication seems to be broken (~line 238). For example, I type in "hm" for the tag and try to add that, I should get an error about the character length. Feel free to try fixing that in this PR if you'd like, but otherwise we should create a ticket to make sure that gets looked at

HanaXu (approved):

PR #8270, opened 2022-03-10
Fixed Select Overflow

bnussman:
## Description

- Fixes `Select` components from having unnecessary x-overflow

> This has been driving me crazy so I had to put this PR up. I swear this use to not be an issues, but somehow it is now. I'm not sure why the extra width was added to the select group CSS, but I just set it to 100% to fix the overflow.

### Before

Screen Shot 2022-03-09 at 10 44 23 PM

### After

Screen Shot 2022-03-09 at 10 44 35 PM

## How to test

- Test `Select` components throughout Cloud Manager
- Test as a restricted user just to be sure 😉😰


tiffwong (approved):
I wasn't able to replicate this on a normal or restricted user but the fix looks good to me

DevDW (approved):
I tested as a restricted user and got hit with the bug from last week which was scary 😧 but it was because the fix hadn't been pulled into this branch, so I merged `develop` in -- LGTM now 👌

PR #8271, opened 2022-03-10
M3-5604: Clean up payment method code

HanaXu:
## Description
- Added PaymentMethodCard component for displaying payment method cards
- Don't report popup closed errors to Sentry
    - There's no exact way to reproduce this. The one time I did get this error, the payment went through anyway. I was also unable to find more info on this specific error so I think we can go ahead and filter these out
- Removed unused classes

## How to test
Go to `/account/billing/make-payment` and ensure the payment methods still display correctly

### DIFF
--- a/packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/PayPalButton.tsx
+++ b/packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/PayPalButton.tsx
@@ -183,7 +183,10 @@ export const PayPalButton: React.FC<Props> = (props) => {
     }
   };
 
-  const onError = (error: Record<string, unknown>) => {
+  const onError = (error: any) => {
+    if (error?.message?.includes('popup close')) {
+      return;
+    }
     reportException(
       'An error occurred when trying to make a one-time PayPal payment.',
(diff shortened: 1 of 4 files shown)


packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/PayPalButton.tsx
```diff
@@ -183,7 +183,10 @@ export const PayPalButton: React.FC<Props> = (props) => {
     }
   };
 
-  const onError = (error: Record<string, unknown>) => {
+  const onError = (error: any) => {
+    if (error?.message?.includes('popup close')) {
+      return;
+    }
```
bnussman, line 189: I looked through Sentry and saw a lot of variations of "closed" related errors. I think what you have is fine because the 'popup close' seemed to be the most frequent one.

Screen Shot 2022-03-14 at 11 06 56 AM

bnussman (approved):
Nice job and thanks for doing the `PaymentMethodCard` component! 🧹

jdamore-linode:
Looks good! :rocket:

I also ran the Cypress account tests to ensure those still pass, and they look good!

PR #8273, opened 2022-03-10
M3-5743: Editable DBaaS cluster labels

DevDW:
## Description
Similar to Linodes, LKE, etc., the labels of DBaaS clusters are now editable from the breadcrumb on the Database Detail page.

## How to test
Please test the success flow and also the error flow by blocking requests to `/database` before attempting to submit the label change.

### DIFF
--- a/packages/manager/src/features/Databases/DatabaseDetail/index.tsx
+++ b/packages/manager/src/features/Databases/DatabaseDetail/index.tsx
@@ -1,22 +1,35 @@
+import { Engine } from '@linode/api-v4/lib/databases/types';
+import { APIError } from '@linode/api-v4/lib/types';
 import * as React from 'react';
+import { matchPath, useHistory, useParams } from 'react-router-dom';
 import Breadcrumb from 'src/components/Breadcrumb';
 import CircleProgress from 'src/components/CircleProgress';
 import TabPanels from 'src/components/core/ReachTabPanels';
 import Tabs from 'src/components/core/ReachTabs';
+import { DocumentTitleSegment } from 'src/components/DocumentTitle';
 import ErrorState from 'src/components/ErrorState';
 import SafeTabPanel from 'src/components/SafeTabPanel';
 import TabLinkList from 'src/components/TabLinkList';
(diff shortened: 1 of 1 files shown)


packages/manager/src/features/Databases/DatabaseDetail/index.tsx
```diff
@@ -81,6 +96,24 @@ export const DatabaseDetail: React.FC = () => {
     history.push(tabs[index].routeName);
   };
 
+  const handleSubmitLabelChange = (newLabel: string) => {
+    return updateDatabase({ allow_list: database.allow_list, label: newLabel })
```
bnussman, line 100: Wow that's unfortunate. I reported this as an API bug. `allow_list` should be optional.

```suggestion
  const handleSubmitLabelChange = (newLabel: string) => {
    // @TODO Update this to only send the label when the API supports it
    return updateDatabase({ allow_list: database.allow_list, label: newLabel })
```

Can we add some kind of comment to remind ourselves to update this at a later date when the API supports an optional `allow_list`?

packages/manager/src/features/Databases/DatabaseDetail/index.tsx
```diff
@@ -114,4 +153,4 @@ export const DatabaseDetail: React.FC = () => {
   );
 };
 
-export default DatabaseDetail;
+export default withEditableLabelState(DatabaseDetail);
```
bnussman, line 156: If you have the time, can you make a hook version of `withEditableLabelState` and use that? It would be nice to use newer patterns in DBaaS

HanaXu (approved):

bnussman (approved):

packages/manager/src/features/Databases/DatabaseDetail/index.tsx
DevDW: I went ahead and created M3-5772 for this since it is used in a few other places and can be replaced there as well. In the meantime I'll go ahead and get this merged 👍

PR #8277, opened 2022-03-15
M3-5495: UI refinements for Volume Create

tiffwong:
## Description

This PR adds contextual help, rearranges the Size fields, makes the Linode dropdown clearable, and removes the Tags section. The width of the fields have also been reduced.

## How to test

There are several type changes to components that are used frequently in Cloud Manager so check the other forms for regressions.


HanaXu (commented):
Should we move the label icon next to the input field (as seen in the label)?

image

HanaXu (changes requested):
There's a bug with the region not clearing if we select a region, then select a linode, and clear the selected linode. However, if we skip selecting a region and select a linode to prefill region, then clear the selected linode, the region clears.

There's also a whitespace issue with the dropdown expanded

https://user-images.githubusercontent.com/14323019/158682531-5d191987-b92c-459d-b527-1a30e221288d.mov

DevDW (approved):
LGTM 👌

bnussman (commented):
In progress of reviewing but have a few comments. 

It would be nice if this paper took up the whole width of the container. This patterns is being used just about everywhere now (Linode Create, Images Create, ...)
Screen Shot 2022-03-18 at 10 15 45 AM

Not really a change but I hope we can find some consistency on this button placement. In some places it's in the paper, in other places it's outside...
Screen Shot 2022-03-18 at 10 19 13 AM

packages/manager/src/components/EnhancedSelect/Select.tsx
```diff
@@ -102,7 +102,7 @@ export interface BaseSelectProps
   /**
    * We require label for accessibility purpose
    */
-  label: string;
+  label: string | JSX.Element;
```
HanaXu, line 105: Do we still need this change?

packages/manager/src/features/Volumes/VolumeCreate/CreateVolumeForm.tsx
```diff
@@ -240,6 +263,18 @@ const CreateVolumeForm: React.FC<CombinedProps> = (props) => {
...
 
+        const handleLinodeChange = (linode: Linode | null) => {
+          if (linode !== null) {
+            setFieldValue('linode_id', linode.id);
+            setFieldValue('region', linode.region);
+            setLinodeId(linode.id);
+          } else {
+            // If the LinodeSelect is cleared, reset the values for Region and Config
+            setFieldValue('linode_id', -1);
+            setLinodeId(initialValueDefaultId);
```
HanaXu, line 274:
```suggestion
            setFieldValue('linode_id', initialValueDefaultId);
            setLinodeId(initialValueDefaultId);
```

HanaXu (approved):
LGTM, left some minor clean up comments 🚢

PR #8282, opened 2022-03-16
M3-5514: Dynamically-shown warning for raw disks in Capture Image tab

DevDW:
## Description
Replace the static `Linode Images cannot be created if you are using raw disks or disks that have been formatted using custom filesystems.` text on the Capture Image tab in the Image Create flow with a warning notice that displays only if the chosen disk is a raw disk. If the warning notice is being displayed, the estimated price also does not display.

## How to test
From the "Capture Image" tab in the Image Create flow, choose a Linode that you have a raw disk added to. Then select that raw disk. The warning notice should display. If you choose a different, non-raw disk, or choose a different Linode, the warning should disappear.


DevDW:
this was accidentally closed

HanaXu (approved):

tiffwong (approved):

PR #8287, opened 2022-03-18
M3-5762: Resolve "try to delete in use disk" E2E test flakiness

jdamore-linode:
## Description

**What does this PR do?**
This reduces or eliminates the test flakiness found in the `try to delete in use disk` test within `linode-storage.spec.ts`.

This test is flaky because navigating to the Linode details page "Storage" tab too quickly after creating a Linode can cause the Linode's disk not to appear (presumably because it has not been created yet), and the disk will not appear until the page is refreshed.

These changes sidestep the issue by waiting for the Linode to boot before navigating to the "Storage" tab. See M3-5762 for more information.

## How to test

```bash
yarn cy:run -s "**/*/linode-storage.spec.ts"
```
Confirm that tests succeed.

Unfortunately, reproducing the issue is difficult and it only occurs once in a while, so it may take some time to confirm that the flakiness is truly resolved.

### DIFF
--- a/packages/manager/cypress/integration/linodes/linode-storage.spec.ts
+++ b/packages/manager/cypress/integration/linodes/linode-storage.spec.ts
@@ -61,18 +61,11 @@ describe('linode storage tab', () => {
       cy.intercept('DELETE', `*/linode/instances/${linode.id}/disks/*`).as(
         'deleteDisk'
       );
-      /* Waiting is necessary because visiting the Linode landing page too
-       * quickly causes the disk not to appear in the `Storage` tab, presumably
-       * because it has not been created yet. This can be improved by waiting
-       * only until an API response confirms that the disk has been created,
-       * or by changing the `Storage` tab to update its contents automatically
-       * without requiring a refresh. In the meantime, this hack should help
-       * reduce flakiness for this test.
-       */
-      // @TODO Remove/improve use of `cy.wait()`. See M3-5762.
-      // eslint-disable-next-line cypress/no-unnecessary-waiting
-      cy.wait(5);
-      cy.visitWithLogin(`/linodes/${linode.id}/storage`);
+      cy.visitWithLogin(`/linodes/${linode.id}`);
+      containsVisible('RUNNING');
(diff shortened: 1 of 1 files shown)


bnussman:
Wow I must be really unlucky because I got the bad behavior first try...
Screen Shot 2022-03-21 at 10 14 12 AM

jdamore-linode:
> Wow I must be really unlucky because I got the bad behavior first try...
> 
> The test ran a second time and passed
> 
> Screen Shot 2022-03-21 at 10 14 12 AM

Ha! That _is_ lucky (or unlucky?). I had the test running on repeat for about 30 minutes on Friday with no sign of the issue!

I'll have to look into this again, but I guess the question is why would the disk not be shown in the "Storage" tab even once the Linode is booted?

I suppose I could modify this so that it waits for the Linode to boot, and then navigates to the "Storage" tab using `cypress.visitWIthLogin()` so that it performs a full page refresh, but we're starting to get into hacky territory with that.

jdamore-linode:
Ok, hoping this really fixes it.

Turns out the data for the "Storage" tab loads as soon as you navigate to the Linode details page, not necessarily when you first navigate to the "Storage" tab itself.

Now I'm waiting for the Linode to boot from the landing page, and navigating to the storage tab after it's reported running. Because there's a full page load in between, and assuming my understanding of this issue is correct, the disk should now always appear in the list for the sake of this test.

jdamore-linode:
> When I run just this test locally, I end up with 4 linodes left on my account created by this one test. Can you confirm those 4 linodes get deleted at some point in the pipeline? I don't want our test accounts getting flooded with dangling Linodes.

Yep! We have hooks set up to clear all the test data _before_ Cypress runs. I'm not certain why we do it this way instead of cleaning up after the tests, but I'm speculating that it's to help with debugging in case a test fails -- we can manually investigate any problematic Linodes/volumes/etc. without having to worry about Cypress deleting everything.

Any thoughts on this?

HanaXu (approved):

PR #8295, opened 2022-03-24
M3-4926: Default values creating access token

codyfinn:
## Description

- [ ] Change default values to `Read/Write` for all
- [ ] When creating a new token the `Select All` radio button for `Read/Write` should be selected

## How to test

1. Login to cloud
2. Navigate to `/profile/tokens`
3. Click the `Create a Personal Access Token` button
4. Ensure that all the radio buttons in the `Read/Write` column are selected including the on in the `Select All` row

 `yarn test APITokens`

### DIFF
--- a/packages/manager/src/features/Profile/APITokens/APITokenDrawer.test.tsx
+++ b/packages/manager/src/features/Profile/APITokens/APITokenDrawer.test.tsx
@@ -1,4 +1,4 @@
-import { fireEvent, within } from '@testing-library/react';
+import { fireEvent, within, waitFor } from '@testing-library/react';
 import * as React from 'react';
 import { APITokenDrawer } from './APITokenDrawer';
 
@@ -59,5 +59,16 @@ describe('API Token Drawer', () => {
     const submit = getByText('Create Token');
     fireEvent.click(submit);
     expect(props.onCreate).toHaveBeenCalledWith('*');
+  });
+  it.only('Should default to read/write for all scopes', async () => {
+    const { findByLabelText } = renderWithTheme(<APITokenDrawer {...props} />);
+    const selectAllReadWriteRadioButton = await findByLabelText(
+      'Select read/write for all'
+    );
+    await waitFor(() => expect(selectAllReadWriteRadioButton).toBeChecked());
+  });
+  it('Should default to never for expiration', () => {
+    // const { getByLabelText } = renderWithTheme(<APITokenDrawer {...props} />);
+    // const experationSelection = getByLabelText('Expiry');
   });
 });
--- a/packages/manager/src/features/Profile/APITokens/APITokenDrawer.tsx
+++ b/packages/manager/src/features/Profile/APITokens/APITokenDrawer.tsx
@@ -30,6 +30,7 @@ import {
   Permission,
   permTuplesToScopeString,
   scopeStringToPermTuples,
+  allScopesAreTheSame,
 } from './utils';
 
 type Expiry = [string, string];
@@ -147,9 +148,10 @@ export class APITokenDrawer extends React.Component<CombinedProps, State> {
   state = {
     scopes: scopeStringToPermTuples(this.props.scopes || '', this.props.perms),
     expiryTups: genExpiryTups(),
-    selectAllSelectedScope: null,
+    selectAllSelectedScope: allScopesAreTheSame(
+      scopeStringToPermTuples(this.props.scopes || '', this.props.perms)
+    ),
   };
-
   /* NB: Upon updating React, port this to getDerivedStateFromProps */
   UNSAFE_componentWillReceiveProps(nextProps: CombinedProps) {
     if (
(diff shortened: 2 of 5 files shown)


packages/manager/src/features/Profile/APITokens/utils.ts
```diff
@@ -160,3 +160,19 @@ export const permTuplesToScopeString = (
...
+    } else if (acc === scope[1] || acc === undefined) {
+      return scope[1];
+    } else {
+      return null;
+    }
+  }, undefined);
+  if (result === undefined) {
+    return null;
+  }
+  return result;
```
codyfinn, line 177: Ideally I would like to refactor this to make it easier to follow

packages/manager/src/features/Profile/APITokens/APITokenDrawer.test.tsx
```diff
@@ -60,4 +60,15 @@ describe('API Token Drawer', () => {
     fireEvent.click(submit);
     expect(props.onCreate).toHaveBeenCalledWith('*');
   });
+  it.only('Should default to read/write for all scopes', async () => {
+    const { findByLabelText } = renderWithTheme(<APITokenDrawer {...props} />);
+    const selectAllReadWriteRadioButton = await findByLabelText(
+      'Select read/write for all'
+    );
+    await waitFor(() => expect(selectAllReadWriteRadioButton).toBeChecked());
```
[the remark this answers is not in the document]
jdamore-linode: Hm... looking at this now.

Not sure if it'll help to illuminate the issue, but it seems that from Jest's point of view the "Select none for all" radio button is checked while the other two are not. Still working to figure out why.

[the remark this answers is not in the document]
jdamore-linode: I haven't actually reviewed any of the code besides the test yet, so I don't have a full understanding of what's going on. But I'm finding that if I add `scopes: '*'` to the `props` object at the top of `APITokenDrawer.test.tsx`, the test passes (additionally, it passes without the need for `async`/`waitFor`). Does that outcome make sense?

packages/manager/src/features/Profile/APITokens/utils.ts
jdamore-linode: What about something like this?

```typescript
/**
 * Determines whether permission scopes all have the same access level.
 *
 * If all scopes have the same access level, the numeric permission level is
 * returned. Otherwise, `null` is returned.
 *
 * @param scopes - Permission scopes for which to check access levels.
 *
 * @returns Access level for the given scopes if they are all the same; `null` otherwise.
 */
export const allScopesAreTheSame = (scopes: Permission[]): number | null => {
  const initialPermissionLevel = scopes[0][1];
  return scopes.reduce((acc, currentScope) => {
    const currentPermissionLevel = currentScope[1];
    if (acc === currentPermissionLevel) {
      return currentPermissionLevel;
    }
    return null;
  }, initialPermissionLevel);
};
```

packages/manager/src/features/Profile/APITokens/APITokenDrawer.test.tsx
```diff
@@ -60,4 +60,15 @@ describe('API Token Drawer', () => {
...
+    const { getByLabelText } = renderWithTheme(<APITokenDrawer {...props} scopes="*"/>);
+    const selectAllReadWriteRadioButton = getByLabelText(
+      'Select read/write for all'
+    );
+    expect(selectAllReadWriteRadioButton).toBeChecked();
+  });
+  it.only('Should default to never for expiration', () => {
+    const { getByText } = renderWithTheme(<APITokenDrawer {...props} scopes="*" expiry="Never" />);
+    const expirySelection = getByText(/Never/i);
+  });
```
codyfinn, line 73: Not sure why the query can't find the text. Also React Select(the library we use to create the dropdown) pulls A Lot of tricks with displaying this data that makes simple queries for the value of the dropdown almost impossible. Not sure how well it would play with a screen reader either.

packages/manager/src/features/Profile/APITokens/utils.ts
jdamore-linode: I know this is unimportant and that this ship has probably sailed, but here's another way to go about this:

```TypeScript
/**
 * Determines whether permission scopes all have the same access level.
 *
 * If all scopes have the same access level, the numeric access level is
 * returned. Otherwise, `null` is returned.
 *
 * @param scopes - Permission scopes for which to check access levels.
 *
 * @returns Access level for the given scopes if they are all the same; `null` otherwise.
 */
export const allScopesAreTheSame = (scopes: Permission[]) => {
  const sample = scopes[0];
  const scopeMatches = (scope: Permission) => scope[1] === sample[1];
  return (scopes.slice(1).every(scopeMatches) ? sample[1] : null);
};
```

I'd be lying if I said this challenge hasn't been in the back of my mind for like the past week lol

jdamore-linode (approved):

packages/manager/src/features/Profile/APITokens/APITokenDrawer.tsx
```diff
@@ -238,7 +252,7 @@ export class APITokenDrawer extends React.Component<CombinedProps, State> {
                 className={classes.noneCell}
               >
                 <Radio
-                  name="Select All"
+                  name="Select Al"
```
bnussman, line 255:
```suggestion
                  name="Select All"
```

bnussman (approved):

PR #8298, opened 2022-03-25
M3-5719: Fixed StackScripts with deprecated Images not showing

bnussman:
## Description

- Shows StackScripts that use deprecated Images
  - Previously we just filtered out StackScripts that used deprecated images and user's StackScripts would not be viewable at all in Cloud Manager
- Adds a new "Deprecated Images" section to the StackScript details page containing any deprecated images the StackScript uses (show in screenshot)
 
Screen Shot 2022-03-25 at 11 04 53 AM

## How to test

- Test viewing account StackScripts
  - Ensure Linode Kubernetes StackScripts do not display when you have LKE clusters deployed
-  Refer to internal ticket

### DIFF
--- a/packages/manager/src/components/StackScript/StackScript.tsx
+++ b/packages/manager/src/components/StackScript/StackScript.tsx
@@ -15,6 +15,8 @@ import ScriptCode from 'src/components/ScriptCode';
 import { useImages } from 'src/hooks/useImages';
 import { useAccountManagement } from 'src/hooks/useAccountManagement';
 import { useReduxLoad } from 'src/hooks/useReduxLoad';
+import HelpIcon from '../HelpIcon';
+import Box from '../core/Box';
 
 const useStyles = makeStyles((theme: Theme) => ({
   root: {
@@ -40,9 +42,8 @@ const useStyles = makeStyles((theme: Theme) => ({
   description: {
     whiteSpace: 'pre-wrap',
   },
-  scriptHeading: {
+  heading: {
     marginBottom: theme.spacing(1),
-    fontSize: '1rem',
   },
   descriptionText: {
     marginBottom: theme.spacing(2),
@@ -73,11 +74,20 @@ const useStyles = makeStyles((theme: Theme) => ({
     display: 'block',
     marginTop: theme.spacing(1),
   },
+  helpIcon: {
+    padding: 0,
+    marginLeft: theme.spacing(),
+  },
 }));
 
 export interface Props {
   data: StackScript;
   userCanModify: boolean;
+}
+
+interface StackScriptImages {
+  available: JSX.Element[];
+  deprecated: JSX.Element[];
 }
 
 type CombinedProps = Props;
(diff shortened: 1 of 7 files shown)


packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
```diff
@@ -220,7 +220,9 @@ export class LinodeCreate extends React.PureComponent<
     this.state = {
       selectedTab: preSelectedTab !== -1 ? preSelectedTab : 0,
       stackScriptSelectedTab:
-        preSelectedTab === 2 && location.search.search('Account') > -1 ? 0 : 1,
+        preSelectedTab === 2 && location.search.search('Community') > -1
+          ? 1
+          : 0,
```
bnussman, line 225: This fixes the Linode Create -> Stackscripts tab from defaulting to "Community"

packages/manager/src/store/image/image.helpers.ts
```diff
@@ -19,17 +19,7 @@ export const filterImagesByType = (
...
-  return Object.keys(images).reduce((acc, eachKey) => {
-    if (!isLinodeKubeImage(images[eachKey])) {
-      acc[eachKey] = images[eachKey];
-    }
-
-    return acc;
-  }, {});
+// Image IDs are in the form linode/debian11
+export const isLinodeKubeImageId = (id: string) => {
+  return id.startsWith('linode/') && Boolean(id.match(/kube/i));
```
bnussman, line 24: Do we think this is sufficient to detect if an Image is a LKE image?

packages/manager/src/components/StackScript/StackScript.tsx
```diff
@@ -193,22 +207,47 @@ export const SStackScript: React.FC<CombinedProps> = (props) => {
...
+                <HelpIcon
+                  text="You must update your StackScript to use a compatible Image to deploy it"
+                  tooltipPosition="bottom"
+                  className={classes.helpIcon}
+                />
+              </Box>
+              {imageChips.deprecated}
+            </>
+          ) : null}
+        </div>
```
codyfinn, line 242: I would rather have the logic the ternary is doing here done by a function that can be tested than in the JSX returned by the component.

packages/manager/src/store/image/image.helpers.ts
HanaXu: I think this is fine for now. We can use the `test` function instead of `match` so that we don't have to cast the result as a boolean

```suggestion
  return id.startsWith('linode/') && /kube/i.test(id);
```

HanaXu (approved):

packages/manager/src/components/StackScript/StackScript.tsx
DevDW: I agree with this as it would tidy up the JSX a bit and make the logic more easily testable, but that can be done in a follow-up PR if you'd like

DevDW (approved):
fixed a couple of typos and I agree with Cody on the ternary, but this looks good to me -- I didn't observe any issues with StackScripts

PR #8299, opened 2022-03-25
M3-5471: Add reference to Gravatar to My Profile settings

tiffwong:
## Description

This PR adds a new section to the Display tab under My Profile to encourage users to using individual profile photos rather than the default user icon. 

Some other UI changes to the Display tab include:
- Moving the other sections into the same `<Paper />` component
- Adding a label and descriptive text to the Timezone selector
- Aligning the buttons to the right to reduce vertical spacing
- Changing the button text to better explain what fields it corresponds to

## How to test

- Go to `/profile/display`
- Create a Gravatar account to update existing profile photo or change the existing photo via the links provided
- Check UI for restricted users

- `yarn test SingleTextFieldForm`
- `yarn test TimezoneForm`

### DIFF
--- a/packages/manager/src/features/Profile/DisplaySettings/DisplaySettings.tsx
+++ b/packages/manager/src/features/Profile/DisplaySettings/DisplaySettings.tsx
@@ -54,6 +54,11 @@ const useStyles = makeStyles((theme: Theme) => ({
     marginLeft: 6,
     padding: 0,
   },
+  tooltip: {
+    '& .MuiTooltip-tooltip': {
+      minWidth: 350,
+    },
+  },
   addImageLink: {
     fontFamily: theme.font.bold,
     fontSize: '1rem',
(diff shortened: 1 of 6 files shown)


HanaXu (commented):
Can we look into using React Query for getting the Gravatar url so that the url will be cached since other components will also need to use the Gravatar url?

DevDW (approved):

tiffwong:
@HanaXu I think that's out of scope for this PR but I can create a follow-up ticket to address that

packages/manager/src/features/Profile/DisplaySettings/DisplaySettings.tsx
```diff
@@ -68,56 +142,105 @@ export const DisplaySettings: React.FC<WithNotifications> = (props) => {
...
+        gravatarURL === undefined ? (
+          <div className={classes.avatar}>
+            <UserIcon />
+          </div>
+        ) : (
+          <div className={classes.avatar}>
+            <img
+              className={classes.gravatar}
+              src={gravatarURL}
+              alt="Gravatar"
```
bnussman, line 171: The image should be rounded

Screen Shot 2022-03-30 at 10 01 49 AM

bnussman (approved):
This looks awesome! 🚀

PR #8300, opened 2022-03-25
M3-5698: Fix Linodes missing from Firewall Linode selection

bnussman:
## The Bug 🐛 

- Prevented users from seeing all of their Linodes in the _Add Linode to Firewall_ drawer
  - Any page that used `<LinodeMultiSelect />` could have been affected by this bug)
- Only occurred **only** when the user navigated directly to a page like `https://cloud.linode.com/firewalls/{id}/linodes` because all Linodes were not in the Redux store 
- Affected user's with **> 100 Linodes** 

## The Fix 🛠 
- Replaces Redux with HOC with React Query to get all of the user's Linodes to display them in the dropdown
 
### Unit Tests
`yarn test LinodeMultiSelect`

### Manual Tests
- Visit a Firewall on a large account (an account with more than 100 Linodes) (`https://cloud.linode.com/firewalls/{id}/linodes`)
- Ensure you see **all** of your Linodes as an option in the _Add Linode to Firewall_ drawer

### DIFF
--- a/packages/manager/src/components/LinodeMultiSelect/LinodeMultiSelect.tsx
+++ b/packages/manager/src/components/LinodeMultiSelect/LinodeMultiSelect.tsx
@@ -42,7 +42,7 @@ export const LinodeMultiSelect: React.FC<Props> = (props) => {
     [allowedRegions, _filteredLinodes, linodes]
   );
 
-  const linodeError = error && error[0]?.reason;
+  const linodeError = error?.[0]?.reason;
 
   const [selectAll, toggleSelectAll] = React.useState<boolean>(false);
 
--- a/packages/manager/src/components/LinodeMultiSelect/LinodeMultiSelect.tsx
+++ b/packages/manager/src/components/LinodeMultiSelect/LinodeMultiSelect.tsx
@@ -27,7 +27,7 @@ export const LinodeMultiSelect: React.FC<Props> = (props) => {
 
   const { data, isLoading, error } = useAllLinodesQuery();
 
-  const linodes = data || [];
+  const linodes = data ?? [];
 
   const _filteredLinodes = filteredLinodes ?? [];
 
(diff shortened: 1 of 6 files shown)


packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/IPSharing.tsx
```diff
@@ -113,7 +113,7 @@ const IPSharingPanel: React.FC<CombinedProps> = (props) => {
     open // Only run the query if the modal is open
   );
 
-  const linodes = Object.values(data?.linodes ?? []);
+  const linodes = data || [];
```
DevDW, line 116: same comment re: `??` vs. `||`

DevDW (approved):

packages/manager/src/queries/linodes.ts
```diff
@@ -29,25 +23,26 @@ export const useLinodesQuery = (
...
+const getAllLinodesRequest = (passedParams: any = {}, passedFilter: any = {}) =>
+  getAll<Linode>((params, filter) =>
+    getLinodes({ ...params, ...passedParams }, { ...filter, ...passedFilter })
+  )().then((data) => data.data);
+
+const getLinodesRequest = (passedParams: any = {}, passedFilter: any = {}) =>
+  getLinodes(passedParams, passedFilter).then((data) => ({
+    linodes: listToItemsByID(data.data),
+    results: data.results,
+  }));
```
codyfinn, line 48: Could these be rewritten as async/await? I generally prefer that syntax but don't know how the rest of the team feels about it.

packages/manager/src/components/LinodeMultiSelect/LinodeMultiSelect.tsx
```diff
@@ -14,37 +13,36 @@ export interface Props extends Partial<BaseSelectProps> {
...
-    linodesError,
-    linodesLoading,
     handleChange,
     showAllOption,
     ...selectProps
   } = props;
 
+  const { data, isLoading, error } = useAllLinodesQuery();
+
+  const linodes = data || [];
```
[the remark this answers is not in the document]
bnussman: I will change this to `??`. 

From DMeechan on StackOverflow
 > The OR operator || uses the right value if left is falsy, while the nullish coalescing operator ?? uses the right value if left is null or undefined.

PR #8301, opened 2022-03-29
Initial commit

tiffwong:
## Description

Scrolling occurs on the Linodes Landing table when there are legacy/long plan names or maintenance scheduled. This PR mitigates the scrolling by moving the inline actions inside the `ActionMenu`. 

In order to prevent confusion regarding which Linode the actions are being applied to, a hover state has been added to the row. I also updated the hover state for the IP address copy icon to be consistent with other clickable icons.

## How to test

Go to `/linodes` and check for scrolling at different viewports with different plan names and statuses.

### DIFF
--- a/packages/manager/src/features/linodes/LinodesLanding/LinodeActionMenu.test.tsx
+++ b/packages/manager/src/features/linodes/LinodesLanding/LinodeActionMenu.test.tsx
@@ -1,16 +1,16 @@
-import * as React from 'react';
 import { screen } from '@testing-library/react';
-import { extendedTypes } from 'src/__data__/ExtendedType';
-import { regionFactory } from 'src/factories/regions';
+import * as React from 'react';
 import { linodeBackupsFactory } from 'src/factories/linodes';
+import { regionFactory } from 'src/factories/regions';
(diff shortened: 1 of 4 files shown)


jdamore-linode:
Still have to take a closer look at different viewports before approving, but pushed a quick change to make the Cypress tests pass in the meantime!

jdamore-linode:
Not sure if it's worth holding up merging, but I was able to get the Linodes table to show a scrollbar when these conditions are met:

* Viewport width is between `960px` and `1035px`
* Side navigation is visible
* A Linode is listed with a long plan name (in my case _Dedicated 32 GB + RTX6000 GPU x1_)

It seems that the scrollbar only appears if all 3 of the above conditions are met.

Screen Shot 2022-03-30 at 10 13 29 AM

I'm approving this anyway since I doubt many people will encounter this and I think these changes are a big improvement overall, but if you make any changes I'd be happy to re-review.

jdamore-linode (approved):

DevDW (approved):
With the mocks on, I'm seeing the scrolling start at <1350px when the sidebar is expanded. The main changes LGTM though

PR #8303, opened 2022-03-30
M3-5412: Events page refinements

HanaXu:
## Description
- Added a GravatarIcon component
- Replaced entity icons with the user's Gravatar in the events page
    - If the user does not have a Gravatar, display a placeholder icon
- Removed 'Duration' column, added 'Relative Date' column, and renamed 'When' column to 'Absolute Date'
- Removed blue line row hover state
- Better accessibility: Show event entity url on hover
- Minor code clean up

## How to test
Check the events page `/events`


jdamore-linode (approved):
Looks good!

packages/manager/src/features/Events/EventRow.tsx
```diff
@@ -117,39 +101,21 @@ export const Row: React.FC<RowProps> = (props) => {
...
       </TableCell>
-
-      <TableCell parentColumn="Duration">
-        <Typography variant="body1">
-          {/* There is currently an API bug where host_reboot event durations are
-          not reported correctly. This patch simply hides the duration. @todo
-          remove this // check when the API bug is fixed. */}
-          {action === 'host_reboot' ? '' : formatEventSeconds(duration)}
-        </Typography>
+      <TableCell parentColumn={'Relative Date'}>
```
tiffwong, line 115:
```suggestion
      <TableCell parentColumn="Relative Date">
```

packages/manager/src/features/Events/EventRow.tsx
```diff
@@ -117,39 +101,21 @@ export const Row: React.FC<RowProps> = (props) => {
...
+        {parseAPIDate(created).toRelative()}
       </TableCell>
-      <TableCell parentColumn={'When'} data-qa-event-created-cell>
+      <TableCell parentColumn={'Absolute Date'} data-qa-event-created-cell>
```
tiffwong, line 118:
```suggestion
      <TableCell parentColumn="Absolute Date" data-qa-event-created-cell>
```

packages/manager/src/features/Events/EventRow.tsx
```diff
@@ -117,39 +101,21 @@ export const Row: React.FC<RowProps> = (props) => {
...
-            </div>
-          </Hidden>
-        </TableCell>
-      )}
+      <TableCell data-qa-event-icon-cell>
+        <Hidden smDown>
+          <GravatarIcon className={classes.icon} />
+        </Hidden>
+      </TableCell>
       <TableCell parentColumn={'Event'} data-qa-event-message-cell>
```
tiffwong, line 110:
```suggestion
      <TableCell parentColumn="Event" data-qa-event-message-cell>
```

tiffwong (commented):
The avatar currently shows the current user's avatar even if the action is done by a different user. Is that the expected behavior?

packages/manager/src/queries/accountUsers.ts
```diff
@@ -16,12 +15,10 @@ export const useAccountUsers = (params: any, withGravatar: boolean = false) =>
     queryPresets.oneTimeFetch
   );
 
-const memoizedGetGravatarURL = memoize(getGravatarUrl);
```
HanaXu, line 19: I think removing this should be fine since React Query handles caching for us and it seems like this is an artifact of before we were using RQ for account users

tiffwong (approved):
Great job lgtm! 🚀

packages/manager/src/utilities/getEventsActionLink.ts
```diff
@@ -19,9 +19,7 @@ export default (
   const label = entity?.label;
 
   if (action.match(/community/gi)) {
-    return () => {
-      window.open(entity!.url, '_blank', 'noopener');
-    };
+    return entity!.url;
```
DevDW, line 22: I don't have any community-related events to test this with at the moment, but will this still open a new tab even without the `window.open(...)` piece that was removed?

packages/manager/src/queries/accountUsers.ts
DevDW: We should definitely keep the memoization here IMO.

If you add a `console.log` statement in

https://github.com/linode/manager/blob/534df6e29a76964850fc60c1a274f796d1a0a4d3/packages/manager/src/utilities/gravatar.ts#L25

the old code makes two logs to the console whereas the current code makes 25, so the function is being called significantly more times when the memoization is removed

packages/manager/src/utilities/getEventsActionLink.ts
HanaXu: Theoretically, it won't open in a new tab. I think this is fine bc that's how it works in the notification drawer anyway and this simplifies the code a lot

bnussman (commented):
On small displays should we drop the first table header item? We also may want to hide one o the date columns so there is more space for the event on small displays

Screen Shot 2022-04-04 at 10 50 02 AM

packages/manager/src/utilities/getEventsActionLink.ts
DevDW: Since the link would lead outside of Cloud, my personal preference as a user would be for it to open in a new tab. But we shouldn't have two different behaviors for the same link in different places

DevDW (commented):
Two other small things I noticed are that in mobile view, the "Events" header is flush with the left edge, and I think we can drop the `Absolute Date` column at a higher viewport width to prevent the text bunching

Screen Shot 2022-04-04 at 1 47 18 PM

looks good aside from this though

DevDW (approved):
Great work! 🚀

PR #8305, opened 2022-04-01
M3-5338: Notification Drawer refinements

tiffwong:
## Description

This PR updates the UI for the notification drawer and it's empty state. The notification drawer is less of a drawer and is more consistent with the user menu. 

**Todo:**
- [x] Refactor `NotificationDrawer` to implement a Menu component (such as from the Reach package) instead of adding conditional logic inside of our `Drawer` component
- [x] Update namings from drawer to menu

**Tasks that will be addressed in a future PR:**
- [ ] Bold the username, action and entity name in the event messages inside the notification drawer

## How to test

Check the notification events view:
- Notifications but a few events (< 5 events)
- Notifications but a lot of events (turn on mocks)
- No notifications and no events (comment out code inside of `serverHandlers.ts`)

Test menu opening/closing:
- Opening the menu via another action other than the bell icon (such as clicking on a Linode's status when it is booting)
- Toggling the menu by double clicking the bell icon
- Clicking inside the menu should not close the menu
    - Unless you clicked a link inside the menu (such as `View all events`), then the menu should close

Other:
- Browsers such as Firefox and Safari
- Mobile view

### DIFF
--- a/packages/manager/src/components/Drawer/Drawer.tsx
+++ b/packages/manager/src/components/Drawer/Drawer.tsx
@@ -9,7 +9,6 @@ import { convertForAria } from 'src/components/TabLink/TabLink';
 
 export interface Props extends DrawerProps {
   title: string;
-  isNotificationDrawer?: boolean;
   wide?: boolean;
 }
 
@@ -54,9 +53,6 @@ const useStyles = makeStyles((theme: Theme) => ({
   backDrop: {
     backgroundColor: theme.color.drawerBackdrop,
   },
-  visuallyHideBackdrop: {
-    backgroundColor: 'rgba(255, 255, 255, 0)',
-  },
 }));
 
 type CombinedProps = Props;
@@ -64,14 +60,7 @@ type CombinedProps = Props;
 const DDrawer: React.FC<CombinedProps> = (props) => {
   const classes = useStyles();
 
-  const {
-    title,
-    children,
-    isNotificationDrawer,
-    onClose,
-    wide,
-    ...rest
-  } = props;
+  const { title, children, onClose, wide, ...rest } = props;
 
   const titleID = convertForAria(title);
 
@@ -80,19 +69,14 @@ const DDrawer: React.FC<CombinedProps> = (props) => {
       anchor="right"
       classes={{ paper: `${classes.paper} ${wide ? classes.wide : ''}` }}
       onClose={(event, reason) => {
-        if (
-          (onClose && reason !== 'backdropClick') ||
-          (onClose && isNotificationDrawer)
-        ) {
+        if (onClose && reason !== 'backdropClick') {
           onClose(event, reason);
         }
       }}
       {...rest}
       ModalProps={{
         BackdropProps: {
-          className: isNotificationDrawer
-            ? classes.visuallyHideBackdrop
-            : classes.backDrop,
+          className: classes.backDrop,
         },
       }}
       aria-labelledby={titleID}
@@ -107,31 +91,27 @@ const DDrawer: React.FC<CombinedProps> = (props) => {
         justifyContent="space-between"
         updateFor={[title, classes]}
       >
-        {!isNotificationDrawer ? (
-          <>
-            <Grid item>
-              <Typography
-                id={titleID}
-                variant="h2"
-                data-qa-drawer-title={title}
-                data-testid="drawer-title"
-              >
-                {title}
-              </Typography>
-            </Grid>
-            <Grid item>
-              <Button
-                className={classes.button}
-                buttonType="secondary"
-                onClick={props.onClose as (e: any) => void}
-                aria-label="Close drawer"
-                data-qa-close-drawer
-              >
-                <Close />
-              </Button>
-            </Grid>
-          </>
-        ) : null}
+        <Grid item>
+          <Typography
+            id={titleID}
+            variant="h2"
+            data-qa-drawer-title={title}
+            data-testid="drawer-title"
+          >
+            {title}
+          </Typography>
+        </Grid>
+        <Grid item>
+          <Button
+            className={classes.button}
+            buttonType="secondary"
+            onClick={props.onClose as (e: any) => void}
+            aria-label="Close drawer"
+            data-qa-close-drawer
+          >
+            <Close />
+          </Button>
+        </Grid>
       </Grid>
       {children}
     </Drawer>
--- a/packages/manager/src/components/Drawer/Drawer.tsx
+++ b/packages/manager/src/components/Drawer/Drawer.tsx
@@ -77,7 +77,7 @@ const DDrawer: React.FC<CombinedProps> = (props) => {
 
   return (
     <Drawer
-      anchor={isNotificationDrawer ? 'top' : 'right'}
+      anchor="right"
       classes={{ paper: `${classes.paper} ${wide ? classes.wide : ''}` }}
       onClose={(event, reason) => {
         if (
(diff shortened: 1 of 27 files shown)


packages/manager/src/components/Drawer/Drawer.tsx
```diff
@@ -2,133 +2,140 @@ import Close from '@material-ui/icons/Close';
...
   return (
     <Drawer
       anchor="right"
+      classes={{ paper: `${classes.paper} ${wide ? classes.wide : ''}` }}
       onClose={(event, reason) => {
-        if (onClose && reason !== 'backdropClick') {
+        if (
+          (onClose && reason !== 'backdropClick') ||
+          (onClose && isNotificationDrawer)
+        ) {
```
codyfinn, line 86:
```suggestion
        if (onClose && (reason !== 'backdropClick' || isNotificationDrawer)) {
```

packages/manager/src/components/NotificationDrawer/NotificationDrawer.tsx
```diff
@@ -42,12 +49,13 @@ export interface Props {
 }
 
 export const NotificationDrawer: React.FC<Props> = (props) => {
-  const { data, open, onClose } = props;
   const classes = useStyles();
+
+  const { data, open, onClose } = props;
+  const { eventNotifications, formattedNotifications } = data;
```
codyfinn, line 55:
```suggestion
const { data: {eventNotifications, formattedNotifications}, open, onClose } = props;
```

bnussman (commented):
When viewing on `smDown`, it seems like the events section is collapsed by default. Can we have it expanded by default?

https://user-images.githubusercontent.com/6440455/162768872-e09a2e52-c88c-42b7-861f-a904a02128d8.mov

bnussman (commented):
I'm a bit confused by the comment 
> Refactor `NotificationDrawer` to implement MUI's `Drawer` component instead of adding conditional logic inside of our `Drawer` component

Shouldn't we refactor this to use a proper `Menu` component (maybe from `@reach/...`) similar to the way `UserMenu.tsx` does it? I feel like, at this point, we're using a Drawer for something it's not made for.

HanaXu:
@bnussman I had the same thought. I'm actually looking into that rn

packages/manager/src/features/NotificationCenter/NotificationSection.tsx
```diff
@@ -109,48 +113,62 @@ export const NotificationSection: React.FC<Props> = (props) => {
...
+                  {showMoreTarget && (
+                    <Typography variant="body1">
+                      <strong>
+                        <Link
+                          to={showMoreTarget}
+                          onClick={() => {
+                            if (onClose) {
+                              onClose();
+                            }
+                          }}
```
codyfinn, line 142:
```suggestion
                          onClick={() => (onClose?.())}
```

packages/manager/src/features/TopMenu/NotificationButton.tsx
```diff
@@ -21,26 +66,72 @@ export const NotificationButton: React.FC<{}> = (_) => {
...
+      // eslint-disable-next-line
+      document.addEventListener('click', handleAnyClick, true);
+
+      // clean up the event when the component is re-rendered
+      return () => document.removeEventListener('click', handleAnyClick, true);
+    } else {
+      return () => null;
+    }
+  }, [handleAnyClick, notificationContext.drawerOpen]);
+
```
HanaXu, line 103: The Reach Menu button exposes the open state to child components, but doesn't let us mutate it (https://reach.tech/menu-button#usemenubuttoncontext).

We need to be able to access the state in other components to programmatically open the menu. So all this is basically a work-around to maintain feature parity

bnussman (commented):
Amazing work on the refactor @HanaXu 🎉 

I'm going to spend some more time looking at this, but would it be possible to dismiss the Menu when the `View all events` is clicked? 

https://user-images.githubusercontent.com/6440455/163015831-9f8fa344-0b9e-4b5f-938d-16494a1e9283.mov

packages/manager/src/features/NotificationCenter/NotificationSection.tsx
```diff
@@ -193,29 +206,31 @@ const ContentBody: React.FC<BodyProps> = React.memo((props) => {
...
-          />
-        </button>
+        <Box display="flex" justifyContent="flex-end">
+          <button
+            className={classes.showMore}
+            onClick={() => setShowAll(!showAll)}
+            aria-label={`Display all ${content.length} items`}
+            data-test-id="showMoreButton"
+          >
+            {showAll ? 'Close' : `${content.length - count} more`}
```
DevDW, line 216: I wonder if we should have it say "Collapse" rather than "Close" 🤔

DevDW (commented):
There are still a handful of files I want to comb through more, but this is looking good so far 👌 

I did notice that the Notification Menu is a bit slow to disappear if you click the User Menu while it is open. The reverse is not true though: the User Menu closes immediately when the Notifications icon is clicked.

https://user-images.githubusercontent.com/32860776/163050919-aacbfa7b-1ff6-4202-aa42-6d4f84072183.mov

packages/manager/src/features/NotificationCenter/NotificationData/RenderNotification.tsx
```diff
@@ -82,11 +75,11 @@ export const RenderNotification: React.FC<Props> = (props) => {
   return (
     <>
       <Grid
-        container
         className={classes.root}
-        data-test-id={notification.type}
+        container
+        alignItems="flex-start"
```
DevDW, line 80: One of the benefits of having the items aligned to the center is that the icons helped as a visual aid for telling where one notification ended and another began when low severity notifications are in the mix. We lose a bit of that now. For example, just glancing at the screenshot, it looks like the "We've updated our policies [...]" notification is part of the preceding one, and "Testing for minor notification" is part of the one that precedes it:

Screen Shot 2022-04-13 at 2 30 00 PM

HanaXu: Spoke with Matthew about this and he said:
> I think the top alignment helps mark the beginning of a notification or event especially for longer messages.

I'm going to leave the icon alignment as is for now and we can revisit this during our next UI/UX review or something

packages/manager/src/features/NotificationCenter/NotificationSection.tsx
HanaXu: The menu might be closing a little bit slower since I had to add a bunch of additional clicking logic, but it's virtually unnoticeable so I think it's fine

packages/manager/src/components/NotificationMenu/NotificationMenu.tsx
```diff
@@ -1,75 +1,55 @@
...
-      marginBottom: theme.spacing(3),
-    },
-  },
-  actionHeader: {
-    display: 'flex',
-    paddingBottom: theme.spacing(),
-    justifyContent: 'flex-end',
-    borderBottom: `solid 1px ${theme.palette.divider}`,
-    marginBottom: theme.spacing(3),
+    padding: 20,
```
bnussman, line 15: Dark mode background color needs to be fixed. This suggested change seems to take care of it. 

Screen Shot 2022-04-15 at 9 59 29 AM



```suggestion
    backgroundColor: theme.bg.bgPaper,
    padding: 20,
```

DevDW (approved):
This LGTM -- I think it'd be worthwhile to look into adding some tests in a follow-up PR for some of the functions with custom click logic

bnussman (approved):
Functionally looks good. Found two behaviors that are not blocking us, but seem odd. 

### Bell icon stays focused when an event link is clicked
https://user-images.githubusercontent.com/6440455/163853800-8120b3cb-a0df-4160-8d11-516a58afdae1.mov

### We should disable links in the activity feed that link to the Linode we are already on

https://user-images.githubusercontent.com/6440455/163854088-4989969c-2759-41d0-a8ec-c1b99de4c076.mov

PR #8307, opened 2022-04-01
Clean up some Cypress support files and functions

jdamore-linode:
## Description
This cleans up some of the code in the `cypress/support` directory in Cloud Manager, primarily removing some redundant constants and functions within `cypress/support/api`. Ultimately I'm hoping to streamline a lot of the code in `cypress/support/api` by relying instead on API functions and types directly from `@linode/api-v4`, and this PR is one small step in that direction.

**What does this PR do?**
- Remove the warning recommending Node versions 10 and 12, replace it with a warning recommending Node 14
- Moves all support functions related to random value generation to `support/util/random.ts`, replaces references to old functions (`makeTestLabel()` and derivatives, `strings.randomPass()`, `strings.randomTitle()`, etc)
- Some import clean-up, replaced some relative imports with absolute imports using TypeScript path aliases

## How to test

**What are the steps to verify the changes?**
Ensure that unit and E2E tests pass. You can either observe the result of GitHub Actions on this PR, or run the tests manually:

```bash
yarn test
yarn cy:run
```


codyfinn (approved):

packages/manager/cypress/integration/firewalls/migrate-linode-with-firewall.spec.ts
```diff
@@ -10,7 +10,7 @@ import {
   containsVisible,
 } from '../../support/helpers';
 import { selectRegionString } from '../../support/ui/constants';
-import { randomTitle } from 'cypress/support/cypresshelpers';
+import { randomString} from 'support/util/random';
```
bnussman, line 13:
```suggestion
import { randomString } from 'support/util/random';
```

packages/manager/cypress/integration/objectStorage/access-keys.spec.ts
```diff
@@ -1,16 +1,16 @@
...
   getClick,
   getVisible,
-} from '../../support/helpers';
+} from 'support/helpers';
+import { randomLabel } from 'support/util/random';
 
 describe('access keys', () => {
   it('create access key', () => {
-    const bucketLabel = 'cy-test-' + makeTestLabel();
+    const bucketLabel = 'cy-test-' + randomLabel();
```
bnussman, line 13: I may be wrong here, but doesn't `randomLabel()` always return a string starting with `cy-test-`? If so, could this be clean up?

```suggestion
    const bucketLabel = randomLabel();
```

bnussman (approved):
Awesome job!

packages/manager/cypress/integration/objectStorage/access-keys.spec.ts
jdamore-linode: Nope, you're absolutely right! That const is definitely being assigned some wacky value like `cy-test-cy-test-fdksfk` -- I meant to circle back and fix that and totally forgot, thank you for catching it!

PR #8309, opened 2022-04-02
Use esbuild for `@linode/api-v4` and `@linode/validation`

bnussman:
## Description

> **Note**: `yarn up` is now **deprecated** with this PR. Use `yarn dev` to develop Cloud Manager

> **Warning**: There's no doubt this has breaking changes. The best we can do it ensure Cloud Manager is not negatively affected by these changes, then we can support users when use `@linode/api-v4` and `@linode/validation` as we get feedback 

- The goal is to make developing, building, and CI faster 💨
- Build `@linode/api-v4` and `@linode/validation` with esbuild
- Use esbuild-jest for unit tests instead of ts-jest
  - Unit tests now take **2 minutes** to run instead of **15 minute**s 🤯🎉
  - No more type checking when running the test, but we invoke `tsc` often anyway and have Typescript in our IDEs, so this should not be a dealbreaker what so ever
- `yarn dev` startup time is much faster
- e2e tests run about 1-3 minutes faster because `yarn up` is quicker
- We now run a dev server for `@linode/validation` so you don't need to run a build for changes to appear in Cloud Manager when running `yarn up`/`yarn dev`
- Due to some `esbuild-jest` issues, we can't use jest.spy so some tests have been skipped
  - In my opinion, I only had to skip a small number of tests so this should not be a blocker for merging 

### Technical Details 🔧
- `@linode/api-v4` and `@linode/validation` targets ES6 instead of ES5 ([babel docs](https://babeljs.io/docs/en/options#no-targets))
  - This means we are dropping support for really older browsers
- `@linode/api-v4` and `@linode/validation` output `commonjs`
- We're no longer shipping a bundled (with dependencies) single `index.js` file
  - This wasn't used by us and I doubt users use or benefit from it


patthiel:
Tests took 5 minutes for me to run locally, which is still a massive improvement. I did however have 1 test fail:

```javascript
 FAIL  src/features/linodes/LinodesDetail/LinodeSummary/ActivitySummary.test.tsx
  ● Test suite failed to run

    TypeError: client.abort is not a function

      at XMLHttpRequestImpl.abort (../../node_modules/jsdom/lib/jsdom/living/xhr/XMLHttpRequest-impl.js:365:14)
      at Object.abort (../../node_modules/jsdom/lib/jsdom/living/xhr/xhr-utils.js:417:13)
      at RequestManager.close (../../node_modules/jsdom/lib/jsdom/browser/resources/request-manager.js:25:21)
      at Window.close (../../node_modules/jsdom/lib/jsdom/browser/Window.js:773:29)
```

bnussman:
Thanks for testing that @patthiel. This change to the unit test loader has caused some flakiness unfortunately so I'll have to take a step back and figure out why. Seems that failures are intermittent somehow

jdamore-linode (approved):

jdamore-linode:
One other thought! Would you be at all agreeable to adding a command that runs Jest but still does type checking at some point? Not to be used often, but just for the sake of occasionally running to make sure no type issues have made their way into the tests.

I was thinking something like `yarn test:types` or similar, and with PR #8353 merged, we could probably use [npm-run-all](https://www.npmjs.com/package/npm-run-all) to run tsc in parallel with Jest.

jdamore-linode:
Ok, one final thing -- this is labeled as `Proof of Concept`, but if we do merge these changes in as-is, we may also want to replace `yarn up` with `yarn dev` in [`.github/workflows/e2e_schedule_and_push.yml`](https://github.com/linode/manager/blob/develop/.github/workflows/e2e_schedule_and_push.yml)

bnussman:
I'm going to look into https://swc.rs/ and see what it offers

jdamore-linode:
I've heard about SWC! Really eager to hear your thoughts

PR #8313, opened 2022-04-05
M3-5574: Update Chip variants

bnussman:
## Description

- Wraps the Material UI Chip component
  - adds `outlined` and `clickable` variant 
- Makes `Upgrade to NVMe` chip the same as the Kubernetes `Upgrade` chip

## How to test

- Check the Chip component in Storybook
- Check all Chip styles in Cloud Manager
  - Use the MSW to help with this

### DIFF
--- a/packages/manager/src/components/PaymentMethodRow/PaymentMethodRow.tsx
+++ b/packages/manager/src/components/PaymentMethodRow/PaymentMethodRow.tsx
@@ -28,9 +28,6 @@ export const useStyles = makeStyles((theme: Theme) => ({
     display: 'flex',
     alignItems: 'center',
   },
-  chip: {
-    fontSize: '0.625rem',
-  },
   container: {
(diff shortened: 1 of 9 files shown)


HanaXu (commented):
We'll also need to update the focus state for the upgrade chip in light/dark mode. To see this state, click the upgrade chip, then click cancel to close the modal

image

bnussman:
I set the focus colors to be the same as the hover color. I don't really know from an accessibility standpoint if that is ok or not. Have any suggestions? @HanaXu

HanaXu:
@bnussman I think that's fine, the focus state also has the dotted outline to differentiate between hover and focus

HanaXu (approved):

DevDW (approved):
LGTM, although I don't think we want `ref` in the Component API args table?

Screen Shot 2022-04-19 at 5 53 57 PM

PR #8322, opened 2022-04-08
M3-5719: Show StackScripts that use deprecated images in Account StackScripts tab

bnussman:
## Description

Take two of #8298

- Shows StackScripts that use deprecated Images in the **Account StackScripts** tab 
  - Previously we just filtered out StackScripts that used deprecated images and user's StackScripts would not be viewable or editable at all in Cloud Manager
  - If a **Community StackScript** does **not** use any non-deprecated images (all of the StasckScripts images are deprecated), it should be hidden in the **Community StackScript** because it is not deployable. 
- Adds a new "Deprecated Images" section to the StackScript details page containing any deprecated images the StackScript uses (show in screenshot)
 
Screen Shot 2022-03-25 at 11 04 53 AM

## How to test

- Test viewing **Account StackScripts**
  - Ensure Linode Kubernetes StackScripts do not display when you have LKE clusters deployed
  - Test viewing an Account StackScript that uses a deprecated image (Refer to internal ticket)
-  Test viewing **Community StackScripts**
   - Behavior should match production exactly
   - You should not see any StackScripts only use deprecated images
- Test Deploying a Linode with the `Deploy New Linode` button on a StackScript
- Test Deploying a Linode by going through the Linode create flow and using the StackScript tab


DevDW (commented):
`alpine3.10` is a deprecated image, but I see the "docker-swarm" Community StackScript appear:

Screen Shot 2022-04-11 at 12 41 00 PM

bnussman:
Good point @DevDW. I took a look and this functionally matches `develop` and appears to be intended behavior. I see this being valuable by allowing deprecated stackscripts to show upon search, but If you disagree, I'd be willing to make a change here.

DevDW:
> Good point @DevDW. I took a look and this functionally matches `develop` and appears to be intended behavior. Less filtering is applied for searches as compared to regular requests. I see this being valuable by allowing deprecated stackscripts to show upon search, but If you disagree, I'd be willing to make a change here.

I think this is fine. I agree. The PR description read as though StackScripts using only deprecated images shouldn't show on the Community tab at all, but search being the exception to that makes sense. I wonder if it'd be beneficial to note this in the tooltip somehow, but that tooltip already has quite a bit of text.

DevDW (approved):
Tested manually in Cloud and looked through the code -- I think the scenarios for testing look good and we're achieving the desired functionality now

codyfinn:
> > Good point @DevDW. I took a look and this functionally matches `develop` and appears to be intended behavior. Less filtering is applied for searches as compared to regular requests. I see this being valuable by allowing deprecated stackscripts to show upon search, but If you disagree, I'd be willing to make a change here.
> 
> I think this is fine. I agree. The PR description read as though StackScripts using only deprecated images shouldn't show on the Community tab at all, but search being the exception to that makes sense. I wonder if it'd be beneficial to note this in the tooltip somehow, but that tooltip already has quite a bit of text.

I think we shouldn't show stackscripts that only have deprecated images. Or if we do we should disable the `Deploy New Linode` button on them since that action is not available.

bnussman:
I looked into disabling the `Deploy New Linode` button, but that would take some bigger changes that I think are out of the scope of this.

codyfinn (approved):

PR #8323, opened 2022-04-08
M3-5776: Cloud Manager doesn't give feedback when resizing a Node Pool over the 100 node limit

codyfinn:
## Description

Fixes a bug with the `EnhancedNumberInput` that would allow numbers above the max to be input. This bug in conjunction with how the component updated its value resulted in truncated large values being submitted. Now when the user tries to input a number larger than the max the component will set the value to the max.

## How to test

1. Navigate to the Kubernetes cluster create page
2. Click a node pool quantity to begin to edit it
3. Type in a number larger than the max (100)
4. Ensure that the value shown in the input does not exceed the max (100)
5. Ensure that when adding that pool to the checkout bar that the value is the max (100)


DevDW (approved):
🚢

bnussman (approved):
🚀

PR #8326, opened 2022-04-12
M3-5282: Linode Detail - Networking - Improved stats empty state

bnussman:
## Description

- Improves the stats unavailable state displayed at `/linodes/{id}/networking` when you create a new Linode
  -  Users were getting confused by the old error state and thought something was wrong with their Linode following it provisioning
- Updates this component to use **React Query**

Screen Shot 2022-04-11 at 8 08 06 PM

## How to test

- Create a new Linode and navigate to its `Networking` tab immediately
- Watch this tab for **7 minutes** and verify that, stats eventually show up in the graph
  - During the first 7 minute period, Cloud Manager should not even hit the Linode API for stats
  - After 7 minutes, you may still get the "stats not ready" message, this may be because the API is still not ready
  - Continue checking that stats eventually show up (should start working after about 10 minutes)
  - Keep in mind, because there is no polling, to get Cloud to re-check for stats, you must refresh, or refocus the window to get react query to refetch
- Verify that all functionality of the `Network Transfer History` graph still works


packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/NetworkingSummaryPanel/TransferHistory.tsx
```diff
@@ -178,19 +191,33 @@ export const TransferHistory: React.FC<TransferHistoryProps> = (props) => {
...
+              ? PendingIcon
+              : undefined
+          }
+          errorText={
+            statsErrorString === STATS_NOT_READY_API_MESSAGE
+              ? STATS_NOT_READY_MESSAGE
+              : statsErrorString
+          }
+          compact
+        />
```
bnussman, line 213: This is a bit messy feeling, but I helps us distinguish a real error from the stats just not being ready.

packages/manager/src/queries/linodes.ts
```diff
@@ -35,6 +49,46 @@ export const useAllLinodesQuery = (
...
+  month: string,
+  linodeCreated?: string
+) => {
+  return useQuery<Stats, APIError[]>(
+    [`${queryKey}-stats`, id, year, month],
+    // If the Linode was created within the last 7 minutes,
+    // mock an API failure so the real API is not actually used.
+    getIsTooEarlyForStats(linodeCreated)
+      ? () => Promise.reject([{ reason: STATS_NOT_READY_MESSAGE }])
+      : () => getLinodeStatsByDate(id, year, month),
```
bnussman, line 72: This approach feels a bit weird, but It allows us to let us use React Query's refetch timer to check if we can hit the API or not. I did this with the goal of making the logic in the components that consume  `useLinodeNetworkStatsByDate` simpler.

HanaXu (commented):
For an existing Linode, CM is still polling stats every 30 seconds. I don't think we should add polling to check if stats are ready since Linodes are only briefly in the provisioning state.

Also, can we fix this artifact by setting overflow-x to `auto` instead of `scroll`?

image

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/NetworkingSummaryPanel/TransferHistory.tsx
codyfinn: I think conditional renders like this would be better served by breaking the components out into their own thing that handles the conditional internally. For getting working code though this is probably the simplest way to do it

DevDW (approved):

HanaXu (approved):

PR #8327, opened 2022-04-13
Show /116s in networking panel

bcmertz:
## Description

Adds back in /116 ranges to a Linode's networking panel. These were accidentally removed when support for routed and shared /64s and /56s was added. This happened because we get our /64 and /56 info from /networking/ipv6/ranges which doesn't surface /116 information

Also adds comment about adding in RDNS info in the future

## Testing

See internal message "Testing https://github.com/linode/manager/pull/8327"


HanaXu (commented):
How do we test/get a 116 range?

bnussman (approved):
I'm seeing the 116 just fine! Is it extected there is no RDNS associated with the range? 

Screen Shot 2022-04-18 at 10 49 20 AM

bcmertz:
@bnussman  yep! that's correct

HanaXu (approved):
Didn't test but logic looks fine to me, approving to move on with release

PR #8329, opened 2022-04-14
Max node pool copy update

codyfinn:
## Description

Changed the copy to reflect the max number of nodes per node pool

## How to test

1. Navigate to the Kubernetes cluster create page
2. Ensure that the copy is correct for the `Add Node Pools` section


DevDW (approved):

HanaXu (approved):

jdamore-linode (approved):
Looks good! Confirmed that I can see "100" in the number field without the text cutting off

tiffwong (approved):

PR #8339, opened 2022-04-19
M3-5748: Use private user attributes for LaunchDarkly fields

codyfinn:
## Description

Makes use of the private user attributes feature of Launch Darkly to limit the collection of PII. In particular the items of `country` and `taxID`. 

## How to test

Instructions listed in ticket

### DIFF
--- a/packages/manager/src/IdentifyUser.tsx
+++ b/packages/manager/src/IdentifyUser.tsx
@@ -77,6 +77,7 @@ export const IdentifyUser: React.FC<{}> = () => {
             custom: {
               taxID: _taxID,
             },
+            privateAttributeNames: ['country', 'taxID'],
           })
           .then(() => setFeatureFlagsLoaded())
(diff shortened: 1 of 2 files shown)


DevDW (commented):
I seem to be running into the same issue @tiffwong and I ran into when we looked into this before -- the `Last Seen` time not being accurate and the attributes not being marked private (specifically `country`) 🤔

tiffwong (approved):
@DevDW it might be related to browser/ad-blocking. I troubleshooted it with @codyfinn and it seems to work with Chrome if you log out and then in from localhost 🤷

DevDW (approved):
I can confirm this works for me in Firefox, but I wasn't able to get it working with my Chrome configuration (tried logging in and out in a regular window and also in Incognito mode). Would be good to get a third set of eyes on this PR for additional testing

PR #8342, opened 2022-04-20
M3-5797: Add alt payment methods API functions/types, factories, and MSW handlers

jdamore-linode:
## Description

**What does this PR do?**
- Adds alt payments functions and types for new API endpoints to the `@linode/api-v4` package
- Adds factories for new types
- Adds MSW handlers for new API endpoints

## How to test

N/A

N/A

### DIFF
--- a/packages/api-v4/src/profile/profile.ts
+++ b/packages/api-v4/src/profile/profile.ts
@@ -146,7 +146,7 @@ export const sendCodeToPhoneNumber = (phoneNumber: string) => {
  * Verifies a one-time password sent using `sendCodeToPhoneNumber`.
  */
 export const verifyPhoneNumberCode = (otpCode: number) => {
-  return Request(
+  return Request<{}>(
     setURL(`${API_ROOT}/profile/phone-number/verify`),
     setMethod('POST'),
     setData({
@@ -174,7 +174,7 @@ export const getSecurityQuestions = () => {
  * questions may be used.
  */
 export const updateSecurityQuestions = (payload: SecurityQuestions) => {
-  return Request(
+  return Request<{}>(
     setURL(`${API_ROOT}/profile/security-questions`),
     setMethod('PUT'),
     setData({
--- a/packages/api-v4/src/profile/profile.ts
+++ b/packages/api-v4/src/profile/profile.ts
@@ -9,7 +9,14 @@ import Request, {
   setXFilter,
 } from '../request';
 import { ResourcePage } from '../types';
-import { Profile, ProfileLogin, TrustedDevice, UserPreferences } from './types';
+import {
+  Profile,
+  ProfileLogin,
+  TrustedDevice,
+  UserPreferences,
+  SecurityQuestions,
+  PhoneNumberVerificationCode,
+} from './types';
 
 /**
  * getProfile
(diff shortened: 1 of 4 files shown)


packages/api-v4/src/profile/profile.ts
```diff
@@ -117,3 +124,61 @@ export const getLogins = (params: any, filter: any) => {
...
 };
+
+/**
+ * sendPhoneNumberCode
+ *
+ * Sends a one-time password via SMS to be used to verify a phone number.
+ */
+export const sendPhoneNumberCode = (phoneNumber: string) => {
+  return Request<PhoneNumberVerificationCode>(
+    setURL(`${API_ROOT}/profile/phone-number`),
```
jdamore-linode, line 135: Should I be using the beta API root for these?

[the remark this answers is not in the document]
bnussman: Non-beta should be correct!

packages/api-v4/src/profile/types.ts
```diff
@@ -21,6 +21,7 @@ export interface Profile {
   authorized_keys: string[];
   two_factor_auth: boolean;
   restricted: boolean;
+  phone_number: string;
```
DevDW, line 24: Do you know how this will be populated for existing users? Wondering if we might need to make this `phone_number?: string` for that case

packages/api-v4/src/profile/profile.ts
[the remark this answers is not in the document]
DevDW: I don't _think_ it's necessary since `/profile` isn't a new endpoint. If this gets changed to `BETA_API_ROOT` (/v4beta) though, if the API doesn't find the endpoint, I believe it just swaps in `API_ROOT` (/v4).

packages/api-v4/src/profile/profile.ts
```diff
@@ -117,3 +124,61 @@ export const getLogins = (params: any, filter: any) => {
     setParams(params)
   );
 };
+
+/**
+ * sendPhoneNumberCode
+ *
+ * Sends a one-time password via SMS to be used to verify a phone number.
+ */
+export const sendPhoneNumberCode = (phoneNumber: string) => {
```
DevDW, line 133: perhaps rename to `sendCodeToPhoneNumber`

packages/api-v4/src/profile/profile.ts
```diff
@@ -117,3 +124,61 @@ export const getLogins = (params: any, filter: any) => {
...
+  );
+};
+
+/**
+ * verifyPhoneNumberCode
+ *
+ * Verifies a one-time password sent using `sendPhoneNumberCode`.
+ */
+export const verifyPhoneNumberCode = (otpCode: number) => {
+  return Request(
```
DevDW, line 149: what's the return type here?

packages/api-v4/src/profile/profile.ts
```diff
@@ -117,3 +124,61 @@ export const getLogins = (params: any, filter: any) => {
...
+};
+
+/**
+ * updateSecurityQuestions
+ *
+ * Updates the current user's security questions. Only expected security
+ * questions may be used.
+ */
+export const updateSecurityQuestions = (payload: SecurityQuestions) => {
+  return Request(
```
DevDW, line 177: what's the return type here? `{}`?

DevDW (commented):
Looks good overall; I imported the methods and turned the MSW on to test. Left a few comments/questions

packages/api-v4/src/profile/types.ts
jdamore-linode: Thanks, good catch! I believe this should be optional -- I recall a conversation where it was specified that `phone_number` will only be populated if the number is verified.

DevDW (approved):
🚢

bnussman (approved):
Looks good, some of this may change as the project progresses, but this should be a great start

PR #8343, opened 2022-04-21
Fix failing firewall E2E test

jdamore-linode:
## Description

**What does this PR do?**
Fixes the failing Cypress test related to firewalls. The tests were failing to select options within an enhanced select; I think Cypress was finding the option elements in the DOM and attempting to interact with them before they were actually visible on-screen.

## How to test

```bash
yarn cy:run -s "cypress/integration/firewalls/migrate-linode-with-firewall.spec.ts"
```

To reproduce the issue, run the above command against `develop` and observe failures (the test doesn't fail every time, but does seem to fail most of the time). To test the fix, run the same command against this PR and confirm that the tests succeed.

### DIFF
--- a/packages/manager/cypress/integration/firewalls/migrate-linode-with-firewall.spec.ts
+++ b/packages/manager/cypress/integration/firewalls/migrate-linode-with-firewall.spec.ts
@@ -192,16 +192,16 @@ describe('Migrate Linode With Firewall', () => {
         fbtClick('Create Firewall');
       });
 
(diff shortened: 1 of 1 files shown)


bnussman (approved):
Passes ✅ I wonder if my work in #8300 has an effect on this?

DevDW (approved):
Ran it against `develop` twice and it passed both times 🤷‍♂️ but it is passing with these changes as well

PR #8345, opened 2022-04-22
feature flagged the beta chips for DBaaS

codyfinn:
## Description

Controls the beta chips for DBaaS using the `databaseBeta` feature flag

## How to test

1. Disable database beta in the dev tools
3. Ensure that the beta chip for databases in the primary nav does not render
3. Navigate to Database Create
4. Ensure that the beta chip in the breadcrumb at the top of the page does not render


DevDW (approved):
LGTM!

In the future with features like this that we know will be in a beta period initially, I'd like to explore structuring the initial feature flag (in this case, `flags.databases`) such that it can include a property in LD indicating whether it is in beta or not. That would potentially allow us to consolidate feature flags.

PR #8349, opened 2022-04-26
M3-5091: Bump postcss version

DevDW:
## Description
We want to upgrade `postcss` to v8.2.10 or higher. This proved to be a bit tricky because it is not a direct dependency and we have several packages that had various different versions as a dependency:

Screen Shot 2022-04-25 at 11 20 04 PM

Even the most recent versions of many of these packages use versions of `postcss` older than v8.2.10.

Before the changes (two instances of `postcss@8.4.6` and one instance of `postcss@7.0.39` listed):
Screen Shot 2022-04-25 at 4 08 13 PM

Given the situation, I opted for [selective dependency resolutions](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/) (also see our ["Managing Dependencies" doc](https://github.com/linode/manager/blob/develop/docs/development-guide/12-manging-dependencies.md)) by adding the desired `postcss` version number to the `resolutions` object in `package.json` and `packages/manager/package.json`. This defines the specified version as an override.

## How to test
Test the app thoroughly to make sure no functionality has been impacted. Additionally, please check the console to make sure no new errors are being logged to the console from this change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,8 @@
     "immer": "^9.0.6",
     "lodash": "^4.17.21",
     "glob-parent": "^5.1.2",
-    "browserslist": "^4.16.5"
+    "browserslist": "^4.16.5",
+    "postcss": "^8.4.6"
   },
   "workspaces": {
     "packages": [
--- a/packages/manager/package.json
+++ b/packages/manager/package.json
@@ -188,7 +188,7 @@
     "chalk": "^2.4.2",
     "circular-dependency-plugin": "^5.2.0",
     "core-js": "^2.6.5",
-    "css-loader": "^6.5.1",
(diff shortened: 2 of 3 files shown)


bnussman (changes requested):
Seems to break Storybook 😖  (screenshot shows the error I get on `yarn storybook`)

Screen Shot 2022-04-27 at 11 27 00 AM

We should probably look into this message on shown on `yarn storybook`. It may be a solution for this.

Screen Shot 2022-04-27 at 11 30 32 AM

DevDW:
As discussed in standup today, going to close this PR out. Listed out the details of steps taken to try to get this working in a comment on the ticket.

PR #8360, opened 2022-04-27
M3-5808: Adjustments to Database Create flow and types due to API changes

DevDW:
## Description
Adjustments to the Database Create flow, Database types, and mocks to reflect forthcoming API changes.

## To-Do:
- [x] Fix various TypeScript types
- [x] Fix issue where changing the DB Engine doesn't update the plan table pricing
- [x] Fix `DatabaseCreate.test.tsx`
- [x] Cleanup (rename some variables, etc.)

## How to test
Go through the Database and Linode Create flows on dev or staging. Ensure that nothing is broken and things behave as you would expect them to.

Please note that the app will crash on dev currently if you choose a MongoDB or PostgreSQL database engine while a plan is selected. This is because the rows have not been populated on dev with data for those engines. Neither DB engine will appear outside of dev presently.

### DIFF
--- a/packages/api-v4/src/databases/types.ts
+++ b/packages/api-v4/src/databases/types.ts
@@ -12,10 +12,13 @@ interface DatabaseClusterSizeObject {
   price: DatabasePriceObject;
 }
 
-export interface DatabaseType extends BaseType {
+interface Engines {
+  [engineType: string]: DatabaseClusterSizeObject[];
+}
+export interface DatabaseType extends Omit<BaseType, 'transfer'> {
   class: DatabaseTypeClass;
-  deprecated: boolean;
-  cluster_size: DatabaseClusterSizeObject[];
+  engines: Engines;
+  transfer?: number;
 }
 
 export type Engine = 'mysql' | 'postgresql' | 'mongodb' | 'redis';
--- a/packages/api-v4/src/databases/types.ts
+++ b/packages/api-v4/src/databases/types.ts
@@ -7,18 +7,15 @@ export interface DatabasePriceObject {
   hourly: number;
 }
 
-interface DatabaseClusterSizeObject {
+export interface DatabaseClusterSizeObject {
   quantity: number;
   price: DatabasePriceObject;
 }
 
-interface Engines {
-  [engineType: string]: DatabaseClusterSizeObject[];
-}
-export interface DatabaseType extends Omit<BaseType, 'transfer'> {
+type Engines = Partial<Record<Engine, DatabaseClusterSizeObject[]>>;
+export interface DatabaseType extends BaseType {
   class: DatabaseTypeClass;
   engines: Engines;
-  transfer?: number;
 }
 
 export type Engine = 'mysql' | 'postgresql' | 'mongodb' | 'redis';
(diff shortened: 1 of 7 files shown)


packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
```diff
@@ -186,6 +190,18 @@ export const SelectPlanPanel: React.FC<CombinedProps> = (props) => {
...
+      ? ((type as unknown) as ExtendedDatabaseType).engines[
+          selectedEngine
+        ].find((cluster: any) => cluster.quantity === 1)
+      : undefined;
+
+    const databaseSubheadings = [
+      `$${databasePrices?.price?.monthly}/mo ($${databasePrices?.price?.hourly}/hr)`,
+      typeLabelDetails(type.memory, type.disk, type.vcpus),
+    ] as [string, string];
+
```
HanaXu, line 204: Really not a fan of this, but refactoring it now might take too long. Can we create a tech debt ticket to refactor/clean this up?

HanaXu (changes requested):
Functionality lgtm 👍 , left some clean up comments

jdamore-linode (approved):
I agree with Hana's suggestions but otherwise think it looks good! Ran the DB and Linode Cypress tests and everything looked good there too

packages/api-v4/src/databases/types.ts
```diff
@@ -12,10 +12,13 @@ interface DatabaseClusterSizeObject {
   price: DatabasePriceObject;
 }
 
-export interface DatabaseType extends BaseType {
+interface Engines {
+  [engineType: string]: DatabaseClusterSizeObject[];
+}
```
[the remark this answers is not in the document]
DevDW: This is arguably cleaner, but is it better from an accuracy POV? If I'm understanding it correctly, `Partial` makes the properties optional, which in this case would be the engineTypes?

packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
DevDW: Is the TypeScript trickery what you're not a fan of? I'm not a fan of it either, we can look into that in the ticket which I'll create later. The logic here is basically the `singleNodePricing` and `subHeadings` logic that used to be in `DatabaseCreate.tsx` though.

packages/api-v4/src/databases/types.ts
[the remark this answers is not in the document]
HanaXu: Good point, I think we can remove the `Partial` bit

packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
HanaXu: Yeah that and calculating `databasePrices` in `DatabaseCreate` and passing that as a prop instead of `selectedEngine` or passing in a function to calculate db prices

HanaXu (approved):
🚢

PR #8363, opened 2022-04-29
Expanding on accessibility and styling docs

tiffwong:


HanaXu (commented):
Is there a difference between MacOs's built-in screen-reader, NVDA, and JAWS? Which one is preferred?

DevDW (approved):

tiffwong:
@HanaXu there is no preferred screen reader since it comes down to a matter of preference but NVDA and JAWS are only available for Windows. Also, I mention in the docs that NVDA is free and open-source while JAWS requires a license if you want to use it for more than 40mins.

HanaXu (approved):

PR #8365, opened 2022-04-29
M3-5796: Bold the Entity Label, Actions, and Username in the Notification Menu

bnussman:
## Description

- Uses markdown to bold words in events in the Notification and Event Menu
- Rewrites an Enzyme test to use newer React Testing Library

Screen Shot 2022-04-29 at 3 32 10 PM

## How to test

- Check Events Menu and Events Landing Table


bnussman:
The more I think about the use of markdown as a solution, the less I like it. I'd like to hear the team's thoughts on this approach

packages/manager/src/features/Events/Event.helpers.ts
```diff
@@ -99,6 +99,6 @@ export const formatEventWithUsername = (
         The event message for Lassie events already includes "by the Lassie Watchdog service",
         so we don't want to add "by Linode" after that.
       */
-      `${maybeRemoveTrailingPeriod(message)} by ${username}.`
+      `${maybeRemoveTrailingPeriod(message)} by **${username}**.`
```
codyfinn, line 102: Maybe a comment here or in the storybook docs to explain how to make words bold using the `**` formatter. Otherwise it is a sort of magical thing that just happens in this one place.

codyfinn:
> The more I think about the use of markdown as a solution, the less I like it. I'd like to hear the team's thoughts on this approach

If the goal is a short term fix to get this formatting in soon with a commitment to add a better pattern for holding certain words later I think it is a simple and effective solution. However as a pattern to use going forward I am less of a fan.

matthewbartholomew:
@bnussman It looks like the approved mockup has a stronger bolding style applied. Please see if that's an option and what the impact might be on our CSS and the UI. Thanks

packages/manager/src/eventMessageGenerator.ts
```diff
@@ -28,160 +28,178 @@ export const eventMessageCreators: { [index: string]: CreatorsForStatus } = {
...
+      )}** is being **added** to Linode **${e.entity!.label}**.`,
     started: (e) =>
-      `${safeSecondaryEntityLabel(e, 'Disk', 'A disk')} is being added to ${
-        e.entity!.label
-      }.`,
+      `**${safeSecondaryEntityLabel(
+        e,
+        'Disk',
+        'A disk'
+      )}** is being **added** to ${e.entity!.label}.`,
```
DevDW, line 79: wouldn't we want to bold the label here too? If not, what are the circumstances under which we want bolded labels vs. regular weight labels? (there are a few other instances where the primary entity's label isn't being bolded)

DevDW (commented):
I think it is an all right pattern, although not necessarily ideal. This is a tricky issue to address.

One thing we can consider with this approach is making an array of standalone words ("canceled", "updated", etc.) that we want to bold in the event messages, and have a function that processes the message text to add the `** **` to those words so it isn't done directly in `eventMessageGenerator.ts` (which could tidy up the file a bit), but I don't think it's a huge deal. Could help in case new events get added and someone forgot to bold the words that should be bolded, though.

packages/manager/src/eventMessageGenerator.ts
```diff
@@ -771,5 +771,48 @@ export default (e: Event): string => {
...
+    'cloned',
+    'enabled',
+    'disabled',
+    'revised',
+    'Upgrade',
+    'failed',
+    'added',
+    'removed',
+    'migrated',
+  ];
```
bnussman, line 801: I may have missed some action words here, If you find any that I may have missed, please let me know!

DevDW: Maybe: `duplicated, reboot(ed), upgraded, rebuilt, reset`?

DevDW (approved):
LGTM 👍

packages/manager/src/eventMessageGenerator.ts
```diff
@@ -771,5 +771,54 @@ export default (e: Event): string => {
...
+    'failed',
+    'added',
+    'removed',
+    'migrated',
+    'duplicated',
+    'reboot',
+    'rebooted',
+    'upgraded',
+    'rebuilt',
+    'reset',
```
HanaXu, line 806: We should add `revoked` to this list

Can we also order this alphabetically so that the action words will be easier to scan in the future?

HanaXu (approved):
Left a minor clean up comment but LGTM 🚀

PR #8369, opened 2022-05-10
M3-5813: Types and related adjustments for PostgreSQL and MongoDB

DevDW:
## Description
- Add new types and tweak existing types to accommodate MongoDB and PostgreSQL
- Tweak mocks, factories, and schema accordingly

## How to test
~~Pointing to prod, test the creation of MongoDB, PostgreSQL, and MySQL databases. Ensure that the pricing is correct and that creation proceeds smoothly and as expected. Double-check the console for errors.~~ Cross-check the types and schema with the spec docs sent to the channel last week.

A few points:
- Mongo and Postgres pricing isn't available in dev right now
- Creating a MySQL and Postgres DB is possible in dev. I'm looking into the Mongo issue right now


packages/validation/src/databases.schema.ts
```diff
@@ -14,9 +14,38 @@ export const createDatabaseSchema = object({
   cluster_size: number()
     .oneOf([1, 3], 'Nodes are required')
     .required('Nodes are required'),
-  replication_type: string()
-    .oneOf(['none', 'semi_synch', 'asynch'])
-    .required('Replication Type is required'),
+  replication_type: string().notRequired().nullable(true),
```
DevDW, line 17: I'm still working through why the `replication_type` validation isn't working as intended. Even with this stopgap measure, Mongo DBs cannot be created.

I'm looking to tweak the commented-out logic below so that it works.

packages/api-v4/src/databases/types.ts
```diff
@@ -132,6 +130,39 @@ export interface Database {
...
+type MongoCompressionType = 'none' | 'snappy' | 'zlib';
+export interface MongoDatabase extends Database {
+  storage_engine: MongoStorageEngine;
+  compression_type: MongoCompressionType;
+}
+
+export type ComprehensiveReplicationType = MySQLReplicationType &
+  PostgresReplicationType;
+
+export type ComprehensiveDatabase = Database &
```
DevDW, line 161: not sure if this is the best name 🤷‍♂️ open to renaming it

packages/api-v4/src/databases/types.ts
```diff
@@ -132,6 +130,39 @@ export interface Database {
   updates: UpdatesSchedule;
 }
 
+export interface MySQLDatabase extends Database {
+  replication_type: MySQLReplicationType;
+}
+
+export type PostgresReplicationType = 'none' | 'synch' | 'asynch';
```
codyfinn, line 137: Given that you want to use these strings in an array in other areas something like this might work better. 
`export const possiblePostgresReplicationTypes = ['none', 'synch', 'asynch'] as const`
`export type PostgresReplicationType = typeof possiblePostgresReplicationTypes[number];`

This way the two are always in sync

packages/api-v4/src/databases/types.ts
codyfinn: Same with the mysql replication type as well

codyfinn (approved):
Left a couple of non-blocking comments. The code works as it is.

packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -174,6 +177,17 @@ const databases = [
       label: `database-${req.params.id}`,
       engine: req.params.engine,
       ssl_connection: true,
+      replication_type:
+        req.params.engine === 'mysql'
+          ? pickRandom(possibleMySQLReplicationTypes)
+          : req.params.engine === 'postgresql'
+          ? pickRandom(possiblePostgresReplicationTypes)
+          : (undefined as any),
```
bnussman, line 185: I removed the cast to `any` and the types seem off here. I don't want this to be a blocker, but something seems off with the `replication_type` of the `ComprehensiveDatabase`

jdamore-linode (approved):

packages/manager/src/mocks/serverHandlers.ts
DevDW: It's a little questionable to me as well. I say let's keep an eye on it to see if it causes any issues. I'll make a note to circle back to it to see if it can be handled more crisply.

PR #8370, opened 2022-05-10
M3-5794: Security Questions component

codyfinn:
## Description

Adds the security questions component for creating and updating account security questions 

## Known Bugs
- Refreshing the page causes the questions the user has defined on their account to not be shown in the form.
- Submitting causes a page crash but successfully sends out the correct PUT request.

## How to test

1. Navigate to the Login and Authentication section of the account page
2. Enable MSW
3. Ensure the Security Questions show up as plain text with an Edit button next to them
4. Ensure the submit button reads 'Update Security Questions' and is disabled
5. Click on one of the Edit buttons
6. Ensure that the question text is replaced by a Select and TextField component
7. Ensure that the previous question text is the default value in the question Select
8. Ensure that the default value of the answer TextField is the answer from the MSW
9. Ensure the submit button is still disabled
10. Click the dropdown for the question Select
11. Ensure the options for the question Select do not contain the question texts of the other questions
12. Click an option
13. Ensure the submit button is now enabled


packages/manager/src/features/Profile/AuthenticationSettings/SecurityQuestions/SecurityQuestions.tsx
```diff
@@ -0,0 +1,164 @@
...
+    initialValues: initalFormValues,
+    onSubmit: async (values) => {
+      try {
+        await updateSecurityQuestions({
+          [values['question-1']]: values['answer-1'],
+          [values['question-2']]: values['answer-2'],
+          [values['question-3']]: values['answer-3'],
+        });
+      } catch (e) {
+        // Do something here I guess
```
DevDW, line 87: We'd likely want to display some sort of error notice if the update fails

DevDW (commented):
Did a bit of testing and glanced over the code – I'm seeing already-selected questions as options for the subsequent questions still (question 2 in the screenshot), and the "Edit" button is a bit misaligned also:

Screen Shot 2022-05-13 at 5 26 12 PM

Screen Shot 2022-05-13 at 5 26 29 PM

packages/manager/src/queries/securityQuestions.ts
```diff
@@ -0,0 +1,63 @@
...
+        security_questions: data.security_questions.map((question) => ({
+          id: question.id,
+          answer: question.answer,
+        })),
+      };
+      return updateUserSecurityQuestions(requestBodyWithoutQuestions);
+    },
+    { onSuccess: updateUserSecurityQuestionsData }
+  );
+};
```
codyfinn, line 63: This was a strange bit of code. The API expects the PUT request to have a request body with the shape of `{ security_questions: [ {id: 1, answer: 'red'}, {id: 2, answer: 'main street'}, ...]}` In order to keep the cache up to date in react query though we need to preserve the question strings themselves in addition to the question IDs. We don't want to save the answers for security reasons so I have to strip them out. I wasn't sure where was the right place to do this data manipulation it felt kinda wrong in the components and feels kinda wrong in the queries

codyfinn: Another option would be to consolidate this data manipulation to the mutation `onsuccess` callback. We could use the cached security questions to get the question strings from the IDs that are in the request body

packages/api-v4/src/profile/types.ts
```diff
@@ -76,4 +76,21 @@ export interface PhoneNumberVerificationCode {
   otp_code: number;
 }
 
-export type SecurityQuestions = Record<string, string>;
+export interface SecurityQuestion {
+  id: number;
+  question: string;
+}
+
+export interface AnsweredSecurityQuesiton extends SecurityQuestion {
```
DevDW, line 84: `AnsweredSecurityQuestion`

DevDW (commented):
Seems like I'm still able to end up in a sort of odd state with the available questions

https://user-images.githubusercontent.com/32860776/173451894-7006be58-9355-4835-a770-0fc2ecf5e5a3.mov

packages/manager/src/features/Profile/AuthenticationSettings/SecurityQuestions/SecurityQuestions.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  const options = securityQuestionsToItems(securityQuestionsData?.security_questions ?? []);
+
+  const initalFormValues = { security_questions: answeredQuestions };
+
+  const formikConfig: FormikConfig<SecurityQuestionsData> = {
+    initialValues: initalFormValues,
+    onSubmit: async (values) => {
+      try {
+        await updateSecurityQuestions(values);
```
bnussman, line 63: Looks like we should be sending `question_id` instead of `id`
Screen Shot 2022-06-16 at 12 02 09 PM

packages/manager/src/features/Profile/AuthenticationSettings/SecurityQuestions/SecurityQuestions.tsx
codyfinn: Yeah toss up an error snackbar probably right?

bnussman:
We need to fix these weird characters 
Screen Shot 2022-06-16 at 2 43 41 PM

We need to filter each select's options so questions are unique
https://user-images.githubusercontent.com/6440455/174143213-92fbdfda-4f99-4340-9804-82f4817b885f.mov

packages/manager/src/features/Profile/AuthenticationSettings/SecurityQuestions/SecurityQuestions.tsx
DevDW: Either an error snackbar or an error notice for the Security Questions section

packages/manager/src/factories/profile.ts
```diff
@@ -34,10 +34,28 @@ export const phoneNumberVerificationCodeFactory = Factory.Sync.makeFactory<Phone
...
-  {
-    'Security question 1': 'Answer 1',
-    'Security question 2': 'Answer 2',
-    'Security question 3': 'Answer 3',
-  }
-);
+export const securityQuestionsFactory = Factory.Sync.makeFactory<
+  SecurityQuestionsData
+>({
+  security_questions: [
```
DevDW, line 40: We can update this array now to just include the six actual ones that are being used, right?

packages/manager/src/features/Profile/AuthenticationSettings/AuthenticationSettings.tsx
```diff
@@ -47,6 +48,7 @@ export const AuthenticationSettings: React.FC = () => {
   const isThirdPartyAuthEnabled = authType !== 'password';
 
   const [success, setSuccess] = React.useState<string | undefined>(undefined);
+  const [areSecuirtyQuestionsAnswered, setAreSecurityQuestionsAnswered] = React.useState(false);
```
DevDW, line 51: `areSecurityQuestionsAnswered`

packages/manager/src/features/Profile/AuthenticationSettings/SecurityQuestions/Answer.tsx
```diff
@@ -0,0 +1,31 @@
...
+  handleChange: any;
+}
+
+const Answer = (props: Props) => {
+  const { isReadOnly, questionResponse, index, handleChange } = props;
+  const label = `Answer ${index + 1}`;
+  const name = `security_questions[${index}].response`;
+
+  if (isReadOnly) {
+    return <></>;
```
DevDW, line 18: What's the rationale for returning an empty fragment instead of `null`?

packages/manager/src/features/Profile/AuthenticationSettings/SecurityQuestions/SecurityQuestions.tsx
```diff
@@ -0,0 +1,164 @@
...
+    onSubmit: async (values) => {
+      try {
+        await updateSecurityQuestions({
+          security_questions: values.security_questions.map((item) => ({
+            question_id: item.id,
+            response: item.response as string,
+          })),
+        });
+        enqueueSnackbar(
+          `Successfully ${buttonCopy.toLowerCase()}ed your security questions`,
```
DevDW, line 83: have to tweak this, right now we're getting "updateed"

Screen Shot 2022-06-16 at 4 37 30 PM

DevDW: Also, after a successful update, I'm unable to do subsequent edits & updates unless I refresh the page:

https://user-images.githubusercontent.com/32860776/174160052-d8bb4e6c-f232-4cfc-9ed3-548d24042f6c.mov

packages/manager/src/features/Profile/AuthenticationSettings/SecurityQuestions/SecurityQuestions.tsx
```diff
@@ -0,0 +1,164 @@
...
+    [theme.breakpoints.down('sm')]: {
+      '& > button': {
+        width: '100%',
+      },
+    },
+  },
+}));
+
+interface Props {
+  setAreSecurityQuestionsAnswered: React.Dispatch<
```
DevDW, line 46: Will we need this?

codyfinn: Yeah the 2FA needs to be disabled and show the Notice instructing the user to answer the security questions to be able to enable 2FA

bnussman: updated this in 4737a4c

jdamore-linode:
I don't think this should block the merge, but the text field / button positioning can look kind of strange in some situations:

Screen Shot 2022-06-17 at 2 34 05 PM

DevDW (approved):

jdamore-linode (approved):
Looks nice! 🎉

bnussman:
I'm going to update some logic that changes how we write to the React Query store. Currently when we write, the order changes unexpectedly upon updating the questions

PR #8378, opened 2022-05-17
M3-5829: Fix Linode Configs Cancel Button

bnussman:
## 🪲 The Bug

- Caused 6 months ago
- Caused when updating code to fix a Material UI deprecation regarding `backdropClick`
- The `onClose` function was not getting passed to our function that generates the modal action buttons

## 📝 Description

- Fixes a flaw in the `ConfirmationDialog.tsx` component by passing `onClose` down to the actions

## 🧪 How to test

- Test action buttons and cancel buttons in Linode dialogs
  - Specifically test the Linode configs deletion dialog

Screen Shot 2022-05-17 at 10 04 38 AM

### DIFF
--- a/packages/manager/src/components/ConfirmationDialog/ConfirmationDialog.tsx
+++ b/packages/manager/src/components/ConfirmationDialog/ConfirmationDialog.tsx
@@ -34,19 +34,17 @@ export interface Props extends DialogProps {
   title: string;
 }
 
-type CombinedProps = Props;
-
-const ConfirmationDialog: React.FC<CombinedProps> = (props) => {
+const ConfirmationDialog: React.FC<Props> = (props) => {
   const classes = useStyles();
 
-  const { title, children, actions, error, onClose, ...dialogProps } = props;
+  const { title, children, actions, error, ...dialogProps } = props;
 
   return (
     <Dialog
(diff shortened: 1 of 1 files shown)


DevDW (commented):
If we take this approach, do we need this anymore?

https://github.com/linode/manager/blob/db006b7614e58132abaf6fc33cb4034e50e9d779/packages/manager/src/features/linodes/LinodesDetail/LinodeAdvanced/LinodeConfigs.tsx#L213

It seems like there are many such cases in the codebase that could be removed if we end up using `dialogProps.onClose()` instead?

bnussman:
@DevDW I'm a bit confused by what your suggesting exactly. The `onClose` is needed so MUI can handle closing the dialog in events such as when the backdrop is clicked or the `esc` key is pressed.

DevDW:
> @DevDW I'm a bit confused by what your suggesting exactly. The `onClose` is needed so MUI can handle closing the dialog in events such as when the backdrop is clicked or the `esc` key is pressed.

yes, my last comment can be disregarded. The discussion about the bug in Slack cleared things up.

DevDW (approved):

HanaXu (approved):

PR #8371, opened 2022-05-11
M3-5805: Remove NVMe Banners

bnussman:
## Description

- Remove NVMe Banners
- Fixes alignment in Volume Row

### Removed Banners
Screen Shot 2022-05-11 at 12 09 29 PM
Screen Shot 2022-05-11 at 12 07 38 PM
Screen Shot 2022-05-11 at 12 07 25 PM

### Fixed Volume Row
Screen Shot 2022-05-11 at 12 08 22 PM
Screen Shot 2022-05-11 at 12 08 38 PM

## How to test

- Check alignment in volume tables
- Ensure volume banners have been removed properly and no critical functionality was accidentally lost

PR #8383, opened 2022-05-19
Fix volumes e2e tests after removing NVMe banners

bnussman:
## Description

- Updates e2e tests for volumes to not use the NVMe banner we removed in #8371 

## How to test

- `yarn cy:run -s "./cypress/integration/volumes/create-volume.smoke.spec.ts"`


jdamore-linode (approved):
Thank you, Banks! Just confirmed that this fixes the test 👍

codyfinn (approved):

PR #8162, opened 2022-02-01
Fix Storybook Dependency Issues

bnussman:
## Description

- Updates Sentry and fixes major dependency issues encountered by @tiffwong and myself
  - Resolves https://github.com/storybookjs/storybook/issues/16837 for Cloud Manager
- Should help us un-block our dependency upgrade tickets

## How to test

- Run `yarn clean`
- Run `yarn`
- Run `yarn up` and check for issues and Typescript errors (especially lookout for react-router related errors)
- Run `yarn storybook` and check Storybook functionality

PR #8385, opened 2022-05-20
Upgrade Storybook to latest

bnussman:
## Description

- Upgrading Storybook because Storybook depends on prismjs
  - Doing this to address dependabot alert about prismjs
- Also just upgrading Storybook to keep it up to date
  - This also gets us off the alpha version of Storybook, which is good (for context of why we were on alpha #8162)

## How to test

- Test all things Storybook related


jdamore-linode:
This is awesome, nice work Banks!

On my system, the tests run in 328 seconds on `develop` and 134 seconds on this branch -- nearly 3 and a half minutes saved!

I did not see the test failure that Pat reported, but I'm not sure if that's because the issue is intermittent or if the changes that have since been merged in from `develop` factor in.

Another interesting observation: without these changes, on my machine Jest exits with this message:

```
A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks.
```

I have been trying to track down the cause of this for weeks without any real progress. But with your changes, I'm not seeing the message. I don't know if this is because of the switch to esbuild, or if one of the tests that is now being skipped is the culprit, but either way I find it very interesting and it makes me want to revisit the issue.

Let me know if you continue having issues with intermittent test failures! I'd be happy to pair up or otherwise help to get to the bottom of it.

codyfinn (approved):
I think you picked up some changes to the cypress tests when you did the merge. Other than that it looks good.

jdamore-linode (approved):
Looks good!

PR #8386, opened 2022-05-20
Allow more sorting options for Databases 

bnussman:
## 📝 Description

- Allows users to sort their Databases by `Configuration` and `Status` in addition to the existing `Label` and `Created`
- We can add this because the API supports it and this table is fully paginated and sorted by the API

## 🔍 Preview 

https://user-images.githubusercontent.com/6440455/169611034-49d0f160-fc86-4fb4-9f9c-337275b9b4ae.mov

## How to test

- Ensure you have a few databases of different types to test with
- Go to `http://localhost:3000/databases`
- Test all sortable columns
  - Test table responsiveness
  - Ensure query params get set properly when a column is changed

### DIFF
--- a/packages/manager/src/features/Databases/DatabaseLanding/DatabaseLanding.tsx
+++ b/packages/manager/src/features/Databases/DatabaseLanding/DatabaseLanding.tsx
@@ -93,7 +93,14 @@ const DatabaseLanding: React.FC = () => {
               Status
             </TableSortCell>
             <Hidden xsDown>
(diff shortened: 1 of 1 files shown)


johnwcallahan (approved):
Nice improvement 👍

DevDW (approved):
Looks good -- in the screen capture in the PR description I noticed a bit of a lag/visual bug with the status icon and "Provisioning" text, but I did not observe that during my own testing

PR #8388, opened 2022-05-25
M3-5814: Refactors in SelectPlanPanel.tsx / DatabaseCreate.tsx

codyfinn:
## Description
Moves the logic for calculating the prices for different DB cluster plans back to the `DatabaseCreate` component. In addition this PR establishes a new `PlanSelectionType` to help standardized the data used to select form different plans. I think more can be done with this but the current iteration is a good first step. I think due to the widespread use of the `SelectPlanPanel` it could be extracted from the `LinodeCreate` structure and generalized to make it easier to possibly create more custom plan selections in the future for new products.

## How to test

Ensure that creation flows for Linodes and Databases work as intended paying attention to pricing and copy. Also ensure that the Linode Resize flow has stayed the same paying close attention to pricing and copy.


packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
```diff
@@ -147,15 +155,14 @@ export const SelectPlanPanel: React.FC<CombinedProps> = (props) => {
...
   } = props;
 
   const classes = useStyles();
 
   const onSelect = (id: string) => () => props.onSelect(id);
 
-  const getDisabledClass = (thisClass: LinodeTypeClass) => {
-    const disabledClasses = props.disabledClasses ?? [];
+  const getDisabledClass = (thisClass: string) => {
+    const disabledClasses = (props.disabledClasses as string[]) ?? []; // Not a big fan of the casting here but it works
```
johnwcallahan, line 165: Would it work to change the type of `PlanSelectionType.class` to `LinodeTypeClass`? Then you don't have to do the casting.

packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
```diff
@@ -225,18 +225,27 @@ const DatabaseCreate: React.FC<{}> = () => {
...
+      const price = clusterPricing?.price || { monthly: null, hourly: null };
+      const subHeadings = [
+        `$${price.monthly}/mo ($${price.hourly}/hr)`,
+        typeLabelDetails(type.memory, type.disk, type.vcpus),
+      ] as [string, string];
       return {
         ...type,
         label: formattedLabel,
         heading: formattedLabel,
+        price: price,
```
johnwcallahan, line 247: Getting a couple of eslint warnings for these. You can use the shorthand here.

Screen Shot 2022-06-02 at 11 04 20 AM

packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx
```diff
@@ -179,9 +181,7 @@ const getEngineOptions = (engines: DatabaseEngine[]) => {
   );
 };
 
-export interface ExtendedDatabaseType extends DatabaseType {
-  heading: string;
-}
+export type DatabasePlanSelectionType = PlanSelectionType;
```
johnwcallahan, line 184: I'd think this type definition is unnecessary; you can just use `PlanSelectionType` in the definition of `displayTypes` below.

packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
```diff
@@ -256,22 +251,12 @@ export const SelectPlanPanel: React.FC<CombinedProps> = (props) => {
...
               </div>
             </TableCell>
-            <TableCell data-qa-monthly>
-              {' '}
-              $
-              {!selectedEngine
-                ? type.price?.monthly
-                : databasePrices?.price.monthly ?? 0}
-            </TableCell>
+            <TableCell data-qa-monthly> $ {type.price?.monthly}</TableCell>
```
johnwcallahan, line 254: There's an extraneous space after the `$`.

johnwcallahan (commented):
Nice work moving the DB types out of the SelectPlanPanel component.

packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
codyfinn: No because it could possibly be a DatabasePlanType. I wanted to leave this as generic as possible so that in the future pulling the component out would be easier. I was trying to avoid defining every possible type of PlanType in this component.

packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
```diff
@@ -316,12 +299,10 @@ export const SelectPlanPanel: React.FC<CombinedProps> = (props) => {
     );
   };
 
-  const renderPlanContainer = (plans: ExtendedTypes) => {
+  const renderPlanContainer = (plans: Array<PlanSelectionType>) => {
```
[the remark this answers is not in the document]
codyfinn: I was just copying what had already been used in this file, but I agree with you that we normally use the bracket notation and prefer it as well. I'll update accordingly.

johnwcallahan:
New changes look good. There's a failing test in GH, but when I run the DatabaseCreate tests locally they pass. Not sure why.. I just re-ran the job.

DevDW (approved):
I left a couple of small comments but this LGTM 👍

jdamore-linode (approved):

PR #8389, opened 2022-05-25
M3-5817: Add support for scoping Database and Firewall permissions

HanaXu:
## Description
Adds Databases to the `Personal Access Token` drawer, which lets the user set smaller scopes such as read_only.

I also went ahead and added Firewall to the drawer since it's a simple change and we'll need to add support for that later anyway.

image

## How to test
- Go to `/profile/tokens` and click on `Create a Personal Access Token`
- You should see `Databases` and `Firewall` in the access table
- Create a token and then click on `View Scopes` for the token you just created. You should see a checkmark for the values you selected when creating the token.
- Make an API request with the token and ensure that the permissions are working as expected


matthewbartholomew:
Make Firewall plural to match.

codyfinn (approved):
🚀

codyfinn (approved):

PR #8393, opened 2022-05-31
Remove notification menu content from DOM when not expanded

jdamore-linode:
## Description

**What does this PR do?**
Since revamping the notification menu (#8305), the notification menu content has been present in the DOM even when hidden.

This PR changes the behavior of the `NotificationButton` component so that the notification menu content is not present in the DOM when the menu is not open.

**Why?**
A few tests are failing or are flaky because `cy.findByText(...)` and similar calls are now selecting elements within the notification menu instead of the elements that were intended. We can (and do) fix these on a case-by-case basis, but it's becoming a game of whack-a-mole as changes to the notification menu and event language continue to cause previously stable tests to break.

## How to test

To reproduce:
1. Checkout develop, run `yarn up`
2. `yarn cy:run`
3. Observe failing/flaky tests

To verify changes:
1. Checkout this branch, run `yarn up`
2. `yarn cy:run`
3. Observe fewer (no?) failing/flaky tests
  (ℹ️ In my case all tests pass, but I did encounter a few flaky tests (e.g. `resizes a linode`) that could fail on subsequent runs)
4. Confirm that notification menu UX is the same as before

### DIFF
--- a/packages/manager/src/components/NotificationMenu/NotificationMenu.tsx
+++ b/packages/manager/src/components/NotificationMenu/NotificationMenu.tsx
@@ -41,7 +41,7 @@ export const NotificationMenu: React.FC<Props> = (props) => {
         dismissNotifications(notifications, { prefix: 'notificationMenu' });
       }
     };
-  }, [open]);
+  }, [dismissNotifications, notifications, dispatch, open]);
 
   return (
     <Paper className={classes.root}>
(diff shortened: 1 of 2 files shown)


HanaXu (commented):
The notification badge doesn't seem to update anymore 🤔 

https://user-images.githubusercontent.com/14323019/171256322-c190f81b-2787-4f7e-bdd1-779e11e8935c.mov

bnussman:
Should probably be fixable by changes in the `NotificationMenu` logic regarding `isOpen` and the `useEffect`

HanaXu (approved):

PR #8396, opened 2022-05-31
Database Create Remove Duplicate State

bnussman:
## Description

- We stored the Engine state in two places on the Database Create page
  - This PR removes one state and pulls the database's engine from Formik's state
 
## How to test

- Test the Database Creation flow for different engines
- Run Database Create Cypress tests


jdamore-linode (approved):
Looks good! Verified that the create flow still works and that the integration tests still pass

HanaXu (approved):

PR #8381, opened 2022-05-17
M3-5795: Phone Verification Component

bnussman:
## 📝 Description

- Adds the phone verification component to the `Login & Authentication` page
- Allows users to verify a phone number and edit it 
- Adds `libphonenumber-js` package to both `manager` and `@linode/validation` for phone number validation and parsing

Screen Shot 2022-06-15 at 11 16 09 AM

## 🧪 How to test

- Turn on the MSW to test the **flow**
- Try adding a phone number to your account at `/profile/auth`

PR #8398, opened 2022-06-01
M3-5842: SMS Messaging Opt Out

bnussman:
## Description 📝

- Allows users to opt out of SMS messaging
- Relates to #8381 
- Essentially, a user is _opted in_ when they have a phone number, and _opted out_ when they don't have a phone number

## Preview 🔍

https://user-images.githubusercontent.com/6440455/171445887-36528660-66dc-456d-bc13-2668ed56317c.mov

## How to test 🧪

- Go to `http://localhost:3000/profile/auth`
- Turn on the MSW
- Try opting out of SMS messaging
  - The flow should be fully functional

### DIFF
--- a/packages/api-v4/src/profile/types.ts
+++ b/packages/api-v4/src/profile/types.ts
@@ -21,7 +21,7 @@ export interface Profile {
   authorized_keys: string[];
   two_factor_auth: boolean;
   restricted: boolean;
-  verified_phone_number?: string | null;
+  verified_phone_number: string | null;
 }
 
 export interface TokenRequest {
--- a/packages/api-v4/src/profile/types.ts
+++ b/packages/api-v4/src/profile/types.ts
@@ -21,7 +21,7 @@ export interface Profile {
   authorized_keys: string[];
   two_factor_auth: boolean;
   restricted: boolean;
-  phone_number?: string;
+  phone_number?: string | null;
 }
 
 export interface TokenRequest {
--- a/packages/api-v4/src/profile/types.ts
+++ b/packages/api-v4/src/profile/types.ts
@@ -21,7 +21,7 @@ export interface Profile {
   authorized_keys: string[];
   two_factor_auth: boolean;
   restricted: boolean;
-  phone_number?: string | null;
+  verified_phone_number?: string | null;
 }
 
 export interface TokenRequest {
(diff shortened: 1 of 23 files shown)


packages/manager/src/features/Profile/AuthenticationSettings/SMSMessaging.tsx
```diff
@@ -0,0 +1,127 @@
...
+            </Button>
+            <Button buttonType="primary" loading={isLoading} onClick={onOptOut}>
+              Opt Out
+            </Button>
+          </ActionsPanel>
+        )}
+      >
+        <Typography>
+          Opting out of SMS messaging will reduce security and limit the ways
+          you can securly access your account.{' '}
```
jdamore-linode, line 110: Just caught this while writing tests! `securly` -> `securely`

I'll come back with a real review for this PR soon, just wanted to point this out before I forgot about it!

bnussman: Thank you! Fixed in b8d99c5

jdamore-linode (approved):
Looking great so far! I know this is still WIP so don't hesitate to request a re-review if you push any big changes 👍

packages/api-v4/src/profile/types.ts
```diff
@@ -21,7 +21,7 @@ export interface Profile {
   authorized_keys: string[];
   two_factor_auth: boolean;
   restricted: boolean;
-  phone_number?: string;
+  verified_phone_number: string | null;
```
bnussman, line 24: Changing this type because it seems the API will **always** return this, but it is nullable

packages/manager/src/features/Profile/AuthenticationSettings/SMSMessaging.tsx
```diff
@@ -0,0 +1,128 @@
...
+          </a>
+        </Typography>
+        <Notice
+          spacingTop={16}
+          spacingBottom={0}
+          className={classes.text}
+          error
+        >
+          <b>Warning:</b> As part of this action, your verified phone number{' '}
+          {profile?.verified_phone_number} will be deleted.
```
DevDW, line 123: Is it possible for us to format the phone number in a more readable way like in the mockups? (if we don't have some existing utility to do this, it can be a follow-up item at a later time)

bnussman: yeah, I'm gonna do this, but I need the `libphonenumber-js` library from #8381

I'll do this once they are both in the feature branch

DevDW (approved):
LGTM 🚢

PR #8403, opened 2022-06-07
M3-5835: Allow users to migrate unattached volumes

bnussman:
## Description 📝

- Allows unattached volumes to be migrated to NVMe instantly by the user
- Refer to internal ticket on what problem this is trying to solve
- This flow is different that an attached volume because it does not associate the action with an entire linode

## Preview 🔍

https://user-images.githubusercontent.com/6440455/172462431-d20492c9-00d7-4f64-b27a-300d32341382.mov

## How to test 🧪

- Turn on MSW, try upgrade flow for the volume called `hdd-unattached`

### DIFF
--- a/packages/manager/src/features/Volumes/UpgradeVolumeDialog.tsx
+++ b/packages/manager/src/features/Volumes/UpgradeVolumeDialog.tsx
@@ -28,7 +28,7 @@ export const VolumeUpgradeCopy = (props: CopyProps) => {
     <Typography>
       {prefix} will be upgraded to high-performance NVMe Block Storage. This is
       a free upgrade and will not incur any additional service charges. Check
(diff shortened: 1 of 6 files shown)


DevDW (changes requested):
We'll need to modify this logic

https://github.com/linode/manager/blob/c21051d23f67621d75b0e86e192a4848d709b642/packages/manager/src/features/Volumes/VolumeTableRow.tsx#L133-L135

so that unattached volumes manually pushed into the migration queue have their chips updated to "Upgrade Pending" under the appropriate conditions

DevDW (approved):
Aside from the typo noted below, I think this should be good to go

jdamore-linode (approved):
Looks good!

PR #8411, opened 2022-06-14
M3-5315: Clean up dismissed notifications more frequently in User Preferences

DevDW:
## Description
Reduce `STALE_DAYS` from 60 to 45 to clean up dismissed notifications more frequently in User Preferences (more discussion in Slack).

Moving forward, we will have to be more vigilant on a few fronts; for example, establishing and adhering to timelines for keeping marketing banners live. Reducing the number of `STALE_DAYS` runs the risk of having previously dismissed notifications/banners reappear sooner in the UI.

## How to test
Dismissal functionality should still work as intended

### DIFF
--- a/packages/manager/src/hooks/useDismissibleNotifications.test.ts
+++ b/packages/manager/src/hooks/useDismissibleNotifications.test.ts
@@ -73,7 +73,7 @@ describe('Dismissible notifications', () => {
       ).toHaveProperty('normal');
     });
 
-    it('should not include notifications created more than 60 days ago', () => {
+    it('should not include notifications created more than 45 days ago', () => {
       expect(
         updateDismissedNotifications(dismissedNotifications, [], {})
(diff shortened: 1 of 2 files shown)


packages/manager/src/hooks/useDismissibleNotifications.test.ts
```diff
@@ -18,7 +18,7 @@ describe('Dismissible notifications', () => {
       ).toBe(true);
     });
 
-    it('Should return false if the notification was created fewer than 60 days ago', () => {
+    it('Should return false if the notification was created fewer than 45 days ago', () => {
```
HanaXu, line 21: We should prob update line 76 as well

image

HanaXu (approved):
LGTM, left one minor comment 🚢

bnussman (approved):

PR #8414, opened 2022-06-15
M3-5652: Display errors in Firewall rules table

HanaXu:
## Description
If there is an error with saving Firewall rules, display those errors in the table. 

Note: There is an edge-case bug where the error displays on the wrong rule (see Jira ticket for more info), that will be addressed in M3-5849.

https://user-images.githubusercontent.com/14323019/173919986-7ba9c633-dff6-422b-9319-863811c7db2d.mov

## How to test
Edit a Firewall rule with an invalid IP and you should see an error displayed after clicking "Save Changes"


DevDW (approved):
This is a definite improvement over not surfacing the error, but the edge case bug you noted is problematic:

Screen Shot 2022-06-16 at 11 46 48 AM

Precluding the possibility of even arriving at this state (hopefully M3-5849 achieves that) would be ideal, and we'll need to test thoroughly at that next stage to ensure that any errors displayed are being displayed on the correct rule.

PR #8418, opened 2022-06-21
(Experimental) Detect excessive Linode APIv4 requests during Cypress tests

jdamore-linode:
## Description

**What does this PR do?**
Tracks the number of Linode APIv4 requests made and prints the total to the Cypress log after each test. It also lays the groundwork to add a maximum request threshold and to begin failing tests if they exceed that value.

You can play with the constants in `request-tracking.ts` to tweak the behavior and thresholds for request tracking. Right now I have the max number of requests set to `null` so that no tests fail, but once we have a better understanding of the number of requests that are typical for our tests we can set an appropriate value for the max.

## How to test

Run manager with `yarn up`, then:
```
yarn cy:debug
```

Run any test, and observe that the total number of Linode APIv4 requests is shown in Cypress after each test runs. If you play with the constants in `request-tracking.ts`, confirm that the behavior matches your expectations.

### DIFF
--- a/packages/manager/cypress/support/index.js
+++ b/packages/manager/cypress/support/index.js
@@ -26,6 +26,7 @@ chai.use(chaiString);
 authenticate();
 
 import './commands';
+import './request-tracking';
 
 // Runs before each test file.
 before(() => {
--- a/packages/manager/cypress/support/request-tracking.ts
+++ b/packages/manager/cypress/support/request-tracking.ts
@@ -1,0 +1,56 @@
+/**
+ * @file Tracks the number of Linode APIv4 requests being made during each test.
+ */
+
+// The number of Linode APIvr requests that are allowed.
+// Set to `null` to allow any number of APIv4 requests.
+const maxRequests: number | null = null;
+
+// Whether a test should fail immediately upon passing `maxRequests` threshold.
+// If `false`, the test will continue running but will fail upon finishing.
+const failImmediately = false;
+
+// Whether or not to report the APIv4 request count at the end of a test.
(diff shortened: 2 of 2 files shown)


patthiel:
very nice idea!

jdamore-linode:
I removed the `Do Not Merge` label -- eager to hear the team's thoughts on this!

HanaXu (commented):
Where exactly do I find the total number of Linode APIv4 requests in Cypress? 🤔

jdamore-linode:
Sorry @HanaXu! Should've been clearer about that -- it only appears if you run Cypress interactively, and it gets shown in the left sidebar where HTTP requests, other log messages, etc., are shown. You may have to wait until all of the tests in a given spec file finish running before you can actually drill down to see it:

Screen Shot 2022-06-29 at 2 13 06 PM

It doesn't appear in the CLI, which is unfortunate :/ but if/when we configure this to fail after a certain number of requests is exceeded, the tests will fail (with a clear explanation) even when invoked via the CLI

HanaXu (approved):
LGTM, nice idea! How exactly would we track the requests number average to determine a typical amount? Randomly checking a couple runs?

bnussman (approved):
Very neat tool to have! Makes me want to try to find ways to cut down the total number of api calls cloud makes 😮‍💨

jdamore-linode:
@HanaXu That's a really good point 🤔 I'm going to see if it would be possible to output a summary after all of the tests run that show the average number of requests made per test or something like that.

PR #8422, opened 2022-06-22
2FA with Security Questions Fixes

bnussman:
## Description 📝

- For existing users with 2FA enabled, we need them to still see the 2FA toggle even if they don't have security questions set
- Pending an API change, a user will not be able to `Reset two-factor authentication` if they don't have security questions setup (shown in screenshots 📷)

## Bug

Screen Shot 2022-06-22 at 8 03 27 AM

### Updated

Screen Shot 2022-06-22 at 11 54 39 AM
Screen Shot 2022-06-22 at 11 55 03 AM
Screen Shot 2022-06-22 at 11 56 12 AM

## How to test 🧪

- Testing 2FA and Security Questions and their various states
  - If an existing user has 2FA enabled but no security questions, they should be able to see the toggle and disable 2fa
  - If an existing user has 2FA enabled but no security questions and they try to click `Reset two-factor authentication`, the API will (not deployed yet) error (shown in screenshots)
  - If the user does not have 2FA setup and has no security questions, they should see a notice saying to set security questions


jdamore-linode (approved):

packages/manager/src/components/Notice/Notice.tsx
```diff
@@ -137,8 +137,8 @@ interface Props extends GridProps {
   className?: string;
   flag?: boolean;
   notificationList?: boolean;
-  spacingTop?: 0 | 8 | 12 | 16 | 24;
-  spacingBottom?: 0 | 8 | 12 | 16 | 20 | 24 | 32;
+  spacingTop?: 0 | 4 | 8 | 12 | 16 | 24;
+  spacingBottom?: 0 | 4 | 8 | 12 | 16 | 20 | 24 | 32;
```
codyfinn, line 141: Not sure 4 is a valid value for these to have. I think our spacing is based off of 8px. I could be wrong though and don't think this blocks the PR

DevDW (approved):
LGTM. Can we request that the API error message have "Two-Factor Authentication" with the hyphen so it matches how we refer to it in Cloud?

PR #8424, opened 2022-06-23
Use SWC for `@linode/validation` and `@linode/api-v4`

bnussman:
## Description 📝

- Putting this up as an alternative to #8309 
- I'm thinking SWC will be **_safer_** to go with because it allows us to compile `@linode/validation` and `@linode/api-v4` to the same format we currently do (`umd` & `ES5`). esbuild does not support this.

## How to test 🧪

> ⚠️ Work in Progress


bnussman:
Using pure SWC would be awesome, but I don't think it's the right direction for us. We prefer features like bundling and dts out of the box and SWC lacks those things. Bundling always threw errors for some reason.

PR #8425, opened 2022-06-23
M3-5462: Keys and secrets modals update UI

codyfinn:
## Description

Adds the ability to download the keys for a token from the text field that displays it.

## How to test

1. Navigate to /profile/tokens
2. Create a new Personal Access Token
3. Ensure that the `TextFields` displaying the keys have both a download and copy `Tooltip` at the end of the `TextField`
4. Click the download `Tooltip`
5. Ensure the file downloaded
    - Contains the string displayed in the `TextField`
    - Is named the label of the `TextField` in snake case

### DIFF
--- a/packages/manager/src/components/CopyableAndDownloadableTextField.tsx
+++ b/packages/manager/src/components/CopyableAndDownloadableTextField.tsx
@@ -25,6 +25,7 @@ const useStyles = makeStyles((theme: Theme) => ({
 type Props = TextFieldProps & {
   className?: string;
   hideIcon?: boolean;
+  fileName?: string;
 };
 
 type CombinedProps = Props;
@@ -34,6 +35,8 @@ export const CopyableAndDownloadableTextField: React.FC<CombinedProps> = (
 ) => {
   const classes = useStyles();
   const { value, className, hideIcon, ...restProps } = props;
+
+  const fileName = props.fileName ?? _.snakeCase(props.label);
 
   return (
     <TextField
@@ -47,7 +50,7 @@ export const CopyableAndDownloadableTextField: React.FC<CombinedProps> = (
             <DownloadTooltip
               text={`${value}`}
               className={classes.copyIcon}
-              fileName={_.snakeCase(props.label)}
+              fileName={fileName}
             />
             <CopyTooltip text={`${value}`} className={classes.copyIcon} />
           </>
--- a/packages/manager/src/components/CopyableAndDownloadableTextField.tsx
+++ b/packages/manager/src/components/CopyableAndDownloadableTextField.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import _ from 'lodash';
 import CopyTooltip from 'src/components/CopyTooltip';
 import DownloadTooltip from 'src/components/DownloadTooltip';
 import { makeStyles, Theme } from 'src/components/core/styles';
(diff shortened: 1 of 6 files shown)


HanaXu (commented):
We'll also need to check the Object Storage Access Keys modal (see below)

The tooltip for the download button says Copy instead of Download and the order is reverse of the mockup
Screen Shot 2022-06-24 at 3 19 41 PM

And for the Object Storage Access Keys modal specifically, would it make more sense to add a secondary download button at the bottom of the modal to download both keys at once instead of adding separate download buttons?

packages/manager/src/components/CopyableAndDownloadableTextField.tsx
```diff
@@ -0,0 +1,61 @@
...
+  props
+) => {
+  const classes = useStyles();
+  const { value, className, hideIcon, ...restProps } = props;
+
+  return (
+    <TextField
+      value={value}
+      {...restProps}
+      className={`${className} ${'copy'} ${classes.removeDisabledStyles}`}
```
DevDW, line 42: I think we can remove the `${' ... '}` wrapping `copy` here

packages/manager/src/features/Profile/SecretTokenDialog/SecretTokenDialog.tsx
```diff
@@ -27,42 +29,64 @@ const useStyles = makeStyles(() => ({
...
 
+const renderActions = (onClose: () => void) => (
+  <ActionsPanel>
+    <Button
+      buttonType="secondary"
+      onClick={onClose}
+      data-qa-cancel
+      data-testid="dialog-cancel"
+    >
+      Cancel
```
DevDW, line 40: I mistakenly referred to it as a Cancel button in my last comment, but the mockups show "Close".

That brings up a worthwhile point, though – clicking either action button results in the modal closing. Maybe we should just have the "I Have Saved My XYZ" button? @matthewbartholomew / @mbartholomew-design

packages/manager/src/features/Profile/SecretTokenDialog/SecretTokenDialog.tsx
```diff
@@ -27,42 +29,64 @@ const useStyles = makeStyles(() => ({
...
       maxWidth="sm"
+      actions={actions}
     >
       <Notice
         spacingTop={8}
         warning
         className={classes.noticeText}
-        text={`For security purposes, we can only display your ${
+        text={`${
+          objectStorageKey ? '' : 'Your keys have been generated.'
```
DevDW, line 73: I thought `Your keys have been generated.` should only be added for the Access Keys modal?

packages/manager/src/features/Profile/SecretTokenDialog/SecretTokenDialog.tsx
HanaXu: Agreed

HanaXu (commented):
Can we include the label in the generated file name? That would make it easier to differentiate between different token files

codyfinn:
> Can we include the label in the generated file name? That would make it easier to differentiate between different token files

It is definitely possible, but I think for the first pass this works well enough. How we should handle naming the file is something I think the team should discuss. Due to how a user may have labeled their bucket or token how we format the rest of the file name could make it look weird or be hard to find.

packages/manager/src/components/DownloadTooltip.tsx
```diff
@@ -0,0 +1,82 @@
+import classNames from 'classnames';
+import { downloadFile } from 'src/utilities/downloadFile';
+import * as React from 'react';
+import FileDownload from 'src/assets/icons/download.svg';
```
bnussman, line 4: The download icon looks a bit to small and maybe a different color than the copy icon? 

Screen Shot 2022-07-07 at 12 39 06 PM

packages/manager/src/components/CopyableAndDownloadableTextField.tsx
```diff
@@ -0,0 +1,65 @@
+import * as React from 'react';
+import _ from 'lodash';
```
bnussman, line 2:
```suggestion
```

Remove this `import`

jdamore-linode (approved):

PR #8426, opened 2022-06-24
Use HTML entity for curly apostrophe in attribute on cancel landing page

rmcintosh:
## Description

Fixes a display issue on the cancel landing page where the raw code for the character is being displayed instead of the rendered character. Since this is within an attribute, I think using the HTML entity code is valid and should work.

## How to test

View the landing page with this commit.


johnwcallahan (dismissed):
Thanks @rmcintosh!

HanaXu (approved):

codyfinn (approved):

PR #8429, opened 2022-06-27
M3-5180: Refactor Domains Landing to begin React Query integration

bnussman:
## Description 📝

- Refactors Domains Landing and associated files to integrate React Query along with new patterns
- API paginates the Domains Landing table for all users (not just large accounts)

## How to test 🧪

- Testing functionality of the Domains landing page
  - `Create`, `Delete`, `Enable`, and `Disable` should work like production

`yarn test Domain`

## Known Issues (that will be addressed in follow up PRs)

> **Note**: Most of these are because we need to update the React Query store instead of the Redux store in many places

- Changes when editing Domains are not reflected on the landing page 
- Cloning a Domain is not reflected in the landing page


jdamore-linode (approved):
This looks good! Also confirmed that the Cypress tests still pass.

Since there are those two issues, is there any plan in place in case this gets merged to `develop` but you don't have time to fix those issues before the release?

PR #8430, opened 2022-06-28
M3-5305: Support Ticket Table Refactor

bnussman:
- To address the table confusion, I made the `Last Updated` column stay visible on small screens
- I took this ticket as opportunity to do a quick clean up and refactor to React Query

## How to test

- Test the two tabs of `http://localhost:3000/support/tickets`
  - Consider testing in a dev environment so you can create tickets without interrupting Linode support
  - Ensure open tickets shown in the **Open** tab and **closed** tickets only show up in the closed tab
  - Test filtering and sorting the table

### DIFF
--- a/packages/manager/src/features/Support/SupportTickets/TicketList.test.tsx
+++ b/packages/manager/src/features/Support/SupportTickets/TicketList.test.tsx
@@ -1,58 +1,81 @@
-import { screen } from '@testing-library/react';
 import * as React from 'react';
+import { screen, waitForElementToBeRemoved } from '@testing-library/react';
+import { mockMatchMedia, renderWithTheme } from 'src/utilities/testHelpers';
+import { TicketList, Props } from './TicketList';
+import { QueryClient } from 'react-query';
+import { rest, server } from 'src/mocks/testServer';
 import { supportTicketFactory } from 'src/factories';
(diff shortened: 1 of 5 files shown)


jdamore-linode:
This might just be because of MSW, but I'm seeing some strangeness with the column sorting. (Haven't been able to create enough support tickets in the dev environment to reproduce because I'm getting 429s, presumably from opening a bunch of tickets in quick succession)

Anyway, when I visit the Tickets page with MSW enabled, it shows _TEST Support Ticket_ 0 through 14. When I click the column header to sort by ticket subject, I now see _TEST Support Ticket_ 30 through 44 (still with ascending sorting). If I click the column header again, I see _TEST Support Ticket_ 0 through 14 again. But at no point was the sorting descending, and I never did see tickets 15-29.

bnussman:
@jdamore-linode Because we're letting the API do all of the pagination and sorting / filtering, testing with the MSW is very inconsistent because it does not paginate and sort at all. It's caching also plays a role in the weirdness.

jdamore-linode:
@bnussman Thought it might be something like that! The sorting was working with the 3 tickets I had opened, I just wasn't able to confirm with a larger amount of real data

jdamore-linode (approved):

codyfinn (approved):

PR #8431, opened 2022-06-28
M3-5841: Create image over estimates cost

codyfinn:
## Description

Removes the cost estimator from the Image capture tab. It is being removed because we cannot accurately determine the disk usage for a given Linode. Instead we inform the user of the per GB per month cost of a custom image and link them to a guide that has information on how they can calculate the price themselves. 

## How to test

- Ensure the cost estimator on the Image Capture tab is gone. 
- Ensure the price per GB per month is displayed on the Image Capture tab.
- Ensure there is a link to the Custom Image guide on the Image Capture tab.

### DIFF
--- a/packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
+++ b/packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
@@ -268,7 +268,6 @@ export const CreateImageTab: React.FC<Props & ImagesDispatch> = (props) => {
       <>
         <TextField
           label="Label"
-          required
           value={label}
           onChange={changeLabel}
           error={Boolean(labelError)}


jdamore-linode:
Looks like on small screens there's an overflow issue with the new blurb that explains the pricing: 

Screen Shot 2022-06-29 at 10 59 24 AM

I also did a quick search and it looks like this is the only code that used `calculateCostFromUnitPrice()` -- any chance we can get rid of `calculateCostFromUnitPrice.ts` and `calculateCostFromUnitPrice.test.ts` altogether?

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
```diff
@@ -294,7 +285,13 @@ export const CreateImageTab: React.FC<Props & ImagesDispatch> = (props) => {
           data-qa-image-description
         />
       </>
-
+      <Typography variant="body1" className={classes.helperText}>
+        Custom Images are billed at $0.10/GB per month. Maximim file size is
```
DevDW, line 289: "maximum"

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
```diff
@@ -294,7 +285,13 @@ export const CreateImageTab: React.FC<Props & ImagesDispatch> = (props) => {
...
+        6GB.{' '}
+        <Link to="https://www.linode.com/docs/products/tools/images/guides/capture-an-image/">
+          Learn more about requirements and considerations.
+        </Link>
+      </Typography>
```
DevDW, line 294: Has this copy been finalized? Also, will John's comment on the ticket re: the `du -h` flag be reflected on this page or is it something better suited for the doc that we're linking to?

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
```diff
@@ -294,7 +288,18 @@ export const CreateImageTab: React.FC<Props & ImagesDispatch> = (props) => {
           data-qa-image-description
         />
       </>
-
+      <Typography variant="body1" className={classes.helperText}>
+        Custom Images are billed at $0.10/GB per month. Learn about requirements
+        and considerations.{' '}
+        <Link to="https://www.linode.com/docs/products/tools/images/guides/capture-an-image/">
+          Learn more about requirements and considerations.{' '}
+        </Link>
```
HanaXu, line 296: Duplicate "Learn about requirements and considerations."

I also have the same question as Dajahi re: adding the `du -h` copy to this page

codyfinn: The du -h should be in the second link

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
```diff
@@ -262,19 +260,15 @@ export const CreateImageTab: React.FC<Props & ImagesDispatch> = (props) => {
...
-            Estimated: {calculateCostFromUnitPrice(0.1, selectedDiskSizeInGB)}
-            /month
-          </Typography>
-        ) : null}
       </Box>
       {isRawDisk ? rawDiskWarning : null}
       <>
         <TextField
           label="Label"
+          required
```
bnussman, line 271: I don't think this field is required. I'm able to capture an Image without completeing this feild. 
```suggestion
```

PR #8432, opened 2022-06-28
M3-5759: Don't make /account requests if restricted user

HanaXu:
## Description
Once Cloud has established that a user is restricted, we should not attempt to request any of the account endpoints

## How to test
Go to `/account` on a restricted user and click through the tabs while verifying that we longer make requests to /account endpoints

### DIFF
--- a/packages/manager/src/queries/account.ts
+++ b/packages/manager/src/queries/account.ts
@@ -5,16 +5,21 @@ import {
 } from '@linode/api-v4/lib/account';
 import { APIError } from '@linode/api-v4/lib/types';
 import { useMutation, useQuery } from 'react-query';
+import { useProfile } from 'src/queries/profile';
 import { getGravatarUrl } from 'src/utilities/gravatar';
 import { mutationHandlers, queryPresets } from './base';
 
 export const queryKey = 'account';
 
-export const useAccount = () =>
-  useQuery<Account, APIError[]>(queryKey, getAccountInfo, {
+export const useAccount = () => {
+  const { data: profile } = useProfile();
+
+  return useQuery<Account, APIError[]>(queryKey, getAccountInfo, {
     ...queryPresets.oneTimeFetch,
     ...queryPresets.noRetry,
+    enabled: !profile?.restricted,
   });
+};
 
 export const useMutateAccount = () => {
   return useMutation<Account, APIError[], Partial<Account>>((data) => {
--- a/packages/manager/src/queries/accountPayment.ts
+++ b/packages/manager/src/queries/accountPayment.ts
@@ -30,7 +30,7 @@ export const useAllPaymentMethodsQuery = () => {
     getAllPaymentMethodsRequest,
     {
       ...queryPresets.oneTimeFetch,
-      enabled: grants?.global.account_access !== null,
+      enabled: grants?.global?.account_access !== null,
     }
   );
 };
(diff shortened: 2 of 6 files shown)


bnussman (changes requested):
Works great and implementation is super simple! 🚀 I found one issue that might apply to some of the queries.

packages/manager/src/queries/account.ts
```diff
@@ -5,16 +5,21 @@ import {
...
 
-export const useAccount = () =>
-  useQuery<Account, APIError[]>(queryKey, getAccountInfo, {
+export const useAccount = () => {
+  const { data: profile } = useProfile();
+
+  return useQuery<Account, APIError[]>(queryKey, getAccountInfo, {
     ...queryPresets.oneTimeFetch,
     ...queryPresets.noRetry,
+    enabled: !profile?.restricted,
```
johnwcallahan, line 20: Is the non-null assertion operator necessary here? (and similar places)

codyfinn: Do you mean the optional chaining operator?

packages/manager/src/queries/account.ts
```diff
@@ -5,16 +5,21 @@ import {
...
+import { useProfile } from 'src/queries/profile';
 import { getGravatarUrl } from 'src/utilities/gravatar';
 import { mutationHandlers, queryPresets } from './base';
 
 export const queryKey = 'account';
 
-export const useAccount = () =>
-  useQuery<Account, APIError[]>(queryKey, getAccountInfo, {
+export const useAccount = () => {
+  const { data: profile } = useProfile();
```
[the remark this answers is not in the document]
bnussman: The profile is fetched and cached whenever Cloud Manager loads (see `AuthenticationWrapper.tsx`). So making queries dependent on the `useProfile` query should be fine. I won't actually make extra https requests unless somehow the profile is evicted from the cache.

packages/manager/src/queries/account.ts
johnwcallahan: Yeah, I was confusing the leading `!` with a non-null assertion; that's my mistake. Thanks for the discussion @codyfinn and @bnussman.

packages/manager/src/queries/accountPayment.ts
```diff
@@ -18,11 +23,14 @@ export const usePaymentMethodsQuery = (params: any = {}, filter: any = {}) => {
...
 
 export const useAllPaymentMethodsQuery = () => {
+  const { data: grants } = useGrants();
+
   return useQuery<PaymentMethod[], APIError[]>(
     queryKey + '-all',
     getAllPaymentMethodsRequest,
     {
       ...queryPresets.oneTimeFetch,
+      enabled: grants?.global.account_access !== null,
```
bnussman, line 33: Screen Shot 2022-07-01 at 12 40 31 PM

The `!== null` seems spot on. I think we just need this optional chain to not crash non-restricted accounts

```suggestion
      enabled: grants?.global?.account_access !== null,
```

bnussman (changes requested):

codyfinn (approved):
Looks good to me! Might want to add tests but I don't think that is blocking.

bnussman (approved):
Nice work on this one 🚀

PR #8438, opened 2022-06-29
M3-5853: Make 2FA / Security Questions Errors Accurate

bnussman:
## Description 📝

- If an existing user with 2FA enabled **and** without security questions tried to press `Reset two-factor authentication`, they would get an error notice saying `You must add Security Questions to your profile in order to enable Two Factor Authentication`. This may be confusing to users because it says `enable` instead of `reset`
- This PR makes it so the error will say `reset` if the API returns `You must add Security Questions to your profile in order to enable Two Factor Authentication` **and** you have 2FA enabled.

## Preview 📷

https://user-images.githubusercontent.com/6440455/176538305-e487e383-340a-4d95-a9b2-6adbb7eb4ec4.mov

## How to test 🧪

- Get an account into a state where it does not have security questions, but it does have 2FA enabled
- Press the `Reset two-factor authentication` button
- Verify that you see `You must add Security Questions to your profile in order to reset Two-Factor Authentication` (specifically it should say `reset` instead of `enable`)

### DIFF
--- a/packages/manager/src/features/Profile/AuthenticationSettings/TwoFactor/TwoFactor.tsx
+++ b/packages/manager/src/features/Profile/AuthenticationSettings/TwoFactor/TwoFactor.tsx
@@ -13,7 +13,6 @@ import { queryKey } from 'src/queries/profile';
 import { useSecurityQuestions } from 'src/queries/securityQuestions';
 import { getAPIErrorOrDefault } from 'src/utilities/errorUtils';
 import getAPIErrorFor from 'src/utilities/getAPIErrorFor';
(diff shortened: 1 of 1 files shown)


bnussman:
@DevDW I did something like that in the first commit 537fcf8, but it felt a bit more clunky. I figured manually setting the error is worth it for cleaner code and one less API call

jdamore-linode (approved):
I think this is a good approach! Determining what to show by checking or modifying the API message would work too but I'd be a little nervous that the API team could change the message at any time and cause our logic to break

PR #8442, opened 2022-06-30
M3-5255: Fix Linode stats graphs to be more responsive

DevDW:
## Description
In prod, when you resize your window while on the "Analytics" tab of a Linode Detail page, there can be periods of up to ~30 seconds where the graphs are distorted in their shapes. This occurs because the graphs did not re-render until the data fetch interval came around again.

This PR adds logic to refetch the stats 1.5 seconds after resizing the window.

## How to test
While on the "Analytics" tab of a Linode Detail page, try resizing the window multiple times. You should see a call to `stats` in the Network tab of Devtools shortly after you do each resize. Compare the behavior on this branch with production.


HanaXu (approved):
🚀

codyfinn (approved):

PR #8444, opened 2022-07-01
M3-5092: `hosted-git-info` - CVE-2021-23362 

bnussman:
## Description 📝

- Pins  `hosted-git-info`  to the latest version at address CVE-2021-23362 

With these types of tickets, it would be nice to be able to just update the dependencies that depend on `hosted-git-info` to resolve this, but packages that use `hosted-git-info` such as `npm-run-all` have not updated `hosted-git-info`  to the latest yet, so it's no use.

## How to test 🧪

- Test functionality of Cloud Manager (`yarn up`)
- Test functionality of Storybook (`yarn storybook`)
  - Storybook depends on `hosted-git-info`

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,8 @@
     "glob-parent": "^5.1.2",
     "browserslist": "^4.16.5",
     "nth-check": "v2.0.1",
-    "trim-newlines": "^4.0.2"
+    "trim-newlines": "^4.0.2",
+    "hosted-git-info": "^5.0.0"
   },
   "workspaces": {
     "packages": [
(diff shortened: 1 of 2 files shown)


jdamore-linode:
Do you know if `npm-run-all` is the only thing using `hosted-git-info`? If so I might be able to find an alternative

bnussman:
@jdamore-linode It's not just `npm-run-all`, its also Storybook and a few other things

jdamore-linode (approved):
Nice! Poked around in Cloud Manager and Storybook and didn't notice any issues or differences, and also ran Cypress without issue. Thanks for taking care of this!

codyfinn (approved):

PR #8448, opened 2022-07-05
M3-5771: Add Cypress tests for account security questions

jdamore-linode:
## Description

**What does this PR do?**
This PR adds Cypress integration tests for account security questions at `/profile/auth`. These tests use mocked data to confirm that the Cloud Manager UI responds accordingly when adding and updating security questions.

## How to test

**How do I run relevant tests?**

**⚠️ Note:** these tests may fail if you have not yet dismissed the banner regarding Linode/Akamai billing. This is because the banner is displayed over the Cloud Manager content, and in some circumstances can block Cypress's click attempts.

To run via CLI:
```
yarn cy:run -s "cypress/integration/account/security-questions.spec.ts"
```

To run interactively:
```
yarn cy:debug
```
Then search for _security-questions_.

### DIFF
--- a/packages/manager/cypress/integration/account/security-questions.spec.ts
+++ b/packages/manager/cypress/integration/account/security-questions.spec.ts
@@ -2,17 +2,321 @@
  * @file Integration tests for account security questions.
  */
 
-import { randomLabel, randomString } from 'support/util/random';
-import { mockGetProfile } from 'support/intercepts/profile';
+import { securityQuestionsFactory } from 'src/factories/profile';
+import {
+  mockGetSecurityQuestions,
+  mockUpdateSecurityQuestions,
+} from 'support/intercepts/profile';
 import { ui } from 'support/ui';
 import { assertToast } from 'support/ui/events';
 
+/**
+ * Finds the "Security Questions" section on the profile auth page.
+ *
+ * @returns Cypress chainable.
+ */
+const getSecurityQuestionsSection = (): Cypress.Chainable => {
+  return cy.contains('h3', 'Security Questions').parent();
+};
+
+/**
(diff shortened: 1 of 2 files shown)


packages/manager/cypress/integration/account/security-questions.spec.ts
```diff
@@ -0,0 +1,322 @@
...
+ *
+ * @param questionNumber - Security question number (1-3) to edit.
+ */
+const editQuestion = (questionNumber: number) => {
+  getSecurityQuestion(questionNumber).within(() => {
+    ui.button
+      .findByTitle('Edit')
+      .should('be.visible')
+      .should('be.enabled')
+      .click({ scrollBehavior: 'center' });
```
bnussman, line 57: This step timed out on it's first run, then passed on it's second. 
Screen Shot 2022-07-12 at 2 55 30 PM

jdamore-linode: Thanks @bnussman! This one is strange, and I'm still trying to figure it out, but I am able to reproduce it. Based on Cypress's message, it appears that the "Edit" element is being detached from the DOM at some point between the time that Cypress finds the element (and asserts that it's visible and enabled, etc) and the time that it dispatches the click events. The only thing that comes to mind is that React is could be re-rendering, but I'm not aware of any reason that it may be doing that. Does anything come to mind from the time you spent working on this component?

DevDW (approved):
I ran the tests 3 times throughout the day and didn't encounter the timeout issue, so I'm approving pending further investigation of that

bnussman (approved):
We can circle back on this if there are issues, but it ran fine for me!

PR #8451, opened 2022-07-07
M3-5180: React Query for Domains Event Handler

bnussman:
## Description 📝

- Ports functionality of `packages/manager/src/store/domains/domains.events.ts` to React Query 
- `packages/manager/src/store/domains/domains.events.ts` will be deleted later on

### Why this code exists
It's pretty edge-casey, but by using our events polling we can do our best to keep Cloud Manager up to date with what is actually true on the backend. An example of this is shown in the video below. Notice how two instances of Cloud Manager update to reflect the same data with no added polling. 

https://user-images.githubusercontent.com/6440455/177845250-3c731fde-8e96-4879-8619-2798932afbe4.mov

## How to test 🧪

- Test **Creating** and **Deleting** Domains


packages/manager/src/queries/domains.ts
```diff
@@ -75,3 +76,48 @@ const updatePaginatedDomainsStore = (id: number, newData: Partial<Domain>) => {
...
+    const data = store[1]?.data;
+    if (data?.some((domain) => domain.id === id)) {
+      return true;
+    }
+  }
+
+  return false;
+};
+
+// @TODO: make this generic
```
bnussman, line 96: After I complete React Query for Domains, I'll resolve these `@todo`s by creating generic functions to reduce code duplication. Domains and Databases will be able to use the newly created generic helpers

jdamore-linode (approved):
Looks good! And domain tests pass 👍

HanaXu (approved):

PR #8452, opened 2022-07-08
M3-5889: Safari on iOS Accessibility and Zoom Fix

bnussman:
## Description 📝

- In Safari on iOS, the browser zooms in when a user clicks into any TextField. This is very annoying and not great.
- At first I thought it was an issue with our `<meta name="viewport" />`, but it's not the main issue.
- After some research I found that Safari will do this if your input font size is less than 16px.
- The fix was to make all inputs have a font size of `1rem` / `16px` on `XS` screen sizes

### Before 🚫

https://user-images.githubusercontent.com/6440455/178029324-c51c5c89-ed39-4ee1-8759-dfc315506630.mp4

### After ✅

https://user-images.githubusercontent.com/6440455/178029338-323030c5-453e-4997-8100-cc3729f21bec.mp4

## How to test 🧪

- Test Cloud Manager's inputs in your computer's browser
- Test Cloud Manager on your mobile phone's browser

### DIFF
--- a/packages/manager/public/index.html
+++ b/packages/manager/public/index.html
@@ -4,7 +4,7 @@
     <meta charset="utf-8" />
     <meta
       name="viewport"
-      content="width=device-width, initial-scale=1, shrink-to-fit=no"
+      content="width=device-width, initial-scale=1.0"
     />
     <meta name="theme-color" content="#000000" />
     <% if (process.env.NODE_ENV === 'production') { %>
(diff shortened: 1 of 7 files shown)


jdamore-linode (approved):
Nice! It's working well for me using the simulator

HanaXu (commented):
Since the font-size is being changed globally, this also affects alignment and inputs on Chrome. And the bigger font on Chrome looks a bit off 🤔 

Before
image

After
image

HanaXu (commented):
@bnussman Can you double check with Matthew regarding the global font size change? It just looks off to me in comparison to surrounding font 🤔

bnussman:
I'm going to make an internal ticket the investigate this. A possible long term solution could be to make inputs have size 16 font **only** on mobile. cc @HanaXu

HanaXu (approved):

PR #8459, opened 2022-07-13
M3-5877: Override Marketplace app doc link with LaunchDarkly

HanaXu:
## Description
Sometimes, there are small mismatches in a Marketplace app's doc link that break the link (missing a dash, capital letter, etc).

This PR adds the ability to override the app's doc link using LaunchDarkly so that we can fix the link without an immediate release/hotfix

## How to test
Go to Marketplace and check the docs link in the Kali Linux detail drawer with the `one-click-apps-docs-override` flag on/off


jdamore-linode (approved):
Looks good!

PR #8460, opened 2022-07-14
PayPal Loading State Improvement

bnussman:
## Description 📝

- Takes PayPal's script loading into account when rendering the PayPal button spinner
- Small difference, but creates less time where there is no spinner and no button at the same time

### Before

https://user-images.githubusercontent.com/6440455/179014587-89562fcf-c50e-46d1-8ab7-101725fdfcb0.mov

### After

https://user-images.githubusercontent.com/6440455/179014568-8f4cc36e-0650-40da-8a10-efd26c2f0f82.mov

## How to test 🧪

- Test PayPal functionality the best you can locally and on the preview link


jdamore-linode (approved):
This is a nice improvement! I couldn't really perceive the difference between the two videos in your PR description, but when I checked it out and compared my local dev with `cloud.linode.com` the difference was definitely noticeable!

PR #8462, opened 2022-07-18
M3-5180: React Query for Domains - Remove Redux, Allow Search, Attach to Ticket

bnussman:
## Description 📝

- Removes domains from Redux store
- Connects React Query store to search
- Make support ticket entity attachment pull from React Query store for domains
- Makes the Domain Records page use React Query when updating the Domain
- Removes functionality to convert domain groups to tags for legacy Classic Manager customers

## How to test 🧪

- Test Search
- Test attaching Domains to support tickets
- Test Domain details
- Test creating, editing, and deleting domain records


jdamore-linode:
Nicely done @bnussman! Seeing [this](https://user-images.githubusercontent.com/97627410/179783397-d12d965e-bf8f-415c-bf1b-e00af080261c.png) never gets old.

I went through the flows you listed and things looked good, and also confirmed that all of the Cypress tests are passing (minus one unrelated image test that would be fixed by merging in the latest changes from `develop`).

The only thing I've noticed is some Jest errors from `SearchLanding.test.tsx`. They're not causing the tests to fail, but they do appear to be new errors (I'm not seeing them on `develop`).

```
yarn test Search/SearchLanding.test
```

Screen Shot 2022-07-19 at 11 02 39 AM

bnussman:
@jdamore-linode Great find. Thankfully those errors are intentional. Previously, I mocked some errors with the MSW when testing Object Storage. I'll update the MSW to not throw these errors.

jdamore-linode (approved):
No longer seeing those Jest console messages!

packages/manager/src/features/Search/SearchLanding.tsx
```diff
@@ -95,12 +96,18 @@ export const SearchLanding: React.FC<CombinedProps> = (props) => {
     data: objectStorageClusters,
     isLoading: areClustersLoading,
     error: objectStorageClustersError,
-  } = useObjectStorageClusters(_isLargeAccount);
+  } = useObjectStorageClusters(!_isLargeAccount);
```
DevDW, line 99: What's the reasoning behind changing this to `!_isLargeAccount`? Did we overlook this before?

DevDW (approved):
LGTM after combing through the code and testing the functionality -- left one question for clarification

packages/manager/src/features/Search/SearchLanding.tsx
bnussman: Yes. I realized this existing logic was not correct. We want to use the React Query store when we don't use the full API search.

PR #8464, opened 2022-07-19
M3-5868: Marketplace Apps list refinement

HanaXu:
## Description
- Marketplace Apps list is now a scrolling panel with three sections:
    - New apps, Popular apps, All apps (there will be some duplication between sections)
- Replaced grayscale app svgs with full color svgs

## How to test
Go to `/linodes/create?type=One-Click` and check the Apps list panel


packages/manager/src/features/linodes/LinodesCreate/SelectAppPanel.tsx
```diff
@@ -179,12 +176,34 @@ class SelectAppPanel extends React.PureComponent<CombinedProps> {
...
               classes={classes}
             />
           ))}
         </Grid>
-      </Panel>
+      </>
+    );
+
+    const newApps = appInstances.filter((app) => {
+      return ['kali linux', 'easypanel', 'liveswitch', 'saltcorn'].includes(
```
codyfinn, line 188: Could this be hooked up to a LD variable so that we don't have to push a commit to update this info?

packages/manager/src/features/linodes/LinodesCreate/SelectAppPanel.tsx
```diff
@@ -164,10 +159,12 @@ class SelectAppPanel extends React.PureComponent<CombinedProps> {
       return null;
     }
 
-    return (
-      <Panel className={classes.panel} error={error} title="Select App">
+    const panelSection = (heading: string, apps: StackScript[]) => (
```
codyfinn, line 162: Should we pull this component out?

packages/manager/src/features/linodes/LinodesCreate/SelectAppPanel.tsx
HanaXu: @codyfinn This will basically only need be updated when there's a PR for a new app, so I think this is fine. I also want to avoid having a bunch of LD flags just for Marketplace

packages/manager/src/features/linodes/LinodesCreate/SelectAppPanel.tsx
HanaXu: @codyfinn I think this component is simple enough to just leave in this file

codyfinn: It should be pulled out of the component at least. As it is a new function is allocated for each instance the parent component is used.

packages/manager/src/features/linodes/LinodesCreate/SelectAppPanel.tsx
```diff
@@ -179,12 +176,34 @@ class SelectAppPanel extends React.PureComponent<CombinedProps> {
...
+    const newApps = appInstances.filter((app) => {
+      return ['kali linux', 'easypanel', 'liveswitch', 'saltcorn'].includes(
+        app.label.toLowerCase().trim()
+      );
+    });
+
+    const popularApps = appInstances.slice(0, 10);
+
+    // sort mutates original array so make a copy first
+    const allApps = [...appInstances].sort((a, b) =>
```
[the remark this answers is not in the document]
codyfinn: This creates a copy of the `appInstances`. Using the `sort` method on an array mutates the array. We want to avoid mutating data we get from props/state so this works around that issue.

codyfinn (approved):

patthiel:
I noticed on certain one click apps the drawer has some alignment issues between the app name and logo:
image

bnussman (commented):
Can we clean up the loading state of this section?

https://user-images.githubusercontent.com/6440455/181601551-87372d08-8792-4f78-a69c-6438ae807ba8.mov

https://user-images.githubusercontent.com/6440455/181601582-19f633e2-9dff-4ffc-b15b-b262c5a7c0b2.mov

https://user-images.githubusercontent.com/6440455/181601599-2e651c29-90a2-4605-ba39-47184adf5602.mov

- Sometimes there is the `<LinearProgress />`. Maybe we should replace this with a centered `<CircularProgress />` now that the height of this section is static?
- Sometimes this section shows `No compatible images are available`, can this be fixed?

HanaXu:
@bnussman The `No compatible images are available` issue is a bit out of scope for this ticket but I created a `ImageEmptyState` component to dedupe some of the code. We can look into this if it gets brought up again

packages/manager/src/features/linodes/LinodesCreate/TabbedContent/ImageEmptyState.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export const ImageEmptyState: React.FC<Props> = (props) => {
+  const { errorText, className } = props;
+  const classes = useStyles();
+
+  return (
+    <Paper className={className}>
+      {/* empty state for images */}
+      {errorText ? <Notice error={true} text={errorText} /> : null}
```
DevDW, line 26: `error` can just be passed by itself to `<Notice />`, no need to pass the boolean explicitly with it I believe

DevDW (approved):
LGTM, these are some nice tweaks to the Marketplace Apps section.

PR #8466, opened 2022-07-20
[Hotfix] Add dismissible banner to Database Landing and Database Create pages re: maintenance

DevDW:
## Description
Add a dismissible banner to the Databases Landing and Database Create pages regarding maintenance.

This PR modifies `ProductInformationBanner` to make it capable of displaying a warning icon to the left of the banner text.

## How to Test
Confirm that the dismissible banner is visible on the Databases Landing page (both the empty state and also when you have DBs) and Database Create page, and that dismissing the banner in one location dismisses it in all of them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [2022-07-20] - v1.71.1
+
+### Added:
+- Banner regarding maintenance to the Databases Landing and Database Create pages
+
 ## [2022-07-12] - v1.71.0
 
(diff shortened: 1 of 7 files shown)


bnussman:
> DBaaS is currently undergoing maintenance. During this time you will not be able to provision new instances.

I'm not sure if this copy is final, but it should probably say something like `Managed Databases` instead of `DBaaS`.

DevDW:
> > DBaaS is currently undergoing maintenance. During this time you will not be able to provision new instances.
> 
> I'm not sure if this copy is final, but it should probably say something like `Managed Databases` instead of `DBaaS`.

I agree but for now we're just matching the API error message and once that's updated, we can update ours in LD

HanaXu (approved):
LGTM! 🚢

bnussman (approved):

jdamore-linode (approved):

PR #8469, opened 2022-07-22
M3-5876: Add Payment Drawer error notice

HanaXu:
## Description
Display an error notice in Add Payment Drawer instead of a toast (and some minor refactoring)

Before
image

After
image

## How to test
Point to dev and try to make/add a third party payment on an account that has the `pays by akamai` flag

Also test normal flow on an account without the flag


DevDW (commented):
Going to test more thoroughly on Monday morning but the first pass looked good

DevDW (approved):
🚢

jdamore-linode (approved):
Nice! Also confirmed that all of the payment related Cypress tests still pass

codyfinn (approved):

PR #8470, opened 2022-07-22
M3-5584: Clear the region in Linode Create flow when switching between Create tabs

DevDW:
## Description
Currently in prod, if you go to `/linodes/create` and fill in the region but then switch Create tabs to create the Linode from a different source, the "Region" field appears to be populated still, but the actual value has been cleared (which is why filling everything out on the most recent Create tab and clicking "Create Linode" results in the user seeing an error on the "Region" field, although it appears to be populated).

This PR clears the value when the region ID does not return a matched region.

## How to test
1. Go to `/linodes/create` and choose a region 
2. Switch Create tabs by clicking "Marketplace", etc.
3. Observe: the "Region" field should no longer be populated on this branch, and filling out the rest of the required information should allow you to successfully create a Linode with no errors.

Additionally, other places in the app where `<RegionSelect />` is used should not be adversely impacted (e.g., Kubernetes Cluster creation, Linode migration, etc.).


bnussman (approved):
Nice job 🚀

HanaXu (approved):

PR #8471, opened 2022-07-26
Cloud Manager Props Clean Up

bnussman:
## Description 📝

- Cleans up props of all components
- Removes `React.FC` where it is unnecessary 

## Why ❓
- https://medium.com/raccoons-group/why-you-probably-shouldnt-use-react-fc-to-type-your-react-components-37ca1243dd13
- This will improve readability of our code
- May help speed up `tsc` a bit with less complex typings

## How to test 🧪

- Run typechecking with `yarn workspace linode-manager typecheck`
- Check basic app functionality
  - The diff looks crazy, but it's just type changes throughout the app. If `tsc` passes, we can safely assume no app functionality will change


codyfinn (approved):
LGTM thanks for this cleanup!

PR #8472, opened 2022-07-27
M3-5143: Fix Linode Backups tab error

DevDW:
## Description
Under certain circumstances outlined in the ticket, a user would see this screen when navigating to the "Backups" tab of the Linode Detail page:

Screen Shot 2022-07-27 at 2 40 19 PM

This PR fixes this bug.

More in-depth explanation and steps for testing are in the Slack thread in the team channel.

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeBackup/LinodeBackup.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeBackup/LinodeBackup.tsx
@@ -794,7 +794,7 @@ class _LinodeBackup extends React.Component<CombinedProps, State> {
     }
 
     const backupsMonthlyPrice =
-      type?.response?.addons?.backups?.price?.monthly ?? 0;
+      type.response?.addons?.backups?.price?.monthly ?? 0;
 
     return (
       <div>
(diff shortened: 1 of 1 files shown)


jdamore-linode (approved):
Looks good! I was able to reproduce the error when navigating to the backups tab on cloud.linode.com and confirmed that it doesn't appear with your changes in place.

Just noting that I'm still unable to actually enable backups on a Linode in this state, but I'm guessing that's an API thing.

packages/manager/src/features/linodes/LinodesDetail/LinodeBackup/LinodeBackup.tsx
```diff
@@ -794,7 +794,7 @@ class _LinodeBackup extends React.Component<CombinedProps, State> {
     }
 
     const backupsMonthlyPrice =
-      type?.response.addons.backups.price.monthly ?? 0;
+      type?.response?.addons?.backups?.price?.monthly ?? 0;
```
HanaXu, line 797: `type` is technically defined, just not the response so we should be able to remove the first optional chain

image

HanaXu (approved):

PR #8473, opened 2022-07-27
M3-5435: Filter `read_only` Linodes From Firewall Select For Restricted Users

bnussman:
## Description 📝

- Filters out `ready-only` Linodes from the `Firewall Create Drawer` and `Add Linode to Firewall Drawer` for restricted users. 
- We are doing this because the API does not allow restricted users to create firewalls with Linodes the user does not have permission to modify. The API error message for this was ambiguous and this change will result in a more useable experience for restricted users using firewalls.

### Un-restricted User (Full Permission)
Screen Shot 2022-07-27 at 3 44 40 PM

### Restricted User where Linode `testing-2` is `read-only` for this user
Screen Shot 2022-07-27 at 3 45 00 PM

## How to test 🧪

- Have a **primary account with full permission** and a **restricted user** with **read only permission on a Linode**
- Give the restricted user has permission to create firewalls
- On the restricted user, try to create a Firewall and attempt to attach a Linode you have `read-only` permission on
  - You should not see this Linode in the dropdown
  - You should see a note in the dropdown saying `Only Linodes you have permission to modify are shown.`

### DIFF
--- a/packages/manager/src/features/Firewalls/FirewallDetail/Devices/AddDeviceDrawer.tsx
+++ b/packages/manager/src/features/Firewalls/FirewallDetail/Devices/AddDeviceDrawer.tsx
@@ -11,6 +11,9 @@ import Notice from 'src/components/Notice';
 import { useStyles } from 'src/components/Notice/Notice';
 import SupportLink from 'src/components/SupportLink';
 import { useGrants, useProfile } from 'src/queries/profile';
+import { getEntityIdsByPermission } from 'src/utilities/grants';
+import { READ_ONLY_LINODES_HIDDEN_MESSAGE } from '../../FirewallLanding/AddFirewallDrawer';
+
 interface Props {
   open: boolean;
   error?: APIError[];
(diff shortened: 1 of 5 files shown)


packages/manager/src/features/Firewalls/FirewallDetail/Devices/AddDeviceDrawer.tsx
```diff
@@ -85,6 +90,15 @@ export const AddDeviceDrawer: React.FC<Props> = (props) => {
     }
   };
 
+  // If a user is restricted, they can not add a read-only Linode to a firewall.
+  const readOnlyLinodeIds = isRestrictedUser
+    ? grants?.linode
+        .filter((grant) => grant.permissions === 'read_only')
+        .map((grant) => grant.id) ?? []
```
codyfinn, line 97: Could this be replaced with the reduce function?

packages/manager/src/features/Firewalls/FirewallLanding/AddFirewallDrawer.tsx
```diff
@@ -91,6 +94,15 @@ const AddFirewallDrawer: React.FC<CombinedProps> = (props) => {
   const firewallHelperText = `Assign one or more Linodes to this firewall. You can add
   Linodes later if you want to customize your rules first.`;
 
+  // If a user is restricted, they can not add a read-only Linode to a firewall.
+  const readOnlyLinodeIds = _isRestrictedUser
+    ? grants?.linode
+        .filter((grant) => grant.permissions === 'read_only')
+        .map((grant) => grant.id) ?? []
```
codyfinn, line 101: This code is used twice within this PR and might be used more throughout the app. Could this be pulled out into a utility function?

DevDW (approved):
This is working well from my testing. Going ahead and approving but there are a couple of possible refactors and extractions of logic that can be done to reduce code duplication

HanaXu (approved):

PR #8474, opened 2022-07-28
M3-5809: Run Cypress via Docker Compose

jdamore-linode:
## Description

**What does this PR do?**
This PR adds the capability to run Cypress via Docker, which will later be used to run Cypress in Jenkins.

Because this was built with Jenkins in mind, it is assumed that dependencies are already installed (for Linux systems) and that Cloud Manager is already built when the Docker container starts.

## How to test

Before you can build and start the containers, you must install Node dependencies for Linux and build Cloud Manager. Because Docker volumes run slowly on MacOS, this will take noticeably longer than usual but thankfully only needs to be done once. Install Linux dependencies and build Cloud Manager using these commands:

```sh
# From `manager` repo root
# Run `yarn clean` to delete Node dependencies for MacOS
yarn clean

# Run `yarn install:all` from inside the web container to install Node dependencies for Linux
docker compose run web yarn install:all

# Run `yarn build` from inside the web container to build Cloud Manager using Linux dependencies
docker compose run web yarn build
```
Running the 3 commands above took 25 minutes on my machine.

Once your environment is set up, you can run Cypress via Docker Compose:

```sh
docker compose build
docker compose up e2e
```

Confirm that Cypress runs and that tests are generally passing.


bnussman:
Do these changes assume Cloud Manager is fully built? in `packages/manager/build`?

jdamore-linode:
@bnussman That's correct! The Jenkins build script will be responsible for installing dependencies and building Cloud Manager. It's WIP, but there's a link to the corresponding Jenkinsfile changes in the comments of M3-5895 if you want to see how the two pieces fit together.

docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+  web:
+    build:
+      context: .
+      dockerfile: ./packages/manager/Dockerfile
+      target: manager
+    container_name: web
+    ports:
+      - "3000:3000"
+    volumes:
+      - .:/home/node/app
```
patthiel, line 14: Are we only mounting a volume here because we want the compiled static assets from the previous jenkins stages to exist in the container context? If so, i still think we may want to be a bit more specific about the volumes we mount.

packages/manager/Dockerfile
```diff
@@ -0,0 +1,27 @@
+# `manager`
+#
+# Serves Cloud Manager.
+FROM node:14.17-bullseye-slim as manager
+ENV NODE_ENV=development
+WORKDIR /home/node/app
+VOLUME /home/node/app
+EXPOSE 3000/tcp
+# Curl is needed for health check.
+RUN apt-get update && apt-get install -y curl
```
patthiel, line 10: Cleaning up after installing will help reduce the image size. We do this in many of our other Dockerfiles. I would look around at other example projects to see what they're doing but here are some quick suggestions!


```suggestion
RUN apt-get update
    && apt-get install -y curl \ 
    && rm -rf /var/cache/apt/* \
    && rm -rf /var/lib/apt/lists/* \
    && apt-get clean
```

bnussman (approved):
Everything built and the tests ran and some passed! I didn't let all of the suite run, but it looked good. Will this new approach allow us to still do parallelization with more than one instance?

jdamore-linode:
> Everything built and the tests ran and they passed! I didn't let the entire suite run, but it looked good when it was running. Will this new approach allow us to still do parallelization with more than one instance?

Thanks! No need to let the tests finish. As of now there's no parallelization, but this ticket lays some groundwork for that and another ticket (M3-5896) details the remaining work that needs to be done to get parallelization working (which really isn't very much).

jdamore-linode:
@patthiel Managed to narrow down the scope of our volumes a little bit -- I was pretty limited by what I could do for the `e2e` container sadly. Let me know if you have any other suggestions!

patthiel (approved):
Discussed offline the possibility of exploring tmpfs volume mounts for performance improvement. But that is non-blocking feedback. If you end up making those changes in this PR, feel free to tag me and i'll re-review. Otherwise 👍  from me!

PR #8477, opened 2022-08-01
M3-5117: Remove __data__/types.ts and update affected files

DevDW:
## Description
Removes `__data__/types.ts` and updates the files (mostly tests) that used it accordingly by bringing back `cachedData/types.json` (see https://github.com/linode/manager/pull/7777/files) to use instead.

## How to test
Ensure that `yarn up` works with no TS errors; run the affected tests and ensure everything passes –

```
yarn test typesHelpers.test.ts
```

```
yarn test GetSearchEntities.test.ts
```

```
yarn test ResizeNodePoolDrawer.test.ts
```

```
yarn test BackupsTable.test.ts
```

```
yarn test BackupsDrawer.test.ts
```

```
yarn test BackupLinodes.test.ts
```


DevDW:
I'm aware of the test failures uncovered by the unsuccessful check -- going to make a couple more pushes today fixing those tests

packages/manager/src/cachedData/types.json
```diff
@@ -0,0 +1,679 @@
+{
```
bnussman, line 1: How did this file get here? I'm thinking the generation of this should be added to `packages/manager/scripts/buildRequests.js`?

packages/manager/src/features/Backups/BackupsTable.test.tsx
```diff
@@ -1,15 +1,22 @@
...
-const linode1: ExtendedLinode = { ...linodes.linode1, typeInfo: type };
-const linode2: ExtendedLinode = { ...linodes.linode2, typeInfo: type };
+const linode1: ExtendedLinode = {
+  ...linodes.linode1,
+  typeInfo: type as LinodeType,
+};
+const linode2: ExtendedLinode = {
+  ...linodes.linode2,
+  typeInfo: type as LinodeType,
+};
```
bnussman, line 19: I'm thinking things like this should just be replaced by a Linode factory and a types factory. That may be out of the scope of this work.

packages/manager/src/features/Backups/BackupDrawer.test.tsx
```diff
@@ -2,8 +2,9 @@ import '@testing-library/jest-dom/extend-expect';
 import { fireEvent, render, waitFor } from '@testing-library/react';
 import * as React from 'react';
 
-import { linode1, linode2, linode3 } from 'src/__data__/linodes';
-import * as types from 'src/__data__/types';
+import { LinodeType } from '@linode/api-v4/lib/linodes';
+import data from 'src/cachedData/types.json';
```
bnussman, line 6: Use type factory here instead?

packages/manager/src/cachedData/types.json
DevDW: I added it back manually after removing it last year as part of the PR linked in the description. We wouldn't want it in `buildRequests.js` because that was causing problems that resulted in its removal back then. I'm not opposed to moving the file outside of the `cachedData` folder though.

packages/manager/src/features/Backups/BackupDrawer.test.tsx
DevDW: Some assertions in the test need the full array of the different types so for now I'll leave it as-is

bnussman (approved):

HanaXu (approved):

PR #8480, opened 2022-08-03
M3-5831: Fix invoice formatting issue

jdamore-linode:
## Description

**What does this PR do?**
This fixes the invoice formatting issue that occurs when any of the (sub)totals listed at the bottom of the invoice overflow their column.

This change removes the hardcoded column width (which was `30` units) and implements new logic to calculate the final widths of the columns.

## How to test

1. Enable MSW
2. Modify the [mocked invoice values](https://github.com/linode/manager/blob/develop/packages/manager/src/factories/billing.ts#L40-L52) to correspond with what's described in M3-5831.
3. Navigate to `/account/billing` and click _Download PDF_ for any of the listed invoices
4. View the generated PDF and confirm that formatting issue is fixed
5. _(Optional)_ Continue modifying the [mocked invoice values](https://github.com/linode/manager/blob/develop/packages/manager/src/factories/billing.ts#L40-L52) and confirm that formatting issue is fixed even with extreme values

### DIFF
--- a/packages/manager/src/factories/billing.ts
+++ b/packages/manager/src/factories/billing.ts
@@ -30,7 +30,7 @@ export const paymentFactory = Factory.Sync.makeFactory<Payment>({
   usd: 5,
 });
 
-const invoiceDate = new Date('2020-01-01T00:00:00');
+const invoiceDate = new Date('2022-08-01T00:00:00');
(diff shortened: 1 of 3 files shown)


bnussman (approved):
Wrapping of the subtotals and totals looks much better! 🚀

DevDW (approved):
Nice work 🚢

PR #8481, opened 2022-08-03
M3-4919: Fix browser Back button behavior on the User Permissions tab

DevDW:
## Description
In prod, if you're on the "Users & Grants" tab on the Account page, click on "User Profile" for a user, click on the "User Permissions" tab, and then use the browser Back button to navigate back, you'll observe odd behavior: the tab doesn't switch, and if you hit the Back button enough, you'll go back further than expected.

The changes I made bring the behavior more in line with what we see on the Object Storage landing page. It's not ideal because it still takes two clicks of the browser Back button to get to the first tab, but this seems to be either a problem with ReachTabs or the way we're implementing it.

## How to test
Test out the "Users & Grants" --> "User Profile" --> "User Permissions" path and then the reverse using the browser Back button. You shouldn't get stuck as much now.

### DIFF
--- a/packages/manager/src/features/Users/UserDetail.tsx
+++ b/packages/manager/src/features/Users/UserDetail.tsx
@@ -182,11 +182,6 @@ const UserDetail: React.FC = () => {
     return Boolean(matchPath(p, { path: location.pathname }));
   };
 
-  const clampTabChoice = () => {
-    const tabChoice = tabs.findIndex((tab) => matches(tab.routeName));
-    return tabChoice < 0 ? 0 : tabChoice;
-  };
-
   const navToURL = (index: number) => {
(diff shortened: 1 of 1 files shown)


bnussman (commented):
Going to look a bit more into this, but at first glance I agree with `"seems to be either a problem with ReachTabs or the way we're implementing it"`

I implemented something like this in a personal project and seemed to have a better experience with a similar implementation using Chakra UI's Tabs. https://github.com/bnussman/beep/blob/main/website/src/routes/admin/users/User.tsx#L287

HanaXu (approved):

bnussman (approved):

jdamore-linode (approved):

PR #8483, opened 2022-08-04
M3-5867: Marketplace refinement search and filtering

codyfinn:
## Description

**What does this PR do?**

## How to test


matthewbartholomew:
Adding these comments here because this ticket is still open. @codyfinn let me know if these changes are possible.

1. Change width of category dropdown so Media & Entertainment is not truncated when selected
2. Change search bar hint text to `Search for app name`
3. Decrease height of Apps window to 450px to optimize for lower resolution viewports

bnussman:
I seem to be getting info weird states when choosing a category and searching at the same time.
- When I clear the search bar, I wouldn't expect the category to automatically clear
- When I clear both the search bar and the category, the sections such as `New apps` are gone and I would expect them to be there


https://user-images.githubusercontent.com/6440455/184375526-dc8c7920-d12a-4465-8800-ca0735caef63.mov

codyfinn:
> I seem to be getting info weird states when choosing a category and searching at the same time.
> 
>     * When I clear the search bar, I wouldn't expect the category to automatically clear
> 
>     * When I clear both the search bar and the category, the sections such as `New apps` are gone and I would expect them to be there
> 
> 
> Screen.Recording.2022-08-12.at.10.28.51.AM.mov

The behavior shown is intended. Currently you can only search by name of filter by category. Not both at the same time

bnussman (commented):
In our loading state, this section should not be scrollable

https://user-images.githubusercontent.com/6440455/184386439-fd808ddc-432e-4595-a164-51e7d4965a95.mov

jdamore-linode:
I'm noticing an issue that's preventing me from resetting my search. To reproduce:

1. Select any category
2. Without clearing the category, type in a search query and press the enter key
3. Observe that your search results appear and the category field is cleared as expected
4. Clear the search field and press the enter key
5. Observe that your search is not reset as expected

If you don't select a category first, or if you clear our your category selection before typing a search query, this issue doesn't occur.

As far as the UX goes, I was kind of thrown off by the fact that the only way to reset your search is to manually clear the fields (and in the case of the search field, submit the cleared value). I'm wondering if having some kind of reset link or button would be more intuitive?

codyfinn:
> I'm noticing an issue that's preventing me from resetting my search. To reproduce:
> 
>     1. Select any category
> 
>     2. Without clearing the category, type in a search query and press the enter key
> 
>     3. Observe that your search results appear and the category field is cleared as expected
> 
>     4. Clear the search field and press the enter key
> 
>     5. Observe that your search is not reset as expected
> 
> 
> If you don't select a category first, or if you clear our your category selection before typing a search query, this issue doesn't occur.
> 
> As far as the UX goes, I was kind of thrown off by the fact that the only way to reset your search is to manually clear the fields (and in the case of the search field, submit the cleared value). I'm wondering if having some kind of reset link or button would be more intuitive?

My latest commit should fix this. Though I am curious if it also fixes the issue you are having where you need to press `Enter` in order to get the query to filter the apps. Let me know if this fixes that as well.

packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
```diff
@@ -148,6 +203,56 @@ class FromAppsContent extends React.PureComponent<CombinedProps, State> {
...
+        (instance) => {
+          return appLabels.includes(instance.label.trim());
+        }
+      );
+      this.setState({
+        categoryFilter: categoryItem,
+        isSearching: false,
+        query: '',
+        filteredApps: instancesInCategory ?? [],
+        isFiltering: categoryItem !== null,
```
DevDW, line 239: Maybe explicitly say `true` here since it's the same condition from the `if` statement the code block is within? and line 247 could explicitly be set to `false`?

packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
```diff
@@ -148,6 +203,56 @@ class FromAppsContent extends React.PureComponent<CombinedProps, State> {
...
+        query: '',
+        // filteredApps: instancesInCategory ?? [],
+        isFiltering: categoryItem !== null,
+      });
+    }
+  };
+
+  shouldComponentUpdate() {
+    return true;
+  }
```
DevDW, line 254: What is making this necessary?

packages/manager/src/features/StackScripts/UserDefinedFieldsPanel/UserDefinedFieldsPanel.tsx
```diff
@@ -57,92 +57,72 @@ interface Props {
...
+      </Grid>
+    );
+  }
+  if (isPasswordField(field.name)) {
     return (
       <Grid item xs={12} lg={5} key={field.name}>
         <UserDefinedText
-          /** see comment above for why we're passing the value prop */
+          /**
+           * we explicitly passing the value to solve for the situation
```
DevDW, line 106: "we explicitly pass [...]"

packages/manager/src/features/StackScripts/UserDefinedFieldsPanel/UserDefinedFieldsPanel.tsx
```diff
@@ -151,14 +131,41 @@ const UserDefinedFieldsPanel = (props: Props) => {
...
-    if (openDrawer !== undefined) {
-      return openDrawer(selectedLabel);
-    }
-    return null;
-  };
+const handleOpenDrawer = (
+  openDrawer: CombinedProps['openDrawer'],
+  selectedLabel: string
+) => () => {
+  openDrawer?.(selectedLabel);
```
DevDW, line 155: this change brings back the linter complaint

Screen Shot 2022-08-16 at 3 38 16 PM

DevDW (commented):
The functionality is looking good from my testing, nice job 👍

jdamore-linode:
> My latest commit should fix this. Though I am curious if it also fixes the issue you are having where you need to press `Enter` in order to get the query to filter the apps. Let me know if this fixes that as well.

Nice! Thanks Cody, confirming that both issues mentioned here are fixed! I'll keep looking at this today after stand up.

bnussman (commented):
Functionality is looking good! I have two UX related things that are not super urgent

In the loading state, I think it is weird that it says `Select (an) App` two times. I think the second `Select App` Should be hidden if the spinner is present. 
Screen Shot 2022-08-17 at 10 51 21 AM


I think it would be good if we could remove this extra white space when searching or filtering
Screen Shot 2022-08-17 at 10 52 43 AM

packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
codyfinn: Nothing it was left over from debugging

packages/manager/src/features/StackScripts/UserDefinedFieldsPanel/UserDefinedFieldsPanel.tsx
DevDW, line 155: just resurfacing my previous comment: https://github.com/linode/manager/pull/8483#discussion_r947184815

packages/manager/src/features/StackScripts/UserDefinedFieldsPanel/UserDefinedFieldsPanel.tsx
```diff
@@ -232,7 +241,7 @@ const isMultiSelect = (udf: UserDefinedField) => {
  *
  * @return nested array [[...requiredUDFs], [...nonRequiredUDFs]]
  */
-const seperateUDFsByRequiredStatus = (udfs: UserDefinedField[]) => {
+const seperateUDFsByRequiredStatus = (udfs: UserDefinedField[] = []) => {
```
DevDW, line 244: `separateUDFsByRequiredStatus`

packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
```diff
@@ -148,6 +203,44 @@ class FromAppsContent extends React.PureComponent<CombinedProps, State> {
...
+    } else {
+      const appsMatchingQuery = this.props.appInstances?.filter((app) =>
+        app.label.toLowerCase().includes(query.toLowerCase())
+      );
+      this.setState({
+        isFiltering: false,
+        filteredApps: appsMatchingQuery,
+        isSearching: true,
+        categoryFilter: null,
+        query: query,
```
DevDW, line 218: can go with the property shorthand here

DevDW (approved):
LGTM aside from a few minor comments!

> Functionality is looking good! I have two UX related things that are not super urgent
> 
> In the loading state, I think it is weird that it says `Select (an) App` two times. I think the second `Select App` Should be hidden if the spinner is present. Screen Shot 2022-08-17 at 10 51 21 AM

Piggybacking off of this a bit, it's also visible in the error state:

Screen Shot 2022-08-23 at 4 40 37 PM

bnussman (approved):
Approving for functionality! Nice job 🎉 We can refine more at a later date

PR #8491, opened 2022-08-10
M3-5714: Generate Jest coverage report via GitHub Actions

DevDW:
## Description
Generate a Jest coverage report when running the CI workflow via GitHub Actions and save it as an artifact.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,7 +123,11 @@ jobs:
           name: packages-api-v4-lib
           path: packages/api-v4
       - run: yarn --frozen-lockfile
-      - run: yarn workspace linode-manager run test
+      - run: yarn workspace linode-manager run test --coverage
+      - uses: actions/upload-artifact@v2
+        with:
+          name: unit-test-coverage-report
+          path: packages/manager
 
   build-manager:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -124,14 +124,16 @@ jobs:
           path: packages/api-v4
       - run: yarn --frozen-lockfile
       - run: yarn workspace linode-manager run test --coverage --outputFile=coverage_report.json --json
+      - run: cp packages/manager/coverage/lcov.info packages/manager
       - uses: actions/upload-artifact@v2
         with:
           name: unit-test-coverage-report
           path: packages/manager/coverage_report.json
       - uses: romeovs/lcov-reporter-action@v0.3.1
         with:
-          lcov-file: packages/manager/coverage/lcov.info
+          lcov-file: packages/manager/lcov.info
           github-token: "${{ secrets.GITHUB_TOKEN }}"
+      - run: rm packages/manager/lcov.info
 
   build-manager:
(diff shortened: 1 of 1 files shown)


bnussman:
Really neat! I noticed that it seemed to add 15 min to the `test-manager` stage of CI which is a bit unfortunate, but not a big deal for now imo

DevDW:
> Really neat! I noticed that it seemed to add 15 min to the `test-manager` stage of CI which is a bit unfortunate, but not a big deal for now imo

I'll look into seeing if we can shorten that time, but one big issue with the initial attempt was the size of the artifact (70+ MB) because it took all the contents of `packages/manager` and spit it out in the artifact. Going to see if the previous commit makes any difference on that front, but once that's sorted I'll circle back to the time considerations

DevDW:
Looks like this adds anywhere from 6-9 minutes to how long the checks take to complete. The artifact also comes in at just under 8 MB.

jdamore-linode:
I think this is a valid tradeoff -- there aren't a lot of cases where we're anxiously waiting for GHA to finish, and if it does end up becoming a burden we can reevaluate.

bnussman (commented):
Nice job! I see the coverage artifact! Is there an easy way to consume and view the json? Did we explore outputting html?

DevDW:
> Nice job! I see the coverage artifact! Is there an easy way to consume and view the json? Did we explore outputting html?

Thanks! I could be wrong, but I don't think we can get HTML output easily -- I don't see a flag for it in https://jestjs.io/docs/cli. I tried it manually just in case it was valid but undocumented and the command (`yarn workspace linode-manager run test --coverage --outputFile=coverage_report.html --html`) failed

jdamore-linode:
> > Nice job! I see the coverage artifact! Is there an easy way to consume and view the json? Did we explore outputting html?
> 
> Thanks! I could be wrong, but I don't think we can get HTML output easily -- I don't see a flag for it in https://jestjs.io/docs/cli. I tried it manually just in case it was valid but undocumented and the command (`yarn workspace linode-manager run test --coverage --outputFile=coverage_report.html --html`) failed

There is an HTML report which is pretty nice, but it includes hundreds (thousands?) of small files and stacks up to around 30MB. It gets generated at `packages/manager/coverage/lcov-report` if you run `yarn test --coverage`, but I don't know how to generate _only_ the HTML report.

bnussman (approved):
Approving because it works, but I'd like to know how we're gonna utilize this data!

jdamore-linode:
Perhaps we can export it as JUnit XML and use something like [JUnit Report Action](https://github.com/marketplace/actions/junit-report-action)?

DevDW:
I was initially thinking the "Resource not accessible by integration" error was related to the `lcov.info` file being in a directory ignored by git, but the error looks to be permissions-related at the GHA integration level: https://www.google.com/search?q=%22Resource+not+accessible+by+integration%22+jest+site%3Agithub.com

Going to continue investigating but this may present a problem:

> Pull requests from public forks are still considered a special case and will receive a read token regardless of these settings.
(from https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/)

PR #8492, opened 2022-08-10
M3-5839: Add More Insightful Credit Card Expiration Date Validation

bnussman:
## Description 📝

- The goal of this PR is to decrease friction when inputting a Credit Card's expiration date
  - Support more formats
  - Allow user to submit form at any time (don't disable button by default)
  - Add validation message when there is no delimiter (`/`)

### The Issue ❗️
- Users were confused when they entered something like `052025` as a credit card expiration date and the `Add Credit Card` button was still disabled

## The Fix 🔧
- Allow users to push the `Add Credit Card` button with a partial expiry inputed so they can get some validation feedback
- If the user does not include some kind of delimiter to separate the month and year, show a validation message (`Date format must have forward slash between the month and year.`)

## Additional Feature ✨
- Users can now user characters other than just a `/` to seperate the month and year
  - The following are now acceptable values for the expiry date: `5 25`, `5-25`, `05 2025`, `05 25`, ...

## Preview 📷

https://user-images.githubusercontent.com/6440455/183980648-81b6fa72-8646-407b-b054-7113c810c063.mov

## How to test 🧪

- **Ensure you can still add a Credit Card to your Linode account**
- Check overall validation of the credit card form
- Ensure when  a user enters something like `052026`, that they are presented with a descriptive error message telling them how to resolve the formatting issue


HanaXu (commented):
How difficult would it be for us to persist the placeholder MM/YY as the user types? (e.g. `MM/YY` -> `0M/YY` -> `05/YY` -> `05/2Y` -> `05/25`)

And thoughts about automatically adding the `/`?

bnussman:
@HanaXu Great feedback, thank you! 

- Matthew and I considered adding the `/` for the user but we don't want Cloud Manager to become "too helpful" to the user. Also, by automatically adding the `/`, we would have to force the user to type a two digit month and we currenly allow a single digit month. 
- I'll look into persisting the placeholder

HanaXu:
@bnussman Gotcha, personally I don't think automatically adding `/` would make Cloud "too helpful" as it's a really minor user improvement. And I think the tradeoff is worth forcing the user to type a two digit month

bnussman:
@HanaXu I don't think persisting the placeholder will be an option, but let's discuss automatically adding the `/` in our next team meeting

packages/manager/src/features/Billing/BillingPanels/PaymentInfoPanel/AddPaymentMethodDrawer/AddCreditCardForm.tsx
```diff
@@ -113,17 +117,32 @@ const AddCreditCardForm: React.FC<Props> = (props) => {
       country: '',
     },
     onSubmit: addCreditCard,
-    validationSchema: CreditCardSchema,
+    validate: async (values) => {
+      const expiryValue = expiryRef?.current?.value;
```
bnussman, line 121: We do this wacky validation here because with Formik here, we have two values: one for month, one for year. We have no way to know the actual text input values with this approach. This is why I used a ref to the expiry input.

DevDW (commented):
There's a case where you address the `/` issue but an inaccurate error about the expiration month seems to persist:

https://user-images.githubusercontent.com/32860776/184725584-f90bc0f7-8e26-444b-8297-9854a29bdf77.mov

bnussman:
@DevDW I believe it resolves itself when you hit the Add button. I think that's just the nature of how Formik's validation works

DevDW (commented):
> @DevDW I believe it resolves itself when you hit the Add button. I think that's just the nature of how Formik's validation works

I was experimenting with the code some more and I think what I observed is because of this line:

https://github.com/linode/manager/blob/d3dbaff0e18713010dd4a2869e27de759bfb0776/packages/manager/src/features/Billing/BillingPanels/PaymentInfoPanel/AddPaymentMethodDrawer/AddCreditCardForm.tsx#L182

for the sake of testing, if you comment that out and try to go through the same steps as in the last video, that inaccurate error doesn't seem to rear its head.

In this scenario, maybe we could look at attempting that non-ideal approach we discussed last week, where we grab the first two numbers as the month and the remaining numbers for the year in the absence of a delimiter? I don't seem to be running into that inaccurate error with this:

```
              if (delimiter?.[0]) {
                const values: string[] = e.target.value.split(delimiter[0]);
                setFieldValue('expiry_month', values[0]);
                setFieldValue('expiry_year', parseExpiryYear(values[1]));
              } else {
                setFieldValue('expiry_month', e.target.value.slice(0, 2));
                setFieldValue(
                  'expiry_year',
                  parseExpiryYear(e.target.value.slice(2))
                );
              }
```

As a side note -- is the `'Expiration year must be 2 for 4 digits.'` error in `account.schema.ts` supposed to say "or" instead of "for"?

bnussman:
@DevDW That proposed solution suffers from similar behavior I believe. We need to keep the `setFieldValue('expiry_month', value);` you referenced so the `Add Credit Card` button becomes active when the user enters something into the expiry field. 

https://user-images.githubusercontent.com/6440455/184977141-f45b7086-3176-4c8a-bc81-8a18be55ad2a.mov

jdamore-linode:
Sorry if this has already been discussed or nixed -- feel free to disregard if so. If we're already working against the grain with Formik and the API expecting the month and year to be separate values, would there be any opportunity to simplify the UX and our code by splitting this into two fields?

Forgive the crappy mockup:
split-exp

My thoughts are that this simplifies our logic and helps convey to the user what is expected for each value, without really hindering the UX.

- We can take better advantage of `maxlength`, `max`, `min`, and `pattern` attributes on the inputs (which can also be `type="Number"` now)
- We don't have to worry about the presence of a delimiter
- UX niceties don't require very complicated logic
  - If the month input is only 1 char when the field loses focus, prepend `0`
  - If the year input is only 2 chars when the field loses focus, prepend `20`
  - If the user types in a valid month, automatically switch focus to the year field(?)

Any thoughts?

packages/manager/src/features/Billing/BillingPanels/PaymentInfoPanel/AddPaymentMethodDrawer/AddCreditCardForm.tsx
codyfinn: Would it be better to join the two formik fields together? We don't have to have them split for any particular reason since we can split them when we actually submit the form.

packages/manager/src/features/Billing/BillingPanels/PaymentInfoPanel/AddPaymentMethodDrawer/AddCreditCardForm.tsx
```diff
@@ -149,11 +168,19 @@ const AddCreditCardForm: React.FC<Props> = (props) => {
...
+              const value = e.target.value;
+              const delimiter = getExpirationDelimiter(value);
+
+              if (delimiter?.[0]) {
+                const values: string[] = e.target.value.split(delimiter[0]);
+                setFieldValue('expiry_month', values[0]);
+                setFieldValue('expiry_year', parseExpiryYear(values[1]));
+              } else {
+                setFieldValue('expiry_month', value);
+              }
```
codyfinn, line 183: Could this be pulled out into a function above instead of being defined on the element itself?

bnussman:
@jdamore-linode We decided to just use the validation error for now after exploring all the options. We think that following this update, we can consider a larger change.

packages/manager/src/features/Billing/BillingPanels/PaymentInfoPanel/AddPaymentMethodDrawer/AddCreditCardForm.tsx
```diff
@@ -91,6 +92,9 @@ const AddCreditCardForm: React.FC<Props> = (props) => {
   //   };
   // });
 
+  const getExpirationDelimiter = (value: string | undefined) =>
+    value?.match(/[^$,.\d]/);
+
```
HanaXu, line 97: We can move this function outside of the component

HanaXu (approved):
LGTM, great idea on allowing multiple delimiters! 🚀

PR #8497, opened 2022-08-15
M3-5919: Update Storybook and Add New Webpack 5 Loader

bnussman:
## Description 📝

- We need to update Storybook to v5 so it has the ability understand our new `exports` field in our `@linode/api-v4` and `@linode/validation` packages following #8484 
- Updates Storybook to `latest`

### Technical Explanation
- With #8484, we now emit only singular `index.js` files so when older compilers try to resolve paths like `@linode/api-v4/lib/linodes`, the actual file `lib/linodes.js` does not exist now!
- We now leverage the `exports` field in our `package.json`s to alias paths such as `@linode/api-v4/lib/**` to the `index.js` file. Webpack 4 does not use or understand  the `exports` field in our `package.json` so that's why we need this upgrade. 

## Preview Link 📷
https://m3-5919-update-storybook-to-.manager-ec2.pages.dev/

## How to test 🧪

- Verify that you can use the Storybook dev server with `yarn storybook`
- Verify that you can build Storybook with `yarn workspace linode-manager run build-storybook`


codyfinn (approved):

DevDW (approved):
LGTM 🚢

PR #8499, opened 2022-08-15
M3-5880: "Any/All" option in StackScript Create flow

DevDW:
## Description
Add an "Any/All" option in the StackScript Create flow in the "Target Images" dropdown.

## How to test
On the StackScript Create page, select "Any/All" for the "Target Images" dropdown. Ensure that no other images can be added. Also test the reverse, i.e., selecting other images and then selecting "Any/All"—doing the latter should clear the other selections out.

Confirm that creation goes through as expected.


bnussman (approved):

HanaXu (approved):
Creation went through 🚢

PR #8506, opened 2022-08-22
`master` --> `develop` following release of Cloud Manager v1.74.0

jdamore-linode:
Merge `master` into `develop` following the release of Cloud Manager v1.74.0


DevDW (approved):
In the docs for the release process, would you mind updating the step(s) you found ambiguous to specify that we should do merges (and not squash & merges)?

HanaXu (approved):

PR #8507, opened 2022-08-24
Hotfix StackScripts User Defined Field Input Bug

bnussman:
## Description

- Fixes bug relating to UDF Input fields clearing when Deploying a Linode from an Account or Community Stackscript that has UDFs
- This bug was my bad. I rewrote a Class component to a functional component and overlooked the fact that the class component was a PureComponent which handles prop updates differently

## How to test

- Test Deploying a Account or Community Stackscript at `http://localhost:3000/linodes/create?type=StackScripts&subtype=Community` and ensure all types of UDF field configurations work
- Ensure the new `any/all` image option still works as expected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [2022-08-24] - v1.74.1
+
+### Fixed:
+- Issue causing user-defined fields to clear erroneously during Linode StackScript deploy
+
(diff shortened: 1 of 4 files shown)


jdamore-linode (dismissed):
Nice! Confirmed that the issue with the user defined fields is fixed (I used the `spiraxcarts123 / squid-proxy-script|user-password` script for testing in case anyone is looking for one to use to reproduce the issue). The `any/all` image option also seems to be working -- I created a stackscript with `any/all` selected, and then was able to deploy a Linode with an arbitrarily selected image using that script.

codyfinn (approved):

DevDW (dismissed):

codyfinn (approved):

DevDW (approved):

PR #8508, opened 2022-08-24
M3-5901: Support Pro Dedicated Plans

HanaXu:
## Description
Add support for `prodedicated` plan types in the Linode Create flow

image

## How to test
Go to `/linodes/create` while pointing to dev or turn on the mocks

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
@@ -142,7 +142,7 @@ const getProDedicated = (types: PlanSelectionType[]) =>
   types.filter((t: PlanSelectionType) => /prodedicated/.test(t.class));
 
 const getDedicated = (types: PlanSelectionType[]) =>
-  types.filter((t: PlanSelectionType) => /dedicated/.test(t.class));
+  types.filter((t: PlanSelectionType) => /^dedicated/.test(t.class));
 
 const getGPU = (types: PlanSelectionType[]) =>
   types.filter((t: PlanSelectionType) => /gpu/.test(t.class));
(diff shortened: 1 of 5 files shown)


packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
```diff
@@ -410,6 +414,24 @@ export const SelectPlanPanel: React.FC<CombinedProps> = (props) => {
...
 
+    if (!isEmpty(proDedicated)) {
+      tabs.push({
+        render: () => {
+          return (
+            <>
+              <Typography data-qa-prodedi className={classes.copy}>
+                Pro Dedicated CPU instances are for very large plans. They only
+                have AMD 2nd generation processors or above.
+              </Typography>
```
bnussman, line 425: Is this copy temporary? It does not sound right to me. I left what I think sounds better below.

```suggestion
              <Typography data-qa-prodedi className={classes.copy}>
                Pro Dedicated CPU instances are for very demanding workloads. They
                have AMD Epyc 2nd generation processors or newer.
              </Typography>
```

bnussman (changes requested):
I don't think `Pro Dedicated CPU` should be the first item in the plans table. Can it go after `Dedicated CPU`? If a customer saw a $960 monthly price as the first plan we show, they may be very disturbed.

packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
HanaXu: Not sure if `AMD Epyc` has been finalized so just going to leave it as `AMD 2nd generation` for now

DevDW (approved):
LGTM assuming the copy has been ran by stakeholders (even if it's still subject to change at the moment)

jdamore-linode (approved):
Looks good!

PR #8510, opened 2022-08-24
M3-5635: Volume Status on Volumes landing

bnussman:
## Description 📝

- Adds Volume Status to the Volumes Landing Page
- Adds initial support to pre-populate the support ticket modal with a pre-selected entity

## Preview 📷

Screen Shot 2022-08-24 at 6 23 50 PM

## How to test 🧪

- Check Volumes Landing for Volume statuses
- Use the MSW to see all possible states

### DIFF
--- a/packages/manager/src/features/Volumes/VolumesLanding.tsx
+++ b/packages/manager/src/features/Volumes/VolumesLanding.tsx
@@ -110,7 +110,7 @@ export const volumeHeaders = [
     label: 'Label',
     dataColumn: 'label',
     sortable: true,
-    widthPercent: 30,
+    widthPercent: 35,
   },
   {
     label: 'Status',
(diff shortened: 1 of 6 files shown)


packages/manager/src/features/Volumes/VolumesLanding.tsx
```diff
@@ -110,7 +110,13 @@ export const volumeHeaders = [
     label: 'Label',
     dataColumn: 'label',
     sortable: true,
-    widthPercent: 40,
+    widthPercent: 30,
```
DevDW, line 113: What about adjusting the width of the "Attached To" column a bit as well as or instead of the "Label" column? This could add some more visual balance to the table, right now it's rather heavy on the left

Screen Shot 2022-08-25 at 11 18 12 AM

DevDW (approved):

DevDW:
Another thing is that we surprisingly aren't hiding columns as the viewport width shrinks for this table. Addressing that may be outside the scope of this particular ticket, but we should get one created to look into addressing that

HanaXu (approved):

PR #8512, opened 2022-08-25
M3-5911: Marketplace UI Refinement: Searching and filtering analytics

codyfinn:
## Description

Adds google analytics events for search and filtering on the marketplace page.

## How to test

1. Navigate to the marketplace page.
2. Click on the app search bar.
3. Ensure that a GA event is sent in the network tab of the developer tools
4. Click the category dropdown
5. Select a category
6. Ensure that a GA event is sent in the network tab of the developer tools

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
@@ -226,10 +226,10 @@ class FromAppsContent extends React.Component<CombinedProps, State> {
   };
 
   handleSelectCategory = (categoryItem: Item<AppCategory>) => {
-    sendMarketplaceSearchEvent('Category Dropdown', categoryItem.label);
     const didUserSelectCategory = categoryItem !== null;
     let instancesInCategory: StackScript[] | undefined = [];
     if (didUserSelectCategory) {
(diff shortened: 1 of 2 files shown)


packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
```diff
@@ -221,6 +226,7 @@ class FromAppsContent extends React.Component<CombinedProps, State> {
   };
 
   handleSelectCategory = (categoryItem: Item<AppCategory>) => {
+    sendMarketplaceSearchEvent('Category Dropdown', categoryItem.label);
```
bnussman, line 229: Clearing the category gives an error. This optional chain should fix it! Maybe we shouldn't send an event if the input is cleared? 

https://user-images.githubusercontent.com/6440455/187231818-5ec7f11a-a970-4d6b-9fc1-e0a0dec3ecd3.mov


```suggestion
    sendMarketplaceSearchEvent('Category Dropdown', categoryItem?.label);
```

bnussman (changes requested):

DevDW (commented):
I agree with not sending an event when the category is cleared

DevDW (approved):

bnussman (approved):

PR #8513, opened 2022-08-25
M3-5908: Remove Type-to-Confirm checkbox from modals

DevDW:
## Description
Remove the "Disable type-to-confirm" checkbox from modals and replace it with text and a link to "My Settings".

The TTC setting will now be controlled solely from the "My Settings" page.

## How to test
Confirm that:

1. When TTC is enabled, you see the text field and `To disable type-to-confirm, go to the Type-to-Confirm section of My Settings.` below it
2. When TTC is disabled, you see only the text `To enable type-to-confirm, go to the Type-to-Confirm section of My Settings.`

in the modals for the Rebuild Linode, Resize Linode, Delete Kubernetes Cluster, Delete Bucket, Restore Database from Backup, and Delete Database flows. The Close Account flow should have the TTC regardless of user preference.

### DIFF
--- a/packages/manager/src/components/TypeToConfirm/TypeToConfirm.tsx
+++ b/packages/manager/src/components/TypeToConfirm/TypeToConfirm.tsx
@@ -1,18 +1,13 @@
 import * as React from 'react';
-import CheckBox from 'src/components/CheckBox';
-import FormControlLabel from 'src/components/core/FormControlLabel';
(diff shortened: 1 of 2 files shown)


bnussman (approved):

HanaXu (approved):
We should remove the line about disabling TTC in the close account modal since it will always have TTC

image

PR #8520, opened 2022-08-31
M3-5503: Add script to increment package versions during release

jdamore-linode:
## Description

**What does this PR do?**
This adds a small script to increment our package versions during the release process. It can be executed like this:

```
yarn package-versions
```
You'll be prompted to specify a version for each package. Alternatively, if you already know which version each package should be, you can invoke it non-interactively like this:

```
yarn package-versions <manager-version> <api-version> <validation-version> -f
```

In either case, a summary will be displayed which reflects the changes made by the script.

Because we now use `*` to specify workspace package dependency versions, no changes get made to the package `dependencies` fields.

This probably could have been done more concisely or with a fancier UI, but I wanted the script to run without any additional dependencies so that running `yarn install` beforehand would not be necessary.

## How to test

Run the command with a variety of option/flag combinations to ensure that the behavior works as expected:

Runs the script interactively:
```
yarn package-versions
```

Runs the script interactively, with Cloud Manager `v1.75.0`, API v4 `v0.77.0`, and Validation `v0.14.0` specified by default:
```
yarn package-versions 1.75.0 0.77.0 0.14.0
```

Runs the script non-interactively, updating Cloud Manager to `v1.75.0`, APIv4 to `v0.77.0`, and Validation to `v0.14.0`:
```
yarn package-versions 1.75.0 0.77.0 0.14.0 -f
```

Runs the script non-interactively, only specify Cloud Manager `v1.75.0`, which should result in an error:
```
yarn package-versions 1.75.0 -f
```

Running the script will modify your `package.json` files, so you'll want to discard those changes afterwards.


scripts/package-versions/index.js
```diff
@@ -0,0 +1,199 @@
...
+      })();
+
+      const prompt = `New version for '${jobName}' package? (${ suggestedVersion })\n`;
+      result = await readlinePromise(prompt);
+      if (!result) {
+        result = suggestedVersion;
+      }
+    }
+    else {
+      if (!jobVersion) {
```
codyfinn, line 162: This needs to happen sooner. Currently running `yarn package-versions 1.75.0 -f` updates the cloud version then fails. We need to make sure that all of the versions are supplied before we start making changes.

codyfinn (changes requested):

scripts/package-versions/index.js
codyfinn: Also it is not uncommon for us to only bump one package when publishing. Should we have throw an error at all?

HanaXu: Agreed with @jdamore-linode that running it interactively would be the most likely (and simplest) use-case

HanaXu (approved):
This is great, I like the addition of the simple UI! 🚀 🤖 

When this gets merged, can you update step 3 of our Deploy docs with these changes?

scripts/package-versions/index.js
codyfinn: I would recommend removing the `-f` flag for the time being then. As is this still will say it failed but change the version numbers for the packages provided (either just `manager` or `manager and api`. That or refactor the code to correctly handle invalid input.

codyfinn (changes requested):

scripts/package-versions/index.js
jdamore-linode: Thank you @codyfinn and @HanaXu for the feedback! Just pushed some changes -- I ended up keeping the `-f` flag in place, but changed the error handling behavior to be more in line with Cody's suggestion.

It'll only write to the package files once it's processed the CLI input, so if the user only provides 1 or 2 arguments and passes the `-f` flag, it'll display an error message but will not write to the package files. There are cases where the user can provide valid input but the script can fail to write the changes (if, for example, there's a file permissions error), and that can still lead to a situation where some of the package files get modified but others don't.

codyfinn (approved):

PR #8521, opened 2022-09-01
M3-5850: Add Cypress tests for two-factor authentication

jdamore-linode:
## Description

**What does this PR do?**
Adds Cypress integration tests for 2FA. Confirms UI flows for enabling, disabling, and resetting 2FA, and confirms that 2FA cannot be enabled when security questions are not answered.

## How to test

**How do I run relevant tests?**
`yarn up`, and then:

```
yarn cy:run -s "cypress/integration/account/two-factor-auth.spec.ts"
```

Or, if you want to run them interactively:

```
yarn cy:debug
```
And search for `two-factor-auth.spec.ts`.

### DIFF
--- a/packages/manager/cypress/integration/account/two-factor-auth.spec.ts
+++ b/packages/manager/cypress/integration/account/two-factor-auth.spec.ts
@@ -2,40 +2,156 @@
  * @file Integration tests for account two-factor authentication functionality.
  */
 
-import { randomNumber, randomLabel, randomString } from 'support/util/random';
+import { SecurityQuestionsData } from '@linode/api-v4/types';
+import {
+  profileFactory,
+  securityQuestionsFactory,
+} from 'src/factories/profile';
 import {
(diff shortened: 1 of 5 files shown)


bnussman (commented):
I saw `POST /v4/profile/tfa-enable 400` and I think it has to do with Security Questions.

Screen Shot 2022-09-01 at 2 11 50 PM

Should this be handled by the test or is there an assumption that my account should security questions completed?

jdamore-linode:
Ah, nice catch! Glad you caught that! This is because I opted not to mock responses for the 2FA enable API request and overlooked the fact that the API will error if security questions aren't set. I'll fix this shortly

jdamore-linode:
@bnussman think that should be fixed now, thanks again!

bnussman (approved):

DevDW (approved):

HanaXu (approved):

PR #8522, opened 2022-09-01
M3-5927: Allow Deletion of Private IPv4 Addresses

bnussman:
## Description 📝

- Show a `Delete` button for Private IPv4 addresses on a Linode

### Before 

Screen Shot 2022-09-01 at 1 47 19 PM

### After

Screen Shot 2022-09-01 at 1 47 44 PM

## How to test 🧪

- Spin up a Linode in dev
- Add a Private IP address
- Verify that you see a `Delete` Button and that you can successfully delete the private IPv4 address


jdamore-linode (approved):
Ran into the same issue you did where I initially got the "Only public IPv4 addresses can be deleted" error, but the changes look good on our side!

DevDW (approved):
🚢

PR #8523, opened 2022-09-01
M3-5843: Allow each state to have their own date for a Tax Collection Banner

codyfinn:
## Description

Changes the format we expect from the `taxCollectionBanner` flag to allow regions within the same country to have different dates. Also only displays the banner for up to 5 weeks after the date specified in the flag data. The new data shape keeps the top level `date` field. This is now used as the default date for all regions in the list. The region list has been changed from a list of strings to a list of objects. The objects have two properties `name` and `date`. The `name` field is the string that we use to denote the region. The `date` field is used to specify a date that is different from the default one specified for the country. If a region has a date specified that will always override the country level one.  

## How to test

Check ticket for testing procedure

### DIFF
--- a/packages/manager/src/featureFlags.ts
+++ b/packages/manager/src/featureFlags.ts
@@ -8,10 +8,15 @@ interface TaxBanner {
   linode_tax_id?: string;
 }
 
+interface TaxCollectionRegion {
+  name: string;
+  date?: string;
+}
+
 interface TaxCollectionBanner {
   date: string;
   action?: boolean;
-  regions?: string[];
+  regions?: TaxCollectionRegion[];
 }
 
 type OneClickApp = Record<string, string>;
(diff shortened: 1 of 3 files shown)


patthiel:
@codyfinn is this ready for review?

codyfinn:
> @codyfinn is this ready for review?

Yes

jdamore-linode (approved):
Looks good! Checked the states mentioned in the ticket and saw the expected notices (or lack of notice, in PA's case). E2E tests involving the billing page are also passing.

My only question is whether the date should have a comma between the day of the month and the year (e.g. `August 1, 2022` vs `August 1 2022`) -- only asking because I notice that the example data in the ticket does have the comma

codyfinn:
> Looks good! Checked the states mentioned in the ticket and saw the expected notices (or lack of notice, in PA's case). E2E tests involving the billing page are also passing.
> 
> My only question is whether the date should have a comma between the day of the month and the year (e.g. `August 1, 2022` vs `August 1 2022`) -- only asking because I notice that the example data in the ticket does have the comma

The comma was just a typo on my part. But the banner will display whatever string is passed

DevDW (approved):
LGTM 🚢

PR #8526, opened 2022-09-06
M3-5787: Improve Timezone Offsets by Pulling Them From `Luxon`

bnussman:
## Description 📝

- Pulls timezone offsets from [Luxon](https://github.com/moment/luxon/) so that they are more accurate
  - Previously, we pulled them from the static `timezones.ts` file, which did not incorporate **Daylight Savings Time**
- Rewrites `TimezoneForm.tsx` to be a FC and use hooks
- Rewrites `TimezoneForm.test.tsx` to use React Testing library and newer patterns

### Before

> **Note**: Notice how the offset for EST is -5

Screen Shot 2022-09-06 at 1 50 22 PM

### After

> **Note**: Notice how the offset for EST is -4. This is because it is taking into account the fact we are current in daylight savings time. If you do some Googling, you will see that this is currently the correct offset

Screen Shot 2022-09-06 at 1 49 59 PM
 
## How to test 🧪

- Go to `http://localhost:3000/profile/display` 
- Check functionality of the timezone input for general use
- Look at the offsets in the dropdown and compare them to what other sources say online about what the current offset is
- It may be useful to compare this PR to production to see how the offsets are different. You should see different offset values for many places

`yarn test packages/manager/src/features/Profile/DisplaySettings/TimezoneForm.test.tsx`

### DIFF
--- a/packages/manager/src/features/Profile/DisplaySettings/TimezoneForm.tsx
+++ b/packages/manager/src/features/Profile/DisplaySettings/TimezoneForm.tsx
@@ -41,8 +41,7 @@ interface Timezone {
   offset: number;
 }
 
-export const formatOffset = ({ label, name }: Timezone) => {
-  const offset = DateTime.now().setZone(name).offset;
+export const formatOffset = ({ label, offset }: Timezone) => {
   const minutes = (Math.abs(offset) % 60).toLocaleString(undefined, {
     minimumIntegerDigits: 2,
     useGrouping: false,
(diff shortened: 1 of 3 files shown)


jdamore-linode (approved):
Looks good! Confirmed that the timezone offsets have changed, the `TimezoneForm.test.tsx` tests pass, and the account/profile related E2E tests still pass.

I did notice that before this change, the timezone drop-down items were sorted by their offset, from GMT -11:00 to GMT +14:00. Since the offsets have changed, but the order of the items has not, they're no longer sorted from lowest offset to highest offset. I don't think this is a big problem, but just wanted to point it out in case having those items sorted by offset is expected for this component!

packages/manager/src/features/Profile/DisplaySettings/TimezoneForm.tsx
```diff
@@ -62,159 +41,104 @@ interface Timezone {
...
 
 const renderTimeZonesList = (): Item[] => {
   return timezones.map((tz: Timezone) => {
-    const label = formatOffset(tz.offset, tz.label);
+    const label = formatOffset(tz);
     return { label, value: tz.name };
   });
 };
 
 const timezoneList = renderTimeZonesList();
```
DevDW, line 63: since `timezoneList` will have some timezones that are out of order now, can we use `.sort()` with an appropriate sorting function in `renderTimeZones()` to return the expected order?

DevDW (commented):
Looks good from my testing -- I do think it's worth getting the list sorted properly so everything is where users would expect them to be. Also, these changes move to a field error as opposed to an error notice in that section, which I think is fine and more in line with our current patterns.

codyfinn (approved):
I agree with @DevDW that we should sort them but all the functionality is there. Nice job

DevDW (approved):
Looks great, thanks for sorting the options! 🚢

PR #8529, opened 2022-09-09
M3-5931: Fix database maintenance window mapping and database_update notification message

DevDW:
## Description

- Adjust day mapping so that Monday through Sunday is 1 through 7, respectively

- Change `status` field in `eventMessageCreators` from `notification` to `finished` for `database_update` notification, per confirmation with the API team. This ensures the correctly-worded notification is displayed in the drawer and on the Events landing page.
- Move database-related events in `eventMessageCreators` so that the list is (more) alphabetical order


bnussman (approved):
Thanks for addressing the event as part of this!

HanaXu (approved):

PR #8531, opened 2022-09-12
M3-5913: Replace `novnc-node` with `react-vnc`

bnussman:
## Description 📝 

- `novnc-node` was last updated 7 years ago so we are replacing it with a more modern and supported alternative
- `react-vnc` will play better with Vite when we make the switch
- Cleans up a lot of old code and simplifies logic while retaining all functionality
- Fixes deep linking so you can go to `http://localhost:3000/linodes/:id/lish/glish` and see Glish instead of Weblish

## How to test 🧪

- Test **Lish** and **Glish** functionality on Linodes
  - Check that it is usable for a long time because tokens expire quickly and Cloud should handle reconnecting when a new token is needed
  - Test with Powered Off Linode
  - Compare to Production and check for any regressions in functionality


jdamore-linode:
Looking good for the most part! Only encountered one issue -- if the Lish/Glish window is not in the foreground, after a few minutes a spinner appears as if it's disconnected, and the only way to reconnect is to click whichever tab is not currently selected, then click back to the tab you were on, at which point it appears to resume your session:

Screen Shot 2022-09-14 at 9 56 44 AM

I wasn't able to reproduce this behavior against Prod -- it appears to remain connected even when the window is in the background for a while.

That's the only issue I ran into! Everything else is looking good, and this change also fixes the horizontal scrollbar which currently appears (erroneously) at the bottom of the window when using Lish.

jdamore-linode (approved):
It looks like your latest commit fixed the issue I was seeing with Lish disconnecting in the background!

DevDW (commented):
I want to comb through the code a bit more, but so far I'm not seeing any regressions from my testing. I did notice that the new queries don't seem to be showing up in the RQ Dev Tool though, are you seeing that as well?

bnussman:
@DevDW I moved the `<ReactQueryDevtools />` so the Lish window will show its queries.

DevDW (approved):
Looks good from my testing, thanks for adding the RQ Dev Tool there 👌

PR #8536, opened 2022-09-13
M3-4935: Add initrd selection to Linode Config modal

DevDW:
## Description
Users were able to, in Classic Manager (and presently via the API), select a disk for `initrd`. This PR adds a dropdown in the Linode Config modal in the "Block Device Assignment" section for initrd. The dropdown is populated by disks with an initrd filesystem.

## To-Do
- [ ] Fix bug where initrd somehow gets set to `null` when you change and save the new disk multiple times in quick succession
- [x] Explanatory PR comments
- [x] Confirm with API that the `initrd` should not explicitly be a number/integer (right now a `typeof` is not specified in the API mapping)

## How to test
- Create a few disks with a filesystem type of "initrd"
- Add or edit a Linode Configuration. Confirm that the "initrd" field is visible in the modal and that it is populated with any disks on that Linode with a filesystem type of "initrd".
- Select a disk and save the changes. Go through the modal flow a few times to ensure the correct disk is saved for initrd and shows up the next time you open the modal.

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
@@ -109,7 +109,7 @@ interface EditableFields {
   useCustomRoot: boolean;
   label: string;
   devices: DevicesAsStrings;
-  initrd: string | null;
+  initrd: string | number | null;
   kernel?: string;
   comments?: string;
   memory_limit?: number;
@@ -329,6 +329,11 @@ const LinodeConfigDialog: React.FC<CombinedProps> = (props) => {
 
     const configData = convertStateToData(values) as LinodeConfigCreationData;
 
+    // If Finnix was selected, make sure it gets sent as a number in the payload, not a string.
+    if (Number(configData.initrd) === 25669) {
+      configData.initrd = 25669;
+    }
+
     if (!regionHasVLANS) {
       delete configData.interfaces;
     }
@@ -468,19 +473,22 @@ const LinodeConfigDialog: React.FC<CombinedProps> = (props) => {
       return {
         label: categoryTitle,
         value: category,
-        options: items.map(({ label, id }) => {
-          return {
-            label,
-            value: String(id) as string | null,
-          };
-        }),
+        options: [
+          ...items.map(({ label, id }) => {
+            return {
+              label,
+              value: String(id) as string | number | null,
+            };
+          }),
+          { label: 'Recovery – Finnix (initrd)', value: '25669' },
+        ],
       };
     }
   );
 
   categorizedInitrdOptions.unshift({
-    value: '',
     label: '',
+    value: '',
     options: [{ label: 'None', value: null }],
   });
 
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
@@ -223,6 +223,7 @@ const LinodeConfigDialog: React.FC<CombinedProps> = (props) => {
     open,
     onClose,
     config,
+    initrdFromConfig,
     kernels,
     linodeConfigId,
     linodeRegion,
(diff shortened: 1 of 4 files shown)


packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
```diff
@@ -1021,6 +1066,8 @@ const enhanced = compose<CombinedProps, Props>(
       ? state.__resources.linodeConfigs[linodeId].itemsById[linodeConfigId]
       : undefined;
 
+    const initrdFromConfig = config?.initrd ? String(config.initrd) : '';
```
DevDW, line 1069: I added this because `config.initrd` comes in as a `number` initially, but when the user clicks "Save Changes," it needs to be submitted as a `string`. Converting it to a string as early as possible helps for the sake of value comparisons throughout the file.

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
```diff
@@ -322,6 +327,9 @@ const LinodeConfigDialog: React.FC<CombinedProps> = (props) => {
 
     const configData = convertStateToData(values) as LinodeConfigCreationData;
 
+    console.log(typeof configData.initrd);
+    console.log(configData.initrd);
+
```
DevDW, line 332: these `console.log()`'s show that at the time the values are submitted, the datum is a string and is not undefined or null.

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
```diff
@@ -491,6 +511,15 @@ const LinodeConfigDialog: React.FC<CombinedProps> = (props) => {
     [setFieldValue]
   );
 
+  const handleInitrdChange = React.useCallback(
+    (selectedDisk: Item<string>) => {
+      console.log(selectedDisk);
+      setFieldValue('initrd', selectedDisk.value);
+      console.log(values);
```
DevDW, line 518: these `console.log()`'s show that when a disk is selected, the selected value is not undefined nor null, and neither is `values.initrd`.

DevDW:
The bug I described in demo prep still exists, but I'm going to go ahead and remove the WIP label here. I may be overlooking what's causing the issue but I'm starting to think it might be something on the back-end because the payloads go through as expected.

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
codyfinn: could `25669` be set as a constant so that it is less of a magic number?

codyfinn (approved):
Left a small request but the code works from what I tested

jdamore-linode:
Encountered a couple small issues, but they may be related to the issue you're already aware of (and if not, might be API issues anyhow)

- After setting up a few initrd disks and creating a configuration, I select one of the initrd disks under "Block Assignment / initrd" and click "Add Configuration", then edit the configuration and observe that the value under "initrd" is "None". Updating or re-setting it seems to work fine, however. (This is the one that I think is the same issue you've described)
- This might just be inherent to initrd disks, but it seems like I'm unable to resize them once they're created. Whether I try to make them bigger or larger, I always seem to get this message in the notifications pane: _A disk on Linode cy-test-initrd could not be resized by prod-test-019._

Since I think the first issue is one you're already aware of, and the second is most likely a backend issue, I'm still gonna approve this -- just wanted to bring attention to these issues just in case.

bnussman (approved):

DevDW:
> Encountered a couple small issues, but they may be related to the issue you're already aware of (and if not, might be API issues anyhow)
> 
> * After setting up a few initrd disks and creating a configuration, I select one of the initrd disks under "Block Assignment / initrd" and click "Add Configuration", then edit the configuration and observe that the value under "initrd" is "None". Updating or re-setting it seems to work fine, however. (This is the one that I think is the same issue you've described)
> * This might just be inherent to initrd disks, but it seems like I'm unable to resize them once they're created. Whether I try to make them bigger or larger, I always seem to get this message in the notifications pane: _A disk on Linode cy-test-initrd could not be resized by prod-test-019._
> 
> Since I think the first issue is one you're already aware of, and the second is most likely a backend issue, I'm still gonna approve this -- just wanted to bring attention to these issues just in case.

The first point is tied into the issue I referenced -- not sure about the second point but that would be related to either a back-end or technical limitation with initrd disks

PR #8538, opened 2022-09-14
Linode create Google Analytics timing POC

codyfinn:
## Description

adds Google Analytics timing to track how long a user takes to successfully create a Linode.
This is achieved by using an instance variable on the `LinodeCreateContainer` class that is set to the current UNIX timestamp on instantiation. Then when the user submits the form we calculate the number of difference between that time and the current time and send that to Google Analytics via the `timing()` function from the `react-ga` package.

## How to test

1. Have a stopwatch ready
2. Have the developer tools open and on the Networking tab
3. Start the stopwatch when you navigate to the Linode create page
4. Fill out the form to create a Linode
5. Stop the stopwatch when you click the `Create Linode` button
6. Ensure that upon successfully creating a Linode a Google Analytics timing is sent and that the value in the params specified as `utt` is equal to the number of milliseconds you timed on your stopwatch

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/LinodeCreateContainer.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/LinodeCreateContainer.tsx
@@ -76,6 +76,7 @@ import {
   queryKey,
   reportAgreementSigningError,
 } from 'src/queries/accountAgreements';
+import { timing } from 'react-ga';
 
 const DEFAULT_IMAGE = 'linode/debian11';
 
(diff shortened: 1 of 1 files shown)


jdamore-linode:
This is cool! Just went through the flow and confirmed that the amount of time spent was tracked and that the `utt` value matched (give or take half a second) my phone's stop watch

bnussman-akamai:
Closing for now due to inactivity. We can revisit this PR if we need this!

PR #8546, opened 2022-09-21
M3-5937: Add Firewall Support for `IPENCAP` Protocol

bnussman:
## Description 📝

- Adds support for `IPENCAP` in Cloud Manager
  - 🌎[Linode API Documentation on Firewall Rules](https://www.linode.com/docs/api/networking/#firewall-rules-update)

## Preview 📷
Screen Shot 2022-09-21 at 11 15 43 AM

## How to test 🧪

- Create or have a Linode Firewall
- Add a rule to that Firewall that uses `IPENCAP` as the `Protocol`
- Confirm that you can add an `IPENCAP` firewall rule
  - Test many different configurations of firewall rules and ensure Cloud Manager's validation holds up
  - Make sure you are pressing `Save Changes` to commit your firewall changes and get validation from the API

### DIFF
--- a/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.tsx
+++ b/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.tsx
@@ -332,7 +332,7 @@ const FirewallRuleForm: React.FC<FirewallRuleFormProps> = React.memo(
         }
 
         setFieldValue('protocol', item?.value);
-        if (item?.value === 'ICMP') {
+        if (item?.value === 'ICMP' || item?.value === 'IPENCAP') {
           // Submitting the form with ICMP and defined ports causes an error
           setFieldValue('ports', '');
           setPresetPorts([]);
--- a/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.tsx
+++ b/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.tsx
@@ -469,12 +469,11 @@ const FirewallRuleForm: React.FC<FirewallRuleFormProps> = React.memo(
           value={presetPorts}
           options={portOptions}
           onChange={handlePortPresetChange}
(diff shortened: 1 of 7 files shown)


packages/manager/src/features/Firewalls/FirewallDetail/Rules/firewallRuleEditor.ts
```diff
@@ -246,7 +246,10 @@ export const removeICMPPort = (
   rules: ExtendedFirewallRule[]
 ): ExtendedFirewallRule[] =>
   rules.map((thisRule) => {
-    if (thisRule.protocol === 'ICMP' && thisRule.ports === '') {
+    if (
+      (thisRule.protocol === 'ICMP' || thisRule.protocol === 'IPENCAP') &&
```
bnussman, line 250: I could do something like this, but I'm going to leave what I have because it's only two items right now. In the future, a change like this could be made if we add more protocols. 

```suggestion
      ['ICMP', 'IPENCAP'].includes(thisRule.protocol) &&
```

packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.tsx
```diff
@@ -332,7 +332,7 @@ const FirewallRuleForm: React.FC<FirewallRuleFormProps> = React.memo(
         }
 
         setFieldValue('protocol', item?.value);
-        if (item?.value === 'ICMP') {
+        if (item?.value === 'ICMP' || item?.value === 'IPENCAP') {
```
DevDW, line 335: let's update the comment below accordingly

DevDW (commented):
Looking good overall -- can we get the same behavior for the Ports field when IPENCAP is selected as when ICMP is selected?

Screen Shot 2022-09-22 at 12 43 08 PM

Screen Shot 2022-09-22 at 12 43 15 PM

DevDW (approved):

codyfinn (approved):

patthiel:
cc. @thorn3r !

PR #8553, opened 2022-09-26
M3-5903: Add ARIA labels for network transfer history previous/next buttons

jdamore-linode:
## Description

**What does this PR do?**

- Adds ARIA labels to the previous/next buttons on the Network Transfer History graph shown on the "Network" tab of the Linode details page
    - The labels are formatted _Show network transfer history for \<Month\> \<Year\>_, or _Show network transfer history for last 30 days_ (whichever is applicable) -- very open to feedback for this language
- Disables the previous and next buttons when there is no data left to navigate to
    - We already apply a disabled class to the buttons in these situations, this change just makes it apparent to screen reader users that the buttons are disabled
- Modifies a utility function and updates tests accordingly

## How to test

To verify button ARIA label and disabled state:
1. Enable mock service workers, and navigate to a Linode details page "Network" tab
2. Confirm that the Network Transfer History graph previous button has an ARIA label that reads _Show network transfer history for \<Previous Month\> \<Previous Year\>_
3. Confirm that the Network Transfer History graph previous button is enabled, except when you navigate all the way back to the end of the history (January 2020)
4. Confirm that the Network Transfer History graph next button has an ARIA label that reads _Show network transfer history for \<Next Month\> \<Next Year\>_
5. Confirm that the Network Transfer History graph next button is disabled only when the most recent data is shown on the graph
6. (Optional) [Enable VoiceOver](https://support.apple.com/guide/voiceover/turn-voiceover-on-or-off-vo2682/mac), navigate to Network Transfer History buttons, verify that VoiceOver/keyboard navigation works as expected and has a good UX
    * In my experience, these changes make it so users navigating using keyboard + screen reader cannot focus the previous or next button if it is disabled. If the user has either button focused and then navigates to a point where the button becomes disabled, the button will remain focused but the user cannot click it (the screen reader reports "Dimmed" in this case).

```
yarn test TransferHistory.test
```

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/NetworkingSummaryPanel/TransferHistory.test.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/NetworkingSummaryPanel/TransferHistory.test.tsx
@@ -39,30 +39,60 @@ describe('combineGraphData', () => {
 
 const now = DateTime.fromISO('2020-07-01T12:00:00');
 
-describe('getYearAndMonthFromOffset', () => {
-  it('returns the current year and month with an offset is 0', () => {
-    const { year, month, humanizedDate } = parseMonthOffset(0, now);
+describe('parseMonthOffset', () => {
+  it('returns the current year and month with an offset of 0', () => {
+    const { year, month, humanizedDate, longHumanizedDate } = parseMonthOffset(
+      0,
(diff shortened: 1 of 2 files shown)


DevDW (approved):
Looks good 🚢  

++ for the detailed testing instructions, especially for the VoiceOver piece

bnussman (approved):

PR #8556, opened 2022-09-27
M3-5950: Fix Object Storage overwrite error

HanaXu:
## Description
Fix Cloud Manager crashing when overwriting an Object Storage file that already exists. This was because the `foundFile` variable was being defined and used in two different cases, but was not being correctly scoped with curly braces.

## How to test
1. Upload an object to a bucket.
2. Upload it again.
3. Click `Replace`


jdamore-linode (approved):
Nice, confirmed this fixes the issue!

bnussman (approved):
I was able to replace a file without issue! 🚀

PR #8558, opened 2022-09-29
M3-5552: Fix Linode Add Disk Drawer UI width bug

mjac0bs:
## Description 📝

**What does this PR do?**
Fixes the bug causing misalignment of Password and SSH Key fields by removing the default MUI theme padding.
For consistency, applied a bottom margin to match the top margin once bottom padding was removed.

## Preview 📷
fixedDrawerMargins
buggyDrawerPadding

## How to test 🧪
Using the browser Developer Tools, view the component with the React Developer Tools > Components tab and search for 'AccessPanel'. Verify that the component's padding has been removed and a bottom margin added in its place.

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/ImageAndPassword.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/ImageAndPassword.tsx
@@ -1,3 +1,4 @@
+import classNames from 'classnames';
 import { GrantLevel } from '@linode/api-v4/lib/account';
 import * as React from 'react';
 import { compose } from 'recompose';
@@ -7,10 +8,28 @@ import withImages, { WithImages } from 'src/containers/withImages.container';
 import { ImageSelect } from 'src/features/Images';
 import { withLinodeDetailContext } from '../linodeDetailContext';
 import LinodePermissionsError from '../LinodePermissionsError';
+import {
+  createStyles,
+  Theme,
+  withStyles,
+  WithStyles,
+} from 'src/components/core/styles';
 
 interface ContextProps {
   permissions: GrantLevel;
 }
+
+type ClassNames = 'root';
+
+const styles = (theme: Theme) =>
+  createStyles({
+    root: {
+      margin: `${theme.spacing(3)}px 0px ${theme.spacing(3)}px 0px`,
+      padding: 0,
+    },
+  });
+
+const styled = withStyles(styles);
 
 interface Props {
   onImageChange: (selected: Item<string>) => void;
@@ -23,12 +42,14 @@ interface Props {
   userSSHKeys: UserSSHKeyObject[];
   requestKeys: () => void;
   sshError?: string;
+  className?: string;
 }
 
-type CombinedProps = Props & ContextProps & WithImages;
+type CombinedProps = Props & ContextProps & WithImages & WithStyles<ClassNames>;
 
 export const ImageAndPassword: React.FC<CombinedProps> = (props) => {
   const {
+    classes,
     imagesData,
     imagesError,
     imageFieldError,
@@ -40,6 +61,7 @@ export const ImageAndPassword: React.FC<CombinedProps> = (props) => {
     userSSHKeys,
     sshError,
     permissions,
+    className,
   } = props;
 
   const disabled = permissions === 'read_only';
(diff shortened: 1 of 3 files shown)


DevDW (changes requested):
Nice work getting your first PR up! 🎉 

The `<AccessPanel />` component is used in a few different files (you can do a keyword search in VSC for "<AccessPanel" to see which ones). Although the changes you made work in the "Add Disk" drawer, in some other contexts the formatting has become misaligned, for example in the Linode Create flow:

Screen Shot 2022-09-30 at 11 36 43 AM

One option would be for us to make a change at a level closer to this specific individual instance the component is used in, perhaps adding some styling in `ImageAndPassword.tsx`, for example.

mjac0bs:
> Nice work getting your first PR up! 🎉
> 
> The `<AccessPanel />` component is used in a few different files (you can do a keyword search in VSC for "<AccessPanel" to see which ones). Although the changes you made work in the "Add Disk" drawer, in some other contexts the formatting has become misaligned, for example in the Linode Create flow:
> 
> Screen Shot 2022-09-30 at 11 36 43 AM
> 
> One option would be for us to make a change at a level closer to this specific individual instance the component is used in, perhaps adding some styling in `ImageAndPassword.tsx`, for example.

Thanks for catching that @DevDW! 

I've now checked that the Create Linode page, the Add Disk Drawer, and the Rebuild from Image or StackScript, where the AccessPanel is used, all seem to be aligned correctly with this updated change. I went ahead and moved the styling to the ImageAndPassword component. Let me know if there are any further revisions I should make there; I modeled my styling changes after the styling in AccessPanel.

Screen Shot 2022-09-30 at 12 18 02 PM
Screen Shot 2022-09-30 at 12 18 18 PM
Screen Shot 2022-09-30 at 12 20 09 PM

DevDW (commented):
Thanks for making that adjustment!

One further improvement we can make from a code pattern perspective is to make use of the `useStyles()` hook instead of `withStyles` wrapper.

The reason why `AccessPanel.tsx` uses the `withStyles` wrapper is because it is currently written as a class component, so hooks can't be used in it. However, we don't have that restriction here in `ImageAndPassword.tsx` since it is a function component.

In general, our current pattern preference is to write function components instead of class components, and to use hooks where possible over wrappers.

A related file that is a function component and uses the `useStyles()` hook that you can refer to as an example is `RebuildFromImage.tsx`.

mjac0bs:
@DevDW Makes a lot of sense! Latest commit reflects those code pattern improvements.

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/ImageAndPassword.tsx
```diff
@@ -23,6 +31,7 @@ interface Props {
   userSSHKeys: UserSSHKeyObject[];
   requestKeys: () => void;
   sshError?: string;
+  className?: string;
```
DevDW, line 34: Given the current approach, we can get rid of this prop

DevDW (approved):
Looks good! 🚀 I left a comment re: one final cleanup item

mjac0bs:
@jdamore-linode 

Could you give this a once over and let me know if my updates to `ImageAndPassword.test.tsx` look correct? I noticed that the test-manager build failed and found this error that seemed to be causing it in the [logs](https://github.com/linode/manager/actions/runs/3174976305/jobs/5172588427):

``` FAIL  src/features/linodes/LinodesDetail/LinodeSettings/ImageAndPassword.test.tsx
  ● Test suite failed to run

    TypeError: theme.spacing is not a function

      12 | const useStyles = makeStyles((theme: Theme) => ({
      13 |   root: {
    > 14 |     margin: `${theme.spacing(3)}px 0 ${theme.spacing(3)}px 0`,
         |                      ^
      15 |     padding: 0,
      16 |   },
      17 | }));
```

I updated the unit test to include the `wrapWithTheme` helper function, like `RebuildFromImage.test.tsx` and re-ran the unit tests locally, and no longer noticed the failure.

jdamore-linode:
@mjac0bs Nicely done, that is correct!

Without `wrapWithTheme(...)`, the `theme` object that gets passed to your `makeStyles(...)` callback is just `{}`, so that's why you were seeing an error saying that `theme.spacing()` doesn't exist. Any component with custom styling that depends on the `theme` object will need to use `wrapWithTheme()` in its tests :)

It's also worth mentioning that this is an older unit test that uses [enzyme](https://www.npmjs.com/package/enzyme) rather than React Testing Library. That fact isn't really relevant to your question, just pointing it out since some of the patterns you see in our older tests won't line up exactly with the patterns we use in our newer tests!

jdamore-linode (approved):
Looks great! 🚢 Nicely done @mjac0bs!

PR #8559, opened 2022-09-29
M3-5941: Linode landing page – Empty state

codyfinn:
## Description 📝

Adds helpful links to the bottom of the Database and Linodes empty landing pages. These links also send GA events.

## Preview 📷

Databases Before:
image
Databases After: 
image

Linodes Before:
image
Linodes After:
image

## How to test 🧪

Ensure that the landing pages for Databases and Linodes are present and take the user to the correct location. In addition check the network tab to ensure that GA events are being sent.


packages/manager/src/features/linodes/LinodesLanding/ListLinodesEmptyState.tsx
```diff
@@ -2,65 +2,126 @@ import * as React from 'react';
...
+          <LinksSubSection
+            title="Getting Started Guides"
+            icon={<DocsIcon />}
+            moreLink={
+              <Link to="https://www.linode.com/docs/">
+                Check out all our Docs <PointerIcon />
+              </Link>
+            }
+          >
+            <List>
```
DevDW, line 66: similar feedback as above re: `.map()`'ing

packages/manager/src/features/linodes/LinodesLanding/ListLinodesEmptyState.tsx
```diff
@@ -2,65 +2,126 @@ import * as React from 'react';
...
+          >
+            <AppsSection />
+          </LinksSubSection>
+          <LinksSubSection
+            title="Getting Started Playlist"
+            icon={<YoutubeIcon />}
+            moreLink={
+              <Link to="https://www.youtube.com/playlist?list=PLTnRtjQN5ieb4XyvC9OUhp7nxzBENgCxJ">
+                View the complete playlist{' '}
+                <ExternalLinkIcon style={{ marginLeft: 8 }} />
```
DevDW, line 121: could be worth making use of `useStyles()` and putting these inline styles into classes

packages/manager/src/features/Databases/DatabaseLanding/DatabaseEmptyState.tsx
```diff
@@ -69,24 +78,80 @@ const DatabaseEmptyState: React.FC = () => {
...
+                  </Link>
+                </ListItem>
+              </List>
+            </LinkSubSection>
+            <LinkSubSection
+              title="Getting Started Playlist"
+              icon={<YoutubeIcon />}
+              moreLink={
+                <Link to="">
+                  View the complete playlist
```
DevDW, line 142: need a `{' '}` here

packages/manager/src/features/Databases/DatabaseLanding/DatabaseEmptyState.tsx
```diff
@@ -69,24 +78,80 @@ const DatabaseEmptyState: React.FC = () => {
...
+                    Check out all out Docs
+                  </Link>
+                </ListItem>
+              </List>
+            </LinkSubSection>
+            <LinkSubSection
+              title="Getting Started Playlist"
+              icon={<YoutubeIcon />}
+              moreLink={
+                <Link to="">
```
DevDW: playlist link?

packages/manager/src/features/Databases/DatabaseLanding/DatabaseEmptyState.tsx
```diff
@@ -69,24 +78,80 @@ const DatabaseEmptyState: React.FC = () => {
...
+                    Overview of Managed Databases
+                  </Link>
+                </ListItem>
+                <ListItem>
+                  <Link to="https://www.linode.com/docs/products/databases/managed-databases/get-started/">
+                    Get Started with Managed Databases
+                  </Link>
+                </ListItem>
+                <ListItem>
+                  <Link to="https://www.linode.com/docs/products/databases/managed-databases/get-started/">
```
DevDW: Should this be `https://www.linode.com/docs/products/databases/managed-databases/guides/database-engines/` instead?

packages/manager/src/features/Databases/DatabaseLanding/DatabaseEmptyState.tsx
```diff
@@ -69,24 +78,80 @@ const DatabaseEmptyState: React.FC = () => {
...
+                    Choosing a Database Engine
+                  </Link>
+                </ListItem>
+                <ListItem>
+                  <Link to="https://www.linode.com/docs/">
+                    Check out all out Docs
+                  </Link>
+                </ListItem>
```
DevDW, line 134: looks like this last `ListItem` is an accidental leftover

packages/manager/src/features/linodes/LinodesLanding/ListLinodesEmptyState.tsx
```diff
@@ -2,65 +2,126 @@ import * as React from 'react';
...
+              </Link>
+            }
+          >
+            <List>
+              <ListItem>
```
DevDW, line 126: these are all linking to docs still instead of the YouTube videos

DevDW (commented):
left some feedback in addition to what was pointed out in the demo today -- I would say that time permitting, some of the refactoring I mention in the comments can be done in this PR, but if there are time constraints early next week then it can possibly be done in a follow-up PR instead.

packages/manager/src/features/Databases/DatabaseLanding/DatabaseEmptyState.tsx
DevDW: Still need to add the space

DevDW (commented):
The functionality is looking good.

There's still unaddressed feedback from my last review; I'm all right with the more involved refactoring being held off on until a follow-up PR but let's get the less involved pieces squared away (feel free to resolve the comment convos as you address them)

packages/manager/src/features/Databases/DatabaseLanding/DatabaseEmptyState.tsx
codyfinn: Applying the `marginLeft` rule instead to be the same as other instances. Later this should be refactored into the `useStyles` hook as you mentioned before.

codyfinn:
>     1. The demo feedback re: section header and header icon horizontal alignment
We decided leave the icon alignment as is since the icons will never look centered due to their proportions. We figured baseline was a better place to have the icons.

packages/manager/src/features/linodes/LinodesLanding/ListLinodesEmptyState.tsx
```diff
@@ -2,65 +2,175 @@ import * as React from 'react';
...
+      <ListItem key={linkData.to}>
+        <Link to={linkData.to} onClick={getLinkOnClick(linkData.text)}>
+          {linkData.text}
+        </Link>
+      </ListItem>
+    ))}
+  </List>
+);
+
+const YoutubeLinks = () => (
```
HanaXu, line 85: Can we have these return just JSX instead of a function?

HanaXu (changes requested):
The spacing between sections seem a bit tight and uneven in the Linode Empty State
image

Dark mode link contrast isn't great, can we update this to look like our other dark mode links? The Marketplace App list colors look off too, we use that styling for inputs but the app links aren't inputs.
image

Here's a dark mode link for reference
image

matthewbartholomew:
vertically center pointer icons on lower case letters (move them down 2px)
Screen Shot 2022-10-03 at 3 12 33 PM

packages/manager/src/features/linodes/LinodesLanding/AppsSection.tsx
```diff
@@ -0,0 +1,134 @@
...
+      <div className={classes.appLinkIcon}>
+        <PointerIcon />
+      </div>
+    </Link>
+  );
+};
+
+const appLinks = appsLinkData.map((linkData) => (
+  <AppLink {...linkData} key={linkData.to} />
+));
```
HanaXu, line 126: This is pretty simple so I think we can just inline this instead of defining a separate variable

HanaXu (approved):

DevDW (approved):
LGTM now, thanks for the cleanup 🚢

PR #8564, opened 2022-10-05
M3-5900: Visual indicator when navigating Linode and Database plan selection tables

DevDW:
## Description 📝
Currently in Firefox, if you try using the Tab key to navigate through the Linode or Database Create flow, you run into an issue when you reach the plan selection table: there is no indicator of which row has focus. You may have to press Tab many times to get through the listed plans and to the next field, but you do not know where you are in the table because of the lack of a visual indicator.

This PR addresses this by highlighting the row that has focus.

## Preview 📷
Screen Shot 2022-10-05 at 4 59 52 PM

Screen Shot 2022-10-05 at 5 00 41 PM

## How to test 🧪
In Chrome and Firefox, use the Tab key to navigate to and through the Linode and Database plan panels in the Create flow. You should observe a visual indicator for the row with focus. If you use the mouse to click the radio button for a row, you should also see the visual indicator.

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
@@ -85,6 +85,11 @@ const useStyles = makeStyles((theme: Theme) => ({
     width: '5%',
     paddingRight: 0,
   },
+  focusedRow: {
+    '&:focus-within': {
+      backgroundColor: theme.bg.lightBlue1,
+    },
+  },
   gpuGuideLink: {
     fontSize: '0.9em',
     '& a': {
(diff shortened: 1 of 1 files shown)


jdamore-linode (approved):
Nice! This looks really nice, and makes navigating the Linode Create page with keyboard so much easier!

packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx
```diff
@@ -212,6 +217,7 @@ export const SelectPlanPanel: React.FC<CombinedProps> = (props) => {
             }
             aria-disabled={isSamePlan || planTooSmall || isDisabledClass}
             className={classNames({
+              [classes.focusedRow]: true,
               [classes.disabledRow]:
                 isSamePlan || planTooSmall || isDisabledClass,
             })}
```
bnussman, line 223: I know this is pretty normal to see in Cloud, but I think we should start to do this when we intend to apply classes unconditionally. I'd like to hear your thoughts!

```suggestion
             className={classNames(classes.focusedRow, {
              [classes.disabledRow]:
                isSamePlan || planTooSmall || isDisabledClass,
            })}
```

bnussman (approved):

PR #6449, opened 2020-05-20
M3-4168 Intercept all verification is required errors and add a link to Support

Jskobos:
## Description

We previously added intercepting logic wherein if the user tried to create a GPU Linode and was unable to due to insufficient reputation, we added a link to Support. Jim has requested that we always do this for this error type. I'm calling this phase 1: any verification errors will have a link to open the ticket drawer, with "Verification request" pre-filled in the summary.

Phase 2 would involve trying to have more situation-specific messages for different situations. It's a little tricky, since we're already inspecting the request config to see what the error was about, since the API error doesn't tell us much.

## Notes

To test, set your account's reputation score to 1 in admin and try to create a Linode or Kubernetes cluster.

PR #8566, opened 2022-10-06
M3-5699: Add support ticket link to notice text

mjac0bs:
## Description 📝

**What does this PR do?**

Links the words "open a support ticket" in error notices to the support page with a modal for opening a new support ticket.

The code additions follow the existing error intercept pattern in `request.tsx` that replaces error text by matching on text from the API's error reason. Now, when any error has a reason that contains the text "open a support ticket", we replace it with the SupportError component. The SupportError component displays the text of the initial error reason, followed by text linking to a support ticket.

## Preview 📷

https://user-images.githubusercontent.com/114685994/194411756-d3fb1fc0-50b2-49e1-a7d2-c48f5c76759f.mov

## How to test 🧪

1. Set the account Thing Limit to 0.
2. Attempt to create a new linode (or firewall, nodebalancer, volume) by filling in the required information.
3. Click 'Create' and the error message "Account Limit reached. Please open a support ticket." should be visible in the notice at the top of the page.
4. Click the linked text in the notice and confirm you are redirected to the support page with an open support ticket modal.

Applies to testing linkable text across multiple error messages (to be included in follow up issue):
5. Reset Thing Limit back to original value. Set Reputation as 0.
6. Attempt to create a new linode of any size.
7. Check that an error message stating "Additional verification is required to add this service. Please open a support ticket" appears at the top of the screen.
8. Click the linked text in the notice and confirm you are redirected to the support page with an open support ticket modal.


mjac0bs:
It looks like linking to a support ticket is done in a couple of different ways throughout the app:
1. Via a Link from React Router
2. Via the SupportLink component
3. Via a button (in one case)

I followed the pattern using the SupportLink as it seemed more recent code does this, but feedback is welcome.

I also started with, then deviated slightly from, the pattern in `request.tsx`  where we matched on a condition that was specific enough to pre-populate the support ticket with a title and description fields based on the intercepted error (see Verification Error in [M3-4168](https://github.com/linode/manager/pull/6449#discussion_r433271851)). To get the desired behavior of linking on any message with the keywords "support ticket", I didn't see a reason or way to pre-populate a more genetic ticket with relevant information, but maybe others have better ideas.

DevDW (commented):
The functionality is working well from my testing and the approach makes sense. In prod currently, the text is bold–I think we'll want to retain that styling:

Screen Shot 2022-10-10 at 4 35 53 PM

HanaXu (changes requested):
It looks like this code might be breaking some changes I made in https://github.com/linode/manager/pull/7826

If you add these lines to `serverHandlers.ts`, turn on the mocks, and try to assign a Firewall to a Linode, the button is infinitely spinning and no error message is shown.

```typescript
  rest.post('*/firewalls/*/devices', (req, res, ctx) => {
    return res(
      ctx.status(400),
      ctx.json({
        errors: [
          {
            field: 'id',
            reason: `Linode singapore_firewall_02 (ID 42343243) is not compatible with Firewalls. Please initiate this Linode's pending migration or open a support ticket.`,
          },
        ],
      })
    );
  }),
```

This branch:

https://user-images.githubusercontent.com/14323019/195178184-a2740bd2-266e-47ea-8f89-6c17c4a18b9e.mov

However, if you add those lines and test on the develop branch, the error message displays as expected:

https://user-images.githubusercontent.com/14323019/195178217-1a5398e2-5112-4d9c-ac32-3e2ccf855559.mov

mjac0bs:
Somehow I managed to accidentally include other commits when pushing my changes... Sorry about that, will look into dealing with it so it doesn't happen again.

mjac0bs:
> The functionality is working well from my testing and the approach makes sense. In prod currently, the text is bold–I think we'll want to retain that styling:
> 

@DevDW Thanks for catching that. Addressed this by applying the same styling that applied to Notice's text.

mjac0bs:
> It looks like this code might be breaking some changes I made in #7826
> 

@HanaXu Can you have a look at my change in `request.tsx` that addressed this breaking change? Want to make sure this is a real fix for this edge case and not likely to create issues elsewhere down the line. 

I have noticed through testing that creating a new Volume or Firewall is not displaying the correct text (shown below), and I've tracked that down in `createVolumeForm.tsx` where the [API error response is caught and the generalError gets assigned](https://github.com/mjac0bs/manager/blob/M3-5699-notice-text-links/packages/manager/src/features/Volumes/VolumeCreate/CreateVolumeForm.tsx#L210-L222). generalError is later passed down as a prop to the NoticePanel component. With the changes in this PR, the errorResponse is a component (SupportError) rather than a string, so the component gets stringified rather than handled correctly.

I was wondering if I could do something like the code below to assign generalError to either a string or a component, but that wasn't quite right, so I'm hoping for other suggestions on how to handle cases like these.
 
```
 const mapErrorToStatus = (generalError: string | JSX.Element) =>
       setStatus({ generalError });
```

mjac0bs:
> > It looks like this code might be breaking some changes I made in #7826
> 
> 
> I have noticed through testing that creating a new Volume or Firewall is not displaying the correct text 

Latest commit fixes this on assignment of generalError's value by determining if an API error is a React element to avoid stringifying JSX. Currently, the conditional check assumes that the SupportError/JSX will be the first error in the apiErrors array, which could cause issues if this is not always the case (e.g. more than one API error).

HanaXu (commented):
Hmm I think making a blanket change to link to `open a support ticket` might be too big of a scope for this ticket as there are many edge cases to account for.

For example, the API may not always return `open a support ticket` as the last few words of an error message, but in the middle of the message instead. You can see this by trying to add another IPv4 address to a Linode in the Network tab of a Linode's detail page

Prod:
image

This branch:
image

It might be simpler to just specifically handle the account limit error for now

mjac0bs:
@HanaXu Agreed that there are many edge cases. And several ways in which errors are handled throughout the app. Do you have any recommendations on good ways to get more familiar with possible errors (specific files, API docs, etc.)?

After reverting back to just handling the Account Limit error, the edge cases mentioned above seem to be addressed. We still need the changes to `formikErrorUtils.ts` in order for the Account Limit error to display correctly for Volumes and Firewalls. 

I can create a follow up ticket to finish addressing the blanket change of "open a support ticket" text to keep this ticket at a smaller scope.

packages/manager/src/components/AccountLimitError.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+import { APIError } from '@linode/api-v4/lib/types';
+import Typography from 'src/components/core/Typography';
+import SupportLink from 'src/components/SupportLink';
+import { useStyles } from 'src/components/Notice/Notice';
+interface Props {
+  errors: APIError[];
+}
+
+export const AccountLimitError: React.FC<Props> = (props) => {
```
HanaXu, line 10: We can keep the component named `SupportError` as there's nothing specific to AccountLimit here

packages/manager/src/utilities/formikErrorUtils.ts
```diff
@@ -1,5 +1,6 @@
 import { APIError } from '@linode/api-v4/lib/types';
 import { reverse } from 'ramda';
+import * as React from 'react';
```
HanaXu, line 3: I'm hesitant to bring React into a TypeScript utility file. Since the error is displaying correctly for Linode Create, we should look at possible changes within Volume/Firewall Create first.

Changing the error handling to something like below fixes the display without touching the util file:
```typescript
import { getErrorMap } from 'src/utilities/errorUtils';

const mapErrorToStatus = () => setStatus({ generalError: getErrorMap([], errorResponse).none });
```

https://github.com/linode/manager/blob/develop/packages/manager/src/features/Volumes/VolumeCreate/CreateVolumeForm.tsx#L212-L213
https://github.com/linode/manager/blob/develop/packages/manager/src/features/Firewalls/FirewallLanding/AddFirewallDrawer.tsx#L89-L90

mjac0bs: @HanaXu Thanks for the suggested fix. Latest changes make this update to error handling at all places where it's relevant in volume and firewall create flows. Changes to the TypeScript utility file have been reverted.

HanaXu (approved):

PR #8567, opened 2022-10-06
M3-5764: Add HA chip to Kubernetes landing page

HanaXu:
## Description 📝
- Add HA chip to the Kubernetes landing page
- Update HA chip size in the Kubernetes details page to match the landing page chip
- Minor refactoring

## Preview 📷
image

image

## How to test 🧪
Check the view in `/kubernetes/clusters` and `/kubernetes/clusters/<cluster_id>/summary`
```
yarn test ClusterRow
```

### DIFF
--- a/packages/manager/src/features/Kubernetes/ClusterList/ClusterRow.tsx
+++ b/packages/manager/src/features/Kubernetes/ClusterList/ClusterRow.tsx
@@ -8,8 +8,6 @@ import Grid from 'src/components/Grid';
 import TableCell from 'src/components/TableCell';
 import TableRow from 'src/components/TableRow';
 import { dcDisplayNames } from 'src/constants';
-import { getKubeHighAvailability } from 'src/features/Kubernetes/kubeUtils';
-import { useAccount } from 'src/queries/account';
 import { ExtendedCluster, PoolNodeWithPrice } from './../types';
 import ActionMenu from './ClusterActionMenu';
 
@@ -44,6 +42,7 @@ const useStyles = makeStyles((theme: Theme) => ({
 export interface Props {
   cluster: ExtendedCluster;
   hasUpgrade: boolean;
+  isClusterHighlyAvailable: boolean;
   openDeleteDialog: (
     clusterID: number,
     clusterLabel: string,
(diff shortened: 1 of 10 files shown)


mjac0bs (approved):
Changes look good to me, though I don't have full code review approval yet.

The one thing I thought of, from the perspective of someone still new to Cloud Manager, is that a user might have to do some searching to find out what "HA" means, if they were unfamiliar with the abbreviation. (Not sure how often that could be the case.) I noticed that the LKE docs page does list High Availability as a LKE feature (without the abbreviation) and a search for HA on the Support page turns up results for HAProxy, which a user could extrapolate the abbreviation from. 

Not sure if it's against our design patterns to include tooltip text on hover over a chip, and it may be overkill in a situation like this, but as a user, I might expect to learn what "HA" is in that way.

packages/manager/src/features/Kubernetes/ClusterList/ClusterRow.tsx
```diff
@@ -54,9 +56,15 @@ type CombinedProps = Props;
...
 export const ClusterRow: React.FunctionComponent<CombinedProps> = (props) => {
   const classes = useStyles();
+  const { data: account } = useAccount();
 
   const { cluster, hasUpgrade, openDeleteDialog, openUpgradeDialog } = props;
 
+  const { isClusterHighlyAvailable } = getKubeHighAvailability(
+    account,
+    cluster
+  );
```
DevDW, line 66: I think it'll be more efficient if we refactor this by getting `isClusterHighlyAvailable` a level higher, in `ClusterList.tsx`, and adding it to the cluster object so that `ClusterRow.tsx` can just check the prop for it

DevDW (commented):
Looks good overall, I left a comment re: a refactor for efficiency/performance

matthewbartholomew:
I think it makes sense to have a tooltip definition of HA though it's likely an edge case. HA is offered and defined as part of the Kubernetes create flow and also when a user decides to upgrade an existing cluster to HA.

CodyFinn-Akamai (approved):

HanaXu:
Agreed with @matthewbartholomew that users being unfamiliar with HA is an edge case. I think a tooltip is unnecessary, but we can revisit this if we get confusion complaints.

DevDW (approved):
🚀

packages/manager/src/features/Kubernetes/ClusterList/ClusterRow.tsx
```diff
@@ -78,6 +90,17 @@ export const ClusterRow: React.FunctionComponent<CombinedProps> = (props) => {
...
             </div>
           </Grid>
+          {isClusterHighlyAvailable ? (
+            <Grid item>
+              <Chip
+                label="HA"
+                variant="outlined"
+                outlineColor="green"
+                size="small"
+                data-testid={'ha-chip'}
```
bnussman, line 100:
```suggestion
                data-testid="ha-chip"
```

bnussman (approved):

PR #8569, opened 2022-10-12
M3-5501: [BUILD] Update the changelog generation script

CodyFinn-Akamai:
## Description 📝

Updates the `generate_changelog.py` script in the following ways:
- Allows the user to specify what sub-repo the changelog should be generated for
- Adds the new naming conventions form PR name standardization proposal 
- Automatically strips Jira labels from changelog
- Automatically inserts current date

the syntax for calling the script has changed it should now take this form
`python generate_changelog.py <version number> <origin> <repo to generate changelog for>`
The repo argument is optional and only needed to specify `api-v4` or `validation`

## How to test 🧪

Ensure you can generate a changelog for the repos without error

### DIFF
--- a/scripts/generate_changelog.py
+++ b/scripts/generate_changelog.py
@@ -11,7 +11,7 @@ except Exception:
     REPO=False
 
 DATE = datetime.datetime.now().strftime('%Y-%m-%d')
-START_INSERT=5 if REPO.lower() == 'manager' else 0
+START_INSERT=0 if (REPO and REPO.lower() != 'manager') else 5
 
 NOT_INCLUDED_IN_LOG = []
 
--- a/scripts/generate_changelog.py
+++ b/scripts/generate_changelog.py
@@ -54,7 +54,7 @@ def remove_pull_request_id(commit_message):
     regexp = "\(#\d+\)"
     return re.sub(regexp, '', commit_message).strip()
 
-def generateChangeLog(release, date, origin, repo=False):
+def generateChangeLog(release, date, origin, repo=''):
     git_log_command = ["git", "log", "--no-merges", "--oneline", "--pretty='%s'", "{}/develop...HEAD".format(origin)]
 
     if (repo):
--- a/scripts/generate_changelog.py
+++ b/scripts/generate_changelog.py
@@ -74,11 +74,11 @@ def generateChangeLog(release, date, origin, repo=False):
     for commit in commits:
         commit = remove_pull_request_id(commit)
 
-        jira_key_regex=re.match('M3-\d{4}', commit)
+        jira_key_regex=re.match('M3-\d{4}: ', commit)
         if (jira_key_regex is not None):
             jira_key=jira_key_regex.group(0)
             jql_query.append(jira_key)
-            commit.lstrip(jira_key)
+            commit = commit.lstrip(jira_key)
 
         if(checkKeyWords(TEST_KEYWORDS, commit.lower())):
             NOT_INCLUDED_IN_LOG.append(commit)
--- a/scripts/generate_changelog.py
+++ b/scripts/generate_changelog.py
@@ -3,7 +3,7 @@ import re
 import sys
 import datetime
 
-RELEASE=sys.argv[1]
+RELEASE='v' + sys.argv[1].strip('v')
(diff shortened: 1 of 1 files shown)


dwiley-akamai (commented):
This is looking solid so far, but I am seeing the release header format in the changelog (e.g., `## [2022-09-19] - v1.76.0`) is different now. Across the packages' changelogs, the format now has the version first and then the date -- e.g., `## [1.79.0] - 2022-10-19`. We should keep it consistent with what it has been historically

scripts/generate_changelog.py
```diff
@@ -93,12 +101,19 @@ def generateChangeLog(release, date, origin):
...
+    else:
+        changelog_file = "packages/{}/CHANGELOG.md".format(repo)
+
+    read_change_log=open(changelog_file, 'r')
     change_log_lines=read_change_log.readlines()
     read_change_log.close()
 
     change_log_lines.insert(START_INSERT,'\n')
-    change_log_lines.insert(incrementLine(),'## [%s] - %s\n'%(release,date))
+    change_log_lines.insert(incrementLine(),'## [%s] - %s\n'%(date, release))
```
dwiley-akamai, line 116: We should either add the `v` prefix here and note in the documentation that we don't have to pass it in, or vice versa, don't add it here but note in the documentation that we'll have to pass the `v` prefix with the version number

dwiley-akamai (commented):
When I run the command without the `repo` argument, e.g. `python3 scripts/generate_changelog.py 1.79.0 origin`, I get this error:

Screen Shot 2022-10-20 at 11 40 14 AM

scripts/generate_changelog.py
CodyFinn-Akamai: What about allowing both? Then adding it if needed?

dwiley-akamai: That works for me, just want to ensure a consistent format and make sure it isn't prone to errors

scripts/generate_changelog.py
```diff
@@ -1,19 +1,24 @@
...
+RELEASE='v' + sys.argv[1].strip('v')
+ORIGIN=sys.argv[2]
+try:
+    REPO=sys.argv[3]
+except Exception:
+    REPO=False
 
-START_INSERT=5
+DATE = datetime.datetime.now().strftime('%Y-%m-%d')
+START_INSERT=5 if REPO.lower() == 'manager' else 0
```
dwiley-akamai, line 14:
```suggestion
START_INSERT=0 if (REPO and REPO.lower() != 'manager') else 5
```

scripts/generate_changelog.py
```diff
@@ -49,8 +54,11 @@ def remove_pull_request_id(commit_message):
     regexp = "\(#\d+\)"
     return re.sub(regexp, '', commit_message).strip()
 
-def generateChangeLog(release, date, origin):
-    git_log_command = ["git", "log", "--no-merges", "--oneline", "--pretty='%s'", "{}/master...HEAD".format(origin)]
+def generateChangeLog(release, date, origin, repo=False):
```
dwiley-akamai, line 57:
```suggestion
def generateChangeLog(release, date, origin, repo=''):
```

scripts/generate_changelog.py
```diff
@@ -93,12 +101,19 @@ def generateChangeLog(release, date, origin):
 
     generateJQLQuery(jql_query)
 
-    read_change_log=open('CHANGELOG.md', 'r')
+    changelog_file = "CHANGELOG.md"
+
+    if (repo == 'manager'):
```
dwiley-akamai, line 106:
```suggestion
    # If the repo isn't explicitly passed in, we default to Manager's CHANGELOG.
    if (repo == 'manager' or repo == '' or repo is False):
```

dwiley-akamai (commented):
LGTM aside from the error I mentioned in the last comment when `repo` isn't passed as an argument. The suggestions below seemed to work for me but may need tweaking or could be less verbose

dwiley-akamai (approved):
Everything looks to be working as intended, nice work 🚀

bnussman-akamai (approved):

PR #8575, opened 2022-10-19
Marketplace Update: add new logos Nextcloud & Plex

hmorris3293:
## Description 
Update logo for Nextcloud & Plex


hana-linode (commented):
Plex logo looks the same to me 🤔 

image

jdamore-linode (changes requested):
I'm seeing an issue with the [Plex SVG](https://raw.githubusercontent.com/linode/manager/396c60a2753790264beb389967309b881121f738/packages/manager/public/assets/plex.svg):

> This page contains the following errors:
>
> error on line 1 at column 222: Attribute height redefined
>
> Below is a rendering of the page up to the first error.

Removing either of the `height` attributes on the root `<svg>` element seems to fix it, but I'm not sure which value is the correct one.

bnussman-akamai:
@hmorris3293 do we still need to take action on updating these logos?

jaalah-akamai:
Correct logos were not provided. @hmorris3293 is going to follow up and re-submit if they ask again.

PR #8576, opened 2022-10-19
M3-5920: Make info button accessible via keyboard nav

mjac0bs:
## Description 📝

**What does this PR do?**
Makes the info button on Marketplace app cards keyboard navigable. Users can now select the button and expand the drawer to access information about the application.

## Preview 📷

https://user-images.githubusercontent.com/114685994/196762382-90ef0a0b-5101-4ae6-808f-b5836b54aebd.mov

## How to test 🧪

1. Visit the Marketplace.
2. Use keyboard navigation to tab to a Marketplace app selection card.
3. Once a card is selected, press tab again and observe that the Info button is now focusable.
4. Press Enter and the drawer will open. Tab to the close button and press Enter; observe the drawer closes.
5. Enable a screen reader such as Voice Over, navigate to the Info button on the card, and observe that the screen reader reads out the ARIA label (e.g. "Info for [App Name]").

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/SelectionCardWrapper.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/SelectionCardWrapper.tsx
@@ -84,6 +84,12 @@ export const SelectionCardWrapper: React.FC<Props> = (props) => {
     openDrawer(label);
   };
 
+  const handleKeyPress = (e: React.KeyboardEvent) => {
+    if (e.key === 'Enter') {
+      openDrawer(label);
+    }
+  };
+
   const renderIcon =
     iconUrl === ''
       ? () => <span className="fl-tux" />
@@ -91,7 +97,13 @@ export const SelectionCardWrapper: React.FC<Props> = (props) => {
 
   const renderVariant = () => (
     <Grid item className={classes.info} xs={2}>
-      <Info onClick={handleInfoClick} />
+      <Info
+        role="button"
+        aria-label={`Info for "${label}"`}
+        onClick={handleInfoClick}
+        onKeyDown={handleKeyPress}
+        tabIndex={0}
+      />
     </Grid>
   );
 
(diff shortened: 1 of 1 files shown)


dwiley-akamai (approved):
LGTM! 🚢

mjac0bs:
> because now it takes two keyboard presses to move to the next app 

@hana-linode That's a good point. I just referred back to the [WebAIM](https://webaim.org/techniques/keyboard/) recommendations, which suggest buttons are a focusable element that should be tab-accessible. (Looks like standards also specify that pressing `space`, in addition to `enter`, should activate the button.) But I was also looking at the ARIA guide and wondering if a `right arrow` key might be the way to go, which is the pattern outlined in [Grids : Interactive Tabular Data and Layout Containers](https://www.w3.org/WAI/ARIA/apg/patterns/grid/).

What do you think: `right arrow` rather than `tab` makes the info button focusable, and then `enter` and `space` interact with the button?

mjac0bs:
Some additional context after looking into manually focusing the Info button via keyboard navigation with `right arrow` instead of `tab`...

The ways to do this seem to be to either:

1. Add an event listener in useEffect
	1. Results in useEffect firing on every keypress for every SelectionCardWrapper, which is not what we want. I think this would require that we manually find the focused app among all SelectionCardWrapper apps in order to evaluate whether the key pressed was the right arrow key and manually trigger focus on Info button.
2. Add an event handler (e.g. onKeyDown, onKeyPress?) to the components' props 
	1. Does not result in the keyboard event listener firing on every app, every time a key is pressed. Does seem to require changes to SelectionCard that might have larger impacts. When attempting to add a onKeyDown event to SelectionCard and SelectionCardWrapper, I was breaking existing behavior: key press did trigger the event handler for the focused app only, but tab functionality broke with an inability to tab to the next app. Attempting to modify onKeyPress leads to other issues.

Manually toggling focus: with a roving tab index is described [here](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_general_within).

bnussman-akamai (approved):

jdamore-linode (approved):

PR #8577, opened 2022-10-19
M3-3166: [CHANGED] Custom User Agent

CodyFinn-Akamai:
## Description 📝

When the api-v4 package is used in a node script the User-Agent header is set to `linodejs`. It also provides users with the ability to prefix the new User-Agent with a custom string if they want to. This does not effect the HTTP requests made in cloud or when the package is used in a browser. 

## How to test 🧪

Ensure that HTTP requests made via manager have their normal browser specific User-Agent.
Ensure that HTTP requests made by using the api-v4 package have `linodejs` set for their User-Agent.

### DIFF
--- a/packages/api-v4/src/request.ts
+++ b/packages/api-v4/src/request.ts
@@ -12,6 +12,18 @@ export const baseRequest = Axios.create({
   baseURL: 'https://api.linode.com/v4',
 });
 
+baseRequest.interceptors.request.use((config) => {
+  const isRunningInNode = typeof process === 'object';
+  const newConfig = {
+    ...config,
+    headers: {
+      ...config.headers,
+      'User-Agent': 'linodejs',
+    },
+  };
+  return isRunningInNode ? newConfig : config;
+});
+
 /**
  * setToken
  *
@@ -275,6 +287,25 @@ export const CancellableRequest = <T>(
         (response) => response.data
       ),
   };
+};
+
+/**
+ * setUserAgentPrefix
+ *
+ * Helper function to set a custom prefix on the user agent
+ *
+ * @param prefix
+ */
+export const setUserAgentPrefix = (prefix: string) => {
+  return baseRequest.interceptors.request.use((config) => {
+    return {
+      ...config,
+      headers: {
+        ...config.headers,
+        'User-Agent': `${prefix}/${config.headers['User-Agent']}`,
+      },
+    };
+  });
 };
 
 export default requestGenerator;


bnussman-akamai (approved):

packages/api-v4/src/request.ts
```diff
@@ -277,4 +289,23 @@ export const CancellableRequest = <T>(
...
 };
 
+/**
+ * setUserAgentPrefix
+ *
+ * Helper function to set a custom prefix on the user agent
+ *
+ * @param prefix
+ */
+export const setUserAgentPrefix = (prefix: string) => {
```
bnussman-akamai, line 299: What is a use case for this function?

dwiley-akamai (commented):
Confirmed from my testing that requests from Cloud have the browser as the user agent and requests using the APIv4 package have `linodejs` as the user agent. 👍 

I did run into this when trying to use `setUserAgentPrefix` though:
Screen Shot 2022-10-25 at 11 01 35 AM

I think this would be a good PR to have a more thorough `How to test` section for as well

packages/api-v4/src/request.ts
CodyFinn-Akamai: It mimics linodego functionality. I believe the purpose is to allow users of the api-v4 package to still add to the user agent header if they need to for whatever reason

CodyFinn-Akamai:
> Confirmed from my testing that requests from Cloud have the browser as the user agent and requests using the APIv4 package have `linodejs` as the user agent. +1
> 
> I did run into this when trying to use `setUserAgentPrefix` though: Screen Shot 2022-10-25 at 11 01 35 AM
> 
> I think this would be a good PR to have a more thorough `How to test` section for as well



> Confirmed from my testing that requests from Cloud have the browser as the user agent and requests using the APIv4 package have `linodejs` as the user agent. +1
> 
> I did run into this when trying to use `setUserAgentPrefix` though: Screen Shot 2022-10-25 at 11 01 35 AM
> 
> I think this would be a good PR to have a more thorough `How to test` section for as well



> Confirmed from my testing that requests from Cloud have the browser as the user agent and requests using the APIv4 package have `linodejs` as the user agent. +1
> 
> I did run into this when trying to use `setUserAgentPrefix` though: Screen Shot 2022-10-25 at 11 01 35 AM
> 
> I think this would be a good PR to have a more thorough `How to test` section for as well

I haven't run into this error. Could you provide how you came across it?

dwiley-akamai (approved):

dwiley-akamai:
> > Confirmed from my testing that requests from Cloud have the browser as the user agent and requests using the APIv4 package have `linodejs` as the user agent. +1
> > I did run into this when trying to use `setUserAgentPrefix` though: Screen Shot 2022-10-25 at 11 01 35 AM
> > I think this would be a good PR to have a more thorough `How to test` section for as well
> 
> I haven't run into this error. Could you provide how you came across it?

Looks like I was calling it in the wrong place, I tried again and it's working without the error so I went ahead and approved

PR #8579, opened 2022-10-24
Update Cypress domain CAA record test to use domain name value

jdamore-linode:
## Description 📝

**What does this PR do?**
Fixes our failing domain CAA record test by supplying a random domain name (rather than a regular old random string) when creating the record. This is necessary as of [API v4.138.0](https://developers.linode.com/changelog/#API-4.138.0), which implements validation for CAA record values.

## How to test 🧪

To reproduce the issue, run the following command against `develop`. To verify the fix, run the same command with this branch checked out:

```
yarn cy:run -s "cypress/e2e/domains/smoke-create-domain-records.spec.ts"
```


bnussman-akamai (approved):
NIce job having `randomDomainName` ready to be used

hana-linode (approved):

PR #8580, opened 2022-10-25
Remove unused `babel` dependencies

bnussman-akamai:
## Description 📝

- Removes Babel because it is no longer needed
  - We no longer depend on Babel to build Cloud Manager
  - We no longer need Babel to transform for unit testing

## How to test 🧪

- Test local development `yarn dev` and ensure it still works as expected
- Test building Cloud Manager for production `yarn build` and serving it with `yarn start:ci`
- Test Storybook development with `yarn storybook`
- Test building Storybook with `yarn build-storybook`

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -16,7 +16,7 @@
   dependencies:
     "@babel/highlight" "^7.16.7"
 
-"@babel/compat-data@^7.13.11", "@babel/compat-data@^7.16.4", "@babel/compat-data@^7.16.8", "@babel/compat-data@^7.9.0":
+"@babel/compat-data@^7.13.11", "@babel/compat-data@^7.16.4", "@babel/compat-data@^7.16.8":
   version "7.17.0"
   resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.17.0.tgz#86850b8597ea6962089770952075dcaabb8dba34"
   integrity sha512-392byTlpGWXMv4FbyWw3sAZ/FrW/DrwqLGXpy0mbyNe9Taqv1mg9yON5/o0cnr8XYCkFTZbC1eV+c+LAROgrng==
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -16,7 +16,7 @@
   dependencies:
     "@babel/highlight" "^7.16.7"
 
-"@babel/compat-data@^7.13.11", "@babel/compat-data@^7.16.4", "@babel/compat-data@^7.16.8", "@babel/compat-data@^7.9.0":
+"@babel/compat-data@^7.13.11", "@babel/compat-data@^7.16.4", "@babel/compat-data@^7.16.8":
```
mjac0bs, line 19: Looking at this generated lock file, I thought it was interesting that we can still have a number of babel dependencies like these that are not themselves dependent on @babel/core. Babel is a little bit of magic to me, though.

mjac0bs (commented):
Most of these commands worked for me, but I received some unexpected output for two, probably due to things not being configured correctly on my end. 

This is the first time I've attempted to build Storybook (though running it via `yarn storybook` worked fine), but it looks like there's potentially some additional set up I need to do there. I was running all commands from the base directory of the repo. Any pointers or suggestions on why I might have been seeing the following?

`yarn build-storybook`:  ```ERR! Error: No configuration files have been found in your configDir.
ERR! Storybook needs either a "main" or "config" file.```

`yarn start:ci`: resulted in `error Command "start:ci" not found`. However, running the full command for the script `yarn serve ./build -p 3000 -s --cors` was successful in generating the "Serving!" messages and addresses, though when I attempted to navigate to the page, I received a 404 Not Found.

yarn.lock
bnussman: I believe babel will remain in our `package.json` because other dependencies such as storybook depend on babel. This is expected!

mjac0bs (approved):

jdamore-linode (approved):

PR #8582, opened 2022-10-25
M3-5969: Prevent unnecessary error for `company` field when updating Account info

dwiley-akamai:
## Description 📝
Updates the Account schema to prevent an unnecessary error on the "Company Name" field when you try updating your billing contact info on an account that did not initially require certain information.

## How to test 🧪
The easiest way to test this would be to create a new Cloud account using your Akamai email address. From that account, you can try editing your billing contact info (not the "Company Name" field) in prod and saving it to observe the unnecessary error on the "Company Name" field. On this branch, you should be able to make those initial changes without receiving an error for that field.

### DIFF
--- a/packages/validation/src/account.schema.ts
+++ b/packages/validation/src/account.schema.ts
@@ -4,9 +4,7 @@ export const updateAccountSchema = object({
   email: string().max(128, 'Email must be 128 characters or less.'),
   address_1: string().max(64, 'Address must be 64 characters or less.'),
   city: string().max(24, 'City must be 24 characters or less.'),
-  company: string()
-    .max(128, 'Company must be 128 characters or less.')
-    .nullable(true),
+  company: string().max(128, 'Company must be 128 characters or less.'),
   country: string()
     .min(2, 'Country code must be two letters.')
     .max(2, 'Country code must be two letters.'),
--- a/packages/validation/src/account.schema.ts
+++ b/packages/validation/src/account.schema.ts
@@ -4,7 +4,9 @@ export const updateAccountSchema = object({
   email: string().max(128, 'Email must be 128 characters or less.'),
   address_1: string().max(64, 'Address must be 64 characters or less.'),
   city: string().max(24, 'City must be 24 characters or less.'),
-  company: string().max(128, 'Company must be 128 characters or less.'),
+  company: string()
+    .max(128, 'Company must be 128 characters or less.')
+    .nullable(true),
   country: string()
     .min(2, 'Country code must be two letters.')
     .max(2, 'Country code must be two letters.'),
--- a/packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/UpdateContactInformationForm/UpdateContactInformationForm.tsx
+++ b/packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/UpdateContactInformationForm/UpdateContactInformationForm.tsx
@@ -231,6 +231,14 @@ class UpdateContactInformationForm extends React.Component<
       filteredRegionResults = regionResults;
     }
 
+    // Prevent edge case field error for "Company Name" when updating billing info
+    // for accounts that did not initially require an address.
+    if (this.state.fields.company === null) {
+      this.setState({
+        fields: { ...this.state.fields, company: '' },
+      });
+    }
+
     return (
       <Grid
         container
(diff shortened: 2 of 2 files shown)


packages/validation/src/account.schema.ts
```diff
@@ -4,7 +4,9 @@ export const updateAccountSchema = object({
   email: string().max(128, 'Email must be 128 characters or less.'),
   address_1: string().max(64, 'Address must be 64 characters or less.'),
   city: string().max(24, 'City must be 24 characters or less.'),
-  company: string().max(128, 'Company must be 128 characters or less.'),
+  company: string()
+    .max(128, 'Company must be 128 characters or less.')
+    .nullable(true),
```
bnussman-akamai, line 9: I think nullable defaults to true if it is used.

```suggestion
    .nullable(),
```

bnussman-akamai (commented):
This change seems fine, but feels inconsistent. If we make `company` nullable, we probably should  make everything else (like `email` for example) nullable because the api allows it. 

Also, api-v4 won't allow `null` a type so I feel like this adds to some inconsistencies
Screen Shot 2022-10-26 at 11 40 55 AM

mjac0bs (approved):

mjac0bs:
@jdamore-linode So I was seeing this at first too... I tried clearing my cache and it seemed to go away. Behavior was then as expected on this branch. Does that work for you?

jdamore-linode:
@mjac0bs No such luck, sadly -- cleared the cache in Chrome and Firefox, and was still able to see the error in both. But it did lead me to discover an interesting (problematic?) behavior.

1. Opened the "Edit Billing Contact Info" drawer
2. Filled out "First Name", "Last Name", "Country", "State", and "Postal Code" fields
3. Clicked "Save Changes"
4. Observed "company must be a `string` type..." error
5. Clicked "Save Changes" again (no page refresh, cache clear, etc. in between)
6. This time the form submitted, allowing me to leave "Company Name" blank
7. Now I'm unable to reproduce the issue

Not clear exactly what's going on and I'll have to create another new account to reproduce if we need it, but my guess is that the schema validation runs before/at the same time as the `this.composeState(...)`, so it still sees the value as `null` instead of `''` and causes the error message to appear, but by the time the user has clicked "Save Changes" again, the state has successfully updated so that `''` gets passed when performing validation the next time. Then the user's company gets updated to `''` upon in the backend, and the error no longer manifests for that user.

dwiley-akamai:
> @jdamore-linode So I was seeing this at first too... I tried clearing my cache and it seemed to go away. Behavior was then as expected on this branch. Does that work for you?

> @mjac0bs No such luck, sadly -- cleared the cache in Chrome and Firefox, and was still able to see the error in both. But it did lead me to discover an interesting (problematic?) behavior.
> 
> 1. Opened the "Edit Billing Contact Info" drawer
> 2. Filled out "First Name", "Last Name", "Country", "State", and "Postal Code" fields
> 3. Clicked "Save Changes"
> 4. Observed "company must be a `string` type..." error
> 5. Clicked "Save Changes" again (no page refresh, cache clear, etc. in between)
> 6. This time the form submitted, allowing me to leave "Company Name" blank
> 7. Now I'm unable to reproduce the issue
> 
> Not clear exactly what's going on and I'll have to create another new account to reproduce if we need it, but my guess is that the schema validation runs before/at the same time as the `this.composeState(...)`, so it still sees the value as `null` instead of `''` and causes the error message to appear, but by the time the user has clicked "Save Changes" again, the state has successfully updated so that `''` gets passed when performing validation the next time. Then the user's company gets updated to `''` in the backend, and the error no longer manifests for that user.

I noticed that behavior too today, yes, what was happening with the previous commit was the `company` field was getting set to an empty string but it only kicked in on the second click of "Save Changes", so the error was still appearing for the first click.

The most recent commit moves the logic into `renderForm()` so that when the form renders, if `company` is `null`, it'll change it to an empty string. This way, it's not dependent upon the clicking of "Save Changes" and it is changed far earlier in the process.

jdamore-linode:
@dwiley-akamai I wasn't able to confirm the latest changes using real data (tried creating a new account `jdamore+1@akamai.com`, but haven't had success receiving the confirmation email...), but I think I was able to confirm it with MSW enabled and the account factory's `company` field set to `null` (ignoring the resulting type errors).

With this branch checked out, I did not see the schema validation error upon updating the contact info. Checking out `develop` with MSW still enabled and configured as above, I do see the error.

Not sure if we have anybody available to check this with real data, but I'll give this an approval and leave it to you whether or not we want to merge this for the release today.

mjac0bs:
@jdamore-linode @dwiley-akamai 
I can see how clearing the cache was a red herring the first time around when I'd already tried to Save Changes once. 

I was able to test this when creating another new account with my linode email since it was still unused. Went through the above steps again to update contact info with "Country", "State" and "Postal Code" fields only. Clicked Save Changes and was able to submit changes successfully the first time without seeing any error message on "Company".

dwiley-akamai:
Additional testing was coordinated via Slack and we've confirmed this looks good with real data and the intended email addresses. Going ahead and merging

PR #8583, opened 2022-10-26
Copy IP from Kubernetes Node Pool

kitknox:
## Description 📝

Makes each row in a node pool more consistent with the Linode instance rows including changing the color of the active row and allowing copy of the IP address for a given node.

## Preview 📷

Screen Shot 2022-10-25 at 5 57 31 PM

## How to test 🧪

Confirm you can now copy the IP address of a selected node from a Kubernetes node pool


jdamore-linode:
This is a nice enhancement, thanks Kit! I did notice a small issue and one minor design inconsistency:

When you create a new cluster, the Linodes aren't assigned an IP right away but the copy button still appears in the UI. Clicking on it shows the "Copied!" message, but nothing is actually copied to the clipboard:
 
Screen Shot 2022-10-27 at 9 53 59 AM

I think ideally the button would not be present when a Linode is still provisioning, and it looks like you could do that with a quick conditional, but there may be a more elegant solution:
```typescript
// NodeTable.tsx ln 285:
) : (displayIP.length > 0 && (
 <div className={classes.ipsWrapper}>
   <IPAddress ips={[displayIP]} showMore />
 </div>
))}
```

And the only other thing that caught my eye is that this IP copy button is a little bit different from others because it has a white background when it's focused or hovered (I believe this button should typically have a transparent background):

Node pool table:
Screen Shot 2022-10-27 at 10 08 01 AM

Linode landing page table:
Screen Shot 2022-10-27 at 10 08 07 AM

Aside from those small things, I think this is a really nice improvement! Thanks again!

dwiley-akamai (commented):
Thanks for opening this PR!

In addition to what Joe pointed out, if you compare this branch with production, there's a slight upward displacement of a few pixels of the IP address so it's no longer vertically centered

kitknox:
> Thanks for opening this PR!
> 
> In addition to what Joe pointed out, if you compare this branch with production, there's a slight upward displacement of a few pixels of the IP address so it's no longer vertically centered

I also noticed that we don't have consistent behavior for the copy IP in various places around Cloud Manager.  The hint color in the Linode instance list is different, etc.  Looks like this new view also ends up with a different highlight color when in dark mode.  So a fair bit to cleanup still.  I was doing work where I kept needing access to the worker node IP and was missing this so was going to make the small change.

Oh, and while we are in here, the running state icons/colors for a worker node don't match that of a regular Linode instance either.

hana-linode (approved):

dwiley-akamai (commented):
@hana-linode It's looking good to me, but in the Linode Details --> Network tab --> IP Addresses table and the Node Pools table, the copy icons on hover don't change color as they do on the Linodes landing table

dwiley-akamai (approved):

PR #8568, opened 2022-10-11
M3-5196: React Query for Volumes - Volumes Landing

bnussman-akamai:
## Description 📝

- The goal with **part 1** is to get the Volumes Landing and all associated actions such as attach, delete, etc working

## How to test 🧪

- Test **only** Volumes Landing features at `http://localhost:3000/volumes`
  - Check all actions/mutations possible in the Volumes Landing Action Menus
  - Check that Landing table updates accordingly following actions such as attach, delete, clone, etc...

PR #8584, opened 2022-10-26
React Query for Volumes - Linode Storage

bnussman-akamai:
## Description 📝

- Makes `http://localhost:3000/linodes/:id/storage` use React Query

## Preview 📷

https://user-images.githubusercontent.com/115251059/198066432-616af01d-d604-47a2-b6d3-2be813113bfd.mov

## How to test 🧪

- Test all possible volume actions from `http://localhost:3000/linodes/:id/storage` 
  - When testing, check back with the `http://localhost:3000/volumes` page to make sure the cache is updating properly as mutations occur
- Test volume actions `http://localhost:3000/volumes` and make sure `http://localhost:3000/linodes/:id/storage` remains up to date

### DIFF
--- a/packages/api-v4/src/volumes/types.ts
+++ b/packages/api-v4/src/volumes/types.ts
@@ -19,6 +19,7 @@ export type VolumeStatus =
   | 'resizing'
   | 'deleting'
   | 'deleted'
+  | 'migrating'
   | 'contact_support';
 
 type VolumeHardwareType = 'hdd' | 'nvme';
--- a/packages/manager/src/features/Volumes/VolumeAttachmentDrawer.tsx
+++ b/packages/manager/src/features/Volumes/VolumeAttachmentDrawer.tsx
@@ -148,12 +148,7 @@ export const VolumeAttachmentDrawer = (props: Props) => {
           name="config"
           id="config"
           errorText={linodeError}
-          disabled={
-            disabled ||
-            readOnly ||
-            selectedLinode === -1 ||
-            configChoices.length === 1
-          }
+          disabled={disabled || readOnly || selectedLinode === -1}
           label="Config"
           isClearable={false}
           isLoading={configsLoading}
--- a/packages/manager/src/features/Volumes/VolumeAttachmentDrawer.tsx
+++ b/packages/manager/src/features/Volumes/VolumeAttachmentDrawer.tsx
@@ -5,7 +5,7 @@ import Button from 'src/components/Button';
 import FormControl from 'src/components/core/FormControl';
 import FormHelperText from 'src/components/core/FormHelperText';
 import Drawer from 'src/components/Drawer';
-import Select, { Item } from 'src/components/EnhancedSelect/Select';
+import Select, { Item } from 'src/components/EnhancedSelect';
 import Notice from 'src/components/Notice';
 import { resetEventsPolling } from 'src/eventsPolling';
 import LinodeSelect from 'src/features/linodes/LinodeSelect';
@@ -37,7 +37,7 @@ export const VolumeAttachmentDrawer = (props: Props) => {
   } = useAttachVolumeMutation();
 
   const [selectedLinode, setSelectedLinode] = React.useState<number>(-1);
-  const [selectedConfig, setSelectedConfig] = React.useState<string>();
+  const [selectedConfig, setSelectedConfig] = React.useState<Item<string>>();
 
   const { data, isLoading: configsLoading } = useAllLinodeConfigsQuery(
     selectedLinode,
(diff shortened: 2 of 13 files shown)


mjac0bs (commented):
I noticed a couple of things. After trying various actions, I think the cache is just updating slowly (multiple minutes) so changes made on one page don't always immediately (but eventually?) reflect on the other without a manual refresh.

When a new volume is created from `http://localhost:3000/linodes/:id/storage`, I watched it quickly become active on `http://localhost:3000/volumes`, but the volume's status remained in the "Creating" stage on `http://localhost:3000/linodes/:id/storage` until I manually refreshed the page. It looks like a few other actions require a manual refresh for the update to be *quickly* visible across pages, as well -- is this expected behavior? Are you seeing something different?

**To quickly see updates on `http://localhost:3000/volumes` after action taken on `http://localhost:3000/linodes/:id/storage`**:
Required refresh: rename, retag, resize
No required refresh: attach, detach, delete

**To quickly see updates on `http://localhost:3000/linodes/:id/storage` after action taken on `http://localhost:3000/volumes`:**
Required refresh: rename, retag, resize, attach (but the toast displays)
No required refresh: detach, delete

Screen Shot 2022-10-28 at 8 51 19 AM

On the Volumes landing page, it looks like we misspell "Attached" in the table header. See prod. vs [the change in a previously merged PR.](https://github.com/linode/manager/pull/8568/files#diff-d028c43b4b85460702d69c80e017739cd1fd1c4a36095760648c848374547fc5R315)

Prod:
Screen Shot 2022-10-28 at 8 20 23 AM

[M3-5196: React Query for Volumes - Volumes Landing](https://github.com/linode/manager/pull/8568/files#top)
Screen Shot 2022-10-28 at 8 16 55 AM

bnussman-akamai:
@mjac0bs Were you testing with multiple instances of Cloud Manager open? I am writing logic that is assuming you are using one singular instances of Cloud Manager.

mjac0bs:
> Were you testing with multiple instances of Cloud Manager open? 

@bnussman-akamai: I was; I had two local browser tabs open. It makes sense that you were assuming the user would be using one instance of Cloud Manager, though. 

I retested the volume actions with just one Cloud Manager open and most of the actions seem to be working as expected! I did notice that volume detachment doesn't complete as quickly in the UI as the other actions. After seeing both the 'started' and 'successful' toasts, the "detached" volume continues to show up for a bit, both in the storage tab of the linode and in the "Attached To" column on the volumes landing page. Even though the update to the page isn't immediate, the volumes on both pages generally seem to be in sync. Refreshing the page gets the UI up to date.

There was one time where the detachment seemed out of sync across the pages (it was not showing up as attached on Linode Details but was listed as attached on Volumes Landing), but it's an edge case that I don't know many customers would realistically face. 

I'll send you some videos of what I've seen. 

While testing, I also noticed something else. Is there a reason why we can attach an existing volume to a linode on Volumes Landing without specifying a config, but can't do the same on Linode Details without receiving a "config_id is not valid" error until a config is selected? Screencaps below. (This might be beyond the scope of this ticket, but wanted to point it out while we're testing volume actions.)

Screen Shot 2022-11-04 at 9 54 55 AM

https://user-images.githubusercontent.com/114685994/200038152-99cf8b9c-6267-4865-91d7-067c39a9bfc7.mov

mjac0bs (commented):
The detach functionality seems to be working nicely now!

It does look like I'm still seeing the config_id issue, though. After investigating a little more, I can see that the error response that shows up in the console traces back to the attachVolumes call in `packages/manager/src/features/Volumes/VolumeDrawer/AttachVolumeToLinodeForm.tsx`, but more precisely, probably [this line](https://github.com/linode/manager/pull/8584/files#diff-30ae90cb256ec5cde1ec29a223c6403e44d6343a6e05ea040337392847c91e2fL69). It looks like here we're using a [different validation schema](https://github.com/linode/manager/pull/8584/files#diff-30ae90cb256ec5cde1ec29a223c6403e44d6343a6e05ea040337392847c91e2fL42-L45) than we are with attachment on the Volumes Landing page, which may be responsible for our issues on the linode details page.

bnussman:
Thank you for all of the great feedback @mjac0bs. Just updated the validation to be more consistent in 
22bea69

CodyFinn-Akamai (approved):
RQ code works just like redux. Nice job!

mjac0bs (approved):
I just went back through and tested all the actions again. Looks good so I'm going to approve this. 

Since our handling of config validation with volume attachment differs from the API, I think it would be good to address in a separate ticket.

PR #8587, opened 2022-10-27
Update `launchdarkly-react-client-sdk` to `latest`

bnussman-akamai:
## Description 📝

- Remove Launchdarkly Patch because it is no longer needed
- Disable postinstall command because we don't need to do any patching
- Updates  `launchdarkly-react-client-sdk` to `latest`

## How to test 🧪

- Have `REACT_APP_LAUNCH_DARKLY_ID` defined in your `.env` (ask me for this value)
- Re-run `yarn dev` to pick up new dependencies and env value
- Verify Launch Darkly is working by 
  - Logging into Launch Darkly
  - Having Cloud Manger open
  - Toggling a flag like `Database Beta` **in the Test environment**
  - Making sure that Cloud Manager updated to reflect the new value


mjac0bs (approved):
Tested with `database-beta` and `Promotional Offers` feature flags in Launch Darkly Test environment and saw the Database beta chip and object storage banner display when feature was toggled on. 

All looks to be working!

jdamore-linode (approved):
Nice! Any thoughts about deleting the `postinstall` script altogether, along with the `postinstall` and `patch-packages` dev dependencies? Or do you think there's a decent chance we'll need them again in the future?

bnussman:
@jdamore-linode I'm all for getting rid of of all of that if the team is ok with it!

PR #8590, opened 2022-10-27
M3-5965: Add Generic Support Text Link Replacement to Notices

mjac0bs:
## Description 📝

**What does this PR do?**
Links the words "open a support ticket" in error notices to the support page with an open ticket modal. We previously did this for the Account Limit error only; this PR intends to link "open a support ticket" for all errors intercepted from the API with these key words that are not field errors, and replace "open a support ticket" anywhere it exists the error message reason.

## Preview 📷
Screen Shot 2022-10-27 at 1 47 13 PM
Screen Shot 2022-10-27 at 2 01 55 PM
Screen Shot 2022-10-27 at 2 02 31 PM

## How to test 🧪

Tests from the [previous PR](https://github.com/linode/manager/pull/8566) still apply, but this can also be tested in `serviceHandlers.ts` by commenting out the payload, linode, and successful mocked response in `rest.post('*/instances', async (req, res, ctx)` and adding the error response below.  Test the following error messages (one at a time) when creating a linode with MSW:

```
    return res(
      ctx.status(400),
      ctx.json({
        errors: [
          // {
          //   reason: 'Open a support ticket for this issue.',
          // },
          // {
          //   reason:
          //     'Additional verification is required to add this service. Please open a support ticket.',
          // },
          // { reason: 'Please open a Support Ticket describing your requirement'}
        ],
      })
    );
```


packages/manager/src/components/SupportError.tsx
```diff
@@ -9,14 +9,24 @@ interface Props {
 
 export const SupportError: React.FC<Props> = (props) => {
   const { errors } = props;
-  const [reason] = errors[0].reason.split(/open a support ticket\./i);
+  const errorMsg = errors[0].reason.split(/(open a support ticket)/i);
+  const ticketLinkText =
```
mjac0bs, line 13: I don't like this line of code. There don't seem to be any current APIv4 error messages where "open a support ticket" is at the beginning of the sentence, but this covers the case that it shows up at the start of the first sentence only. Should fix this to capitalize "Open" if it were to be the first word in any sentence in the message.

bnussman (approved):

PR #8592, opened 2022-10-31
Fix IP Transfer Dialog

bnussman-akamai:
## Bug 🪲

- IPs in Transfer IP Dialog were not correct after a change has been made

https://user-images.githubusercontent.com/115251059/199073105-8dedabb5-cccb-461b-afaf-2c69fe1701d0.mov

## Solution ✅

- Invalidate the React Query cache after an IP has been changed

https://user-images.githubusercontent.com/115251059/199073407-4782eb7f-844c-4875-ad94-fe2ee6d61312.mov

## How to test 🧪

- Test the Transfer IP dialog in depth
- Test swapping IPs between two Linodes twice in a row


jdamore-linode (approved):
Nice! Thanks for tackling this so quickly! I was able to reproduce the issue, and confirmed that these changes fix it.

mjac0bs (approved):
I also was able to reproduce the error and the fix in this PR. Now seeing the IP Transfer work!

dwiley-akamai (approved):

PR #8593, opened 2022-10-31
Re-add `nth-check` Yarn Resolution for CVE-2021-3803

bnussman-akamai:
## Description 📝

- Re-adds a resolution for `nth-check` for CVE-2021-3803
- I removed this in #8587 unintentionally, my fault!

## How to test 🧪

- Rerun `yarn dev` and check Cloud Manager in general

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,8 @@
     "browserslist": "^4.16.5",
     "trim-newlines": "^4.0.2",
     "hosted-git-info": "^5.0.0",
-    "minimatch": "^3.0.5"
+    "minimatch": "^3.0.5",
+    "nth-check": "^2.0.1"
   },
   "workspaces": {
     "packages": [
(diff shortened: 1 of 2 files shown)


mjac0bs (approved):
Didn't see any errors running `yarn dev` and found no unexpected issues when clicking  around in Cloud Manager. 👍 

(Is there a way we could have caught the removal of `nth-check` in the previous PR, other than noticing and pointing out the [line](https://github.com/linode/manager/pull/8587/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L62) of code?)

jdamore-linode:
@mjac0bs Not sure what brought it to Banks's attention, but it looks like Dependabot did notice the change and [created an alert](https://github.com/linode/manager/security/dependabot/3) (though I'm not sure if we get any kind of notification beyond that)

PR #8605, opened 2022-11-08
`develop` ⬅️ `feature/mui-v5`

bnussman-akamai:
## Description 📝

- Updates Material UI from v4.x.x to v5.x.x

> **Note**: Many of the changes in this PR automatically updated by MUI's `codemod`

## How to test 🧪

- Test all aspects of Cloud Manager
  - The goal of testing in this case is to make sure that the UI in this PR is indistinguishable from `develop` following this massive upgade


jdamore-linode (commented):
Nice work, Banks!! I still have to do manual testing and comb through the code, but I'm really impressed so far.

I ran Cypress and there were a few failures, but only one was actually an issue -- I'll open a PR with fixes soon.

jdamore-linode:
Only issue I've found so far is the "Open a Support Ticket" dialog's description field is super tall now, which led to an E2E test failure because the submit button is no longer on-screen:

Screen Shot 2022-11-10 at 12 16 43 PM

jdamore-linode:
Only other question so far -- are the toasts supposed to have changed?

Screen Shot 2022-11-10 at 1 05 34 PM

bnussman:
@jdamore-linode Good catch. I'll make a PR to fix both of those issues.

dwiley-akamai (commented):
I combed through as many places of the app I could think of and noticed a few general differences compared to prod and some specific issues we'll need to address.

**General differences**:
- The spacing between field names and fields is greater compared to prod. Looking at modals and drawers especially here. I don't think this is an issue per se, but something to note.
- The spacing between headers and paragraphs is decreased in some places compared to prod. An example is Managed --> SSH Access --> the Linode Public Key section.

**Specific Issues**:
- Field text color in the Volume Configuration drawer is grayed out
- Increased left padding of entity labels in breadcrumbs
- "X" button in tags is misaligned
Screen Shot 2022-11-15 at 11 40 33 AM
- Dividers in Longview Clients line w/ links and API Key
Screen Shot 2022-11-15 at 12 37 59 PM
- Support page in dark mode — tile titles disappear on hover
Screen Shot 2022-11-15 at 12 51 13 PM

CodyFinn-Akamai (approved):
A massive undertaking! Nice job!

PR #8610, opened 2022-11-11
M3-5197: React Query for Volumes - Misc Final Changes

bnussman-akamai:
## Description 📝

- Wraps up React Query for Volumes
- Remove Volumes from Redux
- Makes final changes needed throughout app for Volumes to come from React Query 

## How to test 🧪

- Test all Volume related features in Cloud Manager


mjac0bs (approved):
I went through and tested all the volume actions (rename, tag, resize, clone, attach, detach, create, delete) on both the volumes landing and linode details page. All the still functionality seemed to be working as expected in the UI! 

Also ran the e2e tests for volumes. Other than the known issue with volume detachment tests failing (haven't followed the status on that), I had a couple of cypress timeouts and tests otherwise passed.

packages/manager/src/store/volume/volume.selector.ts
```diff
@@ -1,9 +1,4 @@
...
 
-export const getVolumesForLinode = (
-  itemsById: Record<string, Volume>,
-  linodeId: number
-) =>
-  Object.values(itemsById).filter(
-    ({ linode_id }) => linode_id && linode_id === linodeId
-  );
+export const getVolumesForLinode = (volumes: Volume[], linodeId: number) =>
+  volumes.filter(({ linode_id }) => linode_id && linode_id === linodeId);
```
dwiley-akamai, line 4: Since volumes are moved to RQ now, I think it'd be good to move this function to somewhere in `src/features/Volumes` (maybe a new volumes utility file in there if there's no place that currently makes sense for it) and delete this file since it's in the store with Redux stuff.

packages/manager/src/utilities/getEntityByIDFromStore.test.ts
```diff
@@ -41,7 +32,7 @@ describe('getEntityByIDFromStore utility function', () => {
   });
 
   it('should work as a curried function', () => {
-    const getVolume = getEntityByIDFromStore('volume');
-    expect(getVolume(volume.id)).toEqual(volume);
+    const getVolume = getEntityByIDFromStore('linode');
```
dwiley-akamai, line 35: this should be `getLinode` now, right?

packages/manager/src/features/linodes/MigrateLanding/MigrateLinode.tsx
```diff
@@ -226,7 +226,7 @@ const MigrateLanding: React.FC<CombinedProps> = (props) => {
         notifications={notifications}
       /> */}
       <CautionNotice
-        linodeVolumes={linode._volumes}
+        linodeId={linodeID}
```
dwiley-akamai, line 229: When you open the Migrate modal, there's a delay of up to a few seconds before the volumes get listed out in the notice. From what I can see this is still an improvement in some cases over current prod behavior, where if the volumes for the given Linode haven't been requested yet, no volumes get listed out.

If we pulled the new volumes logic from `CautionNotice` and got that data in this component instead, and then passed the array of volumes down as a prop, I think it would fix the delay and still avoid the current prod issue. What do you think about that change?

packages/manager/src/features/Volumes/VolumesLanding.tsx
```diff
@@ -293,14 +293,7 @@ export const VolumesLanding: React.FC<CombinedProps> = (props) => {
...
             </TableSortCell>
-            <TableSortCell
-              active={orderBy === 'region'}
-              direction={order}
-              label="region"
-              handleClick={handleOrderChange}
-            >
-              Region
-            </TableSortCell>
+            <TableCell>Region</TableCell>
```
dwiley-akamai, line 296: It's not possible for us to retain the ability to sort by Region?

dwiley-akamai (commented):
There is more code I need to comb through, but it's looking good so far. Left a few comments/questions about a few points

packages/manager/src/features/linodes/MigrateLanding/MigrateLinode.tsx
bnussman-akamai: I think that approach would generally be fine, but I don't want the data to be fetched unless the `<MigrateLinode />` is mounted and I don't want to have to deal with conditionally enabling the query when the modal is open. I also like keeping the queries as close to the use of the data as possible. 

This is a tradeoff between okay UX and great DX as compared to great UX and ok DX if that makes any sense.

packages/manager/src/features/Volumes/VolumesLanding.tsx
bnussman-akamai: Because this data is now paginated and ordered by the API, it's not possible because the API errors when you try to order by `region`. We can ask the API to support this.

packages/manager/src/features/linodes/LinodesDetail/LinodeRescue/RescueDialog.tsx
```diff
@@ -159,7 +151,11 @@ const LinodeRescue: React.FC<CombinedProps> = (props) => {
 
   const devices = {
     disks: linodeDisks ?? [],
-    volumes: filteredVolumes ?? [],
+    volumes:
+      filteredVolumes.map((volume) => ({
+        ...volume,
+        _id: String(volume.id),
+      })) ?? [],
```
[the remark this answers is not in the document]
bnussman-akamai: Good catch. I have no idea what I was thinking when I made this change. Fixed 29095e8

dwiley-akamai (approved):
I think this should be good to go now 👍

PR #8612, opened 2022-11-15
Fix MUI v5 UI Issues

bnussman-akamai:
## Description 📝

- Fixes UI issues pointed out by @DevDW 

## How to test 🧪

- Verify that UI issues from https://github.com/linode/manager/pull/8605#pullrequestreview-1181362210 are resolved

### DIFF
--- a/packages/manager/src/features/Longview/shared/InstallationInstructions.tsx
+++ b/packages/manager/src/features/Longview/shared/InstallationInstructions.tsx
@@ -29,8 +29,8 @@ const useStyles = makeStyles((theme: Theme) => ({
         '&:before': {
           content: "'|'",
           position: 'absolute',
-          top: `calc(${theme.spacing(1)} - 3)`,
-          left: -theme.spacing(1) + 2,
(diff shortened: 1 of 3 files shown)


dwiley-akamai (commented):
I think this is still a WIP since there are some pieces from the last bit of feedback that still need addressing, but it's looking good so far. 

This is minor but I did observe that the "X" on the tags is now a bit bigger than in prod and the hover state doesn't cover the entire box as a result:

Screen Shot 2022-11-15 at 5 27 18 PM

dwiley-akamai (commented):
The only issue I'm seeing now is when you're loading up the Longview landing page, the box for the individual clients is very large until it gets populated (the video doesn't show it all, but you can end up scrolling very far down the page):

https://user-images.githubusercontent.com/114682940/202235867-49d07dd5-cc03-4e66-826b-f8911130bd4a.mov

dwiley-akamai (approved):
LGTM now, nice work! 🚀

PR #8614, opened 2022-11-16
M3-5461: Add Object Storage type to confirm

mjac0bs:
## Description 📝

**What does this PR do?**

Implements type-to-confirm for object storage cancellation in the user's Account Settings.

## Preview 📷

**TTC enabled:**
image

**TTC disabled:**
image

https://user-images.githubusercontent.com/114685994/202314676-dfbcdef3-0265-4034-9ad1-f669646fd71a.mov

## How to test 🧪

- If Object Storage is not already active, create a new Object Storage bucket to enable it. 
- Navigate to My Settings and ensure type-to-confirm is enabled.
- Navigate to Account Settings and click the Cancel Object Storage button.
- Observe the type-to-confirm pattern present on the Cancel Object Storage dialog.
- Navigate to My Settings and toggle the type-to-confirm setting to disabled.
- Navigate to Account Settings and click the Cancel Object Storage button
- Observe the type-to-confirm text field is not visible and a description instructs user to visit My Settings to toggle this setting.
- Confirm that clicking 'Confirm cancellation' does disable object storage and delete data.

- Run `EnableObjectStorage.test.tsx`.
- Run `TypeToConfirm.test.tsx`.

### DIFF
--- a/packages/manager/src/components/TypeToConfirm/TypeToConfirm.tsx
+++ b/packages/manager/src/components/TypeToConfirm/TypeToConfirm.tsx
@@ -66,7 +66,10 @@ const TypeToConfirm: React.FC<Props> = (props) => {
         </>
       ) : null}
       {!hideInstructions ? (
-        <Typography className={classes.description}>
+        <Typography
+          data-testid={'instructions-to-enable-or-disable'}
+          className={classes.description}
+        >
           To {disableOrEnable} type-to-confirm, go to the Type-to-Confirm
           section of <Link to="/profile/settings">My Settings</Link>.
         </Typography>
--- a/packages/manager/src/components/TypeToConfirm/TypeToConfirm.tsx
+++ b/packages/manager/src/components/TypeToConfirm/TypeToConfirm.tsx
@@ -67,7 +67,7 @@ const TypeToConfirm: React.FC<Props> = (props) => {
       ) : null}
       {!hideInstructions ? (
         <Typography
-          data-testid={'instructions-to-enable-or-disable'}
+          data-testid="instructions-to-enable-or-disable"
           className={classes.description}
         >
           To {disableOrEnable} type-to-confirm, go to the Type-to-Confirm
--- a/packages/manager/src/features/Account/EnableObjectStorage.tsx
+++ b/packages/manager/src/features/Account/EnableObjectStorage.tsx
@@ -3,7 +3,6 @@ import { cancelObjectStorage } from '@linode/api-v4/lib/object-storage';
 import { APIError } from '@linode/api-v4/lib/types';
 import * as React from 'react';
 import { Link } from 'react-router-dom';
-import { compose } from 'recompose';
 import Accordion from 'src/components/Accordion';
 import ActionsPanel from 'src/components/ActionsPanel';
 import Button from 'src/components/Button';
@@ -13,17 +12,15 @@ import TypeToConfirm from 'src/components/TypeToConfirm';
 import Typography from 'src/components/core/Typography';
 import ExternalLink from 'src/components/ExternalLink';
 import Grid from 'src/components/Grid';
-import withPreferences, {
-  Props as PreferencesProps,
-} from 'src/containers/preferences.container';
 import { updateAccountSettingsData } from 'src/queries/accountSettings';
+import usePreferences from 'src/hooks/usePreferences';
 import { useProfile } from 'src/queries/profile';
 
 interface Props {
   object_storage: AccountSettings['object_storage'];
 }
 
-type CombinedProps = Props & PreferencesProps;
+type CombinedProps = Props;
 
 interface ContentProps {
   object_storage: AccountSettings['object_storage'];
@@ -73,11 +70,12 @@ export const ObjectStorageContent: React.FC<ContentProps> = (props) => {
 };
 
 export const EnableObjectStorage: React.FC<CombinedProps> = (props) => {
-  const { object_storage, preferences } = props;
+  const { object_storage } = props;
   const [isOpen, setOpen] = React.useState<boolean>(false);
   const [error, setError] = React.useState<string | undefined>();
   const [isLoading, setLoading] = React.useState<boolean>(false);
   const [confirmText, setConfirmText] = React.useState('');
+  const { preferences } = usePreferences();
   const { data: profile } = useProfile();
   const username = profile?.username;
   const disabled =
@@ -104,29 +102,24 @@ export const EnableObjectStorage: React.FC<CombinedProps> = (props) => {
       .catch(handleError);
   };
 
-  const renderActions = (
-    disabled: boolean,
-    loading: boolean,
-    onClose: () => void,
-    onSubmit: () => void
-  ) => (
+  const actions = (
     <ActionsPanel>
       <Button
         buttonType="secondary"
-        onClick={onClose}
-        data-testid={'dialog-cancel'}
+        onClick={handleClose}
+        data-testid="dialog-cancel"
       >
         Cancel
       </Button>
 
       <Button
         buttonType="primary"
-        onClick={onSubmit}
+        onClick={handleSubmit}
         disabled={disabled}
-        loading={loading}
-        data-testid={'dialog-confirm'}
+        loading={isLoading}
+        data-testid="dialog-confirm"
       >
-        Confirm cancellation
+        Confirm Cancellation
       </Button>
     </ActionsPanel>
   );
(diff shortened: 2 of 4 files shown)


CodyFinn-Akamai:
Type to confirm works. All the objects and buckets are deleted. Object Storage is disabled on the account.
It is a little odd being stuck on the modal for so long though. Should we close the modal once the user successfully clicks the confirm button. Then we could show a toast informing the user of success or failure? Also if the user clicks the confirm button should we disable the cancel button since they cannot cancel the request?

packages/manager/src/features/Account/EnableObjectStorage.tsx
```diff
@@ -112,15 +143,35 @@ export const EnableObjectStorage: React.FC<CombinedProps> = (props) => {
...
+          }
+          visible={preferences?.type_to_confirm}
+        />
       </ConfirmationDialog>
     </>
   );
 };
 
-export default React.memo(EnableObjectStorage);
+const enhanced = compose<CombinedProps, Props>(withPreferences());
```
bnussman, line 175: Rather than using `withPreferences` we should use the `usePreferences` **hook** for the sake of using new React patterns. This will be nice when we switch preferences to React Query.

packages/manager/src/features/Account/EnableObjectStorage.tsx
```diff
@@ -88,13 +104,28 @@ export const EnableObjectStorage: React.FC<CombinedProps> = (props) => {
       .catch(handleError);
   };
 
-  const actions = () => (
+  const renderActions = (
+    disabled: boolean,
+    loading: boolean,
+    onClose: () => void,
+    onSubmit: () => void
+  ) => (
```
bnussman, line 112: No need to pass parameters here or even make this a function because the values are in the scope of this function. You approach is fine, but I find this reduces lines of code and makes the component more readable.

```typescript
  const actions = (
    <ActionsPanel>
      <Button
        buttonType="secondary"
        onClick={handleClose}
        data-testid={'dialog-cancel'}
      >
        Cancel
      </Button>

      <Button
        buttonType="primary"
        onClick={handleSubmit}
        disabled={disabled}
        loading={isLoading}
        data-testid={'dialog-confirm'}
      >
        Confirm cancellation
      </Button>
    </ActionsPanel>
  );
```

then use actions just by doing 

```typescript
      <ConfirmationDialog
        open={isOpen}
        error={error}
        onClose={() => handleClose()}
        title="Cancel Object Storage"
        actions={actions}
      >
```

packages/manager/src/features/Account/EnableObjectStorage.tsx
```diff
@@ -112,15 +143,35 @@ export const EnableObjectStorage: React.FC<CombinedProps> = (props) => {
...
-        </Typography>
+        <Notice warning>
+          <Typography style={{ fontSize: '0.875rem' }}>
+            <strong>Warning:</strong> Canceling Object Storage will permanently
+            delete all buckets and their objects. Object Storage Access Keys
+            will be revoked.
+          </Typography>
+        </Notice>
+        <TypeToConfirm
+          data-testid={'dialog-confirm-text-input'}
```
bnussman, line 157:
```suggestion
          data-testid="dialog-confirm-text-input"
```

packages/manager/src/features/Account/EnableObjectStorage.tsx
```diff
@@ -88,13 +104,28 @@ export const EnableObjectStorage: React.FC<CombinedProps> = (props) => {
...
         Cancel
       </Button>
 
-      <Button buttonType="primary" onClick={handleSubmit} loading={isLoading}>
+      <Button
+        buttonType="primary"
+        onClick={onSubmit}
+        disabled={disabled}
+        loading={loading}
+        data-testid={'dialog-confirm'}
```
bnussman, line 127:
```suggestion
        data-testid="dialog-confirm"
```

packages/manager/src/features/Account/EnableObjectStorage.tsx
```diff
@@ -88,13 +104,28 @@ export const EnableObjectStorage: React.FC<CombinedProps> = (props) => {
...
+    loading: boolean,
+    onClose: () => void,
+    onSubmit: () => void
+  ) => (
     <ActionsPanel>
-      <Button buttonType="secondary" onClick={handleClose}>
+      <Button
+        buttonType="secondary"
+        onClick={onClose}
+        data-testid={'dialog-cancel'}
```
bnussman, line 117:
```suggestion
        data-testid="dialog-cancel"
```

packages/manager/src/components/TypeToConfirm/TypeToConfirm.tsx
```diff
@@ -66,7 +66,10 @@ const TypeToConfirm: React.FC<Props> = (props) => {
         </>
       ) : null}
       {!hideInstructions ? (
-        <Typography className={classes.description}>
+        <Typography
+          data-testid={'instructions-to-enable-or-disable'}
```
bnussman, line 70:
```suggestion
          data-testid="instructions-to-enable-or-disable"
```

packages/manager/src/features/Account/EnableObjectStorage.tsx
```diff
@@ -88,13 +104,28 @@ export const EnableObjectStorage: React.FC<CombinedProps> = (props) => {
...
 
-      <Button buttonType="primary" onClick={handleSubmit} loading={isLoading}>
+      <Button
+        buttonType="primary"
+        onClick={onSubmit}
+        disabled={disabled}
+        loading={loading}
+        data-testid={'dialog-confirm'}
+      >
         Confirm cancellation
```
bnussman, line 129: Should we capitalize this?

```suggestion
        Confirm Cancellation
```

bnussman (commented):
Nice job all of this! Everything works great as is, but I have some optional changes that could be made

packages/manager/src/features/Account/EnableObjectStorage.tsx
mjac0bs: Thanks for this feedback. It felt kind of weird when I was writing it, so I'm glad there was a cleaner way to write it.

packages/manager/src/features/Account/EnableObjectStorage.tsx
mjac0bs: I was wondering about this too, so thanks for pointing it out. I went back through the app and found that generally all words are capitalized on our buttons, so I went ahead and made your suggested change here.

mjac0bs:
> Should we close the modal once the user successfully clicks the confirm button. Then we could show a toast informing the user of success or failure? Also if the user clicks the confirm button should we disable the cancel button since they cannot cancel the request?

@CodyFinn-Akamai - I think it would be good, at least for the scope of this ticket, to keep in line with behavior of the other type-to-confirm delete/disable dialogs. I haven't come across one yet that disables the cancel button on confirm (database delete cluster and kube delete cluster still enable it). Let me know if I'm missing others that do.

image

The request does take a couple seconds to complete. I don't think it's an unusual wait, about the same as other delete/disable actions I was trying. Kube delete has no toast and database delete has a toast but the user still has to wait on the action to complete for the modal to close. If we want to improve UI's responsiveness to the delete requests, I'm wondering if that should be its own ticket.

dwiley-akamai (approved):
LGTM, nice work 👌 

There's one potential improvement I see, but I'm going ahead and approving given the scope of the ticket.

Once you cancel OBJ, if you navigate back to the OBJ landing page, you'll still see your Buckets and Access Keys. This seems to be because RQ isn't being updated (unless you manually refresh the page).

In the success callback of `handleSubmit()`, we could address that issue by invalidating the relevant queries with `queryClient.invalidateQueries(...)`

CodyFinn-Akamai (approved):
I agree with you, Banks, and Dajahi the code fits the scope of the ticket. Though I think adding the RQ invalidation Dajahi recommended would be important otherwise we might cause a lot of confusion and generate a lot of tickets for support.

bnussman:
Good call on the invalidation.

```ts
queryClient.invalidateQueries(`object-storage-buckets`);
```

mjac0bs:
@bnussman @dwiley-akamai Can you verify that RQ invalidation is working for you now with my latest changes? I invalidated the queries for both buckets and access keys, since canceling obj storage suggests that both of those things should be deleted.

CodyFinn-Akamai (approved):
Confirmed the RQ invalidation works

dwiley-akamai:
> @bnussman @dwiley-akamai Can you verify that RQ invalidation is working for you now with my latest changes? I invalidated the queries for both buckets and access keys, since canceling obj storage suggests that both of those things should be deleted.

LGTM 🚀

bnussman (approved):

PR #8616, opened 2022-11-17
M3-5996: Cluster Pricing UI updates

dwiley-akamai:
## Description 📝
When an OCA with a UDF regarding clusters is selected, display a notice regarding the number of clusters and the creation of an additional temporary node. Additionally, add text regarding the temporary node and pricing for multiple nodes to the Summary panel.

## Preview 📷
Notice:
Screen Shot 2022-11-18 at 3 08 28 PM

Summary:
Screen Shot 2022-11-18 at 3 09 04 PM

## How to test 🧪
Test the OCA flow with an OCA with clusters and an OCA without clusters. The former should show features like the above screenshots; the latter should be unchanged from their current functionality.

StackScript deployments from other areas of the app should be tested as well to ensure they've not been adversely impacted.

## To-Do
- [x] Test related flows to ensure no adverse effects were created
- [x] Some potential refactors
- [x] Possibly more unit tests

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
@@ -1,11 +1,15 @@
 import { Interface, restoreBackup } from '@linode/api-v4/lib/linodes';
 import { Tag } from '@linode/api-v4/lib/tags/types';
+import { cloneDeep } from 'lodash';
 import * as React from 'react';
 import { connect, MapDispatchToProps, MapStateToProps } from 'react-redux';
 import { RouteComponentProps } from 'react-router-dom';
 import { compose as recompose } from 'recompose';
 import AccessPanel from 'src/components/AccessPanel';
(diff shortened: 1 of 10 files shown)


packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
```diff
@@ -250,6 +254,12 @@ export class LinodeCreate extends React.PureComponent<
     });
   };
 
+  setNumberOfNodesForUDFSummary = (num: number) => {
+    this.setState({
+      numberOfNodes: num,
+    });
+  };
```
dwiley-akamai, line 261: What's basically happening with the number of nodes is that it gets set via default or user selection in `<UserDefinedFieldsPanel />`, so this function is created here to pass it down through as a callback function. `this.state.numberOfNodes` is then consumed in `<CheckoutSummary />` to display the correct number for the temporary node.

bnussman-akamai (approved):

PR #8620, opened 2022-11-21
M3-5994: [Changed] One Click Cluster App Card

CodyFinn-Akamai:
## Description 📝

Adds the ability to add a chip to add components to the right of a card's heading. This is used to add a Chip to the right of the heading on the SelectionCards of one click apps that are targeted at clusters.

## Preview 📷

image

## How to test 🧪

1. In a text editor open the `AppPanelSection.tsx` file
2. On line 53 set the value of `isCluster` to `true`
3. Navigate to the marketplace
4. Ensure Chip indicating that an app is deployed on a cluster is in the correct place.

`yarn test CardBase`

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/AppPanelSection.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/AppPanelSection.tsx
@@ -68,7 +68,7 @@ export const AppPanelSection: React.FC<Props> = (props) => {
               openDrawer={openDrawer}
               disabled={disabled}
               iconUrl={eachApp.logo_url.toLowerCase() || ''}
-              labelDecoration={isCluster ? <Chip label="Cluster" /> : undefined}
+              labelDecoration={isCluster ? <Chip label="CLUSTER" /> : undefined}
             />
           );
         })}
--- a/packages/manager/src/features/linodes/LinodesCreate/AppPanelSection.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/AppPanelSection.tsx
@@ -50,10 +50,16 @@ export const AppPanelSection: React.FC<Props> = (props) => {
       ) : null}
       <Grid className={classes.flatImagePanelSelections} container>
         {apps.map((eachApp) => {
+          const decodedLabel = decode(eachApp.label);
           const isCluster =
-            eachApp.user_defined_fields.find(
+            decodedLabel.endsWith('Cluster ') &&
+            eachApp.user_defined_fields.some(
               (field) => field.name === 'node_options'
-            ) !== undefined;
+            );
+
+          const label = isCluster
+            ? decodedLabel.split(' Cluster')[0]
+            : decodedLabel;
 
           return (
             <SelectionCardWrapper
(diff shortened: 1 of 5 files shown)


packages/manager/src/features/linodes/LinodesCreate/AppPanelSection.tsx
```diff
@@ -48,21 +49,29 @@ export const AppPanelSection: React.FC<Props> = (props) => {
...
+              checked={eachApp.id === selectedStackScriptID}
+              // Decode App labels since they may contain HTML entities.
+              label={decode(eachApp.label)}
+              availableImages={eachApp.images}
+              userDefinedFields={eachApp.user_defined_fields}
+              handleClick={handleClick}
+              openDrawer={openDrawer}
+              disabled={disabled}
+              iconUrl={eachApp.logo_url.toLowerCase() || ''}
+              labelDecoration={isCluster ? <Chip label="CLUSTER" /> : undefined}
```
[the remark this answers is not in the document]
CodyFinn-Akamai: Here is what a smaller one would look like 
image

CodyFinn-Akamai:
> Can we remove the `Cluster` text from the app name as seen in the mockup? The chip text also turns blue when the app is selected
> 
> image
> 
> I just realized this now looks very similar to the `PaymentMethodCard` component which uses `renderVariant` to render the chip. I wonder if we can just use that function to render both the chip and the info icon so that we don't have to add a new `headingDecoration` prop to SelectionCard
> 
> image

The `PaymentMethodCard` component does not include the info icon that needs to be right aligned. The Chip's location is dictated by the end of the Card's label. Putting it in the `renderVariant` causes more issues with how the flexbox works than just putting the Chip in the heading. I think the SelectionCards need a rework given how many sections there are on them. Might be another good candidate for css grid in the future.

packages/manager/src/features/linodes/LinodesCreate/AppPanelSection.tsx
```diff
@@ -48,21 +54,39 @@ export const AppPanelSection: React.FC<Props> = (props) => {
...
-            handleClick={handleClick}
-            openDrawer={openDrawer}
-            disabled={disabled}
-            iconUrl={eachApp.logo_url.toLowerCase() || ''}
-          />
-        ))}
+        {apps.map((eachApp) => {
+          const decodedLabel = decode(eachApp.label);
+          const isCluster =
+            decodedLabel.endsWith('Cluster ') &&
```
bnussman-akamai, line 60: Why do we need the trailing space here?

bnussman-akamai (approved):
The card looks good visually!

packages/manager/src/features/linodes/LinodesCreate/AppPanelSection.tsx
CodyFinn-Akamai: Yeah the `LinodeCreateContainer` has a method called `trimOneClickFromLabel` and it returns it with the extra space.

hana-linode (approved):

mjac0bs (approved):
Confirmed that cluster chip looks good in both dark and light modes, no color change on click. Unit test still looked good. 👍

PR #8621, opened 2022-11-21
M3-5991: SMTP add helper text

mjac0bs:
## Description 📝

**What does this PR do?**
Adds two SMTP restriction notices to Cloud Manager for customers with _**accounts created on or after**_ the next release date.

Notice 1—at the bottom of the Linode create page 
Notice 2—on the Linode detail page, above tab set, with the Network tab selected

## Preview 📷

**Notice 1:**

Without EU agreement:
Screen Shot 2022-11-21 at 2 31 17 PM

With EU agreement:
Screen Shot 2022-11-21 at 2 31 48 PM

**Notice 2:**
Screen Shot 2022-11-21 at 2 32 58 PM

## How to test 🧪

- Turn on MSW to mock a user who created an account on the day of the next release.

Notice 1:
- Navigate to the Linode Create page. Scroll to bottom of page and observe SMTP text.
- Set the new linode's region to London, UK or Frankfurt, DE 
- Scroll to the bottom of the page and observe SMTP text and EU agreement text.
- Adjust the screen size and confirm that the message(s) and Create Linode button at the bottom of the page are still legible and spacing is not weird.

Notice 2:
- Navigate to the Linode Details page of any linode. Click on the Network tab. Observe the dismissible notice. Select a different tab and confirm the notice is not visible.
- On the Network tab, confirm that the notice is dismissible by clicking the 'X'. 
- In `serverHandlers.ts`, adjust the `active_since` value in `rest.get('*/account', (req, res, ctx)` to dates before and after the next release. Ensure that SMTP helper text only appears for accounts active _on or after **11-28-22**_. 

Run `yarn test SMTPRestrictionText.test.ts`.

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodesDetailNavigation.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodesDetailNavigation.tsx
@@ -103,7 +103,7 @@ const LinodesDetailNavigation: React.FC<CombinedProps> = (props) => {
         segment={`${linodeLabel} - ${tabs[getIndex()]?.title ?? 'Detail View'}`}
       />
       {tabs[getIndex()]?.title === 'Network' ? (
-        <SMTPRestrictionText>
+        <SMTPRestrictionText supportLink={true}>
           {({ text }) =>
             !isNoticeDismissed && text !== null ? (
               <Notice
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodesDetailNavigation.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodesDetailNavigation.tsx
@@ -41,6 +41,8 @@ const LinodesDetailNavigation: React.FC<CombinedProps> = (props) => {
     match: { url },
   } = props;
 
+  const [isNoticeDismissed, setIsNoticeDismissed] = React.useState(false);
+
   // Bare metal Linodes have a very different detail view
   const isBareMetalInstance = linodeType?.class === 'metal';
 
@@ -102,13 +104,20 @@ const LinodesDetailNavigation: React.FC<CombinedProps> = (props) => {
       />
       {tabs[getIndex()]?.title === 'Network' ? (
         <SMTPRestrictionText>
-          {({ text }) => (
-            <Notice warning dismissible onClose={() => null} spacingTop={32}>
-              <Grid item xs={12}>
-                {text}
-              </Grid>
-            </Notice>
-          )}
+          {({ text }) =>
+            !isNoticeDismissed && text !== null ? (
+              <Notice
+                warning
+                dismissible
+                onClose={() => setIsNoticeDismissed(true)}
+                spacingTop={32}
+              >
+                <Grid item xs={12}>
+                  {text}
+                </Grid>
+              </Notice>
+            ) : null
+          }
         </SMTPRestrictionText>
       ) : null}
       <div style={{ marginTop: 8 }}>
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodesDetailNavigation.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodesDetailNavigation.tsx
@@ -103,7 +103,7 @@ const LinodesDetailNavigation: React.FC<CombinedProps> = (props) => {
       {tabs[getIndex()]?.title === 'Network' ? (
         <SMTPRestrictionText>
           {({ text }) => (
-            <Notice warning dismissible onClose={() => null}>
+            <Notice warning dismissible onClose={() => null} spacingTop={32}>
               <Grid item xs={12}>
                 {text}
               </Grid>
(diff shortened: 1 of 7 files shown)


matthewbartholomew:
I think these pages look good! I have a couple of comments.

1. **Notice 1:** I think we should remove the link to "open a support ticket" because I believe you need to have created the Linode before you can ask for the ports to be unblocked. I can confirm. 
2. **Notice 1 with EU agreement:** The EU notice is a one-time event so even though it adds a lot of text at the bottom of the page, the stacked structure of the two notices is easily understood.

mjac0bs:
> I believe you need to have created the Linode before you can ask for the ports to be unblocked. I can confirm.

@matthewbartholomew 

The [old SMTP restriction text](https://github.com/linode/manager/pull/8621/commits/86c051eaf3e8def872ee356649d0de187b4fd63d#diff-e7882d3f23e97038a547ea25b6032d3772309f50af5f2b63ed5bcbef61766212L34) from the last time we displayed this notice seems to suggest this. Once you confirm, (a) I can remove the link to the support ticket on the Linode Create page only or (b) remove it entirely, but it seems helpful to include on the Linode Details page, since the linode already exists.

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailNavigation.tsx
```diff
@@ -97,6 +102,24 @@ const LinodesDetailNavigation: React.FC<CombinedProps> = (props) => {
...
+                onClose={() => setIsNoticeDismissed(true)}
+                spacingTop={32}
+              >
+                <Grid item xs={12}>
+                  {text}
+                </Grid>
+              </Notice>
+            ) : null
+          }
+        </SMTPRestrictionText>
```
CodyFinn-Akamai, line 121: Nice use of renderProps!

CodyFinn-Akamai (approved):
Looks good to me!

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailNavigation.tsx
```diff
@@ -38,6 +41,8 @@ const LinodesDetailNavigation: React.FC<CombinedProps> = (props) => {
     match: { url },
   } = props;
 
+  const [isNoticeDismissed, setIsNoticeDismissed] = React.useState(false);
```
dwiley-akamai, line 44: I don't think the dismissal will persist with the current logic (e.g., a user dismisses the notice but then refreshes the page or navigates away and then back to the Network tab), assuming a persisted dismissal is the desired effect for this ticket.

`<DismissibleBanner />` would be better suited to address this, but is contained within a `<Paper />` which would depart from the mockups

jdamore-linode (approved):
Looking good! I have a couple thoughts/questions but nothing that's worthy of holding up the merge.

- It might be good to add unit tests for the component itself. It's good that we have the date comparison logic covered, but a quick test to confirm that the component only renders when `accountCreatedAfterRestrictions()` returns `true` would be good to have too.

- What do these changes mean for people who created their account after 11/6/2019 but before tomorrow's release? Will this mean that port 25, etc., remains blocked for them, but they won't be shown any message to inform them of that? (Sorry if this has already been addressed, I think I was out during the planning meeting(s) that involved this ticket!)

I'm running E2E tests right now (so far so good, 42/51) and I'm assuming the rest will pass, but I'll follow up if any of them fail.

dwiley-akamai (approved):
Confirmed this works as intended with real data -- approving pending resolution of the bolding and capitalization points for the notice.

packages/manager/src/features/linodes/SMTPRestrictionText.tsx
```diff
@@ -1,19 +1,22 @@
...
 import ExternalLink from 'src/components/ExternalLink';
+import SupportLink from 'src/components/SupportLink';
 import { useAccount } from 'src/queries/account';
 import { sendLinodeCreateDocsEvent } from 'src/utilities/ga';
 
 // "In an effort to fight spam, Linode restricts outbound connections on ports 25, 465, and 587 on all Linodes for new accounts created after November 5th, 2019."
 // https://www.linode.com/docs/email/best-practices/running-a-mail-server/
 const MAGIC_DATE_THAT_EMAIL_RESTRICTIONS_WERE_IMPLEMENTED =
-  '2019-11-06T00:00:00.000Z';
+  '2022-11-28T00:00:00.000Z'; // Date of release for Manager v81.0.
```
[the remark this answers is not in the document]
mjac0bs: Yep, addressing these both in a future commit! Dajahi caught the versioning typo is his review but I accidentally force-pushed over it. Thanks for catching these too.

mjac0bs:
@jdamore-linode 

> It might be good to add unit tests for the component itself. It's good that we have the date comparison logic covered, but a quick test to confirm that the component only renders when `accountCreatedAfterRestrictions()` returns `true` would be good to have too.

Thanks for your help with this. We're almost there, but I'm going to add these in a follow-up PR so we be sure to get this one merged before the release.

> What do these changes mean for people who created their account after 11/6/2019 but before tomorrow's release? Will this mean that port 25, etc., remains blocked for them, but they won't be shown any message to inform them of that? 

Yeah, there's been some discussion of this in a channel about SMTP notices. The decision that came out of that was that, since most of our SMTP requests come from new customers + in effort to avoid displaying a potentially irrelevant notice to older customers, we're going to show the notice for new customers where "new" == account created on/after tomorrow's release. Anyone with an account created after 11/6/2019 but before 11/29/22 could still have restrictions in place, but would not see the message.

PR #8622, opened 2022-11-22
M3-5922: Add machine image update and delete E2E tests

jdamore-linode:
## Description 📝

**What does this PR do?**
Adds flows to the machine image E2E test to verify that updating and deleting machine images works as expected.

I opted to add flows to the existing image creation test rather than creating separate tests since the data setup (i.e. creating an image) can take a minute or two.

## How to test 🧪

Run `yarn dev`, and then run:

```bash
yarn cy:run -s "cypress/e2e/images/machine-image-upload.spec.ts"
```
And confirm that all of the tests pass.

To run the tests interactively, run `yarn cy:debug` and search for `machine-image-upload.spec.ts`.

### DIFF
--- a/packages/manager/cypress/e2e/images/machine-image-upload.spec.ts
+++ b/packages/manager/cypress/e2e/images/machine-image-upload.spec.ts
@@ -1,5 +1,5 @@
-import { EventStatus } from '@linode/api-v4/lib/account/types';
-import { ImageStatus } from '@linode/api-v4/lib/images/types';
+import { EventStatus } from '@linode/api-v4/types';
+import { ImageStatus } from '@linode/api-v4/types';
 import { eventFactory, imageFactory } from '@src/factories';
 import { makeResourcePage } from '@src/mocks/serverHandlers';
 import 'cypress-file-upload';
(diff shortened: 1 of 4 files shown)


bnussman-akamai (approved):
All 5 tests passed ✅ 

Generally regarding Cloud's e2e testing, what do we generally try to do about picking regions? Should we use the same region every time, randomize, test all regions?

jdamore-linode:
> All 5 tests passed ✅
> 
> Regarding Cloud's e2e testing, what do we generally try to do about picking regions? Should we use the same region every time, randomize, test all regions?

Thanks! Most of the time we randomly choose a region (this test does too), but we do have quite a few cases where the region used in the test is hardcoded.

There are arguments to be made for or against any approach, but I think using a random region for each test strikes a decent middle ground where we get reasonable coverage across our regions without having to run an excessive number of tests. I think there's an argument to be made that certain critical operations (like creating a Linode) could be tested against every region, though.

hana-linode (approved):
All tests passed 🚢

mjac0bs:
I ran interactively with `yarn cy:debug` and was seeing a Cypress error for the first test (`uploads, updates, and deletes a machine image, end-to-end`) returned at `cypress/e2e/images/machine-image-upload.spec.ts:189:66`where "**cy.click() can only be called on a single element. Your subject contained 2 elements. Pass { multiple: true } if you want to serially click each element**". All 3 attempted runs of that test failed.

Overall, 4/5 tests passed for me due to the failure above. Any idea what could be causing a failing test on my end when others have 5/5 successfully passing?

jdamore-linode:
Thanks @mjac0bs! I'll take a closer look at this tomorrow, but I think I have an understanding of what's going on. By any chance, did you already have one or more machine images on your account when you ran the test?

mjac0bs:
> did you already have one or more machine images on your account when you ran the test?

@jdamore-linode - It looks like I did have one additional machine image already! When I deleted that image and reran the tests, that first test did pass. Based on that original error, it seems like Cypress was trying to select both images or didn't know which to choose.

jdamore-linode:
> > did you already have one or more machine images on your account when you ran the test?
> 
> @jdamore-linode - It looks like I did have one additional machine image already! When I deleted that image and reran the tests, that first test did pass. Based on that original error, it seems like Cypress was trying to select both images or didn't know which to choose.

That's pretty much exactly what's happening! The action menus (the "..." menus) are all present in the DOM, but only visible when they're open -- the test was tripping up because Cypress was selecting the wrong action menu.

Just pushed a fix for that, thanks again! I'm super glad you caught that because it could've led to some hard-to-troubleshoot failures later on.

PR #8623, opened 2022-11-22
M3-5995: UDF copy updates

hana-linode:
## Description 📝
- Change placeholder text from "Enter your password" to "Enter your token" for API token field
- Add tooltip about setting up an API token next to the token input field

## Preview 📷
image

## How to test 🧪
Click on a Marketplace app that has an API token input field under Setup such as Mongodb Cluster

### DIFF
--- a/packages/manager/src/components/TextField.tsx
+++ b/packages/manager/src/components/TextField.tsx
@@ -9,6 +9,7 @@ import InputAdornment from 'src/components/core/InputAdornment';
 import InputLabel from 'src/components/core/InputLabel';
 import { makeStyles, Theme, WithTheme } from 'src/components/core/styles';
 import TextField, { TextFieldProps } from 'src/components/core/TextField';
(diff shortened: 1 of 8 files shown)


packages/manager/src/components/TextField.tsx
```diff
@@ -152,6 +153,7 @@ export const LinodeTextField: React.FC<CombinedProps> = (props) => {
     tooltipPosition,
     tooltipText,
     tooltipClasses,
+    tooltipInteractive,
     tooltipOnMouseEnter,
```
bnussman-akamai, line 157: This is completely optional, but because we have so many tooltip props, we may want to change this to one prop called something like `tooltipProps` that is of type `TooltipProps`.

bnussman-akamai (approved):

packages/manager/src/components/TextField.tsx
CodyFinn-Akamai: I second this

CodyFinn-Akamai (approved):

PR #8629, opened 2022-11-29
M3-5197: React Query Volumes - Fixes and Feedback on Feature Branch

bnussman-akamai:
## Description 📝

- This PR will be used to address any feedback or bugs found in `feature/react-query-volumes`

### DIFF
--- a/packages/manager/src/App.tsx
+++ b/packages/manager/src/App.tsx
@@ -120,7 +120,7 @@ export class App extends React.Component<CombinedProps, State> {
       .subscribe(domainEventsHandler);
 
     /*
-     * Send any Volume events to the Domain events handler in the queries file
+     * Send any Volume events to the Volume events handler in the queries file
      */
     events$
       .filter((event) => event.action.startsWith('volume') && !event._initial)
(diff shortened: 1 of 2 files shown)


bnussman:
@jdamore-linode Does 6bccb45 seem correct to fix the volume attach stability?

jdamore-linode:
@bnussman I still haven't been able to reproduce that failure so it's hard to say :/ But the tests continue to pass for me with your changes in place, so they certainly don't hurt.

I'm still trying to figure it out, but my thought is that this shouldn't stop us from merging the RQ changes as long as we've thoroughly tested volume attaching manually. Hopefully we won't see any failures, but if there is flake it'll be a lot easier to troubleshoot once this is merged to `develop` and we get it logged and recorded in the Cypress dashboard.

For now I'm going to focus on code review/manual testing, but if I have any time left over I may reach out to Mariah (cc @mjac0bs) to take another stab at reproducing/understanding this!

jdamore-linode:
@bnussman Just to clarify, I think the cause of the failure is that the test is clicking the "Attach" button before the form has had a chance to update after selecting a Linode -- if that's the case, I'm pretty sure your commit will fix the issue 👍

bnussman:
I also saw the issue @mjac0bs saw.

The failures make since because with the RQ changes, a Linode's configs need to be loaded to attach a volume. Previously, we didn't enforce configs to be loaded to attach a volume.

dwiley-akamai (approved):
LGTM now 👍

jdamore-linode (approved):

PR #8634, opened 2022-12-05
M3-6000: Add missing region text in Database summary view

carrillo-erik:
## Description 📝
Simple PR to show the region in summary view for Databases.

## Preview 📷

image

### DIFF
--- a/packages/manager/cypress/e2e/databases/update-database.spec.ts
+++ b/packages/manager/cypress/e2e/databases/update-database.spec.ts
@@ -2,7 +2,7 @@
  * @file DBaaS integration tests for update operations.
  */
 
-import { regions } from 'support/constants/regions';
+import { regions, regionsMap } from 'support/constants/regions';
 import {
   randomLabel,
   randomItem,
@@ -176,6 +176,10 @@ describe('Update database clusters', () => {
     cy.visitWithLogin(`/databases/${database.engine}/${database.id}`);
     cy.wait('@getDatabase');
 
+    cy.get('[data-qa-cluster-config]').within(() => {
+      cy.findByText(regionsMap[database.region]).should('be.visible');
+    });
+
     cy.get('[data-qa-connection-details]').within(() => {
       // "Show" button should be enabled to reveal password when DB is active.
       cy.findByText('Show')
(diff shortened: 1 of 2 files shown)


jdamore-linode (approved):
This is a nice improvement! Looks good to me 👍

dwiley-akamai (approved):
Adding the region here might seem like a minor change, but believe it or not, it has been a requested addition for some time now 😅

I let @bavedarnow this will be getting rolled out with our next release and he raised the question of whether an assertion should be added to our E2E tests, maybe somewhere around https://github.com/linode/manager/blob/develop/packages/manager/cypress/e2e/databases/create-database.spec.ts#L143-L153? cc @jdamore-linode

carrillo-erik:
> > This might seem like a minor change, but believe it or not, it has been a requested addition for some time now 😅
> > I let @bavedarnow this will be getting rolled out with our next release and he raised the question of whether an assertion should be added to our E2E tests, maybe somewhere around https://github.com/linode/manager/blob/develop/packages/manager/cypress/e2e/databases/create-database.spec.ts#L143-L153? cc @jdamore-linode
> 
> ++
> 
> @carrillo-erik I'm happy to make this addition, but this is a super simple change. If you're at all curious about how our E2E tests work, this would be a great opportunity to get some hands on experience with them! I'd be happy to help you do this over a ~15 minute meeting if you have the time!

@jdamore-linode Yes, I'd love to go over out E2E tests and I think a 15 minute meeting would be great.

packages/manager/cypress/e2e/databases/update-database.spec.ts
```diff
@@ -247,7 +251,7 @@ describe('Update database clusters', () => {
    * - Confirms that users cannot reset root passwords for provisioning DBs.
    * - Confirms that users cannot change maintenance schedules for provisioning DBs.
    */
-  it('Cannot update database clusters while they are provisioning', () => {
+  it.skip('Cannot update database clusters while they are provisioning', () => {
```
jdamore-linode, line 254:
```suggestion
  it('Cannot update database clusters while they are provisioning', () => {
```
Ah sorry about this, I forgot to remind you to delete the `.skip` once we finished development! Applying this suggestion should take care of it.

Otherwise looking good!

PR #8636, opened 2022-12-06
M3-5988: Add SMTP restriction support ticket form

mjac0bs:
## Description 📝

**What does this PR do?**

Adds form fields specific to lifting SMTP restrictions to a support ticket. This is part of efforts to improve the user experience and reduce the amount of back-and-forth between customers and Support when handling requests to remove these restrictions.

**Notes**: 

- You may notice when testing various types of support tickets that the title and description fields are prepopulated with data that does not belong to that support ticket, but was from a previous support ticket, likely because form values are cached in localstorage and not cleared. This issue seems to exist in production prior to these changes and is beyond the scope of this ticket, but could be addressed in another.
- Will create an additional ticket to transition `SupportTicketDrawer` to Formik for improved form state management.

## Preview 📷
**Trigger and fill out the support ticket:**

https://user-images.githubusercontent.com/114685994/206579933-9f1caa2f-7eac-482f-8d5c-4cabe9b85c33.mov

**Submitted support ticket:**
Screen Shot 2022-12-08 at 2 22 00 PM

## How to test 🧪

1. **Use the dev environment for the following to avoid generating test support tickets in production:**
2. Use an account that has been created since _or temporarily modified_ to have an account creation date of on/after 11-30-22 (the date we activated the SMTP restriction notices for new accounts).
3. Navigate to the linode details page for a linode and select the network tab.
4. Click on the "open a support ticket" link in the SMTP restriction notice.
5. Observe that you've been redirected to the Support page with an open support dialog that has form fields specific to SMTP.
6. Observe that the form does not submit until a user enters all required fields. 
7. Fill out all required fields and submit the ticket. 
8. Observe that the user is redirected to the ticket details page. Check that ticket details look correct and readable.
9. Check that Support can view the ticket and that details look correct and readable.
10. Confirm that a user can access the form dialog for a general support ticket and a support ticket with entities, and the SMTP form does _not_ display. (Turning on MSW and navigating to the Volumes Landing page to click the volume with "Contact Support" in its status is a quick way.)

Units tests:
```
yarn test SMTPRestrictionText
```
```
yarn test Support
```

e2e:
```
yarn cy:run -s "cypress/e2e/helpAndSupport/open-support-ticket.spec.ts"
```


packages/manager/src/features/Support/SupportTickets/SupportTicketDrawer.tsx
```diff
@@ -447,7 +501,63 @@ export const SupportTicketDrawer: React.FC<CombinedProps> = (props) => {
...
+        break;
+      case 'general':
+      default:
+        helperText = general;
+    }
+
+    return (
+      <Typography data-qa-support-ticket-helper-text>{helperText}</Typography>
+    );
+  };
```
hana-linode, line 550: This could be simplified by defining a mapping outside of the component like:

```typescript
interface TicketTypeData {
  dialogTitle: string;
  helperText: string | JSX.Element;
}

const ticketTypeMap: Record<TicketType, TicketTypeData> = {
  smtp: {
    dialogTitle: smtpDialogTitle,
    helperText: smtpHelperText,
  },
  general: {
    dialogTitle: 'Open a Support Ticket',
    helperText: (
      <>
        {`We love our customers, and we\u{2019}re here to help if you need us.
        Please keep in mind that not all topics are within the scope of our support.
        For overall system status, please see `}
        <a
          href="https://status.linode.com"
          target="_blank"
          aria-describedby="external-site"
          rel="noopener noreferrer"
        >
          status.linode.com
        </a>
        .
      </>
    ),
  },
};
```

hana-linode (commented):
SMTP form info seems to display in the desc when clicking on `Open New Ticket` in the Support Ticket landing page 🤔 

image


Should we pass in the Linode entity as well so that it shows up in the `Regarding` column in the Support ticket landing table and in the details page?

image

image

packages/manager/src/features/Support/SupportTickets/SupportTicketDrawer.tsx
```diff
@@ -321,6 +355,26 @@ export const SupportTicketDrawer: React.FC<CombinedProps> = (props) => {
     setEntityID(String(selected?.value) ?? '');
   };
 
+  const handleSMTPFieldChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const { name, value } = e.target;
+    setSMTPFields((smtpFields) => {
+      return { ...smtpFields, [name]: value };
+    });
```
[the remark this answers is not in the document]
mjac0bs: Whoops, good catch, thanks.

jdamore-linode:
In addition to what @hana-linode mentioned about the `Open a Support Ticket` ticket form being pre-populated, I don't seem to be able to get the "Open Ticket" submit button to become enabled, seemingly preventing me from opening any non-SMTP related tickets.

Reproduce:
1. Navigate to `/support/tickets`
2. Click "Open New Ticket"
3. Observe form is pre-populated with SMTP content
4. Observe that "Open Ticket" button is disabled
5. Mess around with the form content, observe that "Open Ticket" button remains disabled

(This is also reproduced in the `open-support-ticket.spec.ts` e2e tests)

packages/manager/src/features/Support/SupportTickets/SupportTicketDrawer.tsx
```diff
@@ -447,7 +501,63 @@ export const SupportTicketDrawer: React.FC<CombinedProps> = (props) => {
...
+      case 'smtp':
+        helperText = smtpHelperText;
+        break;
+      case 'general':
+      default:
+        helperText = general;
+    }
+
+    return (
+      <Typography data-qa-support-ticket-helper-text>{helperText}</Typography>
```
cpathipa, line 548: Suggestion: I think we can simplify this as below

```suggestion
   const isGeneralTicket = ticketType === 'general';

    return (
      <Typography data-qa-support-ticket-helper-text>
        {isGeneralTicket ? general : smtpHelperText}
      </Typography>
    );
```

cpathipa (approved):
Great start @mjac0bs I have couple minor suggestions to simplify setting helper test based on ticketType. Other than that over all LGTM!

I agree with @hana-linode feedback on defining outside of the component

mjac0bs:
@hana-linode and @jdamore-linode 

Latest commit should address issue with inability to submit non-SMTP support tickets - thanks for catching that @jdamore-linode. Can you confirm it's working now?

 I also added in the entity, so it should display now when an SMTP ticket is submitted as @hana-linode suggested.

Still working on the saved description... need to figure out track down that value isn't being reset when the drawer is closed.

packages/manager/src/features/Support/SupportTickets/SupportTicketDrawer.tsx
```diff
@@ -181,6 +241,10 @@ export const SupportTicketDrawer: React.FC<CombinedProps> = (props) => {
     }
   }, [open]);
 
+  React.useEffect(() => {
+    setDescription(formatDescription(smtpFields));
+  }, [smtpFields]);
+
```
hana-linode, line 247: Looks like this might be the culprit for why smtp fields are displaying in the desc for normal support tickets. If I comment this out and clear my application storage, the desc is no longer filled.

hana-linode: Maybe only format the description on submission and not on every change would be a fix?

mjac0bs: Hmm. I tried adding a conditional to only `setDescription(formatDescription(smtpFields));` if the `ticketType === 'smtp'` but found that the desc is still filled because ticketType doesn't get reset from `smtp` to `general`, although that was [my intent](https://github.com/linode/manager/blob/c64e6f9a9fa8179949d54a3b99c96e22bc729ada/packages/manager/src/features/Support/SupportTickets/SupportTicketDrawer.tsx#L350). 

Good point. I had some issues when trying to format the description on submission before, but the code has changed since then, so I'll take another look!

mjac0bs: Okay, so there is different behavior for the dialog based on how the user cancels it, as documented in [this PR](https://github.com/linode/manager/pull/7512):

- If the user hits the Cancel button, the SMTP desc is not visible in the desc of normal support tickets. (Expected behavior).
- If the user hits the 'X', ESC, or clicks outside of the dialog, the contents of the form are saved in local storage rather than cleared. Because the close method didn't know anything about the new SMTP `ticketType`, it was saving that description to local storage and then repopulating it for a normal ticket's description.

Latest changes should fix this.

jdamore-linode (approved):
Ha! I was just about to post about an issue I was seeing with the ticket contents but I think your latest commit fixed it.

Confirmed that the submit button is fixed and that non-SMTP support tickets can be submitted, confirmed that the SMTP ticket template is no longer pre-populated for non-SMTP tickets, and confirmed that the E2E support test is now passing.

Nicely done @mjac0bs!

packages/manager/src/features/Support/SupportTickets/SupportTicketDrawer.tsx
mjac0bs: Disregard the last bit of above ^^  Formatting of the description now only occurs on submission, with an adjustment to the logic that toggles the boolean controlling the submit button from disabled to enabled.

I left the conditional checking `ticketType` in `close()` in order to always clear the title of the SMTP ticket when the form is closed out. Note: there is still an edge case when a form title can repopulate in a new support ticket's title field if the user was first redirected to `support/tickets` with any support ticket having a prefilled title. This behavior exists in prod and predates the SMTP form addition.

mjac0bs:
> Ha! I was just about to post about an issue I was seeing with the ticket contents but I think your latest commit fixed it.

@jdamore-linode Yeah, there was an edge case that either needed another conditional checking `ticketType` in the `useEffect` hook with `smtpFields`... or I could go the better route and make the changes needed to format description only once, `onSubmit`. Second option implemented.

Thanks for confirming that behavior and E2E is looking good to you!

PR #8637, opened 2022-12-06
M3-5511: Improve Linode Migration Time Estimates

bnussman-akamai:
## Description 📝

- Updates the rates that Linodes are migrated at so that our Linode Migration estimates are more accurate

### Before
Screen Shot 2022-12-06 at 11 51 32 AM

### After
Screen Shot 2022-12-06 at 11 51 08 AM

## How to test 🧪

- Verify that the estimates in this PR are lower than estimates in production for the same Linode
  - There is no promise that this estimate is correct, but now it should be closer to the actual time it takes to migrate a Linode


CodyFinn-Akamai (approved):
Time estimates show lower estimates

PR #8639, opened 2022-12-06
M3-5753: Filter deprecated images

cpathipa:
## Description 📝

Goal is to replace manual process of setting flags in DB to get suffix ( "deprecated") added to distro's  in the manager.
- Automatically filters deprecated images 
- Automatically Adds suffix "deprecated"

## Preview 📷

**Before:** 
image

**After:**
image

## How to test 🧪

- Navigate to route /linodes
- Click on create linode
- Validate Images in "Choose a Distribution"

`yarn test components/ImageSelect/ImageSelect`

### DIFF
--- a/packages/manager/src/components/ImageSelect/ImageSelect.tsx
+++ b/packages/manager/src/components/ImageSelect/ImageSelect.tsx
@@ -87,37 +87,29 @@ export const imagesToGroupedItems = (images: Image[]) => {
         draft.push({
           label: thisGroup,
           options: group
-            .filter((thisImage) => {
-              const { eol } = thisImage;
-              if (!eol) {
-                // no end of life is known so show Image to be safe
-                return true;
-              }
-              const diff = DateTime.now().diff(DateTime.fromISO(eol), 'months')
-                .months;
-
+            .reduce((acc: ImageItem[], thisImage) => {
+              const { created, eol, id, label, vendor } = thisImage;
+              const differenceInMonths = DateTime.now().diff(
+                DateTime.fromISO(eol!),
+                'months'
+              ).months;
               // if image is past its end of life, hide it, otherwise show it
-              return diff > MAX_MONTHS_EOL_FILTER ? false : true;
-            })
-            .map((thisImage) => {
-              const { eol } = thisImage;
-              const diff = DateTime.now().diff(DateTime.fromISO(eol!), 'months')
-                .months;
-              let fullLabel = thisImage.label;
-              // Add suffix 'depricated' to the image if end of life is past
-              if (diff > 0) {
-                fullLabel += ' (Deprecated)';
+              if (!eol || differenceInMonths < MAX_MONTHS_EOL_FILTER) {
+                acc.push({
+                  created,
+                  // Add suffix 'depricated' to the image at end of life.
+                  label:
+                    differenceInMonths > 0 ? `${label} (Deprecated)` : label,
+                  value: id,
+                  className: vendor
+                    ? // Use Tux as a fallback.
+                      `fl-${distroIcons[vendor] ?? 'tux'}`
+                    : `fl-tux`,
+                });
               }
-              return {
-                created: thisImage.created,
-                label: fullLabel,
-                value: thisImage.id,
-                className: thisImage.vendor
-                  ? // Use Tux as a fallback.
-                    `fl-${distroIcons[thisImage.vendor] ?? 'tux'}`
-                  : `fl-tux`,
-              };
-            })
+
+              return acc;
+            }, [])
             .sort(sortByImageVersion),
         });
       });
--- a/packages/manager/src/components/ImageSelect/ImageSelect.tsx
+++ b/packages/manager/src/components/ImageSelect/ImageSelect.tsx
@@ -1,5 +1,6 @@
 import { Image } from '@linode/api-v4/lib/images';
 import produce from 'immer';
+import { DateTime } from 'luxon';
 import { equals, groupBy } from 'ramda';
 import * as React from 'react';
 import Paper from 'src/components/core/Paper';
@@ -16,6 +17,7 @@ import getSelectedOptionFromGroupedOptions from 'src/utilities/getSelectedOption
 import { distroIcons } from './icons';
 import ImageOption from './ImageOption';
 
+const maxEOLFilter = 6;
 export type Variant = 'public' | 'private' | 'all';
 
 interface ImageItem extends Item<string> {
@@ -85,6 +87,18 @@ export const imagesToGroupedItems = (images: Image[]) => {
         draft.push({
           label: thisGroup,
           options: group
+            .filter((thisImage) => {
+              const { deprecated, eol } = thisImage;
+              if (deprecated) {
+                const diff = DateTime.now().diff(
+                  DateTime.fromISO(eol),
+                  'months'
+                ).months;
+                return Number.isNaN(diff) || diff <= maxEOLFilter;
+              } else {
+                return true;
+              }
+            })
             .map((thisImage) => {
               const isDeprecated = thisImage.deprecated;
               const fullLabel =
--- a/packages/manager/src/components/ImageSelect/ImageSelect.tsx
+++ b/packages/manager/src/components/ImageSelect/ImageSelect.tsx
@@ -88,21 +88,19 @@ export const imagesToGroupedItems = (images: Image[]) => {
           label: thisGroup,
           options: group
             .filter((thisImage) => {
-              const { deprecated, eol } = thisImage;
-              if (deprecated) {
-                const diff = DateTime.now().diff(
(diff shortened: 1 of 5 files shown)


packages/manager/src/components/ImageSelect/ImageSelect.tsx
```diff
@@ -16,6 +17,7 @@ import getSelectedOptionFromGroupedOptions from 'src/utilities/getSelectedOption
 import { distroIcons } from './icons';
 import ImageOption from './ImageOption';
 
+const maxEOLFilter = 6;
```
hana-linode, line 20: Can we rename this to something like `maxMonthsEOLFilter` so we know what the unit is? And maybe also move this variable to `manager/src/constants.ts`

packages/manager/src/components/ImageSelect/ImageSelect.tsx
```diff
@@ -85,10 +87,20 @@ export const imagesToGroupedItems = (images: Image[]) => {
...
-              const isDeprecated = thisImage.deprecated;
-              const fullLabel =
-                thisImage.label + (isDeprecated ? ' (Deprecated)' : '');
+              const { eol } = thisImage;
+              const diff = DateTime.fromISO(eol!).diff(DateTime.now(), 'months')
+                .months;
+              let fullLabel = thisImage.label;
+              if (diff <= maxEOLFilter) {
+                fullLabel += ' (Deprecated)';
+              }
```
CodyFinn-Akamai, line 103: It is possible to put some of the filtering on the server by using the X-Filter header in the request. `deprecated` is a filterable field. The curl command would look something like this: 

```
curl "https://api.linode.com/v4/images" \
  -H 'X-Filter: { "deprecated": "false" }'
```

packages/manager/src/components/ImageSelect/ImageSelect.tsx
cpathipa: @hana-linode Yes, I agree, `maxMonthsEOLFilter` is more descriptive. Updated accordingly and moved it to `constants`.

packages/manager/src/components/ImageSelect/ImageSelect.tsx
```diff
@@ -85,10 +87,22 @@ export const imagesToGroupedItems = (images: Image[]) => {
...
+              return diff > MAX_MONTHS_EOL_FILTER && eol !== null
+                ? false
+                : true;
+            })
             .map((thisImage) => {
-              const isDeprecated = thisImage.deprecated;
-              const fullLabel =
-                thisImage.label + (isDeprecated ? ' (Deprecated)' : '');
+              const { eol } = thisImage;
+              const diff = DateTime.fromISO(eol!).diff(DateTime.now(), 'months')
```
hana-linode, line 100: There seems to be some kind of bug with the time difference, we're showing some images as Deprecated when they haven't reached EOL yet:

- Alpine 3.14 (EOL 2023-05-01)
- Fedora 36 (EOL 2023-05-16)
- Fedora 35 (2023-12-15)
- Ubuntu 18.04 (EOL 2023-04-30)

We should also figure out if we should add Deprecated if the `deprecated` field is true such as for Slackware 14.1

hana-linode (changes requested):

packages/manager/src/components/ImageSelect/ImageSelect.tsx
```diff
@@ -85,10 +87,22 @@ export const imagesToGroupedItems = (images: Image[]) => {
         draft.push({
           label: thisGroup,
           options: group
+            .filter((thisImage) => {
+              const { eol } = thisImage;
+              const diff = DateTime.now().diff(DateTime.fromISO(eol!), 'months')
+                .months;
+              return diff > MAX_MONTHS_EOL_FILTER && eol !== null
+                ? false
+                : true;
```
bnussman-akamai, line 96: The non-null assertion (`!`) caught my attention. I found that when `eol` was `null`, `diff` would be `NaN`. Optionally, we could check if it is null before we calculate the time difference. This change is up to you, it results in the same behavior.



```suggestion
              const { eol } = thisImage;
              if (!eol) {
                // no end of life is known so show Image to be safe
                return true;
              }
              const diff = DateTime.now().diff(DateTime.fromISO(eol), 'months')
                .months;

              // if image is past its end of life, hide it, otherwise show it
              return diff > MAX_MONTHS_EOL_FILTER ? false : true;
```

bnussman-akamai (approved):
Nice job on this logic!

packages/manager/src/components/ImageSelect/ImageSelect.tsx
cpathipa: Good catch @hana-linode 

- Above time difference issue was fixed
- Yes, we no longer rely on `deprecated` Since it's a manual process of setting in DB for deprecated images. So, we should rely on `eol` field to dynamically add suffix / filter images that are past end of life.

cpathipa:
Thank you @hana-linode and @bnussman-akamai for the feedback. Addressed the changes, enhanced unit test coverage and PR is ready for re-review.

packages/manager/src/components/ImageSelect/ImageSelect.tsx
```diff
@@ -85,10 +87,27 @@ export const imagesToGroupedItems = (images: Image[]) => {
...
-              const fullLabel =
-                thisImage.label + (isDeprecated ? ' (Deprecated)' : '');
+              const { eol } = thisImage;
+              const diff = DateTime.now().diff(DateTime.fromISO(eol!), 'months')
+                .months;
+              let fullLabel = thisImage.label;
+              // Add suffix 'depricated' to the image if end of life is past
+              if (diff > 0) {
+                fullLabel += ' (Deprecated)';
+              }
```
CodyFinn-Akamai, line 110: This could be simplified with a reduce instead of a filter followed by a map

packages/manager/src/components/ImageSelect/ImageSelect.tsx
```diff
@@ -85,10 +87,27 @@ export const imagesToGroupedItems = (images: Image[]) => {
         draft.push({
           label: thisGroup,
           options: group
+            .filter((thisImage) => {
+              const { eol } = thisImage;
+              if (!eol) {
+                // no end of life is known so show Image to be safe
+                return true;
+              }
+              const diff = DateTime.now().diff(DateTime.fromISO(eol), 'months')
```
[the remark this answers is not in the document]
cpathipa: sure, renamed `diff` ->  `differenceInMonths`

packages/manager/src/components/ImageSelect/ImageSelect.tsx
cpathipa: @CodyFinn-Akamai Good call, we can definitely cut down number of operations either with `forEach | reduce`

packages/manager/src/components/ImageSelect/ImageSelect.tsx
```diff
@@ -85,10 +87,27 @@ export const imagesToGroupedItems = (images: Image[]) => {
         draft.push({
           label: thisGroup,
           options: group
+            .filter((thisImage) => {
+              const { eol } = thisImage;
```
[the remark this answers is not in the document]
cpathipa: @hana-linode I like the idea. But, The only downside I see with this is they overridable.

packages/manager/src/components/ImageSelect/ImageSelect.tsx
```diff
@@ -85,10 +87,27 @@ export const imagesToGroupedItems = (images: Image[]) => {
...
+                // no end of life is known so show Image to be safe
+                return true;
+              }
+              const diff = DateTime.now().diff(DateTime.fromISO(eol), 'months')
+                .months;
+
+              // if image is past its end of life, hide it, otherwise show it
+              return diff > MAX_MONTHS_EOL_FILTER ? false : true;
+            })
             .map((thisImage) => {
```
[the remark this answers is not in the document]
cpathipa: @hana-linode The only downside I see is they are overridable. Addressed this within the fun [c8a1955]([c8a1955](https://github.com/linode/manager/pull/8639/commits/c8a195524f9819aa41397f9cf391aef5ed996e82))

cpathipa:
@CodyFinn-Akamai and @hana-linode 
Addressed below feedback changes. 
- Replaced map and filter with reduce
- Renamed diff -> differenceInMonths
- Destructuring

CodyFinn-Akamai (approved):

hana-linode (approved):

packages/manager/src/components/ImageSelect/ImageSelect.tsx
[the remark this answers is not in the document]
hana-linode: Good point, didn't know they were overridable

PR #8640, opened 2022-12-06
M3-5855: Fix delayed loading of /56 ranges in Linode Network tab

hana-linode:
## Description 📝
- Reduce unnecessary requests to a linode instance's `/ips` endpoint
- Fix delayed loading of /56 & /64 IPv6 ranges
    - This was due to a change made in `LinodeNetworking.tsx` (https://github.com/linode/manager/pull/8082)
    - IPv6 ranges were only displayed after we got IPv6 range info for _all_ IPv6 ranges on the account. And getting one IPv6 range info is one request. So, the more IPv6 ranges an account has, the more requests we need to make and the longer it takes for the last request to finish
    - For just viewing the Linode's IPv6 addresses, we didn't need to wait for all the requests to be finished first. The info we needed to display the IP addresses were already available in the Linode instance's `/ips` response

## Preview 📷
(Linode was deleted after this)
Local

https://user-images.githubusercontent.com/115299789/205991561-4032245d-4b33-48f0-8d54-8d94ac072e2a.mov

Prod

https://user-images.githubusercontent.com/115299789/205991622-977dc97d-5d7d-46ac-958c-4cfcffcf2848.mov

## How to test 🧪
1. Go to the Network tab of a Linode with a /56 range (reach out for info on how to get one)
2. In your browser's dev tools, throttle your network connection to `Slow 3G`
3. Reload the page and observe how the /56 ranges display immediately locally compared to prod
4. Ensure that IPv6 IP sharing still works the same as prod (only certain Data Centers are supported and both Linodes must be in the same DC; use Linodes with the `Frankfurt, DE` region)

```
yarn test LinodeNetworking
```

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
@@ -259,7 +259,6 @@ class LinodeNetworking extends React.Component<CombinedProps, State> {
           }
 
           // get info on an IPv6 range; if its shared check if its shared to our Linode
-          // if its not shared (!is_bgp) figure out if its a slaac address or a statically routed range
           const resp = await getIPv6RangeInfo(range.range);
 
           if (
@@ -914,7 +913,7 @@ export const ipResponseToDisplayRows = (
     ipDisplay.push(ipToDisplay(ipv6?.link_local, 'Link Local'));
   }
 
-  // IPv6 ranges and pools (/116s) to display in the networking table
+  // IPv6 ranges and pools to display in the networking table
   ipDisplay.push(
     ...[...(ipv6 ? ipv6.global : [])].map((thisIP) => {
       /* If you want to surface rdns info in the future you have two options:
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
@@ -201,7 +201,7 @@ class LinodeNetworking extends React.Component<CombinedProps, State> {
       ipv6Error: undefined,
     });
 
-    const refreshIPv4 = getLinodeIPs(this.props.linode.id)
+    const linodeIPs = getLinodeIPs(this.props.linode.id)
       .then((ips) => {
         const hasIPv6Range = ips.ipv6 && ips.ipv6.global.length > 0;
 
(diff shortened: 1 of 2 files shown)


CodyFinn-Akamai (approved):
Works nicely!

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
```diff
@@ -201,7 +201,7 @@ class LinodeNetworking extends React.Component<CombinedProps, State> {
       ipv6Error: undefined,
     });
 
-    const refreshIPv4 = getLinodeIPs(this.props.linode.id)
+    const linodeIPs = getLinodeIPs(this.props.linode.id)
```
cpathipa, line 204: `linodeIPs` is more relevant than `refreshIPv4`

cpathipa (approved):
Nice work @hana-linode on this fix. I see seamless loading of IP addresses in the table.

mjac0bs (approved):
Functionality LGTM -- /56 addresses loaded immediately with the throttled connection and IPv6 sharing seemed to have the same behavior as prod. 

When testing this yesterday, I noticed an unexpected mismatched type error message in prod that is getting triggered when I was trying to save multiple /56 IP addresses at once in the IP Sharing dialog. Out of the scope of this ticket, but I'll follow up with you to make sure it's a bug we'd want to fix in the future and haven't already accounted for it.

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
```diff
@@ -929,14 +914,9 @@ export const ipResponseToDisplayRows = (
     ipDisplay.push(ipToDisplay(ipv6?.link_local, 'Link Local'));
   }
 
-  // IPv6 pools (/116s) and routed ranges to display in the networking table
+  // IPv6 ranges and pools (/116s) to display in the networking table
```
dwiley-akamai, line 917: This comment should be updated as well if we're not filtering for `/116`'s anymore

packages/manager/src/features/linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
```diff
@@ -262,12 +261,6 @@ class LinodeNetworking extends React.Component<CombinedProps, State> {
           // get info on an IPv6 range; if its shared check if its shared to our Linode
           // if its not shared (!is_bgp) figure out if its a slaac address or a statically routed range
```
dwiley-akamai, line 262: With the removal of some logic related to `slaac`, this comment should be updated as well

dwiley-akamai (approved):
functionality LGTM 👌

PR #8645, opened 2022-12-08
M3-5182: RQ-ify Images endpoints, Pt. 1

dwiley-akamai:
## Description 📝
The first of two or three PRs related to transitioning Images out of Redux and into React Query.

In terms of functionality, this PR focuses on the Images landing page and the "Capture Image" tab when you go to create an image, and having them rely on RQ instead of Redux.

The next piece in this effort will be the "Upload Image" tab and associated functionality/infrastructure.

## How to test 🧪
Ensure functionality parity of Images landing & the "Capture Image" tab between this branch and prod.

Confirm that the RQ cache is storing and updating the Images data as you would expect. 

## To-Do:
- [x] Fix bug where Image label and descriptions aren't being updated; ensure RQ cache is updated accordingly
- [x] Fix bug where redirection from "Rebuild an Existing Linode" flow doesn't open the modal on Linode Details page 
- [x] Convert `ImagesDrawer.tsx` to a function component
- [x] Fix status bug on landing page after a newly-created Image is ready
- [x] Cleanup

### DIFF
--- a/packages/manager/src/features/Images/ImagesDrawer.tsx
+++ b/packages/manager/src/features/Images/ImagesDrawer.tsx
@@ -412,7 +412,14 @@ export const ImageDrawer: React.FC<CombinedProps> = (props) => {
 
       <ActionsPanel
         style={{ marginTop: 16 }}
-        updateFor={[requirementsMet, classes, submitting, mode]}
+        updateFor={[
+          requirementsMet,
+          classes,
+          submitting,
+          mode,
+          label,
+          description,
+        ]}
       >
         <Button
           onClick={close}
--- a/packages/manager/src/features/Images/ImagesDrawer.tsx
+++ b/packages/manager/src/features/Images/ImagesDrawer.tsx
@@ -90,6 +90,10 @@ export const ImageDrawer: React.FC<CombinedProps> = (props) => {
     enqueueSnackbar,
     changeLabel,
     changeDescription,
+    changeLinode,
+    changeDisk,
+    open,
+    onClose,
   } = props;
 
   // console.log(label);
@@ -143,12 +147,12 @@ export const ImageDrawer: React.FC<CombinedProps> = (props) => {
   }, [props.disks]);
 
   React.useEffect(() => {
-    if (!props.selectedLinode) {
+    if (!selectedLinode) {
       setDisks([]);
     }
 
-    if (props.selectedLinode) {
-      getLinodeDisks(props.selectedLinode)
+    if (selectedLinode) {
+      getLinodeDisks(selectedLinode)
         .then((response) => {
           if (!mounted) {
             return;
@@ -176,22 +180,22 @@ export const ImageDrawer: React.FC<CombinedProps> = (props) => {
           }
         });
     }
-  }, [props.selectedLinode]);
+  }, [selectedLinode]);
 
   const handleLinodeChange = (linodeID: number) => {
     // Clear any errors
     setErrors(undefined);
-    props.changeLinode(linodeID);
+    changeLinode(linodeID);
   };
 
   const handleDiskChange = (diskID: string | null) => {
     // Clear any errors
     setErrors(undefined);
-    props.changeDisk(diskID);
+    changeDisk(diskID);
   };
 
   const close = () => {
-    props.onClose();
+    onClose();
     if (mounted) {
       setErrors(undefined);
       setNotice(undefined);
@@ -204,7 +208,7 @@ export const ImageDrawer: React.FC<CombinedProps> = (props) => {
   const onSubmit = () => {
     setErrors(undefined);
     setNotice(undefined);
-    setSubmitting(false);
+    setSubmitting(true);
 
     switch (mode) {
       case 'edit':
@@ -290,8 +294,6 @@ export const ImageDrawer: React.FC<CombinedProps> = (props) => {
     // When creating an image, disable the submit button until a Linode and
     // disk are selected. When restoring to an existing Linode, the Linode select is the only field.
     // When imagizing, the Linode is selected already so only check for a disk selection.
-    const { mode, selectedDisk, selectedLinode } = props;
-
     const isDiskSelected = Boolean(selectedDisk);
 
     switch (mode) {
@@ -325,7 +327,7 @@ export const ImageDrawer: React.FC<CombinedProps> = (props) => {
   const diskError = hasErrorFor('disk_id');
 
   return (
-    <Drawer open={props.open} onClose={props.onClose} title={titleMap[mode]}>
+    <Drawer open={open} onClose={onClose} title={titleMap[mode]}>
       {!canCreateImage ? (
         <Notice
           error
--- a/packages/manager/src/features/Images/ImagesDrawer.tsx
+++ b/packages/manager/src/features/Images/ImagesDrawer.tsx
@@ -182,7 +182,7 @@ export const ImageDrawer: React.FC<CombinedProps> = (props) => {
     }
   };
 
-  const safeDescription = description ?? ' ';
+  const safeDescription = description ? description : ' ';
 
   const onSubmit = () => {
     setErrors(undefined);
--- a/packages/manager/src/features/Images/ImagesDrawer.tsx
+++ b/packages/manager/src/features/Images/ImagesDrawer.tsx
@@ -17,11 +17,11 @@ import { IMAGE_DEFAULT_LIMIT } from 'src/constants';
 import { resetEventsPolling } from 'src/eventsPolling';
 import DiskSelect from 'src/features/linodes/DiskSelect';
 import LinodeSelect from 'src/features/linodes/LinodeSelect';
+import { useImageAndLinodeGrantCheck } from 'src/hooks/useImages';
 import {
   useCreateImageMutation,
   useUpdateImageMutation,
 } from 'src/queries/images';
-import { getGrantData, getProfileData } from 'src/queries/profile';
 import { getAPIErrorOrDefault } from 'src/utilities/errorUtils';
 import getAPIErrorFor from 'src/utilities/getAPIErrorFor';
 
@@ -97,6 +97,10 @@ export const ImageDrawer: React.FC<CombinedProps> = (props) => {
   const classes = useStyles();
   const { enqueueSnackbar } = useSnackbar();
   const history = useHistory();
+  const {
+    canCreateImage,
+    permissionedLinodes: availableLinodes,
+  } = useImageAndLinodeGrantCheck();
 
   const [mounted, setMounted] = React.useState<boolean>(false);
   const [notice, setNotice] = React.useState(undefined);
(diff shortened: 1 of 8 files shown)


packages/manager/src/queries/base.ts
```diff
@@ -214,7 +214,7 @@ export const itemInListDeletionHandler = <
  */
 export const updateInPaginatedStore = <T extends { id: number | string }>(
   queryKey: string,
-  id: number,
+  id: number | string,
```
dwiley-akamai, line 217: For images, the id's come in as strings, e.g., `private/123456789`, `linode/debian11`, etc., so this change is to account for that.

packages/manager/src/store/index.ts
```diff
@@ -36,6 +36,7 @@ import globalErrors, {
   defaultState as defaultGlobalErrorState,
   State as GlobalErrorState,
 } from 'src/store/globalErrors';
+// Remove
```
dwiley-akamai, line 39: Just placeholder notes. At the end of the RQ-ify'ing process, image-related pieces will be removed from this file and elsewhere in `src/store`.

dwiley-akamai:
Still working through a couple of unfortunate bugs with the `ImagesDrawer.tsx` refactor so I will leave the WIP label on for a tad longer until those are sorted out

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -143,11 +141,65 @@ const defaultDialogState = {
...
+  );
+
+  const { mutateAsync: deleteImage } = useDeleteImageMutation();
+
+  const eventState = useSelector((state: ApplicationState) => state.events);
+  const events = imageEvents(eventState);
+
+  // We want private images and also want the associated events tied in.
+  const imagesItemsById = listToItemsByID(images?.data ?? []);
+  const imagesData = Object.values(imagesItemsById).reduce(
```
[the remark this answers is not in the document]
dwiley-akamai: I do advocate for this approach because of performance (especially because this involves in-progress events, so there can be a lot of looping happening), but were there any specific concerns you had about maintainability? We don't often have to touch this code block and I can't think of a situation where this would really hinder us

dwiley-akamai:
> Great start!
> 
> I think as we continue RQ for Images we should consider refactoring the landing page to use some of our new table patterns such as not using the EntityTable component and start using API sorting (see Volumes for an example). The landing page can also be broken up into 2 useImagesQuery with filters such as `'X-Filter: { "is_public": false, "type": "manual" }'`. I think the comment in line 208 of `ImagesLanding.tsx` refers to this. I'm sure a lot of my comments were already on your radar.

Thanks! I agree with most of this and will make a note to circle back for these changes before the RQ work for Images is wrapped up.

dwiley-akamai:
Going to go ahead and remove the WIP label now.

One minor bug that I'm aware of is when you rename an image and then immediately enter the "Deploy to New Linode" flow, the preselected image will have the image's prior label. This is because that Create flow is reliant on Redux still for Images data (i.e., right now, the data is getting updated in the RQ store but not in the Redux store). I've made a note to circle back to that at the end of this process and make sure the bug is fixed at that time.

packages/manager/src/features/Images/ImagesDrawer.tsx
```diff
@@ -63,189 +56,184 @@ export interface Props {
...
           .filter((thisGrant) => thisGrant.permissions === 'read_write')
           .map((thisGrant) => thisGrant.id) ?? []
       : null;
 
-    if (this.props.disks) {
-      // for the 'imagizing' mode
-      this.setState({ disks: this.props.disks });
-    }
-  }
+    setAvailableImages(images);
```
bnussman-akamai, line 137: Can all of this be moved outside of the `useEffect` and converted to hook form?

packages/manager/src/features/Images/ImagesDrawer.tsx
```diff
@@ -326,161 +308,144 @@ class ImageDrawer extends React.Component<CombinedProps, State> {
...
+      </ActionsPanel>
+    </Drawer>
+  );
+};
 
 export default compose<CombinedProps, Props>(
-  styled,
   withRouter,
-  withImages(),
   withSnackbar,
```
bnussman-akamai, line 449: Remove snackbar references in this file because it does not look like we use it.

packages/manager/src/features/Images/ImagesDrawer.tsx
```diff
@@ -63,189 +56,184 @@ export interface Props {
...
-
-  componentDidMount() {
-    this.mounted = true;
+export const ImageDrawer: React.FC<CombinedProps> = (props) => {
+  const {
+    mode,
+    imageId,
+    label,
+    description,
+    history,
```
bnussman-akamai, line 87:
```suggestion
```

Remove `history` from props and `withRouter` and switch to hook.

bnussman-akamai (approved):
Approving to keep you unblocked but left a few more minor comments

packages/manager/src/features/Images/ImagesDrawer.tsx
dwiley-akamai: We do (Linode Storage tab --> Disks --> Imagize has an "Image scheduled for creation." snackbar upon success), but I replaced the snackbar wrapper with the hook 👌

packages/manager/src/features/Images/ImagesDrawer.tsx
bnussman-akamai: This would just help mitigate extra side-effect renders. For example `canCreateImage` could be derived without needed its own state.

mjac0bs (commented):
I'm not the RQ wizard that Banks is, but I tested the functionality for a while and noticed a couple of things:

1. Updating an image's description field from some string to a blank field does not seem to update the `description` field to an empty string in RQ as it does in Redux. Therefore, when the user opens the same image's Edit Image drawer again, the old description, which should have been deleted, was repopulated. On prod, clearing the `description` field and then reopening the drawer does result in a blank description.

> One minor bug that I'm aware of is when you rename an image and then immediately enter the "Deploy to New Linode" flow, the preselected image will have the image's prior label. This is because that Create flow is reliant on Redux still for Images data (i.e., right now, the data is getting updated in the RQ store but not in the Redux store).

^^ This bug sounded somewhat similar at first, but seems separate since we're still on the Images Landing page and shouldn't be pulling from Redux anymore.

2. When I was messing around with sorting the tables, whenever I hit the combination of filters like `http://localhost:3000/images?order=asc&orderBy=status` or `http://localhost:3000/images?order=desc&orderBy=status`, the data in the tables sorts fine. However, when I refreshed the page *or* tried to navigate directly to that URL containing the query, I received the error state on the page with following error in the browser console

```
    {
        "reason": "Could not apply filter",
        "field": "X-Filter"
    }
```

In prod, filtered images tables do seem to load without errors after a page refresh and when navigating to that URL directly (e.g. `https://cloud.linode.com/images?order=asc&orderBy=status`).

dwiley-akamai:
> 1. Updating an image's description field from some string to a blank field does not seem to update the `description` field to an empty string in RQ as it does in Redux. Therefore, when the user opens the same image's Edit Image drawer again, the old description, which should have been deleted, was repopulated. On prod, clearing the `description` field and then reopening the drawer does result in a blank description.

This was due to me using the nullish coalescing operator 🥲 Just reverted that.

> 2. When I was messing around with sorting the tables, whenever I hit the combination of filters like `http://localhost:3000/images?order=asc&orderBy=status` or `http://localhost:3000/images?order=desc&orderBy=status`, the data in the tables sorts fine. However, when I refreshed the page _or_ tried to navigate directly to that URL containing the query, I received the error state on the page with following error in the browser console
> 
> ```
>     {
>         "reason": "Could not apply filter",
>         "field": "X-Filter"
>     }
> ```
> 
> In prod, filtered images tables do seem to load without errors after a page refresh and when navigating to that URL directly (e.g. `https://cloud.linode.com/images?order=asc&orderBy=status`).

Hmm, I can't replicate that behavior:

https://user-images.githubusercontent.com/114682940/208191769-fecbbaf6-5666-43d3-a61f-35cf6cb5ce55.mov

packages/manager/src/hooks/useImages.ts
```diff
@@ -62,4 +63,23 @@ export const filterImages = (
...
+  const canCreateImage =
+    Boolean(!profile?.restricted) || Boolean(grants?.global?.add_images);
+
+  // Unrestricted users can create Images from any disk;
+  // Restricted users need read_write on the Linode they're trying to Imagize
+  // (in addition to the global add_images grant).
+  const permissionedLinodes = profile?.restricted
+    ? grants?.linode
+        .filter((thisGrant) => thisGrant.permissions === 'read_write')
+        .map((thisGrant) => thisGrant.id) ?? []
```
cpathipa, line 79: Nice to have: Can we replace `filter` `map` either with `foreach` or `reducer`.

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -143,11 +141,65 @@ const defaultDialogState = {
...
+          const matchingEvent = events.find(
+            (thisEvent) =>
+              `private/${thisEvent.secondary_entity?.id}` === thisImage.id ||
+              (`private/${thisEvent.entity?.id}` === thisImage.id &&
+                thisEvent.status === 'failed')
+          );
+          if (matchingEvent) {
+            draft.push({ ...thisImage, event: matchingEvent });
+          } else {
+            draft.push(thisImage);
```
cpathipa, line 197: Nice to have: Can we replace `if else` with `ternary` ?  But, it's up to you, totally optional. 
```suggestion
          draft.push(
            !matchingEvent ? thisImage : { ...thisImage, event: matchingEvent }
          );
```

packages/manager/src/queries/images.ts
```diff
@@ -0,0 +1,124 @@
...
+
+export const useImagesQuery = (params: any, filters: any) =>
+  useQuery<ResourcePage<Image>, APIError[]>(
+    [`${queryKey}-list`, params, filters],
+    () => getImages(params, filters),
+    { keepPreviousData: true }
+  );
+
+// Get specific Image
+export const useSingleImageQuery = (imageID: string) =>
```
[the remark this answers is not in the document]
cpathipa: I agree with that..

cpathipa (approved):
Overall LGTM @dwiley-akamai Nice work! I only left few comments for nit-picky things. But those are completely optional.

dwiley-akamai:
Joe and I were able to replicate the sorting issue this morning that Mariah mentioned. I'm going to investigate it more. If there's a fix to be made on our side I'll make those adjustments in the Pt. 2 PR; if it proves to be an API issue, I'll open a ticket with them regarding it.

Going to merge this now to keep things moving along but the Pt. 2 PR will incorporate some of the other feedback here as well.

PR #8646, opened 2022-12-08
Update imageFactory eol fied as null

cpathipa:
## Description 📝

End of life field is set to null in `imageFactory` As our e2e specs are using this factory to build mock data.
This fixes failed 2e2 test `create-linode-from-image.spec.ts`

## How to test 🧪
Navigate to `imageFactory` validate end of life filed

yarn cypress:debug 
Then run spec file `create-linode-from-image.spec.ts` under Images

### DIFF
--- a/packages/manager/cypress/support/api/images.ts
+++ b/packages/manager/cypress/support/api/images.ts
@@ -6,10 +6,13 @@ export const getImages = (page: number = 1) => getAll(`images?page=${page}`);
 
 export const createMockImage = (
   data?,
+  eol = null,
   label = 'cy-test-image',
   id = 'private/99999999'
 ) => {
-  return makeResourcePage(imageFactory.buildList(1, { label, id, ...data }));
+  return makeResourcePage(
+    imageFactory.buildList(1, { eol, label, id, ...data })
+  );
 };
 
 export const deleteImageById = (imageId: number) =>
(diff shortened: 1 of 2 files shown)


cpathipa:
@jdamore-linode Created PR to address the failed spec `create-linode-from-image.spec.ts.`

**Thought:** Wondering it would be worth if we update the spec that leverages this `imageFactory` and build expected data from it. My thought is we can make this spec more robust and avoid breakages in future because of changing any default values in the `imageFactory`. But, I'm open your thoughts on this.

jdamore-linode:
> @jdamore-linode Created PR to address the failed spec `create-linode-from-image.spec.ts.`
> 
> **Thought:** Wondering it would be worth if we update the spec that leverages this `imageFactory` and build expected data from it. My thought is we can make this spec more robust and avoid breakages in future because of changing any default values in the `imageFactory`. But, open your thoughts on this.

@cpathipa Not a bad idea at all! I would probably suggest modifying the `createMockImage()` helper function in `packages/manager/cypress/support/images.ts` to explicitly set `eol` to `null`, and then it may be worth reconsidering whether we should change the factory default at all

jdamore-linode (changes requested):
Looking good so far!

(Edit) Whoops, disregard the previous post! I think this is good to go once you make that helper function change :)

jdamore-linode (approved):

dwiley-akamai (commented):
I agree with Joe; this sort of change should be made closer to where it is consumed for the E2E test instead of at the factory level.

dwiley-akamai (approved):
🚢

cpathipa:
@jdamore-linode @dwiley-akamai Sounds good to me, I meant the same. As, we have flexibility to build the mock data with custom values from factory w.r.t to test scenario. So, reverted the initial commit made to factory `imageFactory` and updated `createMockImage`

PR #8650, opened 2022-12-12
M3-4876: Vertically center breadcrumb, Docs, and buttons

CodyFinn-Akamai:
## Description 📝

Centers the breadcrumb and action items across the app.

### Before
image

### After
image

## How to test 🧪

Go through the app and verify that the breadcrumb and action buttons at the top of the page are consistent across pages.

### DIFF
--- a/packages/manager/src/features/linodes/LinodesLanding/LinodesLanding.tsx
+++ b/packages/manager/src/features/linodes/LinodesLanding/LinodesLanding.tsx
@@ -309,8 +309,14 @@ export class ListLinodes extends React.Component<CombinedProps, State> {
           <MaintenanceBanner />
         )}
         <Grid container className={classes.root}>
-          <Grid item xs={12}>
+          <Grid item xs={12} style={{ padding: 0 }}>
             <DocumentTitleSegment segment="Linodes" />
+            <LandingHeader
+              title="Linodes"
+              entity="Linode"
+              onAddNew={() => this.props.history.push('/linodes/create')}
+              docsLink="https://www.linode.com/docs/platform/billing-and-support/linode-beginners-guide/"
+            />
             <PreferenceToggle<boolean>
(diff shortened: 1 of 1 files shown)


packages/manager/src/features/linodes/LinodesLanding/LinodesLanding.tsx
```diff
@@ -309,8 +309,7 @@ export class ListLinodes extends React.Component<CombinedProps, State> {
           <MaintenanceBanner />
         )}
         <Grid container className={classes.root}>
-          <Grid item xs={12}>
-            <DocumentTitleSegment segment="Linodes" />
+          <Grid item xs={12} style={{ padding: 0 }}>
```
cpathipa, line 312: @CodyFinn-Akamai 
Only thing I observed with `padding:0` is causing one of the child element spill over from parent element. 
**With padding: 0**
image

**Without padding:0**
image

cpathipa (commented):
@CodyFinn-Akamai

Issue: `padding:0` is causing one of the child element to spill over from the parent element.
I see few screenshots are  attached to the ticket, are we addressing the fix for all  of them as part of this ticket?

packages/manager/src/features/linodes/LinodesLanding/LinodesLanding.tsx
```diff
@@ -309,8 +309,7 @@ export class ListLinodes extends React.Component<CombinedProps, State> {
           <MaintenanceBanner />
         )}
         <Grid container className={classes.root}>
-          <Grid item xs={12}>
-            <DocumentTitleSegment segment="Linodes" />
```
bnussman-akamai, line 313: This may need to be kept. The title is not showing how I would expect it to 

https://user-images.githubusercontent.com/115251059/208707169-fa7b651a-cc02-45f3-bc57-adffe878f506.mov

PR #8656, opened 2022-12-12
Update the docs search URL

wildmanonline:
## Description 📝

Updates the `DOCS_SEARCH_URL` constant to use a more recent URL. It now also searches product docs, api docs, and reference architectures in addition to guides. If there is another place where the URL needs to be updated, please make those changes.

## How to test 🧪

Perform a search where the `DOCS_SEARCH_URL` constant is used (likely within the Help & Support section). Use a product related term (such as `NodeBalancers`) and verify that product documentation appears in the search box.


bnussman-akamai (approved):

bnussman-akamai:
@wildmanonline When would you like this merged?

PR #8657, opened 2022-12-13
M3-6008: Search Marketplace apps by category, alternate name and alternate description keywords

cpathipa:
## Description 📝
Adds the ability to search apps in Linode Marketplace by

- category,
- alternative names (ex. TF2 and team fortress 2), and
- description keywords

## Preview 📷

image

## How to test 🧪

- Navigate to marketplace -> `/linodes/create?type=One-Click`
- Search app by category || name || alt names || alt description keywords
- Validate results
- Example: search by category (Databases) -> expected to see Database related apps

  yarn cy:run


packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
```diff
@@ -209,15 +209,39 @@ class FromAppsContent extends React.Component<CombinedProps, State> {
...
+       * */
+      const matchingOCALabels = oneClickApps.reduce(
+        (acc: string[], { categories, name, alt_name, alt_description }) => {
+          const hasMatchingOCA = JSON.stringify(
+            `${categories.join(
+              ' '
+            )} ${name} ${alt_name} ${alt_description}`.replace(/,/g, '')
+          )
+            .toLocaleLowerCase()
+            .includes(query.toLocaleLowerCase());
```
CodyFinn-Akamai, line 227: Includes isn't the correct method to use here. We might be better served by the `.match` method and using a regexp against the categories string that is made instead to find any matches as opposed to looping through the categories or the query terms

CodyFinn-Akamai: We could also split the query based on spaces and then use lodash array function to compare the arrays in order to see if they have any overalp

carrillo-erik (approved):

packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
```diff
@@ -209,15 +209,47 @@ class FromAppsContent extends React.Component<CombinedProps, State> {
...
+       * */
+      const queryWords = query
+        .replace(/[,.-]/g, '')
+        .trim()
+        .toLocaleLowerCase()
+        .split(' ');
+
+      const matchingOCALabels = oneClickApps.reduce(
+        (acc: string[], { categories, name, alt_name, alt_description }) => {
+          const ocaAppString = JSON.stringify(
```
dwiley-akamai, line 227: Is using `JSON.stringify()` necessary here? I removed it during testing and functionality seems to be unaffected.

packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
```diff
@@ -209,15 +209,47 @@ class FromAppsContent extends React.Component<CombinedProps, State> {
...
+        }
+      );
+
       this.setState({
         isFiltering: false,
-        filteredApps: appsMatchingQuery,
+        filteredApps: filteredAppInstances,
         isSearching: true,
         categoryFilter: null,
         query: query,
```
dwiley-akamai, line 255: Linter points out we can use the property shorthand here

packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
```diff
@@ -209,15 +209,47 @@ class FromAppsContent extends React.Component<CombinedProps, State> {
...
+          if (hasMatchingOCA) {
+            acc.push(name.trim());
+          }
+
+          return acc;
+        },
+        []
       );
+
+      const filteredAppInstances = this.props.appInstances?.filter(
```
dwiley-akamai, line 244: I do find the previous variable name `appsMatchingQuery` to be clearer in this context

packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
```diff
@@ -209,15 +209,47 @@ class FromAppsContent extends React.Component<CombinedProps, State> {
...
+        .replace(/[,.-]/g, '')
+        .trim()
+        .toLocaleLowerCase()
+        .split(' ');
+
+      const matchingOCALabels = oneClickApps.reduce(
+        (acc: string[], { categories, name, alt_name, alt_description }) => {
+          const ocaAppString = JSON.stringify(
+            `${name} ${alt_name} ${categories.join(' ')} ${alt_description}`
+          ).toLocaleLowerCase();
```
dwiley-akamai, line 229: This `ocaAppString` logic is also running _n_ times (where _n_ = the number of apps) on every search query, which is not ideal.

Let's create a function outside the class component that generates the stringified values once, and then the matching in this function can just use that. So, for example, we could do something like:

```
const stringifiedOneClickApps = oneClickApps.map(
  ({ categories, name, alt_name, alt_description }) => {
    return {
      name,
      stringifiedAppData: `${name} ${alt_name} ${categories.join(
        ' '
      )} ${alt_description}`.toLocaleLowerCase(),
    };
  }
);
```

and then here

```
      const matchingOCALabels = stringifiedOneClickApps.reduce(
        (
          acc: string[],
          currentApp: { name: string; stringifiedAppData: string }
        ) => {
          const hasMatchingOCA = queryWords.every(
            (queryWord) => currentApp.stringifiedAppData.indexOf(queryWord) >= 0
          );

          if (hasMatchingOCA) {
            acc.push(currentApp.name.trim());
          }

          return acc;
        },
        []
      );
```

(variable names in this sample code can be improved)

dwiley-akamai (commented):
The functionality is looking good 👍 I left a few comments for potential refactoring and improvements.

packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
CodyFinn-Akamai: We should probably add a check to see if the user typed a category name. Then we can more easily grab filter down the list. Then we can ignore the categories section if they don't lowering the amount of work needed to combine all the search terms. We could also combine all the search terms into a regexp that will match if any of the terms are in the name, alt-name, or alt-description.

cpathipa: @dwiley-akamai 
I like the thought process of decoupling the `create` function outside of component, where it avoids performing unnecessary operations when component re-renders. However, in this case logic of building string is wrapped within the event handler `onSearch` on component re-render it only registers the handler, but doesn't execute it. 

Also, The only downside I see in the above  example,  it performs 2x times more operations w.r.t to `oneClickApps`  length.  (ex: if length of the array is 100, with the combination of map and reduce it performs 200 operations to build and filter the matching app labels.) So, I'm leaning towards existing way as number of operations is purely depends on the size of the input.

@CodyFinn-Akamai I think it's over complicates the logic, we should be good, as long as we are able to filter the apps from custom string. But, definitely worth to considering for future enhancements  incase of discrepancies in searching with category.

packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
cpathipa: Thought: Since, we are returning filtered list from `appInstances` so, renamed it to `filteredAppInstances`. But, `appsMatchingQuery` makes sense to me, as we are serving the apps that are matching to query.  Hence set it back to `appsMatchingQuery`

packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
cpathipa: @dwiley-akamai Good point,  It's always better to cleanup as we get chance. applied property shorthand for `query`.

cpathipa (commented):
@dwiley-akamai Addressed the below feedback, And ready for re-review. 

- Renamed `filteredAppInstances` back to `appsMatchingQuery`. 
- Applied property shorthand to `query`.

packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
cpathipa: @CodyFinn-Akamai 

- With the updated changes we no longer using `.includes` instead we are using `.indexof` to scan the each query word thru the custom build string. We could use`.match`, But, we need additional checks to ensure that results matches all query words. 
- I like the idea of splitting the query string. But, we don't need array function from loadash as we are using string methods for scanning.

packages/manager/src/features/OneClickApps/FakeSpec.ts
```diff
@@ -287,6 +324,8 @@ export const oneClickApps: OCA[] = [
   },
   {
     name: 'CS:GO',
+    alt_name: 'FPS game',
+    alt_description: 'Classic first person shooter game.',
```
CodyFinn-Akamai, line 328: This could have the full name of `Counter Strike: Global Offensive` in the data somewhere to help people find it

packages/manager/src/features/OneClickApps/FakeSpec.ts
```diff
@@ -1609,6 +1801,8 @@ export const oneClickApps: OCA[] = [
   },
   {
     name: 'TF2',
+    alt_name: 'FPS game',
+    alt_description: 'Classic first person shooter game.',
```
CodyFinn-Akamai, line 1805: This could have the full name of `Team Fortress 2` somewhere in the data to help people find it.

dwiley-akamai (approved):

packages/manager/src/features/OneClickApps/FakeSpec.ts
cpathipa: @CodyFinn-Akamai I agree, We can expect that these alternative names,  alternative description keywords can change in future. For initial iteration marketplace team wants to go with pretty minimal.

packages/manager/src/features/OneClickApps/FakeSpec.ts
cpathipa: @CodyFinn-Akamai I agree, We can expect that these alternative names,  alternative description keywords can be expanded in future. For initial iteration marketplace team wants to go with pretty minimal.

CodyFinn-Akamai:
> I like the thought process of decoupling the create function outside of component, where it avoids performing unnecessary operations when component re-renders. However, in this case logic of building string is wrapped within the event handler onSearch on component re-render it only registers the handler, but doesn't execute it.

It doesn't run on re-render but it does run on every search query. Putting some code into `FakeSpec.ts` something like

```
export const searchableAppData = oneClickApps.map(({name, alt_name, alt_description, categories}) => {
  return {
    name: name.trim(),
    dataString: `${name} ${alt_name} ${categories.join(' ')} ${alt_description}`.toLocaleLowerCase(),
  };
});
```
This way it should only be calculated once.

packages/manager/src/features/linodes/LinodesCreate/TabbedContent/FromAppsContent.tsx
CodyFinn-Akamai: > With the updated changes we no longer using .includes instead we are using .indexof to scan the each query word thru the custom build string. We could use.match, But, we need additional checks to ensure that results matches all query words.

Not sure why we need to use `.indexOf`. We are looping through each query word and comparing it to the apps searchable data string. If we get any hits it should be put into the array. We don't need to know where in the string the query word matches the data string. Now that we split the query words into an array and check against each of them `.includes` should be fine and then we don't have to do a comparison against a number.

PR #8659, opened 2022-12-13
M3-6013: Resolve `loader-utils` Dependabot

bnussman-akamai:
## Description 📝

- Updates packages that use `loader-utils` **but** also adds a yarn resolution because many packages we depend on still use a very old version of it
  - See https://github.com/storybookjs/storybook/pull/19891#issuecomment-1349582334 for example
- Fixes some unused imports in some Storybook stories
- Pins `@types/react` to `17.x.x` using Yarn resolution so we can regenerate our yarn.lock without yarn pulling `@types/react@18`
  - https://github.com/facebook/react/issues/24304
-  Forces `tsup` to use an older version because newer versions use rollup 3 which do not allow you to use our version node 14

## `yarn why loader-utils`
These are the packages blocking us from **not** using a yarn resolution 
- `@storybook/builder-webpack4#loader-utils@1.4.2`
- `@storybook/manager-webpack4#loader-utils@1.4.2`
- `babel-loader#loader-utils@1.4.2`
- `webpack#loader-utils@1.4.2`
- `css-loader#loader-utils@1.4.2`

## How to test 🧪

- Test Storybook
- Test Cloud Manager
- Make sure preview link builds and works
  - This helps us prove that a production build will work


packages/manager/src/components/EntityTable/EntityTable.stories.mdx
```diff
@@ -38,47 +35,6 @@ export const getNodeBalancers = () => {
   return nodeBalancerFactory.buildList(10).map((nb) => ({ ...nb, configs }));
 };
 
-# Entity Detail
-
-## Volumes
```
bnussman-akamai, line 43: Volumes and Domains don't use `<EntityTable />` anymore so this story was broken. Removing for now.

packages/manager/package.json
```diff
@@ -232,7 +232,6 @@
     "reselect-tools": "^0.0.7",
     "serve": "^14.0.1",
     "source-map-loader": "^4.0.1",
-    "storybook-react-router": "~1.0.8",
```
bnussman-akamai, line 235: This package is not used

cpathipa (approved):
@bnussman-akamai Overall LGTM! Thank you for doing cleanup work and lifting unused imports.

mjac0bs (approved):
Didn't see any `yarn` errors and clicked through Manager and Storybook without running into issues. And preview link looks good. 🚢

PR #8660, opened 2022-12-14
M3-5990: Remove nvme from volumes landing page

carrillo-erik:
## Description 📝

After Fremont was upgraded from `HDD` to `NVMe`, storage across all regions is now `NVMe` 
This PR removes the UI `NVMe` chip component from the `Volumes` landing page.

## Preview 📷

image00

## How to test 🧪
- Go to the Volumes landing page.
- Visually inspect the UI to verify the UI chip component has been removed.

## Please note ⚠️
-  The `hardware_type` property is now deprecated from `api-v4` after Fremont was upgraded from `HDD` to `NVMe`.  Storage across all regions will be `NVMe`.


packages/api-v4/src/volumes/types.ts
```diff
@@ -10,7 +10,6 @@ export interface Volume {
   updated: string;
   filesystem_path: string;
   tags: string[];
-  hardware_type: VolumeHardwareType;
```
bnussman-akamai, line 13: Do we know if the API is going to stop returning this? Is not, we should probably keep `hardware_type` in the `@linode/api-v4` just so the package is 1:1 with the Linode API

carrillo-erik: Thanks for pointing that out, I do not have an exact answer for that. I made the decision in part because the wording on the JIRA ticket states: 

> When Fremont is upgraded from HDD to NVMe, storage across all regions will be NVMe. Remove NVMe chips from Volumes page because it is no longer needed.

And the only references to this type in the codebase were in the Volumes landing page. I can revert that change and it won't break the other changes I have made for this ticket.

mjac0bs: Might be worth reaching out to the API folks about in our shared channel we wanted to delete it? I didn't see any related tickets attached to this one on the API side, so maybe it's not in their plans or on their radar. 

`hardware_type` is still in the [API docs](https://www.linode.com/docs/api/volumes/#volumes-list__response-samples) so Bank's comment about keeping `hardware_type` in `@linode/api-v4` makes sense to me too.

mjac0bs (approved):
Tested this by creating a test volume in the Newark, NJ region on prod and on this branch. Confirmed that the vNME chip displays on the prod volume only and not this branch. (Change also visible via mocks.) Unit and E2E tests passed when I ran them with the exception of a cypress timeout on `create-volume.smoke` which was probably unrelated.

I'm going ahead and approving with my additional comment on Bank's note.

packages/api-v4/src/volumes/types.ts
carrillo-erik: > Might be worth reaching out to the API folks about in our shared channel if we wanted to delete it? I didn't see any related tickets attached to this one on the API side, so maybe it's not in their plans or on their radar.
> 
> `hardware_type` is still in the [API docs](https://www.linode.com/docs/api/volumes/#volumes-list__response-samples) so Bank's comment about keeping `hardware_type` in `@linode/api-v4` makes sense to me too.

Should I reach out to the api-v4 channel or the api-front-end channel?

dwiley-akamai: Agreed, we should not remove this unless it is being removed from the API also.

I'd say to reach out in the main channel (api-v4) for visibility reasons. @carrillo-erik

packages/manager/src/features/Volumes/VolumeTableRow.tsx
```diff
@@ -75,52 +72,20 @@ export const VolumeTableRow: React.FC<CombinedProps> = (props) => {
...
-  const nvmeUpgradeScheduledByUserInProgress = events.some(
-    (event) =>
-      event.action === 'volume_migrate' &&
-      event.entity?.id === id &&
-      event.status === 'started'
-  );
-
-  // const recentEvent = events.find((event) => event.entity?.id === id);
-
-  // Use this to show a progress bar
```
dwiley-akamai, line 120: We can consider leaving this bit of commented out code in; Banks added it in a recent PR as a starting point for when we get the progress bars working again in this table.

dwiley-akamai (commented):
The table looks good (although RIP to the NVMe chip, it was very aesthetic 🤌), let's just get confirmation on the `hardware_type` field from the API and address it accordingly in this PR.

packages/api-v4/src/volumes/types.ts
carrillo-erik: I posted my question on the api-v4 Slack channel and this was the response I received:

> So long as that framework won't trip over seeing a key back from the API that it doesn't care about, then I'm good. We don't remove things from the API itself except under very specific circumstances, but we can discuss marking this as deprecated if it's no longer in use by anyone.

I'm now open to the team's wishes for removing or leaving it as is.

bnussman-akamai: In my option we should keep `hardware_type` in api-v4. That way @linode/api-v4 is accurate to what the API actually returns. @linode/api-v4 is a public NPM package so there may be people outside of Linode/Akamai that want to use the package. They would probably expect to see all types that the API returns.

dwiley-akamai: I am also in favor of keeping it in. There are other instances in the package where we have properties that aren't used at the moment in Cloud, so we would not be breaking precedent here by keeping it in.

carrillo-erik: I'll go ahead and reintroduce `hardware_type` back into the api-v4 package for the reasons @bnussman-akamai has mentioned above, totally makes sense. Secondly, I'll revert the changes that removed the progress bar work that is in progress. I'll also make mention of the deprecation of the `hardware_type` property in the PR description.

cpathipa (approved):
Changes LGTM! functionality is working as expected. I see visual diff of lifting `nvme` labels in volumes landing page. Only thing I have is, below `cypress e2e` tests are failing on this branch. Please consider them before merging. But, you have my approval though. 
Also, I'm inline keeping `hardware_type` field in [packages/api-v4/src/volumes/types.ts](https://github.com/linode/manager/pull/8660/files#diff-88a1165a72c139b682fee9355ed78ca4bb616b53a2f76530573f3564a4974378) If its still part of API response. 

- kubernetes/smoke-lke-create.spec.ts
- objectStorage/object-storage.smoke.spec.ts
- objectStorage/access-keys.smoke.spec.ts

image
image
image

PR #8661, opened 2022-12-14
M3-5183: React Query for Kubernetes - Part 1

bnussman-akamai:
## Description 📝

- This PR focuses getting React Query to power the Kubernetes landing page table and associated creation and deletion actions
  - > **Warning**: All other Kubernetes flows and actions will be addressed in part 2 of this PR

## Preview 📷

Screen Shot 2022-12-14 at 7 52 51 PM

## How to test 🧪

- Testing Kubernetes Landing
  - Test deleting a Kubernetes Cluster and verify the table updates
  - Test creating a Kubernetes Cluster and verify the table updates


packages/manager/src/features/Kubernetes/KubernetesClusterDetail/DeleteKubernetesClusterDialog.test.tsx
```diff
@@ -0,0 +1,127 @@
...
+  //     <DeleteKubernetesClusterDialog {...props} />
+  //   );
+  //   // Fill in the "are you sure" text
+  //   const input = getByTestId('textfield-input');
+  //   fireEvent.change(input, { target: { value: 'this-cluster' } });
+  //   const button = getByTestId('dialog-confirm');
+  //   // Now this should work.
+  //   fireEvent.click(button);
+  //   expect(props.onDelete).toHaveBeenCalledTimes(1);
+  // });
```
jdamore-linode, line 126: Let me know if you need a hand with either of these!

jdamore-linode (approved):
Looks really good, nice work Banks!

packages/manager/src/features/Kubernetes/UpgradeVersionModal.tsx
```diff
@@ -5,30 +5,31 @@ import ActionsPanel from 'src/components/ActionsPanel';
...
-    nextVersion,
-    isOpen,
-    onClose,
-  } = props;
+  const { clusterID, clusterLabel, currentVersion, isOpen, onClose } = props;
+
+  const { data: versions } = useKubernetesVersionQuery();
   const { enqueueSnackbar } = useSnackbar();
 
   const { updateKubernetesCluster } = useKubernetesClusters();
```
dwiley-akamai, line 29: when this is used later in this file, we'll probably want to invalidate the query in the `.then()` block

dwiley-akamai (commented):
I still have quite a bit of code to comb through, but overall it's looking good so far.

This came up on my Images PR too, but I do think we ought to come to a team consensus for how we should generally handle calculated/extended data that needs to be displayed for entity rows or on detail pages. I definitely see your argument re: keeping objects as plain as possible for readability and maintainability but we have to decide on how to weigh that vs. extra calculations & renderings. The decision won't be a one-size-fits-all-situations one but just a guideline we follow unless there's a good reason to go in the other direction in a specific situation

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/DeleteKubernetesClusterDialog.test.tsx
bnussman-akamai: Because the delete mutation was moved into inside the delete dialog, I am going to change this to check the button disabled state.

packages/manager/src/features/Kubernetes/UpgradeVersionModal.tsx
bnussman-akamai: Great call. Got this added in 5b4489f

dwiley-akamai (approved):

PR #8662, opened 2022-12-14
Fix Linode landing page tests for non-localhost environments

jdamore-linode:
## Description 📝

**What does this PR do?**
This fixes an issue with a test in `smoke-linode-landing-table.spec.ts` which causes it to fail in cases where the test environment is not `localhost:3000`. Also fixes a TypeScript error and does a little bit of refactoring of the test file.

## How to test 🧪

You can verify that the test continues to pass locally by running `yarn dev` and then:

```bash
yarn cy:run -s "cypress/e2e/linodes/smoke-linode-landing-table.spec.ts"
```

If you want to confirm the fix for non-localhost environments, an easy (but slow) way to do it is by running the tests via Docker:

(You can speed this up a lot by temporarily setting `specPattern` in `cypress.config.ts` to `'cypress/e2e/linodes/smoke-linode-landing-table.spec.ts'` before running the commands below)

```bash
yarn build
docker compose up -d e2e
docker logs -f manager-e2e-1
```

When you're done, you can clean things up with:
```bash
docker compose down e2e
```

### DIFF
--- a/packages/manager/cypress/e2e/linodes/smoke-linode-landing-table.spec.ts
+++ b/packages/manager/cypress/e2e/linodes/smoke-linode-landing-table.spec.ts
@@ -1,4 +1,5 @@
 /* eslint-disable sonarjs/no-duplicate-string */
+import { Linode } from '@linode/api-v4/types';
 import { createLinode } from '../../support/api/linodes';
 import {
   containsVisible,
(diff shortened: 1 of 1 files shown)


cpathipa (approved):
@jdamore-linode LGTM! Test passed for me.

bnussman-akamai (approved):

PR #8663, opened 2022-12-15
M3-5170: Filter blocklisted events from Notification Menu

mjac0bs:
## Description 📝

**What does this PR do?**
Removes blocklisted events from the NotificationMenu popover, to reach parity with the list of events on the the Events Landing page. Blocklisted events include events like `linode_migrate_datacenter_create`.

## Preview 📷

**Prod:**
prod-for-m3-5170

**This branch:**
branch-for-m3-5170

## How to test 🧪

1. Create a new nanode or use an existing one.
2. Kick off a migration for the linode.
3. Navigate to the Events Landing page at localhost:3000/events. 
4. Click on the notification icon in the TopMenu and observe that the list of events in the popover notification menu matches the list of events on the Events Landing page.
5. Compare this to prod and observe that in prod, the event for initiating a migration is visible in the popover's event list but not visible on the Events Landing page.

```
yarn cy:run -s "cypress/e2e/notificationsAndEvents/notifications.spec.ts"
```
```
yarn cy:run -s "cypress/e2e/notificationsAndEvents/events.spec.ts"
```

### DIFF
--- a/packages/manager/src/components/NotificationMenu/NotificationMenu.tsx
+++ b/packages/manager/src/components/NotificationMenu/NotificationMenu.tsx
@@ -4,14 +4,12 @@ import Paper from 'src/components/core/Paper';
 import { makeStyles, Theme } from 'src/components/core/styles';
 import Events from 'src/features/NotificationCenter/Events';
 import { NotificationData } from 'src/features/NotificationCenter/NotificationData/useNotificationData';
-import { NotificationItem } from 'src/features/NotificationCenter/NotificationSection';
 import Notifications from 'src/features/NotificationCenter/Notifications';
 import useDismissibleNotifications from 'src/hooks/useDismissibleNotifications';
 import useNotifications from 'src/hooks/useNotifications';
 import usePrevious from 'src/hooks/usePrevious';
 import { markAllSeen } from 'src/store/events/event.request';
 import { ThunkDispatch } from 'src/store/types';
-import { removeBlocklistedEvents } from 'src/utilities/eventUtils';
 
 const useStyles = makeStyles((theme: Theme) => ({
   root: {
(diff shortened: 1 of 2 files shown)


jdamore-linode (approved):
Nice, looks good! Followed the steps and confirmed that the _Migration for Linode ... has been initiated by ..._ event which is present in Prod is not shown in the notifications menu for this PR.

packages/manager/src/components/NotificationMenu/NotificationMenu.tsx
```diff
@@ -24,6 +26,35 @@ export interface Props {
...
+  const filteredEvents = removeBlocklistedEvents(events);
+
+  // Include all notifications that are non-blocklisted events.
+  return eventNotifications.filter((notification) => {
+    return (
+      (isNotificationEventObj(notification.body) &&
+        filteredEvents.includes(notification.body.props.event)) ||
+      !isNotificationEventObj(notification.body)
+    );
+  });
```
cpathipa, line 55: Nice to have:
We can remove explicit `return` from `filter`
```suggestion
  return eventNotifications.filter((notification) => 
      (isNotificationEventObj(notification.body) &&
        filteredEvents.includes(notification.body.props.event)) ||
      !isNotificationEventObj(notification.body)
    );
```

cpathipa (commented):
Nice work @mjac0bs, Left couple comments(nice to have things), other than that overall LGTM!

dwiley-akamai (commented):
I think we can achieve this same goal by using `removeBlocklistedEvents()` at a higher level, which should eliminate the need for the custom logic in this file that drills down checking the object structure, etc.

In `useEventNotifications.tsx`, it looks like a one-line change could do the trick.

https://github.com/linode/manager/blob/d32c3d705bf2667ec24f0d288d1da037e358c4a1/packages/manager/src/features/NotificationCenter/NotificationData/useEventNotifications.tsx#L22

```suggestion
const events = removeBlocklistedEvents(givenEvents ?? useEvents().events);
```

From what I can tell, there shouldn't be any adverse impacts elsewhere based on how and where `useEventNotifications()` is used. The badge number also appears to still be accurate.

mjac0bs:
> I think we can achieve this same goal by using `removeBlocklistedEvents()` at a higher level, which should eliminate the need for the custom logic in this file that drills down checking the object structure, etc.
> 
> In `useEventNotifications.tsx`, it looks like a one-line change could do the trick.

@dwiley-akamai - Ha. Thanks for calling this out. I tested that change with a migration, a resize, and a non-blocklisted event and also observed that the blocklisted events seem to be correctly filtered out of the Notification Menu + the badge number is accurate.

@cpathipa - Thanks for the code clean up feedback -- would have implemented if we didn't have a cleaner one-liner here. Can you confirm that functionality works as expected for you too with the latest commit?

dwiley-akamai (approved):

cpathipa (approved):
Nice work!  Able to see the diff of filtering migration `initiated` notification in Prod and current branch.

PR #8665, opened 2022-12-16
Bugfix/access controls ips fix

ericsund:
## Description 📝

**What does this PR do?**

This PR it ensures `/32` is appended to the end of every IP address in the list.

## Preview 📷

With `/32` filtering enabled:
Screen Shot 2022-12-15 at 3 42 04 PM

Without `/32` filtering enabled (adding one more IP after reverting the changes):
Screen Shot 2022-12-15 at 3 44 14 PM

## How to test 🧪

* Navigate to your `/databases` endpoint and create a new database (I used MySQL but it should work for any DB).
* Click the blue **Manage Access Controls** button under the **Summary** tab.
* Add an IP, like 127.0.0.0 and hit Enter right away without clicking **Update Access Controls**
* When the right sidebar collapses, ensure that `/32` is prepended to the newly added IP in the **Summary** tab.

```
yarn test AccessControls.test
```

### DIFF
--- a/packages/manager/src/features/Databases/DatabaseDetail/AddAccessControlDrawer.tsx
+++ b/packages/manager/src/features/Databases/DatabaseDetail/AddAccessControlDrawer.tsx
@@ -81,15 +81,6 @@ const AddAccessControlDrawer: React.FC<CombinedProps> = (props) => {
 
       return [...acc, ipString];
     }, []);
-      .map(extendedIPToString)
-      .filter((ip) => ip !== '')
-      .map((ip) => {
-        let curr_ip = ip;
-        if (curr_ip.indexOf('/32') === -1) {
-          curr_ip = `${curr_ip}/32`;
-        }
-        return curr_ip;
-      });
 
     updateDatabase({ allow_list: [...allowListRetracted] })
       .then(() => {
--- a/packages/manager/src/features/Databases/DatabaseDetail/AddAccessControlDrawer.tsx
+++ b/packages/manager/src/features/Databases/DatabaseDetail/AddAccessControlDrawer.tsx
@@ -68,9 +68,19 @@ const AddAccessControlDrawer: React.FC<CombinedProps> = (props) => {
     }
   ) => {
     // Get the IP address strings out of the objects and filter empty strings out.
-    const allowListRetracted = _allowList
-      .map(extendedIPToString)
-      .filter((ip) => ip !== '');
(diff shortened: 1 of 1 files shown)


dwiley-akamai (commented):
Thanks for opening this PR! 🎉 

Not sure if it's just a matter of the PR description being outdated now, but the duplicates are prevented by logic on the back-end and Cloud displays the error to users:

Screen Shot 2022-12-16 at 10 44 13 AM

Screen Shot 2022-12-16 at 10 28 32 AM

Were you somehow still ending up with duplicates getting through?

ericsund:
Thanks @dwiley-akamai!  I realized duplicates were already being handled, so I removed code that casts it to a `Set` in my last commit -- just didn't take it out in the PR description.  The issue of `/32` missing is still present, though.

packages/manager/src/features/Databases/DatabaseDetail/AddAccessControlDrawer.tsx
```diff
@@ -68,9 +68,17 @@ const AddAccessControlDrawer: React.FC<CombinedProps> = (props) => {
     }
   ) => {
     // Get the IP address strings out of the objects and filter empty strings out.
+    // Ensure we append /32 to all IPs if not already present.
     const allowListRetracted = _allowList
```
dwiley-akamai, line 72: Since the chained array methods are starting to pile up now (3), we can consider using `.reduce()` instead. We'll avoid looping over the array 3 separate times, while still getting the IP strings, filtering out the empty strings, and appending `/32` if needed.

```suggestion
    const allowListRetracted = _allowList.reduce((acc, currentIP) => {
      let ipString = extendedIPToString(currentIP);
      if (ipString === '') {
        return acc;
      }

      if (ipString.indexOf('/32') === -1) {
        ipString += '/32';
      }

      return [...acc, ipString];
    }, []);
```

dwiley-akamai (approved):

ericsund:
Can we get some feedback from @bnussman as well?

bnussman-akamai (changes requested):
I'm not sure adding `/32` to every IP is exactly the behavior we want. I tried adding a new IP address to an existing allow list and it caused `/32` to be added to a `/16` I already had setup. 

https://user-images.githubusercontent.com/115251059/208482721-2c638369-a6c2-40dc-b472-a6e915cc2f68.mov

bnussman-akamai (approved):

cpathipa (approved):

PR #8673, opened 2022-12-21
M3-5182: RQ-ify Images, Pt. 2

dwiley-akamai:
## Description 📝
Part 2 of the effort to transition Images from Redux to React Query.

At a high level, this PR:
- Allows for the removal of the `useImages` hook, `withImages.container.ts` (used to provide Images data from Redux to class components, and has been replaced by `images.container.ts` to provide Images data from RQ to class components), and Images from `useEntities` hook
- Adds a `useAllImagesQuery()` which is what we want in a lot of these cases, not the `useImagesQuery`
- Makes use of the `useAllImagesQuery()` across many files

## How to test 🧪
This PR touches some core functionality, namely:

1. `ImageSelect` fields (Linode Create flow, Linode Rebuild flow, StackScript Create flow)
2. Search (Images should show up for both normal and large accounts)
3. Migrate Linode modal

When comparing this branch to prod, it's important to ensure that anywhere that images are listed in the UI, you see the same options or results.

## Notes & Up Next
Regarding the [sorting issue that surfaced in Pt. 1](https://github.com/linode/manager/pull/8645#issuecomment-1359661600), there's an API ticket being worked on currently that will allow for sorting by Status. That will probably end up being rolled out before the RQ-ifying Images work is released to prod.

Part 3 will consist of addressing logic related to the Image upload functionality and final cleanup.

## To-Do:
- [x] Audit instances of `useImagesQuery` and `useAllImagesQuery` to ensure correct query is being used in each instance
- [ ] Cleanup
- [x] Additional testing


packages/manager/src/features/TopMenu/SearchBar/SearchBar.tsx
```diff
@@ -303,7 +308,7 @@ export const SearchBar: React.FC<CombinedProps> = (props) => {
 export default compose<CombinedProps, {}>(
   withTypes,
   withRouter,
-  withImages(),
+  withImages,
```
bnussman-akamai, line 312: Is `withImages` needed here?

dwiley-akamai: Nope, I went ahead and removed that

packages/manager/src/features/Images/utils.ts
```diff
@@ -0,0 +1,20 @@
+import { getGrantData, getProfileData } from 'src/queries/profile';
+
+export const useImageAndLinodeGrantCheck = () => {
+  const profile = getProfileData();
+  const grants = getGrantData();
```
bnussman-akamai, line 5: Should we be using the hooks here instead? I remember keeping `getProfileData` and `getGrantData` only so we could get the RQ data in class components.

packages/manager/src/features/linodes/LinodesCreate/LinodeCreateContainer.tsx
```diff
@@ -260,10 +259,12 @@ class LinodeCreateContainer extends React.PureComponent<CombinedProps, State> {
...
-      !this.props.imagesLoading
+      (!this.props.imagesLastUpdated && !this.props.imagesLoading) ||
+      Date.now() - this.props.imagesLastUpdated > REFRESH_INTERVAL
     ) {
-      this.props.requestImages();
+      queryClient.fetchQuery(imagesListQueryKey, () => getImages(), {
+        staleTime: REFRESH_INTERVAL,
+      });
     }
   }
```
bnussman-akamai, line 269: I think we should remove this entire chunk of logic. The `LinodeCreateContainer` should have no business dealing with the freshness of the Images data. 

I think for now, we just remove this chunk of code, and fine tune staleTime of the images query down the road

bnussman-akamai (commented):
I have a few small items and I see that the unit tests are failing. Do you intend to fix those in the next PR?

packages/manager/src/features/linodes/LinodesCreate/LinodeCreateContainer.tsx
dwiley-akamai: That should be fine, I removed this chunk.

dwiley-akamai:
> I have a few small items and I see that the unit tests are failing. Do you intend to fix those in the next PR?

The test that failed before seemed to be an unrelated stray flake. I merged `develop` into the feature branch and rebased so everything should be fine now, but I'll keep an eye on it.

packages/manager/src/features/linodes/LinodeEntityDetail.tsx
```diff
@@ -92,15 +93,17 @@ const LinodeEntityDetail: React.FC<CombinedProps> = (props) => {
     recentEvent,
   } = props;
 
-  useReduxLoad(['images', 'types']);
-  const { images } = useImages();
+  const { data: images } = useAllImagesQuery({}, {});
```
bnussman-akamai, line 96: Why not use `useImageQuery` here? Is it to prevent another fetch?

bnussman-akamai (changes requested):
When I load up Cloud Manager and go to `http://localhost:3000/linodes`, I see 4 listeners to the images query. I think we really should work to make this number 0. Unless something on the screen renders an image's information. We should try to not query until we really need to. Here's what I think we should do:

1. Remove useAllImagesQuery() from `packages/manager/src/features/linodes/LinodesLanding/index.tsx` completely
2. Remove any `withImages` usage from `packages/manager/src/features/linodes/LinodesLanding/LinodesLanding.tsx`
3. The `packages/manager/src/features/linodes/MigrateLanding/MigrateLinode.tsx` dialog should use the `open` prop to conditionally enable the  `useAllImagesQuery` so that data is not fetched until the dialog is open
4. `useAPISearch` should not query images unless someone is searching

Resolving those 4 items will allow us to load `http://localhost:3000/linodes` without making any requests to the images endpoint. This is definitely an over-optimization in some ways but will be very beneficial at scale

packages/manager/src/features/linodes/LinodeEntityDetail.tsx
dwiley-akamai: Yes, that was the intention with this

bnussman-akamai (approved):
Looks good!

I think the "use all" approach makes sense for images considering how they are used and what they are needed for. For further optimization we could start replacing  client side filtering with api filters. For example, replacing the `filteredImages` in `ImageSelect.tsx` with a dynamic API filter.

packages/manager/src/features/Search/SearchLanding.tsx
```diff
@@ -92,6 +94,7 @@ export const SearchLanding: React.FC<CombinedProps> = (props) => {
 
   const classes = useStyles();
   const { _isLargeAccount } = useAccountManagement();
+  // const _isLargeAccount = true;
```
hana-linode, line 97: Can we remove this comment?

hana-linode (approved):
Functionality looks good to me!

Not sure if this is related to the sorting issue but the sorting for the search results seem to be in a different order than prod

Prod:

https://user-images.githubusercontent.com/115299789/214686983-98d2e276-cc55-495d-860b-f0d9d8ac808e.mov

This branch:

https://user-images.githubusercontent.com/115299789/214687037-21d774aa-d72f-4bc3-9af4-7173a92af6fa.mov

dwiley-akamai:
Thanks for the comments here, I'll circle back around and look into these items in Pt. 3.

Issue #8675, opened 2022-12-23
Unable to use with nodejs (commonjs) or typescript


Jishun:
Trying to reference the javascript sdk 
started with `npm i @linode/api-v4`
and try the basic example
```js
import { baseRequest } from '@linode/api-v4'

baseRequest.interceptors.request.use(config => {
  const myToken = 'token';

  return {
    ...config,
    headers: {
      ...config.headers,
      Authorization: `Bearer ${myToken}`
    }
  };
});
```
received below error 
```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/djs/dev/temp/node_modules/@linode/api-v4/lib/index.js from /Users/djs/dev/temp/linode.ts not supported.
index.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /Users/djs/dev/temp/node_modules/@linode/api-v4/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).

    at require.extensions.<computed> [as .js] (/usr/local/lib/node_modules/ts-node/dist/index.js:851:20)

```

the error is kinda clear, it believes itself a ES module so rejected request from commonjs, then I set the project to `module` by adding `"type": "module"` in `package.json`
however it turns out this package references another commonjs module that produces another error: 
```
file:///Users/djs/dev/temp/node_modules/@linode/validation/lib/esm/index.js:219
import { parse as parseIP, parseCIDR } from "ipaddr.js";
                           ^^^^^^^^^
SyntaxError: Named export 'parseCIDR' not found. The requested module 'ipaddr.js' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'ipaddr.js';
const { parse: parseIP, parseCIDR } = pkg;
```
The same behaviour with `js` or `ts`. when using `ts` , `tsc` does compile the code to proper js file. but later executing with `node` gives the above error.  the same error when executing the ts file with `ts-node`

Can't find out what I'm missing. 
is there a sample nodejs or TypeScript project that I can take and have a quick try ?

bnussman:
[...]

PR #8677, opened 2022-12-27
Validate api-v4 package before publishing to NPM

jcallahan-akamai:
## Description 📝

This adds a step to the CI workflow called **validate-sdk** to verify our api-v4 package works before publishing it to NPM.

The new workflow step occurs after the validation and api-v4 packages are built. It does the following:

1. Uses `npm pack` to create a tarball (like the one that will be published to NPM) 
2. Uses `npm init` to create a new Node module 
3. Uses `npm install` to install api-v4 via local tarball
4. Runs new test script in scripts/validatePackages

Steps 2-4 are repeated for CommonJS. 

**Note:** I updated the Node version so to get `npm pack` to work correctly. 

## Preview 📷
Screen Shot 2022-12-27 at 5 54 17 PM

## How to test 🧪

You can observe the workflow running [here](https://github.com/jcallahan-akamai/manager/actions/runs/3790380579). 

### Note
This test doesn’t yet pass since the api-v4 package is currently broken for Node. It will need to be fixed first and then these changes need to be validated again.

Specifically, [this is the error ](https://github.com/jcallahan-akamai/manager/actions/runs/3790380579/jobs/6444983641#step:9:6) reported in https://github.com/linode/manager/issues/8675.


.github/workflows/ci.yml
```diff
@@ -3,7 +3,8 @@ name: ci
 on:
   push:
     branches:
-      - master
+      - feature/validate-api-v4
+      # - master
```
jcallahan-akamai, line 7: Will change this before merge.

.github/workflows/ci.yml
```diff
@@ -99,6 +100,39 @@ jobs:
...
+      - uses: actions/download-artifact@v2
+        with:
+          name: packages-api-v4-lib
+          path: packages/api-v4          
+
+      # Create an api-v4 tarball
+      - run: cd packages/api-v4 && npm pack --pack-destination ../../
+
+      # Test @linode/api-v4 as an ES Module
+      - run: mkdir test-sdk-esm && cd test-sdk-esm && npm init es6 -y && npm install ../$(ls ../ | grep "linode-api-v4-.*\.tgz")
```
jcallahan-akamai, line 127: One way to improve this would be to grab the exact name of the tarball from the output to `npm pack` instead of doing `$(ls | grep "inode-api-v4-.*\.tgz")`, but I was having trouble getting the filename. Ideas welcome.

bnussman-akamai (approved):
This all makes sense to me! I'm mostly still trying to better understand https://github.com/linode/manager/issues/8675

dwiley-akamai (commented):
With a more recent version of `develop` merged in fixing the broken package issue, this gets further than before but fails on the `Run cp scripts/validatePackages/sdk-commonjs.cjs test-sdk-cjs/` step:

Screen Shot 2023-01-20 at 3 42 41 PM

The name of the file in this PR is `scripts/validatePackages/sdk-commonjs.js` but in the command we have `scripts/validatePackages/sdk-commonjs.cjs` -- should the former be renamed to have the `.cjs` filename extension?

bnussman-akamai:
I don't quite understand why `test-manager` is failing now

jaalah-akamai:
There's been quite a few changes since this branch was created - let's try updating it and see where we stand then.

jaalah-akamai (approved):
🚢

PR #8681, opened 2022-12-30
Feature: show login history on Account page

jcallahan-akamai:
## Description 📝

**What does this PR do?**

This consumes the /account/logins API endpoint and displays the data on a new "Logins" tab/page in the Account section.

I used existing patterns for RQ, API paginated tables, etc.

This still needs to be validated with UX.

## Preview 📷

Screen Shot 2022-12-30 at 3 05 35 PM

## How to test 🧪

- Visit localhost:3000/account/logins. 
- Check error states, loading state, responsiveness, etc. 
- Check sortable table columns
- Try viewing it with a restricted user (you'll receive an "Unauthorized" error from the API)
- Populate it with a variety of data by creating a restricted user on your account and logging in with it

No unit tests yet.

### DIFF
--- a/packages/manager/src/features/Account/AccountLanding.tsx
+++ b/packages/manager/src/features/Account/AccountLanding.tsx
@@ -14,6 +14,7 @@ import TabLinkList from 'src/components/TabLinkList';
 import { akamaiBillingInvoiceText } from 'src/features/Billing/billingUtils';
 import { useAccount } from 'src/queries/account';
 import { getGrantData, useProfile } from 'src/queries/profile';
+import AccountLogins from './AccountLogins';
 
 const Billing = React.lazy(() => import('src/features/Billing'));
 const EntityTransfersLanding = React.lazy(
@@ -44,6 +45,10 @@ const AccountLanding: React.FC = () => {
     {
       title: 'Users & Grants',
       routeName: '/account/users',
+    },
+    {
+      title: 'Logins',
+      routeName: '/account/logins',
     },
     {
       title: 'Service Transfers',
@@ -134,6 +139,9 @@ const AccountLanding: React.FC = () => {
             </SafeTabPanel>
             <SafeTabPanel index={++idx}>
               <Users isRestrictedUser={profile?.restricted || false} />
+            </SafeTabPanel>
+            <SafeTabPanel index={++idx}>
+              <AccountLogins />
             </SafeTabPanel>
             <SafeTabPanel index={++idx}>
               <EntityTransfersLanding />
--- a/packages/manager/src/features/Account/AccountLanding.tsx
+++ b/packages/manager/src/features/Account/AccountLanding.tsx
@@ -47,8 +47,8 @@ const AccountLanding: React.FC = () => {
       routeName: '/account/users',
     },
     {
-      title: 'Logins',
-      routeName: '/account/logins',
+      title: 'Login History',
+      routeName: '/account/login-history',
     },
     {
       title: 'Service Transfers',
--- a/packages/manager/src/features/Account/AccountLogins.tsx
+++ b/packages/manager/src/features/Account/AccountLogins.tsx
@@ -17,7 +17,6 @@ import { useOrder } from 'src/hooks/useOrder';
 import usePagination from 'src/hooks/usePagination';
 import { useAccountLoginsQuery } from 'src/queries/accountLogins';
 import AccountLoginsTableRow from './AccountLoginsTableRow';
-import Paper from 'src/components/core/Paper';
 
 const preferenceKey = 'account-logins';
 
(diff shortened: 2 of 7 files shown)


packages/manager/src/queries/accountLogins.ts
```diff
@@ -0,0 +1,11 @@
...
+import { APIError, ResourcePage } from '@linode/api-v4/lib/types';
+import { useQuery } from 'react-query';
+
+export const queryKey = 'account-login';
+
+export const useAccountLoginsQuery = (params?: any, filter?: any) =>
+  useQuery<ResourcePage<AccountLogin>, APIError[]>(
+    [`${queryKey}-list`, params, filter],
+    () => getAccountLogins(params, filter)
+  );
```
bnussman-akamai, line 11: Adding the `keepPreviousData` React Query option will make switching between pages a bit more seamless. We've been doing this on all of our other paginated queries. 

## `{ keepPreviousData: true }`

https://user-images.githubusercontent.com/115251059/210390847-c63db8b6-77a6-43a1-be83-85b90fd94e34.mov

## `{ keepPreviousData: false }` (default)

https://user-images.githubusercontent.com/115251059/210390984-4dcb6169-e732-4be5-8fd5-de9147705a2e.mov

```suggestion
    () => getAccountLogins(params, filter),
    { keepPreviousData: true }
  );
```

packages/manager/src/features/Account/AccountLoginsTableRow.tsx
```diff
@@ -0,0 +1,27 @@
+import { AccountLogin } from '@linode/api-v4/lib/account/types';
+import * as React from 'react';
+import Hidden from 'src/components/core/Hidden';
+import TableCell from 'src/components/TableCell';
+import TableRow from 'src/components/TableRow';
+import formatDate from 'src/utilities/formatDate';
+
+const AccountLoginsTableRow: React.FC<AccountLogin> = (props) => {
```
bnussman-akamai, line 8: Consider moving away from `React.FC` on a component like this because `React.FC` implies that the component takes children. 

```suggestion
const AccountLoginsTableRow = (props: AccountLogin) => {
```

packages/manager/src/features/Account/AccountLogins.tsx
```diff
@@ -0,0 +1,142 @@
...
+    width: '12%',
+  },
+  copy: {
+    lineHeight: '20px',
+    marginTop: theme.spacing(),
+    marginBottom: theme.spacing(2),
+  },
+}));
+
+const AccountLogins: React.FC<{}> = () => {
```
bnussman-akamai, line 35: Same `React.FC` feedback here.

```suggestion
const AccountLogins = () => {
```

bnussman-akamai (approved):
Simple but great addition! Please re-request review if we decide on any UI / UX changes

packages/manager/src/features/Account/AccountLanding.tsx
```diff
@@ -45,6 +46,10 @@ const AccountLanding: React.FC = () => {
       title: 'Users & Grants',
       routeName: '/account/users',
     },
+    {
+      title: 'Logins',
```
hana-linode, line 50: Thoughts on renaming the title to `Login History`?

packages/manager/src/features/Account/AccountLogins.tsx
```diff
@@ -0,0 +1,142 @@
...
+      );
+    } else if (error) {
+      return <TableRowError colSpan={5} message={error[0].reason} />;
+    } else if (data?.results == 0) {
+      return <TableRowEmptyState message="No account logins" colSpan={6} />;
+    } else if (data) {
+      return data.data.map((item: AccountLogin) => (
+        <AccountLoginsTableRow key={`${item.id}`} {...item} />
+      ));
+    }
```
hana-linode, line 80: Since we are returning from this function, we can just use regular ifs to unnest this a little

hana-linode (commented):
Re: UX, IIRC we moved away from the redundant header name and unnecessary paper wrapping

I think something like this would look cleaner (also see the `Users & Grants` tab for something similar)

image

packages/manager/src/features/Account/AccountLogins.tsx
```diff
@@ -0,0 +1,142 @@
...
+          }}
+        />
+      );
+    } else if (error) {
+      return <TableRowError colSpan={5} message={error[0].reason} />;
+    } else if (data?.results == 0) {
+      return <TableRowEmptyState message="No account logins" colSpan={6} />;
+    } else if (data) {
+      return data.data.map((item: AccountLogin) => (
+        <AccountLoginsTableRow key={`${item.id}`} {...item} />
```
bnussman-akamai, line 78: Key does not need to be a string here. It can be `string | number`
```suggestion
        <AccountLoginsTableRow key={item.id} {...item} />
```

hana-linode (approved):
LGTM! 🚢 

Non-blocking, but we could add a little bit of left margin to the login text on mobile screens
image

cpathipa (approved):
Nice work! The feedback I have was already covered.

PR #8683, opened 2023-01-03
Gravatar Clean Up

bnussman-akamai:
## Description 📝

- Cleans up Gravatar Logic and Usage
- We were doing way too much heavy lifting when using gravatars
  - We used axios to see if a gravatar existed but this is super redundant. It will just 404 if it does not exist.
- Use the MUI `<Avatar />` to handle fallback to our default icon

## Preview 📷

- Nothing should appear differently. This is a "under the hood" change

## How to test 🧪

- Test Gravatars throughout Cloud Manager
  - View `/account/users`
  - Make sure your gravatar shows in top navigation menu
  - Make sure gravatars show correctly in the events feeds (`/events`)
  - Make sure gravatars show correctly for support tickets and support ticket replies


packages/api-v4/src/account/types.ts
```diff
@@ -4,8 +4,9 @@ export interface User {
   username: string;
   email: string;
   restricted: boolean;
-  gravatarUrl?: string;
```
bnussman-akamai, line 7: We should not make up types in `api-v4`. `gravatarUrl` is not returned from the API

packages/manager/src/components/GravatarByEmail.tsx
```diff
@@ -0,0 +1,37 @@
...
+    borderRadius: '50%',
+  },
+});
+
+interface Props {
+  email: string;
+  className?: string;
+}
+
+export const GravatarByEmail = (props: Props) => {
```
bnussman-akamai, line 22: It's kind of silly to have `GravatarByUsername` and `GravatarByEmail` components but one needs to query and one does not. As dumb as this is, it keeps logic within each component simple.

packages/manager/src/features/Support/types.ts
```diff
@@ -1,9 +0,0 @@
-import { SupportReply, SupportTicket } from '@linode/api-v4/lib/support/types';
-
-export interface ExtendedTicket extends SupportTicket {
-  gravatarUrl: string | undefined;
```
bnussman-akamai, line 4: The gravatar url of a user is just one `md5` hash away. This is not an expensive operation and can be done at the time of rendering the Avatar, no need to extend types.

packages/manager/src/features/Profile/DisplaySettings/GravatarIcon.tsx
```diff
@@ -1,65 +0,0 @@
...
-  } = useAccountUsers(
-    {
-      page: pagination.page,
-      page_size: pagination.pageSize,
-    },
-    true
-  );
-  const classes = useStyles();
-
-  const gravatarURL = users?.data.find((user) => user.username === username)
```
bnussman-akamai, line 37: This logic would have broken if this account had >25 users. `users` is a paginated list of only 25 items

mjac0bs (commented):
Nice cleanup. I tested without a gravatar to see the default behavior (left a related comment related to the user profile display tab) and then created a gravatar to also confirm top nav, profile, events page, and support tickets show gravatar as expected.

packages/manager/src/components/GravatarByEmail.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const { email, className } = props;
+  const classes = useStyles();
+
+  const url = getGravatarUrl(email);
+
+  return (
+    <Avatar
+      className={classNames(classes.avatar, className)}
+      src={url}
```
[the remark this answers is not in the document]
bnussman-akamai: I'm not seeing this 👁

packages/manager/src/features/Users/UsersLanding.tsx
```diff
@@ -160,17 +157,7 @@ const UsersLanding: React.FC<Props> = (props) => {
...
-              ) : user.gravatarUrl === 'not found' ? (
-                <UserIcon className={classes.avatar} />
-              ) : (
-                <img
-                  alt={`user ${user.username}\u{2019}s avatar`}
-                  src={user.gravatarUrl}
-                  className={classes.avatar}
-                />
-              )}
+              <GravatarByEmail email={user.email} />
```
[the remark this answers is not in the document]
bnussman-akamai: Good catch, thank you!

packages/manager/src/components/GravatarByEmail.tsx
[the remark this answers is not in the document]
mjac0bs: @bnussman-akamai This was weird; you can disregard this. I think I had something going on with my local instance -- I another had other unexpected error that popped up too, and after yarn clean and rebuilding, I can no longer replicate the error. Since you didn't see it and Dajahi didn't mention it, I think this was all on my end.

mjac0bs (approved):

dwiley-akamai (approved):

bnussman-akamai:
@dwiley-akamai  My bad. Updated that in f359d76.

PR #8686, opened 2023-01-04
`develop` ⬅️ `feature/react-query-kubernetes`

bnussman-akamai:
## Description 📝

- Final PR to merge `develop` ⬅️ `feature/react-query-kubernetes`
- Use React Query for all Kubernetes API Calls
- API Paginates and Orders Kubernetes Landing

## Preview 📷

Screen Shot 2023-01-04 at 1 58 44 PM

## How to test 🧪

- Test all Linode Kubernetes related features in Cloud Manager
  - All Kubernetes features should work the same (or better) as compared to production


jdamore-linode:
It seems like there may be an issue with sorting by region on the landing page. If you navigate to the landing page and try to sort by region, it doesn't change the order at all. But what's worse is that it seems like Cloud remembers that as the default sorting option, so when I navigated away from the Kubernetes landing page and then navigated back, I got this error:

Screen Shot 2023-01-06 at 4 17 21 PM

Interestingly clearing local storage made no difference, the API requests to get the clusters still have `x-filter: {"+order_by":"region","+order":"asc"}` so the Kubernetes landing page is essentially broken until I can find a way to unset that filter.

bnussman-akamai:
@jdamore-linode The filter is persisting because it is using user preferences under the hood. I should have clarified this in the PR, but in the next API release, filtering by `region` and `version` will start working. Therefore, when we release on Monday, we will not see this error.

jdamore-linode:
> @jdamore-linode The filter is persisting because it is using user preferences under the hood. I should have clarified this in the PR, but in the next API release, filtering by `region` and `version` will start working. Therefore, when we release on Monday, we will not see this error.
> 
> Edit: To override your preferences, add these query params to your url `?order=desc&orderBy=label`

> @jdamore-linode The filter is persisting because it is using user preferences under the hood. I should have clarified this in the PR, but in the next API release, filtering by `region` and `version` will start working. Therefore, when we release on Monday, we will not see this error.

Ah nice, that's perfect! I confirmed that everything else is working as expected and I'm in the middle of running E2E tests just to be thorough, but I'm assuming they'll all pass and will only circle back if there are any suspicious failures.

Nice work as always Banks!

jdamore-linode (approved):

packages/api-v4/src/kubernetes/kubernetes.ts
```diff
@@ -138,9 +138,15 @@ export const getKubernetesVersion = (versionID: string) =>
  *
  */
 
-export const getKubernetesClusterEndpoints = (clusterID: number) =>
+export const getKubernetesClusterEndpoints = (
+  clusterID: number,
+  params?: any,
+  filters?: any
```
CodyFinn-Akamai, line 144: Are these arguments needed? I don't think this endpoint has any filterable fields and does not accept any parameters.

CodyFinn-Akamai (approved):
Just a minor comment about optional arguments but otherwise everything works great! Nice job!

PR #8688, opened 2023-01-04
M3-5283: Support `completed` Account Maintenance Events

bnussman-akamai:
## Description 📝

- Redesigns `/account/maintenance` to support showing maintenance events with a status of `completed`

## Preview 📷

> **Warning**: Work in progress

## How to test 🧪

> **Warning**: Work in progress

### DIFF
--- a/packages/api-v4/src/account/types.ts
+++ b/packages/api-v4/src/account/types.ts
@@ -399,7 +399,7 @@ export interface SaveCreditCardData {
 
 export interface AccountMaintenance {
   reason: string;
-  status: 'pending' | 'started';
+  status: 'pending' | 'started' | 'completed';
   type: 'reboot' | 'cold_migration' | 'live_migration' | 'volume_migration';
   when: string;
   entity: {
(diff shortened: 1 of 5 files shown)


packages/manager/src/components/LinkButton.tsx
```diff
@@ -23,7 +23,7 @@ interface Props {
   isLoading?: boolean;
   className?: string;
   children: React.ReactNode | string;
-  onClick: () => void;
+  onClick?: () => void;
```
dwiley-akamai, line 26: Did TypeScript complain about it not being optional? From what I can see, every instance of `<LinkButton />` in the codebase has something passed to `onClick` (and given the purpose of the component, it would be odd to be using it without something happening upon a click)

dwiley-akamai (commented):
This is looking good to me, just a couple of minor points. 👌 

Screen Shot 2023-01-06 at 11 00 44 AM

We should add some padding to the left of "Pending" and "Completed" so it isn't flush with the edge in mobile view.

packages/manager/src/components/LinkButton.tsx
bnussman-akamai: I meant to revert this change. Thanks!

dwiley-akamai (approved):
🚢

jdamore-linode (approved):

PR #8689, opened 2023-01-05
M3-6044: Create Button "Create Using Command Line" and show modal 

cpathipa:
## Description 📝

- This story is part of feature API-CLI.
Team, guide me incase I miss setting any process oriented things before kicking off feature development. (feature flag, etc..)
Note: ga events are in my radar, if need following PR's will cover that.

**What does this PR do?**
- Shows API-CLI awareness modal upon clicking "Create Create Using Command Line"

## Preview 📷
image

## How to test 🧪

- Navigate to create Linode page
- Scroll to bottom and click ""Create Create Using Command Line"" button.
- Should show the modal.

Work in progress...

yarn test ApiAwarenessModal

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
@@ -769,12 +769,10 @@ export class LinodeCreate extends React.PureComponent<
               >
                 Create Linode
               </Button>
-              {this.state.isApiAwarenessModalOpen ? (
-                <ApiAwarenessModal
-                  isOpen={this.state.isApiAwarenessModalOpen}
-                  onClose={() => this.setIsApiAwarenessModalOpen(false)}
-                />
-              ) : null}
+              <ApiAwarenessModal
+                isOpen={this.state.isApiAwarenessModalOpen}
+                onClose={() => this.setIsApiAwarenessModalOpen(false)}
+              />
             </div>
           </Box>
         </Grid>


packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
```diff
@@ -373,6 +377,10 @@ export class LinodeCreate extends React.PureComponent<
     this.props.handleSubmitForm(payload, this.props.selectedLinodeID);
   };
 
+  toggleApiAwarenessModal = (value: boolean) => {
+    this.setState({ ...this.state, showApiAwarenessModal: value });
+  };
```
bnussman-akamai, line 382: Toggle functions generally imply no that no value is needed and it will just toggle the value between `true` and `false`. This function should be called something like `setIsApiAwarenessModalOpen` if we are going to take a paramater.

packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
```diff
@@ -733,20 +741,41 @@ export class LinodeCreate extends React.PureComponent<
...
-                (showAgreement && !signedAgreement)
-              }
-            >
-              Create Linode
-            </Button>
+            <div>
+              <Button
+                data-qa-api-cli-linode
+                buttonType="outlined"
+                onClick={this.toggleApiAwarenessModal.bind(null, true)}
```
bnussman-akamai, line 748: Throughout the repo we prefer the following

```suggestion
                onClick={() => this.toggleApiAwarenessModal(true)}
```

As much as I dislike arrow functions, I find this more readable.

packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { renderWithTheme } from 'src/utilities/testHelpers';
+
+import ApiAwarenessModal, { Props } from '.';
+
+const defaultProps: Props = {
+  isOpen: false,
+  onClose: jest.fn(),
+};
+
+const renderComponent = (overrideProps?: Pick<Props, 'isOpen'>) => {
```
bnussman-akamai, line 14: `Pick` may not be the right type here. I think `Partial` makes more sense. `Partial` makes all properties in the Props object optional. 

```suggestion
const renderComponent = (overrideProps?: Partial<Props>) => {
```

packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
```diff
@@ -733,20 +741,41 @@ export class LinodeCreate extends React.PureComponent<
...
+                }
+              >
+                Create Linode
+              </Button>
+              {this.state.showApiAwarenessModal ? (
+                <ApiAwarenessModal
+                  isOpen={this.state.showApiAwarenessModal}
+                  onClose={this.toggleApiAwarenessModal.bind(null, false)}
+                />
+              ) : null}
```
bnussman-akamai, line 777: We need to be careful with this. I think it is generally good to lazily render content, but when we do this on Modals and Drawers it actually cuts off the open and close animations.

bnussman-akamai (commented):
Awesome job. Left some small items of feedback!

packages/manager/src/components/index.ts
```diff
@@ -0,0 +1,6 @@
+import ActionsPanel from 'src/components/ActionsPanel';
```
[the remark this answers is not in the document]
cpathipa: I agree... But, we could definitely improve that iteratively by following certain patter. Also, we have ticket to address this in the backlog M3-6038.

dwiley-akamai (commented):
As a thought, if the intent is to bring awareness about the command line options to other entities as well in the future, maybe `ApiAwarenessModal` can be made a little more generic now by passing the modal copy/contents in as a child.

It'd look something like:

```
<Dialog>
     {children}
     <ActionsPanel>
          ...
     </ActionsPanel>
</Dialog>
```

the JSX for this specific modal would be defined in `LinodeCreate.tsx` and then passed as a children/render prop to `<ApiAwarenessModal />`.

If we take this approach, the component and its test should be moved from the `/LinodesCreate` directory too

packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+      title="Create using command line"
+      open={isOpen}
+      onClose={onClose}
+      maxWidth={'sm'}
+      fullHeight
+    >
+      <Typography variant="body1">
+        You&#39;ll first need to{' '}
+        <ExternalLink
+          link="https://cloud.linode.com/profile/tokens"
```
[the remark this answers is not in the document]
cpathipa: @dwiley-akamai Upon clicking the relative path link, wouldn't it cause user to lose the context by taking away from Modal?

packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
cpathipa: I'm not aware of that.. but, good to know. Are you able to see such behavior here?

packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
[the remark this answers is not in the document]
dwiley-akamai: I just double-checked, using `<Link />` would actually cause that to happen, but we should be fine making that adjustment with `<ExternalLink />`:

```
        <ExternalLink
          link="/profile/tokens"
          text="create an API access token"
          hideIcon
        />
```

The reason we want to use the relative link here is to preserve the environment you're in. For example, right now if you're on your local (localhost:3000), clicking the absolute link takes you to the prod site (cloud.linode.com) instead of keeping you on your local.

cpathipa:
> As a thought, if the intent is to bring awareness about the command line options to other entities as well in the future, maybe `ApiAwarenessModal` can be made a little more generic now by passing the modal copy/contents in as a child.
> 
> It'd look something like:
> 
> ```
> <Dialog>
>      {children}
>      <ActionsPanel>
>           ...
>      </ActionsPanel>
> </Dialog>
> ```
> 
> the JSX for this specific modal would be defined in `LinodeCreate.tsx` and then passed as a children/render prop to `<ApiAwarenessModal />`.
> 
> If we take this approach, the component and its test should be moved from the `/LinodesCreate` directory too

Good call @dwiley-akamai! That was one of the reason for decoupling `ApiAwarenessModal` from `LinodeCreate` in this iteration. But, we could definitely make it as more generic considering future wireframes it will be more clearer to convert `ApiAwarenessModal` as reusable.

packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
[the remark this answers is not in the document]
cpathipa: I see what you mean, Thank you for checking that @dwiley-akamai !

packages/manager/src/components/index.ts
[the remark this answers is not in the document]
cpathipa: @bnussman-akamai , @dwiley-akamai , @carrillo-erik Thank you for providing your opinions on this, appreciate that. I see that we all have common goal of composing optimal components with certain pattern and minimal confusion. We have an opportunity to explore on industrial practices, we could do spike / POC as part of M3-6038 ticket. This imports changes shouldn’t be part of this feature. So, removed from this PR.

packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
bnussman-akamai: The open animation looks good to me, but _i think_ i can see the close animation getting cut off

packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
```diff
@@ -733,20 +741,41 @@ export class LinodeCreate extends React.PureComponent<
...
-                (showAgreement && !signedAgreement)
-              }
-            >
-              Create Linode
-            </Button>
+            <div>
+              <Button
+                data-qa-api-cli-linode
+                buttonType="outlined"
+                onClick={() => this.setIsApiAwarenessModalOpen(true)}
```
carrillo-erik, line 748: I know this is still open, but I did notice that at the moment the `Create Using Command Line` click to action button opens the modal without checking that the required fields have been filled, ie. `Root Password`, `Linode Plan` selection.

packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
dwiley-akamai: Same, the close looks more abrupt than it does for the Rebuild, Rescue, and Migrate modals, for example

carrillo-erik (approved):

packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+import ExternalLink from 'src/components/ExternalLink';
+
+import Typography from 'src/components/core/Typography';
+
+export interface Props {
+  isOpen: boolean;
+  onClose: () => void;
+}
+
+const ApiAwarenessModal = ({ isOpen, onClose }: Props) => {
```
CodyFinn-Akamai, line 15: I think deconstructing in the function definition is a little confusing and we more consistently deconstruct props in the body of the component

dwiley-akamai (approved):

PR #8690, opened 2023-01-05
M3-5980: Update text for Akamai-billed customers

mjac0bs:
## Description 📝

**What does this PR do?**
For _Akamai-billed customers only_, we are making the following changes to the billing information text:
 - Removing the /Account banner that says: "Linode products and services will appear on your Akamai Technologies invoice"
 - Removing the text "Future Linode products and services will appear on your Akamai Technologies invoice" from the  billing table
 - Adding a tooltip to explain invoicing in the billing table with new copy
 - Updating the header of the billing table to correspond with new tooltip copy

## Preview 📷

Previously:

https://user-images.githubusercontent.com/114685994/211042981-70245bc7-6dd6-4410-910d-3626642554a6.mov

This change:

https://user-images.githubusercontent.com/114685994/211069140-b5ca21cb-477d-47b8-8cb1-1698d5dba18e.mov

## How to test 🧪

1. In `src/factories/account.ts`, change the `billing_source` from `linode` to `akamai` and then turn on the MSW to mock an Akamai-billed account.
2. Navigate to http://localhost:3000/account/billing.
3. Observe there is no dismissible banner about Linode/Akamai invoicing at the top of the page and no custom first row in the table with Linode/Akamai invoicing.
4. Observe that the billing table header copy is "Usage History", a usage history tooltip with a dashed underline displays next to the header, and hovering over tooltip displays updated copy with link below the tooltip. **Note:** Close the notices at the top of the page and zoom out browser window to move the mocked page content far enough from the bottom of the window that the tooltip popover will display underneath the tooltip rather than on top, which is MUI Tooltip default behavior to avoid cutting the popover text off.
5. Adjust the window size and check that the table header components still display neatly with the addition of the tooltip.
6. Turn off MSW and confirm that http://localhost:3000/account/billing displays and behaves like prod. New "usage history" copy changes should not be visible to accounts with a Linode billing source; these accounts did not see the Linode/Akamai invoice banner or custom table text to begin with.

```
yarn test TableContentWrapper
```

### DIFF
--- a/packages/manager/src/components/BillingTooltip.tsx
+++ b/packages/manager/src/components/BillingTooltip.tsx
@@ -6,7 +6,6 @@ import Link from 'src/components/Link';
 
 interface Props {
   text: string;
-  className?: string;
 }
 
 const useStyles = makeStyles((theme: Theme) => ({
@@ -21,6 +20,11 @@ const useStyles = makeStyles((theme: Theme) => ({
   flex: {
     display: 'flex',
     width: 'auto !important',
+  },
+  popper: {
+    '& .MuiTooltip-tooltip': {
+      minWidth: 375,
+    },
   },
 }));
 
@@ -46,6 +50,7 @@ export const BillingTooltip: React.FC<Props> = (props) => {
       title={akamaiBillingInvoiceText}
       placement="bottom"
       className={classes.root}
+      classes={{ popper: classes.popper }}
     >
       <Typography>{text}</Typography>
     </ToolTip>
--- a/packages/manager/src/components/BillingTooltip.tsx
+++ b/packages/manager/src/components/BillingTooltip.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { makeStyles, Theme } from 'src/components/core/styles';
 import ToolTip from 'src/components/core/Tooltip';
 import Typography from 'src/components/core/Typography';
-import { akamaiBillingInvoiceText } from 'src/features/Billing/billingUtils';
+import Link from 'src/components/Link';
 
 interface Props {
   text: string;
@@ -27,6 +27,18 @@ const useStyles = makeStyles((theme: Theme) => ({
 export const BillingTooltip: React.FC<Props> = (props) => {
   const classes = useStyles();
   const { text } = props;
+  const akamaiBillingInvoiceText = (
+    <Typography>
+      Charges in the final Akamai invoice should be considered the final source
+      of truth. Linode invoice will not reflect discounting, currency
+      adjustment, or any negotiated terms and conditions. Condensed and
+      finalized invoice is available within{' '}
+      <Link to="https://control.akamai.com/apps/billing">
+        Akamai Control Center &gt; Billing
+      </Link>
+      .
+    </Typography>
+  );
 
   return (
     <ToolTip
--- a/packages/manager/src/components/BillingTooltip.tsx
+++ b/packages/manager/src/components/BillingTooltip.tsx
@@ -1,0 +1,45 @@
+import * as React from 'react';
+import { makeStyles, Theme } from 'src/components/core/styles';
+import ToolTip from 'src/components/core/Tooltip';
+import Typography from 'src/components/core/Typography';
+import { akamaiBillingInvoiceText } from 'src/features/Billing/billingUtils';
+
+interface Props {
+  text: string;
+  className?: string;
+}
+
+const useStyles = makeStyles((theme: Theme) => ({
+  root: {
+    position: 'relative',
+    padding: `4 4 0 4`,
+    backgroundColor: 'transparent',
+    transition: theme.transitions.create(['background-color']),
+    borderRadius: 4,
+    border: 'none',
+    cursor: 'pointer',
+    borderBottom: `1px dotted ${theme.palette.primary.main}`,
+    color: theme.palette.primary.main,
+  },
+  flex: {
+    display: 'flex',
+    width: 'auto !important',
+  },
+}));
+
+export const BillingTooltip: React.FC<Props> = (props) => {
+  const classes = useStyles();
+  const { text } = props;
+
+  return (
+    <ToolTip
+      title={akamaiBillingInvoiceText}
+      placement="bottom"
+      className={classes.root}
+    >
+      <Typography>{text}</Typography>
+    </ToolTip>
+  );
+};
+
+export default BillingTooltip;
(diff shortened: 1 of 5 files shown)


packages/manager/src/components/BillingTooltip.tsx
```diff
@@ -0,0 +1,54 @@
...
+import ToolTip from 'src/components/core/Tooltip';
+import Typography from 'src/components/core/Typography';
+import Link from 'src/components/Link';
+
+interface Props {
+  text: string;
+  className?: string;
+}
+
+const useStyles = makeStyles((theme: Theme) => ({
```
mjac0bs, line 12: Note that the width of the tooltip popover is not as wide as the popover shown in the mockups. I'm still taking a look at this to find the correct CSS to apply -- but if there are any thoughts, much appreciated.

packages/manager/src/features/Billing/BillingPanels/BillingActivityPanel/BillingActivityPanel.tsx
```diff
@@ -341,8 +320,13 @@ export const BillingActivityPanel: React.FC<Props> = (props) => {
     <div className={classes.root}>
       <div className={classes.headerContainer}>
         <Typography variant="h2" className={classes.headline}>
-          Billing &amp; Payment History
+          {`${isAkamaiCustomer ? 'Usage' : 'Billing & Payment'} History`}
         </Typography>
+        {isAkamaiCustomer ? (
+          <div className={classes.headerLeft}>
```
dwiley-akamai, line 326: We can probably do without the padding, but at a minimum we should get rid of it for the mobile view:

Screen Shot 2023-01-06 at 3 05 10 PM

packages/manager/src/features/Billing/BillingPanels/BillingActivityPanel/BillingActivityPanel.tsx
```diff
@@ -341,8 +320,13 @@ export const BillingActivityPanel: React.FC<Props> = (props) => {
...
+            <BillingTooltip text="Usage History may not reflect finalized invoice"></BillingTooltip>
```
dwiley-akamai, line 327: Since there are no children for `BillingTooltip`, we could go with

`<BillingTooltip text="Usage History may not reflect finalized invoice" />`

dwiley-akamai (approved):
LGTM aside from a couple of minor points!

packages/manager/src/components/BillingTooltip.tsx
```diff
@@ -0,0 +1,60 @@
...
+interface Props {
+  text: string;
+}
+
+const useStyles = makeStyles((theme: Theme) => ({
+  root: {
+    position: 'relative',
+    padding: 4,
+    borderRadius: 4,
+    cursor: 'pointer',
```
bnussman-akamai, line 16: Does `pointer` make sense here? This is more hoverable than it is clickable.

CodyFinn-Akamai (approved):

packages/manager/src/components/BillingTooltip.tsx
```diff
@@ -0,0 +1,60 @@
...
+    textDecoration: `underline dotted ${theme.palette.primary.main}`,
```
bnussman-akamai, line 17: I don't think we've ever done this in cloud so we should just be aware we may be introducing a new pattern here (which is completely fine). Typically in these situations I feel like we've historically used normal text and a `<HelpIcon />`

bnussman-akamai (approved):

packages/manager/src/components/BillingTooltip.tsx
bnussman-akamai: Actually we do this a lot, so it makes sense here to be consistent.

packages/manager/src/components/BillingTooltip.tsx
```diff
@@ -0,0 +1,60 @@
...
+      Charges in the final Akamai invoice should be considered the final source
+      of truth. Linode invoice will not reflect discounting, currency
+      adjustment, or any negotiated terms and conditions. Condensed and
+      finalized invoice is available within{' '}
+      <Link to="https://control.akamai.com/apps/billing">
+        Akamai Control Center &gt; Billing
+      </Link>
+      .
+    </Typography>
+  );
```
hana-linode, line 45: Since this is static, we can move this out of the component to reduce unnecessary re-renders.

This component is also very simple and only used in one place, so we could also move this variable declaration to `billingUtils.ts` and just use the `ToolTip` component directly in `BillingActivityPanel`

packages/manager/src/components/BillingTooltip.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const { text } = props;
+  const akamaiBillingInvoiceText = (
+    <Typography>
+      Charges in the final Akamai invoice should be considered the final source
+      of truth. Linode invoice will not reflect discounting, currency
+      adjustment, or any negotiated terms and conditions. Condensed and
+      finalized invoice is available within{' '}
+      <Link to="https://control.akamai.com/apps/billing">
+        Akamai Control Center &gt; Billing
+      </Link>
```
hana-linode, line 42: Since this opens in a new tab we should add an external link icon here

hana-linode (approved):
LGTM, left some non-blocking comments!

cpathipa (approved):
LGTM!

packages/manager/src/components/BillingTooltip.tsx
mjac0bs: Yeah, the ticket for this calls it out as a new pattern, but similar interaction to the `<HelpIcon />`. I'm not sure if the reason for using text over an icon in this case is to increase visibility and clarity for the user rather than hiding the copy behind a ? but @matthewbartholomew might be able to provide more detail.

packages/manager/src/components/BillingTooltip.tsx
mjac0bs: Good point on moving out of the component, will do. Wouldn't moving the variable declaration to `billingUtils.ts` require importing components into the utils file -- is that something we're okay? I haven't seen much of it.

hana-linode: @mjac0bs oh good point, you can disregard the second sentence

packages/manager/src/features/Billing/BillingPanels/BillingActivityPanel/BillingActivityPanel.tsx
mjac0bs: Thanks for the misalignment catch on mobile. I removed the left padding at that breakpoint. I did keep the non-mobile padding and actually increased in a little to more closely match the spacing in the mockup. Without the padding, it feels a little crowded to me with all those words.

PR #8691, opened 2023-01-06
M3-5896: Enable Cypress recording and parallelization via Jenkins

jdamore-linode:
## Description 📝

**What does this PR do?**
This implements the Cloud-specific changes needed to enable parallelization when running our Cypress tests via Jenkins.

Specifically, this makes the following changes:

- Defines additional services in `docker-compose.yml` to facilitate running multiple Cypress test runners at once
- Refactors `docker-compose.yml` to reduce redundancy
- Passes build info to the test runner containers via environment variables

## How to test 🧪
There's no real testing that you can do locally to verify these changes, but feel free to reach out via Slack and I can refer you to successful builds which use this code.


docker-compose.yml
```diff
@@ -1,5 +1,37 @@
...
+  &default-env
+  BUILD_ID: ${BUILD_ID}
+  BUILD_URL: ${BUILD_URL}
+  BUILD_NUMBER: ${BUILD_NUMBER}
+  COMMIT_INFO_BRANCH: ${GIT_BRANCH}
+  COMMIT_INFO_SHA: ${GIT_COMMIT}
+  ghprbPullId: ${ghprbPullId}
+  JENKINS_HOME: ${JENKINS_HOME}
+  JENKINS_VERSION: ${JENKINS_VERSION}
+  HUDSON_URL: ${HUDSON_URL}
```
bnussman-akamai, line 14: What does Hudson enable us to do? I'm not familiar with it

docker-compose.yml
```diff
@@ -1,5 +1,37 @@
...
+version: '3.4'
+
+x-e2e-env:
+  &default-env
+  BUILD_ID: ${BUILD_ID}
+  BUILD_URL: ${BUILD_URL}
+  BUILD_NUMBER: ${BUILD_NUMBER}
+  COMMIT_INFO_BRANCH: ${GIT_BRANCH}
+  COMMIT_INFO_SHA: ${GIT_COMMIT}
+  ghprbPullId: ${ghprbPullId}
```
bnussman-akamai, line 11: What consumes this env variable? It's the only lowercase one 🥲

bnussman-akamai (approved):
So exciting 🚀

There's some stuff in here i've never seen before like `&default-env` but I trust that it all works!

docker-compose.yml
jdamore-linode: Hudson is actually just the name of the software that Jenkins was forked from! So I'm guessing that variable name is just a really old holdover.

The reason I'm passing these environment variables (including `ghprbPullId`) is because they're the ones listed for Jenkins in Cypress's `ci_provider.js` helper [here](https://github.com/cypress-io/cypress/blob/develop/packages/server/lib/util/ci_provider.js#L71) and [here](https://github.com/cypress-io/cypress/blob/develop/packages/server/lib/util/ci_provider.js#L316).

Since Cypress is running in a container, it can't read these environment variables that Jenkins sets implicitly, so I have to manually pass them to the containers for that detection to work. I'm not sure if every single one is required, but since they're what Cypress checks for by default I figured it doesn't hurt to pass them all.

docker-compose.yml
```diff
@@ -29,19 +64,68 @@ services:
...
-    volumes:
-      - ./packages/manager:/home/node/app/packages/manager
-      - ./packages/validation:/home/node/app/packages/validation
-      - ./packages/api-v4:/home/node/app/packages/api-v4
-      - ./package.json:/home/node/app/package.json
-      - ./node_modules:/home/node/app/node_modules
+    entrypoint: yarn cy:ci
+    volumes: *default-volumes
+
+  e2e-1:
```
bavedarnow, line 73: this may be overkill, but I _think_ you can use extensions to DRY these 4 almost-identical services

https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd

cpathipa (approved):
Nice improvement on enabling parallelization. 
Observation: One test failed upon running E2E tests on this branch. I'm sure this not due to the changes in this PR. May be not having latest changes from develop.
image

PR #8692, opened 2023-01-07
feat: Manual Backups for Managed Databases

ericsund:
## Description 📝

**What does this PR do?**

Enables managed database users to trigger backups on demand.  This will hit the `POST` endpoint with a custom label, defined here in the public API: https://www.linode.com/docs/api/databases/#managed-mysql-database-backup-snapshot-create

Since we have a limit on the number of manual backups, this PR also adds the ability to delete some in order to make room for more.  Deletion works for either automatic or manual backups.  It hits the `DELETE` endpoint defined here: https://www.linode.com/docs/api/databases/#managed-mongodb-database-backup-delete

The confirmation dialogs have been changed such that:
* The formatted date should be entered to restore an automatic or manual backup
* The formatted date should be entered to delete a manual backup
* Deleting automatic backups is not supported.  Deleting them does not count against the 3 manual backup limit.

Thoughts on this? ☝️ 

## Preview 📷

https://user-images.githubusercontent.com/5467060/213305403-950a5935-f837-4851-b6c7-d2e36d565add.mov

## How to test 🧪

Head to the managed databases using the left sidebar.  Create a new managed database (or use an existing one).  Select your database, then hit the `Backups` tab.  From here, you should see a `Manual Backup` section, where you'll be able to click the blue `Start Backup` button.  Enter a custom label using only alphanumeric characters and no spaces, then click `Backup Database`.  Observe that the correct endpoint is hit successfully.

`yarn test` (for now)


packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DatabaseBackups.tsx
```diff
@@ -115,13 +140,35 @@ export const DatabaseBackups: React.FC = () => {
...
       </Table>
+      <Paper style={{ marginTop: 16 }}>
+        <Typography variant="h3">Manual Backup</Typography>
+        <Typography style={{ lineHeight: '20px', marginTop: 4 }}>
+          Trigger a manual backup outside of the usual schedule.<br></br>
+        </Typography>
+        <Button
+          buttonType="primary"
+          onClick={() => onBackup()}
+          style={{ marginLeft: 12, marginTop: 12 }}
```
bnussman-akamai, line 151: I think this button looks a bit weird with the left margin. I think it would be okay to align it left with the header and other text.

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DatabaseDeleteDialog.tsx
```diff
@@ -0,0 +1,141 @@
...
+        visible={preferences?.type_to_confirm}
+        placeholder={
+          backup.type === 'snapshot' ? backup.label : formatDate(backup.created)
+        }
+      />
+    </ConfirmationDialog>
+  );
+};
+
+const enhanced = compose<CombinedProps, Props>(withPreferences());
```
bnussman-akamai, line 139: Rather than using withPreferences, can you use the `usePreferences` hook

packages/manager/src/utilities/formatBackupLabel.ts
```diff
@@ -0,0 +1,11 @@
+export const formatBackupLabel = (backup: any): string => {
```
bnussman-akamai, line 1: I think I'd prefer to have this file in `packages/manager/src/features/Databases/` called something more generic like `utils.ts`. That way we can add more DBaaS helper functions in the future.

packages/manager/src/utilities/formatBackupLabel.ts
```diff
@@ -0,0 +1,11 @@
...
+  if (backup.type === 'auto') {
+    return 'Automatic Backup';
+  } else if (backup.type === 'snapshot') {
+    return backup.label;
+  }
+  // The backup type is unknown, so return an empty string.
+  return '';
+};
+
+export default formatBackupLabel;
```
bnussman-akamai, line 11: Lets remove the default export

```suggestion
```

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DatabaseDeleteDialog.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  const {
+    mutateAsync: manualDelete,
+    isLoading,
+    error,
+  } = useDeleteBackupMutation(database.engine, database.id, backup.id);
+
+  const handleDeleteBackup = () => {
+    manualDelete().then(() => {
+      history.push('summary');
```
bnussman-akamai, line 45: I don't think we need to redirect the user back to summary here

```suggestion
```

bnussman-akamai (commented):
Thank you for adding these features! Here are a few comments to start with

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DatabaseDeleteDialog.tsx
ericsund: Can you describe a the benefit of using `usePerferences()` over `withPerferences()`?

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DatabaseBackups.tsx
ericsund: Fair point -- removed the left offset.

packages/manager/src/utilities/formatBackupLabel.ts
ericsund: Good idea, this will indeed grow in the future.

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DatabaseDeleteDialog.tsx
bnussman-akamai: No major benefits, it's just a more modern approach, uses less code, and is easier to trace.

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DatabaseBackups.css
```diff
@@ -0,0 +1,6 @@
+.backups_section {
+  clear: both;
+  display: flex;
+  vertical-align: top;
+  justify-content: space-between;
+}
```
bnussman-akamai, line 6: In Cloud Manager, we try to avoid using `.css` files whenever possible. Can we move this into the same file as a component?

packages/manager/src/features/Databases/databaseUtils.ts
```diff
@@ -0,0 +1,25 @@
...
+  if (backup.type === 'auto') {
+    return 'Automatic';
+  } else if (backup.type === 'snapshot') {
+    return 'Manual';
+  }
+  // The backup type is unknown, so return an empty string.
+  return '';
+};
+
+export const generateManualBackupLabel = (): string => {
```
bnussman-akamai, line 11: Why did we decide to not let users choose their own label? We should let users choose their label if the API supports it.

bnussman-akamai (changes requested):

packages/manager/src/features/Databases/databaseUtils.ts
ericsund: Marking this as resolved: discussed that this approach would make the UI more intuitive for this scope.

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DatabaseBackups.tsx
```diff
@@ -109,27 +136,63 @@ export const DatabaseBackups: React.FC = () => {
...
+            </Hidden>
             <TableCell style={{ width: 100 }}></TableCell>
           </TableRow>
         </TableHead>
         <TableBody>{renderTableBody()}</TableBody>
       </Table>
+      <Paper style={{ marginTop: 16 }}>
+        <Typography variant="h3">Manual Backup</Typography>
+        <div
+          className="backups_section"
```
bnussman-akamai, line 150:
```suggestion
```

bnussman-akamai (changes requested):

bnussman-akamai (approved):
Nice work on this!

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/CreateBackupDialog.tsx
```diff
@@ -0,0 +1,73 @@
...
+        error
+          ? getAPIErrorOrDefault(error, 'Unable to create this backup.')[0]
+              .reason
+          : undefined
+      }
+    >
+      <Typography>
+        Would you like to perform a manual backup of this database? When you
+        reach 3 manual backups, please delete an existing one before adding
+        another.
```
hana-linode, line 69: Copy sounds a bit off, did UX approve of this copy? 🤔

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DatabaseBackups.tsx
```diff
@@ -109,27 +136,62 @@ export const DatabaseBackups: React.FC = () => {
...
+          }}
+        >
+          <Typography style={{ lineHeight: '20px', marginTop: 4 }}>
+            Trigger a manual backup outside of the usual schedule.<br></br>
+          </Typography>
+          <Button buttonType="primary" onClick={() => onBackup()}>
+            Start Backup
+          </Button>
+        </div>
+      </Paper>
```
hana-linode, line 164: Can we hide this while the Database is provisioning?
image

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DeleteBackupDialog.tsx
```diff
@@ -0,0 +1,105 @@
...
+          ? getAPIErrorOrDefault(error, 'Unable to delete this backup.')[0]
+              .reason
+          : undefined
+      }
+    >
+      <Notice warning>
+        <Typography style={{ fontSize: '0.875rem' }}>
+          <strong>Warning:</strong> Deleting a backup is irreversible. You are
+          not deleting the database cluster itself with this action. Deleting a
+          manual backup frees up space that counts toward the limit of 3.
```
hana-linode, line 86: Same here, copy seems a bit verbose/off. Did UX approve this copy?

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DeleteBackupDialog.tsx
```diff
@@ -0,0 +1,105 @@
...
+            <strong>{formatDate(backup.created)}</strong>) from (
+            <strong>{database.label}</strong>) in the field below.
+          </span>
+        }
+        onChange={(input) => setConfirmationText(input)}
+        value={confirmationText}
+        label={`Manual backup date from ${database.label}`}
+        visible={preferences?.type_to_confirm}
+        placeholder={formatDate(backup.created)}
+      />
```
hana-linode, line 102: Can we use the `TypeToConfirmDialog` component here? (might require a little bit of refactoring)

packages/manager/src/features/Databases/databaseUtils.ts
```diff
@@ -0,0 +1,22 @@
...
+    'Manual' +
+    currDate.getFullYear() +
+    +(currDate.getMonth() + 1) +
+    +currDate.getDate() +
+    'H' +
+    currDate.getHours() +
+    'M' +
+    +currDate.getMinutes() +
+    'S' +
+    +currDate.getSeconds()
```
hana-linode, line 20: We can use a template literal here instead of concatenating strings together

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/RestoreFromBackupDialog.tsx
```diff
@@ -100,21 +99,31 @@ export const RestoreFromBackupDialog: React.FC<CombinedProps> = (props) => {
...
+          )
         }
         onChange={(input) => setConfirmationText(input)}
         value={confirmationText}
-        label="Database Label"
+        label={
+          backup.type === 'snapshot'
+            ? `Manual backup date from  ${database.label}`
+            : `Automatic backup date from  ${database.label}`
+        }
```
hana-linode, line 123: We can clean up these `backup.type === 'snapshot'` conditionals with a map variable

Something like:
```typescript
  interface backupDialogData {
    title: string;
    confirmationText: JSX.Element;
    label: string;
  }

  const backupDialogMap: Record<DatabaseBackupType, backupDialogData> = {
    snapshot: {
      title: `Restore Manual Backup ${formatDate(backup.created)}`,
      confirmationText: (
        <span>
          To confirm manual backup restoration, type the date/time of the manual
          backup (<strong>{formatDate(backup.created)}</strong>) from the{' '}
          <strong>{database.label}</strong> database in the field below.
        </span>
      ),
      label: `Manual backup date from  ${database.label}`,
    },
    auto: {
      title: `Restore Automatic Backup ${formatDate(backup.created)}`,
      confirmationText: (
        <span>
          To confirm automatic backup restoration, type the date/time of the
          automatic backup (<strong>{formatDate(backup.created)}</strong>) from
          the <strong>{database.label}</strong> database in the field below.
        </span>
      ),
      label: `Automatic backup date from  ${database.label}`,
    },
  };
```

hana-linode (commented):
Functionality looks good! Left some clean up comments

matthewbartholomew:
This is looking great. UX copy needs to be finalized.

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DatabaseBackups.tsx
bnussman-akamai: Pushed up a03291a to address this @hana-linode. Can you double check my logic in that commit?

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DeleteBackupDialog.tsx
ericsund: Sounds like @matthewbartholomew will follow up.

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/DeleteBackupDialog.tsx
ericsund: Can you describe some differences between `TypeToConfirmDialog` and `TypeToConfirm`?  We use one in `/object-storage/buckets` and it appears functionally the same.  Any advantages of using one over the other?

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/RestoreFromBackupDialog.tsx
ericsund: Nifty!  Thanks, I hadn't thought of doing this.

packages/manager/src/components/TypeToConfirmDialog/TypeToConfirmDialog.tsx
```diff
@@ -82,14 +107,17 @@ const TypeToConfirmDialog: React.FC<CombinedProps> = (props) => {
...
+          TypeToConfirmDialogMap[entity.type].labelPrefix
+        }`}
         confirmationText={
           confirmationText ? (
             confirmationText
           ) : (
             <span>
-              To confirm deletion, type the name of the {entity.type} (
-              <b>{entity.label}</b>) in the field below:
+              To confirm deletion, type the{' '}
```
ericsund, line 118: There are two usages of `TypeToConfirmDialog` in this change: one in `RestoreFromBackupDialog.tsx` and one in `DeleteBackupDialog.tsx`.  In both cases, we provide a custom `confirmationText` prop which overrides the default one, so we don't ever need it.

However, I've made the default dialog more generic by adding a `TypeToConfirmDialogMap`.  The intention here is to make this dialog future-proof so we can use it on more `EntityType`s.  Let me know if this is worthwhile to keep (even though it doesn't actually benefit us now).

dwiley-akamai: I don't think it'd be bad to keep this, seems nice to have. I'd probably move lines 114-120 into a variable outside the `return` statement just to tidy this area of the code up a bit. Could store `TypeToConfirmDialogMap[entity.type].labelPrefix` in a `const` as well.

dwiley-akamai (commented):
We need to handle for the event in the Events dropdown:

Screen Shot 2023-02-15 at 4 31 16 PM

(this would be done in `eventMessageGenerator.ts`)

It also seems that the Backups table doesn't update on its own once the manual backup is created successfully

jaalah-akamai:
@ericsund A few minor changes here and we should be good to go!

jaalah-akamai:
This PR is paused since until we determine who will be taking over this work from @ericsund and if this is the route they want to continue to take.

bnussman-akamai:
This PR can probably be reviewed by the team. All it needs is final copy 📝

PR #8695, opened 2023-01-09
M3-5717: Storybook patterns

CodyFinn-Akamai:
## Description 📝

This PR cleans up and standardizes the format, code patterns, and layout of the Storybook stories.


packages/manager/src/components/CircleProgress/CircleProgress.stories.mdx
```diff
@@ -15,17 +15,7 @@ export const Template = (args) => <CircleProgress {...args}></CircleProgress>;
 <Canvas>
   <Story
     name="Circle Progress"
-    args={{ children: '25%' }}
     argTypes={{
-      children: {
```
mjac0bs, line 20: Was just wondering why you chose to remove the children arg in this story -- is there a best practice for when to include it vs. exclude it? Do we just not display the percentage anywhere in the UI?

packages/manager/src/components/TagsPanel/TagsPanel.stories.mdx
```diff
@@ -23,29 +24,13 @@ import TagsPanel from './TagsPanel';
...
   return (
     <>
-      <TagsPanel tags={tags} updateTags={updateTags} {...args} />
-      <Typography style={{ marginTop: 12 }}>
-        <i>
-          Modifying the tags in this example does not currently work at the
-          moment.
-        </i>
-      </Typography>
+      <TagsPanel updateTags={updateTags} {...args} />
```
mjac0bs, line 33: 🎉 for working functionality on the Tags Panel.

mjac0bs (commented):
🥇 to you for going through this many Storybook components! I left a lot of comments, but they're mostly nitpicks -- typos, punctuation, small suggestions. Also wanted to call out a couple comments I made on `04-component-library.md` about things we may want to include in the "Best Practices" section.

packages/manager/src/components/CircleProgress/CircleProgress.stories.mdx
CodyFinn-Akamai: I could not find anyplace where the percentage is shown. Also the `Circle Progress` documentation says that it is an `indeterminate indicator` so a percentage would break that rule.

mjac0bs (approved):
`develop` has some changes to at least one story (`DismissibleBanner`) that will conflict with these changes, so I didn't recheck out this PR and re-run Storybook. From the file diff, though, things look good and thanks for making the adjustments from initial feedback!

packages/manager/src/components/TagsPanel/TagsPanel.stories.mdx
mjac0bs: One final question: does this fix make existing ticket M3-5731 a noop? 🤔

CodyFinn-Akamai: I think so. Nice catch!

bnussman-akamai (approved):

PR #8696, opened 2023-01-09
Fixes StackScript Create UI following MUI v5 Changes

bnussman-akamai:
## Description 📝

- Fixes broken stackscript create form UI after MUI v5 update

### Before 

Screen Shot 2023-01-09 at 12 20 03 PM

### After 

Screen Shot 2023-01-09 at 12 19 43 PM

## How to test 🧪

- Test and check ui of stackscript create

### DIFF
--- a/packages/manager/src/features/Images/ImageSelect.tsx
+++ b/packages/manager/src/features/Images/ImageSelect.tsx
@@ -19,7 +19,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     marginLeft: -20,
   },
   selectContainer: {
-    width: 415 + theme.spacing(2),
+    width: `calc(415px + ${theme.spacing(2)})`,
     [theme.breakpoints.down('sm')]: {
       width: '100%',
     },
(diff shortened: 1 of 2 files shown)


CodyFinn-Akamai (approved):

dwiley-akamai (commented):
I noticed a couple of things –

The tooltip icon should be centered vertically centered relative to the dropdown (it's not exactly centered right now in prod, but right now it's still further up than it is in prod):
Screen Shot 2023-01-09 at 1 04 11 PM

The "Description" box on the "Capture Image" tab is huge compared to prod:
Screen Shot 2023-01-09 at 1 02 19 PM

bnussman-akamai:
@dwiley-akamai Should be better now! We should fix the ImageSelect help icon alignment as a separate issue.

dwiley-akamai (approved):

jdamore-linode (approved):
Nice work Banks!

Confirmed that the StackScript create page and longview gauges are fixed. There's still a little bit of strangeness with the Target Image field on the StackScript create page, but I'm not sure it's worth holding up the merge:

Screen Shot 2023-01-09 at 2 10 27 PM

* The help tooltip is still slightly misaligned, but it is a lot better than it was
* The "Target Images" label seems to be just slightly cut off at the bottom

PR #8700, opened 2023-01-09
Fix E2E Tests for MUIv5, Region Selection Improvements

jdamore-linode:
## Description 📝

**What does this PR do?**

* Fixes the StackScript test failures related to the MUIv5 upgrade
* Fixes the Image, Migration, and StackScript test failures related to the recent region selection improvements

**MUIv5 Fix**

Since upgrading to MUIv5, our `<HelpIcon />` buttons have an `aria-label` attribute containing the contents of the tooltip when it had previously used the `title` attribute. `cy.findByLabelText` checks against `aria-label`, so it was tripping up on that help icon since its content matches our query.

This PR increases the strictness of the selector, matching against any element with a label which begins with "StackScript Label" (case sensitive). Now only the input field will be selected, not the tooltip button.

**Region Selection Fix**

This adds some helpers to make interacting with region selection drop-downs easier. There's still some discussion around these improvements, so if they should change in a breaking way again it'll only require updating these two util functions to fix the affected tests.

## How to test 🧪

To reproduce, run the command below against `develop`. To confirm the changes, run it against this branch.
```bash
yarn cy:run -s "cypress/e2e/stackscripts/create-stackscripts.spec.ts,cypress/e2e/firewalls/migrate-linode-with-firewall.spec.ts,cypress/e2e/images/machine-image-upload.spec.ts"
```

### DIFF
--- a/packages/manager/cypress/e2e/firewalls/migrate-linode-with-firewall.spec.ts
+++ b/packages/manager/cypress/e2e/firewalls/migrate-linode-with-firewall.spec.ts
@@ -152,7 +152,9 @@ describe('Migrate Linode With Firewall', () => {
     getClick('[data-qa-checked="false"]');
     cy.findByText(`United States: Dallas, TX`).should('be.visible');
     containsClick(selectRegionString);
-    fbtClick('Singapore, SG');
+
+    ui.regionSelect.findItemByRegionName('Singapore, SG').click();
+
     fbtClick('Enter Migration Queue');
     cy.wait('@migrateReq').its('response.statusCode').should('eq', 200);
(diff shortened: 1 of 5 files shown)


bnussman-akamai (approved):
Tests passed! You rock joe! 🪨

jdamore-linode:
Just updated this PR to include fixes related to the recent Region Select improvements

cpathipa (commented):
All tests passed other than `create-database.spec.ts`. Is this due to DB provisioning ? 
image

cpathipa (approved):

PR #8702, opened 2023-01-10
Fix `null` Event `username`

bnussman-akamai:
## Description 📝

- Thank you @mjac0bs for finding this bug!
- For some reason, the API is returning `null` for a small number of events, which it does not claim to do. 
- We need to handle null usernames in some of our logic or else event related features of cloud manager cuse the app to crash

## How to test 🧪

- Test viewing the events menu and events page with a linode migration event recently added


mjac0bs (approved):
These changes do fix the issue I was seeing! Confirmed that the events drawer could be viewed without crashing during both an in-progress migration and completed migration. Thanks for your quick response on this. :) 

To document the additional context:

This bug was discovered when attempting to migrate a Linode and view the corresponding event in the event drawer. The event `linode_migrate_datacenter` includes no username. Not sure if there are other events like this, as the [API docs](https://www.linode.com/docs/api/account/#events-list__responses) suggest all should be strings.

Video of the bug (linode deleted afterwards):

https://user-images.githubusercontent.com/114685994/211448434-9144f5f7-edba-4336-85d0-5c5240950591.mov

API event with `null` username:
Screen Shot 2023-01-09 at 5 35 19 PM

cpathipa (approved):

PR #8710, opened 2023-01-11
M3-6046: dynamically generate cURL and CLI commands

CodyFinn-Akamai:
## Description 📝

Adds functions to generate Linode CLI and cURL commands to create Linodes.
This is a work in progress to just get the basics down. I want to move this to a single function or an object that can generate both cURL and Linode CLI comands.

## How to test 🧪

`yarn test generate-cURL`

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
@@ -13,6 +13,8 @@ import { makeStyles, Theme } from 'src/components/core/styles';
 import Tabs from 'src/components/core/ReachTabs';
 import TabPanels from 'src/components/core/ReachTabPanels';
 import { sendEvent } from 'src/utilities/ga';
+import generateCurlCommand from 'src/utilities/generate-cURL';
+import generateCLICommand from 'src/utilities/generate-cli';
 
 import CodeBlock from '../CodeBlock';
 
@@ -47,6 +49,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     fontSize: '14px !important',
   },
 }));
+
 export interface Props {
   isOpen: boolean;
   onClose: () => void;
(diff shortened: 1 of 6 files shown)


cpathipa:
@CodyFinn-Akamai This should go to `feature/api-cli` branch

packages/manager/src/utilities/generate-cURL.ts
```diff
@@ -0,0 +1,15 @@
...
+const headers = [
+  '-H "Content-Type: application/json"',
+  '-H "Authorization: Bearer $TOKEN"',
+].join('\n');
+
+const generateCurlCommand = (data: {}, path: string) => {
+  const command = `curl ${headers}\n-X POST -d ${JSON.stringify(
+    data
+  )}\n https://api.linode.com/v4/${path}`;
+  return command.trimStart().trimEnd();
```
carrillo-erik, line 12: I think you can simply use the `.trim()` method here instead. I believe it accomplishes the same result you are looking for.

packages/manager/src/utilities/generate-cli.ts
```diff
@@ -0,0 +1,21 @@
...
+};
+
+export const generateCLICommand = (data: {}) => {
+  const dataForCLI = Object.entries(data).map(convertJSONFieldToCLIArg);
+  return `
+linode-cli linodes create
+${dataForCLI.join('\n')}
+  `
+    .trimStart()
+    .trimEnd();
```
carrillo-erik, line 18: Same thing in this instance as my previous comment, the `.trim()` method returns a new string without the whitespaces from beginning and end of the original string.

carrillo-erik (changes requested):

carrillo-erik (approved):

cpathipa:
Hold on merging this PR until M3-6047 merges to the feature branch.

packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
```diff
@@ -62,38 +65,14 @@ const fireGAEvent = (label: string) => {
   });
 };
 const ApiAwarenessModal = (props: Props) => {
-  const { isOpen, onClose, route } = props;
+  const { isOpen, onClose, route, payLoad } = props;
 
   const classes = useStyles();
 
   // This harcoded values will be removed as part of following story that generated dynamic CURL and CLI commands.
```
cpathipa, line 72: We can remove this comment.

cpathipa (approved):
LGTM!

PR #8714, opened 2023-01-12
M3-6053: Standardize usage of dismissible banners and notices

hana-linode:
## Description 📝
- Remove unused/old banners
- Refactor Notice component to accept a dismissible button prop
- Refactor DismissibleBanner component to use the Notice component
    - This allows us to use the same NoticeProps to customize the banner instead of redefining the same props
    - Also standardizes how we use dismissible banners/notices
- Refactor ProductInformationBanner component to leverage the above
- Update Banner and Notice stories in Storybook

## How to test 🧪
Check the banners/notices changed in this PR

run `yarn storybook` and check the updated stories:
- http://localhost:6006/?path=/docs/components-notifications-dismissible-banners--example-beta-banner
- http://localhost:6006/?path=/docs/components-notifications-notices--success


packages/manager/src/components/DismissibleBanner/DismissibleBanner.stories.mdx
```diff
@@ -77,17 +124,7 @@ export const BetaBannerTemplate = () => {
   </Story>
 </Canvas>
 
-#### Beta
-
-Beta banners, along with [beta chips](/docs/elements-chips-beta-chips--default-story), provide notice to users about beta features.
```
mjac0bs, line 82: Deleting [L26](https://github.com/linode/manager/pull/8714/files#diff-3740bd49a2630ac8afd9781d64b2858a38ea88d6ed1774baf025945549bdbae8L82) and [L82](https://github.com/linode/manager/pull/8714/files#diff-3740bd49a2630ac8afd9781d64b2858a38ea88d6ed1774baf025945549bdbae8L82) seems to be reverting a change made in [M3-5101](https://github.com/linode/manager/pull/8603), which had the goal of documenting our beta components (banner + chip and the related SLA) so that we could more easily reference them whenever we next roll out a beta feature. If we want to do this clean up here, we should consider if there's a better way to document UI aspects of a beta feature using internal documentation. Although beta banners might not be in use right now, they likely will be again.

cpathipa (approved):
Nice work @hana-linode simplifying `Notice` and `DismissibleBanner`. Overall looks great! Only thing I have is A11y issue around close icon within the banner. Knowing the fact it is current behavior, feel free to address this as separate ticket if it take longer than expected. But, you have my approval though.

packages/manager/src/components/ProductInformationBanner/ProductInformationBanner.tsx
```diff
@@ -105,24 +51,10 @@ const ProductInformationBanner: React.FC<Props> = (props) => {
...
-    >
-      <>
-        {displayProductInformationWarning ? (
-          <Warning className={classes.icon} data-qa-warning-img />
-        ) : null}
-        <div className={classes.content}>
-          <HighlightedMarkdown textOrMarkdown={thisBanner.message} />
-        </div>
-      </>
+    <DismissibleBanner preferenceKey={thisBanner.key} {...rest}>
```
dwiley-akamai, line 54: Interestingly, I was trying to do some testing with mocked data and TS raised an issue about `key` not existing on the type I added to the `return` statement in `useFlags.ts`:

```
    productInformationBanners: [
      {
        bannerLocation: 'Databases',
        expirationDate: '2023-01-22',
        // eslint-disable-next-line xss/no-mixed-html
        message:
          'Test test test test testguywegfwfnksjbf <a target="_blank" href="https://www.linode.com/docs/products/storage/object-storage/">New Availability: Atlanta</a>',
      },
    ],
```

That's true for the bit of code above, of course, but I also don't recall us ever specifying `key` in LD for these banners 🤔 

In LD I made it so that you should be able to see a dummy banner for "Databases" pages locally right now. The key comes through as `undefined` if you add a `console.log(thisBanner.key)` around line 53 in `ProductInformationBanner.tsx`.

Seems like we should either make the `preferenceKey` prop optional, or be consistent about passing it in and add some logic enforcing it?

dwiley-akamai (commented):
Nice job consolidating and simplifying the various pieces we had here 👍

Are we able to make the position of the "X" consistent across banners?

Screen Shot 2023-01-17 at 2 43 19 PM

packages/manager/src/components/ProductInformationBanner/ProductInformationBanner.tsx
hana-linode: I don't think we should make the `preferenceKey` prop optional so I updated the code in `ProductInformationBanner.tsx` to pass in a key created from the bannerLocation and expirationDate.

This was an edge case since we are relying on LD to give us data. I think TS should catch the other `preferenceKey` issues during development so we shouldn't need to add additional logic to enforce it

hana-linode:
@dwiley-akamai Good catch, the `X` button alignment will be handled as part of a follow-up ticket (M3-6073) to refactor the dismissibleButton prop out of the Notice component

packages/manager/src/components/DismissibleBanner/DismissibleBanner.stories.mdx
```diff
@@ -60,6 +89,26 @@ export const BetaBannerTemplate = () => {
...
+
+#### Call to Action Banner
+
+<Canvas>
+  <Story name="Call to Action Banner">
+    {CallToActionBanner.bind({})}
+  </Story>
+</Canvas>
+
+### Component API
```
dwiley-akamai, line 110: This story needs to be updated to reflect the changes in props (`productInformationIndicator` doesn't exist anymore).

Also as a side note, can we get a description in the Component API ArgsTable for `preferenceKey`? It's marked as required but no additional context is provided.

dwiley-akamai (approved):
I have a couple of minor pieces of feedback but I'm approving pending their being addressed 👍

Maybe this will be included in the follow-up ticket you noted earlier, but there are some inconsistencies with the starting position of text:

Screen Shot 2023-01-18 at 12 54 12 PM
(the last notification is a `ProductInformationBanner`)

packages/manager/src/components/Notice/Notice.stories.mdx
```diff
@@ -16,10 +16,9 @@ import Notice from './Notice';
 
 Types of Notices:
```
mjac0bs, line 17: This is an existing typo in this story, but can we fix the last bullet point in the `Usage` section so that "Considering" -> "Consider"?

mjac0bs (approved):
Seconding Dajahi's comments on the `DismissibleBanner` story as they were also on my list to point out. Otherwise just a couple of non-blocking questions/comments. I had a look at the banners in their various locations and other than the already mentioned issues for the follow up ticket, look good and seem to be behaving as expected.

packages/manager/src/components/AbuseTicketBanner/AbuseTicketBanner.tsx
```diff
@@ -33,20 +30,20 @@ export const AbuseTicketBanner: React.FC<{}> = (_) => {
...
 
   return (
     <Grid item xs={12}>
-      <Notice important warning dismissible onClose={onDismiss}>
+      <DismissibleBanner
+        important
+        warning
+        preferenceKey={preferenceKey}
+        options={{
+          expiry: DateTime.utc().plus({ days: 7 }).toISO(),
```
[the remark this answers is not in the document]
hana-linode: No standard afaik, a week was just a random number I picked for `AbuseTicketBanner` haha

packages/manager/src/components/DismissibleBanner/DismissibleBanner.stories.mdx
```diff
@@ -77,17 +126,7 @@ export const BetaBannerTemplate = () => {
...
-<Canvas>
-  <Story name="Beta Banners">
-    {BetaBannerTemplate.bind({})}
-  </Story>
-</Canvas>
-
-<ArgsTable story="Dismissible Banners" exclude={['className']} />
+<ArgsTable story="Dismissible Banners" include={['preferenceKey', 'children', 'options', 'text', 'success', 'error', 'warning', 'important', ]} />
 
 <!-- The Status Banners story below can be restored once the MSW works in production builds of Storybook. --!>
```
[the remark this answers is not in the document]
hana-linode: Good question, not sure if there is an existing ticket for this

PR #8718, opened 2023-01-12
M3-6069: Fix NodeBalanacer Mode Change Select

bnussman-akamai:
## Description 📝

- Fixes NodeBalanacer Mode Change Select
- Fixed by using function parameters rather than relying on the event

It was caused by this error
Screen Shot 2023-01-12 at 6 24 37 PM

## How to test 🧪

- Test changing the `Mode` of a nodebalancer node
- Verify you can change the mode and save it successfully

### DIFF
--- a/packages/manager/src/features/NodeBalancers/NodeBalancerConfigPanel.tsx
+++ b/packages/manager/src/features/NodeBalancers/NodeBalancerConfigPanel.tsx
@@ -201,11 +201,11 @@ class NodeBalancerConfigPanel extends React.Component<CombinedProps> {
     }
   };
 
-  onNodeModeChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-    const nodeIdx = e.currentTarget.getAttribute(DATA_NODE);
-    if (nodeIdx) {
-      this.props.onNodeModeChange!(+nodeIdx, e.target.value);
-    }
+  onNodeModeChange = (
+    e: React.ChangeEvent<HTMLInputElement>,
+    nodeIdx: number
+  ) => {
+    this.props.onNodeModeChange!(nodeIdx, e.target.value);
   };
 
   addNode = () => {
--- a/packages/manager/src/components/Dialog/Dialog.tsx
+++ b/packages/manager/src/components/Dialog/Dialog.tsx
@@ -9,6 +9,7 @@ import Typography from 'src/components/core/Typography';
 import Grid from 'src/components/Grid';
 import { convertForAria } from 'src/components/TabLink/TabLink';
 import Notice from 'src/components/Notice';
+import classNames from 'classnames';
 
 export interface DialogProps extends _DialogProps {
   className?: string;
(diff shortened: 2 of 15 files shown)


mjac0bs (dismissed):
No longer seeing that error in the console on change of Mode selection and confirmed that I could successfully save my changes to a Nodebalancer's  Backend Node "Mode".

Yer a wizard, @bnussman-akamai. 🧙‍♂️

jcallahan-akamai (commented):
This seems to fix the problem. Do you know where this bug was introduced? And do you know why the component behaves different than other inputs? (this one applies opacity to the rest of the screen and makes it look focused). That's why I thought the regression could have been introduced with the MUI upgrade – I thought these things might be related – but maybe not.

bnussman-akamai:
@jcallahan-akamai The original implementation is pretty questionable. It is using a `<TextField />` with the `select` prop where a `<Select />` should really be used. Both `<TextField />` where a `<Select />`  are MUI components but in most places in cloud, we are using `react-select` components. I believe what broke here was MUI changing the underlying implementation of the `<TextField />`component.

jcallahan-akamai (dismissed):
@bnussman thanks for the explanation and the fix. We can come back and address this and other things on the NB config form.

packages/manager/src/features/NodeBalancers/NodeBalancerConfigPanel.tsx
```diff
@@ -201,11 +201,11 @@ class NodeBalancerConfigPanel extends React.Component<CombinedProps> {
...
-  onNodeModeChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-    const nodeIdx = e.currentTarget.getAttribute(DATA_NODE);
-    if (nodeIdx) {
-      this.props.onNodeModeChange!(+nodeIdx, e.target.value);
-    }
+  onNodeModeChange = (
+    e: React.ChangeEvent<HTMLInputElement>,
+    nodeIdx: number
+  ) => {
+    this.props.onNodeModeChange!(nodeIdx, e.target.value);
```
hkhalil-akamai, line 208: Would it make sense to make this prop required so the `!` assertion wouldn't be necessary?

dwiley-akamai: looks like it can't be made required at the moment because there's an instance of this component being used where `onNodeModeChange` is not a prop that is passed to it (see `NodeBalancerCreate.tsx`)

dwiley-akamai (approved):

packages/manager/src/features/NodeBalancers/NodeBalancerConfigPanel.tsx
hkhalil-akamai: Ah gotcha, I agree that it's not in the scope of the PR. But ideally I'd like to make "invalid state impossible to represent". For example, it shouldn't be possible to set `forEdit` to true but not provide a `onNodeModeChange`

CodyFinn-Akamai (dismissed):

PR #8725, opened 2023-01-13
M3 6047: Code block component Part 1 & 2

cpathipa:
## Description 📝

- Code block component
- Renders cURL and CLI commands as per the design

## Preview 📷
image

## How to test 🧪

- Copy the cURL command and execute it command line (Need to update token before executing)
- Copy CLI command and execute it command line
- Note: For the scope of this story commands are hardcoded. 

`yarn cy:run`


packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
```diff
@@ -121,6 +160,31 @@ const ApiAwarenessModal = (props: Props) => {
...
+            <CodeBlock command={cliCommand} language={'bash'} />
+            <Typography className={classes.guides} variant="h2">
+              Guides
+            </Typography>
+            <Typography>
+              <ExternalLink
+                text="Get Started with the Linode API"
+                link="https://www.linode.com/docs/products/tools/api/get-started/"
+                onClick={() =>
+                  fireGAEvent('Click: Get Started with the Linode API')
```
CodyFinn-Akamai, line 175: This probably needs to be documented better but the fields for GA events are supposed to represent the following:
category: Where in the application this event is coming from
action: Type of interaction and type of contextual help
label: The text displayed on the textual help to the user

So in this case I think the GA event would look something like:
```
{
  category: 'Linode Create API CLI Awareness Modal',
  action: 'Click:link',
  label: 'Get Started with the Linode API'
}
```

packages/manager/src/features/linodes/LinodesCreate/CodeBlock/styles.ts
```diff
@@ -0,0 +1,32 @@
...
+        color: '#e6db74',
+      },
+      '& .hljs-symbol': {
+        color: '#f8f8f2',
+      },
+    },
+  },
+  copyIcon: {
+    position: 'absolute',
+    top: '8px',
```
carrillo-erik, line 26: Changing this to `30px` fixes the UI defect where the copy icon is on top of the command text.

carrillo-erik (changes requested):

packages/manager/src/features/linodes/LinodesCreate/CodeBlock/styles.ts
cpathipa: Actually this causing by hardcoded command using template literals. Once command is generated dynamically headers will be printed in multiple lines.

packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
cpathipa: Thank you @CodyFinn-Akamai for guiding on this.

packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
```diff
@@ -37,25 +47,54 @@ const useStyles = makeStyles((theme: Theme) => ({
...
   route: string;
+  payLoad: CreateLinodeRequest;
 }
 
-const fireGAEvent = (action: string) => {
+const fireGAEvent = (action: string, label: string) => {
   sendEvent({
-    action,
-    category: 'API CLI Awareness Contextual Help',
+    action: `Click:${action}`,
```
CodyFinn-Akamai, line 59: Sorry I was unclear in the last comment. The format for the action property should be:

<ActionTheUserDid>:<TypeOfContextualHelp>

So in this case it would be the string literal `'Click:link'` since the user clicked on a link. If the user hovered over a tooltip the proper action property would be the string literal `'Hover:tooltip'`

packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
```diff
@@ -37,25 +47,54 @@ const useStyles = makeStyles((theme: Theme) => ({
...
   route: string;
+  payLoad: CreateLinodeRequest;
 }
 
-const fireGAEvent = (action: string) => {
+const fireGAEvent = (label: string) => {
   sendEvent({
-    action,
-    category: 'API CLI Awareness Contextual Help',
+    action: 'Click:Link',
```
CodyFinn-Akamai, line 59: It should be `'Click:link'` to be consistent with our other usage

CodyFinn-Akamai (changes requested):

CodyFinn-Akamai (approved):

packages/manager/package.json
```diff
@@ -52,6 +52,7 @@
     "formik": "~2.1.3",
     "he": "^1.2.0",
     "highlight.js": "~10.4.1",
+    "highlightjs-curl": "1.3.0",
```
bnussman-akamai, line 55:
```suggestion
```

I think we can completely remove `highlightjs-curl` and references to curl. cURL is not necessarily even a language. If we just use `language: "bash"` we get good enough syntax highlighting for cURL

packages/manager/src/components/HighlightedMarkdown/HighlightedMarkdown.tsx
```diff
@@ -82,7 +86,13 @@ export const HighlightedMarkdown = (props: HighlightedMarkdownProps) => {
...
   return (
     <Typography
-      className={classNames(classes.root, 'formatted-text', className)}
+      className={classNames(
+        {
+          [classes.root]: true,
+          'formatted-text': true,
+        },
+        className
+      )}
```
bnussman-akamai, line 95: I believe that this works the same and is more simple

```suggestion
      className={classNames(classes.root, 'formatted-text', className)}
```

bnussman-akamai (changes requested):

bnussman-akamai (approved):

packages/manager/src/components/HighlightedMarkdown/highlight.d.ts
```diff
@@ -4,3 +4,4 @@ declare module 'highlight.js/lib/languages/javascript';
 declare module 'highlight.js/lib/languages/yaml';
 declare module 'highlight.js/lib/languages/bash';
 declare module 'highlight.js/lib/languages/nginx';
+declare module 'highlightjs-curl';
```
bnussman-akamai, line 7:
```suggestion
```

PR #8727, opened 2023-01-17
Fix Linodes Landing Extra Padding

bnussman-akamai:
## Description 📝

- The Linodes landing page had extra top and bottom padding that other landing pages did not.
- This has been bugging me for quite some time so I had to put up a fix 😖
- The solution was to just remove a few useless `<Grid />`s 
- The diff may look crazy, but that's just because I removed so many `<Grid />`s  causing less indentation in the file

### Before

Screen Shot 2023-01-17 at 11 00 20 AM

### After

Screen Shot 2023-01-17 at 11 01 48 AM

## How to test 🧪

- Test layout, ui, and responsiveness of the Linodes Landing table (`http://localhost:3000/linodes`)
- Verify that the padding matches other landing pages like Volumes


jdamore-linode (approved):
Nice catch Banks!

dwiley-akamai (approved):
LGTM 🚢

PR #8703, opened 2023-01-10
M3 6045: Add modal content, tabs and validation check

cpathipa:
## Description 📝
Show the API awareness modal after validation check passes.

**What does this PR do?**

- Adds content and respective links to the API awareness modal.
- Adds Tabs component.
- Add installation guides.
- Wired GA events
-  Validation check to show / hide modal

## Preview 📷
image

## How to test 🧪

- Navigate to Create Linode page
- Click on "Create using command line" button - Try with out filling required data.
- Should show the inline error messages.
- Enter data for all required fields.
- Click on "Create using command line"
- Show show the API awareness modal.
- Validate all the links. 

**How do I run relevant unit or e2e tests?**
yarn test

PR #8729, opened 2023-01-18
M3-6067: Linode Create - API/CLI Copy Changes

bnussman-akamai:
## Description 📝

- Updates new API/CLI modal with new copy on the Linode Create Flow
- **To verify copy check the internal ticket**

## Preview 📷

Screen Shot 2023-01-18 at 3 12 47 PM

## How to test 🧪

- Check copy and links in the modal on the Linode Create Flow
- **To verify copy check the internal ticket**

### DIFF
--- a/packages/manager/src/components/Notice/Notice.tsx
+++ b/packages/manager/src/components/Notice/Notice.tsx
@@ -111,6 +111,9 @@ export const useStyles = makeStyles((theme: Theme) => ({
   successList: {
     borderLeft: `5px solid ${theme.palette.status.successDark}`,
   },
+  marketing: {
+    borderLeft: `5px solid ${theme.color.green}`,
+  },
   flag: {
     marginRight: theme.spacing(2),
   },
@@ -133,6 +136,7 @@ interface Props extends GridProps {
   important?: boolean;
   warning?: boolean;
   success?: boolean;
+  marketing?: boolean;
   typeProps?: TypographyProps;
   className?: string;
   flag?: boolean;
@@ -161,6 +165,7 @@ const Notice: React.FC<CombinedProps> = (props) => {
     errorGroup,
     warning,
     success,
+    marketing,
     typeProps,
     children,
     flag,
@@ -228,6 +233,7 @@ const Notice: React.FC<CombinedProps> = (props) => {
         [classes.success]: success && !notificationList,
         [classes.successList]: success && notificationList,
         [classes.warning]: warning && !notificationList,
+        [classes.marketing]: marketing,
         [classes.warningList]: warning && notificationList,
         [classes.informational]: informational && !notificationList,
         [classes.informationalList]: informational && notificationList,
--- a/packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
@@ -125,11 +125,13 @@ const ApiAwarenessModal = (props: Props) => {
           <SafeTabPanel index={1}>
             <Typography variant="body1">
               Before running the command below, the Linode CLI needs to be
-              installed and configured on your system. See the Install and{' '}
+              installed and configured on your system. See the{' '}
               <ExternalLink
-                text="Configure the Linode CLI"
+                text="Install and Configure the Linode CLI"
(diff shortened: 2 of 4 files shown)


packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
```diff
@@ -76,36 +77,13 @@ const ApiAwarenessModal = (props: Props) => {
...
-        />{' '}
-        and Linodes using the{' '}
-        <ExternalLink
-          onClick={() => fireGAEvent('Click: Linode API Link')}
-          link="https://www.linode.com/docs/api/"
-          text="Linode API"
-          hideIcon
-        />
-        .
+        Create a Linode in the command line using either cURL or the Linode CLI
```
mjac0bs, line 82: This is a nitpick, but "Create a Linode in the command line" sounds strange to me. It matches the approved copy on the internal ticket, but it's the word "in" that is throwing me off -- not sure how common it is to hear about doing something "in the command line" versus "via the command line" or "using the command line". (I see why we'd want to avoid "using" since we'd then have "using the command line using", though.) Not sure if others have thoughts on this?

packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
```diff
@@ -114,49 +92,83 @@ const ApiAwarenessModal = (props: Props) => {
...
+          </SafeTabPanel>
+          <SafeTabPanel index={1}>
+            <Typography variant="body1">
+              Before running the command below, the Linode CLI needs to be
+              installed and configured on your system. See the Install and{' '}
               <ExternalLink
-                onClick={() => fireGAEvent('Click: Using The Linode CLI Link')}
-                link="https://www.linode.com/docs/products/tools/cli/get-started/#basic-usage"
-                text="Using the Linode CLI"
+                text="Configure the Linode CLI"
```
mjac0bs, line 130: Our internal ticket specifies that "Install and Configure the Linode CLI" is the link text, but it doesn't look like "Install and" is included in the hyperlink here.

mjac0bs (commented):
Looking pretty good! Leaving a couple of comments - found one discrepancy in a link with internal ticket copy and this branch's copy. I also noticed that the mock prototype had links underlined by default, not just when they're hovered over, but I'm inclined to think it wasn't an intended choice in the mocks?

packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
bnussman-akamai: Oops. Thanks for catching that

packages/manager/src/components/Notice/Notice.tsx
```diff
@@ -111,6 +111,9 @@ export const useStyles = makeStyles((theme: Theme) => ({
   successList: {
     borderLeft: `5px solid ${theme.palette.status.successDark}`,
   },
+  marketing: {
+    borderLeft: `5px solid ${theme.color.green}`,
+  },
```
cpathipa, line 116: I was under same impression,And added `green` color in earlier [PR](https://github.com/linode/manager/pull/8703#discussion_r1068617698). But, realized that we no longer use green for marketing messages instead we use blue.

packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
dwiley-akamai: "via the command line" would be the better option from my POV.

dwiley-akamai (approved):
Approving pending a final decision on the "via the command line" language and updating the banner `borderLeft` color to blue -- https://github.com/linode/manager/pull/8714 was merged so perhaps a rebase of the feature branch on top of `develop` or a merging in of `develop` to the feature branch can be done as part of that update to pull in the changes to the `<Notice />` component.

cpathipa (approved):

packages/manager/src/components/Notice/Notice.tsx
cpathipa: Marking it as resolved, As we need confirmation on color from UX team.

cpathipa:
Merging the PR after discussing with @bnussman

PR #6995, opened 2020-10-13
Remove flag-based dynamic password validation logic

Jskobos:
## Description

I think our zxcvbn validation (which matches the API's) is stable enough now to remove a bunch of conditional logic that's reading a value from LD.

Please check all of our password fields (Linode creation, StackScripts, disk creation, managed credentials) to make sure we're validating as expected.

PR #8733, opened 2023-01-19
refactor: [M3-6070] - Increase minimum acceptable password strength

dwiley-akamai:
## Description 📝

- Increase `MINIMUM_PASSWORD_STRENGTH` from 2 to 4
- Refactors in `StrengthIndicator.tsx` to bring it up-to-date with our modern code conventions
- Add unit tests for `StrengthIndicator.tsx`
- Don't scale scores of 4 down in `PasswordInput.tsx`
- Enforce password complexity/requirements in `LinodeSettingsPasswordPanel.tsx`

`Do Not Merge` until the 3/6 release process begins.

## How to test 🧪
- Ensure that the strength indicators for password inputs throughout the app have not been adversely impacted.
- Run `yarn test strengthindicator`

### DIFF
--- a/packages/manager/src/components/PasswordInput/StrengthIndicator.tsx
+++ b/packages/manager/src/components/PasswordInput/StrengthIndicator.tsx
@@ -64,7 +64,7 @@ const StrengthIndicator: React.FC<CombinedProps> = (props: Props) => {
           <div
             className={classNames({
               [classes.block]: true,
-              [`strength-${strength}`]: !isNil(strength) && idx <= strength,
+              [`strength-${strength}`]: !isNil(strength) && idx < strength,
             })}
           />
         </Grid>
@@ -94,8 +94,9 @@ export const convertStrengthScore = (strength: StrengthValues) => {
     case 1:
       return ' Weak';
     case 2:
-      return ' Fair';
     case 3:
+      return ' Fair';
+    case 4:
       return ' Good';
     default:
       return ' Weak';
--- a/packages/manager/src/components/PasswordInput/StrengthIndicator.tsx
+++ b/packages/manager/src/components/PasswordInput/StrengthIndicator.tsx
@@ -1,5 +1,4 @@
 import classNames from 'classnames';
-import { isNil } from 'ramda';
 import * as React from 'react';
 import { makeStyles, Theme } from 'src/components/core/styles';
 import Typography from 'src/components/core/Typography';
@@ -44,9 +43,7 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
 }));
 
-type CombinedProps = Props;
-
-const StrengthIndicator: React.FC<CombinedProps> = (props: Props) => {
+const StrengthIndicator = (props: Props) => {
   const classes = useStyles();
 
   const { strength, hideStrengthLabel } = props;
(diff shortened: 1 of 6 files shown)


mjac0bs (commented):
Unit test ✅ 

Tested Linode root password input with strength indicators in the Linode Create flow and Linode Details page; Weak and Fair passwords are rejected and Good passwords are accepted, as expected. 👍 

Since users will need to start providing even more complex passwords, the error displayed on attempted submission of weak and fair passwords could ideally be more descriptive. We don't inform users what passwords meet complexity requirements. Is this lack of notice about password requirements a limitation of our method for assessing password complexity? If not, a follow up ticket to look into improving this experience would be helpful.

image

dwiley-akamai:
@mjac0bs Back when we weren't using zxcvbn a few years ago, we had logic displaying info re: password requirements (see `PasswordInput.tsx` [here](https://github.com/linode/manager/pull/6995/files)). However, [zxcvbn](https://github.com/dropbox/zxcvbn) avoids arbitrary password requirement rules for the most part:

`[...] it is more secure, flexible, and usable when sites require a minimal complexity score in place of annoying rules like "passwords must contain three of {lower, upper, numbers, symbols}".`

mjac0bs:
@dwiley-akamai Thanks for the historical context! Interesting, sounds like lack of rules are a zxcvbn feature rather than a limitation. I am curious what users think about it, but that sounds like a question to raise with UX.

mjac0bs (approved):

packages/manager/src/components/PasswordInput/StrengthIndicator.tsx
```diff
@@ -1,66 +1,55 @@
...
-    },
-  });
+  },
+  blockOuter: {
+    padding: '4px !important' as '4px',
+  },
+}));
 
-const styled = withStyles(styles);
+type CombinedProps = Props;
```
bnussman-akamai, line 47:
```suggestion
```

packages/manager/src/components/PasswordInput/StrengthIndicator.tsx
```diff
@@ -1,66 +1,55 @@
...
 
-type CombinedProps = Props & WithStyles<ClassNames>;
+const StrengthIndicator: React.FC<CombinedProps> = (props: Props) => {
```
bnussman-akamai, line 49:
```suggestion
const StrengthIndicator = (props: Props) => {
```

packages/manager/src/components/PasswordInput/StrengthIndicator.tsx
```diff
@@ -89,15 +78,43 @@ const StrengthIndicator: React.FC<CombinedProps> = (props) => {
...
         </Typography>
       </Grid>
     </Grid>
   );
 };
 
-export default styled(StrengthIndicator);
+export default StrengthIndicator;
+
+const scaledStrength = (strength: number) => {
```
dwiley-akamai, line 90: We had a piece of this logic in `maybeStrength()` in `PasswordInput.tsx` previously to scale `zxcvbn` scores of 4 down to 3. We don't want to do any actual scaling down of scores, but because we have 3 strength bars but 5 possible strength scores, we do need to scale the latter strictly for UI purposes so that 1 bar corresponds to `Weak`, 2 bars correspond to `Fair`, and 3 bars correspond to `Good`.

bnussman-akamai (approved):

Issue #7483, opened 2021-03-09
v1.34.0 breaks screen reader accessibility in firewall rules tables


tech10:
Version 1.34.0 prevents a screen reader from properly reading and navigating the firewall rules tables. My specific use case is on NVDA using Google Chrome, though I believe the screen reader and web browser being used would be irrelevant to this particular issue.

To reproduce this, create a firewall and add some rules, either inbound or outbound should work for this. Using a screen reader, the only buttons accessible when arrowing through the tables will be labeled as View Details. Tabbing through the elements does seem to provide a bit more information and a few more buttons, which can't be activated with the screen reader, and the rule being viewed isn't visible to the screen reader.

With the firewall rules interface as it currently is, this makes things far more limited in their use, so far as accessible navigation with a screen reader goes.

tech10:
@tiffwong Since firewalls are now out of beta, I thought I would see if there are any updates to fix this issue soon. Using the Linode CLI to edit firewall rules isn't always desired over the cloud manager, and as it is, it makes firewalls look unattractive to screen reader users with the lack of accessibility.

tiffwong:
@tech10 Thanks so much for bringing this up again, we're prioritizing this and hope to have a solution soon!

tech10:
#8735 does a good job of resolving this. It's definitely useable now, though the layout does look a bit different than the rest of the cloud manager. Also, when moving through the list with arrows using NVDA, in browse mode, the screen reader does tend to state the instructions for dragging continually with every option I move through.

Though it is useable as it is, it isn't perhaps as intuitive as it could be, since you have to enter focus mode before you're able to do anything, with NVDA, at least. This may not be quite as useable for some people as it is for me.

If this is to be the standard layout for the Firewall section, this issue can be closed. If not, and there's more work to be done to improve accessibility with screen readers, I would recommend keeping this issue open, for the time being. I would welcome any questions regarding possible accessibility improvements here.

mjac0bs:
@tech10 Thank you for letting us know that recent changes resolved the issues you were encountering. We also appreciate your feedback on the areas where there is still room for improvement. Some of the unintuitive behavior you have mentioned is built into the library we use for drag and drop functionality; while we don't have further improvements at this time, it's possible an update to that library could help us improve accessibility in the future.

[...]

PR #8735, opened 2023-01-19
M3-6039: Update FirewallRuleTable to support screen readers

mjac0bs:
## Description 📝

**What does this PR do?**
Refactors `FirewallRuleTable.tsx` to support screen readers by converting the underlying structure from a table into an unordered list arranged in a `Grid` to [improve compatibility with react-beautiful-dnd](https://github.com/atlassian/react-beautiful-dnd/issues/2233). Added `aria-label`s and tab navigability where necessary to improve access to fields within a list item. 

Previously, the FirewallRuleTable was [inaccessible for users using screen readers and keyboard navigation](https://github.com/linode/manager/issues/7483#event-4432711610). The table did not always provide enough context for the user to know how to interact with drag & drop functionality or interact with the FirewallRuleAccessMenu buttons, and the fields in each table row were unknown to the user due to lack of focusability.

Detailed list of issue -> fixed behavior for each screen reader:

Both screen readers:
- No label name read for firewall rule row, just 'View details' ➡️ Screen reader reads firewall rule label when rule is focused
- ~No keyboard access to fields in firewall rule: Protocols, Port Range, Sources, Actions ➡️  Keyboard navigate to fields in a firewall rule when rule is focused~ - removed tab focus and made M3-6154; users can still view details about a rule by editing it in the meantime
- No table header focus - keyboard navigation jumped to first rule row with no context for the table ➡️ Keyboard navigate within list header group to read column headers

VoiceOver:
- No instructions about how to interact with drag & drop functionality read by screen reader ➡️ Screen reader reads drag & drop instructions when rule is focused

NVDA:
- Screen reader reads drag and drop instructions, but requires users to toggle focus mode before a firewall rule can be selected or any action buttons pressed. Screen reader instructions are confusing and misleading about dragging a row. ➡️ Screen reader still requires that user enter focus mode to activate a row... but an `aria-roledescription` added to the list item widget, mentioned in the [react-beautiful-dnd screen reader guide](https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/guides/screen-reader.md), provides more context about how to interact with the draggable option.
 
## Preview 📷
<details>
<summary>
Screencaps with screen readers
</summary>
Prod - Mac/Chrome/VoiceOver:

https://user-images.githubusercontent.com/114685994/214673727-4a6b3e6a-124c-414e-8413-42fa125eadca.mov

This branch - Mac/Chrome/VoiceOver:

https://user-images.githubusercontent.com/114685994/214680631-9e482baa-3515-4bd0-a0f3-cb30bf4cb767.mov

Prod - Windows/Chrome/NVDA:

https://user-images.githubusercontent.com/114685994/214682408-555e0587-d508-44d4-a64d-a9b3ca18a3e0.mov

This branch - Windows/Chrome/NVDA:

https://user-images.githubusercontent.com/114685994/214682924-05875a6b-e180-41c3-92bd-a5f11d5012b0.mov
</details>

## How to test 🧪

These changes need to be tested with multiple screen readers (VoiceOver, NVDA) and ideally on multiple browsers (Chrome, Firefox, Safari) to ensure the FirewallRuleTable functionality is accessible. 

VoiceOver
1. Turn on VoiceOver. Shortcut keys are `Command` + `F5` to toggle the screen reader. You can also find it in Mac settings under Accessibility.

NVDA:
1. Use a Windows machine installed with NVDA you have access to  _or_  set up a Windows emulator and install NVDA using our internal documentation on "Accessibility Testing with Virtual Machines and Screen Readers". (Ask me if you don't know where to find this.)
2. Turn on NVDA. Shortcut keys are `Control` + `Alt` + `N`.

With both screen readers:
1. Navigate to the Firewall Landing page.
2. Create a new firewall or use an existing one. Click on the firewall to view the Firewall Detail page.
3. Use the screen reader and keyboard navigation to navigate to either the inbound or outbound rule button. Create at least two new firewall rules with distinct labels (easiest way is to just set different presets, e.g. `SSH` and `HTTPS`). 
4. Navigate using the screen reader + keyboard to the appropriate firewall rules table (list).
5. Confirm that the above list of fixed behaviors are present using each screen reader. Firewall rules should be draggable and droppable, should be able to be edited, deleted, cloned, and reordered. Changes should persist as long as the changes are saved.
6. Confirm that the styling of the table and rules matches the styling in production.

```yarn test FirewallTableRule```

### DIFF
--- a/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleTable.tsx
+++ b/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleTable.tsx
@@ -54,8 +54,17 @@ const useStyles = makeStyles((theme: Theme) => ({
     backgroundColor: 'transparent',
     border: 'none',
   },
+  unmodified: {
+    backgroundColor: '#FFFFFF',
+  },
   highlight: {
     backgroundColor: theme.bg.lightBlue1,
+  },
+  disabled: {
+    backgroundColor: 'rgba(247, 247, 247, 0.25)',
+    '& td': {
+      color: '#D2D3D4',
+    },
   },
   error: {
     '& p': { color: theme.color.red },
@@ -72,6 +81,9 @@ const useStyles = makeStyles((theme: Theme) => ({
   labelCol: {
     paddingLeft: 6,
     whiteSpace: 'nowrap',
+  },
+  ruleGrid: {
+    spacing: 0,
   },
   ruleList: {
     backgroundColor: theme.color.border3,
@@ -92,7 +104,6 @@ const useStyles = makeStyles((theme: Theme) => ({
   // },
   ruleRow: {
     margin: 0,
-    background: '#FFFFFF',
     borderBottom: '1px solid #F4F5F6',
     alignItems: 'center',
     rowSpacing: 0.5,
@@ -228,7 +239,7 @@ const FirewallRuleTable: React.FC<CombinedProps> = (props) => {
       <Grid
         container
         aria-label={`${category} Rules List`}
-        tabIndex={0}
+        // tabIndex={0}
         style={{ margin: 0 }}
       >
         <Grid
@@ -353,7 +364,6 @@ type FirewallRuleTableRowProps = RuleRow &
     innerRef: any;
     isDragging: boolean;
     disabled: boolean;
-    highlight?: boolean | undefined;
   };
 
 const FirewallRuleTableRow: React.FC<FirewallRuleTableRowProps> = React.memo(
@@ -364,7 +374,6 @@ const FirewallRuleTableRow: React.FC<FirewallRuleTableRowProps> = React.memo(
       id,
       action,
       label,
-      // description,
       protocol,
       ports,
       addresses,
@@ -374,10 +383,10 @@ const FirewallRuleTableRow: React.FC<FirewallRuleTableRowProps> = React.memo(
       triggerOpenRuleDrawerForEditing,
       triggerUndo,
       errors,
-      // innerRef,
+      innerRef,
       // isDragging,
       disabled,
-      // originalIndex,
+      originalIndex,
       // ...rest
     } = props;
 
@@ -393,17 +402,19 @@ const FirewallRuleTableRow: React.FC<FirewallRuleTableRowProps> = React.memo(
       <Grid
         container
         key={id}
-        // highlight={
-        //   // Highlight the row if it's been modified or reordered. ID is the
-        //   // current index, so if it doesn't match the original index we know
-        //   // that the rule has been moved.
-        //   status === 'MODIFIED' || status === 'NEW' || originalIndex !== id
-        // }
-        // disabled={status === 'PENDING_DELETION' || disabled}
-        // domRef={innerRef}
-        className={classes.ruleRow}
+        ref={innerRef}
         aria-label={label}
         role="option"
+        className={classNames({
+          [classes.ruleRow]: true,
+          // Highlight the row if it's been modified or reordered. ID is the
+          // current index, so if it doesn't match the original index we know
+          // that the rule has been moved.
+          [classes.unmodified]: status === 'NOT_MODIFIED',
+          [classes.highlight]:
+            status === 'MODIFIED' || status === 'NEW' || originalIndex !== id,
+          [classes.disabled]: status === 'PENDING_DELETION' || disabled,
+        })}
       >
         <Grid
           item
--- a/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleTable.tsx
+++ b/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleTable.tsx
@@ -82,9 +82,6 @@ const useStyles = makeStyles((theme: Theme) => ({
   table: {
     backgroundColor: theme.color.border3,
   },
-  ruleList: {
-    listStyle: 'none',
-  },
   addLabelButton: {
     ...theme.applyLinkStyles,
   },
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleTable.tsx [hunk 1]
```diff
@@ -334,51 +411,85 @@ const FirewallRuleTableRow: React.FC<FirewallRuleTableRowProps> = React.memo(
...
+          )}{' '}
+        </Grid>
         <Hidden lgDown>
-          <TableCell>
+          <Grid
+            item
+            style={{ width: '10%' }}
+            tabIndex={0}
+            aria-label={`Protocol: ${protocol}`}
+          >
```
mjac0bs, line 459: Setting tab index to 0 is not great to make a rule's subitems accessible, since accessibility standards are to make only interactive elements focusable. A [roving tab index (-1)](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_roving_tabindex) that is programmatically accessible is probably the way to go here... just haven't succeeded in achieving that with a `ref` and `.focus`. Thoughts/suggestions welcome.

packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleTable.tsx
```diff
@@ -436,11 +547,15 @@ export const PolicyRow: React.FC<PolicyRowProps> = React.memo((props) => {
...
 
   return (
-    <TableRow className={classes.policyRow}>
-      <TableCell colSpan={colSpan} className={classes.policyText}>
+    <Grid
+      container
+      className={classNames(classes.policyRow, classes.unmodified)}
+      tabIndex={0}
+    >
+      <Grid item /*xs={colSpan}*/ className={classes.policyText}>
```
mjac0bs, line 555: Still trying to fix a couple layout issues here.

packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleTable.tsx
```diff
@@ -183,6 +222,9 @@ const FirewallRuleTable: React.FC<CombinedProps> = (props) => {
 
   const zeroRulesMessage = `No ${category} rules have been added.`;
 
+  const screenReaderMessage =
```
jdamore-linode, line 225: Any idea if we can make this message less specific to NVDA without reducing its usefulness? I'm hearing this message every time I focus any firewall rule row using any screen reader, so I could see it becoming a bit much for users who aren't using NVDA.

This may come down to my NVDA config (which is largely unchanged from the default) or user error, but it doesn't seem like I even need to enter focus mode explicitly in order to drag+drop or otherwise interact with a row when using NVDA+Chrome or Firefox -- In the case of drag+drop, I can simply focus on the row I want, hit space to initiate drag and drop, use the arrow keys to select my desired position, and then hit space again to finalize (and this seems to be the case with VoiceOver and Windows Narrator, and ORCA is also similar except I have to use shift + arrow keys to rearrange items after initiating the drag+drop operation).

jdamore-linode (commented):
Repeated your steps under these setups:

- VoiceOver + Chrome
- NVDA + Chrome
- NVDA + Firefox
- Windows Narrator + Chrome
- ORCA + Firefox

In every case I was able to interact with the rule row (the rule label was announced by the screen reader, and I was able to focus and activate the "Undo", "Edit", "Clone", "Delete" buttons, as well as reorder the rules).

Full disclosure: Even though I can tell clear as day that this branch improves upon the Production experience, I was not able to reproduce all of the issues described in issue #7483 in Prod -- for example, I was able to activate the table rule "Edit", "Clone", and "Delete" buttons in Prod, unlike the user who submitted the issue. This might be something that has been improved since the issue was initially opened, but I'm just pointing this out in case there's an element of user error on my part in play here.

I'll circle back and approve this once those minor layout issues are resolved!

packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleTable.tsx
mjac0bs: > Any idea if we can make this message less specific to NVDA without reducing its usefulness?

Yeah, this is definitely a fair question. It would be ideal if *only* NVDA users heard the message, but I'm not seeing a way to be able to do that. To make it less specific to NVDA, the message could be something like "Interact with this list item by pressing spacebar to begin a drag or tab to access item actions. If the item is not focused, first press Insert + Space to enter Focus Mode." Less specific, but still feels long. I did try to [overwrite](https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/guides/screen-reader.md#drag-starting) the `onDragStart` lifecycle announcement to try to customize the screen reader language without adding an aria tag, but couldn't seem to get a custom announcement to read out. 

> it doesn't seem like I even need to enter focus mode explicitly in order to drag+drop or otherwise interact with a row when using NVDA+Chrome or Firefox

I haven't seen this behavior! Do you have a screencap of it happening? When I've accessed the rules table, there is  no guarantee that the screen reader is already in focus mode, so it's interesting that your screen reader seems to be.

carrillo-erik:
Something I noticed is that NVDA screen reader reads out "blank, Option TCP, 1 of 4 available" and as you move down the dropdown menu it will read out loud "blank, Option UDP, 2 of 4 available". That initial "blank" did not occur on Chrome in MacOS.

carrillo-erik (approved):

jdamore-linode:
> Something I noticed is that NVDA screen reader reads out "blank, Option TCP, 1 of 4 available"

Echoing this, also hearing "blank" when navigating the drop-down when creating a firewall rule using NVDA. Reiterating what we've discussed in Slack, disregard everything I said about NVDA above; got a working installation and everything worked exactly as you described. I do like the idea of making that blurb more generic (maybe even something like "some screen readers may require you to enter focus mode to interact with firewall rule list items") because I figure that would cover any other hypothetical screen reader that has a focus/browse mode mechanism, and still be applicable in cases where the user remaps the key sequence to something other than NVDA+Space (which I discovered is possible, btw!)

mjac0bs:
> > Something I noticed is that NVDA screen reader reads out "blank, Option TCP, 1 of 4 available"
> 
> Echoing this, also hearing "blank" when navigating the drop-down when creating a firewall rule using NVDA.

In the interest of keeping this PR focused on the firewall rules table and not the form fields in the drawers, I went ahead and noted this in new ticket M3-6156 since this behavior exists in production with Windows/NVDA and its scope seems to be every select dropdown for both the Add and Edit rule drawers. Seems like another instance of NVDA-specific interpretation of form labels/fields we'd want to look into a fix for.

I also updated the screen reader message on the firewall list item to be a bit more generic -- @jdamore-linode - good call that it didn't account for custom mappings.

Thanks to you both for your help in testing this; I know emulators can be temperamental.

PR #8716, opened 2023-01-12
M3-6066: Fix `api-v4` and `validation` Node Errors

bnussman-akamai:
## Description 📝

- Uses a default export of the `ipaddr.js` package to prevent node errors. We are thinking that `ipaddr.js` is lacking support for our build system.

- Fixes
  - #8713
  - #8706
  - #8675 

## How to test 🧪

- Checkout this PR
- Run `yarn build && yarn pack` in `packages/validation`
  - Note the absolute path of the built package (for me it is `/Users/bnussman/Development/manager/packages/validation/linode-validation-v0.17.0.tgz`)
  - Set this version in api-v4's package.json in palace of `"*"`
- Run `yarn build && yarn pack` in `packages/api-v4`
  - Note the absolute path of the built package (for me it is `/Users/bnussman/Development/manager/packages/api-v4/linode-api-v4-v0.85.0.tgz`)
- Make a new Node.js + Typescript project outside of Cloud Manager
- Install the built package into your test node project and make sure you can build your node projct with tsc AND run it with node.js without observing any errors 

```json
"dependencies": {
    "@linode/api-v4": "/Users/bnussman/Development/manager/packages/api-v4/linode-api-v4-v0.85.0.tgz"
  },
```

PR #8739, opened 2023-01-23
M3-6135: Fix `api-v4` and `validation` CommonJS not accepted by Node.js

bnussman-akamai:
### The Bug
- Users that want to use api-v4 or validation in plain javascript + node would see an error like this

```sh
index.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs"
```

### The Fix
- Removes tsup's `legacyOutput` option so we get expected file extension (`.cjs`) that work in Node.js

> **Note**: This is completely unrelated to https://github.com/linode/manager/pull/8716

## How to test 🧪

- Checkout this PR
- Build and Pack validation by doing `cd packages/validation && yarn build && yarn pack`
  - Copy output file path to clipboard for next step
- Edit api-v4's package.json to point validation's version to that file
```
  "dependencies": {
    "@linode/validation": "/Users/bnussman/Development/manager/packages/validation/linode-validation-v0.17.1.tgz",
  },
```
- Go to the repo's root and run `yarn` to re-resolve packages
- Build and pack api-v4 with `cd packages/api-v4 && yarn build && yarn pack`
  - Note the output file 
- Make a new folder outside cloud manager's repo
- Init a new npm (or yarn or pnpm) project with `npm init`
- Install api-v4's build into this project with `npm i /Users/bnussman/Development/manager/packages/api-v4/linode-api-v4-v0.85.0.tgz`  (make sure you update that absolute path with the one you noted when you built api-v4)
- Make an index.js with the following content
```javascript
const api = require("@linode/api-v4");

api.setToken("test");

async function run() {
	const l = await api.getLinodes();
	console.log(l)
}

run();
```
- Run that javascript with `node index.js`
- If you did not get the error described above and the JS executed as expected, this bug has been fixed! 🚀

### DIFF
--- a/packages/api-v4/README.md
+++ b/packages/api-v4/README.md
@@ -17,7 +17,7 @@ $ yarn add @linode/api-v4
 or with a CDN:
 
 ```js
-<script src="https://unpkg.com/@linode/api-v4/lib/iife/index.js"></script>
+<script src="https://unpkg.com/@linode/api-v4/lib/index.global.js"></script>
 ```
 
 ## Usage
--- a/packages/api-v4/package.json
+++ b/packages/api-v4/package.json
@@ -23,25 +23,25 @@
   "license": "Apache-2.0",
   "private": false,
   "type": "module",
-  "main": "lib/index.js",
-  "module": "lib/esm/index.js",
+  "main": "lib/index.cjs",
+  "module": "lib/index.js",
   "exports": {
     ".": {
-      "import": "./lib/esm/index.js",
-      "require": "./lib/index.js"
+      "import": "./lib/index.js",
+      "require": "./lib/index.cjs"
     },
     "./lib": {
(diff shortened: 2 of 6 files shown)


bnussman-akamai:
Oh no. Looks like were going to run into issues with our unit tests. I would assume it is because we’re on old jest packages.

hkhalil-akamai (approved):
Confirmed this works using the listed steps for CommonJS and ES6 imports.

mjac0bs (approved):
++ for the detailed testing instructions.

I see node errors anymore and successfully executing JS returning a list of linodes from the API.

jdamore-linode (approved):

PR #8742, opened 2023-01-24
M3-6065: API Tokens Refactors and Fixes

bnussman-akamai:
## Description 📝

- Refactors all of `/profile/tokens` to use functional React components and React Query
- Uses Formik for token forms
- These changes subsequently fix unexpected cache invalidation behavior when revoking API tokens

## Todo ☑️
- [x] Fix broken tests
- [x] Re-evaluate testing and add tests
- [x] Consider adding relative expiry date 

## Preview 📷

Screen Shot 2023-01-24 at 5 19 19 PM

## How to test 🧪

- Test all actions possible on the `/profile/tokens` page
  - Test Token Creation
  - Test Token Deletion
  - Test Token Editing
  - Test ordering
  - Test pagination


packages/manager/src/features/Profile/APITokens/CreateAPITokenDrawer.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  it('Should default to read/write for all scopes', () => {
+    const { getByLabelText } = renderWithTheme(
+      <CreateAPITokenDrawer {...props} />
+    );
+    const selectAllReadWriteRadioButton = getByLabelText(
+      'Select read/write for all'
+    );
+    expect(selectAllReadWriteRadioButton).toBeChecked();
+  });
+  it('Should default to never for expiration', () => {
```
bnussman-akamai, line 73:
```suggestion
  it('Should default to 6 months for expiration', () => {
```

carrillo-erik (approved):

jdamore-linode:
Looks good, Banks! A couple observations, not sure if any of these are major issues though:

- Just noting the styling difference between this branch and prod where the PAT labels are bold in prod but not bold on your branch; I personally prefer it not to be bold since it seems more consistent with our other tables, but pointing it out just in case.
- Usually on RQ-ified pages, there's a behavior where if you create a resource somewhere else (via the CLI, another browser tab, etc.) the page fetches the new data on some interval and automatically updates the page contents to reflect additions/changes to the data -- I'm not seeing that on this page. Not sure if it matters, but also wasn't sure if it was intended to behave that way or not.
- If I create a personal access token, and then immediately create a new personal access token, the token creation form is pre-populated with the data from the previous token I generated (label, scopes, etc.). This doesn't appear to be the case in prod -- there, you're met with a clear form every time you create a PAT.
- This isn't caused by your PR, but this is the first time I've seen the pagination on this page and I was a little thrown off -- because there's no spacing between the pagination controls and the list beneath it, it almost appears as though the pagination controls are for the _Third Party Access Tokens_ list and not the _Personal Access Tokens_ list:
  - Screen Shot 2023-02-01 at 10 20 44 AM

jdamore-linode (approved):

bnussman-akamai:
@jdamore-linode Thank you for the great review. 

- I removed the bold intentionally :) I will make sure this is ok with UI//UX before merging
- I updated the create drawer form to reset  on successful creation. Normally, we'd reset the drawer whenever it is closed. I kind of like this because if someone were to close the drawer on accident, they could re-open it and still have the same state. Does this make sense or should I just follow our normal pattern of just resetting state more often?
- I added some RQ code to invalidate Personal Access Tokens if the API sends an `Event`. We don't get any events for `Third Party Access Tokens` so there is no action to take there. 
- Weird pagination spacing should be fixed.

bnussman-akamai:
@jdamore-linode Agreed. I'll make the behavior match other drawers! Thanks!

PR #8744, opened 2023-01-25
M3-4911: Entity transfer integration & E2E tests

jdamore-linode:
## Description 📝

**What does this PR do?**
Adds integration tests and end-to-end tests for Cloud Manager service transfer functionality.

- Uses real API requests to confirm that the first half of the transfer flow (initiating a transfer) can be completed.
  - This test may take a minute since it needs to create and boot a new Linode before the test can begin. 
- Uses mocked API requests to confirm that Cloud UI responds accordingly during second half of the transfer flow (accepting a transfer) 
- Uses mocked API requests to confirm that content is displayed on transfer landing page as expected.

## How to test 🧪

Run `yarn dev`, then separately run:

```bash
yarn cy:run -s "cypress/e2e/account/service-transfer.spec.ts"
```


packages/manager/cypress/support/util/random.ts
```diff
@@ -199,3 +199,30 @@ export const randomPhrase = (count: number = 5): string => {
     })
     .join(' ');
 };
+
+/**
+ * Generates a random string which resembles a v4 UUID.
+ *
+ * This does generate a valid UUID, nor does it offer the same guarantees as a
```
dwiley-akamai, line 206: Should this be: `This does not generate...` ?

dwiley-akamai (approved):
LGTM, great work! 🚀

packages/manager/cypress/e2e/account/service-transfer.spec.ts
```diff
@@ -0,0 +1,421 @@
...
+          cy.findByText(pendingTransfer.token)
+            .should('be.visible')
+            .closest('tr')
+            .within(() => {
+              cy.findByText('5 Linodes').should('be.visible');
+              cy.findByText(formatDate(pendingTransfer.created)).should(
+                'be.visible'
+              );
+              cy.findByText(formatDate(pendingTransfer.expiry)).should(
+                'be.visible'
```
mjac0bs, line 175: This test ("list service transfers on the landing page") fails for me, and it seems like something is going on with time stamps (and maybe time zones?). While the tests expected `2023-01-25 15:21`, I only have tokens with `created` timestamps of `2023-01-25 23:51`. Ran it via terminal and then with debug to watch and ended up failing both times.

```
1) Account service transfers
       lists service transfers on landing page:
     AssertionError: Timed out retrying after 80000ms: Unable to find an element with the text: 2023-01-25 15:10. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.
```

jdamore-linode: Thanks for catching this Mariah, I'll take a look and fix it soon!

I'm not too surprised about this; I tossed in the `formatDate()` call as a quick attempt to get the tests to pass for me and it just so happens that they did, but I didn't actually do any snooping around in the frontend code to make sure I handled the timestamps in the exact same way.

packages/manager/cypress/support/util/random.ts
jdamore-linode: It definitely should be, thanks Dajahi!

packages/manager/cypress/e2e/account/service-transfer.spec.ts
jdamore-linode: `lists service transfers on landing page` should pass for you now -- the failure was caused by the fact that calling `formatDate()` from a Cypress test fails to take into account your profile timezone; it falls back to eastern time so most of us won't observe the failure, but in your case it leads to a different output than what's displayed in Cloud.

I fixed this with a band-aid, essentially, but I'll come up with a better solution (and clean this up) next time I run into a similar situation.

mjac0bs: The reason why I've run into this issue makes sense. And I see your changes, just trying running the tests again with them. It looks like I'm still seeing a failure at that same spot. Now my `created` time vs the expected cypress time is only 3 hours off, though, not 8. 🤔 

image

jdamore-linode: @mjac0bs Ok, I think I fixed it for real this time, but if you could double check next time you have some time to kill I'd appreciate it!

I did this by modifying the `formatDate()` function to accept a timezone as an option, so that we can specify a timezone via the E2E tests where the normal method doesn't work. I don't really like changing code in `./src` just for the sake of a test here or there (especially in this case, where I honestly can't think of another scenario where specifying the timezone would be useful), so if you or anybody has any opinion on that I'd be happy to revisit.

bnussman-akamai (approved):

packages/manager/cypress/e2e/account/service-transfer.spec.ts
mjac0bs: image

@jdamore-linode Yay for green checks. 🎉 As far as the necessary change to make this pass... I probably don't have enough knowledge of e2e code to have the most informed opinion on reluctance to modify `./src`. Without the modifications to `./src`, would folks outside ET just expect to see that test fail locally every time we ran it, or would it also affect when we merge in a PR and e2e test are run? I think I'm trying to get at: what's the scope of the failures we'd see with something like this?

mjac0bs (approved):

packages/manager/cypress/e2e/account/service-transfer.spec.ts
jdamore-linode: @mjac0bs Without the change, I think the problem would only occur in cases where `formatDate()` is called in a test using a test account with a profile whose timezone's set to anything other than Eastern Time (or if the test mocks the profile and it matches that criteria), as it seems that's the default timezone that gets used by the `formatDate()` function if it doesn't retrieve one from the profile. I don't think there's any major risk of this problem causing failures in CI since we have control over the accounts, but it's definitely a potential footgun and an annoyance for everybody in other timezones who wants to run the tests on their accounts.

PR #8747, opened 2023-01-27
Fix strings in cli commands

CodyFinn-Akamai:
## Description 📝

Wraps the arguments for the generated CLI commands with single quotes so everything is properly escaped for the CLI.

## How to test 🧪

Use the generated CLI command to make a new linode.

run `yarn test generate-cli`

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
@@ -329,7 +329,13 @@ export class LinodeCreate extends React.PureComponent<
   }
 
   getPayload = () => {
-    return {
+    const selectedRegion = this.props.selectedRegionID || '';
+    const regionSupportsVLANs = doesRegionSupportFeature(
+      selectedRegion,
+      this.props.regionsData,
+      'Vlans'
+    );
+    const payload = {
       image: this.props.selectedImageID,
       region: this.props.selectedRegionID,
       type: this.props.selectedTypeID,
@@ -350,17 +356,6 @@ export class LinodeCreate extends React.PureComponent<
       stackscript_id: this.props.selectedStackScriptID,
       stackscript_data: this.props.selectedUDFs,
     };
-  };
-
-  createLinode = () => {
-    const selectedRegion = this.props.selectedRegionID || '';
-    const regionSupportsVLANs = doesRegionSupportFeature(
-      selectedRegion,
-      this.props.regionsData,
-      'Vlans'
-    );
-
-    const payload = this.getPayload();
 
     if (
       regionSupportsVLANs &&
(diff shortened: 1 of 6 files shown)


dwiley-akamai (commented):
Seems like the `authorized_users` field is still causing it to trip up 🤔  

Screen Shot 2023-01-27 at 5 17 51 PM

Screen Shot 2023-01-27 at 5 18 01 PM

CodyFinn-Akamai:
Where I found a good example to base this command syntax off of
https://www.linode.com/docs/api/linode-instances/#linode-create__request-samples

bnussman-akamai (changes requested):
I found that the generated linode-cli command fails if a tag has spaces in it

```sh
linode-cli linodes create \
  --image 'linode/centos-stream8' \
  --region 'us-central' \
  --type 'g6-nanode-1' \
  --label 'centos-us-central' \
  --tags tag with space \
  --root_pass 't5t5t5t5t5t5t5tt5t5' \
  --booted 'true' \
  --backups_enabled 'false' \
  --private_ip 'false'
```

> linode-cli linodes create: error: unrecognized arguments: with space

dwiley-akamai (commented):
`authorized_users` still presents a challenge here:

Screen Shot 2023-02-01 at 3 38 47 PM

Screen Shot 2023-02-01 at 3 37 51 PM

As a side note, and perhaps it was brought up already in another PR, but it also seems that VLANs don't get accounted for in the generated cURL & CLI commands.

CodyFinn-Akamai:
> `authorized_users` still presents a challenge here:
> 
> Screen Shot 2023-02-01 at 3 38 47 PM
> 
> Screen Shot 2023-02-01 at 3 37 51 PM
> 
> As a side note, and perhaps it was brought up already in another PR, but it also seems that VLANs don't get accounted for in the generated cURL & CLI commands.

Is your CLI setup to use the prod-test-015 user account? And the VLANs aren't being passed in to the parsing arguments. If they were then they would be parsed

dwiley-akamai:
> Is your CLI setup to use the prod-test-015 user account? And the VLANs aren't being passed in to the parsing arguments. If they were then they would be parsed

It was set up to use that account but I just fired off a [fresh configuration setup](https://www.linode.com/docs/products/tools/cli/guides/install/#interactive-configuration) and that does seem to have fixed the issue 👍

dwiley-akamai (approved):
The functionality LGTM now. One of the test assertions is failing but I'm approving pending that test being adjusted

Screen Shot 2023-02-03 at 11 00 39 AM

edit: just confirmed the tags with spaces issue that Banks mentioned previously is still there

cpathipa (approved):

dwiley-akamai (approved):

bnussman-akamai (commented):
Screen Shot 2023-02-03 at 3 21 44 PM

Results in error
```
linode-cli linodes create: error: unrecognized arguments: adding a new tag
```

I understand the thought behind escaping everything, but the API docs site does not do this and I don't think customers will love the look of the maximum escaping. Is this our only option? 

Screen Shot 2023-02-03 at 3 23 10 PM

CodyFinn-Akamai:
@bnussman-akamai The docs site doesn't have to deal with more exotic strings like we do. We don't want to give users a command that could cause a major error due to a naming convention they have or a password generated by a password manager. I think until we have a better implementation this will be good enough to launch the feature.

PR #8750, opened 2023-01-27
Api cli close modal on linode create

cpathipa:
## Description 📝
Address UX Feedback
Automatically closes commands modal upon creating Linode from terminal.

## How to test 🧪

- Go to Linode create page and fill required fields.
- Click on Create Command Line button.
- Copy command and execute in terminal.
- Observe behavior of closing modal and navigating user back to Linode's page.

Test creating Linode, clone Linode, Deploy Linode from backup

- yarn test
- yarn cy:run

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
@@ -73,17 +73,13 @@ const ApiAwarenessModal = (props: Props) => {
   const history = useHistory();
   const { events } = useEvents();
 
-  const isLinodeCreated = useMemo(
-    () =>
-      events.filter(
-        (event) =>
-          event.action === 'linode_create' &&
-          event.entity?.label === payLoad.label
-      ).length === 1,
-    [events]
-  );
+  const isLinodeCreated =
+    events.filter(
+      (event) =>
+        event.action === 'linode_create' &&
+        event.entity?.label === payLoad.label
+    ).length === 1;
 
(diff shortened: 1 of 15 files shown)


bnussman-akamai (approved):
Redirect works as expected! Nice changes to the copy icon styles too 🚀

packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
```diff
@@ -95,6 +111,13 @@ const ApiAwarenessModal = (props: Props) => {
     });
   };
 
+  useEffect(() => {
+    if (isLinodeCreated && isOpen) {
+      onClose();
+      history.replace('/linodes');
```
bnussman-akamai, line 117: I wonder if we should send the user to the Linodes Details page (`/linodes/:id`). If a user with hundreds of Linodes got sent back to the Linodes landing page, they may not see the new Linode if it is not on the current page.

jdamore-linode (approved):
Echoing @bnussman-akamai's suggestion, but otherwise this looks really good! Confirmed the redirect occurs after creating the Linode (and that no redirect occurs if the Linode created via CLI's label doesn't match the expected label). In the middle of running e2e tests, but I'm assuming they'll pass and will circle back if I see any suspicious failures.

packages/manager/src/features/linodes/LinodesCreate/ApiAwarenessModal/index.tsx
cpathipa: Make sense, updated as suggested.

cpathipa:
Thank you @bnussman-akamai and @jdamore-linode.

PR #8751, opened 2023-01-27
Upgrade to Typescript 5.0

bnussman-akamai:
## Description 📝

- We are on a 2 year old version of Typescript
- Updates to the latest Typescript v5 which is still in beta
- Use `incremental` mode to speed up type checking
- Fixes any type errors caused by this upgrade
- Removes `reloadableWithRouter`
  - I have yet to understand what this even did so this may break things badly - please test for this

## How to test 🧪

- Check all type changes I made and make sure they make sense


dwiley-akamai (commented):
The first glance and initial testing looked good.

I am going to circle back later and dig into the question of `reloadableWithRouter`. Its removal doesn't seem to have broken anything, but it looks like it had something to do with refreshing the page if URL params changed? I don't think I ever observed that functionality before, so that logic may have not be functional for a long time.

packages/manager/tsconfig.json
```diff
@@ -36,6 +36,10 @@
     "strictNullChecks": true,
     "suppressImplicitAnyIndexErrors": true,
 
+    "ignoreDeprecations": "5.0",
```
jdamore-linode, line 39: Can you elaborate on why we need this? Does this imply that our config and/or code relies on deprecated features?

bnussman-akamai: Yeah, we do rely on a deprecated option unfortunately. It would be relatively easy to fix all of the tsc errors caused by removing `suppressImplicitAnyIndexErrors`, but for the sake of getting on a recent Typescript version, we can use `ignoreDeprecations`.
 
Screenshot 2023-02-21 at 1 05 32 PM

bnussman-akamai: I'm thinking of closing this PR for now and we can revisit when Typescript 5 is out of beta. I mainly opened this to see what performance improvements we get from v5

jdamore-linode: Thanks for the explanation! Totally agreed.

PR #8752, opened 2023-01-31
M3-5193: RQ-ify tags + enhancements

hkhalil-akamai:
## Description 📝

RQ-ifies all components that hit the `getTags` API endpoint, with some UI and code adjustments:

`TagCell`
* Add loading spinner when adding and deleting tags

`TagDrawer`
* Simplified using the newer `TagsPanel` component

`TagsPanel`
* Fix a bug causing error messages to not display
* Use inline errors rather than toast

`Tags`
* Fix a small style issue when hovering over tags

### UI Changes:

<details>
<summary><code>TagCell</code> and <code>TagsPanel</code></summary>

https://user-images.githubusercontent.com/122488130/216154340-3bd91237-24a8-42e5-9dea-b619f1cee791.mov

</details>

<details>
<summary><code>Tags</code></summary>

<details>
<summary>Before</summary>

https://user-images.githubusercontent.com/122488130/216154475-69655671-746d-4d78-900f-9627ac79cd0b.mov

</details>

<details>
<summary>After</summary>

https://user-images.githubusercontent.com/122488130/216154545-97a70ac1-5438-408b-aec2-050a7152fd70.mov

</details>

</details>

## How to test 🧪

* Ensure user tag suggestions work as expected (e.g., in Linodes/NodeBalancers/Kubernetes, etc.)
* Ensure UX of adding and removing tags is clear (e.g., progress indication, errors)


dwiley-akamai (commented):
I'm going to test the functionality and comb through the code in depth tomorrow, but I did notice that on this branch, if you don't first click the "Add a tag" button on the Linode Detail page and instead go straight into the TagDrawer, the dropdown when you click "Add a tag" in the drawer is not populated with any existing tags (in contrast to Prod, where the dropdown is populated).

packages/manager/src/components/Tag/Tag.tsx
```diff
@@ -96,7 +102,7 @@ const useStyles = makeStyles((theme: Theme) => ({
 
 export interface Props extends ChipProps {
   label: string;
-  colorVariant?: Variants;
```
hkhalil-akamai, line 99: All existing usages of this component provide this prop

packages/manager/src/components/TagCell/TagCell.tsx
```diff
@@ -137,69 +145,18 @@ export const TagCell: React.FC<Props> = (props) => {
     [tags]
   );
 
-  return inTableContext ? (
```
hkhalil-akamai, line 140: Proposing to delete this: no existing usages of TagCell are in table context

packages/manager/src/components/TagCell/TagDrawer.tsx
```diff
@@ -33,48 +22,11 @@ export interface TagDrawerProps {
...
-          key={`tag-item-${thisTag}`}
-          colorVariant="lightBlue"
-          label={thisTag}
-          onDelete={() => _deleteTag(thisTag)}
-        />
-      ))}
-      <div className={classes.addTag}>
-        <AddTag tags={tags} addTag={_addTag} label={'Add a tag'} />
-      </div>
+      <TagsPanel tags={tags} updateTags={updateTags}></TagsPanel>
```
hkhalil-akamai, line 29: TagsPanel is a newer component that offers goodies like progress indication

packages/manager/src/components/TagsPanel/TagsPanel.tsx
```diff
@@ -134,56 +133,49 @@ export interface Props {
...
   const [isCreatingTag, setIsCreatingTag] = React.useState(false);
   const [tagInputValue, setTagInputValue] = React.useState('');
-  const [loading, setLoading] = React.useState(false);
+  const [tagLoading, setTagsLoading] = React.useState(false);
+
+  const {
+    data: userTags,
+    isLoading: userTagsLoading,
+    error: userTagsError,
+  } = useTags(!isRestrictedUser());
```
hkhalil-akamai, line 146: What's the difference between `isRestrictedUser` (from `permissionsHelper.ts`) and `_isRestrictedUser` (from `useAccountManagement` hook). If they are duplicated, should they be combined?

packages/manager/src/components/TagsPanel/TagsPanel.tsx
```diff
@@ -195,21 +187,21 @@ const TagsPanel: React.FC<Props> = (props) => {
...
         ]);
-        setLoading(false);
         setTagError('');
       })
-      .catch((_) => {
-        enqueueSnackbar(`Could not delete Tag: ${label}`, {
-          variant: 'error',
-        });
-        setLoading(false);
+      .catch((e) => {
```
hkhalil-akamai, line 199: Moving this to an inline error because other errors on this page (e.g. duplicate tag, invalid tag, error adding tag) are inline.

packages/manager/src/features/linodes/LinodeEntityDetail.tsx
```diff
@@ -847,31 +847,18 @@ export const Footer: React.FC<FooterProps> = React.memo((props) => {
...
-      );
-    },
-    [linodeTags, linodeId, updateLinode, enqueueSnackbar]
-  );
-
-  const deleteTag = React.useCallback(
-    (tag: string) => {
-      const newTags = linodeTags.filter((thisTag) => thisTag !== tag);
-      updateLinode({ linodeId, tags: newTags }).catch((e) =>
+  const updateTags = React.useCallback(
```
hkhalil-akamai, line 850: Merging `addTag` and `deleteTag` into `updateTag` to match conventions in other components (e.g., TagsDrawer). However, as a consequence, the error message becomes less specific ("Error updating tag" instead of "Error adding/deleting tag")

packages/manager/src/components/TagCell/TagDrawer.tsx
dwiley-akamai: since there are no child components:

```suggestion
<TagsPanel tags={tags} updateTags={updateTags} />
```

packages/manager/src/components/TagCell/AddTag.tsx
```diff
@@ -37,41 +37,35 @@ export type CombinedProps = Props;
...
-  }, [_isRestrictedUser]);
-
-  const tagOptions = accountTags.filter(
-    (thisTag) => !tags.includes(thisTag.value)
+  const { data: accountTags, isLoading: accountTagsLoading } = useTags(
+    !_isRestrictedUser
   );
+  // @todo should we toast for this? If we swallow the error the only
+  // thing we lose is preexisting tabs as options; the add tag flow
+  // should still work.
```
dwiley-akamai, line 48: Outdated comment since errors are being displayed as notices and not toasts now

packages/manager/src/components/TagsPanel/TagsPanel.tsx
bnussman-akamai: In my opinion the best way to actually know if a user is restricted is to do the following. I like this way because it is the most direct and 1:1 with the API which aids maintainability. 

```typescript
  const isRestricted = profile?.restricted;
```

Lets avoid using `isRestrictedUser()` when possible because it will not react to changes in the React Query store if we were to mutate it for some reason. `isRestrictedUser()` was created in case we needed to know if a user was restricted in a clunky class component.

The `useAccountManagement` hook is fine. It just uses `useProfile` but has a lot of extra stuff going on.

packages/manager/src/components/TagCell/TagCell.tsx
```diff
@@ -210,11 +167,16 @@ export const TagCell: React.FC<Props> = (props) => {
       alignItems="center"
       wrap="nowrap"
     >
+      {loading && (
```
dwiley-akamai, line 170: optional: we can use a ternary here instead, i.e. `loading ? ( ... ) : null`

A few years ago we started shifting away from `&&` (based on [this post](https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx) IIRC). `loading` is a boolean so we shouldn't ever run into any issues in this specific situation, but personally I just stick with ternaries now.

packages/manager/src/components/TagsPanel/TagsPanel.tsx
```diff
@@ -195,21 +187,21 @@ const TagsPanel: React.FC<Props> = (props) => {
...
+        const tagError = getErrorStringOrDefault(e, 'Error while deleting tag');
+        setTagError(tagError);
+      })
+      .finally(() => {
+        setTagsLoading(false);
```
dwiley-akamai, line 204: What's the argument for not including this in the `catch()` block instead? (same question for the other `.finally()`ies

dwiley-akamai (commented):
This is looking good overall, nice job! I left a few comments.

This interaction does seem a bit odd:

https://user-images.githubusercontent.com/114682940/216455431-598d0e86-f047-4219-b6c8-84a2c41d5163.mov

This is if you block requests to `*tags*` and go to a NodeBalancer Detail page. The error shows up initially, but then disappears once you click "Add a tag" -- it doesn't come back and the dropdown is just empty.

hkhalil-akamai:
> This is looking good overall, nice job! I left a few comments.
> 
> This interaction does seem a bit odd:
> 
>  Screen.Recording.2023-02-02.at.3.51.00.PM.mov 
> This is if you block requests to `*tags*` and go to a NodeBalancer Detail page. The error shows up initially, but then disappears once you click "Add a tag" -- it doesn't come back and the dropdown is just empty.

Good point. Possible solutions:

1. Persisting the error and adding a second line if more errors occur
2. Persisting the latest error on a single line
3. Showing the error in the dropdown instead of "No results"

packages/manager/src/components/TagCell/AddTag.tsx
hkhalil-akamai: This todo refers to the TagCell component (for example, in the Linode details header), which does not have inline errors.

packages/manager/src/queries/tags.ts
```diff
@@ -0,0 +1,22 @@
...
+    ...queryPresets.shortLived,
+    enabled,
+  });
+
+const getAllTags = (passedParams: any = {}, passedFilter: any = {}) =>
+  getAll<Tag>((params, filter) =>
+    getTags({ ...params, ...passedParams }, { ...filter, ...passedFilter })
+  )().then((data) => data.data);
+
+export const updateTagsData = (newData: Tag[]): void => {
```
dwiley-akamai, line 20: Seems like this one doesn't get used anywhere currently.

I'm thinking this should make use of `useMutation` and we should be able to do some refactoring to use `updateTagsData` instead of `updateTags()`

dwiley-akamai (commented):
> Good point. Possible solutions:
> 
> 1. Persisting the error and adding a second line if more errors occur
> 2. Persisting the latest error on a single line
> 3. Showing the error in the dropdown instead of "No results"

I would be in favor of persisting the latest error, if one exists, on a single line.

packages/manager/src/queries/tags.ts
hkhalil-akamai: The naming in the PR is a bit confusing because this hook gets all the tag suggestions, which there's no API endpoint to update. My understanding is that the API collects all the tags from existing linodes, DBs, etc and returns them as the tag suggestions.

Maybe the hook should be renamed to `useTagSuggestions`?

hkhalil-akamai: But yeah, looks like updateTagsData isn't used anywhere so it should probably be deleted

bnussman-akamai:
Everything is looking good to me. One thing I see is that whenever we focus the TagsPanel input, tags are fetched. 

Is there a way we can just fetch all tags initially, then from then on, just rely on manual cache mutations to keep RQ up to date?

hkhalil-akamai:
> Everything is looking good to me. One thing I see is that whenever we focus the TagsPanel input, tags are fetched.
> 
> Is there a way we can just fetch all tags initially, then from then on, just rely on manual cache mutations to keep RQ up to date?

I updated the Tags query to use the `longLived` preset which significantly cuts back on the number of requests to the tags endpoint.

I also added mutations to roughly keep the suggestions up to date, but I'm not entirely sure how the API synthesizes the list of tag suggestions, but I think appending new tags to the list is a reasonable approximation.

bnussman-akamai (approved):
Great work on this!

packages/manager/src/queries/tags.ts
```diff
@@ -0,0 +1,25 @@
...
+  });
+
+const getAllTagSuggestions = (passedParams: any = {}, passedFilter: any = {}) =>
+  getAll<Tag>((params, filter) =>
+    getTags({ ...params, ...passedParams }, { ...filter, ...passedFilter })
+  )().then((data) => data.data);
+
+export const updateTagsSuggestionsData = (newData: Tag[]): void => {
+  const uniqueTags = Array.from(new Set(newData.map((tag) => tag.label)))
+    .sort()
```
bnussman-akamai, line 22: Do we need the sort? What function does it serve?

hkhalil-akamai: The API returns tag suggestions sorted alphabetically, so this replicates that behavior. The use of a set is to avoid duplicate suggestions in the list.

bnussman-akamai: > API returns tag suggestions sorted alphabetically

Ah, nice. I didn't know this! Great job!

packages/manager/src/components/TagsPanel/TagsPanel.tsx
```diff
@@ -326,8 +300,10 @@ const TagsPanel: React.FC<Props> = (props) => {
             />
           );
         })}
-        {tagError && (
-          <Typography className={classes.errorNotice}>{tagError}</Typography>
+        {(tagError.length > 0 || userTagsErrorDisplay.length > 0) && (
+          <Typography className={classes.errorNotice}>
+            {tagError.length > 0 ? tagError : userTagsErrorDisplay}
+          </Typography>
```
dwiley-akamai, line 306: Instead of wrapping it in `<Typography />`, putting it in a `<Notice error text={...} />` might be more in line with how we present these visually in other areas of the app

dwiley-akamai (approved):
LGTM, left one minor suggestion of something to look into

packages/manager/src/components/TagsPanel/TagsPanel.tsx
hkhalil-akamai: In this case, the `<Notice>` doesn't seem to fit well visually:

Before:
Screen Shot 2023-02-17 at 5 08 58 PM

After:
image

PR #8753, opened 2023-02-01
Manually pin `ua-parser-js` to `0.7.33` for CVE-2022-25927

bnussman-akamai:
## Description 📝

- Update `ua-parser-js` to `0.7.33` for CVE-2022-25927
- We prefer to do this instead of approving https://github.com/linode/manager/pull/8741 because this yarn  resolution method allows our `yarn.lock` to remain in sync with our package.json(s)

## How to test 🧪

- Ensure manager builds and works as expected

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "csv-parse": "^4.6.3",
     "node-fetch": "^2.6.7",
     "y18n": "^4.0.0",
-    "ua-parser-js": "^0.7.23",
+    "ua-parser-js": "^0.7.33",
     "trim": "^0.0.3",
     "prismjs": "^1.24.0",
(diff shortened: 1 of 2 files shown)


carrillo-erik:
Interesting observation, I don't think I would have considered the shortcomings with the previous offered solution.

dwiley-akamai (approved):

jdamore-linode (approved):

cpathipa (commented):
LGTM! Able to build manager. 

Only thing I noticed is below warning message not sure if it causing in upstream branch as well. 
image

bnussman-akamai:
I verified that that console error **is** an existing issue in `develop` @cpathipa

PR #8756, opened 2023-02-02
M3-5182: RQ-ify Images Pt. 3

dwiley-akamai:
## Description 📝
Complete the trilogy of PRs for the RQ-ify Images effort. 🎉  🚀 

High-level overview of this PR:

- RQ-ify the Upload Image flow
  - The only thing remaining in Redux is the `set-pending-upload` logic (not a good candidate for moving to RQ)
- RQ-ify Images in search
  -  the `getSearchEntities` piece was a bit challenging, but the best course of action that didn't feel like a messy anti-pattern was to just decouple Images and Linodes from being returned in there, and to instead pass them in where needed in `withStoreSearch`
- `images` removed from Redux store & Image-related files in `src/store` removed

## How to test 🧪
- Ensure the Upload Image, Capture Image, StackScript, and Linode Create flows all work as expected.
- Ensure that the Search functionality works as expected, particularly as it relates to Images and Linodes.


packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -521,13 +461,46 @@ export const ImagesLanding: React.FC<CombinedProps> = (props) => {
...
+              <TableSortCell
+                active={orderBy === 'label'}
+                direction={order}
+                label="Image"
+                handleClick={handleOrderChange}
+              >
+                Image
+              </TableSortCell>
+              <Hidden smDown>
+                <TableCell>Status</TableCell>
```
dwiley-akamai, line 476: We were hoping we'd be able to sort by `Status` with an x-filter via the API. Making this sortable works when you first click on the column cell, but if you navigate away from the Images landing page and come back, it will fail, and navigating directly to a URL with the sort-related params will fail as well.

We don't do any client-side sorting elsewhere in the app in these situations, so I decided to make this column non-sortable. If the API supports it down the line, we can come back and update this accordingly.

bnussman-akamai (commented):
On `ImagesLanding.tsx`, I was expecting to see two `useImagesQuery`s. With the current implementation, I'm not sure how pagination would work in regards to going to the next page of either `Custom` or the `Recovery` section. 

```ts
  const {
    data: automaticImages,
    isLoading: automaticImagesLoading,
    error: automaticImagesError,
  } = useImagesQuery(
    {
      page: pagination.page,
      page_size: pagination.pageSize,
    },
    {
      type: 'automatic',
      is_public: false,
      ...filter,
    }
  );

  const {
    data: manualImages,
    isLoading: manualImagesLoading,
    error: manualImagesError,
  } = useImagesQuery(
    {
      page: pagination.page,
      page_size: pagination.pageSize,
    },
    {
      type: 'manual',
      is_public: false,
      ...filter,
    }
  );
```

Is filtering on `type` bugged or not possible? I can't seem to get it to work

dwiley-akamai:
Confirmed there is an API bug for filtering with automatic private images. The most recent commit has the logic splitting out the automatic and manual images queries though. We can hold off on releasing the Images RQ-ification work until the API bug has been fixed

dwiley-akamai:
The API bug fix has been merged into dev and will be released to prod next week, so I'm removing the `Blocked` label here as this can be tested currently by pointing your local to the `dev` environment

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -141,19 +94,19 @@ const defaultDialogState = {
...
-const preferenceKey = 'images';
-
-export const ImagesLanding: React.FC<CombinedProps> = (props) => {
+export const ImagesLanding: React.FC<CombinedProps> = () => {
   const classes = useStyles();
+  const history = useHistory();
+  const { enqueueSnackbar } = useSnackbar();
 
-  const pagination = usePagination(1, preferenceKey);
+  const pagination = usePagination(1, queryKey);
```
bnussman-akamai, line 102: We will need 2 `usePagination`s on this page becuase each table needs different state to track the page numbers

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -141,19 +94,19 @@ const defaultDialogState = {
...
 
-  const { order, orderBy } = useOrder(
+  const { order, orderBy, handleOrderChange } = useOrder(
```
bnussman-akamai, line 104: Same here applies to `useOrder`. Users may want to sort each table by different keys at the same time. The 3ed paramater of `useOrder` will need to be used here so that the query params have a way to differentiate between the two tables

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -141,78 +95,99 @@ const defaultDialogState = {
...
+      type: 'manual',
+      is_public: false,
+    }
+  );
 
-  const { order, orderBy } = useOrder(
+  // Pagination, order, and query hooks for automatic/recovery images
+  const paginationForAutomaticImages = usePagination(
+    1,
+    `automatic-${queryKey}`
```
bnussman-akamai, line 141: For all of these preference keys, I think we should have the common key first followed by the variant.

```suggestion
    `${queryKey}-automatic`
```

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -141,78 +95,99 @@ const defaultDialogState = {
...
+    handleOrderChange: handleAutomaticImagesOrderChange,
+  } = useOrder(
     {
       orderBy: 'label',
-      order: 'desc',
+      order: 'asc',
     },
-    `${preferenceKey}-order`
+    `automatic-${queryKey}-order`
   );
```
bnussman-akamai, line 153: For both of the `useOrder` hooks, I think we should add a 3ed parameter so that the query params don't override each other when sorting.

```suggestion
    `${queryKey}-automatic-order`,
    'automatic'
  );
```

Screenshot 2023-02-22 at 3 43 33 PM

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -537,12 +546,56 @@ export const ImagesLanding: React.FC<CombinedProps> = (props) => {
...
+                label="label"
+                handleClick={handleAutomaticImagesOrderChange}
+              >
+                Image
+              </TableSortCell>
+              <Hidden smDown>
+                <TableCell>Status</TableCell>
+              </Hidden>
+              <Hidden smDown>
+                <TableCell>Created</TableCell>
```
bnussman-akamai, line 564: Let's let people sort by `created` because the API supports it

```suggestion
                <TableSortCell
                  active={automaticImagesOrderBy === 'created'}
                  direction={automaticImagesOrder}
                  label="created"
                  handleClick={handleAutomaticImagesOrderChange}
                >
                  Created
                </TableSortCell>
```

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -521,12 +489,53 @@ export const ImagesLanding: React.FC<CombinedProps> = (props) => {
...
+                label="label"
+                handleClick={handleManualImagesOrderChange}
+              >
+                Image
+              </TableSortCell>
+              <Hidden smDown>
+                <TableCell>Status</TableCell>
+              </Hidden>
+              <Hidden smDown>
+                <TableCell>Created</TableCell>
```
bnussman-akamai, line 507:
```suggestion
                <TableSortCell
                  active={manualImagesOrderBy === 'created'}
                  direction={manualImagesOrder}
                  label="created"
                  handleClick={handleManualImagesOrderChange}
                >
                  Created
                </TableSortCell>
```

bnussman-akamai (changes requested):

bnussman-akamai (approved):
Looking good. I'll do more testing when we get this merged into `feature/react-query-images` but things look great so far!

jaalah-akamai (approved):

PR #8757, opened 2023-02-02
M3-4090: Object Storage Add Ability to Create "Folders"

bnussman-akamai:
## Description 📝

- Adds ability to create Object Storage "folders"

## Preview 📷

Screen Shot 2023-02-02 at 3 32 05 PM

## How to test 🧪

- Test Object Storage in depth
  - Test folder creation and deletion
  - Verify object functionality still works as expected

## Questions for Team ❓

I don’t think I’m going to do a good job of asking this question, but I need some guidance on this. Which method should we use to create a folder in Cloud Manager?

- Create a fake folder in memory that does not get sent to the API at all so the user can go into the folder to upload an object, resulting in the folder actually being “created”
  - **Pros:** Less hacky from an S3 perspective
  - **Cons:** Folder will not persist if user refreshes the page
- Create a 0 byte object with a name that ends in a forward slash (the PR is currently doing this)
  - **Pros:** “Folder” will persist upon refresh
  - **Cons:** Feels hacky from an S3 perspective, "folders" will count towards the total object count because they are really objects

Object Storage (s3) does not really support “folders” so either solution seems hacky to me.

### DIFF
--- a/packages/manager/src/features/ObjectStorage/BucketDetail/BucketDetail.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketDetail/BucketDetail.tsx
@@ -179,7 +179,7 @@ export const BucketDetail: React.FC = () => {
 
       if (itemsInFolder.data.length > 1) {
         setDeleteObjectLoading(false);
-        setDeleteObjectError('The folder must be empty to delete it');
+        setDeleteObjectError('The folder must be empty to delete it.');
         return;
       }
     }
--- a/packages/manager/src/features/ObjectStorage/BucketDetail/BucketDetail.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketDetail/BucketDetail.tsx
@@ -67,6 +67,11 @@ const useStyles = makeStyles((theme: Theme) => ({
     textDecoration: 'underline',
     cursor: 'pointer',
   },
+  createFolderButton: {
+    [theme.breakpoints.down('md')]: {
+      marginRight: theme.spacing(),
+    },
+  },
 }));
 
 interface MatchParams {
@@ -357,6 +362,7 @@ export const BucketDetail: React.FC = () => {
       <Box display="flex" justifyContent="flex-end" mt={1.5} mb={0.5}>
         <Button
           buttonType="outlined"
+          className={classes.createFolderButton}
           onClick={() => setIsCreateFolderDrawerOpen(true)}
         >
           Create Folder
--- a/packages/manager/src/features/ObjectStorage/BucketDetail/BucketDetail.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketDetail/BucketDetail.tsx
@@ -33,6 +33,7 @@ import ObjectUploader from '../ObjectUploader';
 import {
   displayName,
   generateObjectUrl,
+  isEmptyObjectForFolder,
   tableUpdateAction,
 } from '../utilities';
 import BucketBreadcrumb from './BucketBreadcrumb';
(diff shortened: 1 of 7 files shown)


hkhalil-akamai:
One small thing I'm noticing is when adding a folder, the loading spinner disappears a half second before the drawer actually closes. Not sure if that's intentional though.

hkhalil-akamai:
Would it make sense to add a '..' entry at the top of the file list? This would just be a link to the parent folder.

bnussman-akamai:
@hkhalil-akamai Because we already have a breadcrumb, I think that serves the same purpose.

bnussman-akamai:
@hkhalil-akamai, updated the logic to not wait for the call to the presigned url so that the loading state is better. This should be fine because it is very unlikely the call will fail.

mjac0bs (commented):
Confirming folder creation and deletion works; non-folder object functionality seems to be working as expected. I think you took the better approach with a 0 byte object; folders that disappear on refresh would feel pretty odd from a UX perspective.

A couple of questions/thoughts: (May warrant their own tickets.)

*Create Folder Button*
- Tight spacing between the edge of the screen and the button once the window hits the `sm` breakpoint on down, which makes that right side of the button look a little odd. We might want to align with the edge of the object drop border? While the table also extends to the edge of the screen, it feels less visually squished since it is similar colors to as page rather than outlined in blue. This is minor.

*Folder Deletion*
- Do we want to apply TTC for folder deletion since it is a destructive action, or is it enough to have an error message on a non-empty folder?
- I agree with having some sort of warning/error when trying to delete a non-empty folder, though I do wonder about the efficiency of requiring a folder to be empty for deletion from a UX perspective. If a user had a folder consisting of many objects, they'd need to delete all objects individually before finally deleting the folder itself. Maybe it is just the safest, best option for now.

*No Duplicate Folders*
- What do we want to happen when a user attempts to create a folder with the same name as an existing folder? (Currently: folder creation drawer submits without error, no notice is provided to user about an existing folder with that name, and new folder is not created; the existing folder and objects inside it persist.)

bnussman-akamai:
@mjac0bs 

- I don't think we need TTC for Folder Deletion because you can't delete a folder with objects in it
- You can't delete folders because "folders" are not really a thing in Object Storage and we'd have to implement some kind of recursive delete which would probably be out of the scope of this ticket. 
- I can probably add a manual check to see if the folder object exists before trying to create it
- I can add some styles for the Create button

packages/manager/src/features/ObjectStorage/BucketDetail/BucketDetail.tsx
```diff
@@ -159,6 +171,19 @@ export const BucketDetail: React.FC = () => {
...
 
+    if (objectToDelete.endsWith('/')) {
+      const itemsInFolder = await getObjectList(clusterId, bucketName, {
+        prefix: objectToDelete,
+        delimiter: OBJECT_STORAGE_DELIMITER,
+      });
+
+      if (itemsInFolder.data.length > 1) {
+        setDeleteObjectLoading(false);
+        setDeleteObjectError('The folder must be empty to delete it');
```
mjac0bs, line 182: Can we add a period to the end of this sentence?

mjac0bs (approved):
Thanks for adding the spacing and validation. Tested the duplicate folder creation and confirmed I did see that error alerting users to an already existing folder (within the same directory level). Folders within folders also seem to be working well.

hana-linode (commented):
I'm not seeing an error message when trying to delete a folder with an object inside 🤔 

https://user-images.githubusercontent.com/115299789/219127792-9bb69bc6-6255-4587-8272-f1889fc74881.mov

bnussman-akamai:
@hana-linode Can you provide some steps to reproduce that?

https://user-images.githubusercontent.com/115251059/219136918-a716ad8d-390d-4888-87ce-f560cbdec665.mov

hana-linode:
@bnussman It seems to be when you upload a file within a folder within a folder

mjac0bs:
https://user-images.githubusercontent.com/114685994/219146633-ae3ea2e3-9b4b-4b41-a133-2faa39f45da6.mov



@hana-linode I did test that and still seem to be seeing the error as expected. 🤔 Am I missing something?

hana-linode:
@mjac0bs try having only the folder test in folder-test (i.e. delete Akamai logo 886x306.png)

hana-linode (approved):
Nice work! 🚀 📂

PR #8760, opened 2023-02-03
M3-6086

hkhalil-akamai:


packages/manager/src/components/PreferenceToggle/PreferenceToggle.tsx
```diff
@@ -53,6 +54,10 @@ const PreferenceToggle: React.FC<CombinedProps> = (props) => {
   >(value);
   const [lastUpdated, setLastUpdated] = React.useState<number>(0);
 
+  React.useEffect(() => {
+    getUserPreferences();
+  }, []);
+
```
bnussman-akamai, line 60: I think this is causing **many** fetches to the preferences endpoint. Maybe we need to do something like the following


```ts
  React.useEffect(() => {
    if (!preferences && !preferencesLoading) {
      getUserPreferences();
    }
  }, []);
```

Screen Shot 2023-02-08 at 4 11 37 PM

bnussman-akamai (commented):
Great clean up to make all of this work as it should. This will be much nicer when we get around to React Queryify all of this. Left one item of feedback that should be fixable

PR #8762, opened 2023-02-06
Fix Uneven Breakpoints in Kubernetes Landing Table

bnussman-akamai:
## Description 📝

- FIxes the messed up breakpoints on the Kubernetes Landing Table
- This was likely caused by a combontaion of MUI v5 and React Query for Kubernetes and some conflicts along the way

## Before 
Screen Shot 2023-02-06 at 10 54 45 AM

### After
Screen Shot 2023-02-06 at 10 54 30 AM

## How to test 🧪

- Go to `http://localhost:3000/kubernetes/clusters`
- Have at least one Kubernetes cluster on your account
- Test resizing the viewport and verify the table breaks as you would expect as you change the viewport


hana-linode (commented):
It seems like we're missing sorting now for the Total Memory and Total CPUs column. And the breakpoint now is displaying different columns from prod, is this intentional? 🤔 

This branch:

image

image


Prod:
image

image

bnussman-akamai:
We can't sort on `Total Memory` and `Total CPUs` any more because we are API paginating and sorting this table. I will update breakpoints to match production.

cpathipa (approved):
@bnussman Thank you for fixing this.

PR #8765, opened 2023-02-06
Fix CI by adding new regions `label` field

bnussman-akamai:
## Description 📝

- Fixes our internal CI (makes typecheck pass) because a new field `label` was added to the regions endpoint
- we were getting a static typescript error because we cache a regions.json file and tsc was able to see a difference in the expected types

## How to test 🧪

- Test region related feature of cloud manager
- Make sure all tests pass and internal builds pass


packages/api-v4/src/regions/types.ts
```diff
@@ -18,6 +18,7 @@ export type RegionStatus = 'ok' | 'outage';
 
 export interface Region {
   id: string;
+  label: string;
```
cpathipa, line 21: Is this the source of truth for regions label? If so can we use them in `RegionSelect.tsx`

bnussman-akamai: Yes. A follow up to this would be to remove the `dcDisplayNames` map because the API now returns the region names (like Newark, NJ), but I think we should do this in a separate PR. `dcDisplayNames` is used in **many** places

cpathipa (approved):
LGTM!

dwiley-akamai (approved):

PR #8768, opened 2023-02-07
M3-6079: Creating a Linode via a StackScript - Require `image` Field

bnussman-akamai:
## Description 📝

- Updates the `CreateLinodeSchema` in `validation` to match the Linode API. 
- `image` is now required by the API when you specify a `stackscript_id`, we need to update our client side validation to match this API behavior
- This PR also passes the `errors` prop to both `FromStackScriptContent` panels because they did not previously have errors data passed to them, resulting in image errors not being surfaced
 
### Before
Screen Shot 2023-02-07 at 2 35 22 PM

### After
Screen Shot 2023-02-10 at 10 00 40 AM

## How to test 🧪

- Test Linode Creation
  - Test that you see an error under the image field if you try to deploy a Linode from a Stackscript with no image selected
  - Test all other ways to deploy a Linode and ensure that no other flows were adversely affected

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
@@ -511,6 +511,7 @@ export class LinodeCreate extends React.PureComponent<
                   // error={hasErrorFor.image}
                   accountBackupsEnabled={accountBackupsEnabled}
                   userCannotCreateLinode={userCannotCreateLinode}
+                  errors={errors}
                   {...rest}
                 />
               </SafeTabPanel>
--- a/packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
@@ -530,6 +530,7 @@ export class LinodeCreate extends React.PureComponent<
                           imagesData={imagesData!}
                           regionsData={regionsData!}
                           typesData={typesData!}
+                          errors={errors}
                           {...rest}
                         />
                       </SafeTabPanel>
@@ -543,6 +544,7 @@ export class LinodeCreate extends React.PureComponent<
                           imagesData={imagesData!}
                           regionsData={regionsData!}
                           typesData={typesData!}
+                          errors={errors}
                           {...rest}
                         />
                       </SafeTabPanel>
--- a/packages/validation/src/linodes.schema.ts
+++ b/packages/validation/src/linodes.schema.ts
@@ -84,7 +84,14 @@ export const CreateLinodeSchema = object({
   stackscript_id: number().notRequired(),
   backup_id: number().notRequired(),
   swap_size: number().notRequired(),
-  image: string().notRequired(),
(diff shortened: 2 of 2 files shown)


cpathipa (commented):
Observation:
Noticed that Image field is required to create Linode. Is this expected behavior? 
I think It should be a required filed creating via StackScrip right
image

bnussman-akamai:
@cpathipa Can you provide some steps to reproduce? I can't get into that state on the Linode create page.

bnussman-akamai:
@cpathipa, made a small change to the condition. Can you see if that fixed it?

mjac0bs (commented):
Tested:
- Creating a linode via StackScript: shows validation error without image selected ✔️ 
- Creating a linode via Linode Create form: with the latest changes, does not show validation error without image select, as expected ✔️ 
- Creating a linode from OCA: does not show an error without image selected but also does not create the linode  ❓ -- Is this user error on my part, are we not addressing OCAs here, or should we be accounting for this situation in this PR?

Steps to repro:
- Navigate to Linodes Create > Marketplace and select an OCA like Gopaddle.
- Remove the prepopulated image from the dropdown selection.
- Make any other required selections.
- Click Create Linode and observe that the linode does not create and no validation or error feedback is given in the UI.

bnussman-akamai:
Great catch @mjac0bs, thank you. I'm assuming this is due to the `errors` prop not getting drilled to the correct component. (Refer to the changes I had to make in `packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx`). I'm thinking it will be a similar fix.

mjac0bs (commented):
> Creating a linode from OCA: does not show an error without image selected but also does not create the linode 

Confirming that the latest change fixes this issue; the validation error message displays below the image field as expected. 

One existing thing I noticed when testing: it is possible both in prod and this branch to create a linode from an image via Linodes Create > Images with no selected image. Since `image` is not required by API, I guess this is correct, though it seems counter-intuitive to me given the purpose of that page. 🤔 Any thoughts or context I'm missing?

bnussman-akamai:
@mjac0bs . I am all for mimicking API behavior, but I agree that is definitely weird. For the sake of the scope of this PR, that probably can be handled as a separate issue.

mjac0bs (approved):

cpathipa:
Observation:
Noticed that root pass is disabled not able to enter data when image is not selected. Attached screen capture for reference. 

https://user-images.githubusercontent.com/119517080/218166512-8861e650-37ac-45a2-8558-ed5dd6d0fa99.mov

bnussman-akamai:
@cpathipa  I have noticed that too. It appears to be an existing behavior, so for the sake of this PR's scope, I think we should leave it. 

I think if we were to make a change, root password should **never** be disabled.

bnussman-akamai:
I usually would be willing to make out of scope changes like that, but considering how clunky and sensitive Linode create is, I don't want to make too many changes.

cpathipa (approved):
@bnussman Thank you for fixing the issue. Confirming that Linode create is working as expected via Stack script and Linode Create.

PR #8769, opened 2023-02-07
M3-6009: Add Cypress integration tests for Personal Access Token CRUD

jdamore-linode:
## Description 📝

**What does this PR do?**
This adds a couple integration tests for Cloud Manager PAT CRUD operations. It confirms that UI flows work as expected when users create, rename, and revoke personal access tokens. I opted to mock all of the data involved rather than create actual PATs and risk them leaking into logs, recordings, etc.

## How to test 🧪
Run `yarn up` and then:

```bash
yarn cy:run -s "cypress/e2e/account/personal-access-tokens.spec.ts"
```

### DIFF
--- a/packages/manager/src/features/Profile/APITokens/APITokenTable.tsx
+++ b/packages/manager/src/features/Profile/APITokens/APITokenTable.tsx
@@ -223,7 +223,7 @@ export const APITokenTable = (props: Props) => {
           )}
         </Grid>
       </Grid>
-      <Table aria-label="List of Personal Access Tokens">
+      <Table aria-label={`List of ${title}`}>
         <TableHead>
           <TableRow data-qa-table-head>
             <TableSortCell
--- a/packages/manager/src/features/Profile/APITokens/RevokeTokenDialog.tsx
+++ b/packages/manager/src/features/Profile/APITokens/RevokeTokenDialog.tsx
@@ -32,7 +32,7 @@ export const RevokeTokenDialog = ({ open, onClose, token, type }: Props) => {
   const onRevoke = () => {
     mutateAsync().then(() => {
       onClose();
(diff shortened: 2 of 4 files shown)


packages/manager/src/features/Profile/APITokens/APITokenTable.tsx
```diff
@@ -223,7 +223,7 @@ export const APITokenTable = (props: Props) => {
           )}
         </Grid>
       </Grid>
-      <Table aria-label="List of Personal Access Tokens">
+      <Table aria-label={`List of ${title}`}>
```
jdamore-linode, line 226: This results in the _Third Party Access Tokens_ table getting an `aria-label` value of `List of Third Party Access Tokens` instead of `List of Personal Access Tokens`. I don't believe there are any other places where this component is used where this change would cause any other impact.

packages/manager/src/features/Profile/APITokens/RevokeTokenDialog.tsx
```diff
@@ -32,7 +32,7 @@ export const RevokeTokenDialog = ({ open, onClose, token, type }: Props) => {
   const onRevoke = () => {
     mutateAsync().then(() => {
       onClose();
-      enqueueSnackbar(`Sucessfully revoked ${token?.label}`, {
+      enqueueSnackbar(`Successfully revoked ${token?.label}`, {
```
bnussman-akamai, line 35: Thanks :)

bnussman-akamai (approved):
Nice, thanks for getting around to these tests so quickly!

cpathipa (approved):
Nice work @jdamore-linode Thank you for covering  tests for Personal Access Token.
Confirming that specs passed 
image

PR #8774, opened 2023-02-08
Fix tags with spaces in cli

CodyFinn-Akamai:
## Description 📝

Wraps the tags in CLI commands in case they have spaces in them

1. Navigate to create a new linode
2. Add a tag with just spaces in its name like `tag with spaces`
3. Add a tag with paces and special characters in its name like `\tag\/with&special-+&chars`
4. Click `Create Using Command Line`
5. Click the Linode CLI tab
6. Copy the linode-cli command and try to run it
7. Ensure that the linode is successfully created
8. Ensure that the tags on the linode are as typed in the create form

### DIFF
--- a/packages/manager/src/utilities/generate-cli.ts
+++ b/packages/manager/src/utilities/generate-cli.ts
@@ -1,14 +1,4 @@
-const escapeStringForCLI = (value: string) => {
-  const doesContainEscapableChars = value.match(/[^\w\s]/gi);
-  let parsedValue = value;
-  if (doesContainEscapableChars) {
-    parsedValue = value.replace(/[^\w\s]/gi, function (char) {
-      return '\\' + char;
-    });
-    return parsedValue;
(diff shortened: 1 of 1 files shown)


bnussman-akamai:
The problem I see here is that when I use tag `testing’okay/this\is!a&testing’okay`

```
linode-cli linodes create \
  --image linode\/centos\-stream8 \
  --region us\-east \
  --type g6\-nanode\-1 \
  --label centos\-us\-east \
  --tags 'testing\’okay\/this\\is\!a\&testing\’okay' \
  --root_pass testing\’okay\/this\\is\!a\&testing\’okay \
  --booted true \
  --backups_enabled false \
  --private_ip false
```
results in 

Screen Shot 2023-02-08 at 11 59 26 AM

which isnt the expected tag of `testing’okay/this\is!a&testing’okay`

PR #8775, opened 2023-02-08
CLI Command Handle More Cases

bnussman-akamai:
## Description 📝

- Enables CLI commands to work with special characters in the root password or tags
- Examples Include
   - `testing’okay/this\is!a&testing’okay` 

## How to test 🧪

- Test CLI creation with special characters and spaces in the root password or tags


bnussman-akamai:
vetoed for UX reasons

PR #8778, opened 2023-02-09
Fix dark mode illegible text, undo button spacing on newly added rules

mjac0bs:
## Description 📝

**What does this PR do?**
Fixes a previous fix that was missed in #8776: due to applying the text styling to the wrong CSS rule, a newly created firewall rule row would should up in dark mode with black text rather than `theme.textColors.tableStatic`. 

A set height of 40px for a firewall rule row also caused a strange issue, noticeable in dark mode, for the undo button of newly added firewall rule(s), where the button had a height of 41px and extended beyond the container. Unsetting the row `height` seemed to fix the issue.

## Preview 📷

Fix:

https://user-images.githubusercontent.com/114685994/217683444-9cfca002-125c-466f-a9f2-ab0ee69412bf.mov

Staging:

https://user-images.githubusercontent.com/114685994/217683459-b89f15ab-5e19-4384-adf6-96c90932cb69.mov

## How to test 🧪

Use dark mode and Add an Inbound (or Outbound) Firewall Rule. Ensure the rule displays with legible text in both dark and light mode. Ensure the "undo" button doesn't stick out beyond the bottom of the row.

### DIFF
--- a/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleTable.tsx
+++ b/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleTable.tsx
@@ -70,7 +70,6 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
   unmodified: {
     backgroundColor: theme.bg.bgPaper,
-    color: theme.textColors.tableStatic,
   },
   highlight: {
     backgroundColor: theme.bg.lightBlue1,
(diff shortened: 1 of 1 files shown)


cpathipa (approved):
LGTM!

dwiley-akamai (approved):
The colors and undo button spacing LGTM now.

Not sure if it's related to the previous PR, but I did notice that the policy text starts wrapping very early now (even when the viewport width >1350px) compared to Prod:

Screen Shot 2023-02-09 at 10 17 57 AM

mjac0bs:
> Not sure if it's related to the previous PR, but I did notice that the policy text starts wrapping very early now (even when the viewport width >1350px) compared to Prod:

@dwiley-akamai Thanks for pointing this out. This is likely happened with to the transformation of table -> grid and trying to mimic spacing and alignment. I can have another look at those breakpoints. For now I'm going ahead and merging this PR so as not to hold up the release.

PR #8779, opened 2023-02-09
M3-6094: Add role attribute to top menu icons

carrillo-erik:
## Description 📝

Improve's Cloud Manager's accessibility score in Lighthouse report by adding the `role` attribute to the TopMenuIcon component and appropriately setting its value to `link` or `button`.

## Preview 📷
**Before:**
before-changes

**After:**
after-changes

## How to test 🧪
Run the Lighthouse report on local branch, then checkout this PR and re-run the Lighthouse report and compare accessibility scores.

### DIFF
--- a/packages/manager/src/features/TopMenu/TopMenuIcon.tsx
+++ b/packages/manager/src/features/TopMenu/TopMenuIcon.tsx
@@ -4,7 +4,6 @@ import Tooltip from 'src/components/core/Tooltip';
 interface Props {
   title: string;
   children: JSX.Element;
-  role: string;
 }
 
 export const TopMenuIcon: React.FC<Props> = (props) => {
@@ -15,7 +14,7 @@ export const TopMenuIcon: React.FC<Props> = (props) => {
       disableTouchListener
       enterDelay={500}
       leaveDelay={0}
-      role={'link' || 'button'}
+      describeChild={true}
     >
       <div>{props.children}</div>
     </Tooltip>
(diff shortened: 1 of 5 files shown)


packages/manager/src/features/TopMenu/TopMenuIcon.tsx
```diff
@@ -4,12 +4,19 @@ import Tooltip from 'src/components/core/Tooltip';
...
 export const TopMenuIcon: React.FC<Props> = (props) => {
   const { title } = props;
   return (
-    <Tooltip title={title} disableTouchListener enterDelay={500} leaveDelay={0}>
+    <Tooltip
+      title={title}
+      disableTouchListener
+      enterDelay={500}
+      leaveDelay={0}
+      role={'link' || 'button'}
```
jdamore-linode, line 18: Should this be `role={role}`? My understanding is `role` will always be `'link'` and whatever actually gets specified for that prop is disregarded.

cpathipa, line 18: can we consider role 'tooltip' as well ?

carrillo-erik:
Thanks for looking into this and providing the informative links. I was able to get the same Lighthouse score implementing the `aria-describedby` prop as you mentioned. I will update the PR.

dwiley-akamai (approved):
I got a score of 98 with the `describeChild` prop added. 🚀

jdamore-linode (approved):

cpathipa (approved):

PR #8783, opened 2023-02-09
M3-6086

hkhalil-akamai:


hana-linode (commented):
Splash screen lgtm, I think we should feature flag this though

dwiley-akamai (approved):
LGTM -- per the comments on M3-6082 I think we're fine to not feature flag this piece

PR #8786, opened 2023-02-10
M3-5807: Don't Reset the Firewall Rule Drawer if the same preset is reselected

bnussman-akamai:
## Description 📝

- If the user reselectes the **same preset** in the firewalls rule drawer, **do not reset the form** to the preset defaults
  - This was causing users to accidentally loose their progress when they reselected the same preset

### Before
Note how form resets when the same preset is selected resulting in the users progress being lost

https://user-images.githubusercontent.com/115251059/218135926-12711216-1df3-44aa-900b-70f2005341f9.mov

### After
Note how the form does not change state at all

https://user-images.githubusercontent.com/115251059/218135792-dc2c847c-4fd5-4618-a57a-af2940e1fdda.mov

## How to test 🧪

- Test the behavior and functionality of the Firewall Rules drawer

### DIFF
--- a/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.tsx
+++ b/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.tsx
@@ -278,6 +278,12 @@ const FirewallRuleForm: React.FC<FirewallRuleFormProps> = React.memo(
     const handleTypeChange = React.useCallback(
       (item: Item | null) => {
         const selectedType = item?.value;
+
+        // If the user re-selectes the same preset, don't do anything
+        if (selectedType === values.type) {
+          return;
+        }
+
         setFieldValue('type', selectedType);
 
         if (!selectedType) {
--- a/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.tsx
+++ b/packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.tsx
@@ -318,6 +318,7 @@ const FirewallRuleForm: React.FC<FirewallRuleFormProps> = React.memo(
         category,
         setPresetPorts,
         values.action,
+        values.type,
       ]
     );
 


jdamore-linode (approved):
Nice! Just confirmed that your latest commit fixes the issue I was seeing 👍

mjac0bs (commented):
> If you open the drawer and select a preset, fill out some fields, and then select another preset, everything you have filled out should be reset/replaced by the defaults for your new preset -- in other words, you're discarding your changes when you select a new preset

I would also have the same expectation as @jdamore-linode for the case above, and I'm seeing different behavior currently.

For example, if I:

- Select an initial preset
- Fill out all fields, including adding a Description and changing the Action selection from `Accept` to `Drop`
- Select a different preset
- Observe that the Description and Action changes persist from the previous preset's form fields; although Label, Protocol, and Sources all reset as expected 

Edit: It looks like this is the existing behavior in prod, although unexpected.
The behavior for not resetting when a same preset is reselected LGTM.

bnussman-akamai:
@mjac0bs Do you think I should change that behavior as part of this PR?

mjac0bs (approved):
I'm going to approve this since the couple of form fields not clearing is existing behavior and the desired change for this PR has been made -- but I think we should either fix the unexpected behavior in this PR if it's a small change or make another ticket to fix it if it's more complex or time-consuming.

packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleDrawer.tsx
```diff
@@ -278,6 +278,12 @@ const FirewallRuleForm: React.FC<FirewallRuleFormProps> = React.memo(
...
         const selectedType = item?.value;
+
+        // If the user re-selectes the same preset, don't do anything
+        if (selectedType === values.type) {
+          return;
+        }
+
         setFieldValue('type', selectedType);
 
         if (!selectedType) {
```
cpathipa, line 289: Can we combine both these conditions ?

```
 if (selectedType === values.type || !selectedType) {
          return;
  }
 setFieldValue('type', selectedType);
```

cpathipa (approved):

PR #8788, opened 2023-02-10
M3-6161

cpathipa:
## Description 📝

**What does this PR do?**

## Preview 📷

## How to test 🧪

### DIFF
--- a/packages/manager/src/features/CancelLanding/CancelLanding.tsx
+++ b/packages/manager/src/features/CancelLanding/CancelLanding.tsx
@@ -1,9 +1,11 @@
-import { path } from 'ramda';
+// import { path } from 'ramda';
 import * as React from 'react';
-import { Redirect, useLocation } from 'react-router-dom';
+// import { useLocation } from 'react-router-dom';
 import { compose } from 'recompose';
+import useFlags from 'src/hooks/useFlags';
 import { makeStyles, Theme } from 'src/components/core/styles';
 
+import AkamaiLogo from 'src/assets/logo/akamai-logo.svg';
 import Logo from 'src/assets/logo/logo-footer.svg';
 import Button from 'src/components/Button';
 import Typography from 'src/components/core/Typography';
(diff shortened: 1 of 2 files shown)


packages/manager/src/features/CancelLanding/CancelLanding.tsx
```diff
@@ -37,7 +43,10 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
 }));
 
+type CombinedProps = FeatureFlagConsumerProps;
+
 export const CancelLanding: React.FC<{}> = () => {
```
dwiley-akamai, line 48:
```suggestion
export const CancelLanding: React.FC<CombinedProps> = (props) => {
```

packages/manager/src/features/CancelLanding/CancelLanding.tsx
```diff
@@ -2,12 +2,18 @@ import { path } from 'ramda';
 import * as React from 'react';
 import { Redirect, useLocation } from 'react-router-dom';
 import { compose } from 'recompose';
+import useFlags from 'src/hooks/useFlags';
```
dwiley-akamai, line 5: since we're having to bring the feature flag wrappers in, let's just rely on those fully instead of using the hook

packages/manager/src/features/CancelLanding/CancelLanding.tsx
```diff
@@ -37,7 +43,10 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
 }));
 
+type CombinedProps = FeatureFlagConsumerProps;
+
 export const CancelLanding: React.FC<{}> = () => {
+  const flags = useFlags();
```
dwiley-akamai, line 49:
```suggestion
const { flags } = props;
```

(made available by the flag wrapper/HOC -- let's just rely on that instead of using the `useFlags()` hook)

dwiley-akamai (approved):
I think we should make a few minor changes noted below, but approving since this is working now

packages/manager/src/features/CancelLanding/CancelLanding.tsx
hana-linode, line 5: Agreed with Dajahi, we can remove this line since it's no longer being used
```suggestion
```

packages/manager/src/features/CancelLanding/CancelLanding.tsx
```diff
@@ -37,7 +43,10 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
 }));
 
-export const CancelLanding: React.FC<{}> = () => {
+type CombinedProps = FeatureFlagConsumerProps;
+
+export const CancelLanding: React.FC<CombinedProps> = (props) => {
+const { flags } = props;
```
hana-linode, line 49: indentation

hana-linode (commented):
Looks like there's some TS issues with the `CancelLanding.test.tsx` file

image

packages/manager/src/features/CancelLanding/CancelLanding.tsx
hana-linode, line 49:
```suggestion
  const { flags } = props;
```

hana-linode (approved):

PR #8790, opened 2023-02-13
M3-6163: Standardize Text Tooltip

mjac0bs:
## Description 📝

**What does this PR do?**
Makes a few updates to standardize the tooltip by doing the following:
- Renames from `BillingTooltip` to a more generic `TextTooltip`
- Removes const billing text and replace with prop for popover text
- Replicates default behavior for triggering tooltips on mobile – our other tooltips using button components open popover immediately on click; don't require a long-press which is the MUI Tooltip default
- Also fixes a small alignment issue in the table header at `sm` screen breakpoint; `xs` breakpoint was previously used and should not have been

Previously, this was only used as a billing notice for Akamai customers, but we expect to expand its use.

## Preview 📷

Tooltip in non-mobile view should behave and look unchanged:
Screen Shot 2023-02-13 at 11 25 21 AM

<details>
<summary> Screencaps 📹  </summary>

**Before** (note the long-press)

https://user-images.githubusercontent.com/114685994/218554530-f87f9b20-a943-46bd-87e3-2c56f37b2e38.mov

**After**

https://user-images.githubusercontent.com/114685994/218554508-2b1948c6-1478-4b85-9022-17b3915dddf7.mov

</details>

## How to test 🧪

Since the Billing Activity Panel is currently the only existing place in the app that the tooltip exists, follow the same steps to replicate as #8690: 
1. In `src/factories/account.ts`, change the `billing_source` from `linode` to `akamai` and then turn on the MSW to mock an Akamai-billed account.
2. Navigate to http://localhost:3000/account/billing.
3. Observe that the billing table header copy is "Usage History", a usage history tooltip with a dashed underline displays next to the header, and hovering over tooltip displays updated copy with link below the tooltip. 
4. Use dev tools to toggle device to a couple different mobile devices, at least one of them with a small screen. Click on the tooltip in mobile view and observe that the popover text is triggered immediately, rather than after a delay. Compare this to the behavior when clicking help tooltips elsewhere on the Account page to confirm tooltip behavior is consistent.
5. Observe that on small screens (`xs` and `sm` breakpoint), table header components are consistently aligned to the left, including the tooltip.

### DIFF
--- a/packages/manager/src/components/TextTooltip/TextTooltip.tsx
+++ b/packages/manager/src/components/TextTooltip/TextTooltip.tsx
@@ -1,12 +1,11 @@
 import * as React from 'react';
-import ExternalLinkIcon from 'src/assets/icons/external-link.svg';
 import { makeStyles, Theme } from 'src/components/core/styles';
 import ToolTip from 'src/components/core/Tooltip';
 import Typography from 'src/components/core/Typography';
-import Link from 'src/components/Link';
 
 interface Props {
-  text: string;
+  displayText: string;
+  tooltipText: JSX.Element | string;
 }
 
 const useStyles = makeStyles((theme: Theme) => ({
@@ -29,33 +28,21 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
 }));
 
-const akamaiBillingInvoiceText = (
-  <Typography>
-    Charges in the final Akamai invoice should be considered the final source
-    truth. Linode invoice will not reflect discounting, currency adjustment, or
-    any negotiated terms and conditions. Condensed and finalized invoice is
-    available within{' '}
-    <Link to="https://control.akamai.com/apps/billing">
-      Akamai Control Center &gt; Billing <ExternalLinkIcon />
-    </Link>
-    .
-  </Typography>
-);
-
-export const BillingTooltip: React.FC<Props> = (props) => {
+export const TextTooltip: React.FC<Props> = (props: Props) => {
   const classes = useStyles();
-  const { text } = props;
+  const { displayText, tooltipText } = props;
 
   return (
     <ToolTip
-      title={akamaiBillingInvoiceText}
+      title={tooltipText}
       placement="bottom"
+      enterTouchDelay={0}
       className={classes.root}
       classes={{ popper: classes.popper }}
     >
-      <Typography>{text}</Typography>
+      <Typography>{displayText}</Typography>
     </ToolTip>
   );
 };
 
-export default BillingTooltip;
+export default TextTooltip;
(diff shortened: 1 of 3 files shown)


packages/manager/src/components/TextTooltip/TextTooltip.tsx
```diff
@@ -29,33 +28,21 @@ const useStyles = makeStyles((theme: Theme) => ({
...
-    available within{' '}
-    <Link to="https://control.akamai.com/apps/billing">
-      Akamai Control Center &gt; Billing <ExternalLinkIcon />
-    </Link>
-    .
-  </Typography>
-);
-
-export const BillingTooltip: React.FC<Props> = (props) => {
+export const TextTooltip: React.FC<Props> = (props: Props) => {
```
cpathipa, line 31: Can we remove React.FC ?
```suggestion
export const TextTooltip = (props: Props) => {
```

cpathipa (commented):
Nice work @mjac0bs I left minor comments that are totally nit-picky things.
Observation: Tool tip modal is close towards left side in SM screen. Can we add some spacing to align with the content. 

image

mjac0bs:
> Observation: Tool tip modal is close towards left side in SM screen. Can we add some spacing to align with the content.

@cpathipa I couldn't really come up with clean way to do this - it would require adjusting (probably calculating) the left margin and possibly midWidth of the `MuiTooltip-tooltip` class. (Adding even a little margin also makes a difference on some `xs` devices because the right side of the popper then goes off the screen.) Since we'd only want to adjust modal spacing for this particular billing TextTooltip, we'd have to conditionally apply styling at the `sm` breakpoint for `classes.popper` and add more complexity to the component. I'm inclined to leave it as it, unless there are strong feelings otherwise.

packages/manager/src/features/Billing/BillingPanels/BillingActivityPanel/BillingActivityPanel.tsx
```diff
@@ -203,6 +204,19 @@ export const BillingActivityPanel: React.FC<Props> = (props) => {
...
+      Charges in the final Akamai invoice should be considered the final source
+      truth. Linode invoice will not reflect discounting, currency adjustment,
+      or any negotiated terms and conditions. Condensed and finalized invoice is
+      available within{' '}
+      <Link to="https://control.akamai.com/apps/billing">
+        Akamai Control Center &gt; Billing <ExternalLinkIcon />
+      </Link>
+      .
+    </Typography>
+  );
```
hana-linode, line 218: This can be defined outside the component

hana-linode (approved):
Nice improvement on the mobile experience!

cpathipa (approved):
Nice work! @mjac0bs

PR #8795, opened 2023-02-14
M3-5195: RQ-ify user preferences

hkhalil-akamai:
## Description 📝

Replaces the Preferences Redux store with a React Query hook. Also includes a refactor of `LinodeThemeWrapper` and `PreferenceToggle` to avoid drilling and unnecessary complexity.

## How to test 🧪

Ensure all component/hooks that use user preferences still work as expected:

- `App`
- `DatabaseSettingsDeleteClusterDialog`
- `DeleteKubernetesClusterDialog`
- `DeletionDialog`
- `EnableObjectStorage`
- `LinodeResize`
- `LinodeThemeWrapper`
- `MainContent`
- `MainContentBanner`
- `OrderBy`
- `PreferenceEditor`
- `PreferenceToggle`
- `RebuildFromImage`
- `RebuildFromStackScript`
- `RestoreFromBackupDialog`
- `Settings`
- `TimeRangeSelect`
- `TypeToConfirmDialog`
- `useDismissibleNotifications`
- `useOrder`
- `usePagination`

`yarn test`


packages/manager/src/features/Help/StatusBanners.test.tsx
```diff
@@ -31,7 +31,7 @@ describe('Status banners from statuspage.io', () => {
       expect(screen.getByRole('link')).toHaveAttribute('href', props.href);
     });
 
-    it.skip("clicking a banner's close icon should remove it from view", () => {
+    it("clicking a banner's close icon should remove it from view", () => {
```
hkhalil-akamai, line 34: [Context](https://github.com/linode/manager/pull/8803#discussion_r1109086845)

packages/manager/src/App.tsx
```diff
@@ -94,7 +99,10 @@ export class App extends React.Component<CombinedProps, State> {
       if (event[modifierKey] && event.shiftKey) {
         switch (event.key) {
           case letterForThemeShortcut:
-            this.props.toggleTheme();
+            this.props.updateUserPreferences({
+              ...this.props.preferences,
```
bnussman-akamai, line 103: Do we need to spread and include the old preferences data here? I think the mutation handler should take care of that

```suggestion
```

packages/manager/src/components/DeletionDialog/DeletionDialog.tsx
```diff
@@ -114,5 +112,5 @@ const DeletionDialog: React.FC<CombinedProps> = (props) => {
 
 export default compose<CombinedProps, Props>(
   React.memo,
-  withPreferences()
+  withPreferences
```
bnussman-akamai, line 115: Can we just use the `usePreferences` hook in this component?

packages/manager/src/components/MainContentBanner.tsx
```diff
@@ -104,6 +104,6 @@ const MainContentBanner: React.FC<CombinedProps> = (props) => {
   );
 };
 
-const enhanced = compose<CombinedProps, Props>(React.memo, withPreferences());
+const enhanced = compose<CombinedProps, Props>(React.memo, withPreferences);
 
 export default enhanced(MainContentBanner);
```
bnussman-akamai, line 109: Can we use hooks in this component instead of `withPreferences`?

```suggestion
export default React.memo(MainContentBanner);
```

packages/manager/src/containers/preferences.container.ts
```diff
@@ -1,39 +1,36 @@
...
+const withPreferences = <Props>(
+  Component: React.ComponentType<
+    Props & PreferencesStateProps & PreferencesActionsProps
+  >
+) => (props: Props) => {
+  const { data: preferences, refetch } = usePreferences();
+  const { mutateAsync: updateUserPreferences } = useMutatePreferences();
+  if (preferences === undefined) {
+    return null;
+  }
```
bnussman-akamai, line 26: I don't know if it is great to block rendering if `preferences` is `undefined`. Is this necessary?

packages/manager/src/features/Databases/DatabaseDetail/DatabaseBackups/RestoreFromBackupDialog.tsx
```diff
@@ -115,6 +115,6 @@ export const RestoreFromBackupDialog: React.FC<CombinedProps> = (props) => {
   );
 };
 
-const enhanced = compose<CombinedProps, Props>(withPreferences());
+const enhanced = compose<CombinedProps, Props>(withPreferences);
 
 export default enhanced(RestoreFromBackupDialog);
```
bnussman-akamai, line 120: Let's use the hooks in this component too.

```suggestion
export default RestoreFromBackupDialog;
```

packages/manager/src/features/Databases/DatabaseDetail/DatabaseSettings/DatabaseSettingsDeleteClusterDialog.tsx
```diff
@@ -127,6 +127,6 @@ export const DatabaseSettingsDeleteClusterDialog: React.FC<CombinedProps> = (
   );
 };
 
-const enhanced = compose<CombinedProps, Props>(withPreferences());
+const enhanced = compose<CombinedProps, Props>(withPreferences);
 
 export default enhanced(DatabaseSettingsDeleteClusterDialog);
```
bnussman-akamai, line 132: Use hooks in this component

```suggestion
export default DatabaseSettingsDeleteClusterDialog;
```

packages/manager/src/features/Profile/Settings/Settings.tsx
```diff
@@ -37,6 +34,14 @@ const ProfileSettings: React.FC<Props & { theme: Theme }> = (props) => {
   const { data: profile } = useProfile();
   const { mutateAsync: updateProfile } = useMutateProfile();
 
+  const { data: preferences } = usePreferences();
+  const { mutateAsync: updatePreferences } = useMutatePreferences();
+
+  const toggleTheme = () => {
+    const newTheme = preferences?.theme === 'dark' ? 'light' : 'dark';
+    updatePreferences({ ...preferences, theme: newTheme });
```
bnussman-akamai, line 42: Same feedback as earlier. Do we need to spread preferences here (`...preferences`) if the mutation handler takes care of that?

packages/manager/src/features/linodes/LinodesDetail/LinodeRebuild/RebuildFromStackScript.tsx
```diff
@@ -389,7 +389,7 @@ const enhanced = compose<CombinedProps, Props>(
   withSnackbar,
   withImages(),
   withRouter,
-  withPreferences()
+  withPreferences
```
bnussman-akamai, line 392: Use Hooks in the component

packages/manager/src/features/linodes/LinodesDetail/LinodeRebuild/RebuildFromImage.tsx
```diff
@@ -245,7 +245,7 @@ export const RebuildFromImage: React.FC<CombinedProps> = (props) => {
 
 const enhanced = compose<CombinedProps, Props>(
   withImages(),
-  withPreferences(),
+  withPreferences,
```
bnussman-akamai, line 248: use Hooks in this component

packages/manager/src/features/Profile/Settings/Settings.tsx
```diff
@@ -47,15 +52,12 @@ const ProfileSettings: React.FC<Props & { theme: Theme }> = (props) => {
...
-      })
-      .catch(() => {
-        setSubmitting(false);
-      });
+    indicatePromise(
+      updateProfile({
+        email_notifications: !profile?.email_notifications,
+      }),
+      setSubmitting
+    );
```
bnussman-akamai, line 60: I see what you are accomplishing with `indicatePromise`,  but to me it hurts readability. If I were new to the repo i'd have to jump to `indicatePromise`'s definition to understand how it works. Maybe the team has some thoughts?

packages/manager/src/queries/preferences.ts
```diff
@@ -0,0 +1,48 @@
...
+import { useMutation, useQuery } from 'react-query';
+import { ManagerPreferences } from 'src/types/ManagerPreferences';
+import { queryClient, queryPresets } from './base';
+
+export const queryKey = 'preferences';
+
+const adaptPreferences = <T>(
+  func: (...args: T[]) => Promise<UserPreferences>
+) => (...args: T[]) =>
+  func(...args).then((preferences) => preferences as ManagerPreferences);
```
bnussman-akamai, line 16: Can you explain what this does?

bnussman-akamai (commented):
This is **amazing** clean up. Good job working through all of the unit tests, I'm sure that was not easy. I left a good bit of minor feedback items.

packages/manager/src/features/Profile/Settings/Settings.tsx
hkhalil-akamai: Yeah the idea was that it can be easy to forget to reset the loading state, leading to infinite loading. But I agree that it impacts readability.

Another (albeit more complicated) way to accomplish this is to globally define a `Promise.prototype.indicate` method.

packages/manager/src/queries/preferences.ts
hkhalil-akamai: This was purely to make TypeScript happy and cast from API's `UserPreferences` type to the new `ManagerPreferences` interface. Removed since this doesn't seem necessary anymore.

jdamore-linode (commented):
Seeing an issue with the preference editor at `/profile/settings?preferenceEditor=true` -- there's a little "Loading..." indicator that appears and disappears, causing a flickering effect, and it seems like each time that happens it resets the contents of the text box, making it difficult (impossible?) to edit:

https://user-images.githubusercontent.com/97627410/220438076-57b926ef-9446-4ec6-be97-d9530c9d1c4d.mp4

(It also seems like each flicker corresponds to a GET request to `profile/preferences`, so be careful not to linger too long on that dialog)

jdamore-linode:
E2E tests pass, and manually went through flows involving all of the components you listed in the PR description. Didn't notice anything aside from that issue above! I'll circle back to test any new changes and approve this once you've had a chance to check out that issue and Banks's suggestions. Awesome work!

packages/manager/src/containers/preferences.container.ts
hkhalil-akamai: Looks like the original behavior was to pass the empty object `{}` while preferences were loading, would that be preferable to render-blocking?

packages/manager/src/App.tsx
hkhalil-akamai: There appears to be some inconsistency between the `profile` and `preferences` APIs. Whereas `updateProfile` only updates the provided fields, `updateUserPreferences` will totally overwrite the existing record with the provided data. Hence the spread is currently necessary.

We could possibly move the spread to the `useMutatePreferences` hook to locally emulate this behavior.

packages/manager/src/features/Profile/Settings/Settings.tsx
hkhalil-akamai: [See above](https://github.com/linode/manager/pull/8795#discussion_r1113612423)

hkhalil-akamai:
Not strictly related to this PR, a concern from testing the `PreferenceEditor` is that users can set the preference blob to invalid values (e.g., setting `theme` to `"not a real theme"`) and thus break the Cloud Manager for themselves. Arguably, the user preferences type should be defined as:

```
export type UserPreferences = Record<string, unknown>;
```

instead of:

```
export type UserPreferences = Record<string, any>;
```

This forces components to validate preference values before using them.

packages/manager/src/containers/preferences.container.ts
bnussman-akamai: I think it would be best for the HOC to match RQ as much as possible, so we just pass whatever React query returns. For example `PreferencesStateProps` would become

```ts
export interface PreferencesStateProps {
  preferences: ManagerPreferences | undefined;
}
```

bnussman-akamai (changes requested):
Found a bug. I think what's happening is that the `updatePreferences` is being called without the old preferences being spread, causing all of the existing preferences to go away. This is in the `useOrder` hook. 

I think the best solution for this would be to do what you recommended so that other consumers of the `useMutatePreferences` don't cause this same bug in the future.

> We could possibly move the spread to the useMutatePreferences hook to locally emulate this behavior.

Something like this (double check what I did here)
```
export const useMutatePreferences = () => {
  const { data: preferences } = usePreferences();

  return useMutation<ManagerPreferences, APIError[], ManagerPreferences>(
    (data) => updateUserPreferences({ ...(preferences ?? {}), data }),
    {
      onMutate: updatePreferenceData,
    }
  );
};
```


https://user-images.githubusercontent.com/115251059/220737563-f55f8c07-ecd7-4773-9ffe-bb2abefdca89.mov

packages/manager/src/queries/preferences.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+export const useMutatePreferences = () => {
+  const { data: preferences } = usePreferences();
+
+  return useMutation<
+    ManagerPreferences,
+    APIError[],
+    Partial<ManagerPreferences>
+  >((data) => updateUserPreferences({ ...(preferences ?? {}), data }), {
```
bnussman-akamai, line 26: Oops, I think I messed you up here. The data value should be spread.

```suggestion
  >((data) => updateUserPreferences({ ...(preferences ?? {}), ...data }), {
```

packages/manager/src/features/Profile/Settings/PreferenceEditor.tsx
```diff
@@ -24,23 +20,24 @@ const PreferenceEditor: React.FC<CombinedProps> = (props) => {
...
+  }, []);
 
   const handleSavePreferences = () => {
     try {
       const parsed = JSON.parse(userPrefs);
       setSubmitting(true);
       setSuccessMessage('');
       setErrorMessage('');
-      updateUserPreferences(parsed as any)
+      updatePreferences(parsed as any)
```
bnussman-akamai, line 40: Type is already `any` so can just remove the cast

```suggestion
      updatePreferences(parsed)
```

bnussman-akamai (approved):
Massive improvement to our codebase. This does a really nice job cleaning up our index and App level files. 🎉

packages/manager/src/queries/preferences.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const updatePreferenceData = (
+  newData: Partial<ManagerPreferences>,
+  replace?: boolean
+): void => {
+  queryClient.setQueryData<ManagerPreferences>(
+    queryKey,
+    (oldData: ManagerPreferences) => ({
+      ...(!replace ? oldData : {}),
+      ...newData,
```
bnussman-akamai, line 46: This just occurred to me, but do you think a deep merge is needed/desirable here? From my testing, this is working fine so I don't think we need to make a change. Just a thought.

packages/manager/src/App.tsx
```diff
@@ -53,7 +56,9 @@ type CombinedProps = Props &
   StateProps &
   RouteComponentProps &
   WithSnackbarProps &
-  FeatureFlagConsumerProps;
+  FeatureFlagConsumerProps &
+  PreferencesStateProps &
```
hkhalil-akamai, line 60: App doesn't use preferences

packages/manager/src/queries/preferences.ts
hkhalil-akamai: This is just to make sure `updatePreferenceData` is consistent with `useMutatePreferences` now with the the `replace` flag.

jaalah-akamai (approved):

jdamore-linode (approved):

PR #8796, opened 2023-02-15
M3-5912: Vite

bnussman-akamai:
## Description 📝

- Use Vite instead of Webpack


bnussman-akamai:
ugh, node version needs to be changed for internal CI

PR #8797, opened 2023-02-15
ga click event for "Create Using Command Line" Button

cpathipa:
## Description 📝

Track events for clicking Create using command line Button

**What does this PR do?**

Tracks click events for Create using command line Button


bnussman-akamai (approved):

mjac0bs (approved):
Confirmed that dev tools show event on button click. 👍  Is this just a ticketless follow up from the initial work?

cpathipa:
> Confirmed that dev tools show event on button click. 👍 Is this just a ticketless follow up from the initial work?

@mjac0bs Good call. I will create a defect for it

PR #8800, opened 2023-02-16
Add Cloud-init compatible check box

cpathipa:
## Description 📝

**What does this PR do?**

image

## How to test 🧪

### DIFF
--- a/packages/manager/src/features/Images/ImageUpload.tsx
+++ b/packages/manager/src/features/Images/ImageUpload.tsx
@@ -1,10 +1,14 @@
 import { APIError } from '@linode/api-v4/lib/types';
 import * as React from 'react';
+import { useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { useHistory } from 'react-router-dom';
 import ActionsPanel from 'src/components/ActionsPanel';
 import Button from 'src/components/Button';
+import CheckBox from 'src/components/CheckBox';
+import HelpIcon from 'src/components/HelpIcon';
 import ConfirmationDialog from 'src/components/ConfirmationDialog';
+import FormControlLabel from 'src/components/core/FormControlLabel';
 import Paper from 'src/components/core/Paper';
 import { makeStyles, Theme } from 'src/components/core/styles';
 import Typography from 'src/components/core/Typography';
@@ -55,6 +59,12 @@ const useStyles = makeStyles((theme: Theme) => ({
     ...theme.applyLinkStyles,
     fontWeight: 700,
   },
+  cloudInitCheckboxWrapper: {
+    marginTop: theme.spacing(2),
+  },
+  cloudInitCheckbox: {
+    marginRight: 0,
+  },
 }));
 export interface Props {
   label: string;
@@ -85,6 +95,11 @@ export const ImageUpload: React.FC<Props> = (props) => {
   const [linodeCLIModalOpen, setLinodeCLIModalOpen] = React.useState<boolean>(
     false
   );
+  const [isCloudInitChecked, setIsCloudInitChecked] = useState<boolean>(false);
+
+  const handleCloudInitChange = () => {
+    setIsCloudInitChecked(!isCloudInitChecked);
+  };
 
   const showAgreement = Boolean(
     !profile?.restricted && agreements?.eu_model === false && isEURegion(region)
@@ -151,6 +166,13 @@ export const ImageUpload: React.FC<Props> = (props) => {
   const cliRegion = formatForCLI(region, 'region');
   const linodeCLICommand = `linode-cli image-upload --label ${cliLabel} --description ${cliDescription} --region ${cliRegion} FILE`;
 
+  const cloudInitTooltipMessage = (
+    <Typography>
+      Copy TBD-Need to have installed cloud-init on the distro and change the
+      config to use our data service. <Link to="/">Link to doc</Link>
+    </Typography>
+  );
+
   return (
     <>
       <Prompt
@@ -213,6 +235,20 @@ export const ImageUpload: React.FC<Props> = (props) => {
             errorText={errorMap.description}
             disabled={!canCreateImage}
           />
+          <div className={classes.cloudInitCheckboxWrapper}>
+            <FormControlLabel
+              className={classes.cloudInitCheckbox}
+              control={
+                <CheckBox
+                  checked={isCloudInitChecked}
+                  onChange={handleCloudInitChange}
+                  data-qa-check-backups="This image is Cloud-init compatible"
+                />
+              }
+              label="This image is Clound-init compatible"
+            />
+            <HelpIcon text={cloudInitTooltipMessage} />
+          </div>
 
           <RegionSelect
             label="Region"


packages/manager/src/features/Images/ImageUpload.tsx
```diff
@@ -55,6 +57,9 @@ const useStyles = makeStyles((theme: Theme) => ({
     ...theme.applyLinkStyles,
     fontWeight: 700,
   },
+  cloudInitCheckboxWrapper: {
+    marginTop: theme.spacing(2),
```
mjac0bs, line 61: What do you think about adding a `marginLeft: 3` here? For some reason, it looks like our checkbox is juusssttt outside the aligned container on the left and I found that a tiny adjustment to the spacing seems to better align the fields visually. 

image

packages/manager/src/components/CheckBox/CheckBox.tsx
```diff
@@ -63,7 +70,9 @@ const LinodeCheckBox: React.FC<Props> = (props) => {
           }
           label={props.text}
         />
-        {toolTipText && <HelpIcon text={toolTipText} />}
+        {toolTipText && (
```
mjac0bs, line 73: While we're making improvements to this code, could we [implement this with a ternary rather than `&&`](https://github.com/linode/manager/blob/develop/docs/development-guide/13-coding-standards.md#react)?

packages/manager/src/features/Images/ImageUpload.tsx
```diff
@@ -213,6 +230,15 @@ export const ImageUpload: React.FC<Props> = (props) => {
             errorText={errorMap.description}
             disabled={!canCreateImage}
           />
+          <div className={classes.cloudInitCheckboxWrapper}>
+            <CheckBox
+              checked={isCloudInitChecked}
+              onChange={handleCloudInitChange}
+              data-qa-check-backups="This image is Cloud-init compatible"
```
mjac0bs, line 237: I don't think [L237](https://github.com/linode/manager/pull/8800/files#diff-031b9f02b49e3c8b0a9ec390547ca68b34a7c57ed730dd3e63e9803d3daf639fR237) should be here, looks like leftovers from copy & pasting.

packages/manager/src/features/Images/ImageUpload.tsx
```diff
@@ -151,6 +161,13 @@ export const ImageUpload: React.FC<Props> = (props) => {
   const cliRegion = formatForCLI(region, 'region');
   const linodeCLICommand = `linode-cli image-upload --label ${cliLabel} --description ${cliDescription} --region ${cliRegion} FILE`;
 
+  const cloudInitTooltipMessage = (
+    <Typography>
+      Copy TBD-Need to have installed cloud-init on the distro and change the
+      config to use our data service. <Link to="/">Link to doc</Link>
+    </Typography>
+  );
+
```
mjac0bs, line 170: Declare outside of component?

mjac0bs (commented):
I realize this PR is still in draft 😃 Just wanted to mention some initial things while working on M3-6149 since it's so similar.

mjac0bs (commented):
I'll circle back to this once we get the final copy approved and I left a couple nitpicky naming questions for us to decide on. 

Functionality looks good! Verified that images can be uploaded with cloud init capabilities. When the image does not have cloud-init enabled, it is not sent in the request. 👍

packages/manager/src/features/Images/ImagesCreate/ImageCreate.tsx
```diff
@@ -51,6 +57,8 @@ export const ImageCreate: React.FC<CombinedProps> = (props) => {
           description={description}
           changeLabel={handleSetLabel}
           changeDescription={handleSetDescription}
+          cloudInitChangeHandler={handleCloudInitChange}
```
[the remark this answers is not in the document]
jaalah-akamai: My opinion on this (fwiw) is that when exposing a components API (in this case coming from `ImageUpload`), it makes sense to expose a prop that is consistent with typical naming conventions. So since this handler is used for the cloud init checkbox, I personally would do something like `onCloudInitChange` and the handler stays as is `handleCloudInitChange`. That way the API is consistent not only with traditional conventions but the broader community who may look at this. 

With that said, let's continue this discussion elsewhere because it's a good point!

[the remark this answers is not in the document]
cpathipa: @mjac0bs  Yes, was trying to improve some readability in event handlers here. @jaalah-akamai  I like the naming convention `onCloudInitChange` But, for now to avoid confusion I'm going with existing pattern  `changeIsCloudInit`.
This could be a good topic to discuss in our frontend cafe.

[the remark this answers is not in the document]
mjac0bs: Agreed with you both. I like `onCloudInitChange` too. And I see it's already on the agenda for the next cafe; thanks @jaalah-akamai!

hana-linode (commented):
@cpathipa Mariah merged her PR into the metadata feature branch. Can you pull in those changes and address the merge conflicts?

jaalah-akamai (approved):

packages/manager/src/features/Images/ImagesCreate/ImageCreate.tsx
```diff
@@ -54,6 +54,8 @@ export const ImageCreate: React.FC<CombinedProps> = (props) => {
           description={description}
           changeLabel={handleSetLabel}
           changeDescription={handleSetDescription}
+          changeIsCloudInit={() => setIsCloudInit(!isCloudInit)}
+          isCloudInit={isCloudInit}
```
mjac0bs, line 58: These [changes](packages/manager/src/features/Images/ImagesCreate/ImageCreate.tsx) should already be merged into the feature branch so we wouldn't expect changes in this file. I think you just ordered the props differently here, but not a big deal.

packages/manager/src/features/Images/ImageUpload.tsx
```diff
@@ -64,8 +64,9 @@ const useStyles = makeStyles((theme: Theme) => ({
 
 const cloudInitTooltipMessage = (
   <Typography>
-    Copy TBD-Need to have installed cloud-init on the distro and change the
-    config to use our data service. <Link to="/">Link to doc</Link>
+    Only check this box if your Custom Image that is compatible with cloud-init,
```
jdamore-linode, line 67: Should this be `Only check this box if your Custom Image is compatible with cloud-init`?

mjac0bs: I think you're right; that sentence doesn't make sense with "that is compatible". 

Also: Noting that we are expecting a final copy update to finalize the docs links in this PR and others.

mjac0bs (approved):
Functionality looks good, so I'm approving with the minor comments below.

After looking at the tooltip spacing again, I do feel like it's the lesser of two evils to revert the change I made in [this commit](https://github.com/linode/manager/pull/8807/commits/f0f8c65e83e9e4507c21bd5b4f54136f7b162bad) on L36-38 and L72 that removed the `helpIcon` styles. The downfall is that then we continue to have a [stretched shadow](https://github.com/linode/manager/pull/8807#discussion_r1117258727) to the icon when the button is tabbed to, which is present throughout the app and should be fixed at some point. To me, that shadow issue is not as noticeable as the extra padding that is applied if we don't explicity set `paddingLeft` to `8px`.

jdamore-linode (approved):
Following @mjac0bs's lead -- totally fine fixing the copy later if we'd rather merge this sooner

cpathipa:
Thank you @mjac0bs for checking on copy updates.

PR #8801, opened 2023-02-16
M3-6174: Remove the `ip_whitelist_enabled` Profile Toggle

bnussman-akamai:
## Description 📝

- Removes the deprecated `ip_whitelist_enabled` feature 

## Preview 📷

The `IP Allowlisting (Legacy)` section, which nobody has probably ever seen, will be permanently gone
Screen Shot 2023-02-16 at 12 26 49 PM

## How to test 🧪

- Before these changes you could use the MSW by adding `ip_whitelist_enabled: true` to your profile. Doing this now will do nothing
- Verify that everything on the `/profile/auth` route works as you'd expect


dwiley-akamai (approved):

mjac0bs (approved):
> which nobody has probably ever seen

👋 , legacy feature.

Confirmed that the IP Allowlisting no longer shows up on `/profile/auth` even when enabled via mocks. The rest of the functionality on the page seems to be working as expected.

Non-blocking question about clean-up: since we're removing this feature entirely, did we consider removing the `ip_whitelist_enabled: false` line from `/src/factories/profile.ts` and the related test from `AuthenticationSettings.text.tsx`?

bnussman-akamai:
@mjac0bs We should keep `ip_whitelist_enabled: false` in the factories because the API still returns this data, but good call on the unit test.

PR #8803, opened 2023-02-16
M3-6175: Update Cloud Manager to Node.js 18.14.1 LTS

bnussman-akamai:
## Description 📝

- Updates Cloud Manager's Node.js version to the current latest LTS version
  - Currently we are on `14.17.4` and we will update to the latest LTS version which is `18.14.0`
- This update will unblock the ability to use Vite

## How to test 🧪

- Use nvm or volta to switch node version to 18.14.0
  - For me, I ran `volta install node` in the repo's root directory
- Verify Storybook still works
- Verify Github actions CI passes
- Verify internal build works


packages/api-v4/package.json
```diff
@@ -15,9 +15,6 @@
     "type": "git",
     "url": "https://github.com/linode/manager/tree/develop/packages/api-v4"
   },
-  "engines": {
-    "node": ">= 14.17.4"
-  },
```
bnussman-akamai, line 20: Removed this because IMO, it's not our business to tell people what version of node to use

packages/manager/src/features/Help/StatusBanners.test.tsx
```diff
@@ -31,7 +31,7 @@ describe('Status banners from statuspage.io', () => {
       expect(screen.getByRole('link')).toHaveAttribute('href', props.href);
     });
 
-    it("clicking a banner's close icon should remove it from view", () => {
+    it.skip("clicking a banner's close icon should remove it from view", () => {
```
bnussman-akamai, line 34: I'm pretty sure this is caused by preferences in redux being really weird. I'm confident that this will be fixed by https://github.com/linode/manager/pull/8795

jdamore-linode (approved):
Awesome work! Messed around with the app and everything looks good. Unit tests and e2e tests all pass, and Storybook looks good too!

jaalah (approved):
Glad to bump the version 🚀

jaalah-akamai (approved):
I don't know who the above guy is, but I agree! 😅

packages/api-v4/package.json
dwiley-akamai: I'm not wedded to keeping it, but I always thought of this more as guidance of what is confirmed to be compatible.

dwiley-akamai (approved):
🚢

packages/manager/src/features/Help/StatusBanners.test.tsx
hkhalil-akamai: Seems like this is the case.

PR #8805, opened 2023-02-17
M3-6173: Improve Cypress support for alternative environments

jdamore-linode:
## Description 📝

**What does this PR do?**
This improves support for running our Cypress tests against alternative environments. Because I had to touch nearly every test file, there's also a lot of clean-up and some other misc improvements.

**Changes to Support Alternative Environments:**
* Added an `apiMatcher()` utility function to generate Cypress intercept matchers against APIv4 URLs. This is more flexible than our previous pattern and supports API requests for environments other than our local Cloud instances.
* Modified every intercept to use the new matcher

**Clean-Up:**
* Replaced direct calls to `cy.intercept()` with intercept/mock utilities from `support/intercepts/` for all `account` tests. I realized pretty quickly that I bit off more than I could chew here, so I'll continue gradually replacing calls to `cy.interept` in more tests as part of subsequent clean-up PRs
* Cleaned up a bunch of imports
* Deleted some unused utility functions, as well as `support.spec.ts`, which didn't do anything useful

**Other Misc. Changes & Improvements:**
* API v4 now exports the types defined in `types.ts` from `index.ts`
* Fixed the crash caused by `creates first image w/ drawer, and fail because POST is stubbed` test, renamed test to reflect improvement
* Rename `object-storage-errors.ts` to `object-storage-errors.spec.ts`. Whoops!

## How to test 🧪

Confirm that Cypress tests continue passing locally and elsewhere. Refer to the ticket for some examples.

```bash
yarn cy:run
```

```bash
CYPRESS_BASE_URL=<environment url here> yarn cy:run
```


dwiley-akamai (approved):
LGTM! 🚢

bnussman-akamai (approved):

PR #8807, opened 2023-02-17
feat: [M3-6149] - Add Cloud Init checkbox to Image Capture/Create flow

mjac0bs:
## Description 📝

**What does this PR do?**
Adds a non-required checkbox field (and accompanying help tooltip) to the Image **Capture** form so that a user can indicate if an image supports cloud init. By default, a captured image does not support cloud init.

Note that #8800 is the related PR for the Image **Upload** form.

TODO:
- [x] Finalize UX copy. Any changes (e.g. final links) will be addressed in a final copy PR to come.

## Preview 📷

Screen Shot 2023-02-27 at 3 25 12 PM

## How to test 🧪

1. Check out this PR and switch your environment to Dev to test these changes. Ask if you need more details on testing these changes with a properly configured dev account.
2. Have at least one Linode with an attached Disk on your account.
3. Navigate to http://localhost:3000/images/create/disk to the Capture Image form.
4. Open the browser's dev tools, navigate to the network tab, and use the search bar to filter traffic by keyword "images" to make it easier to find the requests we want to verify.
5. Observe that the Capture Image form now includes a checkbox for cloud-init. Verify that the checkbox is unchecked by default and the UI (checkbox + tooltip) matches the mock in our internal issue. (Check back if you want to verify this; copy not yet finalized.)
7. Fill out the form with the required fields  **check** the cloud-init checkbox.
8. Examine the POST request in the dev tools Networks tab to verify that the request contains a value `cloud_init: true` and the image response contains `capabilities: ['cloud-init']`.
9. Fill out the form with the required fields and **do not check** the cloud-init checkbox.
10. Examine the POST request in the dev tools Networks tab to verify that the request contains **does not contain** a value `cloud_init` (since this is an optional field, we do not need to send `cloud_init: false`) and the image response **does not contain** `capabilities: ['cloud-init']`.
11. Change screen size and make sure UI looks consistent.

### DIFF
--- a/packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
+++ b/packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
@@ -23,6 +23,7 @@ import { useGrants, useProfile } from 'src/queries/profile';
 import { getAPIErrorOrDefault } from 'src/utilities/errorUtils';
 import getAPIErrorFor from 'src/utilities/getAPIErrorFor';
 import Link from 'src/components/Link';
+import CheckBox from 'src/components/CheckBox';
 
 const useStyles = makeStyles((theme: Theme) => ({
   container: {
@@ -57,7 +58,18 @@ const useStyles = makeStyles((theme: Theme) => ({
       width: '100%',
     },
   },
+  cloudInitCheckboxWrapper: {
+    marginTop: theme.spacing(2),
+    marginLeft: 3,
+  },
 }));
+
+const cloudInitTooltipMessage = (
+  <Typography>
+    Copy TBD-If you deployed this from one of our cloud-init compatible disks
+    then it probably is. <Link to="/">Link to doc</Link>
+  </Typography>
+);
 
 export interface Props {
   label?: string;
@@ -84,6 +96,9 @@ export const CreateImageTab: React.FC<Props & ImagesDispatch> = (props) => {
 
   const [selectedLinode, setSelectedLinode] = React.useState<Linode>();
   const [selectedDisk, setSelectedDisk] = React.useState<string | null>('');
+  const [isCloudInitChecked, setIsCloudInitChecked] = React.useState<boolean>(
+    false
+  );
   const [disks, setDisks] = React.useState<Disk[]>([]);
   const [notice, setNotice] = React.useState<string | undefined>();
   const [errors, setErrors] = React.useState<APIError[] | undefined>();
@@ -138,6 +153,10 @@ export const CreateImageTab: React.FC<Props & ImagesDispatch> = (props) => {
     // Clear any errors
     setErrors(undefined);
     setSelectedDisk(diskID);
+  };
+
+  const handleCloudInitChange = () => {
+    setIsCloudInitChecked(!isCloudInitChecked);
   };
 
   const onSubmit = () => {
@@ -265,6 +284,14 @@ export const CreateImageTab: React.FC<Props & ImagesDispatch> = (props) => {
         />
       </Box>
       {isRawDisk ? rawDiskWarning : null}
+      <Box className={classes.cloudInitCheckboxWrapper}>
+        <CheckBox
+          checked={isCloudInitChecked}
+          onChange={handleCloudInitChange}
+          text="This image is Cloud-init compatible"
+          toolTipText={cloudInitTooltipMessage}
+        />
+      </Box>
       <>
         <TextField
           label="Label"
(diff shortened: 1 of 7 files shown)


packages/manager/src/features/Images/ImagesCreate/ImageCreate.tsx
```diff
@@ -37,8 +42,10 @@ export const ImageCreate: React.FC<CombinedProps> = (props) => {
         <CreateImageTab
           label={label}
           description={description}
+          isCloudInit={isCloudInit}
           changeLabel={handleSetLabel}
           changeDescription={handleSetDescription}
+          changeCloudInit={handleSetCloudInit}
```
mjac0bs, line 48: Ultimately, [#8800 ](https://github.com/linode/manager/pull/8800/files#diff-f608b295d346993372302a211f8bf899244e4d5dd0942a7263e9aed7c7637efd) and this PR will use the same parameter names. I preferred these to follow existing naming conventions, but we can come to an agreement.

bnussman-akamai (approved):
Followed the testing instructions and everything worked as expected

hana-linode (approved):
Functionality LGTM! Approving pending final UX copy

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
```diff
@@ -265,6 +282,14 @@ export const CreateImageTab: React.FC<Props & ImagesDispatch> = (props) => {
         />
       </Box>
       {isRawDisk ? rawDiskWarning : null}
+      <Box className={classes.cloudInitCheckboxWrapper}>
+        <CheckBox
+          checked={isCloudInit}
+          onChange={changeIsCloudInit}
+          text="This image is Cloud-init compatible"
+          toolTipText={cloudInitTooltipMessage}
```
dwiley-akamai, line 290: Since `<HelpIcon />` is included in the `<CheckBox />` component if there is `toolTipText`, we'll need to add another optional prop for `CheckBox` to indicate if we want the tooltip to be interactive (so that it can be passed down as a prop to the `HelpIcon` instance in `<CheckBox />`).

Another option would be to check if the `toolTipText` contains a `<Link [...]` substring (maybe an `<a [...]` as well) somewhere in it, but I'd probably go with the option above.

It looks like there will be a link in this tooltip based on the placeholder, but without the `interactive` prop on `HelpIcon`, a user would never be able to hover over the tooltip contents to be able to click it.

dwiley-akamai (commented):
The `cloud-init` functionality looks good 👍 , left a comment re: the tooltip functionality

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
mjac0bs: Good catch here, thank you for this! I hadn't paid much attention to that tooltip since the copy wasn't finalized, but you're right. Will make that update.

hkhalil-akamai:
Just curious, what's the reason for only having this option available for captured images and not uploaded images? Is it a technical limitation?

mjac0bs:
> Just curious, what's the reason for only having this option available for captured images and not uploaded images? Is it a technical limitation?

This ticket was created to add the cloud-init checkbox for captured images, specifically, though there is quite a bit of overlap with the ticket to add the cloud-init checkbox for uploaded images. This is also being done. (See #8800.)

dwiley-akamai (approved):
🚀

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
jaalah-akamai: Agree with @dwiley-akamai on this. I would go a step further and say that by default `disabledInteractive` [here](https://github.com/linode/manager/blob/develop/packages/manager/src/components/HelpIcon/HelpIcon.tsx#L68) should always false by default, so removing the `!` should fix this. If we disable interactivity we're going to be failing accessibility guidelines as pointed out in the [docs](https://mui.com/material-ui/react-tooltip/#interactive).

jaalah-akamai: Nice feedback - I agree with everything you said, let's move this work into a separate ticket to address this 👍

jaalah-akamai (approved):

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
mjac0bs: @jaalah-akamai Sounds good, I created M3-6200 for this.

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
```diff
@@ -57,21 +58,37 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+    marginTop: theme.spacing(2),
+    marginLeft: 3,
+  },
 }));
 
+const cloudInitTooltipMessage = (
+  <Typography>
+    Many Linode supported distributions are compatible with cloud-init by
+    default, or you may have installed cloud-init.{' '}
+    <Link to="/">Learn more.</Link>
```
mjac0bs, line 71: This link will be finalized in one final copy follow up PR.

PR #8810, opened 2023-02-21
refactor: M3-6185: Standardize Query String Parsing and Stringifying

bnussman-akamai:
## Description 📝

- Currently we use `querystring` (from Node.js) and package called `qs` to handle query string manipulation and parsing. This is bad because `querystring` is from Node.js and we don't want to have to deal with polyfills and we don't want to use `qs` because we don't want extra dependences. 
- We really need to be using `URLSearchParams` because it is built into the browser. 
- We need to make this change because Vite will complain about us using a Node.js package in code intended for the browser
- Switch instances of `querystring` and `qs` to use `URLSearchParams`
- Remove any `querystring` or `qs` related packages
 
Screenshot 2023-02-21 at 2 54 53 PM

## How to test 🧪

```sh
yarn test queryParams
```

- When you scroll through the diff of this PR, check the corresponding functionality in Cloud Manager


packages/manager/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     /* Language and Environment */
     "jsx": "react",
-    "lib": ["dom", "es2017"],
+    "lib": ["dom", "es2019"],
```
bnussman-akamai, line 5: This gives us the ability to call `Object.fromEntries()`.

packages/manager/src/store/linodeCreate/linodeCreate.reducer.ts
```diff
@@ -8,55 +7,44 @@ export interface State {
 }
 
 export const getInitialType = (): CreateTypes => {
```
bnussman-akamai, line 9: This function could really use some simplification but I'm not going to rewrite anything in hopes to not break anything.

yarn.lock
```diff
@@ -14937,7 +14937,7 @@ qrcode.react@^0.8.0:
     prop-types "^15.6.0"
     qr.js "0.0.0"
 
-qs@6.11.0, qs@^6.10.0, qs@^6.6.0:
+qs@6.11.0, qs@^6.10.0:
```
dwiley-akamai, line 14940: Are these (and `"@types/qs@*", "@types/qs@^6.9.5"`) still needed?

packages/manager/src/features/StackScripts/StackScripts.tsx
```diff
@@ -22,11 +21,12 @@ export const StackScripts: React.FC<Props> = (props) => {
   } = props;
 
   // Redirects to prevent breaking old stackscripts?type=whatever bookmarks
-  const searchParams = parseQueryParams(search);
-  if (searchParams.type === 'community') {
+  const searchParams = new URLSearchParams(search);
```
dwiley-akamai, line 24: Since `parseQueryParams` was refactored to make use of `URLSearchParams`, is there any particular benefit to not using it in this file aside from cutting an import?

dwiley-akamai (approved):

packages/manager/src/features/StackScripts/StackScripts.tsx
bnussman-akamai: Yeah, my original intent was to go as far as removing `parseQueryParams` and the entire `packages/manager/src/utilities/queryParams.ts` file, but I realized that would just require a lot of extra changes. I probably made this change before I realized that I was gonna keep `packages/manager/src/utilities/queryParams.ts` around.

yarn.lock
bnussman-akamai: Sadly, these will still show up in our `yarn.lock` because things like Storybook and Cypress depend on them

jaalah-akamai (approved):

PR #8811, opened 2023-02-22
feat: [M3 6143] - Add User Data Accordion to Linode Create page

hana-linode:
## Description 📝
Add User Data accordion to the Linode Create flow so that a user can configure their Linode using cloud-init.

Todo:
- [x] Finalize UX copy

## Preview 📷
image

## How to test 🧪
Testing Create via Distribution
1. Point to the dev environment (you might also need additional account setup, reach out for more info)
2. Navigate to the Linode create page http://localhost:3000/linodes/create
3. Select a distribution that supports cloud-init
    - Arch Linux, Cent OS 7, and Ubuntu 16.04 LTS have been marked as cloud-init compatible for testing purposes
4. If the distribution supports cloud-init, you should see an `Add User Data` accordion
    - If the distribution doesn't support cloud-init, you should _not_ see the accordion
5. Enter some text (can be anything) into the User Data input field and fill out the rest of the required Linode Create fields
6. Click on Create Linode and observe the POST payload to `/linode/instances`. You should see a metadata field with the user_data object
    - If you did not enter anything into the User Data input, you should _not_ see any metadata field in the payload.
    - You can just block the network request to `/linode/instances` and check the payload w/o actually creating another Linode
7. There are no changes to the API response so you will not see a metadata field returned

image

Testing Create via Image
- Navigate to the Linode Create Image tab http://localhost:3000/linodes/create?type=Images
- Select an image that supports cloud-init
  - You can create and mark an image as cloud-init compatible in https://github.com/linode/manager/pull/8807
- Follow steps 4-7 above

### DIFF
--- a/packages/manager/src/features/linodes/UserDataAccordion.tsx
+++ b/packages/manager/src/features/linodes/UserDataAccordion.tsx
@@ -28,7 +28,7 @@ const useStyles = makeStyles(() => ({
   },
 }));
 
-const UserDataAccordion: React.FC<Props> = (props) => {
+const UserDataAccordion = (props: Props) => {
   const { disabled, userData, onChange } = props;
   const classes = useStyles();
 
--- a/packages/manager/src/features/linodes/UserDataAccordion.tsx
+++ b/packages/manager/src/features/linodes/UserDataAccordion.tsx
@@ -1,0 +1,59 @@
+import * as React from 'react';
+import Accordion from 'src/components/Accordion';
+import { makeStyles } from 'src/components/core/styles';
+import Typography from 'src/components/core/Typography';
+import HelpIcon from 'src/components/HelpIcon';
+import TextField from 'src/components/TextField';
+
+interface Props {
+  userData: string | undefined;
+  onChange: (userData: string) => void;
+  disabled?: boolean;
+}
+
+const useStyles = makeStyles(() => ({
+  helpIcon: {
+    padding: '0px 0px 4px 8px',
+    '& svg': {
+      fill: 'currentColor',
+      stroke: 'none',
+    },
+  },
+}));
+
+const UserDataAccordion: React.FC<Props> = (props) => {
+  const { disabled, userData, onChange } = props;
+  const classes = useStyles();
+
+  const accordionHeading = (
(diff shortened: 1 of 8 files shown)


packages/manager/src/features/linodes/UserDataAccordion.tsx
```diff
@@ -0,0 +1,59 @@
...
+  },
+}));
+
+const UserDataAccordion: React.FC<Props> = (props) => {
+  const { disabled, userData, onChange } = props;
+  const classes = useStyles();
+
+  const accordionHeading = (
+    <>
+      Add User Data{' '}
```
mjac0bs, line 30: Fairly minor, but stood out to me: we have a difference in section heading size and alignment for the User Data accordion versus existing sections of the Linode Create form. I'd guess the accordion uses Typography with `h3` and the other headers are `h2`. What do you think about adding some styling to make spacing and font-size more consistent here?

Screen Shot 2023-02-24 at 8 57 05 AM

mjac0bs (approved):
Functionality LGTM on both the Linode create from distro and with #8807 to create from image. I did observe the network request as described, though did not go so far as to check that the API received the user data. 

Also noticed one small UI inconsistency. Going to approve pending that and the final UX copy are addressed.

dwiley-akamai (approved):
LGTM 🚢

packages/manager/src/features/linodes/UserDataAccordion.tsx
```diff
@@ -0,0 +1,59 @@
...
+  helpIcon: {
+    padding: '0px 0px 4px 8px',
+    '& svg': {
+      fill: 'currentColor',
+      stroke: 'none',
+    },
+  },
+}));
+
+const UserDataAccordion: React.FC<Props> = (props) => {
```
jaalah-akamai, line 24: Nice job! Very small change request... for new features let's stop using `React.FC` since it's [unnecessary](https://github.com/facebook/create-react-app/pull/8177) and even removed out of Facebook's suggested use of it.

jaalah-akamai (approved):

PR #8814, opened 2023-02-22
Remove `wait-on` package for CVE-2023-25166

bnussman-akamai:
## Description 📝

- Removes `wait-on` package to solve CVE-2023-25166 dependabot
- `wait-on` depended on `@sideway/formula` and `@sideway/formula` has a ReDoS Vulnerability
- Thankfully, we no longer use `wait-on` so we can just remove it, resulting in us no longer depending on `@sideway/formula` 
  - For context, we previously used `wait-on` for e2e pipelines but we now use docker livelesness checks

## How to test 🧪

- Test general features of Cloud Manager 
- Verify e2es pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,8 @@
     "minimatch": "^3.0.5",
     "nth-check": "^2.0.1",
     "@types/react": "^17",
-    "loader-utils": "^2.0.4"
+    "loader-utils": "^2.0.4",
+    "@sideway/formula": "^3.0.1"
   },
   "workspaces": {
     "packages": [
--- a/package.json
+++ b/package.json
@@ -61,8 +61,7 @@
     "minimatch": "^3.0.5",
     "nth-check": "^2.0.1",
     "@types/react": "^17",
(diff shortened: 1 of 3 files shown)


jaalah-akamai:
Do we use/need wait-on as a dependency, can we remove it?

bnussman-akamai:
@jaalah-akamai I was wondering the same thing, and I verfied with @jdamore-linode we should be able to remove it!

jaalah-akamai:
> @jaalah-akamai I was wondering the same thing, and I verfied with @jdamore-linode we should be able to remove it!

I didn't know if we needed it for some CLI work, but if we do and we're having trouble bumping the external library - `concurrently` might be an alternative.

bnussman-akamai:
@jaalah-akamai Pushed 09e5b8c to remove the dependency and updated the PR description with some context

jaalah-akamai (approved):

jdamore-linode (approved):

PR #8819, opened 2023-02-23
feat: [M3-6147] - Add warning for invalid user data

hkhalil-akamai:
## Description 📝

The User Data of a linode will generally be used by a something like cloud-init to initialize a VM. There is an extremely broad set of valid configuration formats ([described here](https://cloudinit.readthedocs.io/en/latest/explanation/format.html)). Custom formats are also possible if the user image uses something besides cloud-init.

To helpfully guide Metadata users, this feature uses basic string matching to determine if User Data does not match one of the most common formats and displays a warning to the user. The user is able to ignore the warning and proceed anyways if they wish.

~*Note: Since I built this branch off [M3-6143](https://github.com/Hana-Linode/manager/tree/M3-6143-add-user-data-linode-create), this PR currently includes extra commits until that branch is reviewed and merged.*~

## Preview 📷

https://user-images.githubusercontent.com/122488130/222278964-c0c05f1b-63f7-446f-a539-26c29f04c9b9.mov

## How to test 🧪

[For now, you must be using the dev API.]

1. In the Create Linode flow, select an image that is cloud-init compatible. This will enable the Cloud Data dropdown.
2. In the Cloud Data, verify entering a User Data script that doesn't match a common cloud-config format causes the warning to appear.
3. Verify correcting the format causes the warning to disappear.


packages/manager/src/features/linodes/UserDataAccordion.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  const classes = useStyles();
+
+  const checkFormat = (userData: string, isOnChange: boolean) => {
+    const userDataLower = userData.toLowerCase();
+    if (
+      userData.length > 0 &&
+      !userDataLower.startsWith('#cloud-config') &&
+      !userDataLower.startsWith('content-type: text/') &&
+      !userDataLower.startsWith('#!')
```
hkhalil-akamai, line 38: Matches YAML and bash user data formats

packages/manager/src/features/linodes/UserDataAccordion.tsx
```diff
@@ -0,0 +1,94 @@
...
+  },
+}));
+
+const UserDataAccordion: React.FC<Props> = (props) => {
+  const { disabled, userData, onChange } = props;
+  const [formatWarning, setFormatWarning] = React.useState(false);
+
+  const classes = useStyles();
+
+  const checkFormat = (userData: string, isOnChange: boolean) => {
```
jaalah-akamai, line 32: Two nits for readability: 
- `isOnChange` suggests that the function is being called when the value of an input field changes. Since that's not the case, a more meaningful name could be `isInputValueChanged` or `hasInputValueChanged`.
- This may not be a big deal since there's not a lot of parameters. But it's hard to know what `true` and `false` do at a glance if let's say `checkFormat` was in some util or something. One option would be to deconstruct this so the function call would be `checkFormat({ userData: e.target.value, hasInputValueChanged: true });`

jaalah-akamai (changes requested):

packages/manager/src/features/linodes/UserDataAccordion.tsx
jaalah-akamai: We could remove some of the redundant checks by moving the strings into an array called `validPrefixes` and then using the `some` method to check if any of the prefixes match the start of the string.
```
const isUserDataValid = validPrefixes.some((prefix) =>
  userDataLower.startsWith(prefix)
);
```
Then we can just use the `!isUserDataValid && !hasInputValueChanged` variables in the if/else statement to toggle only the state.

jaalah-akamai (changes requested):

packages/manager/src/features/linodes/UserDataAccordion.tsx
jaalah-akamai: Guess Github wants to duplicate things... 😮‍💨

packages/manager/src/features/linodes/UserDataAccordion.tsx
```diff
@@ -0,0 +1,94 @@
...
+          checkFormat(e.target.value, true);
+          onChange(e.target.value);
+        }}
+        value={userData}
+        disabled={Boolean(disabled)}
+        onBlur={(e) => checkFormat(e.target.value, false)}
+        data-qa-user-data-input
+      />
+      {formatWarning ? (
+        <Notice warning important>
```
mjac0bs, line 80: What do you think about adding a little spacing between the bottom of the `TextField` and top of the warning `Notice`, or getting some feedback from UX on their thoughts? 

**Now**:
Screen Shot 2023-02-27 at 7 49 35 AM


**With some margin** (this was a top margin of 24 to match default bottom margin styling for `Notice`, but may be too much space):

```suggestion
        <Notice warning important spacingTop={24}>
```
Screen Shot 2023-02-27 at 7 59 51 AM

packages/manager/src/features/linodes/UserDataAccordion.tsx
```diff
@@ -47,11 +67,26 @@ const UserDataAccordion: React.FC<Props> = (props) => {
...
+          checkFormat(e.target.value, true);
+          onChange(e.target.value);
+        }}
         value={userData}
         disabled={Boolean(disabled)}
+        onBlur={(e) => checkFormat(e.target.value, false)}
         data-qa-user-data-input
       />
+      {formatWarning ? (
+        <Notice warning important>
```
hana-linode, line 80: I don't think we need to make this warning important, a regular warning notice should suffice

hana-linode (changes requested):

hkhalil-akamai:
Blocked pending UX consideration.

jaalah-akamai (approved):

jaalah-akamai:
Nice work @hkhalil-akamai 🚀

mjac0bs (approved):
Functionality looks good. Warning spacing and placement on top of the text field definitely feel more standard; thanks for getting the feedback and making the change.  🚀

packages/manager/src/features/linodes/UserDataAccordion.tsx
```diff
@@ -76,14 +100,25 @@ const UserDataAccordion = (props: Props) => {
         user group(s). <br /> Accepted formats are YAML and bash.{' '}
         <Link to="https://www.linode.com/docs">Learn more.</Link>
       </Typography>
+      {formatWarning ? (
+        <Notice warning spacingTop={16} spacingBottom={16}>
+          This user data may not be in a format accepted by cloud-init.
```
bnussman-akamai, line 105: Is this warning message final?  I think it is too vague and could be frustrating to users. The warning says that something is wrong but It doesn't say what or instructions on how to correct it.

hkhalil-akamai: There is a link to the cloud-init format docs that was originally in this message but was moved to the helper text above it.

PR #8820, opened 2023-02-23
fix: [M3-5840] - Fix validateDOMNesting warning in console

dwiley-akamai:
## Description 📝
When opening the Notifications dropdown currently, a `validateDOMNesting` warning gets logged to the console because there's a case of `<p>` being a descendant of `<p>`.

It seems this was being caused in `RenderEvent.tsx` by `<Typography>` wrapping `<HighlightedMarkdown>`, which returns `<Typography>`. 

## How to test 🧪
Confirm the warning displayed below is no longer logged to the console when opening the Notifications dropdown on this branch.

Screenshot 2023-02-23 at 6 14 45 PM

### DIFF
--- a/packages/manager/src/features/NotificationCenter/NotificationData/RenderEvent.tsx
+++ b/packages/manager/src/features/NotificationCenter/NotificationData/RenderEvent.tsx
@@ -63,13 +63,13 @@ export const RenderEvent: React.FC<Props> = (props) => {
   }
 
   const eventMessage = (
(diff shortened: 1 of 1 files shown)


bnussman-akamai (approved):
Thanks for fixing, randomly just noticed this today too

jaalah-akamai (approved):

cpathipa (approved):

PR #8821, opened 2023-02-24
refactor: M3-6195 - Add tss-react and refactor Button to styled API

jaalah-akamai:
## Description 📝
This refactors the `Button` component to the new [sx](https://mui.com/material-ui/react-box/#the-sx-prop) and [styled](https://mui.com/system/styled/#difference-with-the-sx-prop) API. I also added [tss-react](https://github.com/garronej/tss-react) so that we can begin migrating the older makeStyles API which will give us better type safety and autocomplete than MUI v4, while providing an almost seamless transition in moving away from the deprecated JSS API.

## Major Changes 🔄 
- Added `tss-react`
- Added `Button.test.tsx` for more coverage
- Refactored `Button` to use `styled` API
- Added `sx` prop to `HelpIcon` for one-off styling
- Added util `isPropValid` to filter out props that are not valid DOM attributes
- Created `packages/manager/src/styles` folder for all future shared `keyframes`

## Preview 📷
| Before & After |
| ----------- |
| Screen Shot 2023-02-23 at 11 00 12 PM | 

## How to test 🧪
1. Start local environments and storybook to ensure `Button` stayed the same.
2. Run unit test in root `yarn test packages/manager/src/components/Button` and observe unit tests pass.
3. Run E2E tests in root `yarn cy:debug`

### DIFF
--- a/packages/manager/src/components/Button/Button.tsx
+++ b/packages/manager/src/components/Button/Button.tsx
@@ -1,121 +1,101 @@
-import classNames from 'classnames';
-import { always, cond, propEq } from 'ramda';
 import * as React from 'react';
 import Reload from 'src/assets/icons/reload.svg';
-import _Button, { ButtonProps } from 'src/components/core/Button';
-import { makeStyles, Theme } from 'src/components/core/styles';
+import _Button, { ButtonProps } from '@mui/material/Button';
 import HelpIcon from 'src/components/HelpIcon';
+import { useTheme, styled } from '@mui/material/styles';
+import { SxProps } from '@mui/system';
+import { isPropValid } from '../../utilities/isPropValid';
+import { rotate360 } from '../../styles/keyframes';
 
 export interface Props extends ButtonProps {
   buttonType?: 'primary' | 'secondary' | 'outlined';
   className?: string;
+  sx?: SxProps;
   compactX?: boolean;
   compactY?: boolean;
   loading?: boolean;
   tooltipText?: string;
 }
 
-const useStyles = makeStyles((theme: Theme) => ({
-  '@keyframes rotate': {
-    from: {
-      transform: 'rotate(0deg)',
-    },
-    to: {
-      transform: 'rotate(360deg)',
-    },
-  },
-  loading: {
+const StyledButton = styled(_Button, {
+  shouldForwardProp: (prop) =>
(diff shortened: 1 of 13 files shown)


packages/manager/src/components/Button/Button.tsx
```diff
@@ -1,121 +1,101 @@
-import classNames from 'classnames';
-import { always, cond, propEq } from 'ramda';
 import * as React from 'react';
 import Reload from 'src/assets/icons/reload.svg';
-import _Button, { ButtonProps } from 'src/components/core/Button';
```
hkhalil-akamai, line 5: It looks like `src/components/core/Button` isn't used anywhere anymore. Should it be deleted?

packages/manager/src/features/Profile/APITokens/RevokeTokenDialog.tsx
```diff
@@ -3,7 +3,7 @@ import ConfirmationDialog from 'src/components/ConfirmationDialog';
 import ActionsPanel from 'src/components/ActionsPanel';
 import Typography from 'src/components/core/Typography';
 import { Token } from '@linode/api-v4/lib/profile/types';
-import { Button } from 'src/components/Button/Button';
+import Button from 'src/components/Button/Button';
```
hkhalil-akamai, line 6: Just to make this import consistent with others

```suggestion
import Button from 'src/components/Button';
```

hkhalil-akamai (approved):
LGTM and I agree with this direction for styles moving forward!

packages/manager/src/utilities/isPropValid/isPropValid.ts
```diff
@@ -0,0 +1,18 @@
+/**
```
carrillo-erik, line 1: I like this, in the back of my mind I was worrying about the props drilling, but this should address those concerns.

carrillo-erik (approved):

packages/manager/src/components/Button/Button.tsx
jaalah-akamai: I think it's still used in `src/components/PromotionalOfferCard/PromotionalOfferCard.tsx` and `/src/components/Tile/Tile.tsx`, but I agree! It will 100% get removed in the larger migration when we get to those files specifically. Good call out 👍

dwiley-akamai (approved):
LGTM 🚢

PR #8823, opened 2023-02-24
M3-6081 Dynamic Lish URLs

jcallahan-akamai:
## Description 📝

**What does this PR do?**

Lish URLs have been updated to accept the scheme: `<region.id>.webconsole.linode.com`. Previously, it used a custom, hardcoded prefix (that isn't returned from the /regions endpoint). This way, we don't have to update this hardcoded map every time we open a new datacenter.

This is Part 1 of the changes described in the ticket. Part 2 will be to update region display names to make use of the new "label" field on the /regions endpoint.

## Preview 📷

Screen Shot 2023-02-24 at 1 54 59 PM

## How to test 🧪

Create new Linodes in different DCs, click "Launch LISH console" and make sure it works. Use the "LISH Console via SSH" command in your own terminal and verify it works.

### DIFF
--- a/packages/api-v4/src/networking/types.ts
+++ b/packages/api-v4/src/networking/types.ts
@@ -1,18 +1,3 @@
-export type ZoneName =
-  | 'newark'
-  | 'dallas'
-  | 'fremont'
-  | 'atlanta'
-  | 'london'
-  | 'tokyo'
-  | 'singapore'
-  | 'frankfurt'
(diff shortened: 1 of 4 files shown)


bnussman-akamai (approved):
Confirmed I was able to use Cloud Manager's LISH and SSH LISH

jaalah-akamai (approved):
This lish good to me!

PR #8825, opened 2023-02-24
refactor: M3-6199 - Update from React `17` to `18`

bnussman-akamai:
## Description 📝

- Updates Cloud  Manager to use React 18
- Uses codemod to fix our types `https://github.com/eps1lon/types-react-codemod`
  - In React 18, `children` no longer are a type on React.FC
 
### React 18 Major Features and Changes
- Automatic state update batching
- Concurrent rendering
- `React.FC` no longer has `children` by default
  - In all components from now on, we should be explicit with our `children` type 

```typescript
interface Props = {
  ...
  children?: React.ReactNode
}

const MyComponent = (props: Props) => {}
```

### Motivation for Updating
- Potential performance improvements
- Compatibility with modern packages
- To remain on latest React version

## How to test 🧪

- Test all features of Cloud Manager
- Verify CI passes
- Consider running some e2e tests


hkhalil-akamai:
Locally merged #8795 and confirmed that PR will solve the preferences loop.

bnussman-akamai:
@jdamore-linode  The e2e tests came in clutch and saved us from some bugs 🐛🎉 

- [ ] security-questions.spec.ts
  - @jdamore-linode I'm gonna need your help on this one 😣
- [ ] create-database.spec.ts
  - Should be fixed after https://github.com/linode/manager/pull/8829 is merged in
- [ ] service-transfer.spec.ts
  - This was just a flake, no worries on this one
- [ ] machine-image-upload.spec.ts
  - Fixed in 535f116
- [ ] migrate-linode-with-firewall.spec.ts
  - Fixed in ef7548f
- [ ] rescue-linode.spec.ts
  - Fixed in ef7548f

jdamore-linode:
@bnussman-akamai I'll take a look later today! That security questions failure is especially interesting to me because we had been having some flakiness with that test which looked identical to these failures, except it happened so rarely that it was really difficult to reproduce -- I'm kind of relieved that it's happening consistently now.

jdamore-linode:
Hey @bnussman-akamai, still trying to figure out the security questions thing but in the course of that identified a really minor but random issue with the user menu: opening the user menu from the billing page the first time causes it to initially open on the left side of the screen, and then shift to the right side. I'm not sure if it only happens on the billing page, but it isn't happening on other pages I've tried.

I uploaded a quick video to the ticket!

bnussman-akamai:
@jdamore-linode I think this is happening because Reach UI isn't necessarily React 18 compatible and is no longer maintained. I updated to the latest and that did not fix any problems. I am pushing up a fix, but we really need to switch the Cloud Manager off of Reach UI to just MUI imo

jdamore-linode:
@bnussman-akamai Still no explanation for the security questions test, but this is what I'm seeing so far:

- When Cypress clicks the "Edit" button next to a security question, Cloud apparently sends a `POST` request to `/profile/security-questions` to update the security questions
- The `POST` request succeeds after a moment, which causes the toast notification to appear and the security questions form to no longer be in its edit state -- I'm pretty sure this is what prompts the "cy.type() failed because the page updated as a result of this command, but you tried to continue the command chain" error message, and it's a red herring
- This unexpected `POST` request occurs even if I comment out `lns 258-276` and `lns 302-322` -- initially I thought Cypress may be jumping the gun and somehow executing commands too soon, causing the "Update Security Questions" button to somehow get clicked prematurely, but this point seems to discredit that idea
- I have not been able to reproduce this outside of Cypress; when I click any of the "Edit" buttons, I'm not seeing any `POST` request like I'm seeing from within Cypress
- This happens even if I wait 30 seconds before interacting with the page after it loads
- I'm no longer convinced that this is the same issue that I'd seen before and mentioned in my comment this morning, although both issues seem to involve these "Edit" buttons.

I was hoping that whatever the issue was, I'd be able to narrow it down from Cypress. Tomorrow I'll probably mess with the security questions component(s) and see if I can learn more from there.

bnussman-akamai:
e2es are back in a good state 🎉 cc @jdamore-linode 

Screenshot 2023-03-01 at 6 26 42 PM

dwiley-akamai (commented):
Didn't finish combing through everything yet but I haven't observed any breaks so far

packages/manager/src/components/AbuseTicketBanner/AbuseTicketBanner.tsx
```diff
@@ -10,7 +10,7 @@ import getAbuseTicket from 'src/store/selectors/getAbuseTicket';
 
 const preferenceKey = 'abuse-tickets';
 
-export const AbuseTicketBanner: React.FC<{}> = (_) => {
+export const AbuseTicketBanner: React.FC<React.PropsWithChildren<{}>> = (_) => {
```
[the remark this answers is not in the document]
jaalah-akamai: Yes! However, it will be something done over time and outside the scope of this PR.

We could do either options to avoid the verbosity of the above:
```
interface Props { ... }
type CombinedProps = React.PropsWithChildren<Props>;

const MyComponent = (props: CombinedProps) => { ... }
```
or even
```
interface Props = {
  ...
  children?: React.ReactNode
}

const MyComponent = (props: Props) => {}
```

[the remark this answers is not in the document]
bnussman-akamai: Agree 👆🏼 The whole situation is a bit annoying but it is giving us a good excuse to progressively transition to a more explicit and readable way to type props. 

Following this PR I wouldn't have a problem with a clean up PR that goes throughout the app and cleans up some things like `React.FC<React.PropsWithChildren<{}>>` or unnecessary `CombinedProps ` usage.

packages/manager/src/components/FileUploader/FileUploader.tsx
```diff
@@ -271,7 +274,11 @@ const FileUploader: React.FC<CombinedProps> = (props) => {
           autoHideDuration: 6000,
         });
 
-        dispatchAction(setPendingUpload(false));
+        // React force a render so that `pendingUpload` is false when navigating away
+        // from the upload page.
+        flushSync(() => {
+          dispatchAction(setPendingUpload(false));
+        });
```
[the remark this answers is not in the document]
bnussman-akamai: This solution made the most sense to me. I'm sure we could re-work this component where we wouldn't have to do this but I did this because it seems like the path of least resistance

packages/manager/src/components/CheckoutBar/DisplaySectionList.tsx
```diff
@@ -6,13 +6,13 @@ interface Props {
...
 
-export const DisplaySectionList: React.FC<Props> = ({ displaySections }) => {
+export const DisplaySectionList: React.FC<React.PropsWithChildren<Props>> = ({ displaySections }) => {
   if (!displaySections) {
     return null;
   }
   return (
     // eslint-disable-next-line react/jsx-no-useless-fragment
-    <>
+    (<>
```
jaalah-akamai, line 15: Nit: Remove these extra parens

packages/manager/src/components/LineGraph/LineGraph.tsx
```diff
@@ -287,7 +287,7 @@ const LineGraph: React.FC<CombinedProps> = (props: CombinedProps) => {
   return (
     // Allow `tabIndex` on `<div>` because it represents an interactive element.
     // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
-    <div
+    (<div
```
jaalah-akamai, line 290: Same with this parens

packages/manager/src/components/SupportError.tsx
```diff
@@ -7,30 +7,30 @@ interface Props {
...
-export const SupportError: React.FC<Props> = (props) => {
+export const SupportError: React.FC<React.PropsWithChildren<Props>> = (props) => {
   const { errors } = props;
   const errorMsg = errors[0].reason.split(/(open a support ticket)/i);
 
   const classes = useStyles();
 
   return (
-    <Typography className={classes.noticeText}>
+    (<Typography className={classes.noticeText}>
```
jaalah-akamai, line 17: Same here - the code-mod must've added these in several places

packages/manager/src/components/LandingLoading/LandingLoading.tsx
```diff
@@ -50,7 +50,7 @@ export const LandingLoading: React.FC<Props> = ({
   return showLoading ? (
     !!children ? (
       /** allows us to pass a custom Loader if we please */
-      <React.Fragment>{children}</React.Fragment>
+      (<React.Fragment>{children}</React.Fragment>)
```
jaalah-akamai, line 53: Same

packages/manager/src/components/TagCell/TagCell.tsx
```diff
@@ -133,7 +133,7 @@ export const TagCell: React.FC<Props> = (props) => {
   const [addingTag, setAddingTag] = React.useState<boolean>(false);
   const [loading, setLoading] = React.useState<boolean>(false);
   const overflowRef = React.useCallback(
-    (node) => {
+    (node: any) => {
```
jaalah-akamai, line 136:
```
const overflowRef: OverflowRefCallback = useCallback((node: HTMLElement | null) => {
 ...
}
```

packages/manager/src/components/TransferDisplay/TransferDisplay.tsx
```diff
@@ -147,14 +147,14 @@ interface DialogProps {
...
-export const TransferDialog: React.FC<DialogProps> = React.memo((props) => {
+export const TransferDialog: React.FC<React.PropsWithChildren<DialogProps>> = React.memo((props) => {
   const classes = useStyles();
   const { isOpen, onClose, poolUsagePct, quota, used } = props;
 
   const daysRemainingInMonth = getDaysRemaining();
 
   return (
-    <Dialog
+    (<Dialog
```
jaalah-akamai, line 157: Here too

packages/manager/src/features/Account/CloseAccountDialog.tsx
```diff
@@ -64,7 +64,7 @@ const CloseAccountDialog: React.FC<CombinedProps> = (props) => {
   }, [inputtedUsername, profile]);
 
   const inputRef = React.useCallback(
-    (node) => {
+    (node: any) => {
```
jaalah-akamai, line 67: `node: HTMLElement | null`

jaalah-akamai (approved):
Nice work @bnussman 🚀

cpathipa (approved):
Nice work! @bnussman Over all looks good to me. left few comments related to double wrapping fragments. 

Observations: 
Couple tests failed in E2E suite. But, running them individually got passed. 

image

dwiley-akamai (approved):

bnussman-akamai:
Looks like we might be blocked by two things
- Our old Typescript version
- `@mui/styles`

PR #8826, opened 2023-02-27
fix: [M3-6178] - Disable/hide showAll for  PaginationFooter

cpathipa:
## Description 📝

- Avoids browser getting crash  or slowness when tried to render 1000+ records using PaginationFooter component.
- Improves performance of DisplayLinodes component.

**What does this PR do?**
Disable / Hides "Show All" option in `PaginationFooter` component. 
## Preview 📷
image

## How to test 🧪
- Checkout this branch and navigate to Linode landing page
- Set the mock service or choose an account that has 1000+ Linode's
- Check the page performance of loading Linod's page
- Open the pagination menu at the bottom and observe it shouldn't show "Show All".
- Also test DisplayGroupedLinodes, GroupedEntitiesByTag (http://localhost:3000/firewalls/) and DomainRecords (http://localhost:3000/domains )
- Incase of any question for testing please reach out. 

- Checkout develop branch
- Navigate to Linode landing page
- Set the mock service or choose an account that has 1000+ Linode's
- Check the page performance of loading Linod's page
- Open the pagination menu at the bottom and observe it shouldn't show "Show All".

yarn test


packages/manager/src/components/EntityTable/GroupedEntitiesByTag.tsx
```diff
@@ -132,7 +132,8 @@ export const GroupedEntitiesByTag: React.FC<CombinedProps> = (props) => {
                                   pageSize={pageSize}
                                   page={page}
                                   eventCategory={'Entity table'}
-                                  showAll
+                                  // Disabling showAll as it is impacting page perofrmance.
+                                  showAll={false}
```
dwiley-akamai, line 136: "performance"

dwiley-akamai (commented):
Although disabling `showAll` is a good step, it doesn't solve the performance issue if a user already had that option saved as a preference. This is what you would end up seeing (notice the "Select a value" since the value of `Infinity` never gets pushed to the array since `showAll` is disabled):

Screenshot 2023-02-27 at 5 14 53 PM

I think we'll want some logic in these components where `<PaginationFooter />` gets used so that if `pageSize === Infinity`, we set the page size to `100` in such a way that we downsize their preference from `Infinity` to `100` (should be able to use `handlePageSizeChange` that is exposed as a prop from the `<Paginate />` component or by the `usePagination` hook for this)

cpathipa:
> Although disabling `showAll` is a good step, it doesn't solve the performance issue if a user already had that option saved as a preference. This is what you would end up seeing (notice the "Select a value" since the value of `Infinity` never gets pushed to the array since `showAll` is disabled):
> 
> Screenshot 2023-02-27 at 5 14 53 PM
> 
> I think we'll want some logic in these components where `<PaginationFooter />` gets used so that if `pageSize === Infinity`, we set the page size to `100` in such a way that we downsize their preference from `Infinity` to `100` (should be able to use `handlePageSizeChange` that is exposed as a prop from the `<Paginate />` component or by the `usePagination` hook for this)

@dwiley-akamai Thank you for the feedback. I see what you mean, I will look into that...

cpathipa:
@dwiley-akamai Addressed the above issue.
https://user-images.githubusercontent.com/119517080/223127902-2cf8ca29-6de2-44a9-a119-3f6d051ebd07.mov

dwiley-akamai (approved):

packages/manager/src/hooks/useInfinitePageSize.ts
```diff
@@ -2,8 +2,13 @@ import { useMemo } from 'react';
...
 
 export const useInfinitePageSize = () => {
+  // Setting infinitePageSize to 100 since show all effecting page performance.
+  const infinitePageSize =
+    storage.infinitePageSize.get() === Infinity
+      ? 100
+      : storage.infinitePageSize.get();
   return {
-    infinitePageSize: useMemo(storage.infinitePageSize.get, []),
+    infinitePageSize: useMemo(() => infinitePageSize, []),
```
jaalah-akamai, line 11: `infinitePageSize` is a constant value that never changes, memoizing it is unnecessary here. Returning it directly would be more performant.
```
return {
    infinitePageSize,
    setInfinitePageSize: storage.infinitePageSize.set,
}
```

packages/manager/src/features/linodes/LinodesLanding/DisplayLinodes.tsx
```diff
@@ -80,12 +80,14 @@ const DisplayLinodes: React.FC<CombinedProps> = (props) => {
...
-    return acc;
-  }, 0);
+  const numberOfLinodesWithMaintenance = React.useMemo(() => {
+    return data.reduce((acc, thisLinode) => {
+      if (thisLinode.maintenance) {
+        acc++;
+      }
+      return acc;
+    }, 0);
+  }, [JSON.stringify(data)]);
```
jaalah-akamai, line 90: This works, but we could probably simplify this using `filter`. What do you think? 
```
  return data.filter(thisLinode => thisLinode.maintenance).length;
```
Another option is that since the linodes with maintenance most likely will be very few (almost definitely less than the 25 limit), we could break out of the loop once we find the first one. 
```
 const numberOfLinodesWithMaintenance = React.useMemo(() => {
    let count = 0;
    for (let i = 0; i < data.length; i++) {
      if (data[i].maintenance) {
        count++;
        if (count > 1) {
          break;
        } // exit early if we've found more than one
      }
    }
    return count;
  }, [data]);
```

cpathipa: We could simplify this. I think using `some` would be best here as it covers both options mentioned above:
```
const hasMaintenance = data.some(thisLinode => thisLinode.maintenance && thisLinode.status === 'running');
```
As I see it, this also impacts `getMinimumPageSizeForNumberOfItems`, and we have similar logic in three other places - `DisplayGroupedLinodes` and `PaginationFooter`. I feel like we should normalize this at some central location and reference it.

jaalah-akamai: Agreed 👍

jaalah-akamai (approved):
Good improvement - thanks for changes!

PR #8829, opened 2023-02-27
fix: Allow DBaaS e2e tests to pass regardless of engine minor version

jdamore-linode:
## Description 📝

**What does this PR do?**
This fixes a failing DBaaS test triggered by today's maintenance. The tests run against specific database engine versions, and today's maintenance changed which versions are available.

This fix increases the leniency so that the tests should continue to pass regardless of which minor versions of our database engines are available, so we won't have to update the tests every time we upgrade our available database engines. 

## How to test 🧪

Confirm that the following tests pass:
```bash
yarn cy:run -s "cypress/e2e/databases/create-database.spec.ts"
```

### DIFF
--- a/packages/manager/cypress/e2e/databases/create-database.spec.ts
+++ b/packages/manager/cypress/e2e/databases/create-database.spec.ts
@@ -1,8 +1,4 @@
-import {
-  ClusterSize,
-  Engine,
-  CreateDatabasePayload,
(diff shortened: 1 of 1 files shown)


jaalah-akamai (approved):

bnussman-akamai (approved):
Screenshot 2023-02-27 at 9 41 16 PM

PR #8840, opened 2023-03-03
refactor: [M3-5194] - RQ-ify Types endpoints

hkhalil-akamai:
## Description 📝

Migrate all Linode Type endpoints to React Query. Types refer to the plans available when configuring a Linode/LKE cluster (i.e., shared, GPU, bare metal, etc.)

The main complication I found during this refactor was the problem of "shadow plans." These are legacy linode types that aren't returned by the main `/types` and `/types-legacy` endpoints but can be fetched individually through `/types/{ID}`.

Originally, this problem was solved by a bit of code `AuthenticationWrapper` that would load all the account linodes on app startup, individually fetch all the unresolved types, and append them to the Redux store.

Now, this is handled through a second RQ hook called `useSpecificTypes()` which takes in an array of type IDs and fetches each one individually. I made sure to use this hook everywhere a "shadow plan" might be present, but I don't have any real examples of a linode with such a type in my account to test on, so I'm not certain all edge cases are handled.

Open Questions:

- [x] Are there still Linodes with "shadow plans" out there in the wild?
- [x] If so, is it possible to configure a new Linode with a such a plan for testing purposes?

## How to test 🧪
1. Ensure all affected components/hooks fetch and display plan details and pricing info as expected:

- `AddNodePoolDrawer`
- `BackupDrawer`
- `CreateCluster`
- `EnableBackupsDialog`
- `KubeCheckoutBar`
- `KubeClusterSpecs`
- `KubernetesClusterRow`
- `LinodeActionMenu`
- `LinodeCreateContainer`
- `LinodeEntityDetail`
- `LinodeResize`
- `LinodeTransferTable`
- `MigrateLinode`
- `MutationNotification`
- `NodePoolsDisplay`
- `RenderProgressEvent`
- `ResizeNodePoolDrawer`
- `SearchBar`
- `SearchLanding`
- `useAPISearch`
- `useExtendedLinode`

2. Unit tests: `yarn test`
3. E2E tests: `yarn cy:e2e`


packages/manager/src/features/linodes/LinodesDetail/LinodeResize/LinodeResize.tsx
```diff
@@ -539,7 +565,11 @@ export const isSmallerThanCurrentPlan = (
 };
 
 export default compose<CombinedProps, Props>(
-  withTypes,
+  // this awkward line avoids fetching all types until this dialog is opened which
+  // is important since LinodeResize is mounted on the default landing page
+  (component: React.ComponentType<CombinedProps>) => (props: CombinedProps) =>
+    withTypes(component, { enabled: props.open })(props),
```
hkhalil-akamai, line 571: TODO: is this the right way to accomplish this?

packages/manager/src/containers/types.container.ts
```diff
@@ -1,18 +1,54 @@
...
 
-export default connect((state: ApplicationState) => {
-  return {
-    typesData: state.__resources.types.entities,
-    typesLoading: state.__resources.types?.loading ?? false,
-    typesError: state.__resources.types?.error,
-  };
-});
+export interface WithSpecificTypesProps {
+  setRequestedTypes: (types: string[]) => void;
```
hkhalil-akamai, line 18: TODO: discuss this pattern in cafe

packages/manager/src/utilities/cleanArray.ts
```diff
@@ -0,0 +1,4 @@
+import { isNil } from 'ramda';
+
+export default <T>(array: (T | null | undefined)[]): T[] =>
+  array.filter((el) => !isNil(el)) as T[];
```
jaalah-akamai, line 4: Let's do this in javascript without ramda, we want to eventually get rid of that library:
`el !== null && el !== undefined`

packages/manager/src/features/linodes/index.tsx
```diff
@@ -38,13 +39,17 @@ const LinodesLandingWrapper: React.FC = React.memo(() => {
     { status: { '+or': ['pending, started'] } }
   );
   const { linodes } = useLinodes();
-  const { typesMap } = useTypes();
+  const typesQuery = useSpecificTypes(
+    cleanArray(Object.values(linodes.itemsById).map((linode) => linode.type))
```
jaalah-akamai, line 43: Curious if the team wants to adopt this approach to cleaning the array, may not be necessary.

jaalah-akamai (changes requested):

packages/manager/src/features/linodes/index.tsx
bnussman-akamai: I'm okay with not cleaning the array if it doesn't break anything.

packages/manager/src/queries/types.ts
```diff
@@ -0,0 +1,112 @@
...
+};
+
+export interface TypesQueryOptions {
+  includeLegacy?: boolean;
+  enabled?: boolean;
+}
+
+export const useAllTypes = (options?: TypesQueryOptions) => {
+  return useQuery<ExtendedType[], APIError[]>(
+    [queryKey, options ?? {}],
```
bnussman-akamai, line 90: React Query relation options should not be in the query key.

packages/manager/src/queries/types.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export const queryKey = 'types';
+
+export interface ExtendedType extends LinodeType {
+  heading: string;
+  subHeadings: string[];
+  isDeprecated: boolean;
+}
+
+export const extendType = (type: LinodeType): ExtendedType => {
```
bnussman-akamai, line 29: This may add complexity to the PR, but would it be possible for us to not "extend" the types data at the React Query level? I think it makes more sense to do the "extending" at the component level.

jaalah-akamai: Agreed - for clarity... the reason for this is that this can produce unintended consequences for other parts that will rely on this as the app grows and changes. We just want to do this in a more controlled way at the component level.

packages/manager/src/features/linodes/LinodesDetail/LinodeResize/LinodeResize.tsx
dwiley-akamai: the general approach seems correct to me, I will circle back to this though

dwiley-akamai (commented):
I still have a lot of code to sift through and testing to do, but I left a few comments for what I took a look at so far.

With regards to the question about shadow plans out in the wild, I'm going to reach out offline about how we might be able to check that.

packages/manager/src/features/linodes/index.tsx
hkhalil-akamai: The problem here is that the `linode.type` property is of type `string | null`. Without `cleanArray`, we would have to change the signature of `useSpecificTypes` to accept null IDs.

packages/manager/src/queries/types.ts
```diff
@@ -0,0 +1,112 @@
...
+  ])
+    .then((results) =>
+      results.reduce((acc: LinodeType[], cur) => {
+        acc.push(...cur.data);
+        return acc;
+      }, [])
+    )
+    .then((allTypes) => allTypes.map(extendType));
+
+const getSingleType = async (type: string, queryClient: QueryClient) => {
```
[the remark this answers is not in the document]
hkhalil-akamai: Actually this touches on a larger refactoring I wanted to bring up. Currently we're using a single global instance of `queryClient` but I think the recommended pattern is to use the `useQueryClient` hook to get the query client instance from context. This would allow use to use multiple distinct query client instances in our app if we needed to, or allow us to write unit tests with a custom query client.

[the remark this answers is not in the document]
hkhalil-akamai: But if the team feels like we should continue our current pattern, I can make the requested change.

packages/manager/src/queries/types.ts
hkhalil-akamai: This raises an interesting formatting consideration... the API returns plan labels like 'Linode 1GB' (w/o a space between 1 and 'GB'), whereas the `extendLinode` function adds a space to make it 'Linode 1 GB'. I've changed `extendLinode` to match the API for consistency but I think it's worth discussing which direction we want to go.

hkhalil-akamai: Moved to `src/utilities/extendType.ts`

packages/manager/src/features/linodes/index.tsx
jaalah-akamai: What do we think about having the `useSpecificTypes` hook filter for null or undefined which would simplify the usage of the hook? 

In other area's where we're directly calling `cleanArray`, I suppose I'm ok with... although I think this would achieve the same result:
```
const types = typesQuery
  .map((result) => result.data)
  .filter(isNotNullOrUndefined)
  .map(extendType);
```

jaalah-akamai (approved):
Nice, thanks for making the changes

packages/manager/src/App.tsx
```diff
@@ -302,7 +303,9 @@ export default compose(
   withSnackbar,
   withFeatureFlagProvider,
   withFeatureFlagConsumer,
-  withPreferences
+  withPreferences,
+  (component: React.ComponentType<WithTypesProps>) =>
+    withTypes(component, false)
```
bnussman-akamai, line 308: In my opinion I don't think it's worth blocking the app from rendering just because types has an error

```suggestion
```

bnussman-akamai (commented):
As much as I'm against the "fetch all" approach, I think it would be nice if we could do that for types. Is the only reason we are not doing that because of the shadow plans?

packages/manager/src/App.tsx
hkhalil-akamai: Just to be clear, you think L223 should be deleted as well?

bnussman-akamai: Hmm, honestly yeah. Now that I'm looking at it, a lot of this logic in `App.tsx` is broken anyway. For example, `volumesErrors` and `bucketsError` are not a thing anymore now that those have been React Queryified

bnussman-akamai (approved):
Nice work, there are a lot of conflicts now somehow :0

hkhalil-akamai:
Tested and confirmed working with an "shadow-plan" Linode instance.

dwiley-akamai (commented):
Still doing a final look-through of the code but the functionality is looking good to me so far.

One thing I did notice is that it seems we've had a regression on the Linodes landing page compared to prod, losing a space between the number and "GB" for the plan type:

Screenshot 2023-03-15 at 2 13 48 PM

packages/manager/src/features/Backups/BackupDrawer.tsx
```diff
@@ -103,6 +121,9 @@ export class BackupDrawer extends React.Component<CombinedProps, {}> {
       dismissSuccess();
       close();
     }
+    if (prevProps.linodesWithoutBackups !== this.props.linodesWithoutBackups) {
```
dwiley-akamai, line 124: Since `linodesWithoutBackups` is an array, do we need to do a deep equality check here?

packages/manager/src/features/Search/SearchLanding.tsx
```diff
@@ -147,9 +150,15 @@ export const SearchLanding: React.FC<CombinedProps> = (props) => {
...
-  const types = useSelector((state: ApplicationState) =>
-    Object.values(state.__resources.types.entities)
+
+  const typesQuery = useSpecificTypes(
+    linodes.map((linode) => linode.type).filter(isNotNullOrUndefined)
   );
+  const types = typesQuery
+    .map((result) => result.data)
+    .filter(isNotNullOrUndefined)
+    .map(extendType);
```
dwiley-akamai, line 160: We repeat this `.map().filter().map()` in many different files. Could we abstract it out into a utility function?

dwiley-akamai (commented):
I see that `{size}GB` regression in the Service Transfers table as well:

Screenshot 2023-03-16 at 11 20 50 AM

dwiley-akamai (approved):
LGTM, great work! Thanks for addressing the feedback 🚀

PR #8842, opened 2023-03-06
docs: [M3-6206] - Add commit standards to docs

carrillo-erik:
## Description 📝

**What does this PR do?**
Add the git commit format standards to the repo docs.

## How to test 🧪
Inspect the PULL_REQUEST_TEMPLATE and CONTRIBUTING docs


docs/CONTRIBUTING.md
```diff
@@ -13,7 +13,12 @@ Feel free to open an issue to report a bug or request a feature.
...
 3. Create a branch from `develop`, e.g. `$ git checkout develop && git pull && git checkout -b feature/my-feature`.
-4. Make your changes, commit them, then push them to your fork.
-5. Open a pull request against `develop`.
+4. Make your changes, commit them following the standards below, and then push them to your fork.
+5. Commit message format standard: `<commit type>: [JIRA-ticket-number] - <description>`
+
+    **commit types:** `feat`, `fix`, `refactor`, `style`, `build`, `chore`, `ci`, `docs`, `perf`, `test`
+    
+    e.g. `feat: [M#-1234] - Allow user to view their login history`
+6. Open a pull request against `develop`.
```
jaalah-akamai, line 22: For point 6 - I think we want to make clear that the title of the PR should also be in the format you suggested above

carrillo-erik: Thanks for pointing that out. I've made the changes.

docs/CONTRIBUTING.md
```diff
@@ -13,7 +13,12 @@ Feel free to open an issue to report a bug or request a feature.
...
 2. Clone your fork to your local machine.
 3. Create a branch from `develop`, e.g. `$ git checkout develop && git pull && git checkout -b feature/my-feature`.
-4. Make your changes, commit them, then push them to your fork.
-5. Open a pull request against `develop`.
+4. Make your changes, commit them following the standards below, and then push them to your fork.
+5. Commit message format standard: `<commit type>: [JIRA-ticket-number] - <description>`
+
+    **commit types:** `feat`, `fix`, `refactor`, `style`, `build`, `chore`, `ci`, `docs`, `perf`, `test`
+    
+    e.g. `feat: [M#-1234] - Allow user to view their login history`
```
cpathipa, line 21: Do we need to include JIRA tickets in our commits? Additionally, if we push changes with multiple commits to the same branch, should we follow the same pattern like below example?"

example: 
`feat: [M#-1234] - <commit message 1>`
`chor: [M#-1234] - <commit message 2>`
`chor: [M#-1234] - <commit message 3>`

docs/CONTRIBUTING.md
```diff
@@ -13,7 +13,12 @@ Feel free to open an issue to report a bug or request a feature.
 1. Fork this repository.
 2. Clone your fork to your local machine.
 3. Create a branch from `develop`, e.g. `$ git checkout develop && git pull && git checkout -b feature/my-feature`.
-4. Make your changes, commit them, then push them to your fork.
-5. Open a pull request against `develop`.
+4. Make your changes, commit them following the standards below, and then push them to your fork.
+5. Commit message format standard: `<commit type>: [JIRA-ticket-number] - <description>`
+
+    **commit types:** `feat`, `fix`, `refactor`, `style`, `build`, `chore`, `ci`, `docs`, `perf`, `test`
```
cpathipa, line 19: Would adding descriptions for each commit type help new team members understand the context?

- feat: New feature for the user (not a part of the code, or ci, ...)
- fix: Bugfix for the user (not a fix to build something, ...)
- docs: Changes to the documentation
- style: This could be the code's styling or general styling changes. Does not change any functionality.
- refactor: Refactoring production code. For example: Renaming a variable
- test: Only changes current or new tests. Does not change the production code
- chore: Does not impact production.

jaalah-akamai: Yea I think we should add definitions that are concise 👍

docs/CONTRIBUTING.md
jaalah-akamai: I think this is good practice, but since we should be squashing the commits it probably doesn't matter. I will probably be doing this though

dwiley-akamai: Since we're squashing commits I think it should come down to developer preference (for the commits prior to the final one upon merge)

docs/PULL_REQUEST_TEMPLATE.md
```diff
@@ -11,3 +11,15 @@
...
+
+**Remove this section before opening the pull request**
+
+**Make sure your commit message and PR title are formatted as shown below**
+
+`<commit type>: [JIRA-ticket-number] - <description>`
+
+**commit types:** `feat`, `fix`, `refactor`, `style`, `build`, `chore`, `ci`, `docs`, `perf`, `test`
+    
+e.g. `feat: [M#-1234] - Allow user to view their login history`
```
dwiley-akamai, line 25: Is there a reason we wouldn't be explicit about `M3` here? Just for the sake of clarity to avoid PRs being opened with `M-1234` in the title

docs/CONTRIBUTING.md
hana-linode: @carrillo-erik Can we add `change` as a commit type for changes that don't fit into the above?

docs/PULL_REQUEST_TEMPLATE.md
```diff
@@ -11,3 +11,25 @@
...
+`feat`: New feature for the user (not a part of the code, or ci, ...).
+`fix`: Bugfix for the user (not a fix to build something, ...).
+`refactor`: Refactoring production code. For example: Renaming a variable.
+`style`: General styling changes. Does not change any functionality.
+`build`: Changes to the application build.
+`chore`: Other changes that do not impact production code.
+`ci`: Changes to the CI pipeline.
+`docs`: Changes to the documentation docs.
+`perf`: Performance changes.
+`test`" New tests or changes to existing tests. Does not change the production code.
```
mjac0bs, line 33:
```suggestion
`test`: New tests or changes to existing tests. Does not change the production code.
```

mjac0bs (approved):
Looks like one stray `"` should be a `:`, but otherwise LGTM. @hana-linode mentioned including a `change` commit type for anything that doesn't fall into the other categories. I'm not entirely opposed, but feel like this is a long enough list of commit types that commits should probably fit into *something* already listed. 

I'm also assuming that the Markdown formatting for commit types will display as a list with each type on its own line, as it did when I checked out this branch and looked at in my editor. When I tried to "View File" in this PR, Github wanted to display the markdown for the commit types like below, which was odd. 
image

mjac0bs (approved):

PR #8849, opened 2023-03-07
test: [M3-5952] - Object Storage object upload, access, delete tests

jdamore-linode:
## Description 📝

**What does this PR do?**
This adds a new end-to-end test to confirm that Object Storage objects behave as expected. It confirms that objects can be uploaded, deleted, replaced, accessed when public, and that folders can be created and deleted.

## How to test 🧪

`yarn up`, and then:

```bash
yarn cy:run -s "cypress/e2e/objectStorage/object-storage.e2e.spec.ts"
```

Alternatively, to run the tests interactively run `yarn cy:debug` and search for `object-storage.e2e.spec.ts`.

### DIFF
--- a/packages/manager/cypress/e2e/objectStorage/object-storage.e2e.spec.ts
+++ b/packages/manager/cypress/e2e/objectStorage/object-storage.e2e.spec.ts
@@ -2,6 +2,7 @@
  * @file End-to-end tests for Object Storage operations.
  */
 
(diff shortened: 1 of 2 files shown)


dwiley-akamai (approved):

cpathipa (approved):
Nice work! @jdamore-linode The comments were very descriptive and made it easy to understand the scope and goal of the test.
image

PR #8850, opened 2023-03-07
feat: [M3-6146] - Add User Data to Linode Rebuild flow

dwiley-akamai:
## Description 📝
Adds User Data section to the Linode Rebuild flow when an image with `cloud-init` capability is selected.

## Preview 📷
Screenshot 2023-03-07 at 11 33 22 AM

## To-Do
- [x] Add unit tests

## How to test 🧪
- Ensure that the existing User Data work elsewhere in the app has not been adversely impacted by changes made on this branch
- Confirm that Linode rebuild requests go through with the `{ metadata: { user_data: [encoded string ...] } }` property (or `null` instead of encoded string when the Reuse checkbox is marked)

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeRebuild/RebuildFromImage.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeRebuild/RebuildFromImage.tsx
@@ -108,7 +108,7 @@ export const RebuildFromImage: React.FC<CombinedProps> = (props) => {
     setUserData(userData);
   };
 
-  const handleReuseUserDataChange = () => {
+  const handleShouldReuseUserDataChange = () => {
     setShouldReuseUserData((shouldReuseUserData) => !shouldReuseUserData);
   };
 
@@ -268,7 +268,7 @@ export const RebuildFromImage: React.FC<CombinedProps> = (props) => {
                       <Box>
                         <CheckBox
                           checked={shouldReuseUserData}
-                          onChange={handleReuseUserDataChange}
+                          onChange={handleShouldReuseUserDataChange}
                           text="Reuse user data previously provided for this Linode."
                           sxFormLabel={{ paddingLeft: '2px' }}
                         />
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeRebuild/RebuildFromImage.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeRebuild/RebuildFromImage.tsx
@@ -30,6 +30,7 @@ import {
 } from 'src/utilities/formikErrorUtils';
 import scrollErrorIntoView from 'src/utilities/scrollErrorIntoView';
 import { extendValidationSchema } from 'src/utilities/validatePassword';
+import Divider from 'src/components/core/Divider';
 
 const useStyles = makeStyles((theme: Theme) => ({
   root: {
@@ -102,7 +103,7 @@ export const RebuildFromImage: React.FC<CombinedProps> = (props) => {
   };
 
   const handleReuseUserDataChange = () => {
-    setReuseUserData(!reuseUserData);
+    setReuseUserData((reuseUserData) => !reuseUserData);
   };
 
   React.useEffect(() => {
@@ -133,6 +134,11 @@ export const RebuildFromImage: React.FC<CombinedProps> = (props) => {
         .filter((u) => u.selected)
         .map((u) => u.username),
     };
+
+    // If no user data has been added, do not include the metadata property in the payload.
+    if (!userData && !reuseUserData) {
+      delete params['metadata'];
+    }
 
     // @todo: eventually this should be a dispatched action instead of a services library call
     rebuildLinode(linodeId, params)
@@ -239,14 +245,18 @@ export const RebuildFromImage: React.FC<CombinedProps> = (props) => {
                 passwordHelperText={passwordHelperText}
               />
               {shouldDisplayUserDataAccordion ? (
-                <UserDataAccordion
-                  userData={userData}
-                  onChange={handleUserDataChange}
-                  flowSource="rebuild"
-                  reuseUserData={reuseUserData}
-                  onReuseUserDataChange={handleReuseUserDataChange}
-                  disabled={reuseUserData}
-                />
+                <>
+                  <Divider spacingTop={40} />
+                  <UserDataAccordion
+                    userData={userData}
+                    onChange={handleUserDataChange}
+                    reuseUserData={reuseUserData}
+                    onReuseUserDataChange={handleReuseUserDataChange}
+                    disabled={reuseUserData}
+                    renderNotice
+                    renderCheckbox
+                  />
+                </>
               ) : null}
               <ActionsPanel className={classes.actionPanel}>
                 <TypeToConfirm
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodeRebuild/RebuildFromImage.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodeRebuild/RebuildFromImage.tsx
@@ -1,4 +1,8 @@
-import { rebuildLinode, RebuildRequest } from '@linode/api-v4/lib/linodes';
+import {
+  rebuildLinode,
+  RebuildRequest,
+  UserData,
+} from '@linode/api-v4/lib/linodes';
 import { RebuildLinodeSchema } from '@linode/validation/lib/linodes.schema';
 import { Formik, FormikProps } from 'formik';
 import { useSnackbar } from 'notistack';
@@ -13,6 +17,7 @@ import Grid from 'src/components/Grid';
 import ImageSelect from 'src/components/ImageSelect';
 import TypeToConfirm from 'src/components/TypeToConfirm';
 import { resetEventsPolling } from 'src/eventsPolling';
+import UserDataAccordion from 'src/features/linodes/UserDataAccordion';
 import userSSHKeyHoc, {
   UserSSHKeyProps,
 } from 'src/features/linodes/userSSHKeyHoc';
@@ -55,11 +60,15 @@ export type CombinedProps = Props & UserSSHKeyProps;
 interface RebuildFromImageForm {
   image: string;
   root_pass: string;
+  metadata?: UserData;
 }
 
 const initialValues: RebuildFromImageForm = {
   image: '',
   root_pass: '',
+  metadata: {
+    user_data: '',
+  },
 };
 
 export const RebuildFromImage: React.FC<CombinedProps> = (props) => {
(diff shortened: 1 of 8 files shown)


packages/manager/src/features/linodes/UserDataAccordion.tsx
```diff
@@ -1,40 +1,71 @@
...
+    paddingLeft: 2,
   },
-  accordionDetail: {
-    padding: '0px 24px 24px 24px',
+});
+
+const StyledDiv = styled('div')({
+  '& .notice': {
+    padding: '8px !important',
+    marginBottom: '0px !important',
```
dwiley-akamai, line 54: not ideal, open to styling tweaks that would circumvent the need for these `!important`'s

packages/manager/src/features/linodes/LinodesDetail/LinodeRebuild/RebuildFromImage.tsx
```diff
@@ -101,6 +126,9 @@ export const RebuildFromImage: React.FC<CombinedProps> = (props) => {
     const params: RebuildRequest = {
       image,
       root_pass,
+      metadata: {
+        user_data: userData && !reuseUserData ? window.btoa(userData) : null,
+      },
```
hana-linode, line 131: We only want to add metadata to the request if the user enters User Data or clicks on the reuse button (and only then do we send null). Right now, if we rebuild without touching the User Data input, we send null.

packages/manager/src/features/linodes/UserDataAccordion.tsx
```diff
@@ -1,40 +1,71 @@
...
-import Notice from 'src/components/Notice';
 import Link from 'src/components/Link';
+import Notice from 'src/components/Notice';
 import TextField from 'src/components/TextField';
 
 interface Props {
   userData: string | undefined;
   onChange: (userData: string) => void;
   disabled?: boolean;
+  flowSource?: 'rebuild' | null;
```
hana-linode, line 16: Thoughts on adding a `renderCopy`, `renderNotice`, `renderAction` prop instead of `flowSource` to make this component more generic?

hana-linode (changes requested):
Can we make the spacing between Add User Data and SSH Keys the same?

image

packages/manager/src/features/linodes/UserDataAccordion.tsx
jaalah-akamai: I would keep this as a normal `div` in the JSX with no styles attached to it. Then allow `Notice` to accept an `sx` prop where you pass the `padding` and `marginBottom` down to it. You won't need the important tags and there will be no need to reference the `.notice` class which we need to avoid doing from now on as we refactor.

packages/manager/src/features/linodes/UserDataAccordion.tsx
```diff
@@ -56,50 +89,41 @@ const UserDataAccordion = (props: Props) => {
...
-          root: classes.accordionSummary,
-        },
+        sx: { padding: '5px 24px 0px 24px' },
       }}
       detailProps={{
-        classes: {
-          root: classes.accordionDetail,
+        sx: renderNotice
+          ? { padding: '0px 24px 24px 0px' }
+          : { padding: '0px 24px 24px 24px' },
```
jaalah-akamai, line 105: You can simplify this by doing something like:
```
const sxDetails = {
    padding: `0px 24px 24px ${renderNotice ? 0 : 24}px`,
  }
```

jaalah-akamai (changes requested):

dwiley-akamai:
> I would keep this as a normal div in the JSX with no styles attached to it. Then allow Notice to accept an sx prop where you pass the padding and marginBottom down to it. You won't need the important tags and there will be no need to reference the .notice class which we need to avoid doing from now on as we refactor.

`<Notice />` is built largely with `<Grid />`, which is a custom implementation to mimic MUIv4's grid behavior, and `src/components/core/Grid.tsx` returns a `<div>`. If I'm understanding correctly, the `sx` prop in `Notice.tsx` would be applied to the top-level `<Grid />` instance. Is the `sx` prop approach still viable here @jaalah-akamai (since we're not using the MUIv5 Grid)?

packages/manager/src/features/linodes/LinodesDetail/LinodeRebuild/RebuildFromImage.tsx
```diff
@@ -242,3 +313,9 @@ export const RebuildFromImage: React.FC<CombinedProps> = (props) => {
 const enhanced = compose<CombinedProps, Props>(userSSHKeyHoc);
 
 export default enhanced(RebuildFromImage);
+
+const StyledNotice = styled(Notice)({
+  // @TODO: Remove the !important's once Notice.tsx has been refactored to use MUI's styled()
+  padding: '8px !important',
+  marginBottom: '0px !important',
+});
```
dwiley-akamai, line 321: As we introduce `styled()` components into the codebase, my position is that we ought to position them at the end of our files. There will be many times where multiple styled components are used in a single component, and the beginning of the file can get cluttered, disrupting the visually logical flow of imports --> Props interfaces/types --> component --> utility functions at the end. That flow at the top gives you the high-level view of the lay of the land that you need to jump into a component, understand it, and start making logic changes you may have to make.

I added this as a topic for our Cafe so we can decide on a convention.

If we decide to go with this, I'll move `StyledHelpIcon` in `UserDataAccordion.tsx` down in that file as well.

jaalah-akamai (approved):

packages/manager/src/features/linodes/LinodesDetail/LinodeRebuild/RebuildFromImage.tsx
```diff
@@ -203,6 +248,32 @@ export const RebuildFromImage: React.FC<CombinedProps> = (props) => {
...
+                    renderNotice={() => (
+                      <StyledNotice
+                        success
+                        text="Adding new user data is recommended as part of the rebuild process."
+                      />
+                    )}
+                    renderCheckbox={() => (
+                      <Box>
+                        <CheckBox
+                          checked={reuseUserData}
```
bnussman-akamai, line 267: Can we use a more descriptive variable name for `reuseUserData`? Maybe something like `shouldReuseUserData`? reuseUserData sounds more like a function name to me.

Also what makes us need to use state here over the formik form?

```suggestion
                          checked={reuseUserData}
```

dwiley-akamai: With regards to the state for `checked`, since we wouldn't want to submit `shouldReuseUserData` as part of the payload, I think we're in a similar position here as we were in for `AccessKeyDrawer.tsx`

packages/manager/src/features/linodes/LinodesDetail/LinodeRebuild/RebuildFromImage.tsx
```diff
@@ -101,11 +132,20 @@ export const RebuildFromImage: React.FC<CombinedProps> = (props) => {
...
       authorized_users: userSSHKeys
         .filter((u) => u.selected)
         .map((u) => u.username),
     };
 
+    // If no user data has been added, do not include the metadata property in the payload.
+    if (!userData && !shouldReuseUserData) {
+      delete params['metadata'];
+    }
+
```
hana-linode, line 148: Sorry, looks like I misunderstood the rebuild endpoint 😅 
1. By default, if no user data has been added/user hasn't clicked on the re-use checkbox, we should send user_data as null to the rebuild endpoint
2. If the user checks the reuse checkbox, then we should not send any metadata/user data
3. If metadata/user data is given, we send user_data encoded in base64

hana-linode (changes requested):

packages/manager/src/features/linodes/LinodesDetail/LinodeRebuild/RebuildFromImage.tsx
dwiley-akamai: This should be good to go now 🚀

cpathipa (commented):
Overall, it looks great. I only have a few comments left related to the naming of JSX elements.

packages/manager/src/features/linodes/UserDataAccordion.tsx
```diff
@@ -121,8 +87,43 @@ const UserDataAccordion = (props: Props) => {
...
+    user group(s).
+  </Typography>
+);
+
+const acceptedFormatsCopy = (
+  <Typography>
+    <br /> Accepted formats are YAML and bash.{' '}
+    <Link to="https://www.linode.com/docs">Learn more.</Link>
+  </Typography>
+);
```
cpathipa, line 129:
```suggestion
const AcceptedFormatsCopy = () =>  (
  <Typography>
    <br /> Accepted formats are YAML and bash.{' '}
    <Link to="https://www.linode.com/docs">Learn more.</Link>
  </Typography>
);
```

packages/manager/src/features/linodes/UserDataAccordion.tsx
```diff
@@ -121,8 +87,43 @@ const UserDataAccordion = (props: Props) => {
...
+const userDataExplanatoryCopy = (
+  <Typography>
+    <Link to="https://cloudinit.readthedocs.io/en/latest/reference/examples.html">
+      User Data
+    </Link>{' '}
+    is part of a virtual machine&rsquo;s cloud-init metadata that contains
+    anything related to a user&rsquo;s local account, including username and
+    user group(s).
+  </Typography>
+);
```
cpathipa, line 122:
```suggestion
const UserDataExplanatoryCopy = () => (
  <Typography>
    <Link to="https://cloudinit.readthedocs.io/en/latest/reference/examples.html">
      User Data
    </Link>{' '}
    is part of a virtual machine&rsquo;s cloud-init metadata that contains
    anything related to a user&rsquo;s local account, including username and
    user group(s).
  </Typography>
);
```

packages/manager/src/features/linodes/UserDataAccordion.tsx
```diff
@@ -56,50 +39,33 @@ const UserDataAccordion = (props: Props) => {
...
-        </Link>{' '}
-        is part of a virtual machine&rsquo;s cloud-init metadata that contains
-        anything related to a user&rsquo;s local account, including username and
-        user group(s). <br /> Accepted formats are YAML and bash.{' '}
-        <Link to="https://www.linode.com/docs">Learn more.</Link>
-      </Typography>
+      {renderNotice ? (
+        <div data-testid="render-notice">{renderNotice()}</div>
+      ) : (
+        userDataExplanatoryCopy
```
cpathipa, line 66:
```suggestion
        <UserDataExplanatoryCopy />
```

packages/manager/src/features/linodes/UserDataAccordion.tsx
```diff
@@ -56,50 +39,33 @@ const UserDataAccordion = (props: Props) => {
...
+      )}
+      {acceptedFormatsCopy}
```
cpathipa, line 68:
```suggestion
      <AcceptedFormatsCopy />
```

packages/manager/src/features/linodes/LinodesDetail/LinodeRebuild/RebuildFromImage.tsx
```diff
@@ -168,6 +208,13 @@ export const RebuildFromImage: React.FC<CombinedProps> = (props) => {
           handleRebuildError(status.generalError);
         }
 
+        const shouldDisplayUserDataAccordion = Boolean(
+          values.image &&
+            _imagesData
+              ?.find((image) => image.id === values.image)
+              ?.capabilities?.includes('cloud-init')
+        );
+
```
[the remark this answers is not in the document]
dwiley-akamai: We wouldn't be able to use `useMemo` in this spot since it's currently within a callback, and I don't think we're experiencing a performance hit here so a refactor to be able to use `useMemo` seems like an over-optimization to me

packages/manager/src/features/linodes/UserDataAccordion.tsx
dwiley-akamai: Our current precedent in the codebase seems to point away from doing this so I'll keep the casing as-is for now, but it can be a Cafe item if we want to revisit it

hana-linode (approved):
🚀

cpathipa (approved):

PR #8860, opened 2023-03-09
deps: [M3-6230] - Remove unused packages + update lint-staged

jaalah-akamai:
## Description 📝
- Removes old unused packages from root `package.json`.
- Fixed misspelling of `z-index`


bnussman-akamai (commented):
Can you run `yarn` in the repo's root to update the lockfile?

bnussman-akamai (approved):

yarn.lock
```diff
@@ -9193,11 +9181,6 @@ is-obj@^1.0.1:
   resolved "https://registry.yarnpkg.com/is-obj/-/is-obj-1.0.1.tgz#3e4729ac1f5fde025cd7d83a896dab9f4f67db0f"
   integrity sha512-l4RyHgRqGN4Y3+9JHVrNqO+tN0rV5My76uW5/nuO4K1b6vw5G8d/cmFjP9tRfEsdhZNt0IFdZuK/c2Vr4Nb+Qg==
 
-is-obj@^2.0.0:
```
mjac0bs, line 9196: Is this one fine to remove? `yarn why` had this output:
```
=> Found "is-obj@1.0.1"
info Reasons this module exists
   - "_project_#lint-staged#stringify-object" depends on it
   - Hoisted from "_project_#lint-staged#stringify-object#is-obj"
info Disk size without dependencies: "16KB"
info Disk size with unique dependencies: "16KB"
info Disk size with transitive dependencies: "16KB"
info Number of shared dependencies: 0
✨  Done in 0.37s.
```

jaalah-akamai: This is an interesting one because it's needed for `stringify-object` which is used by `lint-staged`. However, in the latest versions of `lint-staged`, this dependency has been removed. So yes we need it, but if we upgrade to latest `lint-staged` and migrated safely, we could remove it.

jaalah-akamai: Just checked, we can upgrade to latest since we're using node v18+ - I'll update

bnussman-akamai (approved):
Nice job with the upgrade! Verified with `yarn why` that the removed yarn resolutions are no longer needed ✅

mjac0bs (approved):

PR #8861, opened 2023-03-09
feat: [M3-6159] - Improve Linode disk downsize error messaging

mjac0bs:
## Description 📝

**What does this PR do?**
Increases the amount of feedback provided to users on Linode disk resize actions, particularly failed disk resizes.

Summary of changes:
- Adds helper text to the `LinodeDiskDrawer` to provide (1) a link to documentation on disk resizing and (2) a text tooltip for command to find minimum disk size. The goal of these changes is to provide more context for the user on disk resizing limits to help mitigate a resize error in the first place.
- Adds `HelpIcon` tooltip for disabled Add a Disk button to replace tooltip on disabled button hover to make presence of tooltip more visible to the user. 
- Adds `HelpIcon` tooltip for disabled Resize button to make the reason for disabling clearer: the Linode must be powered off. There is already a tooltip on the Delete action, but the Resize action was missing this context, requiring a user to guess why Resize was unavailable for a powered on Linode.
- Adds persistent error `Toast` notifying user of failed disk resize. The goal of this is to provide more visible error messaging to the user when a disk cannot be resized. (Note: This was changed from an error `Notice` in the mocks to align with our current method of displaying failed event errors.) Previously, the only visible feedback was in the notification menu with an event message on failure of the`disk_resize` action; no error message was returned on the screen itself.
- Update `disks_resize` failure event message in notification menu/events landing to include a docs link about disk resizing. This same doc is also linked in the error `Toast`. 
- Updates tests to after moving some event message logic around and adding the ability to append text to an event message when generated. 
- Creates new `Code` component (used in "Check disk usage" text tooltip for determining minimum disk size) to apply to some styling to text that is a Linux command.

TODO: (We can start reviewing without these completed.)
- [x] Finalize docs links
- [ ] Finalize copy 
- [x] Add GA tags

## Preview 📷
Tooltips:
Screenshot 2023-03-13 at 10 41 30 AM

Disk drawer copy updates:
Screenshot 2023-03-13 at 10 37 23 AM

Error messages:
Screenshot 2023-03-13 at 10 32 34 AM

## How to test 🧪

1. Have at least one Linode on your account. For the purposes of verifying step 3, using a nanode is probably easiest.
2. Navigate to the Storage tab of the Linode Details page of a powered on Linode. 
3. Verify that the help icon next to the disabled Add a Disk button displays a tooltip about disk space. Verify that hovering over the button no longer displays this tooltip.
4. Verify that the help icon next to the disabled Resize button displays a tooltip about powering down the Linode. This copy should match the tooltip copy for the Delete action.
5. Power down the Linode (or navigate to the details page for a different Linode that is already off).
6. Verify that the Resize action is no longer disabled and does not display a help icon next to it. 
7. Click Resize and verify the copy changes and text tooltip in the Resize drawer match the internal mocks and the docs link redirects as expected. 
8. Change the size to something that will be below the minimum size possible for the disk (e.g. `1 MB`) and submit the resize. 
9. Wait for the resize action to attempt to complete. Verify that when the resize fails, a *persistent* error toast displays. Verify the link in the error toast redirects to a docs page on resizing a disk.
10. Check the notification menu, the events landing page, and the Activity Feed tab (on the linode details page) to verify that an event displays for the resize failure and that it contains a valid docs link that also redirects to the docs page on resizing a disk.

```
yarn test Event 
```


packages/manager/src/components/Code/Code.tsx
```diff
@@ -0,0 +1,21 @@
...
+  return <Span>{children}</Span>;
+};
+
+export default Code;
+
+const Span = styled('span')({
+  backgroundColor: '#EEEEEE',
+  fontWeight: 'bold',
+  fontFamily: 'Courier',
+  color: '#000000',
```
mjac0bs, line 20: @jaalah-akamai  Any feedback here on styling is appreciated. My understanding is that something like `theme.color.grey8` is no longer accessible using the styles [API](https://mui.com/material-ui/customization/theming/#theme-configuration-variables). The closest I could get was `theme.palette`. I attempted to update [theme.ts](https://github.com/linode/manager/blob/develop/packages/manager/src/themes.ts#L160) and themeFactory.ts to add these new code colors for `text` and `background ` to avoid hardcoding the hexes in this file, but wasn't successful there. I might have been taking the wrong approach.

jaalah-akamai: Does this give you access to the theme? 
```
const StyledButton = styled('span')(({ theme }) => ({
  // ... 
}));
```

mjac0bs: Yeah, sorry, should have mentioned that I did try accessing theme like that, but these are the available options, which does not include `.color`: 
image

mjac0bs: Update: `theme.color.[value]` is valid code, just displaying as a tsc error in my editor for some reason. Updated [9dc5ee6](https://github.com/linode/manager/pull/8861/commits/9dc5ee6bfe354be68dbc16570d381c178a85af29) to use theme colors; leaving this unresolved for now to remind myself to test with system theme PR in anticipation of text color issues.

jdamore-linode:
I haven't figured out the cause, but I'm seeing a few instances where toasts that are supposed to appear are no longer appearing:

- `Volume <volume label> successfully attached.` upon `volume_attach` event
- `Volume successfully deleted.` upon `volume_delete` event
- `Image <image label> is now available.` upon `image_upload` event

(These were all caught by tests, so there may be other toasts that are affected that I'm not aware of)

To run the impacted tests:

```bash
yarn cy:run -s "cypress/e2e/images/machine-image-upload.spec.ts,cypress/e2e/volumes/attach-volume.spec.ts,cypress/e2e/volumes/delete-volume.spec.ts"
```

packages/manager/src/features/ToastNotifications/ToastNotifications.tsx
```diff
@@ -98,6 +110,14 @@ class ToastNotifications extends React.PureComponent<WithSnackbarProps, {}> {
...
               `Error creating Image ${secondaryLabel}.`
             );
+          case 'disk_resize':
+            return _toastWithPersist(
+              false,
+              true,
+              `Disk ${secondaryLabel} resized successfully.`,
+              `Disk resize failed. <a href="https://www.linode.com/docs/products/compute/compute-instances/guides/disks-and-storage/" target=_blank>Learn more about resizing restrictions.</a>`,
+              true // includesLink
+            );
```
jaalah-akamai, line 120: I realize we're following an older pattern here, but this is something we should avoid/cleanup in the future. As noted by your comment, it's very difficult to know what these parameters do without referencing the original function. We should move to something like:

```
_toastWithPersist({
  enqueueSnackbar: () => {},
  eventStatus: '...'
  persistSuccessMessage: false,
  persistFailureMessage: true,
  successMessage: '...'
  failureMessage: '...'
  includesLink: true
})
```

jaalah-akamai (approved):

jaalah-akamai:
Changes look really good, nice job on this 🚀

cpathipa (approved):
LGTM! I left a couple of comments, but they're only nit-picky.

packages/manager/src/features/ToastNotifications/ToastNotifications.tsx
mjac0bs: @jaalah-akamai Could you give another look at [`ToastNotifications.tsx`](https://github.com/linode/manager/pull/8861/files#diff-3080c7f10764113de7db38e895488b809d33d9ac32b94a0a8b377981c3f79e2a)? I ended up refactoring the `toastSuccessAndFailure` function and removed the use of currying. I thought it was best to include these changes in this PR because of the issues that @jdamore-linode found when running E2E tests and discovering vanished toasts. It seemed that, using the curried function, even though some parameters were optional, they were not actually optional and instead required `undefined` (or `false`, in case of bool) to be passed in to the `_toast` or `_toastWithPersist` functions if we weren't including a link in the toast. 

After these changes, the previously failing E2Es are passing for me and I'm seeing expected behavior with toasts.

jdamore-linode (approved):
Approved pending a fix for the TS error in `InlineMenuAction.tsx`

Confirmed that tests pass and that the toasts that were missing before are showing up now. Nice work @mjac0bs!

jaalah-akamai (approved):
Thanks @mjac0bs for making these changes! This improves the readability and extensibility of those functions. 🚀

PR #8862, opened 2023-03-09
chore: [M3-6232] - Update `VolumeStatus` type and logic

bnussman-akamai:
## Description 📝

- We were not in sync with what the API actually returns for Volume statuses
- This PR updates the api-v4 types
- Also updates cloud manager to be more dynamic in the way Volume statuses are handled

## Preview 📷

- No visual changes

## How to test 🧪

- `yarn test volume`

- Turn on the **MSW** to see all of the possible volume statuses on `/volumes`

### DIFF
--- a/packages/api-v4/src/volumes/types.ts
+++ b/packages/api-v4/src/volumes/types.ts
@@ -19,10 +19,8 @@ export type VolumeStatus =
   | 'creating'
   | 'active'
   | 'resizing'
-  | 'deleting'
-  | 'deleted'
   | 'migrating'
-  | 'contact_support';
+  | 'offline';
 
 export interface VolumeRequestPayload {
   label: string;
--- a/packages/manager/src/features/Volumes/VolumeTableRow.tsx
+++ b/packages/manager/src/features/Volumes/VolumeTableRow.tsx
@@ -12,7 +12,7 @@ import TableRow from 'src/components/TableRow';
 import { formatRegion } from 'src/utilities';
 import VolumesActionMenu, { ActionHandlers } from './VolumesActionMenu';
 import SupportLink from 'src/components/SupportLink';
-import { Volume } from '@linode/api-v4/lib/volumes/types';
+import { Volume, VolumeStatus } from '@linode/api-v4/lib/volumes/types';
 // import useEvents from 'src/hooks/useEvents';
 
 export const useStyles = makeStyles({
@@ -60,9 +60,7 @@ export const volumeStatusIconMap: Record<Volume['status'], Status> = {
   resizing: 'other',
   migrating: 'other',
   creating: 'other',
-  contact_support: 'error',
-  deleting: 'other',
-  deleted: 'inactive',
+  offline: 'inactive',
 };
 
 export const VolumeTableRow = (props: CombinedProps) => {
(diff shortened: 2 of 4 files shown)


packages/api-v4/src/volumes/types.ts
```diff
@@ -19,10 +19,8 @@ export type VolumeStatus =
   | 'creating'
   | 'active'
   | 'resizing'
-  | 'deleting'
-  | 'deleted'
   | 'migrating'
```
mjac0bs, line 22: Are `status` of `migrating` and `offline` just not documented in the [API docs](https://www.linode.com/docs/api/volumes/#volumes-list__responses)? The only statuses I saw in the enum were `creating`, `active`, and `resizing`, though I may have missed something.

I was curious when a volume would return a `status` of `migrating`; I had tried to migrate a Linode with an attached volume, the message said 
Screenshot 2023-03-09 at 9 30 59 AM

bnussman-akamai: You're right. The API docs are wrong here. I test against my local API to verify what the API may actually return.

The status of `migrating` is  left over from the HDD to NVME migration but is still possible from a technical perspective

mjac0bs: Got it, thanks for clarifying!

packages/manager/src/features/Volumes/VolumeTableRow.tsx
```diff
@@ -97,16 +95,16 @@ export const VolumeTableRow = (props: CombinedProps) => {
...
-    resizing: 'Resizing',
-    creating: 'Creating',
-    contact_support: (
-      <SupportLink text="Contact Support" entity={{ type: 'volume_id', id }} />
-    ),
-    deleting: 'Deleting',
-    deleted: 'Deleted',
-    migrating: 'Migrating',
+  const getVolumeStatus = (status: VolumeStatus) => {
+    if (status === 'offline') {
```
mjac0bs, line 99: I'm having some hesitation with the `StatusIcon`  for `offline` status being gray, while the text is "Contact Support". It somehow seems less clear that/why there's a problem. But I also can't think of a better way to handle this and match up with the statuses the API returns, other than doing something like "Offline - Contact Support" -- any thoughts on whether that would be a worthwhile change in wording?

mjac0bs (approved):
Unit tests looked good and so do the MSW updates. I called out one note about the `StatusIcon`, but it's a minor copy change suggestion and so I'm going to go ahead and approve.

packages/manager/src/features/Volumes/VolumeTableRow.tsx
bnussman-akamai: Yeah, I had all of those same thoughts. The API use to return `contact_support` as a status, but not it returns `offline` in place of `contact_support`. This status exists due to the nvme effort we did, but that is over now. 

Because all volumes have been migrated to nvme, I wonder if we should just remove the "contact support" completely and just show "Offline" if the status is actually offline.

mjac0bs: Given that context, I think showing "Offline" seems like the best thing to show for volumes that are actually offline.

dwiley-akamai (approved):
LGTM 🚢 

Can we give the Docs team a heads-up re: the accuracy of the `status` field enums?

PR #8866, opened 2023-03-09
chore: [M3-6210] - Update Cypress recommended Node version message

jdamore-linode:
## Description 📝

**What does this PR do?**
This updates the Node version warning that shows when you run Cypress. It will now recommend using Node 18 to align with our recent Node version improvements.

Throwing this question out there too: do any of us get any value out of this warning? Do we think it's helpful for other potential contributors?

## Preview 📷
This message will no longer be shown if you're running Node 18.x:

Screen Shot 2023-03-09 at 3 13 08 PM

## How to test 🧪

1. Run `yarn cy:run` (note: the dev server does not need to be running for this)
2. If you're running Node 18, confirm that no warning is shown
3. If you're not running Node 18, confirm that a warning is shown and that Node 18 is recommended

### DIFF
--- a/packages/manager/cypress.config.ts
+++ b/packages/manager/cypress.config.ts
@@ -41,7 +41,7 @@ const loadConfiguration = () => {
  * Displays a warning if tests are running on an unsupported version of Node JS.
  */
 const nodeVersionCheck = () => {
-  const recommendedVersions = [14];
+  const recommendedVersions = [18];
   const versionString = process.version.substr(1, process.version.length - 1);
   const currentVersion = versionString
     .split('.')


bnussman-akamai (approved):
I don't see the warning anymore ✅

I'd be okay with removing or keep this. It seems like node versioning/compatibility isn't a huge problem we run into so I feel like it's not a super valuable warning

jdamore-linode:
> I feel like it's not a super valuable warning

This is basically where I'm at.

I'm trying to picture a scenario where the user is able to spin up Cloud Manager using an unsupported Node version and encounter no issues, and then run Cypress and hit compatibility issues there -- it seems to me that if a user is actually going to run into compatibility issues due to their version of Node, it would be a lot more likely that they'd encounter those problems well before running Cypress.

jaalah-akamai (approved):
🧪 Thanks Joe

PR #8867, opened 2023-03-09
ci: Re-add cache to `build-validation` CI step

bnussman-akamai:
## Description 📝

- Re-adds yarn cache step to `build-validation` step
- For some unknown reason, we removed it https://github.com/linode/manager/pull/7475 but I can't figure out why
- I think it's worth trying this again, if things go bad we can just revert this PR
- This could save us some time in CI

## How to test 🧪

- Verify that the Github Actions ci passes for this PR
- Well just have to observe this change once it's merged

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,11 @@ jobs:
       - uses: actions/setup-node@v1
         with:
           node-version: "18.14"
+      - uses: actions/cache@v3
+        with:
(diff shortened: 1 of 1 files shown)


jdamore-linode (approved):
Nice! Not sure why we got rid of this either, but happy to see the CI workflow still succeeding

hana-linode (approved):

PR #8869, opened 2023-03-10
feat: Allow users to select `system` as a theme option

bnussman-akamai:
## Description 📝

- Allow users to select `system` as a theme option to use their system's theme

## Preview 📷

https://user-images.githubusercontent.com/115251059/224392649-7fed124b-06be-447d-9e7b-2848d97d6830.mov

### Tests
- `yarn test LinodeTheme`

### Manual Testing
- Test theme switching on the `/profile/settings` page
- Verify that the selection persists when the page is reloaded


packages/manager/src/LinodeThemeWrapper.tsx
```diff
@@ -1,19 +1,23 @@
...
 declare module '@mui/styles/defaultTheme' {
   // eslint-disable-next-line @typescript-eslint/no-empty-interface
   interface DefaultTheme extends Theme {}
 }
 
-export type ThemeChoice = 'light' | 'dark';
+export type ThemeChoice = 'light' | 'dark' | 'system';
 
-const themes: Record<ThemeChoice, Theme> = { light, dark };
+const themes: Record<'light' | 'dark', Theme> = { light, dark };
```
hana-linode, line 20: Can we rename `ThemeName` in https://github.com/linode/manager/blob/develop/packages/manager/src/themeFactory.ts#L5 to `type ThemeType = 'light' | 'dark'` and import that type here?

hana-linode (commented):
If I set the theme to system and my system is in Dark mode, when I try to switch the theme with the keyboard shortcut (option + shift + d), it shifts from system to dark instead of system to light 🤔 

https://user-images.githubusercontent.com/115299789/224425960-9ae4f173-26f9-4222-80ff-22d80d4a2df6.mov

jaalah-akamai:
I know you didn't initially write this, but we should really be removing [the listener](https://github.com/linode/manager/blob/a28f4c23512465547732c23d1c2fa0bd52a93027/packages/manager/src/App.tsx#L100) on `componentWillUnmount`

dwiley-akamai (approved):
LGTM 👌

hkhalil-akamai:
+1 since I love websites that have this option. I also wonder whether it might be better to have a radio group (or even a [toggle button](https://mui.com/material-ui/react-toggle-button/)) here instead of a dropdown:

> If there are 3 or fewer items to select, use radio buttons rather than drop-down menus.
> *-Storybook*

bnussman-akamai:
@hkhalil-akamai That's a good point. I went with a Select because that seems to be standard on other websites/apps i've used.

bnussman-akamai:
Updated to use Radios

https://user-images.githubusercontent.com/115251059/224830530-84675f4f-0bb6-4d23-8e9f-3c3afd578f80.mov

PR #8872, opened 2023-03-13
refactor: [M3-5196] - Remove vlans from redux

carrillo-erik:
## Description 📝

As part of the RQ-ify initiative, this PR removes any references of the VLANs data from the redux store. Previous work created and implemented the queries for the VLAN endpoints. Also the `SelectVLAN.tsx` component was removed because there was no references to it anywhere in the codebase. 

## How to test 🧪
1. Create a new Linode in one of the following regions where VLANS are available:
    Mumbai, IN;
    Toronto, ON;
    Sydney, AU;
    Washington, DC;
    Atlanta, GA;
    Newark, NJ;
    London, UK;
    Singapore, SG;
    Frankfurt, DE.
2. In the `Attach a VLAN` section, use the dropdown menu to select a previously created VLAN or create a new one.
3. Complete all other required fields and create the Linode.
4. Once your Linode is running, visit the `Configurations` tab and inspect that the VLAN is listed under the `Network Interfaces`.
5. For further inspection, feel free to add a new configuration or edit the current configuration to attach a different VLAN to the current Linode.


bnussman-akamai (commented):
Nice clean up 🎉

One thing that I just realized is that VLANs probably don't work for users who have > 25 VLANs. By default the API returns the first 25 items in a paginated list and that's what our query is requesting. We probably should should use the `getAll` utility for now to fix this. (You can look at `packages/manager/src/queries/linodes.ts` for an example of how to use `getAll`)

Screenshot 2023-03-13 at 5 45 50 PM

Can we also get rid of the default export in `packages/manager/src/queries/vlans.ts`, I think we should stick to named exports here

hkhalil-akamai (approved):
Approved, pending the additions mentioned by @bnussman-akamai.

jaalah-akamai:
Agreed with the others... by the way `+1 / -276` changes 💯 🎉

packages/manager/src/queries/vlans.ts
```diff
@@ -2,16 +2,15 @@ import { getVlans, VLAN } from '@linode/api-v4/lib/vlans';
...
 import { useQuery } from 'react-query';
 import { queryPresets } from './base';
+import { getAll } from 'src/utilities/getAll';
 
 export const queryKey = 'vlans';
 
-export const _getVlans = (): Promise<VLAN[]> =>
-  getVlans().then(({ data }) => data);
+const _getVlans = (): Promise<VLAN[]> =>
+  getAll<VLAN>(() => getVlans())().then(({ data }) => data);
```
bnussman-akamai, line 10: It is crucial that the params value from the `getAll` utility gets passed to the `getVlans` function. We need to make either of the following changes for this to work

```suggestion
  getAll<VLAN>((params) => getVlans(params))().then(({ data }) => data);
```

Or 

```suggestion
  getAll<VLAN>(getVlans)().then(({ data }) => data);
```

carrillo-erik: @bnussman-akamai I wasn't noticing the difference in the response at first, but adding the `params` is what sets the page size to 200.

bnussman-akamai (approved):

dwiley-akamai (approved):
LGTM 🚢

PR #8874, opened 2023-03-14
fix: [M3-6141] - Use our custom dialog instead of MUI's

jaalah-akamai:
## Description 📝
Close button for dialog was collapsing to new line on mobile / and small breakpoints.

## Changes:
- Use our custom dialog component to fix issue
   - You'll notice the overlay is nicer and button is styled to our typical design 

## Preview 📷
| Before      | After |
| ----------- | ----------- |
| <video src="https://user-images.githubusercontent.com/125309814/224872309-61cad267-d6bb-423f-afa5-d178083aa651.mp4" /> | <video src="https://user-images.githubusercontent.com/125309814/224872326-87516d0a-ea63-43d6-9416-3d346615e2f7.mp4" /> |

## How to test 🧪
- Go to http://localhost:3000/linodes
- Scroll to bottom and click "Monthly Network Transfer Pool" link

Screen Shot 2023-03-13 at 9 57 43 PM


bnussman-akamai (commented):
Looks good, but while we're here I think there is some clean up that could be done

- Some classes like`poolUsageProgress` could probably be deleted
- Remove `React.FC`s

The new dialog is also much wider than the previous. Maybe we should add the `      maxWidth="sm"` prop to make it more similar to what we had before?

jaalah-akamai:
> Looks good, but while we're here I think there is some clean up that could be done
> 
> * Some classes like`poolUsageProgress` could probably be deleted
> * Remove `React.FC`s
> 
> The new dialog is also much wider than the previous. Maybe we should add the ` maxWidth="sm"` prop to make it more similar to what we had before?

Thanks for review, was able to remove a lot of unused classes! Fixed the width of the dialog as well.

packages/manager/src/components/TransferDisplay/TransferDisplay.tsx
bnussman-akamai, line 1: There is still one more `React.FC` in `packages/manager/src/components/TransferDisplay/TransferDisplay.tsx` that could be removed

bnussman-akamai (approved):
Great change. This dialog always sketched me out because it was so different than the others.

cpathipa (approved):
LGTM!

PR #8875, opened 2023-03-14
refactor: [M3-6229] - Remove /core/styles abstraction for tss-react codemod

jaalah-akamai:
## Description 📝
This is strictly a PR that updates imports.

We import styles from `src/components/core/styles`, but the codemod will not accurately be able to transform the file’s because it's expecting to find `@mui` as the path. We have to move away from this layer of abstraction and directly import them back into the files.

### Removed:
- `packages/manager/src/components/core/styles/index.ts`

### Changed:
- All JSS support has been moved to a legacy package `@mui/styles` with new package coming from `@mui/material/styles`.
- `useMediaQuery` now comes from `import useMediaQuery from '@mui/material/useMediaQuery';`
- `Theme`, `ThemeProvider`, and `useTheme` now come from `@mui/material/styles`

## How to test 🧪
- Run `yarn typecheck`
- Run locally, ensure pages still load and nothing looks obviously broken
- There are a few linting errors caused by previous work that we'll tackle in separate PR.


jaalah-akamai:
I appear to be getting type errors in the file `packages/manager/src/mocks/serverHandlers.ts`. This seems to be happening in `develop` as well, which is preventing me from fixing conflicts.

packages/manager/src/components/SideMenu.stories.mdx
```diff
@@ -64,5 +64,5 @@ export const SideMenuSourceCode = `
 <SideMenu
   collapse={collapse}
   open={open}
-  closeMenu={() => setOpen(false)} 
+  closeMenu={() => setOpen(false)}
```
cpathipa, line 67: Looks like its an indentation change by codemod.

cpathipa (approved):
LGTM! I verified that all the changes are related to importing makeStyles and Theme from @mui instead of the wrapper styles component.

PR #8876, opened 2023-03-14
refactor: Infinitely Loaded Entity Select

bnussman-akamai:
## Description 📝

- Introduces a new pattern to infinite load entities in Select components
- Why?
  - For large customers our `useAll` queries are loading tens of thousands of items which will cause Cloud Manager so slow down or crash when rendering large numbers of items in Select components 

Main Points
- Infinity loads items in the Volumes Select component
- Never loads all volumes on the account 
- Allows you to search for volumes using an API filter 
- Uses MUI and not react-select 🥹

> **Note**: I know I did not add any debouncing. We can add that if we think it is important

## Preview 📷

https://user-images.githubusercontent.com/115251059/224890641-fd246800-d238-4f26-a10d-2a9b842d41c5.mov

## How to test 🧪

- Test attaching an existing volume to a Linode form `http://localhost:3000/linodes/:id/storage`

### DIFF
--- a/packages/manager/src/features/Volumes/VolumeDrawer/AttachVolumeToLinodeForm.tsx
+++ b/packages/manager/src/features/Volumes/VolumeDrawer/AttachVolumeToLinodeForm.tsx
@@ -129,7 +129,6 @@ const AttachVolumeToLinodeForm: React.FC<CombinedProps> = (props) => {
 
             <VolumeSelect
               error={touched.volume_id ? errors.volume_id : undefined}
-              name="volumd_id"
               value={values.volume_id}
               onBlur={handleBlur}
               onChange={(v) => setFieldValue('volume_id', v)}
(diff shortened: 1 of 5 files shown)


jaalah-akamai (approved):
This looks good after testing 🎉  This would be a perfect candidate for `IntersectionObserver` which would be even better because we wouldn't have to listen to the scroll event - Maybe I'll create a `useIntersectionObserver` hook for the future.

bnussman-akamai:
Yeah, I actually got triggered by the fact MUI did not add a `onEndReached` prop or something. @jaalah-akamai

hkhalil-akamai (approved):
Tested and LGTM! Pagination will be very important for our larger customers. Looks like there are some more places that need pagination (e.g., the `LinodeVolumes` and `Volumes` components). For another PR.

PR #8877, opened 2023-03-14
chore: Jest configuration is now decoupled from package.json and can be easily customized and maintainable.

cpathipa:
## Description 📝
Decoupled jest configurations from package.json.
This improves maintainability of jest configurations.

## How to test 🧪

- Validate running unit test
- Validate building the app and packages.
- Validate running cypress tests

yarn test
yarn cy:run


packages/manager/package.json
```diff
@@ -212,96 +212,7 @@
     "vite-plugin-svgr": "^2.4.0"
   },
   "jest": {
```
bnussman-akamai, line 214: I think we can completely remove the `jest` section from the `package.json`.

bnussman-akamai (commented):
Nice!

Did using `jest.config.ts` not work? Our jest version might be too old to understand a `.ts` config file

cpathipa:
> Nice!
> 
> Did using `jest.config.ts` not work? Our jest version might be too old to understand a `.ts` config file

I think .ts may need ts-node package to run the tests. May be that's due to older version.

jaalah-akamai:
> I think .ts may need ts-node package to run the tests. May be that's due to older version.

Yea we'll need `ts-node` to run TypeScript files directly. You will also need `ts-jest` for this as well.

```
import type { Config } from '@jest/types';

const config: Config.InitialOptions = {
 // Configs...

PR #8879, opened 2023-03-14
(style): [M3 6187] - Update maintenance and account activation screen logo

mjac0bs:
## Description 📝

**What does this PR do?**

- Updates the logo on the account activation and maintenance screens. 
- Standardizes the size and placement of the logo. 
- Makes a minor copy update on the maintenance screen.

## Preview 📷

## How to test 🧪

1. Check out this PR.
2. Verify the logo and the copy change on the `MaintenanceScreen` by making the condition `true` on L255 of `MainContent.tsx`. 
3. Verify the logo change on the `AccountLandingScreen` by making the condition `true` on L221 of `MainContent.tsx`. 
4. Resize the window and check to make sure the page layout looks good at various breakpoints. 
5. Check to make sure alignment of the `ErrorState` component throughout the app was not adversely impacted by the margin adjustment that was made to fix a misalignment issue on the maintenance and account activation screens.

### DIFF
--- a/packages/manager/src/components/ErrorState/ErrorState.tsx
+++ b/packages/manager/src/components/ErrorState/ErrorState.tsx
@@ -17,6 +17,7 @@ const useStyles = makeStyles((theme: Theme) => ({
   root: {
     width: '100%',
     padding: theme.spacing(10),
+    marginLeft: 0,
   },
   compact: {
     padding: theme.spacing(5),
--- a/packages/manager/src/MainContent.tsx
+++ b/packages/manager/src/MainContent.tsx
@@ -3,7 +3,7 @@ import { isEmpty } from 'ramda';
 import * as React from 'react';
 import { Redirect, Route, RouteComponentProps, Switch } from 'react-router-dom';
 import { compose } from 'recompose';
-import Logo from 'src/assets/logo/logo.svg';
+import Logo from 'src/assets/logo/akamai-logo.svg';
 import Box from 'src/components/core/Box';
 import {
   makeStyles,
--- a/packages/manager/src/MainContent.tsx
+++ b/packages/manager/src/MainContent.tsx
@@ -129,6 +129,13 @@ const useStyles = makeStyles((theme: Theme) => ({
       margin: '0 auto',
     },
   },
+  bgStyling: {
+    backgroundColor: theme.bg.main,
+    minHeight: '100vh',
+    display: 'flex',
+    justifyContent: 'center',
+    flexDirection: 'column',
+  },
 }));
 
 interface Props {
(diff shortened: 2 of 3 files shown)


bnussman-akamai (commented):
Changes look great!

I'm gonna have to push back on the existing red checkmark. Is there anything we can do about it before this is merged? It reminds me of this:

Screenshot 2023-03-14 at 12 53 43 PM

mjac0bs:
> I'm gonna have to push back on the existing red checkmark

@bnussman-akamai Honestly, I agree with you! I'll follow up with UX.

packages/manager/src/components/ErrorState/ErrorState.tsx
```diff
@@ -11,6 +11,7 @@ interface Props {
   cozy?: boolean;
   CustomIcon?: React.ComponentType<SvgIconProps>;
   CustomIconStyles?: React.CSSProperties;
+  className?: string;
```
bnussman-akamai, line 14: Should we call `className` something like `iconClassName` because to me, `className` implies it will be applied to the root component

packages/manager/src/components/ErrorState/ErrorState.tsx
```diff
@@ -53,7 +54,12 @@ const ErrorState = (props: Props) => {
         <div className={classes.iconContainer}>
           {CustomIcon ? (
             <CustomIcon
-              className={classes.icon}
+              className={classNames(
+                {
+                  [classes.icon]: true,
+                },
+                className
+              )}
```
bnussman-akamai, line 62: I think this can be simplified

```suggestion
              className={classNames(classes.icon, className)}
```

packages/manager/src/components/ErrorState/ErrorState.tsx
mjac0bs: Yeah, good call on the naming here. Addressed this and simplified the syntax in the `className` prop assignment in [b97ebda](https://github.com/linode/manager/pull/8879/commits/b97ebda7ca581043496033d6f2cdda1adb70bf7c).

bnussman-akamai (approved):

jdamore-linode (approved):
Nice!

packages/manager/src/components/AccountActivation/AccountActivationLanding.tsx
```diff
@@ -49,16 +52,17 @@ const AccountActivationLanding: React.FC<CombinedProps> = (props) => {
 
   return (
     <ErrorState
+      iconClassName={classes.icon}
```
hana-linode, line 55: It looks like we can use the `CustomIconStyles` prop instead of defining a new prop
```suggestion
      CustomIconStyles={{ color: '#63A701' }}
```

hana-linode (approved):

packages/manager/src/components/AccountActivation/AccountActivationLanding.tsx
mjac0bs: 🥇 Thanks for catching this. Made that change.

PR #8882, opened 2023-03-14
refactor: Volumes Query Key Factory 🏭

bnussman-akamai:
## Description 📝

- A proof of concept for a "Query Key factory"
  - The idea behind this is to create a single source of truth for query keys
  - Currently we have no "safety" ensuring you invalidate or use the correct query key
  - Hopefully we can use this idea of a "query key factory" to make our query keys more consistent across the app as we scale React Query
- The easy part of this is creating the queryKey, the hard part is finding a good way to invalidate data using the query key "factory"
- Looking to hear options from the team (cc @hkhalil-akamai and @jaalah-akamai )
- Inspired by https://twitter.com/TkDodo/status/1448232665686106119 and https://github.com/lukemorales/query-key-factory

> **Warning**: This is just a proof of concept, I hope we can find ways to improve upon this
  
## How to test 🧪

- Test Volume features of Cloud Manager

### DIFF
--- a/packages/manager/src/queries/volumes.ts
+++ b/packages/manager/src/queries/volumes.ts
@@ -160,7 +160,7 @@ export const volumeEventsHandler = (event: Event) => {
   if (event.action === 'volume_clone') {
     setTimeout(() => {
       queryClient.invalidateQueries(queryKey);
-    }, 10000);
+    }, 8000);
   }
 };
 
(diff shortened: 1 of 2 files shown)


jaalah-akamai:
After thinking about this a little more, I think we should just pull in the [Query Key Factory](https://tanstack.com/query/v4/docs/react/community/lukemorales-query-key-factory) package you mentioned above. It has zero dependencies, an API that is really nice, and it's suggested as the default solution by TanStack.

PR #8884, opened 2023-03-15
fix: [M3-6204] - Add Cypress clean-up for service transfer requests

cliu-akamai:
## Description 📝

- Added `cancelAllTestEntityTransfers()` function to `deleteAllTestData()`.
- Changed `deleteAllTestData()` to run after each test.

## How to test 🧪

N/A

Run command: yarn cy:run -s "cypress/e2e/account/service-transfer.spec.ts"

### DIFF
--- a/packages/manager/cypress/e2e/account/service-transfer.spec.ts
+++ b/packages/manager/cypress/e2e/account/service-transfer.spec.ts
@@ -23,6 +23,7 @@ import { ui } from 'support/ui';
 import { pollLinodeStatus } from 'support/util/polling';
 import { randomItem, randomLabel, randomUuid } from 'support/util/random';
 import { visitUrlWithManagedEnabled } from 'support/api/managed';
+import { cancelAllTestEntityTransfers } from 'support/api/entityTransfer';
 
 // Service transfer landing page URL.
 const serviceTransferLandingUrl = '/account/service-transfers';
--- a/packages/manager/cypress/e2e/account/service-transfer.spec.ts
+++ b/packages/manager/cypress/e2e/account/service-transfer.spec.ts
@@ -23,7 +23,6 @@ import { ui } from 'support/ui';
 import { pollLinodeStatus } from 'support/util/polling';
 import { randomItem, randomLabel, randomUuid } from 'support/util/random';
 import { visitUrlWithManagedEnabled } from 'support/api/managed';
-import { cancelAllTestEntityTransfers } from 'support/api/entityTransfer';
 
 // Service transfer landing page URL.
 const serviceTransferLandingUrl = '/account/service-transfers';
--- a/packages/manager/cypress/support/api/entityTransfer.ts
+++ b/packages/manager/cypress/support/api/entityTransfer.ts
@@ -1,0 +1,66 @@
+import { EntityTransfer } from '@linode/api-v4/types';
+import {
+  getEntityTransfers,
+  cancelTransfer,
+  getLinode,
+} from '@linode/api-v4/lib';
+import { depaginate } from 'support/util/paginate';
+import { isTestLabel } from './common';
+
+/**
+ * Cancel all entity transfers whose labels are prefixed "cy-test-".
+ *
+ * @returns Promise that resolves when entity transfers have been cancelled or rejects on HTTP error.
+ */
+export const cancelAllTestEntityTransfers = async (): Promise<void> => {
+  const entityTransfers = await depaginate<EntityTransfer>((page: number) =>
+    getEntityTransfers({ page_size: 500, page })
+  );
+  const testEntityTransfers: EntityTransfer[] = [];
+  for (let i = 0; i < entityTransfers.length; i++) {
+    // eslint-disable-next-line no-await-in-loop
+    const isTest = await areAllLinodesWithTestLabel(
+      entityTransfers[i]?.entities?.linodes
+    );
+    if (isTest) {
+      testEntityTransfers.push(entityTransfers[i]);
+    }
+  }
+  for (const testEntityTransfer of testEntityTransfers) {
+    // We want to send these requests sequentially
+    // to avoid overloading the API.
+    // eslint-disable-next-line no-await-in-loop
+    await cancelTransfer(testEntityTransfer.token);
+  }
+};
+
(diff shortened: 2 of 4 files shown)


packages/manager/cypress/support/e2e.js
```diff
@@ -32,3 +33,8 @@ import './request-tracking';
 before(() => {
   deleteAllTestData();
 });
+
+// Runs after each test file.
+after(() => {
+  cancelAllTestEntityTransfers();
```
jdamore-linode, line 39: Do you think we could move the call to `cancelAllTestEntityTransfers()` so that it's invoked from [inside `deleteAllTestData()`](https://github.com/linode/manager/blob/develop/packages/manager/cypress/support/ui/common.ts#L73)?

It seems counterintuitive, but doing our clean-up before tests run instead of after offers a few advantages: it can be more reliable, we can examine leftover resources after tests finish in case we need to do troubleshooting, etc. [Cypress's best practices guide](https://docs.cypress.io/guides/references/best-practices#Using-after-or-afterEach-hooks) does a great job explaining the advantages, and it's a great resource in general.

packages/manager/cypress/e2e/account/service-transfer.spec.ts
```diff
@@ -23,6 +23,7 @@ import { ui } from 'support/ui';
 import { pollLinodeStatus } from 'support/util/polling';
 import { randomItem, randomLabel, randomUuid } from 'support/util/random';
 import { visitUrlWithManagedEnabled } from 'support/api/managed';
+import { cancelAllTestEntityTransfers } from 'support/api/entityTransfer';
```
jdamore-linode, line 26:
```suggestion
```

Looks like this may have been left over! I think we can also remove the import in `e2e.js` once the call to `cancelAllTestEntityTransfers()` gets moved

packages/manager/cypress/support/api/entityTransfer.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Cancel all entity transfers whose labels are prefixed "cy-test-".
+ *
+ * @returns Promise that resolves when entity transfers have been cancelled or rejects on HTTP error.
+ */
+export const cancelAllTestEntityTransfers = async (): Promise<void> => {
+  const entityTransfers = await depaginate<EntityTransfer>((page: number) =>
+    getEntityTransfers({ page_size: 500, page })
+  );
```
jdamore-linode, line 18: Potential improvement:

```suggestion
  const entityTransfers = (await depaginate<EntityTransfer>((page: number) =>
      getEntityTransfers({ page_size: 500, page })
    ))
    .filter((entityTransfer: EntityTransfer) => entityTransfer.status === 'pending');
```

We can `filter` out any entity transfers that aren't pending here, which will help limit the number of GET requests we make to fetch Linodes later on (and I think it will eliminate requests that result in 404 responses).

packages/manager/cypress/support/api/entityTransfer.ts
```diff
@@ -0,0 +1,66 @@
...
+  const testEntityTransfers: EntityTransfer[] = [];
+  for (let i = 0; i < entityTransfers.length; i++) {
+    // eslint-disable-next-line no-await-in-loop
+    const isTest = await areAllLinodesWithTestLabel(
+      entityTransfers[i]?.entities?.linodes
+    );
+    if (isTest) {
+      testEntityTransfers.push(entityTransfers[i]);
```
jdamore-linode, line 26: Would it be possible to call `await cancelTransfer(entityTransfers[i].token)` here instead? If that works, I think we could get rid of the `testEntityTransfers` array and the `for... of` loop below!

packages/manager/cypress/support/api/entityTransfer.ts
```diff
@@ -0,0 +1,66 @@
...
+    // eslint-disable-next-line no-await-in-loop
+    await cancelTransfer(testEntityTransfer.token);
+  }
+};
+
+const areAllLinodesWithTestLabel = async (
+  linodeIds: number[]
+): Promise<boolean> => {
+  if (linodeIds?.length > 0) {
+    const filteredLinodeIds = await linodeFilter(linodeIds);
```
jdamore-linode, line 41: I think we can simplify this logic a bit using [`Array.every()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every).

Something like this might work without the need to use `linodeFilter()`:

```typescript
const linodes = await Promise.all(linodeIds.map((id: number) => getLinode(id)));
return linodes.every((linode: Linode) => isTestLabel(linode.label));
```

jdamore-linode (commented):
Nice work @cliu-akamai! Confirmed that your `cancelAllTestEntityTransfers` function succeeds at deleting service transfers for test resources without deleting other service transfers, and posted a few questions/suggestions.

packages/manager/cypress/e2e/account/service-transfer.spec.ts
cliu-akamai: Good catch! I will remove that.

packages/manager/cypress/support/api/entityTransfer.ts
cliu-akamai: Sounds good! Let me try that. Thank you.

packages/manager/cypress/support/api/entityTransfer.ts
```diff
@@ -0,0 +1,49 @@
...
+    (entityTransfer: EntityTransfer) => entityTransfer.status === 'pending'
+  );
+
+  for (const entityTransfer of entityTransfers) {
+    // eslint-disable-next-line no-await-in-loop
+    const isTest = await areAllLinodesWithTestLabel(
+      entityTransfer?.entities?.linodes
+    );
+    if (isTest) {
+      console.log(`entityTransfer - details: ${entityTransfer}`);
```
jdamore-linode, line 30:
```suggestion
```
I'm not sure we need this `console.log` here, and we might want to remove it out of an abundance of caution in case it causes actual tokens to get logged

jdamore-linode (approved):
Looks good, great work @cliu-akamai !

PR #8885, opened 2023-03-15
refactor: [M3-6256] - Provide a Typescript type for `api-v4` params and filters

bnussman-akamai:
## Description 📝

- Adds a typescript type for api-v4's params and filters
- Removes Redux for Firewalls because firewalls are in React Query
- Removes `getAllWithArguments` because it was confusing

## How to test 🧪

- Verify typecheck passes
- Look at the type definition for `Params` and `Filter` and let me know if you have any improvements that could be made


packages/api-v4/src/request.ts
```diff
@@ -64,7 +64,7 @@ export const setMethod = (method: 'GET' | 'POST' | 'PUT' | 'DELETE') =>
   set('method', method);
 
 /** Param */
-export const setParams = (params: any = {}) => set('params', params);
+export const setParams = (params: Params | undefined) => set('params', params);
```
hkhalil-akamai, line 67:
```suggestion
export const setParams = (params?: Params) => set('params', params);
```

packages/api-v4/src/request.ts
```diff
@@ -141,7 +141,7 @@ const mapYupToLinodeAPIError = ({
 });
 
 /** X-Filter */
-export const setXFilter = (xFilter: any) => {
+export const setXFilter = (xFilter: Filter | undefined) => {
```
hkhalil-akamai, line 144:
```suggestion
export const setXFilter = (xFilter?: Filter) => {
```

packages/api-v4/src/types.ts
```diff
@@ -25,3 +25,66 @@ export interface ResourcePage<T> {
...
+  '+neq'?: string;
+}
+
+export type Filter = LinodeFilter | LinodeFilter[];
+
+type LinodeFilter =
+  | {
+      [key in keyof FilterConditionTypes]: FilterConditionTypes[key];
+    }
+  | { [key: string]: string | number | boolean | Filter | undefined };
```
hkhalil-akamai, line 53: Is the more generic filter type used somewhere? Because it's so general, it makes the `FilterConditionTypes` definition redundant.

packages/api-v4/src/types.ts
```diff
@@ -25,3 +25,66 @@ export interface ResourcePage<T> {
...
+//         },
+//         {
+//           vcpus: {
+//             '+lte': 20,
+//           },
+//         },
+//       ],
+//     },
+//   ],
+// };
```
hkhalil-akamai, line 90: I'm assuming the comments are examples?

bnussman-akamai: Yeah, I used those examples to verify the type is correct. I figured it wouldn't hurt to leave them

packages/api-v4/src/types.ts
bnussman-akamai: I'm not completely sure I understand what you mean.

Are you saying because `string` is a valid key that `FilterConditionTypes` is essentially useless?

That is true, but I added `FilterConditionTypes` so that you get nice auto-completion when building filters even though any `string` is technically valid

Screenshot 2023-03-20 at 1 44 51 PM

carrillo-erik:
Nice clean up of the api-v4 package. With the introduction of the `FilterConditionTypes` interface, the code resembles the Linode API docs much better.

carrillo-erik (approved):

PR #8889, opened 2023-03-16
refactor: [M3-6257] - React Query for Firewalls

bnussman-akamai:
## Description 📝

- Firewalls used react query but this moves firewalls away from the "fetch all" approach

## How to test 🧪

- Test all firewall create, update, delete actions
- Test adding and removing Linodes from a Firewall
- yarn test Firewall
- Run `migrate-linode-with-firewall.spec.ts` with `yarn cy:debug`


packages/manager/cypress/e2e/firewalls/migrate-linode-with-firewall.spec.ts
bnussman-akamai, line 1: @jdamore-linode  I could use some guidance here. I rewrote `<LinodeMultiSelect />` to use an infinite query with an API filter. 

I need the linode label to be typed, then **wait** for (multiple) API calls to `linode/instances*` as we "search" the API, and selected the first option when resultus have loaded.

jdamore-linode: @bnussman-akamai I think the most reliable way to handle these interactions would be to split them each into two chains: one that selects the text field and types the Linode label, and another that selects the autocomplete drop-down popper then finds and clicks the expected Linode label inside of it; that will let you sidestep waiting for any API requests since you can just wait for the expected text to become visible instead.

I'm not seeing any good way to select the popper element, so if it doesn't interfere with your work I'd suggest adding a `[data-qa-*]` attribute to the popper using the `Autocomplete` component's `PopperComponent` prop:

```tsx
// LinodeMultiSelect.tsx
import Popper from '@mui/material/Popper';

// ...

return (
  <Autocomplete
    multiple
    clearOnBlur
    // ...
    PopperComponent={(popperProps) => <Popper {...popperProps} data-qa-autocomplete-popper />}
    // ...
  />
```
(I didn't test this too thoroughly in case there are any unintended side effects, I just confirmed that the attribute shows up)

Then you can interact with the text field from Cypress like this:

```typescript
cy.get('[data-testid="textfield-input"]:last')
  .should('be.visible')
  .click()
  .type(linode.label);

cy.get('[data-qa-autocomplete-popper]')
  .findByText(linode.label)
  .should('be.visible')
  .click();
```

packages/manager/src/components/LinodeMultiSelect/LinodeMultiSelect.tsx
```diff
@@ -1,164 +1,119 @@
...
+
+  const options = data?.pages
+    .flatMap((page) => page.data)
+    .map(({ id, label }) => ({ id, label }));
+
+  const selectedLinodeOptions =
+    options?.filter((option) => value.includes(option.id)) ?? [];
 
   return (
-    <Select
```
jdamore-linode, line 89: Are we gradually replacing all of our multi selects with `<TextField />`s wrapped in `<Autocomplete />`s, or is there a specific technical need/UX convention/etc that dictates when to use one over the other?

jdamore-linode:
Related to the question above, I'm noticing some visual differences between the old multi selects and the new auto complete text fields. The screenshots are from the Attach Volume form, but also applies to the LinodeMultiSelect:

**Old:**
Screen Shot 2023-03-17 at 11 55 54 AM

**New:**
Screen Shot 2023-03-17 at 11 55 47 AM

Specifically seeing differences in the placeholder text color, arrow/chevron icon on the right, size and positioning of the popper, and the selection highlight style.

packages/manager/src/components/LinodeMultiSelect/LinodeMultiSelect.tsx
bnussman-akamai: I moved to a `<Autocomplete />` simply because it is built into MUI and is much more easy to use then our crazily abstracted Select component that has no type safety.

jaalah-akamai:
If we want to use the same icon as before, we can do `popupIcon={<DropdownIndicator />}` except we'll either have to add an `sx` prop to remove the `marginRight` or some custom class for now. The other styles should be easy to replicate 🤞

jaalah-akamai (approved):
Approving pending style updates - Solid changes!

cpathipa (approved):
Nice work! @bnussman. Overall, it looks great! I only left a comment for a minor change to improve readability.

PR #8890, opened 2023-03-17
feat: [M3-5746] - Disable Download CA Certificate when DB is provisioning

dwiley-akamai:
## Description 📝
Disable the "Download CA Certificate" button while the database is be provisioned.

## Preview 📷
Screenshot 2023-03-17 at 10 35 53 AM

## How to test 🧪
You can either spin up a new DB instance or utilize the MSW. I believe right now we have it so that if you go to a Database Details page with the MSW on, it randomizes the database status and view according to the status. You can either refresh until you get `Provisioning` (shouldn't be more than 2-3 refreshes) or make a change in `serverHandlers.ts`.

Confirm that DBs in statuses other than provisioning have not been impacted.

### DIFF
--- a/packages/manager/src/features/Databases/DatabaseDetail/DatabaseSummary/DatabaseSummaryConnectionDetails.tsx
+++ b/packages/manager/src/features/Databases/DatabaseDetail/DatabaseSummary/DatabaseSummaryConnectionDetails.tsx
@@ -45,33 +45,6 @@ const useStyles = makeStyles((theme: Theme) => ({
     justifyContent: 'flex-end',
     padding: `${theme.spacing(1)} 0`,
   },
-  actionBtn: {
-    cursor: 'pointer',
-    display: 'flex',
-    alignItems: 'center',
-    borderRight: '1px solid #c4c4c4',
-    '&:hover': {
-      opacity: 0.7,
-    },
-    '&:last-child': {
-      borderRight: 'none',
-    },
-    marginLeft: theme.spacing(2),
-  },
-  actionBtnDisabled: {
-    display: 'flex',
-    '& g': {
-      stroke: theme.color.disabledText,
-    },
-  },
-  actionText: {
-    color: theme.textColors.linkActiveLight,
-    whiteSpace: 'nowrap',
-  },
-  actionTextDisabled: {
-    color: theme.color.disabledText,
-    whiteSpace: 'nowrap',
-  },
   connectionDetailsCtn: {
     padding: '8px 15px',
     background: theme.bg.bgAccessRow,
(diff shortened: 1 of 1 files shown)


packages/manager/src/features/Databases/DatabaseDetail/DatabaseSummary/DatabaseSummaryConnectionDetails.tsx
```diff
@@ -180,6 +191,53 @@ export const DatabaseSummaryConnectionDetails: React.FC<Props> = (props) => {
...
+      <Grid
+        item
+        onClick={() => {
+          if (disableDownloadCACertificateBtn) {
+            return;
+          }
+
+          handleDownloadCACertificate();
+        }}
+        role="button"
```
dwiley-akamai, line 205: An argument can be made that we should look into refactoring this; I'm willing to look into doing so if others think it's worth doing right now.

bnussman-akamai (approved):
Button is disabled while the Database provisions ✅

The behavior of that "button" seems a bit weird. For example, when I hover over it while it is disabled, I see the text selection cursor. I'd expect to see a normal cursor. From my experience most UI frameworks allow you to pass an icon prop to a `<Button />` component. It really seems like we should be using a button here

mjac0bs (commented):
One thing to note is that in dark mode, our makeshift "button" is a little misleading because the light grey disabled text is a similar shade to the rest of the white text on the page. It's difficult to tell that the button is in disabled state. 

I'd be for the refactor here that @bnussman-akamai mentioned to use an actual button.
Screenshot 2023-03-20 at 11 28 45 AM

packages/manager/src/features/Databases/DatabaseDetail/DatabaseSummary/DatabaseSummaryConnectionDetails.tsx
```diff
@@ -346,25 +404,7 @@ export const DatabaseSummaryConnectionDetails: React.FC<Props> = (props) => {
         </Typography>
       </Grid>
       <div className={classes.actionBtnsCtn}>
```
hana-linode, line 406: Can we center align the items in this container? The help icon looks a little off center

image

packages/manager/src/features/Databases/DatabaseDetail/DatabaseSummary/DatabaseSummaryConnectionDetails.tsx
hana-linode: I think it's worth looking into

packages/manager/src/features/Databases/DatabaseDetail/DatabaseSummary/DatabaseSummaryConnectionDetails.tsx
dwiley-akamai: looks centered with the refactor
Screenshot 2023-03-21 at 4 29 44 PM

packages/manager/src/features/Databases/DatabaseDetail/DatabaseSummary/DatabaseSummaryConnectionDetails.tsx
```diff
@@ -97,6 +80,37 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+    marginLeft: theme.spacing(),
+    fontFamily: theme.font.normal,
+    fontWeight: theme.typography.fontWeightRegular,
+    fontSize: '0.875rem',
+    lineHeight: '1.125rem',
+    minHeight: 'auto',
+    minWidth: 'auto',
+    padding: 0,
+    '&[disabled]': {
+      // Override disabled background color defined for dark mode
```
jaalah-akamai, line 101: Wish we didn't have to do this, but this will be something I will make sure Design accounts for when we review the UI this week. 👍

jaalah-akamai (approved):

PR #8891, opened 2023-03-17
feat: [M3-5258] - Add delete warning to LKE

carrillo-erik:
## Description 📝

The new warning dialog informs users that deleting an LKE cluster does not delete any Block Volume Storage or NodeBalancer that is attached to their cluster. Therefore, the user will have to delete them separately to remove them from their account.

## Preview 📷
**Before:**
old

**After:**
new

## How to test 🧪
1. Create a new LKE cluster.
2. Attach a Block Volume Storage or NodeBalancer instance.
3. Try to delete the LKE cluster and confirm the new warning message is shown to the user.

### DIFF
--- a/packages/manager/src/features/Kubernetes/KubernetesClusterDetail/DeleteKubernetesClusterDialog.tsx
+++ b/packages/manager/src/features/Kubernetes/KubernetesClusterDetail/DeleteKubernetesClusterDialog.tsx
@@ -30,7 +30,7 @@ export const getTotalLinodes = (pools: KubeNodePoolResponse[]) => {
 export const DeleteKubernetesClusterDialog = (props: Props) => {
   const { clusterLabel, clusterId, open, onClose } = props;
 
-  const { data: pools, isLoading } = useAllKubernetesNodePoolQuery(clusterId, {
+  const { data: isLoading } = useAllKubernetesNodePoolQuery(clusterId, {
     enabled: open,
   });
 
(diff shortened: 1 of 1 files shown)


packages/manager/src/features/Kubernetes/KubernetesClusterDetail/DeleteKubernetesClusterDialog.tsx
```diff
@@ -30,7 +30,7 @@ export const getTotalLinodes = (pools: KubeNodePoolResponse[]) => {
 export const DeleteKubernetesClusterDialog = (props: Props) => {
   const { clusterLabel, clusterId, open, onClose } = props;
 
-  const { data: pools, isLoading } = useAllKubernetesNodePoolQuery(clusterId, {
+  const { data: isLoading } = useAllKubernetesNodePoolQuery(clusterId, {
```
mjac0bs, line 33:
```suggestion
  const { isLoading } = useAllKubernetesNodePoolQuery(clusterId, {
```

mjac0bs: Looks like we're getting an infinitely loading delete modal (also why the unit tests failed) because we're going to need to delete the `data` param from the query hook now that we've removed `pools`.

mjac0bs (changes requested):
Let's remove the `useKubernetesNotePoolQuery` entirely, and then update the `ConfirmationDialog` since we'll no longer need the conditional render of the `CircleProgress` based on whether or not the pool query `isLoading`. I think there will also be unit test updates to make.

carrillo-erik:
I took care of removing the tests which are no longer required due the removal of the pools query. Also, removed the `CircleProgress` component as well.

mjac0bs (approved):
Unit tests are passing and the confirmation dialog renders with our updated notice, so ✅ .

One thing we might think about adding to the unit test is check that a warning notice exists within the kubernetes deletion dialog. Although we weren't checking for the notice explicitly before, we were implicitly checking when we tested for specific text we'd find within the notice.

bnussman (approved):
Nice. The code is much cleaner now that we don't need to worry about loading any API data

PR #8892, opened 2023-03-17
refactor: [M3-6258] - React Query for SSH Keys

bnussman-akamai:
## Description 📝

- Uses React Query for SSH Keys
- Wow, this was a lot more difficult than expected. SSH keys are used on Linode creation flows

## How to test 🧪

- Test all actions possible on `http://localhost:3000/profile/keys`
- Test the following with at least one SSH Key on your account and please perform these actions as **both* a normal user and as a restricted user
  - Test Creating a Linode with an SSH Key
  - Test rebuilding a Linode from an Image with an SSH Key
  - Test rebuilding a Linode from a Stackscript with an SSH Key


packages/manager/src/queries/profile.ts
```diff
@@ -77,3 +84,55 @@ export const useVerifyPhoneVerificationCodeMutation = () =>
   useMutation<{}, APIError[], VerifyVerificationCodePayload>(
     verifyPhoneNumberCode
   );
+
+export const useSSHKeysQuery = (params?: any, filter?: any, enabled = true) =>
```
dwiley-akamai, line 88: I'm sure it was on your radar already, but since your #8885 PR got merged, we can pull in the latest `develop` on this branch and give these the proper types

dwiley-akamai (commented):
Going to take a deeper look later today or sometime tomorrow but left a couple of small comments

packages/manager/src/utilities/stringUtils.test.ts
```diff
@@ -26,4 +26,14 @@ describe('truncateAndJoinList', () => {
     const result = truncateAndJoinList(bigStrList);
     expect(result).toMatch(/, plus 900 more/);
   });
+
+  it('supports overriding the total ammount', () => {
```
dwiley-akamai, line 30: "amount"

dwiley-akamai (commented):
- I tried rebuilding with an Image & SSH key and with a StackScript & SSH key but ran into a `Service unavailable [1.10]` API error earlier, so I will circle back around to testing that.

- If you edit an SSH key, clear out the label, and click "Save", the error persists if you close the drawer and go back to edit it. It also persists across different SSH keys (i.e., if you get up to the error on one, then you exit that drawer and go to edit another key, the error will show up there too)

cpathipa (commented):
Not able to run the app locally.
image

bnussman-akamai:
@cpathipa You may need to run `rm -rf node_modules/.vite` from the repo's root to clear the vite cache. The app runs fine for me.

cpathipa (approved):

cpathipa:
@bnussman-akamai FYI - A few files have conflicts that should be resolved.

dwiley-akamai (commented):
The edit --> empty label interaction is still a bit off from what I'd expect; in this clip, I clear the label, try to save, get the error, and close the drawer. But when I re-open the drawer, although the error is cleared, the label field itself is still empty (instead of having the current label):

<details>
<summary>Edit "Label" field interaction</summary>

https://user-images.githubusercontent.com/114682940/227618746-a7750421-8d4a-4d1b-8255-83e7e12ae4d5.mov
</details>

Aside from that, looks like something odd is going on for restricted users in the "SSH Keys" table in the Linode Create and Rebuild flows?

Screenshot 2023-03-22 at 12 26 32 PM

hkhalil-akamai:
Only seeing one minor regression...
Screen Shot 2023-03-24 at 3 43 00 PM

hkhalil-akamai (approved):
Really great clean-up/refactoring; unit tests are a plus

bnussman-akamai:
Very helpful feedback, thank you @dwiley-akamai. Should all be addressed in 5d25bf3

dwiley-akamai (approved):
Looks good, thanks for addressing that feedback! 🚀

PR #8893, opened 2023-03-18
refactor: [M3-6241] - MUI v5 Migration - Features > Account

jaalah-akamai:
## Description 📝
Migrates `SRC > Features > Account` from JSS to tss-react (Emotion)

## Changes
- Bumped `tss-react` version to get latest fixes from author 🎉 
   - https://github.com/garronej/tss-react/issues/160 
   - New updated documentation for class components: https://docs.tss-react.dev/api/withstyles#with-class-components 
- Removed `@mui/styles` imports
- Removed `React.FC`
- Removed `any` and added better typechecking to `Account/Agreements/withAgreements.ts`
- Removed deprecated `createStyles` and unnecessary `withStyles`
- Other minor cleanup

## How to test 🧪
- Go to `/account/` locally and ensure all functionality still works

### DIFF
--- a/packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx
+++ b/packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx
@@ -15,7 +15,7 @@ import TableRowEmptyState from 'src/components/TableRowEmptyState';
 import Typography from 'src/components/core/Typography';
 import { AccountMaintenance } from '@linode/api-v4/lib/account/types';
 import { CSVLink } from 'react-csv';
-import { makeStyles } from '@mui/styles';
+import { makeStyles } from 'tss-react/mui';
 import { Theme } from '@mui/material/styles';
 import { cleanCSVData } from 'src/components/DownloadCSV/DownloadCSV';
 import { useOrder } from 'src/hooks/useOrder';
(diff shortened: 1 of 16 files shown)


packages/manager/src/features/Account/NetworkHelper.tsx
```diff
@@ -1,76 +1,65 @@
...
-      textDecoration: 'underline',
-    },
-    icon: {
-      display: 'inline-block',
-      fontSize: '0.8em',
-      marginLeft: `calc(${theme.spacing(1)} / 3)`,
-    },
-  });
+const useStyles = makeStyles()((theme: Theme) => ({
+  root: {},
```
bnussman-akamai, line 11: We could remove empty classes when we see them

packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx
```diff
@@ -238,3 +238,5 @@ export const MaintenanceTable = ({ type }: Props) => {
     </>
   );
 };
+
+export { MaintenanceTable };
```
dwiley-akamai, line 242: Any particular reason to go with `export { MaintenanceTable }` instead of `export default MaintenanceTable` or staying with `export const MaintenanceTable`?

jaalah-akamai: For the most part I left things as they were (default export) and in this case it's essentially the same thing; we're exporting a named function. In my opinion with our use of typescript, migrations, and future architecture of where we want to be, named functions provide much more benefits that we should all talk about. I strongly suggest we move towards that paradigm.

PR #8897, opened 2023-03-20
perf: Decrease bundle size for largest assets > 100kb

jaalah-akamai:
## Description 📝
I'm going to be making incremental improvements to our bundle size starting with `lodash`.

## Major Changes
- We were importing the whole Lodash library when all we needed were a few methods

## Preview 📷
| Before      | After |
| ----------- | ----------- |
|Screen Shot 2023-03-20 at 5 26 22 PM|Screen Shot 2023-03-20 at 5 32 15 PM|

## Testing
- Run `npx vite-bundle-visualizer`


packages/manager/src/components/PasswordInput/PasswordInput.tsx
bnussman-akamai, line 1: I'm not sure how effective this change is in practice because everywhere `<PasswordInput />` is used, it is already lazy loaded

bnussman-akamai (approved):
Nice. One thing about `lodash` is that we don't actually specify it in manager's `package.json` but we really should because we import it. I think the reason we haven't seen issues is because some other packages probably depend on it, therefore it is installed without us directly specify it. This problem can be addressed in this PR or we can worry about it another time.

dwiley-akamai (approved):
I see the difference between this branch and `develop` when running `npx vite-bundle-visualizer` ✅ 

In case anyone else is having trouble trying to run that command, I was able to get it to run by making sure I was on Node 18.14.1 (`nvm use 18.14.1`) and in the `manager/packages/manager` directory

jaalah-akamai:
> Nice. One thing about `lodash` is that we don't actually specify it in manager's `package.json` but we really should because we import it. I think the reason we haven't seen issues is because some other packages probably depend on it, therefore it is installed without us directly specify it. This problem can be addressed in this PR or we can worry about it another time.

Yea that's true, I'll do it in this PR 👍

PR #8901, opened 2023-03-21
test: [M3-6234] - Add Cypress LKE creation end-to-end tests

jdamore-linode:
## Description 📝

**What does this PR do?**

- Adds e2e test for LKE cluster creation flow
- Cleans up the existing LKE smoke test
- Adds a `title` attribute to the LKE node pool removal button

## How to test 🧪

Unit tests:
```bash
yarn test
```

Cypress tests:
```bash
yarn cy:run -s 
```


jaalah-akamai:
Here is my output I'm getting: https://app.warp.dev/block/pLeKMWStXvLDPAi9QQDLQb after running
```
yarn cy:run -s cypress/e2e/kubernetes/lke-create.spec.ts
```

jdamore-linode:
> Here is my output I'm getting: https://app.warp.dev/block/pLeKMWStXvLDPAi9QQDLQb after running
> 
> ```
> yarn cy:run -s cypress/e2e/kubernetes/lke-create.spec.ts
> ```
> 
> Anything additional I should be running, or is this accurate?

That sounds like our banner's obscuring one of the elements being selected by the tests. If you run the test via `yarn cy:debug`, you should be able to dismiss the banner and the tests should start passing again.

bnussman-akamai (approved):
The tests ran well!

It did seem like the test spun up some **massive** clusters. Do we need to worry about keeping the size of these clusters down?

jaalah-akamai (approved):

jdamore-linode:
Pushed a change so that only LKE clusters that have finished provisioning (i.e. all nodes have `ready` status) get deleted during clean up. This could lead to the accumulation of resources (especially if an LKE gets stuck in a non-ready state, for example) but I don't see it causing us any real trouble.

PR #8903, opened 2023-03-22
refactor: [M3-6252] - Update `react-query@3` to `@tanstack/react-query@4`

bnussman-akamai:
## Description 📝

- Updates our React Query version to v4
- The main change that affects us is that query keys must be arrays. They can no longer be strings
- Because so many query keys had to be migrated, I standardized our query key pattern a bit more... The general format we should follow is shown in the example below

| API Call                         | useQuery Hook                         | Query Key                                           |
|----------------------------------|---------------------------------------|-----------------------------------------------------|
| /v4/linode/instances             | useLinodesQuery(params, filter)       | ['linodes', 'paginated', params, filter]            |
| /v4/linode/instances             | useAllLinodesQuery()                  | ['linodes', 'all']                                  |
| /v4/linode/instances/:id         | useLinodeQuery(id)                    | ['linodes', 'linode', id]                           |
| /v4/linode/instances/:id/backups | useLinodeBackupsQuery(params, filter) | ['linodes', 'linode', id, 'backups', params, filter |

## How to test 🧪

- This touches so much so we're going to have to just manually test all of the app
- We should closely examine any e2e failures


bnussman-akamai:
Classic unit tests...

PR #8906, opened 2023-03-22
feat: [M3-6072] - New tab for plans table

cpathipa:
## Description 📝
Introduces new plan in Plans table

## How to test 🧪
Validate Plan tabs as per the Jira ticket.

- yarn test
- yarn cy:run


mjac0bs (approved):
Confirmed that the new plan type is visible after following instructions on our internal ticket. Did we determine that this work doesn't need to be feat branched and instead this PR can point to develop?

PR #8908, opened 2023-03-23
refactor: [M3-6327, M3-6328] - MUI v5 Migration - Components > Drawer, DrawerContent 

mjac0bs:
## Description 📝

**What does this PR do?**

- Runs the codemod on `Drawer.tsx`.
- Runs the codemod on `DrawerContent.tsx` but no styling updates here -- this component didn't have styles to begin with!
- Removes `React.FC` from `Drawer` and `DrawerContent`.
   - Updates the unit test for `DrawerContent` due to the need to explicitly declare the `children` prop with the removal of `React.FC`.
- A couple other small code clean up items (renaming 'DDrawer' to 'Drawer', destructuring rather than using `props.`)

## How to test 🧪

Visually compare prod and this branch in a few places the components are used throughout the app. For example:
- The Add/Resize/Rename a Disk on a linode drawer
- The Add Payment Method drawer: https://cloud.linode.com/account/billing/add-payment-method
- The View Kubeconfig drawer (the only place `DrawerContent` component is used)

There should be no visual changes.

```
yarn test DrawerContent
```

### DIFF
--- a/packages/manager/src/components/Drawer/Drawer.tsx
+++ b/packages/manager/src/components/Drawer/Drawer.tsx
@@ -1,8 +1,8 @@
 import Close from '@mui/icons-material/Close';
 import * as React from 'react';
 import Button from 'src/components/Button';
-import Drawer, { DrawerProps } from 'src/components/core/Drawer';
-import { makeStyles } from '@mui/styles';
+import _Drawer, { DrawerProps } from 'src/components/core/Drawer';
+import { makeStyles } from 'tss-react/mui';
 import { Theme } from '@mui/material/styles';
 import Typography from 'src/components/core/Typography';
 import Grid from 'src/components/Grid';
@@ -13,7 +13,7 @@ export interface Props extends DrawerProps {
   wide?: boolean;
 }
 
-const useStyles = makeStyles((theme: Theme) => ({
+const useStyles = makeStyles()((theme: Theme) => ({
   paper: {
     width: 300,
     padding: theme.spacing(2),
@@ -59,17 +59,15 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
 }));
 
-type CombinedProps = Props;
-
-const DDrawer: React.FC<CombinedProps> = (props) => {
-  const classes = useStyles();
+const Drawer = (props: Props) => {
+  const { classes } = useStyles();
 
   const { title, children, onClose, wide, ...rest } = props;
 
   const titleID = convertForAria(title);
 
   return (
-    <Drawer
+    <_Drawer
       anchor="right"
       classes={{ paper: `${classes.paper} ${wide ? classes.wide : ''}` }}
       onClose={(event, reason) => {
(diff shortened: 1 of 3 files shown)


packages/manager/src/components/Drawer/Drawer.tsx
```diff
@@ -13,7 +13,7 @@ export interface Props extends DrawerProps {
   wide?: boolean;
 }
 
-const useStyles = makeStyles((theme: Theme) => ({
+const useStyles = makeStyles()((theme: Theme) => ({
```
mjac0bs, line 16: I have noticed one visual difference in spacing for the `Drawer` component, which seems to be caused by the [`drawerHeader`](https://github.com/linode/manager/pull/8908/files#diff-6284e075b32ab7b54d10cf6e2a4c49ce666800bd1a86f45ed031cd4e321f84fcR39-R41) rule that defines a `marginBottom` somehow being overwritten now. 

I'm not sure what's causing this or if there's a best approach  to take to reapply the styling, assuming we want to avoid `!important` -- feedback welcome.

This branch:
Screenshot 2023-03-23 at 7 45 06 AM

Prod:

Screenshot 2023-03-23 at 7 48 14 AM

bnussman-akamai: I think that might be related to the work @jaalah-akamai is working on to fix the `<Grid />`.

packages/manager/src/components/DrawerContent/DrawerContent.test.tsx
```diff
@@ -3,17 +3,16 @@ import * as React from 'react';
...
-      <div>Content</div>
-    </DrawerContent>
-  );
+  renderWithTheme(<DrawerContent {...props} />);
 
 const props: Props = {
   loading: true,
   error: false,
   title: 'my-drawer',
+  children: defaultChildren,
```
bnussman-akamai, line 15: Optional change but because the const `defaultChildren` is only used once, we could just put the jsx here

```suggestion
  children: <div>Content</div>,
```

packages/manager/src/components/DrawerContent/DrawerContent.test.tsx
```diff
@@ -3,17 +3,16 @@ import * as React from 'react';
 import { renderWithTheme } from 'src/utilities/testHelpers';
 import { DrawerContent, Props } from './DrawerContent';
 
+const defaultChildren = <div>Content</div>;
```
bnussman-akamai, line 6:
```suggestion
```

bnussman-akamai (approved):
Overall, the styling looks good! I left one tiny item of feedback

dwiley-akamai (approved):

packages/manager/src/components/Drawer/Drawer.tsx
jaalah-akamai: Yea I'm refactoring Grid, but it's a monster component. For now you can do this to bump up the specificity.
```
'&&': { // TODO: Remove after Grid is refactored
   marginBottom: theme.spacing(2),
},
```

PR #8909, opened 2023-03-23
fix: [M3-6429] - Fix Tag drawer in LinodesLanding summary view

hkhalil-akamai:
## Description 📝

Fixes the tag drawer in the `LinodesLanding` component when in summary view (this issue was previously concealed by ambiguous types).

Improves type-safety throughout the app by replacing several `any`s with parameterized types in `OrderBy` and related components.

## How to test 🧪

In LinodesLanding:
1. Enter summary view using the button on the right side of the table header.
2. Add several tags to a Linode until the overflow "..." button appears.
3. Click on the overflow button to open the tag drawer.
4. Ensure that the drawer header shows the Linode label.
5. Ensure that tags can be added or deleted.

`yarn test`

`yarn cy:e2e`


bnussman-akamai (approved):

packages/manager/src/hooks/useExtendedLinode.ts
```diff
@@ -22,31 +23,38 @@ export interface ExtendedLinode extends LinodeWithMaintenance {
...
-  const { linodes } = useLinodes();
-  const linode = linodes.itemsById[linodeId];
-  const typesQuery = useSpecificTypes(linode?.type ? [linode.type] : []);
-  const type = typesQuery[0]?.data;
+  const { linodes: linodesMap } = useLinodes();
+  const linodes = (linodeIds ?? Object.keys(linodesMap.itemsById))
+    .map((linodeId) => linodesMap.itemsById[linodeId])
+    .filter(isNotNullOrUndefined);
+  const typesQuery = useSpecificTypes(
+    linodes.map((linode) => linode.type).filter(isNotNullOrUndefined)
```
cpathipa, line 33: I think Boolean also works instead of having isNullOrUndefined. For example, [1, 2, null, undefined].filter(Boolean) outputs [1, 2].
```suggestion
    linodes.map((linode) => linode.type).filter(Boolean)
```

packages/manager/src/hooks/useExtendedLinode.ts
```diff
@@ -22,31 +23,38 @@ export interface ExtendedLinode extends LinodeWithMaintenance {
...
+export const useExtendedLinodes = (linodeIds?: number[]): ExtendedLinode[] => {
   const { data: grants } = useGrants();
-  const { linodes } = useLinodes();
-  const linode = linodes.itemsById[linodeId];
-  const typesQuery = useSpecificTypes(linode?.type ? [linode.type] : []);
-  const type = typesQuery[0]?.data;
+  const { linodes: linodesMap } = useLinodes();
+  const linodes = (linodeIds ?? Object.keys(linodesMap.itemsById))
+    .map((linodeId) => linodesMap.itemsById[linodeId])
+    .filter(isNotNullOrUndefined);
```
cpathipa, line 31: Same here as well.

cpathipa (commented):
Nice work! @hkhalil-akamai TY Improves type-safety. Left a very minor comment of using Boolean instead of `isNullOrUndefined`

packages/manager/src/hooks/useExtendedLinode.ts
hkhalil-akamai: Looks like this works from a javascript perspective, but typescript isn't convinced that the result array doesn't include null:
Screen Shot 2023-03-28 at 11 53 23 AM

cpathipa: Ah I see. Good catch!

cpathipa (approved):

PR #8912, opened 2023-03-23
fix: [M3-6433] - Fix crash for Events with a `null` community post `entity`

bnussman-akamai:
### The Bug 🐛
- Users who use https://www.linode.com/community/ were experiencing crashes when they opened the events menu when they clicked the bell icon in the top right of Cloud Manager
- Cloud Manager would crash because a `community_question_reply` event would exist where `entity` is `null`. When we did `entity!.url`, this would cause a crash because you can't access a property on null

### The Fix 🧰
- The actual fix is safely accessing the `url` value on line `22` of `packages/manager/src/utilities/getEventsActionLink.ts` (this is what was causing the crash)
- This PR also includes some minor improvements:
  - Uses named export instead of default export 
  - Improves some event messages for community related events

## Preview 📷

Screenshot 2023-03-23 at 3 10 22 PM

## How to test 🧪

- Turn the MSW **on**
- Open the Event/Notification menu by clicking the bell icon in the top right of Cloud Manager
- If Cloud Manager does not crash, we can conclude this bug is fixed
  
### 
  > **Note**: I updated the events mocks to test this. I will restore them to more sane defaults before merging

### DIFF
--- a/packages/manager/src/eventMessageGenerator.ts
+++ b/packages/manager/src/eventMessageGenerator.ts
@@ -55,19 +55,19 @@ export const eventMessageCreators: { [index: string]: CreatorsForStatus } = {
     notification: (e) =>
       e.entity
         ? `There has been a reply to your thread "${e.entity.label}".`
-        : `There has been a reply to your thread`,
+        : `There has been a reply to your thread.`,
   },
   community_like: {
     notification: (e) =>
       e.entity
         ? `A post on "${e.entity.label}" has been liked.`
-        : `There has been a like on your community post`,
+        : `There has been a like on your community post.`,
   },
   community_mention: {
     notification: (e) =>
       e.entity
-        ? `You have been mentioned in a Community post: ${e.entity.label}`
-        : `You have been mentioned in a Community post`,
+        ? `You have been mentioned in a Community post: ${e.entity.label}.`
+        : `You have been mentioned in a Community post.`,
   },
   credit_card_updated: {
     notification: (e) => `Credit card information has been updated.`,
--- a/packages/manager/src/eventMessageGenerator.ts
+++ b/packages/manager/src/eventMessageGenerator.ts
@@ -2,7 +2,7 @@ import { Event } from '@linode/api-v4/lib/account';
 import { path } from 'ramda';
 import { isProductionBuild } from 'src/constants';
 import { reportException } from 'src/exceptionReporting';
-import createLinkHandlerForNotification from 'src/utilities/getEventsActionLink';
+import { getLinkForEvent } from 'src/utilities/getEventsActionLink';
 import {
   formatEventWithUsername,
   formatEventWithAppendedText,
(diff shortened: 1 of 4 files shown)


packages/manager/src/eventMessageGenerator.ts
```diff
@@ -53,14 +53,21 @@ export const eventMessageCreators: { [index: string]: CreatorsForStatus } = {
   },
   community_question_reply: {
     notification: (e) =>
-      `There has been a reply to your thread "${e.entity!.label}".`,
+      e.entity
+        ? `There has been a reply to your thread "${e.entity.label}".`
+        : `There has been a reply to your thread`,
```
hkhalil-akamai, line 58:
```suggestion
        : `There has been a reply to your thread.`,
```

packages/manager/src/eventMessageGenerator.ts
```diff
@@ -53,14 +53,21 @@ export const eventMessageCreators: { [index: string]: CreatorsForStatus } = {
...
   },
   community_like: {
-    notification: (e) => e.entity!.label,
+    notification: (e) =>
+      e.entity
+        ? `A post on "${e.entity.label}" has been liked.`
+        : `There has been a like on your community post`,
```
hkhalil-akamai, line 64:
```suggestion
        : `There has been a like on your community post.`,
```

packages/manager/src/eventMessageGenerator.ts
```diff
@@ -53,14 +53,21 @@ export const eventMessageCreators: { [index: string]: CreatorsForStatus } = {
...
   },
   community_mention: {
     notification: (e) =>
-      `You have been mentioned in a Community post: ${e.entity!.label}`,
+      e.entity
+        ? `You have been mentioned in a Community post: ${e.entity.label}`
+        : `You have been mentioned in a Community post`,
```
hkhalil-akamai, line 70: This one didn't have punctuation originally but maybe it should?

packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -799,49 +798,38 @@ export const handlers = [
...
-      percent_complete: 65,
+    const brokenEvent = eventFactory.build({
+      action: 'community_question_reply',
+      message: 'some community reply',
+      username: 'randomperson',
+      status: 'notification',
+      percent_complete: null,
+      entity: null,
+      seen: false,
+      read: false,
```
hkhalil-akamai, line 809: Maybe we should keep the broken event? If this is something we have to deal with in prod it would be good to have it here. Also, should there be a unit test ensuring `EventRow` gracefully handles it.

hkhalil-akamai (approved):
Approved pending some minor feedback.

I agree that we shouldn't use the `!` assertion, even if we're confident a given value is not null/undefined. Definitely more robust to handle it with an if or conditional.

cpathipa (commented):
LGTM!

cpathipa (approved):

PR #8914, opened 2023-03-23
refactor: Update the `usePagination` hook to use Query Params instead of state

bnussman-akamai:
## Description 📝

- Updates the `usePagination` hook so it is more useful
  - Uses query params instead of state so page persists when users use browser navigation
- Updates `<SupportTicketLanding />` to be a functional component and improves its use of query params
  - I had to do this because I wanted the back button to work correctly
- Adds Pagination controls to Firewalls landing because I forgot to do that in https://github.com/linode/manager/pull/8889

### Before
Notice how when the user goes **back**, they don't return to the page they were perviously on

https://user-images.githubusercontent.com/115251059/227382914-40bf433c-fe2d-471f-9987-6af60b542ac3.mov

### After
Notice how when the user goes **back**, they **do** return to the page they were previously on

https://user-images.githubusercontent.com/115251059/227383013-0608cac0-9c68-4e93-a754-6eb3891e2d64.mov

## How to test 🧪

- Verify that the SupportTicketLanding page still works 
- Spin up at least 25 of either an Image, LKE Cluster, Database, Firewall, Volume, or any other entity that uses the `usePagination` hook
  - Verify that pagination works as expected
  - Verify that browser navigation works as expected (going back from a details page should bring you back to the previous page you were on)
  - Verify that changing the page size works and that is persists in user preferences


packages/manager/src/features/Firewalls/FirewallLanding/FirewallLanding.tsx
```diff
@@ -208,6 +173,14 @@ const FirewallLanding = () => {
           ))}
         </TableBody>
       </Table>
+      <PaginationFooter
+        count={data?.results ?? 0}
+        handlePageChange={pagination.handlePageChange}
+        handleSizeChange={pagination.handlePageSizeChange}
```
jaalah-akamai, line 179: Based on our team conversation, should these be `onPageChange` and `onSizeChange`?

packages/manager/src/features/Support/SupportTickets/SupportTicketsLanding.tsx
```diff
@@ -1,203 +1,106 @@
...
+  /** ?drawerOpen=true to allow external links to go directly to the ticket drawer */
+  const parsedParams = getParamsFromUrl(location.search);
+
+  const stateParams = location.state;
+
+  const [drawerOpen, setDrawerOpen] = React.useState(
+    stateParams ? stateParams.open : parsedParams.drawerOpen === 'true'
+  );
+
+  // why do we stores these in state?
```
jaalah-akamai, line 30: Agreed - not necessary. We can do this in a separate PR, but probably should add a TODO here. 👍

packages/manager/src/hooks/usePagination.ts
```diff
@@ -1,42 +1,74 @@
...
+  const searchParams = new URLSearchParams(location.search);
+  const searchParamPage = searchParams.get(pageKey);
+  const searchParamPageSize = searchParams.get(pageSizeKey);
+
+  const page = searchParamPage ? Number(searchParamPage) : initialPage;
+  const pageSize = searchParamPageSize
+    ? Number(searchParamPageSize)
+    : preferenceKey
+    ? preferences?.pageSizes?.[preferenceKey] ?? 25
+    : 25;
```
jaalah-akamai, line 42: I wonder if this should be a constant?

cpathipa (approved):
LGTM! working as expected.

PR #8915, opened 2023-03-24
refactor: [M3-6436] - React Query for Notifications

bnussman-akamai:
## Description 📝

- Uses React Query for Account Notifications
- Does not do any fancy pagination, just fetches all notifications

## Preview 📷

- No visual changes

## How to test 🧪

- Testing notifications is difficult. Here are a few things I'd recommend to try to test
  - Resizing a Linode - this will trigger a notification
  - Test the notification menu with the MSW 
- Test the Billing Contact information form - I rewrote this to a FC and it would be bad if it broke


packages/manager/src/queries/accountNotifications.ts
```diff
@@ -0,0 +1,15 @@
+import { useQuery } from 'react-query';
+import { getAll } from 'src/utilities/getAll';
+import { queryKey as accountQueryKey } from './account';
+import { APIError, getNotifications, Notification } from '@linode/api-v4';
+
+export const queryKey = [accountQueryKey, 'notifications'];
+
+export const useNotificationsQuery = () =>
+  useQuery<Notification[], APIError[]>({
```
hkhalil-akamai, line 9: What's the default cache window here? Should we be applying any presets?

packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/UpdateContactInformationForm/UpdateContactInformationForm.tsx
```diff
@@ -381,39 +275,37 @@ class _UpdateContactInformationForm extends React.Component<
...
               isClearable={false}
-              onChange={this.updateState}
+              onChange={(item: Item) =>
+                formik.setFieldValue('state', item.value)
+              }
               options={filteredRegionResults}
               placeholder={
-                account.state ||
-                `Select ${fields.country === 'US' ? 'state' : 'province'}`
+                formik.values.country === 'US' ? 'state' : 'province'
```
hkhalil-akamai, line 303: Minor regression here:
```suggestion
                `Select ${formik.values.country === 'US' ? 'state' : 'province'}`
```

hkhalil-akamai (approved):
Overall LGTM, left some minor feedback. Great work on modernizing `UpdateContactInformationForm`, that must have been some effort.

packages/manager/src/App.tsx
```diff
@@ -228,15 +218,7 @@ export class App extends React.Component<CombinedProps, State> {
...
-    if (
-      hasOauthError(
-        linodesError,
-        notificationsError,
-        volumesError,
-        bucketsError,
-        nodeBalancersError
-      )
-    ) {
+    if (hasOauthError(linodesError, nodeBalancersError)) {
```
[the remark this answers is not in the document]
bnussman-akamai: Keeping those two because they are the only ones in Redux. All of the other ones have been migrated to react query so the values like `notificationsError` I assume were doing nothing

packages/manager/src/components/AbuseTicketBanner/AbuseTicketBanner.test.tsx
```diff
@@ -1,85 +1,109 @@
...
 
 const TICKET_TESTID = 'abuse-ticket-link';
 
 describe('Abuse ticket banner', () => {
-  it('should render a banner for an abuse ticket', () => {
-    const { queryAllByText } = render(
-      wrapWithTheme(<AbuseTicketBanner />, {
-        customStore: makeMockStore(abuseTicketNotificationFactory.buildList(1)),
+  it('should render a banner for an abuse ticket', async () => {
+    server.use(
```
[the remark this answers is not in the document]
bnussman-akamai: I think that would be bad because if someone decided to change a mock, it would break the test. Defining the mock API data in the test itself will ensure we have a predictable testing environment that won't randomly change or break.

packages/manager/src/queries/accountNotifications.ts
bnussman-akamai: I've been using the React Query defaults because I think they are fine and they keep things simple. I think once we finish our big React Query effort, we can fine tune settings as needed.

packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/UpdateContactInformationForm/UpdateContactInformationForm.tsx
bnussman-akamai: I did that because the placeholder of `account.state` will never show because the `value` of this input is essentially `account.state`. I may be misunderstanding the intent of the original code but I think this change makes sense

jaalah-akamai (approved):

PR #8921, opened 2023-03-24
feat: Add Metadata region capability

hana-linode:
## Description 📝
Not every region supports Metadata. This PR adds an additional check for region capability before showing the Add User Data Accordion.

## How to test 🧪
- Turn on the MSW and go to the Linode Create page
- Check the Distributions, Images, Backups, and Clone Linode flow
    - Select the metadata-test image and a region. The Add User Data accordion should only display if the selected region supports Metadata (currently, that is only London, UK)


packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -368,7 +378,16 @@ export const handlers = [
...
+    const metadataLinodeWithCompatibleImage = linodeFactory.build({
+      image: 'metadata-test-image',
+      label: 'metadata-test-image',
+    });
+    const metadataLinodeWithCompatibleImageAndRegion = linodeFactory.build({
+      region: 'eu-west',
+      image: 'metadata-test-image',
+      label: 'metadata-test-region',
+    });
+    const onlineLinodes = linodeFactory.buildList(40, {
```
mjac0bs, line 390: What's the reason behind mocking 40 linodes instead of 60?

bnussman-akamai (approved):

mjac0bs (approved):
Only seeing the Add User Data accordion now for our currently metadata enabled region for the listed flows. ✅

packages/manager/src/mocks/serverHandlers.ts
hana-linode: This was part of me trying to make the metadata-test linodes display first. Turns out I just had to move the declaration above to get a lower id number

PR #8934, opened 2023-03-29
fix: Linode Details Notifications showing all notifications

bnussman-akamai:
## Description 📝

- When I did https://github.com/linode/manager/pull/8915 I caused a this regression
- `packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/Notifications.tsx` is only suppose to show notifications regarding a specific Linode but in my refactor, I accidently render **all** notifications. This fix restores previous behavior which is only showing notifications on a Linode details page that are associated with a Linode

## How to test 🧪

- Turn **on** the MSW
- Go to `http://localhost:3000/linodes/0`
- Verify that you **do not** see a notice saying anything about a "past due balance"
- If you are able to confirm that, you can consider this fix to be working

### DIFF
--- a/packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/Notifications.tsx
+++ b/packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/Notifications.tsx
@@ -15,6 +15,12 @@ const Notifications = () => {
 
   const { data: notifications, refetch } = useNotificationsQuery();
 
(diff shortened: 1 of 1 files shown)


jdamore-linode (approved):
Nice fix @bnussman-akamai ! Confirming that in `develop`, the past due notice is shown, and on this branch it is not

dwiley-akamai (approved):

PR #8935, opened 2023-03-29
chore: Update `yarn.lock` so lockfile is up to date

bnussman-akamai:
## Description 📝

- Brings our `yarn.lock` into sync with our dependencies
- This isn't a big deal but when the lockfile is out of sync, `yarn dev` will have to re-evaluate dependencies every time it is ran, which can be annoying and can slowdown development

## How to test 🧪

- Run `yarn up` or `yarn dev`and verify things function normally
- View the diff and verify that it is what you'd expect

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1270,17 +1270,6 @@
     "@emotion/weak-memoize" "^0.3.0"
     stylis "4.1.3"
 
-"@emotion/cache@*":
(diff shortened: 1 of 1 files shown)


dwiley-akamai (approved):
LGTM 🚢

mjac0bs (approved):

PR #8937, opened 2023-03-29
feat: [M3-6196] Add Cross Data Center Clone warning

hana-linode:
## Description 📝

**What does this PR do?**

## Preview 📷

## How to test 🧪

### DIFF
--- a/packages/manager/src/components/SelectRegionPanel/SelectRegionPanel.tsx
+++ b/packages/manager/src/components/SelectRegionPanel/SelectRegionPanel.tsx
@@ -1,11 +1,13 @@
 import { Region } from '@linode/api-v4/lib/regions';
+import { Theme } from '@mui/material/styles';
+import { createStyles, withStyles, WithStyles } from '@mui/styles';
 import * as React from 'react';
 import { compose } from 'recompose';
+import Box from 'src/components/core/Box';
 import Paper from 'src/components/core/Paper';
-import { createStyles, withStyles, WithStyles } from '@mui/styles';
-import { Theme } from '@mui/material/styles';
 import Typography from 'src/components/core/Typography';
 import RegionSelect from 'src/components/EnhancedSelect/variants/RegionSelect';
+import Notice from 'src/components/Notice';
 import RenderGuard, { RenderGuardProps } from 'src/components/RenderGuard';
 import { sendLinodeCreateDocsEvent } from 'src/utilities/ga';
 
@@ -31,6 +33,7 @@ interface Props {
   selectedID?: string;
   disabled?: boolean;
   helperText?: string;
+  warningNoticeText?: string;
 }
 
 export const regionHelperText = (onClick?: () => void) => (
@@ -60,6 +63,7 @@ const SelectRegionPanel: React.FC<Props & WithStyles<ClassNames>> = (props) => {
     helperText,
     regions,
     selectedID,
+    warningNoticeText,
   } = props;
 
   if (props.regions.length === 0) {
@@ -72,6 +76,11 @@ const SelectRegionPanel: React.FC<Props & WithStyles<ClassNames>> = (props) => {
         Region
       </Typography>
       {regionHelperText(() => sendLinodeCreateDocsEvent('Speedtest Link'))}
+      {warningNoticeText ? (
+        <Box marginTop="16px">
+          <Notice warning text={warningNoticeText} />
+        </Box>
+      ) : null}
       <RegionSelect
         errorText={error}
         disabled={disabled}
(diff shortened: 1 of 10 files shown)


packages/manager/src/features/linodes/LinodesCreate/LinodeCreateContainer.tsx
```diff
@@ -105,6 +105,7 @@ interface State {
   vlanIPAMAddress: string | null;
   authorized_users: string[];
   userData: string | undefined;
+  showCrossDataCenterCloneWarning: boolean;
```
bnussman-akamai, line 108: I don't think we should have something like `showCrossDataCenterCloneWarning` in state when this value is something we can derive from the state we already have available.

It would fundamentally change a lot of the work you did here but I think it is important we don't add more state to Linode create flow unless we really have to

packages/manager/src/components/SelectRegionPanel/SelectRegionPanel.tsx
bnussman-akamai, line 1: This is what I mean by my other comment. I used react query for this example, but we'd use actually use redux probably

```typescript
interface Props {
  linodeId?: number;
  createType: CreateTypes;
}

//...

const SelectRegionPanel: React.FC<Props & WithStyles<ClassNames>> = (props) => {
  const {
    // ...
    linodeId,
    createType,
  } = props;

  const { data: linode } = useLinodeQuery(
    linodeId ?? -1,
    linodeId !== undefined
  );

  const isCloningToDifferentRegion =
    linode?.region !== selectedID && createType === 'fromLinode';

  return (
    <Paper className={classes.root}>
      <Typography variant="h2" data-qa-tp="Region">
        Region
      </Typography>
      {regionHelperText(() => sendLinodeCreateDocsEvent('Speedtest Link'))}
      {isCloningToDifferentRegion ? (
        <Box marginTop="16px" data-testid="region-select-warning">
          <Notice warning text={CROSS_DATA_CENTER_CLONE_WARNING} />
        </Box>
      ) : null}
      <RegionSelect
        errorText={error}
        disabled={disabled}
        handleSelection={handleSelection}
        regions={regions}
        selectedID={selectedID || null}
        helperText={helperText}
      />
    </Paper>
  );

};
```

hana-linode: @bnussman-akamai Since the state can be derived from the url, I went with that approach instead

packages/manager/src/features/linodes/LinodesCreate/LinodeCreateContainer.tsx
```diff
@@ -799,6 +811,7 @@ class LinodeCreateContainer extends React.PureComponent<CombinedProps, State> {
             accountBackupsEnabled={getAccountBackupsEnabled()}
             setAuthorizedUsers={this.setAuthorizedUsers}
             updateUserData={this.setUserData}
+            updateSearchParams={this.setSearchParams}
```
bnussman-akamai, line 814: Do we need to pass this as a prop? Can `<LinodeCreate />` just have its own `setSearchParams` so we don't need another prop?

```suggestion
            updateSearchParams={this.setSearchParams}
```

packages/manager/src/components/SelectRegionPanel/SelectRegionPanel.tsx
```diff
@@ -31,6 +36,7 @@ interface Props {
   selectedID?: string;
   disabled?: boolean;
   helperText?: string;
+  warningNoticeText?: string;
```
bnussman-akamai, line 39:
```suggestion
```

packages/manager/src/components/SelectRegionPanel/SelectRegionPanel.tsx
```diff
@@ -62,16 +67,27 @@ const SelectRegionPanel: React.FC<Props & WithStyles<ClassNames>> = (props) => {
...
 
+  const location = useLocation();
+
   if (props.regions.length === 0) {
     return null;
   }
 
+  const params = getParamsFromUrl(location.search);
+  const showCrossDataCenterCloneWarning =
+    /clone/i.test(params.type) && params.regionID !== selectedID;
```
bnussman-akamai, line 78: This is very minor, but it seems like `selectedID` is sometimes an empty string when `params.regionID` is `undefined`, resulting in the Notice showing when is does not need to




https://user-images.githubusercontent.com/115251059/229572465-e02b324e-43ae-4468-9a63-740163f95a8b.mov


I found that adding a check for `selectedID` fixes this.
 ```suggestion
     /clone/i.test(params.type) && selectedID && params.regionID !== selectedID;
```

bnussman-akamai (approved):

cpathipa (approved):
LGTM! Confirming that cloning a Linode to a different data center shows warning notification.

PR #8941, opened 2023-03-29
perf: Only fetch `/v4/profile/grants` if a user is `restricted`

bnussman-akamai:
## Description 📝

- This is a small tweak to our grants query that only enables the query when a user is restricted

### Why ❓
- For unrestricted users, the `/v4/profile/grants` endpoint returns an empty response because the user is unrestricted 
  - https://www.linode.com/docs/api/profile/#grants-list
- This change will save Cloud Manager from making an unnecessary API call resulting in possible better performance and less load on the Linode API

## Preview 📷

| Before | After |
|--------|-------|
| Screenshot 2023-03-29 at 6 40 49 PM     | after     |

## How to test 🧪

- As an unrestricted user
  - Login and verify that Cloud does **not** make an API call to `/v4/profile/grants` 
- As a restricted user
  - Login and verify that Cloud **does** make an API call to `/v4/profile/grants`

### DIFF
--- a/packages/manager/src/queries/profile.ts
+++ b/packages/manager/src/queries/profile.ts
@@ -43,12 +43,13 @@ export const updateProfileData = (newData: Partial<Profile>): void => {
   }));
 };
 
-export const useGrants = () =>
-  useQuery<Grants, APIError[]>(
-    [queryKey, 'grants'],
-    listGrants,
-    queryPresets.oneTimeFetch
-  );
+export const useGrants = () => {
+  const { data: profile } = useProfile();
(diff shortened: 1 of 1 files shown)


mjac0bs (approved):
Confirmed that the request to `/v4/profile/grants` now happens for only restricted, not unrestricted, user. 🚀

jaalah-akamai (approved):
Confirmed, no grants as unrestricted user 👍 
Screen Shot 2023-03-31 at 3 20 23 PM

PR #8943, opened 2023-03-30
chore:[M3-6313] - Add column head to Plan Name

cpathipa:
## Description 📝
Add plan column name to SelectPlanPanel component

## Preview 📷
image

## How to test 🧪

- Validate SelectPlanPanel component all flows of Linode Create, rebuild, clone etc.

yarn test
yarn cy:run

### DIFF
--- a/packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
@@ -709,6 +709,7 @@ export class LinodeCreate extends React.PureComponent<
             disabledClasses={this.props.disabledClasses}
             isCreate
             showTransfer
+            header="Linode Plan"
             docsLink={
               <DocsLink
                 href="https://www.linode.com/docs/guides/choosing-a-compute-instance-plan/"
--- a/packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
+++ b/packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
@@ -709,11 +709,10 @@ export class LinodeCreate extends React.PureComponent<
             disabledClasses={this.props.disabledClasses}
             isCreate
             showTransfer
-            header="Plan"
             docsLink={
               <DocsLink
                 href="https://www.linode.com/docs/guides/choosing-a-compute-instance-plan/"
-                label="Plan"
+                label="Choosing a Plan"
                 onClick={() => {
                   sendEvent({
                     category: 'Linode Create Flow',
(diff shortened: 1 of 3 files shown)


mjac0bs (commented):
As long as the decision was made to keep the varied headings for the SelectionPlanTable ("Linode Plan" in the Linode Create flows, "Plan" in the Kubernetes flow, and "Choose a Plan" in the Databases flow), then this change looks like it gets the job done. 

@cpathipa Do you know if UX had any input on the above? It seems a little strange to me to have these slight copy variations in 3 different places.

packages/manager/src/features/linodes/LinodesCreate/LinodeCreate.tsx
```diff
@@ -709,7 +709,7 @@ export class LinodeCreate extends React.PureComponent<
             disabledClasses={this.props.disabledClasses}
             isCreate
             showTransfer
-            header="Linode Plan"
+            header="Plan"
```
mjac0bs, line 712: It looks like we're going to want to edit [`DatabaseCreate.tsx`](https://github.com/linode/manager/blob/develop/packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx#L519) to change the header prop passed in there too, as it still reads "Choose a Plan".

But to help maintain consistency across components, what do you think about modifying [`SelectPlanPanel.tsx`](https://github.com/linode/manager/blob/develop/packages/manager/src/features/linodes/LinodesCreate/SelectPlanPanel.tsx#L347-L349) to give our current header a default of "Plan" to fall back on if no header prop is provided? And then removing the header prop in places where it's currently passed in to instances of `<SelectPlanPanel/>`.

Something like:
```
    <TableCell className={classes.headerCell} data-qa-plan-header>
      {header ?? 'Plan'}
    </TableCell>
```

cpathipa: For the second point, IMO, now that we have a unified column name of "Plan" across the board, I would suggest that we can completely eliminate the need to rely on the `header` prop for rendering the column name.

`<TableCell className={classes.headerCell} data-qa-plan-header>
      Plan
    </TableCell>`

mjac0bs: @cpathipa That's fair, and probably the better move to prevent us from straying from our unified copy going forward. Let's just then go ahead and do the clean up to remove the header prop from the component and its instances where it was needed.

jaalah-akamai (approved):

mjac0bs (approved):
Approving following an offline discussion about the need to keep passing header props down to `TabbedPanel` from `SelectPlanPanel` to display an accurate panel header. Addressing prop drilling that's happening here could be tackled in a tech story; this change looks good for the scope of this ticket. 🚢

PR #8944, opened 2023-03-30
test: [M3-6247] - Add Cypress Firewall e2e Create/Delete tests

cliu-akamai:
## Description 📝

Add Cypress end-to-end tests which confirm the expected behavior of Cloud Manager Firewall features.

Test cases:
- Linodes can be added and removed from a firewall
- Inbound rules can be added and removed from a firewall
- Outbound rules can be added and removed from a firewall
- Firewalls can be enabled and disabled, and their status is reflected on the landing page (/firewalls)

## How to test 🧪

`yarn cy:run -s "cypress/e2e/firewalls/create-firewall.spec.ts"`
`yarn cy:run -s "cypress/e2e/firewalls/delete-firewall.spec.ts"`


packages/manager/cypress/e2e/firewalls/create-firewall.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    interceptCreateFirewall().as('createFirewall');
+    cy.visitWithLogin('/firewalls/create');
+
+    // An error message appears when attempting to create a Firewall without a label
+    getClick('[data-testid="create-firewall-submit"]');
+    cy.findByText('Label is required.');
+
+    // Fill out and submit firewall create form.
+    containsClick('Label').type(firewall.label);
+    getClick('[data-testid="create-firewall-submit"]');
```
jdamore-linode, line 31: There's some potential for flake here if this test runs on an account which already has one or more Firewalls -- `containsClick('Label')` will select both the _Label_ table column header on the landing page, and the _Label_ text field in the _Create Firewall_ drawer.

We can fix this by selecting the _Create Firewall_ drawer and only making these assertions in the context of that drawer:

```suggestion
    ui.drawer
      .findByTitle('Create Firewall')
      .should('be.visible')
      .within(() => {
        // An error message appears when attempting to create a Firewall without a label
        getClick('[data-testid="create-firewall-submit"]');
        cy.findByText('Label is required.');

        // Fill out and submit firewall create form.
        containsClick('Label').type(firewall.label);
        getClick('[data-testid="create-firewall-submit"]');
      });
```

(This also applies to lines `72`-`76`, and you'll need to import the `ui` object from `support/ui` to use the drawer helper)

packages/manager/cypress/e2e/firewalls/create-firewall.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+      label: randomLabel(),
+    };
+
+    cy.defer(createLinode(linodeRequest)).then((linode) => {
+      interceptCreateFirewall().as('createFirewall');
+      cy.visitWithLogin('/firewalls/create');
+
+      // Fill out and submit firewall create form.
+      containsClick('Label').type(firewall.label);
+      containsClick('Select a Linode').type(`${linode.label}{enter}`);
```
jdamore-linode, line 68: When I run this locally, Cypress is failing to select the Linodes drop-down -- I think there are a couple issues at play:

- I think we should be searching for `Select Linodes` rather than `Select a Linode`
- Once we start typing the Linode label, we need to wait until the desired Linode is shown in the drop-down before we make our selection by pressing `enter`, otherwise it proceeds as if no Linodes were selected

Banks and I had a quick discussion about this UI pattern a couple weeks ago:
https://github.com/linode/manager/pull/8889#discussion_r1140363483

If you have any trouble getting this working, we can take a closer look during our 1:1 tomorrow!

packages/manager/cypress/e2e/firewalls/delete-firewall.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+            .findButtonByTitle('Cancel')
+            .should('be.visible')
+            .should('be.enabled')
+            .click();
+        });
+
+      // Confirm that firewall is still listed and initiate deletion again.
+      cy.findByText(firewall.label).should('be.visible');
+      fbtVisible('Delete');
+      fbtClick('Delete');
```
jdamore-linode, line 45: This may also lead to flakiness in situations where the user already has one or more firewalls; Cypress won't know which _Delete_ button to press. This can be fixed using a pattern similar to what you're using in `create-firewall.spec` by selecting only the relevant table row:

```suggestion
      cy.findByText(firewall.label)
        .should('be.visible')
        .closest('tr')
        .within(() => {
          fbtVisible('Delete');
          fbtClick('Delete');
        });
```

(This also applies to lines `43`-`45`)

jdamore-linode (commented):
Nice work @cliu-akamai! I encountered an issue with interactions surrounding the _Select Linodes_ drop-down in the Create Firewall drawer, and also noticed a couple spots that might lead to flakiness in certain circumstances, but otherwise this is an excellent start!

packages/manager/cypress/support/ui/autocomplete-popper.ts
```diff
@@ -0,0 +1,16 @@
...
+ */
+export const autocompletePopper = {
+  /**
+   * Finds a autocomplete popper that has the given title.
+   */
+  findByTitle: (title: string): Cypress.Chainable => {
+    return cy
+      .findByText(`${title}`)
+      .closest('[data-qa-autocomplete-popper="true"]');
+  },
```
jdamore-linode, line 15:
```suggestion
  /**
   * Finds an autocomplete popper item that has the given title.
   */
  findItemByTitle: (title: string): Cypress.Chainable => {
    return cy
      .document()
      .its('body')
      .find('[data-qa-autocomplete-popper]')
      .findByText(title);
  },
```

Quick suggestion for the new UI helper! Aside from the function name and docs tweaks, selecting the popper from anywhere in the document body will allow us to use this utility from inside `within(() => { ... })` callbacks which will be more convenient in certain situations, but will still work the way you're using it now, too.

I think in the future there is potential to use this to interact with other drop-downs and autocomplete fields as well (especially the Linode region select) -- I'm expecting this to be a pretty big time saver. Nice work!

jdamore-linode (approved):
Nice work @cliu-akamai! Just pushed a quick fix to narrow the scope when searching for the "Delete" button for the first time in the delete test -- I totally missed it when I reviewed it the first time.

bnussman-akamai (commented):
Tests ran and passed ✅

One thing I noticed is that the test called `adds linode to firewall - real data` in `migrate-linode-with-firewall.spec.ts` appears to do the same thing as the test called `creates a firewall assigned to a linode` in `create-firewall.spec.ts`. Are these tests duplicates? If so, should we remove one of them?

hkhalil-akamai (approved):
Great work!

jdamore-linode:
> Tests ran and passed ✅
> 
> One thing I noticed is that the test called `adds linode to firewall - real data` in `migrate-linode-with-firewall.spec.ts` appears to do the same thing as the test called `creates a firewall assigned to a linode` in `create-firewall.spec.ts`. Are these tests duplicates? If so, should we remove one of them?

Ah great catch @bnussman-akamai, that's on me for missing this test when I wrote the ticket. I think I'd opt to remove the old test since it uses some older patterns and needs clean up anyway (M3-6219), don't mind either way whether we remove it now or later though

PR #8947, opened 2023-03-30
fix: [M3-6453] - Filtering by `Status` on Linode Details Volumes Table

bnussman-akamai:
## Description 📝

- Fixes a bug that caused users to see `Could not apply filter` when filtering by `Status` on the `/linodes/:id/storage` page in the Volumes Table
- This bug was caused by misspelling `status` as `Status` which resulted in an API error because `Status` is not a field returned by the API but `status` is
- This PR also adds pagination controls to Firewalls Landing because they were missing

## The Bug 🐛

https://user-images.githubusercontent.com/115251059/228903288-5cd57927-8c13-48ed-8b38-058fe3871424.mov

## How to test 🧪

- You may need a volume attached to a Linode
- Go to  `/linodes/:id/storage` for that Linode and stort the volumes table by `Status`
- Verify that the API returns a successful response in the dev tools and that the table shows no errors

### DIFF
--- a/packages/manager/src/features/Firewalls/FirewallLanding/FirewallLanding.tsx
+++ b/packages/manager/src/features/Firewalls/FirewallLanding/FirewallLanding.tsx
@@ -19,6 +19,7 @@ import { TableSortCell } from 'src/components/TableSortCell/TableSortCell';
 import TableCell from 'src/components/TableCell/TableCell';
 import TableBody from 'src/components/core/TableBody';
 import Hidden from 'src/components/core/Hidden';
+import PaginationFooter from 'src/components/PaginationFooter';
 
 const preferenceKey = 'firewalls';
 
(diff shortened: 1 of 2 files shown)


afolksetapart (approved):

bnussman-akamai:
Pushed a0fc52a to fix missing pagination footer on Firewalls landing page

jdamore-linode (approved):
Nice work, thanks @bnussman-akamai! Able to confirm the 400 error exists when sorting in Prod due to the "Status" order by value, and confirmed that it's fixed locally after changing it to "status".

cpathipa (approved):

carrillo-erik:
If I filter while a newly created volume has a 'Creating' status, it disappears from the Volumes list for a second and then reappears with an 'Active' status.

If I hit the filter button once again, it changes the status back to 'Creating'. Hitting the filter button again changes the status back to 'Active'.

Could be that the Volume status is updated faster than the response from the filter.

https://user-images.githubusercontent.com/119514965/228931222-8c71e2af-19ee-4c37-9f15-2a5bc7fec226.mov

PR #8951, opened 2023-03-31
test: [M3-6428] - Replace Jest with Vitest

jdamore-linode:
## Description 📝

**What does this PR do?**
Experimentally replaces Jest with Vitest. This is most of the way there; the tests are passing (with a rare exception involving the theme wrapper, explained below), but there's a lot going on that I've elaborated on below.

#### Test Changes
Most of the tests didn't require any major changes besides swapping the `jest` keyword with `vi`. Some tests did require modification to pass with Vitest, and they generally required one or more of the following changes:
* When mocking a module, Vitest requires that the entire module is mocked (instead of just the relevant exports) and Jest does not have this requirement. A handful of tests were impacted by this, and were fixed by using `vi.importActual()` to mock everything that we don't explicitly mock ourselves ([example](https://github.com/linode/manager/pull/8951/files#diff-28e03abae1706e21386268e3caee156060966a97a2b83ebc7af5ba1d50c59d4fR12-R17)).
* Some tests that expect environment variables to be set a certain way now use `vi.stubEnv()` to ensure they're set up correctly ([example](https://github.com/linode/manager/pull/8951/files#diff-c464febe8ebc791d6c6e0c6483407bf3fd9406f628c1ab248a1c0bbcda72071e))

I also did some other test clean-up that's not strictly related to Vitest. If we don't go forward with this, we still may want to implement these changes, which fall into these categories:
* Moving all calls to `renderWithTheme()` and Enzyme's `shallow()`, `mount()`, etc. functions inside of `it(...)` callbacks to fix console messages about HTTP requests that are not getting captured by MSW, because the MSW context does not exist when a component is initialized outside of `it(...)`
* Fixed some issues involving `act()` and calling `done()` in async tests

#### Src Changes
* Rewrote `confirmObjectStorage` in `src/features/ObjectStorage/utilities.ts` to use async/await syntax. There was an actual issue with the original implementation (that I think manifested in the tests) but I can't remember the details -- I'll have to take a closer look, and I'll modify this section once I remember more. I verified that the Object Storage enable confirmation works as expected, but another set of eyes is always appreciated.
* I made a pretty small change to `devToolsEnabled()` in `load.ts` to coerce the function to always return a `boolean`; in environments where `window` isn't defined (basically only tests), iirc it was returning the literal value of the `DEV` environment variable which was a string containing the character `1`.
### Development Impact
* `jest` keyword should be replaced with `vi`, which also needs to be imported
  * In most cases `vi` is a drop-in replacement for `jest`, but there are a couple exceptions:
    * `jest.requireActual` is replaced with `vi.importActual`
    * Vitest's spying (and mocking) behavior has a couple minor differences/limitations that came into play with the `SearchHOC.test.tsx` and `refinedSearch.test.ts` tests
* No more `xit`, have to go back to `it.skip` (RIP, we hardly knew you)

#### Skipped Tests
There are a couple tests that I had to skip because I wasn't able to get them working right away. Off the top of my head, the tests related to Algolia search were difficult because of Vitest's spying/mocking limitations. There may have been some others that I couldn't figure out, so I'll circle back and list them to make sure they aren't forgotten.

#### Theme Wrapper Issues
The only other issue involves the `toggleTheme()` function in [`LinodeThemeWrapper.tsx`](https://github.com/linode/manager/blob/develop/packages/manager/src/LinodeThemeWrapper.tsx#L88-L93). This function creates a 500ms timer when the theme is initialized (or toggled later). When a Vitest test uses `renderWithTheme()` and finishes before 500ms has passed, the timer's callback causes an error because the DOM environment has been torn down so `document` no longer exists.

Right now I'm mocking timers to work around this issue, but it causes a bunch of new warnings to appear and I've still seen the error occur once or twice, so I'm not satisfied with this as a solution.

(I believe this is the same issue which causes the **A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks.** warning at the end of many of our Jest runs.)

#### To-Do Before Merge
Things to do before merging if we decide to proceed:

- [ ] Merge in `develop` / resolve conflicts
- [x] Delete Jest dependencies and config in package.json
- [ ] Update docs
- [ ] Revisit skipped tests, see if we can get them working
- [ ] Investigate coverage report generation, including cobertura reports
- [ ] Find a real solution to the theme wrapper issue described above
- [ ] Investigate CI performance
- [x] Fix CI `--colors` arg issue
- [ ] Investigate `eslint-plugin-vitest`
- [x] Jest -> Vitest for API v4 package

#### Future
We may want to look into replacing `jsdom` with `happy-dom` sometime in the future. I tried using it, and while it created new failures (maybe a dozen or two) which I didn't investigate too closely, it did seem to perform faster than JSDom.

## How to test 🧪

1. Check out this branch
2. Run `yarn`
3. Run `yarn test`


bnussman-akamai:
I think this is really great. Even if Github Actions is a bit slower, I think we will still benefit in the long run. This upgrade gives us a lot of new functionality included in Vitest and most importantly solves all of our tech debt related to Jest the crazy amount of boilerplate required for Jest to run.  

I don't see many reasons not to go ahead and merge this as long as we can fix our internal pipeline (`Unknown option --colors`).

I'm looking forward to seeing all of the Jest configuration and dependencies get deleted. Will that be a seperate PR if we continue with this?

jdamore-linode:
> I think this is really great. Even if Github Actions is a bit slower, I think we will still benefit in the long run. This upgrade gives us a lot of new functionality included in Vitest and most importantly solves all of our tech debt related to Jest the crazy amount of boilerplate required for Jest to run.
> 
> I don't see many reasons not to go ahead and merge this as long as we can fix our internal pipeline (`Unknown option --colors`).
> 
> I'm looking forward to seeing all of the Jest configuration and dependencies get deleted. Will that be a seperate PR if we continue with this?

Thanks, @bnussman-akamai! Just updated the description to provide more info, but I pretty much agree on all counts. I don't see any reason not to delete the Jest dependencies as part of this effort (I was just eager to open the PR on Friday) but I figured I'd wait until our cafe Thursday to talk about it with the team before investing any more effort.

Curious what you and the team's take on the theme wrapper issue might be; I have yet to find a great solution that works without modifying the source, but I'm excited that (I think) we've finally tracked down the cause of that Jest warning that's been a pain in our side forever (even if it took messing with Vitest to identify it lol).

jdamore-linode:
Still working on this but currently blocked on 2 issues:

- Updating to the latest version of JSDom seems to cause the tests to hang. We've been using version `16.7.0`, and it appears to continue working through version `19.0.0`, but stops working on versions `20.x.x` and `21.x.x`. There's no reason we can't continue using an older version but this is something we may need to sort out eventually.
- Running `yarn test --no-threads` seems to cause a _lot_ of failures and I haven't been able to understand why. The debug output makes it look like `renderWithTheme()` is causing components to be rendered twice (as in existing twice in the DOM, not re-rendering in the React sense).

bnussman-akamai (changes requested):
Testing banksbot

bnussman-akamai (changes requested):
Testing, i'm so sorry

bnussman-akamai (approved):

jaalah-akamai:
@jdamore-linode Should we close this for the time being?

jdamore-linode:
> @jdamore-linode Should we close this for the time being?

Yeah I think we should -- I'm planning to revisit this but it will probably be a lot easier to start from scratch using this branch as guidance than it would be to get this branch up-to-date.

PR #8952, opened 2023-04-03
refactor: [M3-6441] - Use Query Client from Context

hkhalil-akamai:
## Description 📝

Removes the global instance of `queryClient` and `store` and pulls it from the `QueryClientProvider` and `Provider` contexts instead.

[Reasons why this pattern is preferable. 
](https://stackoverflow.com/a/71787324)

## How to test 🧪

1. Ensure data fetching works as expected throughout the app.
2. `yarn test`
3. `yarn cy:e2e`


packages/manager/src/App.tsx
```diff
@@ -101,29 +101,38 @@ export class App extends React.Component<CombinedProps, State> {
      * Send any Database events to the Database events handler in the queries file
      */
     events$
```
hkhalil-akamai, line 103: The `events$` subject is now of the type `EventWithStore` which includes a reference to the active store and queryClient.

packages/manager/src/components/MaintenanceBanner/MaintenanceBanner.tsx
```diff
@@ -134,7 +135,9 @@ const generateIntroText = (
        * we're going to display both the raw and humanized versions of the date
        * to the user here.
        */
-      const rawDate = formatDate(start);
+      const rawDate = formatDate(start, {
+        timezone,
```
hkhalil-akamai, line 139: Previously, the user's timezone preference was pulled from the global queryClient.

packages/manager/src/features/Billing/BillingPanels/PaymentInfoPanel/PayPalChip.tsx
```diff
@@ -95,16 +96,17 @@ export const PayPalChip = (props: Props) => {
 
   const onApprove = async (
     data: OnApproveBraintreeData,
-    actions: OnApproveBraintreeActions
+    actions: OnApproveBraintreeActions,
+    queryClient: QueryClient
```
bnussman-akamai, line 100: No need to pass the `queryClient` here as a param. It is already in the scope of the component. 

```suggestion
```

dwiley-akamai (commented):
I still have a lot more to sift through, but what I've looked at looks good so far

packages/manager/src/store/linodes/linodes.events.ts
```diff
@@ -6,12 +6,16 @@ import { getAllLinodeDisks } from 'src/store/linodes/disk/disk.requests';
 import { requestLinodeForStore } from 'src/store/linodes/linode.requests';
 import { EventHandler } from 'src/store/types';
 import { deleteLinode } from './linodes.actions';
-import { queryClient } from 'src/queries/base';
-import { queryKey as volumesQueryKey } from 'src/queries/volumes';
 import { queryKey as accountNotificationsQueryKey } from 'src/queries/accountNotifications';
 import { EntityEvent } from '../events/event.types';
+import { QueryClient } from 'react-query';
```
jaalah-akamai, line 11: I think in places where we're importing this type (and it's not needed for some runtime value), it may make sense to add a `type` to have it removed from runtime code. Also it makes it clear that this is being used as a time as opposed to a value: 
```
import type { QueryClient } from 'react-query';
```

packages/manager/src/store/linodes/linodes.events.ts
```diff
@@ -180,13 +185,6 @@ const handleLinodeDelete = (
 ) => {
   const found = state.__resources.linodes.itemsById[id];
 
-  // Removed unneeded volume stores
-  queryClient.removeQueries([`${volumesQueryKey}-list`, 'linode', id]);
-
-  // A Linode that was deleted may have had volumes attached.
-  // Invalidate paginated volume stores so they will reflect there is no attached Linode.
-  queryClient.invalidateQueries([`${volumesQueryKey}-list`]);
-
```
jaalah-akamai, line 189: Just making sure, we just moved this into `linodes.request.ts` now?

packages/manager/src/queries/tags.ts
```diff
@@ -20,7 +20,10 @@ const getAllTagSuggestions = (
     getTags({ ...params, ...passedParams }, { ...filter, ...passedFilter })
   )().then((data) => data.data);
 
-export const updateTagsSuggestionsData = (newData: Tag[]): void => {
+export const updateTagsSuggestionsData = (
+  newData: Tag[],
+  queryClient: QueryClient
```
dwiley-akamai, line 25: Is there a specific reason we're passing `queryClient` in as an argument here vs. doing `const queryClient = useQueryClient();` in the body of the function? Seems like this PR leans towards the latter for the vast majority of cases

packages/manager/src/components/AuthenticationWrapper/AuthenticationWrapper.test.tsx
```diff
@@ -15,6 +20,8 @@ const component = shallow<AuthenticationWrapper>(
     checkAccountSize={jest.fn()}
     pendingUpload={false}
     linodes={[]}
+    queryClient={queryClient}
+    store={store}
```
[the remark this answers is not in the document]
hkhalil-akamai: The query client itself is instantiated in `index.tsx`. This prop comes from `WithQueryClientProps` which normally would be provided by the `withQueryClient` HOC but it's being passed manually here for testing.

packages/manager/src/features/Account/EnableObjectStorage.tsx
```diff
@@ -115,7 +116,7 @@ export const EnableObjectStorage = (props: Props) => {
 
       <Button
         buttonType="primary"
-        onClick={handleSubmit}
+        onClick={() => handleSubmit(queryClient)}
```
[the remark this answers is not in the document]
hkhalil-akamai: No we don't! Good catch

packages/manager/src/features/linodes/LinodesDetail/LinodeBackup/LinodeBackup.tsx
```diff
@@ -227,15 +232,15 @@ class _LinodeBackup extends React.Component<CombinedProps, State> {
...
         [
           'linode_snapshot',
           'backups_enable',
           'backups_cancel',
           'backups_restore',
-        ].includes(e.action)
+        ].includes(event.action)
       )
-      .filter((e) => !e._initial && e.status === 'finished')
+      .filter(({ event }) => !event._initial && event.status === 'finished')
```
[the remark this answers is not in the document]
hkhalil-akamai: I think the two forms are equivalent, but I think the current way is more readable -- each condition is separated out into its own `.filter()` -- @bnussman-akamai @jaalah-akamai thoughts?

packages/manager/src/store/linodes/linodes.events.ts
hkhalil-akamai: Oh, neat! I didn't know about this syntax. Only trouble is that it's a little tricky to know when to apply it since VSCode doesn't add it automatically. Most of the time, when you import a type it's just an interface

packages/manager/src/store/linodes/linodes.events.ts
hkhalil-akamai: That's correct and the advantage of doing it here is that it may take several seconds for the event to come through during which time the volumes data will not update.

packages/manager/src/features/Account/EnableObjectStorage.tsx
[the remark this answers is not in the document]
hkhalil-akamai: No we don't! Good catch

packages/manager/src/features/linodes/LinodesDetail/LinodeBackup/LinodeBackup.tsx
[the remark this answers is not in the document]
bnussman-akamai: I think either is fine. We'll probably be changing this logic soon when we react query events and linodes anyway

dwiley-akamai (commented):
I got one test failure when running `yarn test`:

Screenshot 2023-04-06 at 3 28 24 PM

packages/manager/src/queries/tags.ts
hkhalil-akamai: Good catch! -- I was going too quickly and didn't notice this was a hook

bnussman-akamai (commented):
looking good, will keep test

packages/manager/src/features/Billing/BillingPanels/PaymentInfoPanel/PayPalChip.tsx
```diff
@@ -101,10 +102,10 @@ export const PayPalChip = (props: Props) => {
 
     return actions.braintree
       .tokenizePayment(data)
-      .then((payload) => onNonce(payload.nonce));
+      .then((payload) => onNonce(payload.nonce, queryClient));
   };
 
-  const onNonce = (nonce: string) => {
+  const onNonce = (nonce: string, queryClient: QueryClient) => {
```
bnussman-akamai, line 108:
```suggestion
  const onNonce = (nonce: string) => {
```

The query client is already in the scope here so we don't need it as a parameter to this function

cpathipa (commented):
Overall looks great! Thank you for addressing the feedback. 

Observation: Noticed below E2E tests are failed. 

image
image
image

bnussman-akamai (approved):
Screenshot 2023-04-12 at 11 39 58 AM

cpathipa (approved):

jaalah-akamai (approved):

packages/manager/src/App.tsx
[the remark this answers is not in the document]
hkhalil-akamai: I suppose the two are equivalent but I like tightly coupling the event with the associated store when the event is emitted.

dwiley-akamai (approved):
Nice work 🚀

PR #8957, opened 2023-04-04
fix: [M3-6251] - Domain Records not updating when navigating

bnussman-akamai:
## Description 📝

- Fixes Domain Records bug and does some additional clean up and updating of our React Query patterns for Domains

### The Bug 🦟

- Domain records were not updating in certain conditions because of how we called `refreshDomainRecords` in a `useEffect`
- In the video, notice that when switching between two domains with search, the records do not update when switching back to `banks-test.com`

https://user-images.githubusercontent.com/115251059/229861500-3803bb04-8921-4c90-b9dc-3189120741e7.mov

## The Fix 🔧
- Use React Query to fetch domain records so we don't have to rely on a `useEffect`

## How to test 🧪

- Have two domains on your account
- Setup some records in each domain
- Switch between the two domains using the search bar
- Verify that the Domains Details page show the correct records for the domain when switching between the two domains via search **and** the browser navigation buttons


mjac0bs (approved):
Nice clean up! 🧼 Confirmed that domain records are correctly updating when switching between multiple domains via search and browser name.

PR #8958, opened 2023-04-04
chore: Repo Clean Up

bnussman-akamai:
## Description 📝

- Removes old “Import Groups as Tags” Feature
  - This is an old feature that was created to help customers switching from Classic Manager to Cloud Manager that allowed customers to convert their `group`s to `tags` on Linode entities. Groups have been deprecated for a very long time so this feature is no longer needed
- Removes regions redux files
  - Regions has been in React Query for a while so we don't need this Redux stuff
- Removes `packages/manager/src/utilities/isProdApi.ts` which will resolve https://github.com/linode/manager/security/code-scanning/21

## How to test 🧪

- Check general functionality of Cloud Manager


cpathipa (approved):
LGTM!

PR #8964, opened 2023-04-04
refactor: [M3-5188] - React Query for NodeBalancers

bnussman-akamai:
## Description 📝

- Modernizes and React Query-ifys most NodeBalancer features of Cloud Manager

## Improvements ✨
- Improves the UI of the NodeBalancer Create page
  - It now matches the Linode Create Page

## Bug Fixes 🐛
- Clicking on a port configuration from the NodeBalancers landing page now brings to the details page **with the click config open**

## How to test 🧪

- Test all NodeBalancer create, update, and delete actions


packages/manager/src/features/NodeBalancers/NodeBalancerDetail/NodeBalancerSettings.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  React.useEffect(() => {
+    if (label !== nodebalancer?.label) {
+      setLabel(nodebalancer?.label);
+    }
+    if (connectionThrottle !== nodebalancer?.client_conn_throttle) {
+      setConnectionThrottle(nodebalancer?.client_conn_throttle);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [nodebalancer]);
```
jaalah-akamai, line 62: Instead of disabling the deps, we can wrap `setLabel` and `setConnectionThrottle` within a `nodebalancer` block to avoid having to include it in the dependency array.
```
React.useEffect(() => {
  if (nodebalancer) {
    setLabel((currentLabel) =>
      currentLabel !== nodebalancer.label ? nodebalancer.label : currentLabel
    );
    setConnectionThrottle((currentThrottle) =>
      currentThrottle !== nodebalancer.client_conn_throttle
        ? nodebalancer.client_conn_throttle
        : currentThrottle
    );
  }
}, [nodebalancer]);
```

packages/manager/src/features/NodeBalancers/NodeBalancerDetail/NodeBalancerSettings.tsx
```diff
@@ -0,0 +1,150 @@
...
+        open={isDeleteDialogOpen}
+        onClose={() => setIsDeleteDialogOpen(false)}
+      />
+    </div>
+  );
+};
+
+const controlClientConnectionThrottle = compose(clamp(0, 20), (value: string) =>
+  defaultNumeric(0, value)
+);
```
jaalah-akamai, line 148: Up to you, but let's just use lodash `defaultTo` and `clamp` as we're trying to get rid of ramda.

jaalah-akamai (commented):
Just a few minor changes, overall it's good and we'll refactor more later 👍

packages/manager/src/features/NodeBalancers/NodeBalancerDetail/NodeBalancerSettings.tsx
jaalah-akamai: I think something like this might work, but you'll have to check:
```
const controlClientConnectionThrottle = (value: string) => {
  const numericValue = defaultTo(0, Number(value));
  return clamp(numericValue, 0, 20);
}
```

packages/manager/src/features/NodeBalancers/NodeBalancerDetail/NodeBalancerSettings.tsx
bnussman-akamai: Won't this cause extra renders if `label` and `client_conn_throttle` don't change?

packages/manager/src/features/NodeBalancers/NodeBalancerDetail/NodeBalancerSettings.tsx
bnussman-akamai: I went ahead and removed the clamping because it was already broken in prod and I think an error message provides the user with more clear expectations

packages/manager/src/features/NodeBalancers/NodeBalancerDetail/NodeBalancerSettings.tsx
jaalah-akamai: Yea I'm just looking at ways to not disable the deps. In that case, I don't see why we wouldn't just add `[nodebalancer, label, connectionThrottle]` to the dependency array, I don't think that's going to cause unnecessary re-renders in this case - what are your thoughts?

bnussman-akamai: I could rewrite this section to use formik or something if we don't want to have to concern ourselves with keeping the state up to date. What do you think?

packages/manager/src/features/Search/withStoreSearch.tsx
```diff
@@ -67,28 +70,20 @@ export default () => (Component: React.ComponentType<any>) => {
...
-  const connected = connect((state: ApplicationState) => {
-    return {
-      entities: getSearchEntities(state.__resources),
-      entitiesLoading: entitiesLoading(state.__resources),
-      errors: entitiesErrors(state.__resources),
-    };
-  });
-
   return compose<SearchProps, {}>(
-    connected,
```
hkhalil-akamai, line 79: Removing this HOC is causing search to crash since the `SearchProps` interface is no longer fully implemented.

A couple of thoughts here:
- all usages of this HOC are FC's, it would be nice if this HOC could be converted to a hook
- recompose's `compose` utility, which we use a lot throughout Manager, really sucks from a type-safety perspective; it doesn't enforce the interface at all and takes the type parameter at face value
    - ramda's `compose` does offer type-safety for up to 6 compositions. If we don't want to use ramda, we can implement this ourselves
    - a lot of our HOCs don't have the correct signature and would make type-safety require a lot of refactoring which may not be worth it if we're transitioning to functional components

hkhalil-akamai (changes requested):
Left some minor feedback and one regression. Overall, changes look good.

packages/manager/src/features/Search/withStoreSearch.tsx
bnussman-akamai: Wow my fault. Search is such a dumpster fire in this repo. I think I resolved this crash. Can you double check behind me?

packages/manager/src/features/NodeBalancers/NodeBalancerDetail/NodeBalancerSettings.tsx
jaalah-akamai: haha understood

packages/manager/src/features/Search/withStoreSearch.tsx
hkhalil-akamai: Yeah, this seems to fix it.

Agreed that search needs some overhauling--if we expect significant feature development that would affect it. Part of that could be turning HOCs like these into hooks.

hkhalil-akamai (approved):
LGTM!

PR #8965, opened 2023-04-05
refactor: Un-Redux-ify Object Storage Drawer

bnussman-akamai:
## Description 📝

- Removes Object Storage Bucket Create Drawer form Redux because it created unnecessary complexity in a simple flow

## How to test 🧪

- Test Creating an Object Storage Bucket
- Test deep linking of `http://localhost:3000/object-storage/buckets/create` (drawer should be open)
- Test enabling object storage for the first time (I used the MSW to mock this)


packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -910,12 +910,12 @@ export const handlers = [
...
         managed: true,
         network_helper: true,
-        object_storage: 'active',
+        object_storage: 'disabled',
       })
     );
   }),
   rest.put('*/account/settings/*', (req, res, ctx) => {
-    return res(ctx.json({}));
+    return res(ctx.json(req.body as any));
```
jaalah-akamai, line 918: I wonder if we can satisfy this type by doing:
```
type Payload = {
  entities: Entity[];
};
const payload: Payload = req.body;
return res(ctx.json(payload));
```

bnussman-akamai: Probably a good idea but I think this may be best if we handled all of the types in this file when we do a clean up of it.

jaalah-akamai (approved):

packages/manager/src/queries/objectStorage.ts
```diff
@@ -84,6 +85,8 @@ export const useCreateBucketMutation = () => {
     ObjectStorageBucketRequestPayload
   >(createBucket, {
     onSuccess: (newEntity) => {
+      // Invalidate account settings becuae it contains obj information
```
dwiley-akamai, line 88: "because"

hana-linode (approved):

dwiley-akamai (changes requested):
We lost some logic that prevented the creation of buckets with non-unique names.

To observe this, try to create a bucket with the same name as one you already have, in the same region. In prod, we prevent that with a field error; here, it goes through and appears successful, but if you refresh the page, the new bucket(s) disappear.

The [API docs](https://www.linode.com/docs/api/object-storage/#object-storage-bucket-create__request-body-schema) suggest that an error should be returned in this situation too, but that doesn't seem to be the case at the moment 🤔

bnussman-akamai:
@dwiley-akamai  The API docs actually explain the weird behavior

> If the bucket already exists and is owned by you, this endpoint returns a 200 response with that bucket as if it had just been created.

I'll work on adding this validation back. Really wish the API work take care of it tho

dwiley-akamai:
> I'll work on adding this validation back. Really wish the API work take care of it tho

ah, I see that part in the docs now. Odd behavior. I think it's worth asking them about possibly revisiting that

dwiley-akamai (approved):
The dupe label validation is working 👌  in prod it only appears once you click "Create Bucket," but I don't mind that it appears upon input here

jdamore-linode (approved):

PR #8967, opened 2023-04-05
fix: [M3-5970] - Fix alignment issues with Node Pool table and buttons

carrillo-erik:
## Description 📝

Fixes alignment and spacing issues with Node Pools table and action buttons in table header component. The initial ask for this ticket was to fix the alignment for the Delete Cluster button. However, after investigation it was deemed an implementation flaw since the button is styled to look like a link, giving it a white background and causing the text to look misaligned. 

## Preview 📷

**Before:**
a
b

**After:**
c
d

## How to test 🧪
Navigate to the Kubernetes Summary view page and inspect the Node Pools table component. Verify that the Add A Node Pool button does not have any extra margin on its right hand side.

Next, reduce the viewport width and inspect that the row elements wrap around correctly and align with each other. Also, notice the removal of the extra white space around the table component.


bnussman-akamai (commented):
I think it would be great if we could fix the placement of these buttons as part of this PR. We never intended for them to be right aligned so weirdly

OMG

jaalah-akamai:
> I think it would be great if we could fix the placement of these buttons as part of this PR. We never intended for them to be right aligned so weirdly
> 
> OMG

If that's the case, let's do it! Wish we had the old design comps >.<

dwiley-akamai (approved):
Thanks for the adjustments 🚢

bnussman-akamai:
The UI in general still looks odd to me.

Screenshot 2023-04-10 at 3 40 29 PM
Screenshot 2023-04-10 at 3 43 03 PM

bnussman-akamai (approved):
After looking around the app a bit more, I really think we should left align `Node Pools` without the padding.

Screenshot 2023-04-11 at 11 44 19 AM

carrillo-erik:
@bnussman-akamai I've changed the left margin to align it with the `LandingHeader`.

PR #8972, opened 2023-04-07
fix: NodeBalancer Create e2e tests and associated bug

bnussman-akamai:
## Description 📝

- Caused by https://github.com/linode/manager/pull/8964
- Fixes e2e test failure caused by updates
- Removes some render guards because the equals function that the render guard is using does not work as expected

## How to test 🧪

- Look at cypress results of this PR
  - If the nodebal spec passes, we can assume I fixed the bug and the e2e test

### DIFF
--- a/packages/manager/cypress/e2e/nodebalancers/smoke-create-nodebal.spec.ts
+++ b/packages/manager/cypress/e2e/nodebalancers/smoke-create-nodebal.spec.ts
@@ -13,11 +13,10 @@ import { randomLabel } from 'support/util/random';
 
 const deployNodeBalancer = () => {
   // This is not an error, the tag is deploy-linode
-  cy.get('[data-qa-deploy-linode]').click();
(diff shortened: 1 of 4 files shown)


packages/manager/cypress/e2e/nodebalancers/smoke-create-nodebal.spec.ts
```diff
@@ -13,11 +13,10 @@ import { randomLabel } from 'support/util/random';
 
 const deployNodeBalancer = () => {
   // This is not an error, the tag is deploy-linode
```
hkhalil-akamai, line 15:
```suggestion
```

hkhalil-akamai (approved):

dwiley-akamai (approved):
Confirmed the nodebal test passed and Nodebal creation still looks good 🚢

PR #8976, opened 2023-04-07
feat: [M3-6450] - Add icon to images not cloud-init compatible

mjac0bs:
## Description 📝

**What does this PR do?**
Adds a cloud-init-incompatible icon and tooltip to designate which of a user's custom images are not compatible with cloud-init.

## Preview 📷

| Light Mode  | Dark Mode  |   
|---|---|
| Screenshot 2023-04-10 at 7 13 50 AM  | Screenshot 2023-04-10 at 7 14 13 AM  |   

## How to test 🧪

- Use a dev test account that is configured for metadata.
- Turn on the MSW _or_ create/have at least one custom image marked as cloud-init compatible and one that is not marked as cloud-init compatible. (cloud-init compatible images will have `capabilites["cloud-init"]`)
- Go to http://localhost:3000/linodes/create?type=Images and open the image select dropdown.
- Verify that cloud-init compatible images do not have the tooltip icon, and cloud-init incompatible images do.
- Go to http://localhost:3000/linodes/create?type=Distributions and, with the MSW off, verify that all distros except for distros marked cloud-init compatible (currently: Ubuntu 22.10 - CloudInit Special) display the tooltip icon.
- Check both light & dark mode and selected & focus state to make sure icon displays as expected (correct color, etc.).


packages/manager/src/components/ImageSelect/ImageOption.tsx
```diff
@@ -35,15 +50,16 @@ interface ImageOptionProps extends OptionProps<any, any> {
 
 type CombinedProps = ImageOptionProps;
 
-export const ImageOption: React.FC<CombinedProps> = (props) => {
+const ImageOption = (props: CombinedProps) => {
```
bnussman-akamai, line 53:
```suggestion
const ImageOption = (props: ImageOptionProps) => {
```

packages/manager/src/components/TooltipIcon/TooltipIcon.tsx
bnussman-akamai, line 1: Should we make it so the `icon` prop overrides the icon no matter what the `status` is?

mjac0bs: Do you mean rendering `icon` rather than `renderIcon` when the `icon` prop exists, rather than assigning `icon` to `renderIcon` when checking the `status`? I could do that; it would just make the purpose of the status of "other" less clear .

packages/manager/src/components/ImageSelect/ImageSelect.tsx
```diff
@@ -96,10 +104,14 @@ export const imagesToGroupedItems = (images: Image[]) => {
...
                   created,
-                  // Add suffix 'depricated' to the image at end of life.
+                  // Add suffix 'deprecated' to the image at end of life.
                   label:
                     differenceInMonths > 0 ? `${label} (deprecated)` : label,
                   value: id,
+                  isCloudInitCompatible:
+                    thisGroup === 'My Images'
+                      ? capabilities?.includes('cloud-init')
+                      : true, // Default to cloud-init compatible for distro images.
```
hana-linode, line 114: We should also check the distro images instead of defaulting to true. Since at launch, API expects about half a dozen regions to not have metadata bc they won't be supporting legacy networks

mjac0bs: Thanks for the additional context -- made that change in [c21fb39](https://github.com/linode/manager/pull/8976/commits/c21fb39d952d4b2dbea087babdd0e4af0ec907be).

packages/manager/src/components/ImageSelect/ImageOption.tsx
```diff
@@ -57,11 +71,27 @@ export const ImageOption: React.FC<CombinedProps> = (props) => {
           justifyContent: 'flex-start',
         }}
       >
-        <span className={`${props.data.className} ${classes.icon}`} />
+        <span className={`${data.className} ${classes.distroIcon}`} />
         <Box>{label}</Box>
+        {!data.isCloudInitCompatible ? (
```
hana-linode, line 76: Can we add a check for flags.metadata?

mjac0bs: Yeah, good call! Done in [639fa63](https://github.com/linode/manager/pull/8976/commits/639fa638d818db5f5af6a25dd25771843750e596).

hana-linode (approved):

cpathipa (approved):
The functionality looks good to me! It will be beneficial to show a visual indication in the Create Linode flow. I also think that we should bring a similar visual indication for the clone and backup flows. Has there been any discussion with UX regarding this?

dwiley-akamai (approved):
🚀

PR #8979, opened 2023-04-10
chore: Clean Up Repo Root

bnussman-akamai:
### Why?
- I'm trying to clean up the root of our repo because it is bloated with lots of random config files
- We have a nice new readme and this will make it easier to see without needing to scroll on Github

## Overview
- Consolidates two eslint files into one
- Deletes `.npmignore` because it is empty and I'm 99% sure it is doing nothing meaningful
- Deletes old changelog tooling because we don't use it and haven't used it in years

## How to test 🧪

- Not much to test here but you can test the eslint still works
  - One way to do this is to create a duplicate variable  in a random component and ensure that your IDE shows an eslint warning


jdamore-linode:
I'm thinking the empty `.npmignore` may have been put there deliberately to [prevent npm from ignoring files that match our `.gitignore` patterns](https://docs.npmjs.com/cli/v7/using-npm/developers#keeping-files-out-of-your-package:~:text=If%20you%20want%20to%20include%20something%20that%20is%20excluded%20by%20your%20.gitignore%20file%2C%20you%20can%20create%20an%20empty%20.npmignore%20file%20to%20override%20it.)

bnussman-akamai:
I don't think it matters because api-v4 and validation's package.jsons specify `"files": [ "lib" ]` so when it comes to shipping those packages, they should pull only those files.

jaalah-akamai (approved):

PR #8983, opened 2023-04-11
fix: async await console erros in unit test

cpathipa:
## Description 📝
Resolves async await console erros in unit test

image

## How to test 🧪

- yarn test useAPIRequest.test.ts


bnussman-akamai (approved):
Nice, much cleaner. Is this change okay with Vitest @jdamore-linode?

jdamore-linode (approved):
> Nice, much cleaner. Is this change okay with Vitest @jdamore-linode?

Yeah this is totally fine! I think the Vitest branch will fix most (all?) of the remaining instances of these errors but it's great to get these cleaned up in the meantime too 🎉

PR #8986, opened 2023-04-11
refactor: [M3-6478] - Make `<Select />` type safe

bnussman-akamai:
## Why ❓
- You may have not noticed, but the Select component had basically **zero type-safety** resulting in very bad developer experience and opens the possibility for bugs. It has been like this for **years** 😥
- Refer to the before and after

## Description 
- Does a refactor / improvement of the "Enhanced" Select component
- The goal isn't to make the component perfect (because it is super messed up), but the goal is to make it better 
- Putting it up as a draft to get e2e test feedback and feedback from team

## Impact 📷

| Before 👎🏼  | After 🎉 |
|--------|-------|
| Screenshot 2023-04-11 at 8 47 10 PM      | Screenshot 2023-04-11 at 8 46 17 PM     |

| Before 👎🏼  | After 🎉 |
|--------|-------|
| Screenshot 2023-04-11 at 9 01 57 PM      | Screenshot 2023-04-11 at 9 03 07 PM     |

### Before 👎🏼  
https://user-images.githubusercontent.com/115251059/231321727-f9d26d44-9214-4a99-a7d1-cb8c673eba50.mov

### After 🎉
https://user-images.githubusercontent.com/115251059/231321739-e72bc643-febb-434e-9b51-ecc55303fba7.mov

## How to test 🧪

- Test places in the app that use a "Enhanced" Select Component
  - I recommend looking at the diff and going to each of the `<Select />` components that were modified and testing the functionality.
- The e2e tests (once they are working) should also help check for bugs


bnussman-akamai:
e2e tests results 9e2b98d

Screenshot 2023-04-11 at 10 35 55 PM

bnussman-akamai:
Little bonus feature worth mentioning. If the select is `clearable`, the `onChange` type updates dynamically ✨

If a select is `clearable`, the `onChange` value could be null (the user hits the clear button)
If a select is **NOT** `clearable`, the `onChange` value will always be a defined item. 

The types now properly reflect this 🎉 Notice how in the video the type of `item` is different depending on the `clearable` prop

https://user-images.githubusercontent.com/115251059/231343809-12f3d0fc-6b82-4fe3-95a6-d1f86e0d6f3e.mov

packages/manager/src/features/Images/ImageSelect.tsx
```diff
@@ -26,19 +26,31 @@ interface Props {
...
+  imageFieldError?: string;
+  isMulti: true;
+  helperText?: string;
+  value?: Item[];
+  disabled?: boolean;
+  onSelect: (selected: Item[]) => void;
+  label?: string;
+  required?: boolean;
+  anyAllOption?: boolean;
+}
```
jaalah-akamai, line 51: Can we just combine this into one interface?
```
interface Props {
  images: Image[];
  imageError?: string;
  imageFieldError?: string;
  helperText?: string;
  value?: Item | Item[];
  disabled?: boolean;
  onSelect: (selected: Item | Item[]) => void;
  label?: string;
  required?: boolean;
  anyAllOption?: boolean;
  isMulti?: boolean;
}
```

packages/manager/src/features/Profile/AuthenticationSettings/PhoneVerification/PhoneVerification.tsx
```diff
@@ -164,7 +164,7 @@ export const PhoneVerification = () => {
     }),
     singleValue: (provided: React.CSSProperties) => ({
       ...provided,
-      textAlign: 'center',
+      textAlign: 'center' as React.CSSProperties['textAlign'],
```
jaalah-akamai, line 167: I think you can just do `as const` for simplicity, but this works too

jaalah-akamai (approved):

packages/manager/src/features/Images/ImageSelect.tsx
bnussman-akamai: No because we are using a discriminated union to make `value` and `onChange` change depending on `isMulti` but I can still clean it up.

packages/manager/src/features/Profile/AuthenticationSettings/PhoneVerification/PhoneVerification.tsx
bnussman-akamai: Switch to as const in 9f687b5

packages/manager/src/features/Images/ImageSelect.tsx
bnussman-akamai: Cleaned up in 9f687b5

cpathipa (commented):
Nice work!

cpathipa (approved):

PR #8989, opened 2023-04-12
chore: [M3-6462] - Migrate GA to AA

mjac0bs:
## Description 📝

**What does this PR do?**
Internal testing - ignore for now.

## Preview 📷

## How to test 🧪


packages/manager/src/App.tsx
```diff
@@ -83,6 +84,15 @@ export class App extends React.Component<CombinedProps, State> {
...
     }
 
+    // Load Adobe Analytics Launch Script
+    if (!!ADOBE_ANALYTICS_URL) {
+      const script = document.createElement('script');
+      // eslint-disable-next-line scanjs-rules/assign_to_src
+      script.src = `${ADOBE_ANALYTICS_URL}`;
+      script.async = true;
+      document.head.appendChild(script);
+    }
```
bnussman-akamai, line 94: Did you try the `loadScript`function for this?

mjac0bs: I'll give it another look, but I think the issue was that `loadScript` appends the script to the document `body` and Adobe wants this script in the document `head`.

hana-linode (approved):
Adobe Launch script loaded and console logs are visible after following the test steps 🚀 

image

image

Can we also add `REACT_APP_ADOBE_ANALYTICS_URL` to the `.env.example` file?

dwiley-akamai (approved):
I am seeing the expected outputs logged to the console 👍

packages/manager/src/App.tsx
mjac0bs: I think we could use the existing `loadScript` function, but modify it to accept an optional `options` argument that contains `setStatus` and `location`, where `location` is then either the "head" or (default) "body".

dwiley-akamai (approved):

hana-linode (approved):

PR #8990, opened 2023-04-12
refactor: [M3-6418] - MUI v5 Migration - Components > Toggle

bnussman-akamai:
## Description 📝

- I don't even know if this counts as a react tss migration but a ticket existed for the `<Toggle />` but there are no styles applied to it so I just did a clean up and modernization of the component

## How to test 🧪

- Find places in the app that use toggles, verify the toggles still look normal and function normally


jaalah-akamai (approved):

packages/manager/src/components/Toggle/index.ts
```diff
@@ -1,2 +1 @@
-import Toggle from './Toggle';
-export default Toggle;
+export { Toggle, ToggleProps } from './Toggle';
```
hana-linode, line 1: Exporting ToggleProps like this displays an error in the console and the app doesn't render. I think this is why we import and export interfaces with underscore in other components

image

bnussman-akamai: Oh interesting, I'm gonna fix it by separating the exports

```ts
export { Toggle } from './Toggle';
export type { ToggleProps } from './Toggle';
```

bnussman-akamai: Another reason I don't want to keep the index files around 😖

bnussman-akamai: Thanks for catching this @hana-linode

PR #8982, opened 2023-04-11
chore: [M3-6186] - Update logo in TPA Providers screen

cpathipa:
## Description 📝
Updates language and icon in login authentication tab.

## Preview 📷
image

**What are the steps to reproduce the issue or verify the changes?**

- Navigate to http://localhost:3000/profile/auth
- Validate the TPA provider (Logo) as per the design mockup. 

**How do I run relevant unit or e2e tests?**

- yarn test
- yarn cy:run

PR #8991, opened 2023-04-12
fix: `smoke-deep-link.spec.ts` after copy changes

bnussman-akamai:
## Description 📝

- Fixes `smoke-deep-link.spec.ts` after copy changes caused by https://github.com/linode/manager/pull/8982

## How to test 🧪

- Run `yarn cy:debug` and run the `smoke-deep-link.spec.ts` test

### DIFF
--- a/packages/manager/cypress/support/ui/constants.ts
+++ b/packages/manager/cypress/support/ui/constants.ts
@@ -115,7 +115,7 @@ export const pages = [
           loadAppNoLogin(url);
           getVisible('[data-qa-header="My Profile"]');
           containsVisible(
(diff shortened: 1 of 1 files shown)


mjac0bs (approved):
This looked good when I ran it. Thanks for fixing! I remember running some of the account/ related E2Es when reviewing that PR, wondering if copy changes would break something. Didn't think to check this file.

image

jdamore-linode (approved):

jdamore-linode:
We can ignore the failed check; segfault during Jest stage :/

PR #8995, opened 2023-04-12
refactor: [M3-6414, M3-6415] - MUIv5 Migration - Components > TagsInput, TagsPanel

hana-linode:
## Description 📝
- Migrate/modernize TagsPanel component from JSS to tss-react (Emotion)
- Modernize TagsInput since it had no styles to migrate
- Use named exports instead of default
- Fix BetaChip story not rendering

## How to test 🧪
- Check places in the app where TagsInput and TagsPanel are used
- Check that the TagsInput, TagsPanel, and BetaChip story renders in Storybook


packages/manager/src/components/TagsInput/TagsInput.tsx
```diff
@@ -26,7 +26,7 @@ export interface Props {
   menuPlacement?: 'bottom' | 'top' | 'auto' | undefined;
```
mjac0bs, line 26: Can we remove `undefined` here since this is already an optional prop?

mjac0bs (approved):
No regressions that I could see. 🚢  And thanks for fixing the stories.

I have noticed that the BetaChip story is messed up somehow (actually, I think the component might be) because the chip in dark mode is not any different from different from light mode. And toggling the `color` in the Component API has no effect. But that's pretty far out of the intended scope this PR. 

Screenshot 2023-04-12 at 2 48 15 PM

packages/manager/src/components/TagsInput/index.ts
```diff
@@ -1,5 +1,8 @@
...
-export interface Tag extends _Tag {}
-export default TagsInput;
+import {
+  TagsInput,
+  TagsInputProps as _TagsInputProps,
+  Tag as _Tag,
+} from './TagsInput';
+export type TagsInputProps = _TagsInputProps;
+export type Tag = _Tag;
+export { TagsInput };
```
jaalah-akamai, line 8: Back to what we discussed recently, what's the lift of removing this index file? Might be worth looking into that as well.

bnussman-akamai (approved):

mjac0bs:
@hana-linode Let's add a Tech Stories section and changelog entry before this is merged in!

PR #8999, opened 2023-04-13
feat: [M3-6427] - Add No Results section for Marketplace Search

cpathipa:
## Description 📝
Add No Results section for Marketplace Search

## Preview 📷
image

## How to test 🧪

- Navigate to Marketplace search
- Search for app that doesn't exists (ex; datadog)
- Should display "No Results" message. 

- yarn test
- yarn cy:run


jdamore-linode:
Nice, it works well!

Would it be possible to center the "No Results" text?

Screen Shot 2023-04-13 at 4 11 29 PM

Do we need any other input from the UX team as far as wording and design should go?

mjac0bs (commented):
I'm glad we're doing this! Seconding @jdamore-linode's request to center the text horizontally. And I'm wondering about the overall placement within that container. 

Mock (assume actual center alignment):
Screenshot 2023-04-13 at 3 09 00 PM

cpathipa:
Thank you @jdamore-linode and @mjac0bs thank you for the feedback, I will circle back with UX on this.

dwiley-akamai (commented):
I also agree with checking on the centered-within-container option

cpathipa:
Thank you for the feedback. Based on the UX feedback, updated the no results copy and adjusted alignment to center. 
image

hana-linode (commented):
Thoughts on left aligning the no results section? The selection cards are left aligned as well as the "no compatible images" text below

image

mjac0bs:
Nitpicks on copy: I'm wondering if clearer copy might read: "Sorry, no results matched your search for "datadog". Please try another term."

Thought process being:
1. "no <search term> results" sounds kind of strange and is slightly harder to scan as a reader than "no results"
2. telling the user to "please try again" feels misleading because if the user were to try the same action again, the results would be the same. It feels more clear to state what they should actually try to yield different results.

What do others (and UX) think?

hana-linode:
@mjac0bs Agreed that "Sorry, no results matched your search for "datadog" sounds clearer. I don't think we even need the second sentence as it's pretty obvious that the user should search for a different term

packages/manager/src/features/linodes/LinodesCreate/AppPanelSection.tsx
```diff
@@ -46,38 +48,46 @@ export const AppPanelSection: React.FC<Props> = (props) => {
...
+              />
+            );
+          })
+        ) : (
+          <Grid container xs={12} columnSpacing={0}>
+            <Typography>
+              {`Sorry, no results matching "${searchValue}" were found.`}
+            </Typography>
+          </Grid>
+        )}
```
jaalah-akamai, line 90: I think we can simplify the JSX a little here by doing:
```
{apps.length > 0 ? (
  <AppPanelGrid container spacing={2}>
    {apps.map((eachApp) => { ... })}
  </AppPanelGrid>
) : (
  <Typography>
    {`Sorry, no results matching "${searchValue}" were found.`}
  </Typography>
)}
```
We won't need the extra Grid in this case

jaalah-akamai (changes requested):

jaalah-akamai:
Let's remember to update changelog too 👍

jaalah-akamai (approved):

mjac0bs (approved):
Just a changelog update and this looks like it's good to go.

PR #9001, opened 2023-04-13
refactor: [M3-6417] - SRC > Components > Tile

cpathipa:
## Description 📝

- Migrate Tile component from JSS to tss-react
- Use named exports instead of default

## How to test 🧪

- Check places in the app where Tile are used
- Check that the Tile story renders in Storybook


bnussman-akamai (approved):
Cloud and Storybook both matched prod!

abailly-akamai (approved):
👍 also compared storybook local & prod

PR #9002, opened 2023-04-14
refactor: [M3-6304] - MUI v5 Migration - `ConditionalWrapper`

bnussman-akamai:
## Description 📝

- Deletes `<ConditionalWrapper />` because it was a bit extra and only consumed once
- Cleans up components that consume it
- Fixes some styles that were broken when we migrated from MUIv4 to MUIv5

## How to test 🧪

- The main component this affects is Docs links
  - Verify that docs links look okay


packages/manager/src/components/IconTextLink/IconTextLink.tsx
```diff
@@ -12,7 +10,7 @@ const useStyles = makeStyles()((theme: Theme) => ({
     display: 'flex',
     alignItems: 'flex-start',
     cursor: 'pointer',
-    padding: theme.spacing(1) + theme.spacing(0.5),
```
bnussman-akamai, line 15: This was probably left over from MUI v4. `theme.spacing()` use to return a `number`, now it returns a `string` so this line actually caused string concatenation which is **not** what we want. This might result in some minor UI changes, but this restores the original intention.

packages/manager/src/components/IconTextLink/IconTextLink.tsx
```diff
@@ -48,7 +46,7 @@ const useStyles = makeStyles()((theme: Theme) => ({
     },
   },
   left: {
-    left: -(theme.spacing(1) + theme.spacing(0.5)),
```
bnussman-akamai, line 51: Same as other comment. string concatenation here was probably not intended

dwiley-akamai (approved):
The Docs button appears to have drifted a few pixels to the left, but this LGTM otherwise

jaalah-akamai (approved):

PR #9004, opened 2023-04-14
chore: [M3-6496] - Add Adobe Analytics custom event tracking

mjac0bs:
## Description 📝

**What does this PR do?**
This PR enables custom event tracking in Adobe Analytics using a Direct Call rule and the Adobe [`_satellite`](https://experienceleague.adobe.com/docs/experience-platform/tags/client-side/satellite-object.html?lang=en) object. 

Everything but the changes in `packages/manager/src/utilities/ga.ts` will be merged before this PR in step 1: #8989, which sets up basic Adobe page view tracking on Cloud Manager. Step 3, removal of Google Analytics, will come in a later PR.

This PR is marked as Do Not Merge for now to roll out the analytics migration in phases. It can still be reviewed but should not be merged until **after** 4/26 (date next release branch is cut) and is scheduled for the **5/15** release. I will update this PR if this timeline changes.

## How to test 🧪

1.  Check out this PR locally.
2. Test that there are no unexpected errors when running Cloud Manager prior to defining an Adobe Analytics env var for local testing.
3. Set the `REACT_APP_ADOBE_ANALYTICS_URL` environment variable in `.env`. Reach out if you don't have this.
4. Open the browser tools Network tab, filter requests by "adobe", and check that successful network requests have been made to load the launch script.
5. In the browser console, type `_satellite.setDebug(true)`. 
6. Refresh the page. You should see something like the following, indicating that Adobe Analytics is integrated into the website and is performing basic tracking of page views:
image
7. Test out a handful of custom events defined in `packages/manager/src/utilities/ga.ts`. After triggering each event, check that the Adobe Analytics debug log in the browser console has fired a corresponding direct call custom event with the additional details (category, action, label [if applicable], and value [if applicable]). For example, `sendCreateVolumeEvent` triggers the following direct call rule when the `sendEvent` function is called on "Create Volume" button click:
image
9. Verify that the E2E tests pass without any unexpected error.

``` 
yarn cy:run
```


packages/manager/src/utilities/ga.ts
```diff
@@ -12,6 +12,20 @@ interface AnalyticsEvent {
  * Will throw error unless analytics is initialized
  */
 export const sendEvent = (eventPayload: AnalyticsEvent): void => {
+  // Send a Direct Call Rule if our environment is configured with an Adobe Launch script
+  if (!!ADOBE_ANALYTICS_URL) {
+    try {
```
jaalah-akamai, line 17: So this try block isn't really doing anything useful; It's just silently discarding any errors. What if returned a console error if `_satellite` and `track` are not defined.
```
if (!ADOBE_ANALYTICS_URL) {
  return;
}

if (!window._satellite || !window._satellite.track) {
  console.error('Adobe Launch script not loaded');
  return;
}

// Tracking Event ...
```

mjac0bs: (Looking into logging this error to Sentry, rather than the console.)

dwiley-akamai (commented):
This is looking good to me; once that error gets logged in Sentry and your other PR gets merged, this should be good to go after a rebase/merge

packages/manager/src/utilities/ga.ts
mjac0bs: This is now set up ([e39a7f1](https://github.com/linode/manager/pull/9004/commits/e39a7f119e8eaf5f90d3d52c9e23db6d02b63fa9)) to log the exception either to Sentry (in production envs) or to the console (in local development with the value for `REACT_APP_SENTRY_URL` also set in `.env`) if the `ADOBE_ANALYTICS_URL` *is* defined but the direct call rule fails. This would indicate that something is wrong with the Adobe launch script because the `_satellite` object and its `track` function should be accessible via `window` once the script has loaded.

Without `ADOBE_ANALYTICS_URL` env defined, `sendEvent` will immediately return, so users developing locally without the analytics configuration - or the e2e tests - should not see crashes or errors.

image

hkhalil-akamai (approved):
Tested, LGTM!

bnussman-akamai (approved):

PR #9009, opened 2023-04-17
fix: [M3-6480] - Fix blank Node Pool plan selection

hkhalil-akamai:
## Description 📝

Fixes a bug introduced in #8840 that causes the `AddNodePool` drawer to not show any plans in certain situations.

Also included are basic unit tests that validate the fix and small amendments to the recent `useQueryClient` PR.

## Preview 📷

| Before      | After |
| ----------- | ----------- |
|Screenshot 2023-04-17 at 1 30 24 PM|Screenshot 2023-04-17 at 1 30 07 PM|

## How to test 🧪

1. Under Kubernetes, navigate to an existing cluster or create a new one.
2. Refresh the page to clean the RQ store.
3. Click on 'Add A Node Pool'.
4. Verify the `AddNodePoolDrawer` appears and displays all plan types and details.


packages/manager/src/features/Kubernetes/CreateCluster/NodePoolPanel.tsx
```diff
@@ -55,11 +46,13 @@ const RenderLoadingOrContent: React.FunctionComponent<CombinedProps> = (
 const Panel: React.FunctionComponent<CombinedProps> = (props) => {
   const { addNodePool, apiError, types, isOnCreate } = props;
 
-  const [_types, setNewType] = React.useState<ExtendedTypeWithCount[]>(
-    addCountToTypes(types)
+  const [typeCountMap, setTypeCountMap] = React.useState<Map<string, number>>(
+    new Map()
```
hkhalil-akamai, line 50: Decided to use a `Map` here instead of an index signature object because those can lead to runtime errors without [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) enabled:

```
const map1: { [key: string]: number } = {"type1": 0};
const type2Count: number = map1["type2"]; # no type error
```

vs with `Map`:

```
const map1: Map<string, number> = new Map([[ "type1", 0 ]]);
const type2Count: number = map1.get("type2"); # type error: number | undefined is not assignable to number.
```

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/NodePoolsDisplay/AddNodePoolDrawer.tsx
```diff
@@ -109,39 +110,28 @@ export const AddNodePoolDrawer = (props: Props) => {
...
-      return { ...thisType, count: 0 };
-    });
-    setNewType(newTypes);
-    // If everything's empty, we need to reset the selected type.
-    if (newTypes.every((thisType) => thisType.count === 0)) {
-      setSelectedType(undefined);
-    } else {
-      setSelectedType(planId);
-    }
+    setSelectedTypeInfo({ planId, count: newCount });
```
bnussman-akamai, line 124: Would doing this allow us to get rid of the `useEffect` that runs `setSelectedTypeInfo(undefined)`

```suggestion
    if (newCount === 0) {
      setSelectedTypeInfo(undefined);
    } else {
      setSelectedTypeInfo({ planId, count: newCount });
    }
```

bnussman-akamai (changes requested):

bnussman-akamai (approved):
I did not notice any regressions, thank you for for the fix!

PR #9018, opened 2023-04-18
refactor: [M3-6306] - MUI v5 Migration SRC > Components > CopyableTextField

cpathipa:
## Description 📝

- Migrate CopyableTextField component from JSS to tss-react.
- Use named export instead of default

- yarn test
- yarn cy:run


jaalah-akamai:
Since this is a small but important component, what do you think about refactoring this all the way using `styled` and `sx`? We could remove the hard-coded `copy` class along with these overrides: https://github.com/linode/manager/blob/develop/packages/manager/src/themeFactory.ts#L764-L766 and https://github.com/linode/manager/blob/develop/packages/manager/src/themes.ts#L340-L342

The `copy` styles can be placed at the component level IMO - open to suggestions though!

cpathipa:
> Since this is a small but important component, what do you think about refactoring this all the way using `styled` and `sx`? We could remove the hard-coded `copy` class along with these overrides: https://github.com/linode/manager/blob/develop/packages/manager/src/themeFactory.ts#L764-L766 and https://github.com/linode/manager/blob/develop/packages/manager/src/themes.ts#L340-L342
> 
> The `copy` styles can be placed at the component level IMO - open to suggestions though!

That sounds good to me!

packages/manager/src/components/CopyableTextField/CopyableTextField.tsx
bnussman-akamai, line 1: Can we remove `CombinedProps` and just have `Props`?

```typescript
export const CopyableTextField = (props: Props) => {
```

mjac0bs: And rename `Props` to `CopyableTextFieldProps`?

dwiley-akamai (commented):
I agree with all of the changes suggested as well

packages/manager/src/components/CopyableTextField/CopyableTextField.tsx
cpathipa: Renamed to `CopyableTextFieldProps`

bnussman-akamai, line 1: | Before | This PR |
|--------|-------|
| Screenshot 2023-04-24 at 10 13 34 AM       |  Screenshot 2023-04-24 at 10 13 30 AM   |




The styles now look significantly different. I honestly don't mind the new styles.

hana-linode: The contrast isn't great but I think that's expected of disabled styling 🤔 

image

packages/manager/src/components/CopyableTextField/CopyableTextField.tsx
```diff
@@ -1,10 +1,10 @@
...
 import CopyTooltip from 'src/components/CopyTooltip';
-import { makeStyles } from '@mui/styles';
+import { makeStyles } from 'tss-react/mui';
 import { Theme } from '@mui/material/styles';
 import TextField, { Props as TextFieldProps } from 'src/components/TextField';
 
-const useStyles = makeStyles((theme: Theme) => ({
+const useStyles = makeStyles()((theme: Theme) => ({
   removeDisabledStyles: {
     '& .MuiInput-input': {
```
jaalah-akamai, line 9: @cpathipa Just add an `&&` in both the places here and you're good. This is a conflict between Emotion and JSS because the underlying TextField hasn't been updated yet.

jaalah-akamai (approved):

bnussman-akamai (approved):

dwiley-akamai (approved):
LGTM if we're holding off on the full refactor to `styled` & `sx` for this one

dwiley-akamai:
@cpathipa We will need an entry for `CHANGELOG.md` before merging though

jaalah-akamai (approved):
Confirmed changes look much better!

PR #9028, opened 2023-04-18
refactor: [M3-6302] - MUIv5 Migration - Components > CircleProgress

hana-linode:
## Description 📝
- Migrate CircleProgress component from JSS to styled components (Emotion)
- Use named export instead of default

## How to test 🧪
- Check places in the app where the CircleProgress component is used
- Ensure that the CircleProgress story renders in Storybook


CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - Event entities should only be linked for true labels
 
 ### Tech Stories:
+- MUIv5 Migration - Components > CircleProgress #9028
+- MUIv5 Migration - Components > StatusIcon #9014
```
mjac0bs, line 19: Nice catch on the missed changelog update.

packages/manager/src/components/CircleProgress/index.ts
```diff
@@ -1,2 +1,2 @@
-import CircleProgress from './CircleProgress';
-export default CircleProgress;
+import { CircleProgress } from './CircleProgress';
+export { CircleProgress };
```
mjac0bs, line 2: We should be able to write this as one line instead of two:
```suggestion
export { CircleProgress } from './CircleProgress';
```

packages/manager/src/components/CircleProgress/CircleProgress.tsx
```diff
@@ -75,54 +14,108 @@ interface Props extends CircularProgressProps {
...
+const StyledMiniCircularProgress = styled(
+  CircularProgress
+)<CircleProgressProps>(({ theme, ...props }) => ({
+  padding: `calc(${theme.spacing()} * 1.3)`,
+  ...(props.noPadding && {
+    padding: 0,
+  }),
+  ...(props.tag && {
+    width: '12px !important',
+    height: '12px !important',
```
mjac0bs, line 113: Does using `&&` here rather than `!important` work? I thought it seemed to, but might have missed something because I also tried to see if we could avoid `!important` for the `StyledCircularProgress` at the `sm` breakpoint and I wasn't seeing that work.
```suggestion
'&&': {
    width: '12px',
    height: '12px',
    }
```

packages/manager/src/components/CircleProgress/index.ts
bnussman-akamai, line 2:
```suggestion
export { CircleProgress } from './CircleProgress';
```

packages/manager/src/components/ExtendedAccordion/ExtendedAccordion.test.tsx
```diff
@@ -1,7 +1,7 @@
 import { shallow } from 'enzyme';
 import * as React from 'react';
 
-import CircleProgress from 'src/components/CircleProgress';
+import { CircleProgress } from 'src/components/CircleProgress';
```
bnussman-akamai, line 4: Unrelated to your work, but I've never been a fan of the spaces between imports

```suggestion
import { CircleProgress } from 'src/components/CircleProgress';
```

bnussman-akamai (approved):

packages/manager/src/components/CircleProgress/CircleProgress.tsx
hana-linode: @mjac0bs It looks like this style wasn't even being used in https://github.com/linode/manager/blob/develop/packages/manager/src/components/Tag/Tag.tsx#L150-L161 so I went ahead and removed it. Can you check again?

packages/manager/src/components/ExtendedAccordion/ExtendedAccordion.test.tsx
hana-linode: I'm not against it, we add spaces when destructuring objects too

jaalah-akamai (approved):

packages/manager/src/components/CircleProgress/CircleProgress.tsx
mjac0bs: I think you're right. Your changes LGTM.

mjac0bs (approved):

PR #9032, opened 2023-04-19
fix: [M3-6446] - Fix code scanning alert that DOM text is reinterpreted as HTML

bnussman-akamai:
## Description 📝

- Fixes https://github.com/linode/manager/security/code-scanning/23
- Uses Braintree's `sanitize-url` because it already is already baked into our repo
  - I considered `sanitize-html`, which we also have installed, but Braintree's `sanitize-url` seems like a better fit for this use case

## Major Changes 🔄
- The `<ExternalLink />` component now sanitizes any link it is given

## How to test 🧪

- Verify that the `<ExternalLink />` component still works as expected
  - You can test this by clicking links in the Footer
- Verify that `SupportSearchLanding` (http://localhost:3000/support/search/?query=) still works as expected because this is what triggered the alert


jaalah-akamai (approved):
🧼 Nice and clean (no pun intended)

hana-linode (approved):

PR #9033, opened 2023-04-19
feat: [M3-6501] - Add Metadata migrate warning

hana-linode:
## Description 📝
Add warning text to the Migrate Linode dialog when migrating from a Data Center that supports Metadata to a Data Center that doesn't

## Preview 📷
https://user-images.githubusercontent.com/115299789/233193496-a8128f36-c8c9-4fcb-b6eb-b657c4cf8945.mov

## How to test 🧪
- Turn on the MSW and migrate a Linode from a Metadata supported DC (London, UK) to a Non-Metadata supported DC (any other DC).
-  You should see the warning notice update with an additional bullet. Reopening the Dialog should clear the additional bullet
- You should _not_ see the additional bullet when Migrating from a Non-Metadata supported DC to a supported Metadata DC


abailly-akamai:
@hana-linode what is MSW and how to you turn it on?

hana-linode:
@abailly-akamai It's the Mock Service Worker and we can turn it on in the lower left corner of the browser

https://user-images.githubusercontent.com/115299789/233415619-a3480a46-90fb-4908-b724-3db8af9c9b75.mov

packages/manager/src/features/linodes/MigrateLinode/MigrateLinode.tsx
```diff
@@ -116,11 +124,38 @@ const MigrateLinode = (props: Props) => {
...
+    }
+    const currentRegionSupportsMetadata =
+      regionsData
+        ?.find((region) => region.id === linode.region)
+        ?.capabilities.includes('Metadata') ?? false;
+
+    const selectedRegionSupportsMetadata =
+      regionsData
+        ?.find((region) => region.id === selectedRegion)
+        ?.capabilities.includes('Metadata') ?? false;
```
abailly-akamai, line 145: You could make this cleaner like so:
```
const regionSupportsMetadata = (_region: string) =>
    regionsData
      ?.find((region) => region.id === _region)
      ?.capabilities.includes('Metadata') ?? false;

const currentRegionSupportsMetadata = regionSupportsMetadata(linode.region);
const selectedRegionSupportsMetadata = regionSupportsMetadata(
  selectedRegion
);
```

packages/manager/src/features/linodes/MigrateLinode/MigrateLinode.tsx
```diff
@@ -116,11 +124,38 @@ const MigrateLinode = (props: Props) => {
...
+
+    const metadataWarning =
+      currentRegionSupportsMetadata && !selectedRegionSupportsMetadata
+        ? 'The selected Data Center does not support Metadata. If your Linode is rebuilt, it will boot without using any associated User Data.'
+        : undefined;
+
+    if (metadataWarning) {
+      setMetadataMigrateWarning(metadataWarning);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
abailly-akamai, line 155: Not a blocker at all: wondering why disabling this rule since it is not configured in eslint - I added an item in the Front end cafe session to discuss if we want the rule

abailly-akamai (commented):
Looks good, but somehow the new warning is not showing after turning on the MSW and testing the feature (I wonder if I missed step to test this properly?) - also left a couple of non-blocking comments

Screenshot 2023-04-20 at 12 06 46 PM

hana-linode:
@abailly-akamai Do you have the metadata feature flag turned on in the dev tools? 
image

abailly-akamai:
@hana-linode ah yes, that was it. I had it on but did not realize it gets turned off after a page reload - looks good!

abailly-akamai (approved):

mjac0bs (approved):
This LGTM -- observed the warning bullet point's existence and non-existence when migrating as described.

packages/manager/src/features/linodes/MigrateLinode/MigrateLinode.tsx
```diff
@@ -116,11 +124,38 @@ const MigrateLinode = (props: Props) => {
     if (open) {
       setAPIError('');
       setRegionError('');
+      setMetadataMigrateWarning('');
       setConfirmed(false);
       handleSelectRegion(null);
     }
   }, [open]);
 
+  React.useEffect(() => {
```
hkhalil-akamai, line 133: Could we compute `metadataMigrateWarning` in a `useMemo` instead of having a `useEffect` that calls `setMetadataMigrateWarning`?

Essentially this would combine the `useState` and `useEffect` into a single hook.

hkhalil-akamai:
I know I'm coming late to this PR, and I don't want to cause it to be delayed, but has anyone from UX given feedback? I'm concerned that a user might miss the additional bullet point (since they may open the migration dialog, read the warning and then go about customizing the options).

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [Unreleased]
 
+### Added:
+- Metadata migrate warning #9033
```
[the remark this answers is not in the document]
hana-linode: We don't need to include every small Metadata change but I think we should keep track of Metadata as a whole being unreleased so I updated the bullet to be more generic

hana-linode:
@hkhalil-akamai I haven't heard back from Matthew but Joe Michelini (original reporter) approved so I'm going to go ahead and merge this

PR #9036, opened 2023-04-19
fix: Fix Managed e2e navigation test flake on SSH tab

jdamore-linode:
## Description 📝
Fixes test flake that occurs in `managed-navigation.spec.ts` when both "Unauthorized" notices on the SSH tab load and appear at the same time; Cypress expected only one instance of the notice to appear and would fail when both appeared at the same time.

## How to test 🧪
Run `yarn && yarn build && yarn start:manager:ci` and then:

```bash
yarn cy:run -s "cypress/e2e/managed/managed-navigation.spec.ts"
```

Confirm that each test passes, preferably on the first attempt.


bnussman-akamai (approved):

cpathipa (approved):
Confirming that the spec passed in first attempt. 
image

PR #9038, opened 2023-04-20
chore: Upgrade Cypress to v12.11

jdamore-linode:
## Description 📝
This upgrades Cypress and some related dependencies. I'm hoping to get Cypress up to date before diving too deeply into the Chrome renderer crash issue just so that we can eliminate old/already fixed Cypress issues as possible causes of the issue.

## How to test 🧪
To test locally, run `yarn && yarn build && yarn start:manager:ci`, then run:

```bash
yarn cy:run
```

Alternatively, we can lean on the results for the Cypress run that will trigger automatically against this PR.


cpathipa (approved):

jdamore-linode:
I've been unable to get this to pass without getting a Chrome renderer process crash, so for now I think I'm going to wait to tackle that issue first (hopefully resolved by PR #9076 ) and then circle back to this to see if the problem is resolved.

PR #9041, opened 2023-04-21
chore: assets and utilities clean up

bnussman-akamai:
## Description 📝

- Clean up files in `packages/manager/src/assets/`  and `packages/manager/src/utilities/`
- Deleted util functions that were not used and that I don't think were worth keeping
- Removed some old unused assets
- Removed some default exports

## Preview 📷

- There should be no visual changes 🚨

## How to test 🧪

- Verify automated testing passes
- Do a quick once-over of the entire App and check for crashes and regressions


packages/manager/src/utilities/escapeStringForCLI.ts
```diff
@@ -1,5 +1,3 @@
-const escapeStringForCLI = (value: string) => {
+export const escapeStringForCLI = (value: string) => {
   return value.replace(/(["'$`\\])/g, '\\$1');
 };
```
jaalah-akamai, line 3: Just a potential for some more cleanup, possibly in a separate PR: This util essentially does the same as https://github.com/linode/manager/blob/develop/packages/manager/src/utilities/generate-cli.ts#L2-L10. However the one in `generate-cli` is more comprehensive in what it handles. Do we want to move forward with just one solution?

jaalah-akamai (approved):
Really good cleanup 🧼

mjac0bs (approved):
Those were some colorful distro `.png`s!

Verified that the deleted files are not used and didn't see any unexpected errors when clicking through Cloud to test on Friday; just tested Linode region functionality and that still looks good too.

packages/manager/src/utilities/escapeStringForCLI.ts
bnussman-akamai: Made myself a note to revisit this

PR #9042, opened 2023-04-21
chore: Update Typescript to `latest`

bnussman-akamai:
## Description 📝

- Updates Typescript to `latest`
- We're not going to use any new features right now but we may benefit from a performance boost

## Major Changes 🔄
- Updates Typescript to `latest`

## How to test 🧪
- Check Github Actions
- Make sure tsc does not get slower

### DIFF
--- a/packages/manager/src/features/Billing/PdfGenerator/PdfGenerator.ts
+++ b/packages/manager/src/features/Billing/PdfGenerator/PdfGenerator.ts
@@ -188,7 +188,10 @@ export const printInvoice = (
       unit: 'px',
     });
 
-    const convertedInvoiceDate = invoice.date && dateConversion(invoice.date);
+    const convertedInvoiceDate = invoice.date
+      ? dateConversion(invoice.date)
+      : null;
+
     const TaxStartDate =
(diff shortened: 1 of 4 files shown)


packages/manager/src/features/Billing/PdfGenerator/PdfGenerator.ts
```diff
@@ -188,7 +188,10 @@ export const printInvoice = (
       unit: 'px',
     });
 
-    const convertedInvoiceDate = invoice.date && dateConversion(invoice.date);
+    const convertedInvoiceDate = invoice.date
```
bnussman-akamai, line 191: Had to make this change because `convertedInvoiceDate` was type`boolean | ""`

packages/manager/tsconfig.json
```diff
@@ -33,7 +33,10 @@
     "noImplicitThis": true,
     "noUnusedLocals": true,
     "strictNullChecks": true,
+
+    /* Backwards Compatability */
     "suppressImplicitAnyIndexErrors": true,
+    "ignoreDeprecations": "5.0",
```
bnussman-akamai, line 39: Going to do this because I don't think I can fix 150 tsc errors today 😖

PR #9046, opened 2023-04-22
refactor: [M3-5181] - RQify Events

hkhalil-akamai:
## Description 📝
This PR migrates all Events endpoints to two React Query hooks (`useEvents` for fetching event notifications and `useEventsPolling` for registering event handlers) while hugely simplifying the current implementation.

The current Redux-based events system in Cloud Manager is pretty complex, making it difficult to understand and develop. Functionality is defined across several files, including `events.ts`, `eventsPolling.ts`, `event.request.ts`, `event.reducer.ts`, `event.actions.ts` and `event.helpers.ts`. Event handlers are used for important app functions like snackbar/toasts and cache invalidation but are defined via two redundant mechanisms (an rxjs subject and a custom Redux middleware).

The current system handles a lot of considerations and edge cases that must be accounted for:

- Polling rate backoff to minimize API load
- Increasing polling during background operations
- Updates to in-progress events
- Avoiding 'replaying' previously seen events
- Separate read statuses for separate users under a single account

The plan is to reduce complexity by consolidating all Events-related functionality to these two hooks, remove the singular usage of the rxjs library and leverage the API-provided `seen` and `read` flags on events to avoid the complicated logic needed for maintaining an up-to-date events store.

### DIFF
--- a/packages/manager/src/queries/events.ts
+++ b/packages/manager/src/queries/events.ts
@@ -1,16 +1,29 @@
 import { APIError } from '@linode/api-v4/lib/types';
 import { QueryClient, useQuery, useQueryClient } from 'react-query';
-import { Entity, Event } from '@linode/api-v4/lib/account';
-import { requestEvents } from 'src/events';
+import { Entity, Event, getEvents } from '@linode/api-v4/lib/account';
 import { useEffect } from 'react';
+import { generatePollingFilter } from 'src/utilities/requestFilters';
+import { DateTime } from 'luxon';
+import { ISO_DATETIME_NO_TZ_FORMAT } from 'src/constants';
+import {
+  isCompletedEvent,
+  isInProgressEvent,
+  isLongPendingEvent,
+  mostRecentCreated,
+  updateEvents,
+} from 'src/store/events/event.helpers';
+import { parseAPIDate } from 'src/utilities/date';
+import { events } from 'src/__data__/events';
 
 export interface EventsState {
   events: ExtendedEvent[];
-  mostRecentEventTime: number;
-  countUnseenEvents: number;
-  inProgressEvents: Record<number, number>;
-  pollingInterval: number;
-  requestDeadline: number;
+  mostRecentEventTime: number | undefined;
+  unseenEventsCount: number;
+
+  /**
+   * Map from Event ID to percentage complete.
+   */
+  inProgressEvents: Map<number, number>;
 }
 
 export interface ExtendedEvent extends Event {
@@ -37,27 +50,162 @@ const dispatchEvents = (
 
 const eventsCacheKey = [queryKey, 'cache'];
 export const useEventsPolling = (
-  eventHandler: (event: ExtendedEvent) => void
+  eventHandler?: (event: ExtendedEvent) => void
 ) => {
   const queryClient = useQueryClient();
 
   const eventHandlers = getEventHandlers(queryClient);
-  useEffect(() => {
-    eventHandlers.add(eventHandler);
-    return () => {
-      eventHandlers.delete(eventHandler);
-    };
-  }, [eventHandler, eventHandlers]);
+  useEffect(
+    () =>
+      eventHandler &&
+      (eventHandlers.add(eventHandler),
+      () => {
+        eventHandlers.delete(eventHandler);
+      }),
+    [eventHandler, eventHandlers]
+  );
 
-  return useQuery<EventsState, APIError[]>(eventsCacheKey, async () => {
-    const eventsCache = queryClient.getQueryData<EventsState>(queryKey);
-    const { eventsState, newEvents } = await requestEvents(eventsCache);
-    dispatchEvents(newEvents, eventHandlers);
-    return eventsState;
-  });
+  return useQuery<EventsState, APIError[]>(
+    eventsCacheKey,
+    async () => {
+      const eventsStateCache = queryClient.getQueryData<EventsState>(queryKey);
+      const { eventsState, newEvents } = await requestEvents(eventsStateCache);
+      dispatchEvents(newEvents, eventHandlers);
+      return eventsState;
+    },
+    {
+      refetchInterval: 5000,
+      retryDelay: 5000,
+      refetchOnMount: false,
+      refetchOnWindowFocus: false,
+    }
+  );
 };
 
 const eventHandlersKey = [queryKey, 'handlers'];
 const getEventHandlers = (queryClient: QueryClient) =>
   queryClient.getQueryData<Set<EventHandler>>(eventHandlersKey) ??
   queryClient.setQueryData<Set<EventHandler>>(eventHandlersKey, new Set());
+
+/**
+ * Will send a filtered request for events which have been created on or after the most recent existing
+ * event or the epoch if there are no stored events. Exclude events already in memory with a "+neq" filter.
+ */
+const requestEvents = (
+  eventsStateCache: EventsState | undefined
+): Promise<{
+  eventsState: EventsState;
+  newEvents: ExtendedEvent[];
+}> =>
+  getEvents(
+    { page_size: 25 },
+    eventsStateCache && createEventsFilter(eventsStateCache)
+  )
+    .then((response) => response.data)
+    /**
+     * There is where we set _initial on the events. In the default state of events the
+     * mostRecentEventTime is set to epoch. On the completion of the first successful events
+     * update the mostRecentEventTime is updated, meaning it's impossible for subsequent events
+     * to be incorrectly marked as _initial. This addresses our reappearing toast issue.
+     */
+    .then((events) =>
+      events.map((e) => ({ ...e, _initial: eventsStateCache === undefined }))
+    )
+    .then((events) => ({
+      eventsState: updateEventsState(eventsStateCache, events),
+      newEvents: events,
+    }));
+
+const createEventsFilter = (eventsStateCache: EventsState) => {
+  const {
+    mostRecentEventTime,
+    inProgressEvents,
+    events: _events,
+  } = eventsStateCache;
+
+  // Regardless of date created, we request events that are still in-progress.
+  const inIds = Object.keys(inProgressEvents).map((thisId) => +thisId);
+
+  // We need to keep polling the event for any database that is still creating.
+  // The same event will change its status from `notification` to `finished`.
+  const databaseEventIds = _events
+    .filter(
+      (event) =>
+        event.action === 'database_create' && event.status === 'notification'
+    )
+    .map((event) => event.id);
+
+  const includeIds = [...inIds, ...databaseEventIds];
+
+  // Generate a list of event IDs for the "+neq" filter. We want to request events created
+  // on or after the most recent created date, minus any events we've already requested.
+  // This is to catch any events that may be "lost" if the request/query lands at just the
+  // right moment such that we receive some events with a specific created date, but not all.
+  const neqIds: number[] = [];
+  if (_events.length > 0) {
+    _events.forEach((thisEvent) => {
+      const thisEventCreated = parseAPIDate(thisEvent.created).valueOf();
+
+      if (
+        thisEventCreated === mostRecentEventTime &&
+        !isInProgressEvent(thisEvent) &&
+        !isLongPendingEvent(thisEvent)
+      ) {
+        neqIds.push(thisEvent.id);
+      }
+    });
+  }
+
+  return generatePollingFilter(
+    DateTime.fromMillis(mostRecentEventTime ?? 0, { zone: 'utc' }).toFormat(
+      ISO_DATETIME_NO_TZ_FORMAT
+    ),
+    includeIds,
+    neqIds
+  );
+};
+
+const updateEventsState = (
+  eventsStateCache: EventsState | undefined,
+  newEvents: ExtendedEvent[]
+): EventsState => {
+  const updatedEvents = updateEvents(eventsStateCache?.events ?? [], newEvents);
+  return {
+    events: updatedEvents,
+    mostRecentEventTime: events.reduce(
+      mostRecentCreated,
+      eventsStateCache?.mostRecentEventTime
+    ),
+    unseenEventsCount: getNumUnseenEvents(updatedEvents),
+    inProgressEvents: updateInProgressEvents(
+      eventsStateCache?.inProgressEvents ?? new Map(),
+      events
+    ),
+  };
+};
+
+/**
+ * Iterate through new events.
+ * If an event is "in-progress" it's added to the inProgressEvents map.
+ * If an event is "completed" it is removed from the inProgressEvents map.
+ * Otherwise the inProgressEvents is unchanged.
+ */
+export const updateInProgressEvents = (
+  inProgressEvents: Map<number, number>,
+  events: Event[]
+) => {
+  return events.reduce((acc, e) => {
+    if (isCompletedEvent(e)) {
+      const updatedMap = new Map(acc);
+      updatedMap.delete(e.id);
+      return updatedMap;
+    }
+
+    return isInProgressEvent(e)
+      ? new Map(acc).set(e.id, e.percent_complete)
+      : acc;
+  }, inProgressEvents);
+};
+
+export const getNumUnseenEvents = (events: Pick<Event, 'seen'>[]) =>
+  events.reduce((result, event) => (event.seen ? result : result + 1), 0);
--- a/packages/manager/src/queries/events.ts
+++ b/packages/manager/src/queries/events.ts
@@ -1,0 +1,63 @@
+import { APIError } from '@linode/api-v4/lib/types';
+import { QueryClient, useQuery, useQueryClient } from 'react-query';
+import { Entity, Event } from '@linode/api-v4/lib/account';
+import { requestEvents } from 'src/events';
+import { useEffect } from 'react';
+
+export interface EventsState {
+  events: ExtendedEvent[];
+  mostRecentEventTime: number;
+  countUnseenEvents: number;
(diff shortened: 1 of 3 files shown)


packages/manager/src/queries/events.ts
```diff
@@ -0,0 +1,133 @@
...
+const queryKey = 'events';
+
+export const useEventsPolling = (
+  eventHandler?: (event: ExtendedEvent) => void
+) => {
+  useQuery<ExtendedEvent[], APIError[]>(
+    [queryKey, 'polling'],
+    async () => {
+      const events = await requestUnseenEvents();
+      await markCompletedEventsAsSeen(events);
```
bnussman-akamai, line 35: Should we be marking the events as `read` here? I think `seen` is the one currently used. I don't think we should change the behavior of it if possible

packages/manager/src/queries/events.ts
```diff
@@ -0,0 +1,133 @@
...
+export interface EntityEvent extends Omit<Event, 'entity'> {
+  entity: Entity;
+}
+
+const queryKey = 'events';
+
+export const useEventsPolling = (
+  eventHandler?: (event: ExtendedEvent) => void
+) => {
+  useQuery<ExtendedEvent[], APIError[]>(
```
bnussman-akamai, line 31: Can we avoid using extended types?

```suggestion
  useQuery<Event[], APIError[]>(
```

packages/manager/src/queries/events.ts
```diff
@@ -0,0 +1,133 @@
...
+      retryDelay: 5000,
+      refetchOnMount: false,
+      refetchOnWindowFocus: false,
+      onSuccess: (events) =>
+        eventHandler && events.reverse().forEach(eventHandler),
+    }
+  );
+};
+
+export const useEvents = () => {
```
bnussman-akamai, line 49: Instead of doing so much logic to make `useEvents` update the cache properly, what if we just invalidate the query if `useEventsPolling` gets any new events? This would allow us to keep the cache up to date with new/updated events without needing to manually manage it.

packages/manager/src/queries/events.ts
```diff
@@ -0,0 +1,110 @@
...
+          eventLocation.eventIdx
+        ] = event;
+        queryClient.setQueryData<InfiniteData<ResourcePage<Event>>>(
+          eventsQueryKey,
+          eventsCache
+        );
+      } else {
+        queryClient.invalidateQueries(eventsQueryKey);
+      }
+    }
```
bnussman-akamai, line 65: I think it would be cleaner for this logic to live in its own function or in the `onSuccess` function of `useEventsPolling`. I don't think `useEvents` should necessarily be dependent on `useEventsPolling` like it is right now.

packages/manager/src/queries/events.ts
```diff
@@ -0,0 +1,110 @@
...
+      retryDelay: 5000,
+      refetchOnMount: false,
+      refetchOnWindowFocus: false,
+      onSuccess: (events) =>
+        eventHandler && events.reverse().forEach(eventHandler),
+    }
+  );
+};
+
+export const useEvents = () => {
```
bnussman-akamai, line 41: Can we make this hook's name more specific?

```suggestion
export const useEventsInfiniteQuery = () => {
```

packages/manager/src/queries/events.ts
```diff
@@ -0,0 +1,110 @@
...
+} from '@linode/api-v4/lib/account';
+import { isInProgressEvent } from 'src/store/events/event.helpers';
+
+export interface EntityEvent extends Omit<Event, 'entity'> {
+  entity: Entity;
+}
+
+const queryKey = 'events';
+
+export const useEventsPolling = (eventHandler?: (event: Event) => void) => {
```
bnussman-akamai, line 22: Because `useEventsPolling` is just a wrapper over `useQuery`, I think for the sake of maintainability, we should try to keep this wrapper as "React Query"-like as possible. I think it would be good if we keep the param here of type `UseQueryOptions`. If you think `(event: Event) => void` is truly better, we can keep it, but I generally like to keep things as "React Query"-like as possible.

packages/manager/src/queries/events.ts
```diff
@@ -0,0 +1,110 @@
...
+  Entity,
+  Event,
+  getEvents,
+  markEventSeen,
+} from '@linode/api-v4/lib/account';
+import { isInProgressEvent } from 'src/store/events/event.helpers';
+
+export interface EntityEvent extends Omit<Event, 'entity'> {
+  entity: Entity;
+}
```
bnussman-akamai, line 18: I don't think this is needed.

```suggestion
```

packages/manager/src/queries/events.ts
```diff
@@ -0,0 +1,110 @@
...
+      await markCompletedEventsAsSeen(events);
+      return events;
+    },
+    {
+      refetchInterval: 5000,
+      retryDelay: 5000,
+      refetchOnMount: false,
+      refetchOnWindowFocus: false,
+      onSuccess: (events) =>
+        eventHandler && events.reverse().forEach(eventHandler),
```
bnussman-akamai, line 36: Is reverseing needed?

packages/manager/src/queries/events.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  const [intervalMultiplier, setIntervalMultiplier] = React.useState(1);
+
+  useQuery<Event[], APIError[]>(
+    [queryKey, 'polling', filter ?? {}],
+    async () => {
+      const events = await requestUnreadEvents(filter);
+      await markCompletedEventsAsRead(events);
+      setIntervalMultiplier(Math.min(intervalMultiplier + 1, 16));
+      return events;
```
bnussman-akamai, line 29: Also, I'm wondering if `await markCompletedEventsAsRead(events);` should be moved into the `onSuccess`. I think that would be more appropriate because we don't want this query to report an error if `markCompletedEventsAsRead` throws an error

packages/manager/src/queries/events.ts
```diff
@@ -0,0 +1,116 @@
...
+}) => {
+  const { eventHandler, filter } = options;
+
+  const [intervalMultiplier, setIntervalMultiplier] = React.useState(1);
+
+  useQuery<Event[], APIError[]>(
+    [queryKey, 'polling', filter ?? {}],
+    async () => {
+      const events = await requestUnreadEvents(filter);
+      await markCompletedEventsAsRead(events);
```
bnussman-akamai, line 27: Testing on my person account  with thousands of unread events caused events to be continuously marked as read every time the internal runs. 

This isn't that bad because it does all of this pretty gracefully but I worry we will run into a rate limit or hit the API pretty hard.  

I think we talked about this before and a possible solution may be to add a `{ created: { '+gte': timestampMountedAt } }`

bnussman-akamai (commented):
Things are shaping up nicely. Awesome job so far.

packages/manager/src/queries/events.ts
hkhalil-akamai: Do you anticipate any issues with the dependency? I think composing the hooks like this de-deduplicates some code.

packages/manager/src/queries/events.ts
hkhalil-akamai: Should we enforce this naming scheme (i.e., '-Query' after all query hooks and '-Mutation' after all mutation hooks) across the app?

packages/manager/src/queries/events.ts
hkhalil-akamai: I am more in favor of keeping the signature and usage of this hook as simple as possible. Even though it uses RQ under the hook, it's not really relevant to components that use it.

packages/manager/src/queries/events.ts
hkhalil-akamai: This is done to ensure the callback is called in chronological order; events are sorted with the most recent events coming first.

packages/manager/src/queries/events.ts
hkhalil-akamai: This was how I did it initially but the `onSuccess` callback is executed multiple times if there are multiple instances of this hook mounted. Having in the actual request avoids this. I investigated alternatives to `onSuccess` that only execute once but I don't think RQ has that option.

packages/manager/src/queries/events.ts
hkhalil-akamai: Capturing our discussion here for future reference:

I implemented this idea but @bnussman-akamai pointed out that filtering out events that occur before the hook was mounted would ignore previous in-progress events.

The current working solution to this that the `useEventsInfiniteQuery` hook will report in-progress events to the `useEventsPolling` hook to poll separately from new events.

bnussman-akamai: makes since but I worry about multiple sources of polling. I'm starting to lean towards one more API change to solve the edge case of capturing in-progress events created before mount.

packages/manager/src/features/Events/EventsLanding.tsx
hkhalil-akamai, line 1: Over half this file consisted of Redux logic that can now be removed!

packages/manager/src/features/NotificationCenter/NotificationData/useEventNotifications.tsx
```diff
@@ -47,6 +49,7 @@ const formatEventForDisplay = (
   event: ExtendedEvent,
   onClose: () => void
 ): NotificationItem => ({
+  originalId: event.id,
```
hkhalil-akamai, line 52: This field is added to find the original event and mark it as read.

packages/manager/src/features/TopMenu/NotificationMenu/NotificationMenu.tsx
```diff
@@ -34,6 +34,23 @@ const NotificationIconWrapper = styled(StyledTopMenuIconWrapper, {
...
+          backgroundColor: 'unset',
+        },
+      }
+    : {
+        '&:hover:not([aria-expanded="true"])': {
+          backgroundColor: 'unset',
+        },
+      }),
+}));
+
```
hkhalil-akamai, line 53: ✨ Enhancement: this fixes inconsistent hover-state background color of the Bell icon while the menu is open

packages/manager/src/queries/events.ts
```diff
@@ -0,0 +1,208 @@
...
+    filter: {
+      created: undefined,
+      '+or': [
+        ...generateInFilter(
+          'id',
+          incompleteEvents?.map((event) => event.id) ?? []
+        ),
+      ],
+    },
+  });
```
hkhalil-akamai, line 111: Using two instances of the polling hook here: the first one checks for new events and the second one checks for updates to previous in-progress events (e.g., an in-progress volume resize that was initiated before this hook was mounted).

packages/manager/src/features/Events/EventsLanding.tsx
```diff
@@ -54,223 +52,33 @@ const useStyles = makeStyles((theme: Theme) => ({
...
-          nextReactEvents
-        ),
-        eventsFromRedux: nextReduxEvents,
-        inProgressEvents: nextInProgressEvents,
-        mostRecentEventTime: nextMostRecentEventTime,
-      };
-  }
-};
-
 export const EventsLanding: React.FC<CombinedProps> = (props) => {
```
dwiley-akamai, line 63:
```suggestion
export const EventsLanding = (props: CombinedProps) => {
```

dwiley-akamai (commented):
The functionality is looking solid from my testing so far and this refactoring cleans things up quite a bit 👏

packages/manager/src/features/TopMenu/Help.tsx
```diff
@@ -6,16 +6,16 @@ import { StyledTopMenuIconWrapper } from './TopMenuIcon';
...
   return (
-    <Link aria-label="Link to Linode Support" to="/support">
-      <StyledTopMenuIconWrapper
-        sx={{
-          marginLeft: [null, null, '8px'],
-        }}
-      >
-        <TopMenuIcon title={'Help & Support'}>
+    <Link aria-label="Link to Linode Support" to="/support" title="Help">
+      <TopMenuIcon title={'Help & Support'}>
```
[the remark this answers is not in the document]
hkhalil-akamai: Agreed. But for some reason, the original order was causing the hover tooltip to not appear. Maybe we can rename the Wrapper to something that makes more sense?

packages/manager/src/features/TopMenu/NotificationMenu/NotificationMenu.tsx
```diff
@@ -34,6 +34,23 @@ const NotificationIconWrapper = styled(StyledTopMenuIconWrapper, {
   color: props.isMenuOpen ? '#606469' : '#c9c7c7',
 }));
 
+const StyledButton = styled(Button)(({ theme }) => ({
```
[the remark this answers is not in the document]
hkhalil-akamai: Thanks for reminding me about this style rule! I was the one who wrote it and I totally forgot 😂

dwiley-akamai (approved):
LGTM 🎉 

The only issue I'm seeing right now is when you power off a Linode, the percentage starts at 100% before adjusting to 0% when it first begins. I think this popped up on another PR recently but I don't see it in prod or `develop`, so merging the latest `develop` into the feature branch after this PR gets merged should probably solve that.

packages/manager/src/features/TopMenu/Help.tsx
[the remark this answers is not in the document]
dwiley-akamai: Yes a renaming is warranted here, maybe something like `StyledTopMenuSvgWrapper`.

mjac0bs:
> when you power off a Linode, the percentage starts at 100% before adjusting to 0% when it first begins. I think this popped up on another PR recently

This [fix](https://github.com/linode/manager/pull/9136/commits/f5a38ca02a784c6fe222ba82a97b5a7e8d9c7e94) was in the v1.93.1 hotfix PR.

Also [fixed in that PR was the regression where clicking on a secondary status](https://github.com/linode/manager/pull/9136/commits/3b12b8bfadd5e35dcf205beaa5ca8c18e4f825ce) (e.g. Migrating (0%)) on Linodes Landing or Linodes Detail didn't open the notification menu. We'll just want to make sure that fix is still looking good once develop is merged into this feature branch.

One thing I noticed on this branch but not prod or develop:

| This Branch  | Prod |
|  Screenshot 2023-05-22 at 12 51 34 PM | Screenshot 2023-05-22 at 12 52 02 PM  |

When a linode is cloned, once the event is complete, the linode we're cloning from has a secondary status of Cloning (100%) that doesn't go away on the Linode Details page. I think this is most likely a bug we introduced and then fixed in the v1.93.0 (re)release process, but I hadn't noticed it before this branch, so wanted to note it.

packages/manager/src/features/NotificationCenter/Events.tsx
```diff
@@ -1,18 +1,23 @@
...
   return (
     <NotificationSection
       content={events.slice(0, NUM_EVENTS_DISPLAY)}
       header="Events"
-      showMoreTarget="/events"
-      showMoreText="View all events"
+      showMore={{
+        target: '/events',
+        text: 'View all events',
+        onClick: notificationContext.closeMenu,
```
mjac0bs, line 19: Good call. 👍

packages/manager/src/features/TopMenu/NotificationMenu/NotificationMenu.tsx
mjac0bs: I was trying to figure out what the inconsistency was here, but couldn't seem to spot it. 🔍  I'll trust that it's fixed and this wasn't meant to address the blue button background in dark mode, which we do still need to fix but have a ticket for.

mjac0bs (approved):
Tested notification functionality in various scenarios and pages and observed all the things you noted in the steps to test. Also confirmed that the issues we were seeing before (starting at 100% to power on/off, Cloning stuck at 100%, and the secondary status not opening the menu) were all resolved once develop was merged in.

Looks like you have a few other comments on the code to circle back to, but I'm going to go ahead and approve this. Thanks for your clear PR description, that was great!

jaalah-akamai (approved):

packages/manager/src/features/TopMenu/NotificationMenu/NotificationMenu.tsx
hkhalil-akamai: @mjac0bs actually I think it was exactly that 😬. Should I mark that ticket done with this one then?

|Before|After|
|------|-----|
|<video src="https://github.com/linode/manager/assets/122488130/e5c19c46-17f5-40d2-917e-a8b95d3ce19f" />|<video src="https://github.com/linode/manager/assets/122488130/8878a92d-11eb-432e-921e-e299e0f2d6e4" />

mjac0bs: Ah, got it, I see the difference and improvement! What we do still need to fix is the blue background showing up at all, so let's not mark that ticket (M3-6542) as done yet unless you want to address that here. Like light mode, we want just the bell svg icon to change color when hovered/selected and the background color should remain the same as the rest of the app.

hkhalil-akamai:
I think I discovered the cause for the failing test ("uploads machine image, mock expired upload event"):

This test intercepts the getEvents call, but since these calls are duplicated for the time being ([see here for why](https://github.com/linode/manager/pull/9046#:~:text=(currently%2C%20polling%20is%20duplicated%20since%20the%20existing%20Events%20architecture%20is%20still%20enabled%3B%20it%20will%20be%20removed%20in%20Part%202).)), it intercepts the wrong call 50% of the time. I'm going ahead with the merge since this failure should be addressed in part 2 when I remove the previous events polling.

cc @jdamore-linode

PR #9051, opened 2023-04-24
refactor: [M3-6300] - MUI v5 Migration - `Components > CheckoutBar` 

dwiley-akamai:
## Description 📝
Migrates `SRC > Components > CheckoutBar` from JSS to styled components.

## Major Changes 🔄
- Removed `packages/manager/src/components/CheckoutBar/index.ts`
- Made use of styled components in `CheckoutBar.tsx`, housed in `/CheckoutBar/styles.ts`

## How to test 🧪
Go to `/kubernetes/create` and ensure the checkout bar looks and functions the same as in prod.

## To-Do:
- [x] Fix total amount display bug
- [x] Some `sx` refactoring

### DIFF
--- a/packages/manager/src/components/CheckoutBar/DisplaySection.tsx
+++ b/packages/manager/src/components/CheckoutBar/DisplaySection.tsx
@@ -1,7 +1,6 @@
-import { useTheme } from '@mui/material/styles';
 import * as React from 'react';
 import Typography from 'src/components/core/Typography';
-import { StyledDiv2 } from './styles';
+import { StyledDiv2, SxTypography } from './styles';
 
 export interface Props {
   title: string;
@@ -10,8 +9,6 @@ export interface Props {
 
 const DisplaySection = React.memo((props: Props) => {
   const { title, details } = props;
-
-  const theme = useTheme();
 
   return (
     <StyledDiv2>
@@ -24,11 +21,7 @@ const DisplaySection = React.memo((props: Props) => {
         <Typography
           component="span"
           data-qa-details={details}
-          sx={{
-            color: theme.color.headline,
-            fontSize: '.8rem',
-            lineHeight: '1.5em',
-          }}
+          sx={SxTypography}
         >
           {details}
         </Typography>
--- a/packages/manager/src/components/CheckoutBar/DisplaySection.tsx
+++ b/packages/manager/src/components/CheckoutBar/DisplaySection.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import Typography from 'src/components/core/Typography';
-import { StyledDiv2, SxTypography } from './styles';
+import { StyledCheckoutSection, SxTypography } from './styles';
 
 export interface DisplaySectionProps {
   title: string;
@@ -11,7 +11,7 @@ const DisplaySection = React.memo((props: DisplaySectionProps) => {
   const { title, details } = props;
 
   return (
-    <StyledDiv2>
+    <StyledCheckoutSection>
       {title && (
         <Typography variant="h3" data-qa-subheading={title}>
           {title}
@@ -26,7 +26,7 @@ const DisplaySection = React.memo((props: DisplaySectionProps) => {
           {details}
         </Typography>
       ) : null}
-    </StyledDiv2>
+    </StyledCheckoutSection>
   );
 });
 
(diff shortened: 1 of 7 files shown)


packages/manager/src/components/CheckoutBar/styles.ts
```diff
@@ -20,42 +22,29 @@ export const useStyles = makeStyles((theme: Theme) => ({
...
     },
-  },
-  sidebarTitle: {
-    color: theme.color.headline,
-    fontSize: '1.125rem',
-    wordBreak: 'break-word',
-  },
-  checkoutSection: {
-    animation: '$fadeIn 225ms linear forwards',
-    opacity: 0,
```
dwiley-akamai, line 31: It wasn't clear to me if `animation` and `opacity` were actually doing anything here, or what the benefit of keeping them would be, so I removed those properties.

packages/manager/src/components/CheckoutBar/DisplaySection.tsx
```diff
@@ -1,33 +1,33 @@
 import * as React from 'react';
 import Typography from 'src/components/core/Typography';
-import useStyles from './styles';
+import { StyledDiv2, SxTypography } from './styles';
 
 export interface Props {
```
mjac0bs, line 5: Same with `DisplaySectionProps`.

packages/manager/src/components/CheckoutBar/DisplaySectionList.tsx
```diff
@@ -1,12 +1,12 @@
 import * as React from 'react';
 import Divider from '../core/Divider';
-import DisplaySection from './DisplaySection';
+import { DisplaySection } from './DisplaySection';
 
 interface Props {
```
mjac0bs, line 5: And `DisplaySectionListProps`.

packages/manager/src/components/CheckoutBar/styles.ts
mjac0bs: Not totally sure either, but it seems like the "Create Cluster" button does a slight fade in when the component first loads in prod. I don't think this is necessary and initial rendering + interacting with the CheckoutBar looks smooth to me without any animation

mjac0bs (commented):
This looked good when testing with the preview link.

I noticed with local development (this branch and on `develop`), we don't display the HA section for kube checkout even though `showHighAvailability` is `true`. Do you know why that is?

![Screenshot 2023-04-25 at 8 53 23 AM](https://user-images.githubusercontent.com/114685994/234334082-782b6df0-9323-4989-95ad-0859d6968f49.jpg

dwiley-akamai:
> I noticed with local development (this branch and on `develop`), we don't display the HA section for kube checkout even though `showHighAvailability` is `true`. Do you know why that is?

It displays on mine; `REACT_APP_LKE_HIGH_AVAILABILITY_PRICE` probably isn't defined in your `.env` file

packages/manager/src/components/CheckoutBar/styles.ts
```diff
@@ -1,16 +1,18 @@
...
-    },
+const StyledButton = styled(Button)(({ theme }) => ({
+  marginTop: 18,
+  [theme.breakpoints.up('lg')]: {
+    width: '100%',
   },
-  root: {
+}));
+
+const StyledDiv1 = styled('div')(() => {
```
hkhalil-akamai, line 12: Do we have a convention for descriptively naming these styled components? For example, `Root` and `CheckoutSection` (the old JSS classnames) could be used instead of `StyledDiv1` and `StyledDiv2` and might make these names more meaningful.

mjac0bs: Thanks for asking this @hkhalil-akamai -- I think this is a good point and don't love `StyledDiv1` or `StyledDiv2` either. I do think it's useful to include the element name in the styled component name. I wonder if something like `RootStyledDiv` and `CheckoutSectionStyledDiv` would help address both of these points?

mjac0bs (approved):
> `REACT_APP_LKE_HIGH_AVAILABILITY_PRICE` probably isn't defined in your `.env` file

Yep, that'll do it! 

Pending some decision on the discussion about styled component naming (maybe this is a potential cafe topic?), the component in the UI looks good, so I'm approving.

packages/manager/src/components/CheckoutBar/styles.ts
jaalah-akamai: Agreed or `StyledRoot` and `StyledCheckoutSection`... just something more descriptive 👍

PR #9053, opened 2023-04-25
fix: [M3-6510] - Modifying Configuration Profile in Cloud breaks with VLANs

cpathipa:
## Description 📝
Fix - Modifying Configuration Profile in Cloud breaks with VLANs

## Preview 📷

| Before  | After   |
| ------ | ------- |
| image | image |

## How to test 🧪

**How to reproduce the issue (if applicable)?**
- Checkout to develop branch
- Create a Linode in Cloud Manager with a standard image (Ubuntu, for example) and no VLAN
- After the Linode provisions, edit the configuration profile in Cloud Manager, adding a VLAN (with or without an IPAM address), then Save the configuration profile
- Edit the configuration profile again, then try to either a) add another VLAN, b) remove the existing VLAN by switching eth1 to None, or c) do literally nothing other than immediately Save the configuration profile (with no changes made).

**How to verify changes?**

- Follow the above steps to create Linode, configure and update with VLAN

**How to run Unit or E2E tests?**

- yarn test
- yarn cy:run


packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
```diff
@@ -425,6 +425,11 @@ const LinodeConfigDialog: React.FC<CombinedProps> = (props) => {
             (thisOption) => thisOption.value === config?.root_device
           )
         );
+        const configInterfaces = config?.interfaces?.map(
```
bnussman-akamai, line 428: Looks like we should update the type of `config.interfaces` so that this makes more sense. The API returns an `id` but our types don't show that.

bnussman-akamai (changes requested):

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - Event entities should only be linked for true labels
 - Radio button hover effect #9031
 - Prevent form submission unless action was taken (IP transfer & IP sharing modals) #5976
+- fix: [M3-6510] - Modifying Configuration Profile in Cloud breaks with VLANs #9053
```
dwiley-akamai, line 24:
```suggestion
Inability to edit and save Linode Configurations #9053
```

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
dwiley-akamai: Interestingly, it's also [undocumented](https://www.linode.com/docs/api/linode-instances/#configuration-profiles-list) in the API docs 🤔

dwiley-akamai (commented):
Curious as to what made this break out of nowhere 🤔

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
cpathipa: Also, as per the documentation Payloads doesn't accept` id` as well.

hkhalil-akamai:
> Curious as to what made this break out of nowhere 🤔

It appears to be related to this API ticket ARB-3629. The interface was modified to add the `id` field but the docs have not been updated.

hkhalil-akamai (approved):
Tested to confirm the fix is working as intended.

Approved pending `api-v4` suggestion by @bnussman-akamai.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - Event entities should only be linked for true labels
 - Radio button hover effect #9031
 - Prevent form submission unless action was taken (IP transfer & IP sharing modals) #5976
+Inability to edit and save Linode Configurations #9053
```
bnussman-akamai, line 24:
```suggestion
- Inability to edit and save Linode Configurations #9053
```

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
```diff
@@ -438,7 +443,7 @@ const LinodeConfigDialog: React.FC<CombinedProps> = (props) => {
             virt_mode: config.virt_mode,
             helpers: config.helpers,
             root_device: config.root_device,
-            interfaces: interfacesToState(config.interfaces),
+            interfaces: interfacesToState(configInterfaces),
```
bnussman-akamai, line 446: Instead of doing the `map` before this to setup the data, I think we should move all of that data manipulation to `interfacesToState`

bnussman-akamai (commented):
I think the next steps for this PR would to be to update the `api-v4` types like the following

```ts
export interface Interface {
  id: number;
  label: string | null;
  purpose: InterfacePurpose;
  ipam_address: string | null;
}

type InterfacePayload = Omit<Interface, 'id'>;
```

Then update `LinodeConfigCreationData` to use `InterfacePayload`

cpathipa:
Great feedback! Thank you @bnussman-akamai and @hkhalil-akamai for taking a deeper look into this PR.

packages/manager/src/features/linodes/LinodesDetail/LinodeSettings/LinodeConfigDialog.tsx
```diff
@@ -193,10 +193,15 @@ const pathsOptions = [
...
+    ({ ipam_address, label, purpose }) => {
+      return { ipam_address, label, purpose };
+    }
+  );
+  if (!processedInterfaces || processedInterfaces.length === 0) {
     return defaultInterfaceList;
   }
-  return padInterfaceList(interfaces);
+  return padInterfaceList(processedInterfaces);
 };
```
bnussman-akamai, line 205: `interfacesToState` could use some clean up, but looks good to me otherwise

```typescript
const interfacesToState = (interfaces?: Interface[]) => {
  if (!interfaces || interfaces.length === 0) {
    return defaultInterfaceList;
  }
  const interfacesPayload = interfaces.map(
    ({ ipam_address, label, purpose }) => ({ ipam_address, label, purpose })
  );
  return padInterfaceList(interfacesPayload);
};
```

bnussman-akamai (approved):

PR #9058, opened 2023-04-26
fix: [M3-6454] - Show error for PayPal payments

carrillo-erik:
## Description 📝
Display an error Notice to the user when PayPal fails to launch after clicking the `PayPalChip` in the `AddPaymentMethodDrawer` or the `PayPalButton` in the `PaymentDrawer`. Previously a pop-up window would flash and the user would be unaware that an error had occurred.

Also, if the `BrainTreePayPalButtons` fail to load initially, an error `ToolTipIcon` is shown to the user.

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screen Shot 2023-04-27 at 9 11 05 AM | Screen Shot 2023-04-27 at 8 30 46 AM |
| Screen Shot 2023-04-27 at 9 06 38 AM | Screen Shot 2023-04-27 at 8 30 25 AM |
| Screen Shot 2023-04-27 at 9 05 00 AM | Screen Shot 2023-04-27 at 8 36 56 AM |

## How to test 🧪
1. Navigate to the Billing page on your local environment and click on either the `Make A Payment` or `Add Payment Method` buttons.
2. Click on the PayPal option.
3. You should see a PayPal modal flash and then the error Notice should be displayed in the appropriate payment drawer.
4. The `BrainTreePayPalButtons` sometimes fail to load which would result in an empty space, try reloading the page a few times and if the failure occurs, the error `TooltipIcon` should appear in their place.


bnussman-akamai (approved):
Thanks for doing this and deleting the old PayPal code!

Did we decide that we need to keep the `PayPalErrorBoundary`?

jaalah-akamai (approved):

PR #9059, opened 2023-04-27
fix: [M3-5852] - Phone Verification - Verification Code State Does Not Reset

bnussman-akamai:
## Description 📝

- Clear phone verification code error so error does not persist on new flow

## How to test 🧪
- Go to `http://localhost:3000/profile/auth`
- First make sure you have your phone number verified
- `Edit` your verified phone number
- Send yourself a verification code
- Enter an **incorrect code** so an error shows
- Click `Enter a different phone number` or `Canel` to start the flow over
- Send yourself a verification code again
- **Verify that you don't see the same error as before**
  - If you don't see the same error as before, this bug has been fixed


dwiley-akamai (approved):
This takes care of the bug 👍 

Should we also consider clearing out the verification code field of the previous attempt's input?

mjac0bs (approved):
Followed the steps and confirmed that the previous error message and verification code text input is cleared after starting the flow again.

PR #9067, opened 2023-05-01
fix: [M3-6540] - Surface general errors in the Object Storage Bucket Create Drawer

bnussman-akamai:
## Description 📝

- Surfaces any general errors as a notice in the Object Storage Create Bucket drawer
  - Previously, only errors specific to the `label` and `region` would show

## Preview 📷

Screenshot 2023-05-01 at 11 15 09 AM

## How to test 🧪
- Enable the MSW
- Uncomment the mock error in `serverHandlers.ts` 
- Try to create an Object Storage Bucket
- Observe that a general error is shown as a notice

### DIFF
--- a/packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.test.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.test.tsx
@@ -22,7 +22,10 @@ describe('CreateBucketDrawer', () => {
   it('Should show a general error notice if the API returns one', async () => {
     server.use(
       rest.post('*/object-storage/buckets', (req, res, ctx) => {
-        return res(ctx.status(500), ctx.json({ errors: [{ reason: 'omg' }] }));
+        return res(
+          ctx.status(500),
+          ctx.json({ errors: [{ reason: 'Object Storage is offline!' }] })
+        );
       }),
       rest.get('*/regions', async (req, res, ctx) => {
         return res(
@@ -61,6 +64,7 @@ describe('CreateBucketDrawer', () => {
 
     userEvent.type(getByLabelText('Label'), 'my-test-bucket');
 
+    // We must waitFor because we need to load region and cluster data from the API
     await waitFor(() =>
       userEvent.selectOptions(
         getByPlaceholderText('Select a Region'),
@@ -72,6 +76,6 @@ describe('CreateBucketDrawer', () => {
 
     userEvent.click(saveButton);
 
-    await findByText('omg');
+    await findByText('Object Storage is offline!');
   });
 });
--- a/packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.test.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.test.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { CreateBucketDrawer } from './CreateBucketDrawer';
-import { waitFor } from '@testing-library/react';
+import { act, waitFor } from '@testing-library/react';
 import { renderWithTheme } from 'src/utilities/testHelpers';
 import userEvent from '@testing-library/user-event';
 import { rest, server } from 'src/mocks/testServer';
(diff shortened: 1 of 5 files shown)


jdamore-linode (approved):
Nice improvement!

packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -635,6 +635,15 @@ export const handlers = [
     const buckets = objectStorageBucketFactory.buildList(10);
     return res(ctx.json(makeResourcePage(buckets)));
   }),
+  rest.post('*/object-storage/buckets', (req, res, ctx) => {
+    // return res(
```
abailly-akamai, line 639: Rather than leaving this comment here, can we have a unit test handle this case?

bnussman-akamai: Pushed up a unit test for this but I can't figure out whats wrong with it. Help would be appreciated

abailly-akamai: sure thing i'll take a look in a bit

packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    );
+
+    const {
+      getByTestId,
+      getByLabelText,
+      getByPlaceholderText,
+      findByText,
+    } = renderWithTheme(<CreateBucketDrawer {...props} />);
+
+    userEvent.type(getByLabelText('Label'), 'my-test-bucket');
```
abailly-akamai, line 62: Shouldn't those be wrapped in an `act()`?

bnussman-akamai: When I wrap `userEvent.type(getByLabelText('Label'), 'my-test-bucket');` in an `act` it does not update properly

bnussman-akamai: This is where I'm stuck. Something about the mock error isn't correct but I don't know what it is. 

Screenshot 2023-05-01 at 5 37 48 PM

abailly-akamai: am not getting that error by doing something like this:

```
const labelField = getByTestId('textfield-input');
const saveButton = getByTestId('create-bucket-button');
await act(async () => {
  userEvent.type(labelField, 'my-test-bucket');
  await waitFor(() =>
    userEvent.selectOptions(
      getByPlaceholderText('Select a Region'),
      'Newark, NJ (us-east-1)'
    )
  );
  userEvent.click(saveButton);
});

await findByText('something bad happened!');
```

but it does not find the element. am confused as to what's going on here

abailly-akamai: nevermind this is erroneous code ^

bnussman-akamai: yeah, I've never seen anything like this. Feels like an Axios interceptor issue or something

bnussman-akamai: Okay, yeah. Seems like the `error` is an actual `Error`, not in the `APIError` shape like we'd expect. This is probably caused by the Axios interceptors not being applied to transform the error to the expected shape. I don't know if this is a new problem or an existing one.

bnussman-akamai: Got it fixed in 18edd23. Axios interceptors were not setup...

abailly-akamai: ah thanks, nice fix - strange that this is the only test relying on those

abailly-akamai (approved):

PR #9069, opened 2023-05-01
fix: [M3-6022] - Add compression to logo when rendering in JPG format during invoice PDF generation

rmcintosh:
## Description 📝
Currently these PDFs are >1 MB in size. With this change, my testing showed the resulting PDF to be closer to 50kb in size. Logo quality looks the same to me 🤷🏽 

## How to test 🧪
Generate a PDF of the same invoice with and without this change.


mjac0bs (approved):
Thank you for making this change, @rmcintosh! We had a ticket in our backlog for it that had not yet been addressed.

Confirmed that the invoices are a much more reasonable size now with the JPG compression.  
Screenshot 2023-05-01 at 1 45 59 PM

cpathipa (approved):
LGTM!

bnussman-akamai (approved):

PR #9075, opened 2023-05-02
feat: [M3-5655] - Allow user to download DNS zone file

cpathipa:
## Description 📝
This enables user to Download DNS Zone file. 

| Before  | After   |
| ------- | ------- |
| image | image |

## How to test 🧪

- Navigate to http://localhost:3000/domains
- Click on any domain.
- Able to see "Download DNS Zone file" link type button.
- Upon clicking, should be able to download the DNS zone file.


packages/manager/src/features/Domains/DownloadDNSZoneFileButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+    const zoneFileContent = data?.zone_file.join('\n');
+    if (zoneFileContent) {
+      const blob = new Blob([zoneFileContent], { type: 'text/plain' });
+      const url = URL.createObjectURL(blob);
+      const a = document.createElement('a');
+      a.href = url;
+      a.download = 'DNSzonefile.txt';
+      document.body.appendChild(a);
+      a.click();
+      document.body.removeChild(a);
```
bnussman-akamai, line 22: Can we use our `downloadFile` helper function here?

packages/manager/src/features/Domains/DownloadDNSZoneFileButton.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import { LinkButton } from 'src/components/LinkButton';
+import { useDNSZoneFileQuery } from 'src/queries/domains';
+
+type Props = {
+  id: number;
+};
+
+export const DownloadDNSZoneFileButton = ({ id }: Props) => {
+  const { data } = useDNSZoneFileQuery(id);
```
bnussman-akamai, line 10: This might be one of the rare cases where we don't really need to use React query. We could just call `await getDNSZoneFile()` in handleClick

packages/api-v4/src/domains/domains.ts
```diff
@@ -99,3 +100,15 @@ export const importZone = (data: ImportZonePayload) =>
     setURL(`${API_ROOT}/domains/import`),
     setMethod('POST')
   );
+
+/**
+ * Imports a domain zone from a remote nameserver.
+ *
+ * @param domain { string } The domain to import.
+ * @param remote_nameserver { string } The remote nameserver that allows zone transfers (AXFR).
+ */
```
jdamore-linode, line 109: I think this might've been carried over from some older code and not updated for the new function

jdamore-linode (approved):
Looks good! Approving pending that suggestion from Banks. Nice work!

packages/manager/src/features/Domains/DownloadDNSZoneFileButton.tsx
```diff
@@ -0,0 +1,20 @@
...
+type Props = {
+  id: number;
+};
+
+export const DownloadDNSZoneFileButton = ({ id }: Props) => {
+  const handleClick = async () => {
+    const data = await getDNSZoneFile(id);
+    const zoneFileContent = data?.zone_file.join('\n');
+    if (zoneFileContent) {
+      downloadFile('DNSzonefile.txt', zoneFileContent);
```
bnussman-akamai, line 15: Can we have the file name be in the format `[domain].txt`? For example, `linode.com` would be  `linode.com.txt`. `DNSzonefile.txt` isn't very descriptive

```suggestion
      downloadFile(`${domain}.txt`, zoneFileContent);
```

packages/manager/src/features/Domains/DomainRecords.tsx
```diff
@@ -326,6 +327,7 @@ class DomainRecords extends React.Component<CombinedProps, State> {
           },
         },
       ],
+      link: () => <DownloadDNSZoneFileButton id={this.props.domain.id} />,
```
bnussman-akamai, line 330: Just realized this now and it's not your fault, but the location of this button makes no sense. Downloading the entire zone file isn't specific to the "SOA records"... 

 I think this button should go beside the `Delete Domain` button at the bottom of the page

Screenshot 2023-05-02 at 2 14 03 PM

bnussman-akamai (changes requested):

packages/manager/src/features/Domains/DomainRecords.tsx
cpathipa: Good call. I agree, it shouldn't be specific to SOA records. IMO, we showing it next to docs link may be more visible to user to download DNS. I will circle back with UX on this. 

image

packages/manager/src/features/Domains/DownloadDNSZoneFileButton.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <DownloadDNSZoneFileButton domainLabel="test.com" domainId={1} />
+    );
+    fireEvent.click(getByText('Download DNS Zone File'));
+    await waitFor(() =>
+      expect(
+        require('src/utilities/downloadFile').downloadFile
+      ).toHaveBeenCalledTimes(1)
+    );
+    expect(
+      require('src/utilities/downloadFile').downloadFile
```
bnussman-akamai, line 37: That's cool that this works, I'm just unsure if it is recommended. `require` might not work nicely in a non commonjs environment in the future.

```suggestion
      require('src/utilities/downloadFile').downloadFile
```

bnussman-akamai (approved):
Awesome! 🚀

packages/manager/src/features/Domains/DownloadDNSZoneFileButton.test.tsx
cpathipa: Good call, replaced `require` with `import`.

PR #9080, opened 2023-05-03
refactor: [M3-6222] - Refactor Images E2E Test Intercepts

cliu-akamai:
## Description 📝
- Refactor Images tests to use defined intercepts methods

## Major Changes 🔄
- Created a `images.ts` file.
- Updated code in `linodes.ts` file.
- Changed to use the intercepts methods in the file above.

## How to test 🧪
`yarn cy:run -s "cypress/e2e/images/create-linode-from-image.spec.ts"`
`yarn cy:run -s "cypress/e2e/images/machine-image-upload.spec.ts"`
`yarn cy:run -s "cypress/e2e/images/smoke-create-image.spec.ts"`

### DIFF
--- a/packages/manager/cypress/e2e/images/smoke-create-image.spec.ts
+++ b/packages/manager/cypress/e2e/images/smoke-create-image.spec.ts
@@ -1,8 +1,32 @@
-import type { Linode } from '@linode/api-v4/types';
+import type { Image, Linode, Disk } from '@linode/api-v4/types';
 import { imageFactory } from 'src/factories/images';
 import { createLinode, deleteLinodeById } from 'support/api/linodes';
-import { apiMatcher } from 'support/util/intercepts';
+import { interceptCreateImage, mockGetImages } from 'support/intercepts/images';
+import { mockGetLinodeDisks } from 'support/intercepts/linodes';
 import { randomLabel, randomNumber, randomPhrase } from 'support/util/random';
+
+const diskLabel = 'Debian 10 Disk';
+
+const mockDisks: Disk[] = [
+  {
+    id: 44311273,
+    status: 'ready',
+    label: diskLabel,
+    created: '2020-08-21T17:26:14',
+    updated: '2020-08-21T17:26:30',
+    filesystem: 'ext4',
+    size: 81408,
+  },
+  {
+    id: 44311274,
+    status: 'ready',
+    label: '512 MB Swap Image',
+    created: '2020-08-21T17:26:14',
+    updated: '2020-08-21T17:26:31',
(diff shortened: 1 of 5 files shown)


packages/manager/cypress/support/intercepts/linodes.ts
```diff
@@ -96,3 +96,27 @@ export const interceptRebootLinodeIntoRescueMode = (
...
+): Cypress.Chainable<null> => {
+  return cy.intercept(
+    'GET',
+    apiMatcher(`linode/instances/${linodeId}/disks*`),
+    {
+      results: disks.length,
+      data: disks,
+      page: 1,
+      pages: 1,
+    }
```
jdamore-linode, line 120:
```suggestion
    paginateResponse(disks)
```
Even though this works, it's probably better to stick to the `paginateResponse()` util so that we only need to touch the code in one place if the shape of our paginated responses ever changes (admittedly unlikely).

packages/manager/cypress/e2e/images/smoke-create-image.spec.ts
```diff
@@ -22,40 +46,18 @@ describe('create image', () => {
...
-    }).as('getImages');
-    cy.intercept('POST', apiMatcher('images'), mockNewImage).as('createImage');
+    const mockImages = new Array(2).fill(null).map(
+      (item: null, index: number): Image => {
+        return imageFactory.build({
+          label: `Image ${index}`,
+        });
+      }
+    );
+    mockGetImages(mockImages).as('getImages');
```
jdamore-linode, line 56:
```suggestion
    const mockImages = imageFactory.buildList(2);
    mockGetImages(mockImages).as('getImages');
    
    // Alternatively, we could even use an empty array:
    mockGetImages([]).as('getImages');
```
This could be simplified by using `imageFactory.buildList(2)` since we don't need to manually set the label for the created images for any particular reason.

For some additional context, the `Array(n).fill(null).map()` pattern is really only needed when we need to create an array of objects using a factory and set a specific value on a per item basis, which isn't really the case in this instance. But before we switched to Vite, we were unable to import [`factory.ts`](https://www.npmjs.com/package/factory.ts) from within our Cypress tests (which gives us handy utils like `Factory.each()`), so following this pattern was the next best choice. I think the issue with importing `factory.ts` has probably been resolved since switching to Vite, so I'm thinking we might not need to follow this pattern going forward.

jdamore-linode (approved):
Nice work Cassie! Nice to see these tests cleaned up! Posted a couple small suggestions but otherwise everything is looking good and running well!

PR #9082, opened 2023-05-03
refactor: [M3-6400, M3-6401, M3-6404, M3-6408] - MUI v5 Migration - Components > Table, TableBody, TableHead, TableCell, TableSortCell, TableRow

bnussman-akamai:
## Description 📝

- MUI v5 Migration for most Table related components 

## Major Changes 🔄
- Runs the TSS migration on Table, TableCell, TableSortCell, TableRow
- Removes `src/components/core/[Table].ts` files because they are a duplicate point of abstraction and a potential way for developers to import an unstyled/un-linode-ifyed component

## Preview 📷

Screenshot 2023-05-03 at 4 56 39 PM

## How to test 🧪
- Check different tables throughout the app for styling issues
- I recommend opening production in one tab and this PR in another and comparing the table styles 
- Verify CI ad e2es pass

PR #9085, opened 2023-05-05
fix: Linode Details Access Table UI

bnussman-akamai:
## Description 📝

- Fixes UI regression in the Linode Details Entity Header access table caused by https://github.com/linode/manager/pull/9082
- This could probably be fixed other ways but making the easy change because it matches the original intention of the component and we may rebuild some UI soon and it can be fixed then

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-05-05 at 9 21 08 AM | Screenshot 2023-05-05 at 9 20 50 AM |

## How to test 🧪
- Check the UI shown above


jdamore-linode (approved):

abailly-akamai (approved):
nice and good to know!

PR #9087, opened 2023-05-05
change: [M3-6057] - Avoid saving/uploading Cypress recordings for successful specs in CI

jdamore-linode:
## Description 📝
This disables uploading Cypress recordings for specs who run without any flake or failures. Before this change, all recordings get uploaded to the Cypress dashboard regardless of whether they are successful or not. Because we don't need recordings for passed tests, we can save some time by skipping the compression/recording step in these cases.

This change only applies when the `CI` environment variable is present; recording behavior is unchanged for local runs (but can be tested locally by setting the `CI` env var).

## How to test 🧪
This can be tested using the automatic run against this PR, but can also be tested locally:

```bash
CI=true yarn cy:run
```
Confirm that the CLI output only shows recording information for tests that fail or are flaky, and that there is no mention of a recording being saved/compressed for spec files without flake and failures. (Without `CI=true`, you should see a note about saving/compressing a recording for every spec file in the output, regardless of pass/fail)

### DIFF
--- a/packages/manager/cypress.config.ts
+++ b/packages/manager/cypress.config.ts
@@ -1,10 +1,11 @@
 /* eslint-disable no-console */
 import { defineConfig } from 'cypress';
 import { resolve } from 'path';
-// switch to import syntax when esModuleInterop": true
-// eslint-disable-next-line @typescript-eslint/no-var-requires
-const vitePreprocessor = require('cypress-vite');
 import * as dotenv from 'dotenv';
+
+// Dependencies used in hooks have to use `require()` syntax.
+const vitePreprocessor = require('cypress-vite'); // eslint-disable-line
+const fs = require('fs'); // eslint-disable-line
 
 /**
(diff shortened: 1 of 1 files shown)


bnussman-akamai (commented):
Neat! Could the config option `videoUploadOnPasses` be useful to us? What difference does `videoUploadOnPasses` have compared to your solution here?

jdamore-linode:
> Neat! Could the config option `videoUploadOnPasses` be useful to us? What difference does `videoUploadOnPasses` have compared to your solution here?

Sadly no, because when that's set to `false` it also stops uploading recordings of flaky tests, and we lean on those pretty heavily for troubleshooting.

I should have linked to it in the description, but there's an open Cypress issue that discusses this more, and is also where I found this workaround:

https://github.com/cypress-io/cypress/issues/16377

bnussman-akamai (approved):

jaalah-akamai (approved):

PR #9099, opened 2023-05-08
refactor: [M3-6488] - React Query for Linode Details - General Details

bnussman-akamai:
## Description 📝
- Switches the "top half" of the Linodes details page to use React Query
- Once we make more progress, another "general" PR like this one will probably be needed to address the comments I left
- The code might still looks and feel a bit crusty but we'll be able to get things cleaner as we migrate more components

## Preview 📷
 > **Note**: There should be no visual changes
 
Screenshot 2023-05-08 at 6 19 19 PM

## How to test 🧪
- Verify that the linodes details page is still working as expected
- Verify e2es pass


packages/manager/src/features/linodes/LinodeEntityDetail.tsx
```diff
@@ -138,20 +125,19 @@ const LinodeEntityDetail: React.FC<CombinedProps> = (props) => {
...
           ipv4={linode.ipv4}
           ipv6={trimmedIPv6}
           linodeId={linode.id}
-          username={username ? username : 'none'}
         />
       }
       footer={
         <Footer
-          linodePlan={linodePlan}
+          linodePlan={linode.type}
```
hkhalil-akamai, line 140: Do we want to be displaying the type ID here (i.e., 'g6-nanode-1') instead of the label ('Nanode 1 GB') as before?

packages/manager/src/features/linodes/LinodeEntityDetail.tsx
```diff
@@ -33,74 +31,65 @@ import { pluralize } from 'src/utilities/pluralize';
...
+
+  const { data: linode } = useLinodeQuery(id);
+
+  const { data: volumesData } = useLinodeVolumesQuery(id);
 
   const { data: images } = useAllImagesQuery({}, {});
-  const imagesItemsById = listToItemsByID(images ?? []);
 
-  const typesQuery = useSpecificTypes(linode.type ? [linode.type] : []);
-  const type = typesQuery[0]?.data ? extendType(typesQuery[0].data) : undefined;
```
hkhalil-akamai, line 88: One downside of not extending the type is that the label is not getting formatted consistently across the app. The formatted label, which is used in most other places of the app, includes a space before the storage unit (e.g., 'Nanode 1 GB' instead of 'Nanode 1GB').

It's a very minor thing but it does introduce some inconsistency and might before more important in the future if there are significant changes to the formatted label.

hkhalil-akamai: This might be worth a brief Cafe discussion.

dwiley-akamai: I think we should try to hold off on changing the formatting until a wider convo is had with UX. In the meantime, we should be able to make use of `formatStorageUnits()` to retain our current formatting.

packages/manager/src/features/linodes/LinodesDetail/LinodesDetailHeader/MutationNotification.tsx
```diff
@@ -1,105 +1,79 @@
...
-  const {
-    classes,
+  const { data: linode } = useLinodeQuery(linodeId);
+
+  const { data: currentTypeInfo } = useTypeQuery(
+    linode?.type ?? '',
+    linode !== undefined
+  );
+
+  const { data: sucessorTypeInfo } = useTypeQuery(
```
dwiley-akamai, line 38: `successorTypeInfo`

packages/manager/src/features/linodes/LinodesDetail/LinodesDetail.tsx
```diff
@@ -3,47 +3,54 @@ import { useDispatch } from 'react-redux';
...
+  }
 
-  if (!linode) {
-    return <NotFound />;
+  if (isLoading || !linode || !extendedLinode) {
+    return <CircleProgress />;
   }
 
-  const ctx: LinodeDetailContext = createLinodeDetailContext(linode, dispatch);
+  // We will delete this as soon a we react query all consumers of this context
```
dwiley-akamai, line 49: "[...] as soon as [...]"

packages/manager/src/features/linodes/LinodeEntityDetail.tsx
bnussman-akamai: It pains me to introduce tech debt like this but I made the change to keep RQ moving 👍🏼 f7e7054

dwiley-akamai (approved):

PR #9101, opened 2023-05-09
refactor: [M3-6563] - Create util function to filter plan selecting t…

cpathipa:
## Description 📝
 Common util function `getPlanSelectionsByPlanType` servers plan selections to Linode, Database and LKE plan tables.

## How to test 🧪

- Validate Linode Create / Clone / Update / resize / backup flows
- Check Plans table is rendering all plan types as expected.
- Validate LKE Plans table rendering all plan types as expected.
- Validate Database plans table rendering all plan types as expected.
- yarn test
- yarn cy:run


packages/manager/src/utilities/filterPlanSelectionsByType.ts
```diff
@@ -0,0 +1,47 @@
...
+    } else if (/^dedicated/.test(type.class)) {
+      filteredPlansByType.dedicated.push(type);
+    } else if (/gpu/.test(type.class)) {
+      filteredPlansByType.gpu.push(type);
+    } else if (/metal/.test(type.class)) {
+      filteredPlansByType.metal.push(type);
+    } else if (/premium/.test(type.class)) {
+      filteredPlansByType.premium.push(type);
+    }
+  });
```
bnussman-akamai, line 44: What do you think about using a for loop here? I find them more readable.

Also, I think we'd be okay to check for strict equality.

I wonder if a `switch` statement would work nicely here?

```suggestion
  for (const type of types) {
    if (type.class === 'nanode') {
      filteredPlansByType.nanodes.push(type);
    } else if (type.class === 'standard') {
      filteredPlansByType.standard.push(type);
    } else if (type.class === 'highmem') {
      filteredPlansByType.highMem.push(type);
    } else if (type.class === 'prodedicated') {
      filteredPlansByType.proDedicated.push(type);
    } else if (type.class === 'dedicated') {
      filteredPlansByType.dedicated.push(type);
    } else if (type.class === 'gpu') {
      filteredPlansByType.gpu.push(type);
    } else if (type.class === 'metal') {
      filteredPlansByType.metal.push(type);
    } else if (type.class === 'premium') {
      filteredPlansByType.premium.push(type);
    }
  }
```

cpathipa: @bnussman-akamai yes we could use for of with switch statement.

packages/manager/src/utilities/filterPlanSelectionsByType.ts
```diff
@@ -0,0 +1,58 @@
...
+interface FilteredPlansTypes {
+  nanodes: PlanSelectionType[];
+  standard: PlanSelectionType[];
+  highMem: PlanSelectionType[];
+  proDedicated: PlanSelectionType[];
+  dedicated: PlanSelectionType[];
+  gpu: PlanSelectionType[];
+  metal: PlanSelectionType[];
+  premium: PlanSelectionType[];
+}
```
bnussman-akamai, line 12: This type could be simplifed to 
```suggestion
type FilteredPlansTypes = Record<LinodeTypeClass, PlanSelectionType[]>;
```

if you are willing to change `nanodes` to `nanode` and `highMem` to `highmem` and `prodDedicated` to `prodedicated`.

I think this change is optional, but I I find that matching the API's types as much as possible help keep things simple and more understandable

bnussman-akamai (commented):
Looking clean 🧼

packages/manager/src/utilities/filterPlanSelectionsByType.ts
```diff
@@ -0,0 +1,123 @@
...
+  highMem: ExtendedType[];
+  proDedicated: ExtendedType[];
+  dedicated: ExtendedType[];
+  gpu: ExtendedType[];
+  metal: ExtendedType[];
+  premium: ExtendedType[];
+}
+
+/*
+The function 'filterPlanSelectionsByType' supports the Database plan selection table and the Linode create Plans table,
```
cpathipa, line 27: Any thoughts or suggestions on consolidating the functions `filterPlanSelectionsByType` and `filterLKEPlansSelectionsByType` into a single utility function would be appreciated. Currently, there are some type errors among the `DatabaseCreate`, `SelectPlanPanel`, and `SelectPlanQuantityPanel` components as some of the types are mismatching when served as a single utility function.

bnussman-akamai: I think a Typescript generic may be helpful to you here. 

Here is what I think `packages/manager/src/utilities/filterPlanSelectionsByType.ts` could look like...

```typescript
import type { LinodeTypeClass } from '@linode/api-v4';

type FilteredPlansTypes<T> = Record<LinodeTypeClass, T[]>;

export const filterPlanSelectionsByType = <
  T extends { class: LinodeTypeClass }
>(
  types: T[]
) => {
  const filteredLKEPlansTypes: FilteredPlansTypes<T> = {
    nanode: [],
    standard: [],
    highmem: [],
    prodedicated: [],
    dedicated: [],
    gpu: [],
    metal: [],
    premium: [],
  };

  for (const type of types) {
    switch (type.class) {
      case 'nanode':
        filteredLKEPlansTypes.nanode.push(type);
        break;
      case 'standard':
        filteredLKEPlansTypes.standard.push(type);
        break;
      case 'highmem':
        filteredLKEPlansTypes.highmem.push(type);
        break;
      case 'prodedicated':
        filteredLKEPlansTypes.prodedicated.push(type);
        break;
      case 'dedicated':
        filteredLKEPlansTypes.dedicated.push(type);
        break;
      case 'gpu':
        filteredLKEPlansTypes.gpu.push(type);
        break;
      case 'metal':
        filteredLKEPlansTypes.metal.push(type);
        break;
      case 'premium':
        filteredLKEPlansTypes.premium.push(type);
        break;
      default:
        break;
    }
  }

  return filteredLKEPlansTypes;
};
```

cpathipa: Thank you @bnussman-akamai

hkhalil-akamai (approved):
Looks good, thanks for adding tests!

packages/manager/src/utilities/filterPlanSelectionsByTypes.test.ts
```diff
@@ -0,0 +1,21 @@
+import { getPlanSelectionsByPlanType } from './filterPlanSelectionsByType';
+import { extendedTypes } from 'src/__data__/ExtendedType';
```
bnussman-akamai, line 2: Would it be possible to use a factory instead of stuff from the `src/__data__/` folder? We are trying to deprecate using `src/__data__/` in favor of factories.

bnussman-akamai (approved):

PR #9102, opened 2023-05-09
refactor: [M3-6347] – MUI v5 - Components > IconButton

bnussman-akamai:
## Description 📝

Style migration for `<IconButton />`

## Major Changes 🔄
- Removes any "wrapping" to the `<IconButton />` because it was not needed
- This PR is a great example of why not to wrap components when MUI's theming can handle everything
- Updates imports in consuming components
- Removes `src/components/core/IconButton` because it acted as an unnecessary second source of abstraction 

## Preview 📷

Here's one example of an `<IconButton />`

Screenshot 2023-05-09 at 11 31 23 AM

## How to test 🧪
- Check for UI differences as compared to production

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 - MUI v5 Migration - `Components > DownloadCSV` #9084
 - MUI v5 Migration - `Components > Notice` #9094
 - MUI v5 Migration - `Components > IconButton` #9102
+- MUI v5 Migration - `Components > CheckoutSummary` #9100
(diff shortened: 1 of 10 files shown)


packages/manager/src/themeFactory.ts
```diff
@@ -835,6 +835,9 @@ export const base: ThemeOptions = {
           marginRight: 0,
         },
       },
+      defaultProps: {
+        size: 'large',
+      },
```
bnussman-akamai, line 840: Doing this because previously we enforced `size="large"` in `packages/manager/src/components/IconButton/IconButton.tsx` (which is now deleted)

jaalah-akamai (approved):
👍

cpathipa (approved):

PR #9103, opened 2023-05-09
refactor: [M3-6393] - MUI v5 - `Components > Snackbar`

jaalah-akamai:
## Description 📝
Refactors `Snackbar` to use `styled` api and cleaned up the overall usage of the component.

## Major Changes 🔄
- Renamed `SnackBar` to `Snackbar` to follow naming convention
- Update to latest version of `"notistack": "^3.0.1"`
- Migrated `withSnackbar` to just call `enqueueSnackbar` directly
- `WithSnackbarProps` is no longer needed, but if it is, we can use `ProviderContext` from `notistack`
- Converted `ToastNotifications` Storybook MDX to 7.0 CSF
- Using named export/imports

## Pending Changes
- [x] Success color is now green instead of blue (pending UX review)
    - Edit: Keeping as blue until UX revisits this.

## Preview 📷
Should be no visual changes

Screen Shot 2023-05-09 at 1 50 23 PM

## How to test 🧪
**Option 1:**
1. Run storybook `yarn storybook`
2. Go to http://localhost:6006/?path=/story/components-notifications-toasts--default

**Option 2:**
1. Run app locally and start a migration or resize on a linode (or anything that triggers a toast). See list here: https://github.com/linode/manager/blob/develop/packages/manager/src/features/ToastNotifications/ToastNotifications.tsx#L78
2. Verify toast messages still appear


packages/manager/src/components/FileUploader/FileUploader.tsx
```diff
@@ -210,17 +210,17 @@ const FileUploader: React.FC<CombinedProps> = (props) => {
     }) exceeded`;
 
     if (wrongFileType) {
-      props.enqueueSnackbar(fileTypeErrorMessage, {
+      enqueueSnackbar(fileTypeErrorMessage, {
```
bnussman-akamai, line 213: I'm a bit confused here on the fact that we are importing `enqueueSnackbar` directly but also adding `ProviderContext` to props. 

If we can import `enqueueSnackbar` directly, then do we need `ProviderContext`?  How does this compare to calling `enqueueSnackbar` when using `const { enqueueSnackbar } = useSnackbar();`

jaalah-akamai: Good point, this specific file can actually just use the hook - I'll also make this adjustment for other places where it makes sense. The `ProviderContext` was a search / replace, but yea I agree it may not be needed anymore.

dwiley-akamai (approved):

cpathipa (approved):

jaalah-akamai:
I have a bit more work to do in order to get our tests to pass. The new version of Snackbar isn't allowing `data-*` to be applied to our DOM elements.

jaalah-akamai:
I think the failing tests are due to the new version of `notistack`. I see a few related PRs in their repo waiting to be addressed. Hoping that solves the issues we're seeing here.

jaalah-akamai:
This PR has been open for a while and conflicts would be easier to resolve with a new PR

PR #9108, opened 2023-05-11
chore: [M3-6574] - Send Adobe Analytics Page Views

bnussman-akamai:
## Description 📝

- Send page views to Adobe Analytics
- We are making this change to match what we did for Google Analytics
- This is pretty typical SPAs to need this

## How to test 🧪
- Check out this PR
- Run `_satellite.setDebug(true)` in the dev console
- Refresh the page
- Navigate around the app 
- Verify you see `🚀 "page view" does not match any direct call identifiers.` in the console
  - This sounds like an error, but I think it's just because configuration is needed on the Adobe Analytics side


jaalah-akamai (approved):

mjac0bs (approved):
Thanks for making this change -- confirmed the direct call rule is firing in the console when viewing a page. 

Can we add a changelog entry under "Fixed" for this?

PR #9124, opened 2023-05-16
refactor: [M3-6421 & M3-6422] – MUI v5 Migration - `Components > TypeToConfirm` & `Components > TypeToConfirmDialog`

dwiley-akamai:
## Description 📝
Use styled components in `TypeToConfirm.tsx`; update a few files according to our current code conventions; update import paths accordingly.

## How to test 🧪
Ensure TTC instances throughout the app appear as expected compared to prod.


dwiley-akamai:
My code commits accidentally got included with the changeset commit, but things should be good apart from that.

packages/manager/.changeset/pr-NaN-tech-stories-1684253040635.md
```diff
@@ -0,0 +1,5 @@
+---
+"@linode/manager": Tech Stories
+---
+
+MUI v5 Migration - `Components > TypeToConfirm` and `Components > TypeToConfirmDialog` ([#NaN](https://github.com/linode/manager/pull/NaN))
```
jaalah-akamai, line 5: May have to run the changeset again to remove `NAN`?

packages/manager/src/components/TypeToConfirm/TypeToConfirm.tsx
```diff
@@ -20,13 +19,7 @@ export interface Props {
   [propName: string]: any;
```
bnussman-akamai, line 19: Unrelated to your PR, but yikes... We should fix this

packages/manager/src/components/TypeToConfirm/TypeToConfirm.tsx
```diff
@@ -67,16 +58,15 @@ const TypeToConfirm: React.FC<Props> = (props) => {
         </>
       ) : null}
       {!hideInstructions ? (
-        <Typography
-          data-testid="instructions-to-enable-or-disable"
-          className={classes.description}
-        >
+        <StyledTypography data-testid="instructions-to-enable-or-disable">
```
bnussman-akamai, line 61: Is it better to prefer styled here?  I'm a big fan of styling via props, but I don't know what direction we want to prefer for things like this going forward.

```typescript
        <Typography
          marginTop={1}
          data-testid="instructions-to-enable-or-disable"
        >
```

bnussman-akamai (approved):

packages/manager/src/components/TypeToConfirm/TypeToConfirm.tsx
bnussman-akamai: Here is an untested fix

```typescript
import { styled } from '@mui/material/styles';
import Link from 'src/components/Link';
import TextField, { Props } from 'src/components/TextField';

export interface TypeToConfirmProps extends Omit<Props, 'onChange'> {
  confirmationText?: JSX.Element | string;
  textFieldStyle?: React.CSSProperties;
  typographyStyle?: React.CSSProperties;
  visible?: boolean | undefined;
  title?: string;
  hideInstructions?: boolean;
  onChange: (value: string) => void;
}

export const TypeToConfirm = (props: TypeToConfirmProps) => {
  const {
    confirmationText,
    textFieldStyle,
    typographyStyle,
    title,
    visible,
    hideInstructions,
    onChange,
    ...rest
  } = props;

  /*
    There was an edge case bug where, when preferences?.type_to_confirm was undefined,
    the type-to-confirm input did not appear and the language in the instruction text
    did not match. If 'visible' is not explicitly false, we treat it as true.
  */

  const showTypeToConfirmInput = visible !== false;
  const disableOrEnable = showTypeToConfirmInput ? 'disable' : 'enable';

  return (
    <>
      {showTypeToConfirmInput ? (
        <>
          <Typography variant="h2">{title}</Typography>
          <Typography style={typographyStyle}>{confirmationText}</Typography>
          <TextField
            style={textFieldStyle}
            onChange={(e) => onChange(e.target.value)}
            {...rest}
          />
        </>
      ) : null}
      {!hideInstructions ? (
        <StyledTypography data-testid="instructions-to-enable-or-disable">
          To {disableOrEnable} type-to-confirm, go to the Type-to-Confirm
          section of <Link to="/profile/settings">My Settings</Link>.
        </StyledTypography>
      ) : null}
    </>
  );
};

const StyledTypography = styled(Typography)(({ theme }) => ({
  marginTop: theme.spacing(),
}));

```

packages/manager/src/components/TypeToConfirm/TypeToConfirm.tsx
dwiley-akamai: This is a good point. I'm not aware of any comparative benefits to preferring the `styled` API in this situation so I don't think we should use use it if the component supports the needed properties like `Typography` does.

jaalah-akamai: This is one of those components that support system props, so technically it's ok. I bet even `mt={1}` would work here. My preference would lean towards the use of the `sx` prop instead.

PR #9125, opened 2023-05-16
refactor: [M3-6335] - MUI v5 Migration - `Components > EntityIcon`

jaalah-akamai:
## Description 📝
Migrate `EntityIcon` to `styled` api and remove `jss`

## Major Changes 🔄
- Update import/exports
- Converted to styled component
- Refactored storybook page
- Renamed `Variant` to `EntityVariants`

## Preview 📷
There should be no visual changes.

Screen Shot 2023-05-16 at 12 14 25 PM

## How to test 🧪
1. `yarn storybook`
2. Go to: http://localhost:6006/?path=/docs/components-entityicon--documentation


bnussman-akamai (approved):

packages/manager/src/components/EntityIcon/EntityIcon.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+  component: EntityIcon,
+  argTypes: {},
+  args: {},
+};
+
+export default meta;
+
+type Story = StoryObj<typeof EntityIcon>;
+
+const variantList = [
```
mjac0bs, line 19: I was noticing that editing the props in Storybook seems to either have no effect (e.g. `size` doesn't change) or, for a few of the apps where the display name and name differ (kubernetes, object storage, marketplace), the Linode icon is shown by default when the `variant` type is changed, rather than the correct icon. 

Entity icons look consistent in Cloud itself.

https://github.com/linode/manager/assets/114685994/5d7e609c-9393-49dc-9355-68cb37e05220

jaalah-akamai: @mjac0bs You're right, there were some inconsistencies in the way we named things - have a look now

mjac0bs: The variant issue looked fixed! I was still seeing that the other properties of the EntityIcon weren't editable via the args table. The changes I suggested seem to fix this, so that I could alter something like the `marginTop` or `size` and see it reflected in Storybook. I might still not be fully understanding the new method of writing stories via autodocs, though, so feel free to correct me if I missed something important here.

jaalah-akamai: This is all good

PR #9130, opened 2023-05-17
fix: [M3-6581] - Fix Cloud Manager Maintenance Mode

bnussman-akamai:
## Description 📝

- Wow this literally took me almost entire workday to figure out 😖
- Fixes API Maintenance mode (I think) 🎉🚀
  - I say I think because I can't perfectly reproduce our production load balancer configuration locally so if other problems lie at the loadbalancer level, maintenance mode may still not work. I'm pretty confident this will fix it.

## Major Changes 🔄
- If the `if` statement that calls `client.identify` can't run, just mark feature flags as loaded in an `else`
  - This allows `MainContent.tsx` to render, therefore allowing `<MaintenanceScreen />` to render when our axios interceptor intercepts the maintenance header

## Preview 📷

> **Note**: Before, Cloud Manager would just show the splash screen loading spinner infinitely 

Screenshot 2023-05-16 at 10 30 43 PM

## How to test 🧪

Contact me, trust me and approve, or wait for testing instructions to be updated

### DIFF
--- a/packages/manager/.changeset/pr-9130-fixed-1684419720777.md
+++ b/packages/manager/.changeset/pr-9130-fixed-1684419720777.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Fixed
+---
+
+fix cloud manager maintenance mode ([#9130](https://github.com/linode/manager/pull/9130))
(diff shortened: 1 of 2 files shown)


jaalah-akamai (approved):
We need to add some cool animations and jazz to this maintenance screen! This is quite disappointing lol...

abailly-akamai (approved):

PR #9140, opened 2023-05-18
fix: [M3-6576]: Fix Bucket Access Component Reloading API Data too often

bnussman-akamai:
## Description 📝

- This one is kind of hard to explain, but `packages/manager/src/features/ObjectStorage/BucketDetail/index.tsx` named imported `<BucketAccess />` which was **not** memoized, which resulted in this component's `useEffect`running more often then it should. It made an API call in the useEffect which caused the loading state to happen every time the parent component updated. 
- This fix is just to remove default imports and memoize the component so that we are guaranteed to import a memoized version of the component.
- Another fix probably could have been to React Query ify `<AccessSelect />` so that is does not rely on props and useEffects to track its state
  - I'll make a ticket and react query-ify this at a later date 

## The Bug 🐛

https://github.com/linode/manager/assets/115251059/34a74534-9dec-4aab-8f4b-be46d659dee7

## How to test 🧪

> **Note**: To recreate the bug, you must have an event that is being polled

- Create a Linode (or trigger any long-running event that will be polled)
- **_Quickly_** go to a valid`https://cloud.linode.com/object-storage/buckets/:cluster/:bucket/access` page
- Verify the loading state does not repeatedly happen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 ### Fixed:
 - LISH Console via SSH containing `None` as the username #9148
 - Inability of restricted users with NodeBalancer creation permissions to add NodeBalancers #9150
+- Bucket Access unnecessarily refreshing #9140
 
 ## [2023-05-22] - v1.93.2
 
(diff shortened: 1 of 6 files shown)


hkhalil-akamai (approved):
Tested, looks good! `BucketAccess` might be a good candidate for RQ-ification.

abailly-akamai (approved):
nice fix 👍

jaalah-akamai (approved):
Approved

hkhalil-akamai (approved):

abailly-akamai (approved):

PR #9143, opened 2023-05-21
hotfix: [M3-6594] - Enable linode action menu "reboot" on mobile

jaalah-akamai:
## Description 📝
Seems to be a slight bug introduced with [a refactor](https://github.com/linode/manager/commit/b9fcf70f3d51094067b578fd10fbd2fd0da7c1ac#diff-af5757c9f2be53593a6a9655c8ea746d322078c01a070da0400f8325d4e127b3L167-R125). We no longer should need to check for `matchesSmDown` anymore.

## Major Changes 🔄
- Reboot should be disabled if there's readonly access or if the linodes not running

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screen Shot 2023-05-21 at 6 13 29 PM | Screen Shot 2023-05-21 at 6 14 18 PM |

## How to test 🧪
1. Go to a linode detail page
2. Reduce screen size below < 960px
3. Open action menu and observe "Reboot" is enabled

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [2023-05-22] - v1.93.2
+
+### Fixed:
+- Issue where linode "Reboot" button was disabled #9143
+
 ## [2023-05-18] - v1.93.1
 
 ### Fixed:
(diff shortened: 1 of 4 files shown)


hkhalil-akamai (approved):
LGTM. I still can't figure out what the original condition before the refactor did:

```
(!hasMadeConfigsRequest && matchesSmDown)
```

Wonder why the viewport size would have anything to do with the reboot action...

bnussman-akamai (approved):
Interesting. Thank you for the fix!

cpathipa (approved):
Thank you @jaalah for the quick fix. Confirming that "reboot" is visible in small and medium screens.

PR #9145, opened 2023-05-22
change: [OCA-1137] - deprecate unifi marketplace app

tbaka:
## Description 📝
**Deprecates Unifi Network Manager Marketplace App**

## Major Changes 🔄
- remove Unifi Network app


jaalah-akamai:
@tbaka I think we can comment this reference too: https://github.com/linode/manager/blob/develop/packages/manager/src/features/OneClickApps/FakeSpec.ts#L2102-L2123

If this is an app that won't be returning, we might be able to remove the logos too:
https://github.com/linode/manager/blob/develop/packages/manager/public/assets/white/unifi.svg
https://github.com/linode/manager/blob/develop/packages/manager/public/assets/unifi.svg

jaalah-akamai (approved):
🚀

hana-linode (approved):
~We can just remove the commented out code~ just pushed the changes!

cpathipa (approved):

PR #9147, opened 2023-05-22
refactor: [M3-6585] - Allow Cypress region selection to be overridden

jdamore-linode:
## Description 📝
**Brief description explaining the purpose of the changes**
This modifies the way that Cypress tests handle regions when creating resources, etc. Additionally, it allows the chosen region to be overridden (instead of the typical behavior of selecting regions at random) using the `CY_TEST_REGION_ID` and `CY_TEST_REGION_NAME` environment variables.

Some tests (like Obj tests) still use hardcoded regions since they're not available in all regions.

## How to test 🧪
Make sure tests run as expected with no regions defined:

```bash
yarn cy:run
```

Make sure tests run as expected with a region defined:

```bash
CY_TEST_REGION_ID=us-east CY_TEST_REGION_NAME='Newark, NJ' yarn cy:run
```


jaalah-akamai (approved):
Well, it took  an hour or so to run all the specs, but they all passed except a `account/security-questions.spec.ts` which I would expect to timeout on the security questions

bnussman-akamai (approved):

PR #9151, opened 2023-05-22
hotfix: Linode Multi Select not working with large filters

bnussman-akamai:
## Description 📝

- Uses a fetch all for the Linode Multi Select because the Linode API could not handle large filters when filtering out linodes
- As a fix for now, I think we can just let the client do the searching and filtering rather than the API
- This fixes the ability for some users to attach linodes to firewalls that already consist of hundreds of linodes

## How to test 🧪
- Verify you can add and remove linodes to a firewall
- Check the functionality of the Select component that displays these linodes

### DIFF
--- a/packages/manager/src/components/LinodeMultiSelect/LinodeMultiSelect.tsx
+++ b/packages/manager/src/components/LinodeMultiSelect/LinodeMultiSelect.tsx
@@ -1,9 +1,8 @@
 import * as React from 'react';
-import { useInfiniteLinodesQuery } from 'src/queries/linodes/linodes';
+import { useAllLinodesQuery } from 'src/queries/linodes/linodes';
 import Autocomplete from '@mui/material/Autocomplete';
 import Popper from '@mui/material/Popper';
 import TextField from 'src/components/TextField';
-import { isNumeric } from 'src/utilities/stringUtils';
 
 export interface Props {
   allowedRegions?: string[];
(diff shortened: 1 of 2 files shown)


packages/manager/src/components/LinodeMultiSelect/LinodeMultiSelect.tsx
```diff
@@ -28,30 +27,6 @@ const LinodeMultiSelect = (props: Props) => {
...
-          : [
-              { label: { '+contains': inputValue } },
-              { tags: { '+contains': inputValue } },
-            ],
-      }
-    : {};
-
-  const linodesFilter =
-    filteredLinodes && filteredLinodes.length > 0
-      ? { '+and': filteredLinodes.map((id) => ({ id: { '+neq': id } })) }
```
bnussman-akamai, line 52: If this component was passed hundreds of `filteredLinodes`, the API would not respond with data. It is likely that the API can't handle filters that are this large. 

This is what caused the bug

jdamore-linode (approved):
I haven't been able to reproduce the original issue yet but I did confirm that this (and the select) work

jaalah-akamai (approved):
Thanks @bnussman-akamai for putting together the test branch - Confirmed this fixes the issue

hkhalil-akamai (approved):
Testing looks good.

PR #9156, opened 2023-05-23
chore: Resolve `yaml` dependabot alert `CVE-2023-2251`

bnussman-akamai:
## Description 📝

- Resolve `yaml` dependabot alert CVE-2023-2251
- Removed some yarn resolutions that are no longer needed
- Updates `patch-package` and `lint-staged` because they depended on an old `yaml` but we still need a resolution thanks to the following packages
  - `react-select` depends on the old yaml version
  - `tsup` somehow still depends on  the old yaml version

## How to test 🧪
- Check basic repo functionality
- Verify that the yarn resolutions I removed do not result in a downgrade of those respective packages
- Consider testing `patch-package` and `lint-staged` because they were upgraded
  - I tested them as I made commits to this PR


yarn.lock
```diff
@@ -8415,7 +8405,7 @@ gunzip-maybe@^1.4.2:
     pumpify "^1.3.3"
     through2 "^2.0.3"
 
-handlebars@^4.4.3, handlebars@^4.7.7:
+handlebars@^4.7.7:
```
dwiley-akamai, line 8408: Shouldn't `handlebars` have gone away here?

dwiley-akamai (approved):

yarn.lock
bnussman-akamai: Storybook still depends on `handlebars "^4.7.7"` so it will still be in the `yarn.lock`

jaalah-akamai (approved):

PR #9158, opened 2023-05-23
poc: Cleaner way to handle MSW endpoints in `serverHandlers.ts`

bnussman-akamai:
## Description 📝

- Simple proof of concept for a more simple way to maintain mock service worker endpoints

## Major Changes 🔄
- Uses an array of objects to maintain msw endpoints

## How to test 🧪
- Turn on the MSW


bnussman-akamai:
would like to hear thoughts from the team if this is worth continuing

PR #9161, opened 2023-05-23
fix: [M3-6586] - Improve error handing for loadScript and Adobe Analytics

mjac0bs:
## Description 📝
This is an attempt to improve the `loadScript` function and error handling for its use with Adobe Analytics.

## Major Changes 🔄
- Removes the error logged when a custom event fires unsuccessfully
   - This will get rid of the annoying Sentry errors that are currently being triggered when any user blocking the analytics scripts performs an action that would fire a custom event to track, if analytics were enabled.
- Modifies `loadScript` function to return a promise.
- Attempts to verify that all three scripts necessary for Adobe (the launch script and its two extensions) are present in the DOM if the script claims it has loaded successfully with a resolved promise.
   - "Attempts" because I haven't found a way to successfully stop all 3 scripts from loading, even with `assets.adobedtm.com/extensions/*`requests blocked, to come closer to mimicking what we saw with the extensions not fully loading (200 response but in `pending` state and not present in the DOM) a week ago.

## How to test 🧪
    - Check this PR out locally and have the REACT_APP_ADOBE_ANALYTICS_URL environment variable defined in your .env.
   - Type `_satellite.setDebug(true)` in the developer console and refresh the page. You should see some debug logs to indicate that Adobe Analytics is running.
   - To verify an error on custom events is logged:
      - Check out `develop` locally.
      - Use dev tools to block requests to `assets.adobedtm.com`.
      - Trigger a custom event (e.g. create a linode or volume).  
      - Observe that [an exception](https://github.com/linode/manager/pull/9161/files#diff-a30d9e66e4612f3a3181b2cabe2770862cef00550bb1797c8faca8a602deb9f5L28) is logged to the browser console; in prod, this would be logged to Sentry. 
3. **How to verify changes?**
    - To verify an error on custom events is *not* logged:
      - Use dev tools to block requests to `assets.adobedtm.com`.
      - Trigger a custom event (e.g. create a linode or volume).  
      - Observe that [an exception](https://github.com/linode/manager/pull/9161/files#diff-a30d9e66e4612f3a3181b2cabe2770862cef00550bb1797c8faca8a602deb9f5L28) is no longer logged.
   - Check that there have been no regressions in the places the app calls `useScript`, such as the Google Pay button and chip on the Billing page. (Try blocking network requests and testing.) I had to slightly modify this hook when I changed the `loadScript` function.
   - With requests to `assets.adobedtm.com` blocked, confirm that the app works as expected and no new console errors/warnings are logged.

### DIFF
--- a/packages/manager/src/App.tsx
+++ b/packages/manager/src/App.tsx
@@ -94,14 +94,12 @@ export class App extends React.Component<CombinedProps, State> {
           const adobeScriptTags = document.querySelectorAll(
             'script[src^="https://assets.adobedtm.com/"]'
           );
-
           // Log an error; if the promise resolved, the _satellite object and 3 Adobe scripts should be present in the DOM.
           if (
             data.status !== 'ready' ||
             !(window as any)._satellite ||
             adobeScriptTags.length !== NUM_ADOBE_SCRIPTS
           ) {
-            adobeScriptTags.forEach((script) => script.remove());
             reportException(
               'Adobe Analytics error: Not all Adobe Launch scripts and extensions were loaded correctly; analytics cannot be sent.'
             );
--- a/packages/manager/src/App.tsx
+++ b/packages/manager/src/App.tsx
@@ -94,11 +94,16 @@ export class App extends React.Component<CombinedProps, State> {
           const adobeScriptTags = document.querySelectorAll(
             'script[src^="https://assets.adobedtm.com/"]'
           );
-          // Log an error; if the promise resolved, three Adobe scripts should be present in the DOM.
-          if (data.status !== 'ready' || adobeScriptTags.length !== 3) {
+
+          // Log an error; if the promise resolved, the _satellite object and 3 Adobe scripts should be present in the DOM.
+          if (
+            data.status !== 'ready' ||
+            !(window as any)._satellite ||
+            adobeScriptTags.length !== 3
+          ) {
             adobeScriptTags.forEach((script) => script.remove());
             reportException(
-              'Not all Adobe Launch scripts and extensions were not loaded correctly; analytics cannot be sent.'
+              'Adobe Analytics error: Not all Adobe Launch scripts and extensions were loaded correctly; analytics cannot be sent.'
             );
           }
         })
--- a/packages/manager/src/App.tsx
+++ b/packages/manager/src/App.tsx
@@ -44,6 +44,7 @@ import { oauthClientsEventHandler } from './queries/accountOAuth';
 import { ADOBE_ANALYTICS_URL } from './constants';
 import { linodeEventsHandler } from './queries/linodes/events';
 import { supportTicketEventHandler } from './queries/support';
+import { reportException } from './exceptionReporting';
 
 interface Props {
   location: RouteComponentProps['location'];
@@ -88,7 +89,22 @@ export class App extends React.Component<CombinedProps, State> {
 
     // Load Adobe Analytics Launch Script
     if (!!ADOBE_ANALYTICS_URL) {
-      loadScript(ADOBE_ANALYTICS_URL, { location: 'head' });
+      loadScript(ADOBE_ANALYTICS_URL, { location: 'head' })
+        .then((data) => {
+          const adobeScriptTags = document.querySelectorAll(
+            'script[src^="https://assets.adobedtm.com/"]'
+          );
+          // Log an error; if the promise resolved, three Adobe scripts should be present in the DOM.
+          if (data.status !== 'ready' || adobeScriptTags.length !== 3) {
+            adobeScriptTags.forEach((script) => script.remove());
+            reportException(
+              'Not all Adobe Launch scripts and extensions were not loaded correctly; analytics cannot be sent.'
+            );
+          }
+        })
+        .catch(() => {
+          // Do nothing; a user may have analytics script requests blocked.
+        });
     }
 
     /**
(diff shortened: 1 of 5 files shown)


packages/manager/src/App.tsx
```diff
@@ -88,7 +89,27 @@ export class App extends React.Component<CombinedProps, State> {
...
+          // Log an error; if the promise resolved, the _satellite object and 3 Adobe scripts should be present in the DOM.
+          if (
+            data.status !== 'ready' ||
+            !(window as any)._satellite ||
+            adobeScriptTags.length !== 3
+          ) {
+            adobeScriptTags.forEach((script) => script.remove());
+            reportException(
+              'Adobe Analytics error: Not all Adobe Launch scripts and extensions were loaded correctly; analytics cannot be sent.'
+            );
```
mjac0bs, line 107: I logged an exception at this point because this would indicate that the analytics scripts were not blocked by the user and did (supposedly) load on the page, but still aren't set up as expected, so something is really wrong. I haven't been able to figure out a way to trigger any of these conditions, though, and am open to feedback on whether we'd want to log this to Sentry at all.

jdamore-linode (approved):

packages/manager/src/App.tsx
```diff
@@ -88,7 +89,27 @@ export class App extends React.Component<CombinedProps, State> {
...
+            'script[src^="https://assets.adobedtm.com/"]'
+          );
+
+          // Log an error; if the promise resolved, the _satellite object and 3 Adobe scripts should be present in the DOM.
+          if (
+            data.status !== 'ready' ||
+            !(window as any)._satellite ||
+            adobeScriptTags.length !== 3
+          ) {
+            adobeScriptTags.forEach((script) => script.remove());
```
jdamore-linode, line 104: I know this is all hypothetical and challenging to actually test, but do we know if this actually makes any impact? If the launch script has already loaded and set up the `_satellite` object, do we know that removing the `<script />` element actually does anything to stop its execution going forward?

mjac0bs: If the scripts don't exist, the `_satellite` object doesn't either -- I think it's loaded in the AppMeasurement.tsx extension. The thought behind removing all the scripts that do exist at this point is mostly that if the launch script were to load, but either/both of the extensions, for some reason, didn't, then we'd just remove the launch script all together since it won't work without the extensions (and maybe prevent a crash from continuing to try to load them?). 

But this is hypothetical, you're right, and if there are reasons we'd want to consider not doing this, I'm happy to rethink this.

mjac0bs:
@dwiley-akamai My bad on not specifying what the exception was more clearly in the repro steps (just fixed). The exception you should see on develop is [this one](https://github.com/linode/manager/pull/9161/files#diff-a30d9e66e4612f3a3181b2cabe2770862cef00550bb1797c8faca8a602deb9f5L28) when blocking requests. And on the PR’s branch, you shouldn’t see the above exception or [this exception](https://github.com/linode/manager/pull/9161/files#diff-106b5b793d9c13dae5657b9f0bbb47f6e358158c5c6faf7adfe8aa26467dc55eR106) just because you’ve blocked analytics from tracking. You will still see the Adobe debug log error in both cases - we don't control that one.

packages/manager/src/App.tsx
```diff
@@ -88,7 +89,27 @@ export class App extends React.Component<CombinedProps, State> {
...
+        .then((data) => {
+          const adobeScriptTags = document.querySelectorAll(
+            'script[src^="https://assets.adobedtm.com/"]'
+          );
+
+          // Log an error; if the promise resolved, the _satellite object and 3 Adobe scripts should be present in the DOM.
+          if (
+            data.status !== 'ready' ||
+            !(window as any)._satellite ||
+            adobeScriptTags.length !== 3
```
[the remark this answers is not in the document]
mjac0bs: Yeah, that's a good question -- asked internally about this and am waiting on a response. 

Thinking of a scenario where a decision is made to modify the launch script and add another extension... if that was communicated to our team in advance, we'd need to remember to update this number (which I should have defined in `constants.ts`). If, for some reason, anything changed with Adobe unexpectedly and we had a different number of scripts, it would stop data collection. In either case, I wouldn't expect a *frequent* change to the number of scripts. We *could* opt to not remove any scripts, just log the error, and hopefully rely on monitoring tests in the future to let us know if scripts are causing any issues with loading the app.

packages/manager/src/hooks/useScript.ts
```diff
@@ -13,62 +13,62 @@ interface ScriptOptions {
...
+export const loadScript = (
+  src: string,
+  options?: ScriptOptions
+): Promise<any> => {
+  return new Promise((resolve, reject) => {
+    // Allow falsy src value if waiting on other data needed for
+    // constructing the script URL passed to this hook.
+    if (!src) {
+      options?.setStatus?.('idle');
+      return;
```
hkhalil-akamai, line 27: Should there be a `resolve()` here?

hkhalil-akamai (approved):
Confirming that the exception is no longer thrown.

cpathipa (approved):
Changes LGTM! I don't see any regressions in Google pay and billing page.
Observation: Was not able to produce the exception either in develop as well as PR branch.

packages/manager/src/App.tsx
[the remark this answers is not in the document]
mjac0bs: I took out the logic to completely remove the scripts for the time being considering this point and Joe's question. Keeping the exception as is to log the issue, though.

PR #9163, opened 2023-05-23
test: [M3-6549] - Add Cypress Integration tests for OAuth Apps

cliu-akamai:
## Description 📝
Add new cypress tests for OAuth Apps

## Major Changes 🔄
- Add `OAuth Apps` addition tests.
- Add `OAuth Apps` editing  tests.
- Add `OAuth Apps` resetting tests.
- Add `OAuth Apps` deletion tests.

## How to test 🧪
`yarn cy:run -s "cypress/e2e/profile/*.spec.ts"`

### DIFF
--- a/packages/api-v4/src/account/types.ts
+++ b/packages/api-v4/src/account/types.ts
@@ -373,6 +373,7 @@ export interface OAuthClient {
   thumbnail_url: string | null;
   public: boolean;
   status: 'disabled' | 'active' | 'suspended';
+  secret: string | null;
 }
 
 export interface OAuthClientRequest {
--- a/packages/api-v4/src/account/types.ts
+++ b/packages/api-v4/src/account/types.ts
@@ -373,7 +373,7 @@ export interface OAuthClient {
   thumbnail_url: string | null;
   public: boolean;
   status: 'disabled' | 'active' | 'suspended';
-  secret: string | null;
+  secret: string;
 }
 
 export interface OAuthClientRequest {
--- a/packages/manager/cypress/e2e/profile/add-oauth-app.spec.ts
+++ b/packages/manager/cypress/e2e/profile/add-oauth-app.spec.ts
@@ -1,9 +1,14 @@
 import { fbltClick } from 'support/helpers';
 import { oauthClientFactory } from '@src/factories';
 import 'cypress-file-upload';
-import { interceptGetProfile } from 'support/intercepts/profile';
+import {
+  interceptGetProfile,
+  mockCreateOAuthApp,
+  mockGetOAuthApps,
+} from 'support/intercepts/profile';
 import { ui } from 'support/ui';
 import { randomLabel } from 'support/util/random';
+import { OAuthClient } from '@linode/api-v4/types';
 
 /**
  * Creates an OAuth App with the given parameters.
@@ -11,11 +16,9 @@ import { randomLabel } from 'support/util/random';
  * This assumes that the user has already navigated to the profile/clients
  * page.
  *
- * @param label - Label for the oauth app.
- * @param url - Callback URL for the oauth app.
- * @param isPublic - Access for the oauth app.
+ * @param oauthClient - OAuth app to mock.
  */
-const createOAuthApp = (label: string, url: string, isPublic: boolean) => {
+const createOAuthApp = (oauthApp: OAuthClient) => {
   ui.button
     .findByTitle('Add an OAuth App')
     .should('be.visible')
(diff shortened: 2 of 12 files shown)


packages/manager/cypress/support/util/random.ts
```diff
@@ -226,3 +226,23 @@ export const randomUuid = (): string => {
...
+ * Returns a random secret of fixed length.
+ *
+ * @param length - Length of the secret strings.
+ *
+ * @returns Random secret.
+ */
+export const randomSecret = (length: number = 64): string => {
+  const hexNumber = '0123456789abcdef';
+
+  const characterSelection = [hexNumber].join('').split('');
```
jdamore-linode, line 240:
```suggestion
  const characterSelection = hexNumber.split('');
```
I think this could be simplified a bit! Is there anything major that distinguishes this from the `randomString()` util? If I'm understanding correctly we might be able to use `randomString(64)`

jdamore-linode (changes requested):
Great start Cassie! I haven't been able to actually run your tests right now due to unrelated issues so I will circle back later this afternoon once everything's working, but in the meantime I left a few suggestions

packages/api-v4/src/account/types.ts
```diff
@@ -373,6 +373,7 @@ export interface OAuthClient {
   thumbnail_url: string | null;
   public: boolean;
   status: 'disabled' | 'active' | 'suspended';
+  secret: string | null;
```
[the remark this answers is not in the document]
cliu-akamai: Got it.

jaalah-akamai (approved):

packages/manager/cypress/e2e/profile/add-oauth-app.spec.ts
jdamore-linode, line 1: This test doesn't seem to be using mock data and is actually creating real OAuth apps. I think we're going to want some mock utils in `intercepts/profile.ts` for this since we don't want Cypress to actually create and delete OAuth apps (mostly due to concerns with having the secrets exposed in the recordings).

hkhalil-akamai:
Looks really good. Just one minor concern: the tests do not check that the dialog/drawer for the add/edit/reset/delete actions actually close when the action is completed/cancelled (e.g., by checking that the dialog/drawer is no longer rendered or visible).

Also, should the tests try closing the dialog/drawer using the 'X' in the corner?
Screenshot 2023-06-08 at 6 54 00 PM
Screenshot 2023-06-08 at 6 54 32 PM

cliu-akamai:
> Looks really good. Just one minor concern: the tests do not check that the dialog/drawer for the add/edit/reset/delete actions actually close when the action is completed/cancelled (e.g., by checking that the dialog/drawer is no longer rendered or visible).
> 
> Also, should the tests try closing the dialog/drawer using the 'X' in the corner? Screenshot 2023-06-08 at 6 54 00 PM Screenshot 2023-06-08 at 6 54 32 PM

That's a good point, I will update that.

packages/manager/cypress/e2e/profile/add-oauth-app.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    .should('be.enabled')
+    .click();
+  ui.drawer
+    .findByTitle('Create OAuth App')
+    .should('be.visible')
+    .within(() => {
+      fbltClick('Label').clear().type(oauthApp.label);
+      fbltClick('Callback URL').clear().type(oauthApp.redirect_uri);
+    });
+  ui.drawerCloseButton.find().click();
```
hkhalil-akamai, line 59: Thank you for testing the 'X' button in Add and Edit, can you add this check to the Delete and Reset dialog as well?

hkhalil-akamai (approved):
Thanks for making the changes! I think you might have missed testing the 'X' button in Delete and Reset. It would be great if you could add those checks, but I will approve now so you don't have to wait on me!

packages/manager/cypress/support/api/profile.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Deletes all oauth apps which are prefixed with the test entity prefix.
+ *
+ * @returns Promise that resolves when oauth apps have been deleted.
+ */
+export const deleteAllTestOAuthApps = async (): Promise<void> => {
+  const oauthApps = await depaginate<OAuthClient>((page: number) =>
+    getOAuthClients({ page_size: pageSize, page })
+  );
+
+  console.log(`oauth apps: ${JSON.stringify(oauthApps)}`);
```
jdamore-linode, line 20:
```suggestion
```
I don't think this is too big a deal since the secret gets redacted but probably good to get rid of this just in case!

packages/manager/cypress/support/ui/common.ts
```diff
@@ -87,6 +88,7 @@ export const deleteAllTestData = async () => {
       deleteAllTestAccessKeys(),
       deleteAllTestTags(),
       deleteAllTestVolumes(),
+      deleteAllTestOAuthApps(),
```
jdamore-linode, line 91: I don't think we really need this since the tests don't make OAuth apps anymore, but it doesn't hurt anything by being there either

jdamore-linode (approved):
Nice work @cliu-akamai !

PR #9169, opened 2023-05-24
refactor: [M3-6389] - MUI v5 Migration - `Components > ShowMore`

jaalah-akamai:
## Description 📝
Migrate `ShowMore` to `styled/sx` api and remove `jss`

## Major Changes 🔄
- Added new storybook page
- Updated imports/exports
- Converted class to functional component
- Removed hardcoded classes: 
    - `.chip` → `.MuiChip-root`
    - `.active` → Uses `sx` prop to handle this now

## Preview 📷
No visual changes

| Storybook Page   | Search Page
| ------- | ------- |
| Screen Shot 2023-05-24 at 1 08 40 PM | Screen Shot 2023-05-24 at 1 18 18 PM |

## How to test 🧪
1. Go to http://localhost:3000/search?query=debian
2. Ensure you have a linode with lots of tags to see the `ShowMore`

### DIFF
--- a/packages/manager/src/components/ShowMore/ShowMore.tsx
+++ b/packages/manager/src/components/ShowMore/ShowMore.tsx
@@ -1,18 +1,18 @@
-import classNames from 'classnames';
 import * as React from 'react';
 import Chip, { ChipProps } from 'src/components/core/Chip';
 import Popover from '@mui/material/Popover';
-import { styled } from '@mui/material/styles';
+import { styled, useTheme } from '@mui/material/styles';
 
 interface ShowMoreProps<T> {
+  ariaItemType: string;
+  chipProps?: ChipProps;
   items: T[];
   render: (items: T[]) => any;
-  chipProps?: ChipProps;
-  ariaItemType: string;
 }
 
 export const ShowMore = <T extends unknown>(props: ShowMoreProps<T>) => {
   const { render, items, chipProps, ariaItemType } = props;
+  const theme = useTheme();
   const [anchorEl, setAnchorEl] = React.useState<HTMLElement | null>(null);
 
   const handleClick = (event: React.MouseEvent<HTMLElement>) => {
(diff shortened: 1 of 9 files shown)


bnussman-akamai (commented):
Going to take a look. Can you fix the failing unit tests?

packages/manager/src/components/ShowMore/ShowMore.tsx
```diff
@@ -1,133 +1,105 @@
...
-  ariaItemType: string;
 }
 
-export class ShowMore<T> extends React.Component<
-  Props<T> & WithStyles<CSSClasses>
-> {
-  state = {
-    anchorEl: undefined,
-  };
+export const ShowMore = <T extends unknown>(props: ShowMoreProps<T>) => {
```
dwiley-akamai, line 13: Would [`T extends {}`](https://stackoverflow.com/questions/61648189/typescript-generic-type-parameters-t-vs-t-extends) be preferable here?

dwiley-akamai (commented):
LGTM aside from the aforementioned unit tests

packages/manager/src/components/ShowMore/ShowMore.tsx
jaalah-akamai: In this case, it should be good 👍

dwiley-akamai (approved):
Screenshot 2023-05-31 at 9 55 51 AM

This test failed during the most recent run, but going to approve this PR pending a fix for that test

jaalah-akamai:
@bnussman-akamai @dwiley-akamai fixed and converted the test to RTL

cpathipa:
Observations:
- I feel like the 'Display all tags (Kabab button)' should appear as the last item instead of the first.
- The font of the last tag ('test5') appears faded.
- The hover background of the delete button for the last tag is not consistent with the other tag buttons.

image

hover background of last tag:
image
hover background of other tags:
image

cpathipa (commented):
Overall LGTM! left few observations as noticed.

jaalah-akamai:
@cpathipa 

1. That's been the layout since I arrived (below are SS from February) - We can run it past UX to see if they would like to change it.
2. That's because when there's a "more" button, there's also a fade that goes over the buttons (always been there as well)
3. Same as point 2


| Large Breakpoint | Small Breakpoint |
| ----------- | ----------- |
| Screen Shot 2023-06-01 at 7 59 30 PM | Screen Shot 2023-06-01 at 8 00 04 PM |

cpathipa:
> @cpathipa
> 
> 1. That's been the layout since I arrived (below are SS from February) - We can run it past UX to see if they would like to change it.
> 2. That's because when there's a "more" button, there's also a fade that goes over the buttons (always been there as well)
> 3. Same as point 2
> 
> Large Breakpoint	Small Breakpoint
> Screen Shot 2023-06-01 at 7 59 30 PM	Screen Shot 2023-06-01 at 8 00 04 PM

Thank you @jaalah-akamai yea, its existing behavior,  spoke with Andrew and created a ticket to handle them separately. 
- The gradient look of the close button on the last tag is barely visible to naked eye.
image
- For smaller screens, more buttons should be positioned to the right of the tags.

cpathipa (approved):

PR #9171, opened 2023-05-24
feat: [M3-6077] - Add Maintenance Mode Environment Variable and Clean Up `<MaintenanceScreen />`

bnussman-akamai:
## Description 📝

- Adds `REACT_APP_ENABLE_MAINTENANCE_MODE` to allows us to explicitly enable maintenance mode
- Cleans up  and modernizes `<MaintenanceScreen />`
- Makes the 🔧 a better color so it looks good in dark mode
- Removes the word `scheduled` from the copy because it may not always be true

## Preview 📷

Screenshot 2023-05-24 at 2 06 37 PM

## How to test 🧪
- Checkout this PR
- Spin up dev server with `yarn up` or `yarn dev`
- Open Cloud Manager to any page
- Add `REACT_APP_ENABLE_MAINTENANCE_MODE=true` to your `packages/manager/.env`
- Verify your Cloud Manager refreshed and went into maintenance mode
- Comment out `REACT_APP_ENABLE_MAINTENANCE_MODE=true`
- Verify that Cloud Manager is no longer in maintenance mode


packages/manager/src/MainContent.tsx
```diff
@@ -238,7 +238,10 @@ const MainContent = (props: CombinedProps) => {
   }
 
   // If the API is in maintenance mode, return a Maintenance screen
-  if (props.globalErrors.api_maintenance_mode) {
+  if (
+    props.globalErrors.api_maintenance_mode ||
+    import.meta.env.REACT_APP_MAINTENANCE_MODE
```
mjac0bs, line 243: As far as:
> Add REACT_APP_MAINTENANCE_MODE=true to your packages/manager/.env

Really, we're not evaluating that the env var is "true", we're evaluating that it exists at all, right? It could be `REACT_APP_MAINTENANCE_MODE=42` and we'd still display maintenance mode. I was just thinking about this in the context that someone could mistakenly think that `REACT_APP_MAINTENANCE_MODE=false` would toggle off maintenance, when it would really require removing the env var completely.

mjac0bs (commented):
This looks good in both light and dark mode, and thanks for making that copy change.

packages/manager/src/MainContent.tsx
bnussman: Hmm. What is the best way to go about this. Check for the string `"true"`? Is it even possible to have a truly `boolean` env var anyway?

mjac0bs: > Is it even possible to have a truly boolean env var anyway?

Looks like 'no', and we could either check for the string `"true"` or write a function to convert `"true"` or `true` to an actual `boolean` for the env var.

dwiley-akamai (approved):
LGTM 🚢

mjac0bs (approved):
🚀

jaalah-akamai (approved):

PR #9184, opened 2023-05-26
feat: [M3 6643] - Improve error messages when Image creation or upload fails

mjac0bs:
## Description 📝
It can be difficult for users to identify the specific reason why a Linode Image creation attempt has failed. With more guidance in Cloud Manager, users will be able to quickly identify an issue related to Image size limits without needing to open a ticket.

This ticket focuses on small tweaks we can make to improve our error messaging in the UI based on what is currently returned from the backend.

TODO:
- [ ] Append units at the end of a number.
- [ ] Figure out why image toasts (or all error toasts? or all persistent error toasts?) are displaying multiple times (5?) when they should be displaying 1. 

## Major Changes 🔄
- Updated helper text on Images -> Create -> Upload Image to inform users that the maximum image size is 6 GB (uncompressed).
- Updated `image_upload` and `disk_imagize` event messages on failed action.
- Updated `image_upload` toast message on failed action.
   - Error toast is persistent, requiring the user to close it to dismiss, which ensures the user can see an error has occurred.

## Preview 📷

> **Note**: Use `<video src="" />` tag when including recordings in table

| Before  | After   |
| ------- | ------- |
| Content | Content |

## How to test 🧪
3. **How to verify changes?**


packages/manager/src/eventMessageGenerator.ts
```diff
@@ -272,7 +278,10 @@ export const eventMessageCreators: { [index: string]: CreatorsForStatus } = {
   image_upload: {
     scheduled: (e) => `Image ${e.entity?.label ?? ''} scheduled for upload.`,
     started: (e) => `Image ${e.entity?.label ?? ''} is being uploaded.`,
-    failed: (e) => `There was a problem uploading ${e.entity?.label ?? ''}.`,
+    failed: (e) =>
+      `There was a problem uploading ${
```
abailly-akamai, line 282: Should we consolidate the messaging and choose between:

`Error [creating/uploading] Image [...]` and 
`There was a problem [creating/uploading] [...]`

packages/manager/src/features/Images/ImageUpload.tsx
```diff
@@ -273,7 +273,8 @@ export const ImageUpload: React.FC<Props> = (props) => {
 
           <Typography className={classes.helperText}>
             Image files must be raw disk images (.img) compressed using gzip
-            (.gz). The maximum file size is 5 GB (compressed).
+            (.gz). The maximum file size is 5 GB (compressed) and maximum image
+            size is 6 GB (uncompressed).
```
abailly-akamai, line 277: I would suggest making this a `<Notice warning />` VS a helper text so it's more prominent

packages/manager/src/features/ToastNotifications/ToastNotifications.tsx
```diff
@@ -137,10 +146,9 @@ class ToastNotifications extends React.PureComponent<WithSnackbarProps, {}> {
               successMessage: `Image ${label} is now available.`,
               failureMessage: isDeletion
                 ? undefined
-                : `There was a problem processing image ${label}: ${event.message?.replace(
-                    'cancelled',
-                    'canceled'
-                  )}`,
+                : `There was a problem uploading image ${label}: ${event.message
+                    ?.replace('cancelled', 'canceled')
```
abailly-akamai, line 150: could we make an API ticket for this typo?

abailly-akamai (commented):
Thanks for working on this, looks good! Couple comments and questions

packages/manager/src/features/ToastNotifications/ToastNotifications.tsx
abailly-akamai: actually:

> Canceled or cancelled is the past tense of the verb to cancel. Both spellings are correct; Americans favor canceled (one l), while cancelled (two l's) is preferred in British English and other dialects.

🤷

mjac0bs: Agreed that it's not ideal to be adjusting copy on the front-end when the error message could be updated directly. And while both are correct spellings, I think it's best to maintain consistency throughout Cloud Manager, where we use "canceled" most often. Looks like API returns "cancelled" in a couple of places and we replace it - I can ticket this.

packages/manager/src/features/Images/ImageUpload.tsx
mjac0bs: image

Good thought. I like this change. I'm going to run it by UX to make sure we're in agreement and there's no concern about including the image size text in a warning and leaving the billing rate text below as helper text.

hana-linode (changes requested):
The placement of the error toast links are a bit off in mobile view

image

mjac0bs:
> a bit off in mobile view

@hana-linode Good catch. Default styling of the snackbar wasn't handling flex wrapping well in messages with an additional link component. Adjusted some styling at the smaller breakpoints to fix this - can you have another look?

https://github.com/linode/manager/assets/114685994/d144426f-c40c-4055-9ff2-8956b5895230

PR #9193, opened 2023-05-30
refactor: [M3-6485, M3-6651] - RQ-ify + Refactor Linode Select

hkhalil-akamai:
## Description 📝
This PR introduces a new `LinodeSelect` component that merges the existing functionality of `LinodeSelect` and `LinodeMultiSelect` in a single type-safe and and RQ-based component.

This new component includes a slightly changed design aimed at clearly and explicitly reflecting user selections (pending UX consideration).

Usage of this new component is demonstrated in the `CreateDomain`, `CreateImage`, `CreateFirewallDrawer` (multi-select) and `AddDeviceDrawer` (multi-select) components. As a precautionary measure, other uses of the previous components will be migrated after these changes are validated in prod.

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Single-Select |
| <video src="https://github.com/linode/manager/assets/122488130/924fa803-5da6-4d4d-a289-477ca1792fac" /> |<video src="https://github.com/linode/manager/assets/122488130/b32a8199-4666-4eb6-ab8e-1bd41fbda73f" />|
| Multi-Select  | 
| <video src="https://github.com/linode/manager/assets/122488130/0e30c4a8-e62d-4481-b8e0-c0ae01316b8f" /> |<video src="https://github.com/linode/manager/assets/122488130/f5d74950-749d-4d4d-b659-ddaab8471edc" />|

## How to test 🧪
1. Test the usages of this component in the Create Domain, Create Image and Create/Update Firewall flows -- ensure there are no regressions from prod.
2. Test all states (empty, single linode selected, multiple linodes selected) of the single- and multi-select variations of this component in both dark and light themes.


hkhalil-akamai:
Not sure whether to create a storybook entry for this component. Currently, there are no linodes/feature-specific components in storybook.

packages/manager/src/features/linodes/LinodeSelect/LinodeSelect.new.tsx
```diff
@@ -0,0 +1,196 @@
...
+  Autocomplete,
+  Box,
+  Popper,
+  PopperProps,
+  useTheme,
+} from '@mui/material';
+import React from 'react';
+import TextField from 'src/components/TextField';
+import { useInfiniteLinodesQuery } from 'src/queries/linodes/linodes';
+import { isNotNullOrUndefined } from 'src/utilities/nullOrUndefined';
```
abailly-akamai, line 21: Let's alphabetize our imports

packages/manager/src/features/linodes/LinodeSelect/LinodeSelect.new.tsx
```diff
@@ -0,0 +1,196 @@
...
+  loading?: boolean;
+  required?: boolean;
+  onBlur?: (e: React.FocusEvent) => void;
+  filter?: Filter;
+  placeholder?: string;
+  helperText?: string;
+  errorText?: string;
+  noOptionsMessage?: string;
+  optionsFilter?: (linode: Linode) => boolean;
+} & (LinodeMultiSelectProps | LinodeSingleSelectProps);
```
abailly-akamai, line 34: Let's alphabetize our props + add JSdoc comments to populate a new stories (which you need) for this new component

see https://github.com/linode/manager/pull/9060/files for implementation detail

packages/manager/src/features/linodes/LinodeSelect/LinodeSelect.new.tsx
```diff
@@ -0,0 +1,196 @@
...
+): Linode[] | Linode | null => {
+  const linodeMap = new Map(linodes.map((linode) => [linode.id, linode]));
+  if (Array.isArray(ids)) {
+    return ids.map((id) => linodeMap.get(id)).filter(isNotNullOrUndefined);
+  } else if (ids !== null) {
+    return linodeMap.get(ids) ?? null;
+  } else {
+    return null;
+  }
+};
```
abailly-akamai, line 196: Let's move this a utility file and add a test for it

packages/manager/src/foundations/themes/light.ts
```diff
@@ -382,13 +382,45 @@ export const lightTheme: ThemeOptions = {
...
+          },
+          '&.Mui-focused': {
+            backgroundColor: `${primaryColors.main} !important`,
+            color: '#fff',
+          },
+          ':hover': {
+            backgroundColor: `${primaryColors.main} !important`,
+            color: '#fff',
+          },
+        },
```
abailly-akamai, line 399: Does this change affect every option for every select in the app? could there be any regression introduced by this?

packages/manager/src/foundations/themes/light.ts
```diff
@@ -382,13 +382,45 @@ export const lightTheme: ThemeOptions = {
...
+              fontSize: '28px',
+              opacity: 0.5,
+              ':hover': {
+                opacity: 1,
+              },
+            },
+          },
+          svg: {
+            color: '#aaa',
+          },
```
abailly-akamai, line 423: same question as above

packages/manager/src/foundations/themes/light.ts
```diff
@@ -826,6 +858,9 @@ export const lightTheme: ThemeOptions = {
       styleOverrides: {
         input: {
           height: 'auto',
+          '&::placeholder': {
+            opacity: 1,
+          },
```
abailly-akamai, line 863: same question as above

packages/manager/src/features/linodes/LinodeSelect/LinodeSelect.new.tsx
```diff
@@ -0,0 +1,196 @@
...
+        ...(props.style ?? {}),
+        ...(props.style?.width
+          ? typeof props.style.width == 'string'
+            ? { width: `calc(${props.style.width} + 2px)` }
+            : { width: props.style.width + 2 }
+          : {}),
+      }}
+    />
+  );
+};
```
abailly-akamai, line 181: We should move this to its own file as a styled component

packages/manager/src/features/linodes/LinodeSelect/LinodeSelect.new.tsx
```diff
@@ -0,0 +1,196 @@
...
+  multiple: true;
+  value: number[];
+  handleChange: (selected: Linode[]) => void;
+}
+
+interface LinodeSingleSelectProps {
+  multiple?: false;
+  value: number | null;
+  handleChange: (selected: Linode | null) => void;
+}
```
abailly-akamai, line 46: same comment about props as above

packages/manager/src/features/linodes/LinodeSelect/LinodeSelect.new.tsx
```diff
@@ -0,0 +1,196 @@
+/**
+ * The plan is to eventually replace the existing `LinodeSelect` and
+ * `LinodeMultiSelect` components with this one. Once that happens,
+ * remove 'new' from this component's name.
+ */
```
abailly-akamai, line 5: Can we name it `LinodeSelectV2`?

packages/manager/src/features/linodes/LinodeSelect/LinodeSelect.new.tsx
```diff
@@ -0,0 +1,196 @@
...
+
+  const linodes = linodesData?.pages.flatMap((page) => page.data);
+  const filteredLinodes = optionsFilter
+    ? linodes?.filter(optionsFilter)
+    : linodes;
+
+  const defaultNoOptionsMessage =
+    !error && !linodesDataLoading && !filteredLinodes?.length
+      ? 'You have no Linodes to choose from'
+      : 'No options';
```
abailly-akamai, line 84: curious if 'No options' is the proper thing to display if there's an actual error. we should probably have a switch statement to display a different copy for each case

packages/manager/src/features/linodes/LinodeSelect/LinodeSelect.new.tsx
```diff
@@ -0,0 +1,196 @@
...
+      disableCloseOnSelect={multiple}
+      renderOption={(props, option, { selected }) => (
+        <li {...props}>
+          <Box
+            component={DoneIcon}
+            sx={{ width: 17, height: 17, mr: '5px', ml: '-2px' }}
+            style={{
+              visibility: selected ? 'visible' : 'hidden',
+            }}
+          />
```
abailly-akamai, line 138: Would be great to make use of styled components for each of those Box instances instead of relying both on SX and Style props

packages/manager/src/features/Firewalls/FirewallDetail/Devices/AddDeviceDrawer.tsx
```diff
@@ -47,16 +50,18 @@ export const AddDeviceDrawer = (props: Props) => {
...
 
-  const [selectedLinodes, setSelectedLinodes] = React.useState<number[]>([]);
+  const [selectedLinodeIds, setSelectedLinodeIds] = React.useState<number[]>(
+    []
+  );
 
   const handleSubmit = async () => {
     await Promise.all(
-      selectedLinodes.map((thisLinode) =>
+      selectedLinodeIds.map((thisLinode) =>
```
abailly-akamai, line 59:
```suggestion
      selectedLinodeIds.map((id: number) =>
```

abailly-akamai (commented):
Nice! I will test it some more later in the browser, but added some comments to enhance this first iteration. Importantly we need more test coverage and add a story for it

hkhalil-akamai:
Tagging @jaalah-akamai because this PR contains changes to theme files.

packages/manager/src/foundations/themes/light.ts
```diff
@@ -382,13 +382,45 @@ export const lightTheme: ThemeOptions = {
...
       styleOverrides: {
+        option: {
+          fontSize: '0.9rem',
+          padding: '10px !important',
+          '&[aria-selected="true"]': {
+            color: primaryColors.main,
+          },
+          '&.Mui-focused': {
+            backgroundColor: `${primaryColors.main} !important`,
+            color: '#fff',
```
jaalah-akamai, line 393: update to `primaryColors.white`

packages/manager/src/foundations/themes/light.ts
```diff
@@ -382,13 +382,45 @@ export const lightTheme: ThemeOptions = {
...
+          },
+          ':hover': {
+            backgroundColor: `${primaryColors.main} !important`,
+            color: '#fff',
```
jaalah-akamai, line 397: update to `primaryColors.white`

packages/manager/src/foundations/themes/light.ts
hkhalil-akamai: Looking at some of the other places we use `MuiAutocomplete`, these style changes seem to have a positive effect on the appearance of the select, making them look more consistent with our current Single Select:

| Before | After |
|-------|-------|
|Screenshot 2023-06-05 at 7 28 10 PM|Screenshot 2023-06-05 at 7 28 28 PM|
|*Notice that the option is selected but the background color change is very faint*|*The selected option has a higher contrast blue color*|

However, this component is still missing some of the other UX improvements introduced in the new `LinodeSelect`:
* The listbox is slightly narrower than the input element
* The down arrow icon is different than the one used in the single select
* No checkmark next to the selected option

Once the UX changes introduced in new `LinodeSelect` are finalized, we can abstract these fixed in a new `BaseSelect` component which will bring these improvements to all select inputs across the app.

packages/manager/src/features/Linodes/LinodeSelect/LinodeSelectV2.tsx
```diff
@@ -0,0 +1,174 @@
...
+    isLoading: linodesDataLoading,
+    error,
+    hasNextPage,
+    fetchNextPage,
+  } = useInfiniteLinodesQuery(filter);
+
+  const linodes = linodesData?.pages.flatMap((page) => page.data);
+  const filteredLinodes = optionsFilter
+    ? linodes?.filter(optionsFilter)
+    : linodes;
```
bnussman-akamai, line 89: This could work fine but I think there could be some inherent risk in combining API filtering with client side filtering.

For example, if a user with 100,000 linodes opens this Select but the first 20,000 are filtered out with `optionsFilter`, then we have the potential to make many consecutive requests with `fetchNextPage`. 

I don't know if this scenario is likely or will be a problem, but I think it is something to be aware of.

hkhalil-akamai: I looked at all usages of this prop (and the corresponding props in the previous `LinodeSelect` and `LinodeMultiSelect`), and almost all instances filter on `id` or `region` which can be turned into server-side filters. The one possible exception is `ConfigNodeIPSelect` which has a regex-based filter on `ipv4` that I believe won't work as a server-side filter.

bnussman-akamai: We had this setup to API filter on `id` but unfortunately we had to hotfix it out (https://github.com/linode/manager/pull/9151)

hkhalil-akamai: Oh right 🤦🏽‍♂️

Honestly, the changes in this PR shouldn't introduce any issues that weren't already present since it replicates the existing behavior of Linode select (but paginates requests instead of requesting all Linodes at once).

bnussman:
Your PR has a changeset ✅

packages/manager/src/features/Linodes/LinodeSelect/LinodeSelectV2.tsx
```diff
@@ -0,0 +1,176 @@
...
+  optionsFilter?: (linode: Linode) => boolean;
+  /* Displayed when the input is blank. */
+  placeholder?: string;
+  /* Displays an indication that the input is required. */
+  required?: boolean;
+}
+
+export interface LinodeMultiSelectProps extends LinodeSelectProps {
+  /* Called when the value changes */
+  handleChange: (selected: Linode[]) => void;
```
jaalah-akamai, line 42: Can we change this to `onChange` to follow our common pattern?

jaalah-akamai: Maybe `onSelectionChange` may even be a bit more descriptive

packages/manager/src/features/Linodes/LinodeSelect/LinodeSelectV2.tsx
bnussman-akamai: _Should_ be fine, but the case I explained above still could be a minor problem although I don't think it will break anything for customers. At the worst many requests will have to fire off before the customer sees some Linodes

bnussman-akamai: Ah. I also realized that the "search" ability won't really search every Linode on the account if we infinitely load the data

jaalah-akamai (approved):

packages/manager/src/features/Linodes/LinodeSelect/LinodeSelectV2.tsx
hkhalil-akamai: I don't think this is the case -- the scroll detection will automatically pull the next page until either the Listbox is filled or there are no more pages matching the search.

bnussman-akamai: Ohh good point. My bad! 🧠

bnussman-akamai (approved):

PR #9194, opened 2023-05-30
refactor: [M3-6322] – MUI v5 Migration – `Components > EnhancedSelect`

dwiley-akamai:
## Description 📝
`tss-react`'ify the `components/EnhancedSelect` directory and update code conventions and imports in a few related files.

## Notable Changes 🔄
- Convert `DropdownIndicator.tsx` and `LoadingIndicator.tsx` from class components to function components and use the `styled()` API
- `styled()` API for `Guidance.tsx`, `SelectPlaceholder.tsx`
- Increased contrast of placeholder text & made it consistent across Search bars (see `/support` path in particular)

## How to test 🧪
Audit the `<Select />` components in the app for appearance and functionality

## To-Do:
- [x] Increase contrast for search bar placeholder text
- [x] Add changeset
- [x] Investigate and rectify E2E test failure (`create-stackscripts.spec.ts`)
- [x] Investigate and rectify the numerous new console errors on the Create StackScript page

### DIFF
--- a/packages/manager/src/components/EnhancedSelect/Select.styles.ts
+++ b/packages/manager/src/components/EnhancedSelect/Select.styles.ts
@@ -3,11 +3,6 @@ import { makeStyles } from 'tss-react/mui';
 
 // TODO jss-to-tss-react codemod: usages of this hook outside of this file will not be converted.
 export const useStyles = makeStyles()((theme: Theme) => ({
-  '@keyframes dash': {
-    to: {
-      'stroke-dashoffset': 0,
-    },
-  },
   root: {
     width: '100%',
     position: 'relative',
--- a/packages/manager/src/components/EnhancedSelect/components/DropdownIndicator.tsx
+++ b/packages/manager/src/components/EnhancedSelect/components/DropdownIndicator.tsx
@@ -1,7 +1,12 @@
 import KeyboardArrowDown from '@mui/icons-material/KeyboardArrowDown';
 import { styled } from '@mui/material/styles';
+import * as React from 'react';
 
-export const DropdownIndicator = styled(KeyboardArrowDown)(() => ({
+export const DropdownIndicator = () => {
+  return <StyledKeyboardArrowDown />;
+};
+
+const StyledKeyboardArrowDown = styled(KeyboardArrowDown)(() => ({
   color: '#aaa !important',
   width: 28,
   height: 28,
--- a/packages/manager/src/components/EnhancedSelect/Select.tsx
+++ b/packages/manager/src/components/EnhancedSelect/Select.tsx
@@ -21,7 +21,7 @@ import MultiValueRemove from './components/MultiValueRemove';
 import NoOptionsMessage from './components/NoOptionsMessage';
 import Option from './components/Option';
 import Control from './components/SelectControl';
-import { SelectPlaceholder } from './components/SelectPlaceholder';
+import { SelectPlaceholder as Placeholder } from './components/SelectPlaceholder';
 
 export interface Item<T = string | number, L = string> {
   value: T;
@@ -43,7 +43,7 @@ export interface NoOptionsMessageProps {
 const _components = {
   Control,
   NoOptionsMessage,
-  SelectPlaceholder,
+  Placeholder,
   MultiValueLabel,
   MultiValueRemove,
   MenuList,


packages/manager/src/components/EnhancedSelect/components/DropdownIndicator.tsx
```diff
@@ -1,38 +1,21 @@
...
-
-const styled = withStyles(styles);
-
-export default styled(DropdownIndicator);
+export const DropdownIndicator = (props: Props) => {
+  return <StyledKeyboardArrowDown />;
+};
+
+const StyledKeyboardArrowDown = styled(KeyboardArrowDown)(() => ({
+  color: '#aaa !important',
```
jaalah-akamai, line 13: Might not need important here anymore

packages/manager/src/components/EnhancedSelect/components/DropdownIndicator.tsx
```diff
@@ -1,38 +1,21 @@
...
-    const { classes } = this.props;
-
-    return <KeyboardArrowDown className={classes.enhancedSelectDropdown} />;
-  }
-}
-
-const styled = withStyles(styles);
-
-export default styled(DropdownIndicator);
+export const DropdownIndicator = (props: Props) => {
```
jaalah-akamai, line 8: I think we can remove this and just export the styled component, right?

packages/manager/src/components/EnhancedSelect/components/LoadingIndicator.tsx
```diff
@@ -1,37 +1,12 @@
+import { styled } from '@mui/material/styles';
 import * as React from 'react';
-import { IndicatorProps } from 'react-select';
 import CircularProgress from 'src/components/core/CircularProgress';
-import { createStyles, withStyles, WithStyles } from '@mui/styles';
-import { Theme } from '@mui/material/styles';
 
-type ClassNames = 'root';
+export const LoadingIndicator = () => {
```
jaalah-akamai, line 5: Same here, can we remove and just export StyledCircularProgress? You could even rename the styled component to `LoadingIndicator`

packages/manager/src/components/EnhancedSelect/components/DropdownIndicator.tsx
dwiley-akamai: Removing `!important` results in a color regression on dropdowns like this one

Screenshot 2023-06-01 at 12 04 40 PM

packages/manager/src/components/EnhancedSelect/components/LoadingIndicator.tsx
dwiley-akamai: Because we're passing in a `test-id` and a `size` prop here the situation is a little bit different than `DropdownIndicator`

packages/manager/src/components/EnhancedSelect/Select.styles.ts
```diff
@@ -1,7 +1,8 @@
-import { makeStyles } from '@mui/styles';
 import { Theme } from '@mui/material/styles';
+import { makeStyles } from 'tss-react/mui';
 
-export const useStyles = makeStyles((theme: Theme) => ({
+// TODO jss-to-tss-react codemod: usages of this hook outside of this file will not be converted.
+export const useStyles = makeStyles()((theme: Theme) => ({
   '@keyframes dash': {
```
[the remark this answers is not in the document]
dwiley-akamai: It's not clear to me if was actually doing anything at all, maybe I'm missing the visual effect here? If it's in fact not doing anything, we could just remove it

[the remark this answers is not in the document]
jaalah-akamai: Yea you can remove it, it's not being used

packages/manager/src/components/EnhancedSelect/components/DropdownIndicator.tsx
```diff
@@ -1,38 +1,13 @@
...
-    return <KeyboardArrowDown className={classes.enhancedSelectDropdown} />;
-  }
-}
-
-const styled = withStyles(styles);
-
-export default styled(DropdownIndicator);
+import { styled } from '@mui/material/styles';
+
+export const DropdownIndicator = styled(KeyboardArrowDown)(() => ({
```
jaalah-akamai, line 4: So we used to have a wrapper here which limited the props applied to the SVG. That has been removed and the way `react-select` operates is that it spreads all the props to the components it consumes. So the SVG is getting a lot of invalid props in the DOM. You can solve this by just doing

```
export const DropdownIndicator = () => {
  return <StyledDropdownIndicator />;
};
```

cpathipa (approved):
LGTM!

jaalah-akamai (approved):

PR #9196, opened 2023-05-30
refactor: [M3-6630] - Allow Cloud UI tests to contain multiple suites, improve region handling, improve CLI UX

jdamore-linode:
## Description 📝
This is a refactor of the Cypress config, test organization, and region handling meant to facilitate adding and running additional test suites. Some improvements to region handling and UX have also been implemented.

## Major Changes 🔄
**Test Suites**
- Cloud Manager's existing tests have been moved from `cypress/e2e` to `cypress/e2e/core` and are now considered the "core" test suite
- Region and synthetic monitoring test suites have been added at `cypress/e2e/region` and `cypress/e2e/synthetic`, respectively, but do not contain any meaningful tests
- Cloud Manager's core test suite will always run by default; `yarn cy:run` will continue to run the core tests (and only the core tests) by default
- Other test suites can be run using the `CY_TEST_SUITE` environment variable, e.g. the region tests can be run with `CY_TEST_SUITE=region yarn cy:run`

**Cypress Configuration**
- Refactored our Cypress configuration at `packages/manager/cypress.config.ts`, moving plugins and other custom logic to `packages/manager/cypress/support/plugins`.
- Added new logic to improve user experience and fetch data prior to tests running
- Modified retry behavior so that failed tests are only attempted when running Cypress via CI

**Region Handling**
- Cypress tests now use the Linode APIv4 SDK's `Region` type instead of the custom `CloudRegion` interface
- Cypress fetches all available regions via the Linode API to use during tests; regions are no longer hardcoded
- When overriding region selection, Cypress can now infer which region the user desires using only the ID (previously both the ID and label needed to be specified); consequently, region override is now specified using `CY_TEST_REGION` instead of `CY_TEST_REGION_ID`, and `CY_TEST_REGION_NAME` has been removed.

**Developer/User Experience Improvements**
All of the changes above allow the following small UX improvements:
- Only `CY_TEST_REGION` must be specified when overriding a region, `CY_TEST_REGION_ID` and `CY_TEST_REGION_NAME` have been removed.
- If an invalid region is specified via `CY_TEST_REGION`, Cypress will detect it immediately and fail
- If a valid region is specified via `CY_TEST_REGION`, Cypress will note the overridden region in the CLI output
- If `MANAGER_OAUTH` is not defined, Cypress will detect it immediately and fail

## How to test 🧪
1. Confirm that Cloud Manager tests run normally when no overrides, etc. are specified (we can rely on the automated run for this).
2. Confirm that you can specify an alternative test suite to run using `CY_TEST_SUITE`:
    * Run region test suite, confirm that placeholder test passes and correctly identifies itself as a region test:
      ```bash
      CY_TEST_SUITE=region yarn cy:run
      ```
    * Run synthetic monitoring test suite, confirm that placeholder test passes and correctly identifies itself as a synthetic monitoring test:
      ```bash
      CY_TEST_SUITE=synthetic yarn cy:run
      ```
3. Confirm that you can override the region used by the tests via the `CY_TEST_REGION` environment variable:
    * Run region test suite for `us-east`, confirm that override is noted at the top of Cypress's CLI output:
      ```bash
      CY_TEST_REGION=us-east yarn cy:run
      ```
    * Run region test suite for an invalid region, confirm that Cypress notes the error and the test run fails immediately:
      ```bash
      CY_TEST_REGION=this-is-not-a-region yarn cy:run
      ```


bnussman-akamai (approved):
Lots of really nice changes! 🚀

jaalah-akamai (approved):

jdamore-linode:
I added the "Wait to Merge" label since I'll be OOO the next two weeks, and merging this may interfere with the testing I mentioned in standup if it's not coordinated with @cliu-akamai (essentially because of the change from using `CY_TEST_REGION_ID`/`CY_TEST_REGION_NAME` to just `CY_TEST_REGION`)

PR #9198, opened 2023-05-31
refactor: [M3-6491] - React Query Linodes - Configurations

bnussman-akamai:
## Description 📝

- React Query-ify Linode Configurations
- Similar to Redux, we fetch all of a Linode's configs on the Linode details page. We could paginate this, but I don't think it's really worth it...

## Major Changes 🔄
- Uses React Query hooks for all Linode Config related components
- Breaks up components more
- Moves Config specific files to config specific folder

## Preview 📷

> **Note**: There should be no changes in UI or functionality

Screenshot 2023-05-31 at 1 55 01 PM

## How to test 🧪
- Test **creating**, **updating**, and **deleting** Linode Configs


jdamore-linode (approved):
Nice! Compared the create/edit/delete behavior against Prod and didn't notice any differences. Noticed this is a draft, happy to re-review if you make any changes/additions

packages/manager/src/features/linodes/LinodesDetail/LinodeConfigs/DeleteConfigDialog.tsx
```diff
@@ -0,0 +1,57 @@
...
+    <ConfirmationDialog
+      title="Confirm Delete"
+      error={error?.[0].reason}
+      open={open}
+      onClose={onClose}
+      actions={actions}
+    >
+      <Typography>
+        Are you sure you want to delete &quot;
+        {config?.label}?&quot;
```
mjac0bs, line 53: This is also a minor typo in prod, but we could fix it here! The question mark comes before the closing quote instead of after:
Screenshot 2023-06-05 at 9 28 58 AM

Also noticed that the "Cancel" button doesn't seem to be working in prod, but is fixed on this branch. 🎉 
```suggestion
        {config?.label}&quot;?
```

mjac0bs (approved):
Functionality LGTM and nice clean up!  🧹 

 I noticed a small typo on the confirmation dialogs that I called out -- unrelated to this PR, but would be nice to fix while we're editing the files.

PR #9204, opened 2023-06-01
chore: [M3-6363, M3-6371] - Remove unused `MonkeyWrench`, `PanelErrorBoundary` components 

mjac0bs:
## Description 📝
Rather than migrate these unused components to MUIv5, we can remove them from the codebase.

I went through the rest of the `src/components` tickets, and it looks like all we have left are at least in use somewhere, even if they may not have styles to be migrated.

## Major Changes 🔄
- Removes `MonkeyWrench` from `src/components`
- Removes `PanelErrorBoundary` from `src/components`

## How to test 🧪
Verify that these components had no current use in Cloud and are fine to delete.


bnussman-akamai (approved):
Nice, I never used either of these things.

hana-linode (approved):
🧹

PR #9205, opened 2023-06-01
fix: [M3-6653] - Encode `api-v4` Path Parameters 

bnussman-akamai:
## Description 📝

- Encode all `api-v4` path parameters

## Details 👀
- User input used in URL parameters for the `api-v4` package are not escaped when creating dynamic URLs
  - If `getLinode` is called with the input `123/disks`, the SDK will actually call `linode/instances/123/disks` instead of the intended endpoint 
- No Akamai or Linode services have been affected or compromised by this vulnerability
- `api-v4` versions `<= 0.93.0` are affected

Screenshot 2023-05-31 at 5 04 55 PM

## How to test 🧪

- Check that e2es pass 
- Do a general check of Cloud Manager
- Look over the `api-v4` diff


jaalah-akamai (approved):
🚀

cpathipa (approved):

PR #9214, opened 2023-06-06
fix: [M3-6555] - Don't conditionally render VLANs depending on Account Capabilities

bnussman-akamai:
## Description 📝
- Don't conditionally render VLANs depending on Account Capabilities

## Why ❓
- Restricted users do not have permission to GET /account which does **not** allow them to see account capabilities
  - Therefore it's not valid to check for permissions for restricted users using account capabilities
- This resulted in us hiding VLANs related UI for restricted users unintentionally

## The Fix 🔧
- Simply just don't perform any checks against `account.capabilities` because it is not a valid way to check permissions when considering restricted users

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-06-06 at 1 20 16 PM | Screenshot 2023-06-06 at 1 20 23 PM |

## How to test 🧪
- Create a restricted user at `http://localhost:3000/account/users` if you don't have one
- Login to your restricted user in a private window or different browser
- Verify you can see the VLANs section on the Linode Create page

### DIFF
--- a/packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx
+++ b/packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx
@@ -33,7 +33,6 @@ import TextField from 'src/components/TextField';
 import { Toggle } from 'src/components/Toggle';
 import DeviceSelection from 'src/features/Linodes/LinodesDetail/LinodeRescue/DeviceSelection';
 import { titlecase } from 'src/features/Linodes/presentation';
-import { useAccount } from 'src/queries/account';
 import { useRegionsQuery } from 'src/queries/regions';
 import { queryKey as vlansQueryKey } from 'src/queries/vlans';
 import { useAllVolumesQuery } from 'src/queries/volumes';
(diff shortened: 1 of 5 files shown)


jaalah-akamai (approved):
🧹

packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx
```diff
@@ -245,22 +244,19 @@ export const LinodeConfigDialog = (props: Props) => {
...
-  // Doing this so that we can toggle our flag without enabling vlans for all customers.
-  const capabilities = account?.capabilities ?? [];
   const regionHasVLANS = regions.some(
     (thisRegion) =>
       thisRegion.id === linode?.region &&
       thisRegion.capabilities.includes('Vlans')
   );
-  const showVlans = capabilities.includes('Vlans') && regionHasVLANS;
+
+  const showVlans = regionHasVLANS;
```
dwiley-akamai, line 259: could get rid of this and use `regionHasVLANS` for the conditional on L859

dwiley-akamai (approved):

PR #9215, opened 2023-06-06
fix: [M3-5972] - Better/correct timestamps for finished event messages

mjac0bs:
## Description 📝
The original ask was to show an accurate completion timestamp for a Backup Restore event in the Activity Feed. Currently, these events show the created timestamp, even when the event has completed. The time between a backup restore starting and completing is most noticeable for a large linode (over 10 GB) and the inaccurate information can be misleading and confusing to users.

Because we always displayed the `created` timestamp and had no method of calculating or updating it for completed events, other events that take some time to complete (e.g. a linode migration) also had inaccurate timestamps. API does not return a completed timestamp, but it does return an event `duration` (the total duration in seconds that it takes for the Event to complete).

This PR fixes that by adding a util function to update the timestamp based on the duration of a finished event. It does the same thing for a failed event with a duration in order to more accurately reflect the content of the event message. 

## Major Changes 🔄
- In the notification menu, events landing page, and activity feed, the relative and absolute timestamps for `finished` and `failed` events now reflects the time of completion (or failure) rather than the time of event creation.

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screenshot 2023-06-07 at 1 33 50 PM | Screenshot 2023-06-07 at 1 32 14 PM |

## How to test 🧪

1. **How to reproduce the issue (if applicable)?**
   -   These are the original steps reported on the M3 ticket to see the issue:
         - Create linode, enable backups and take backup snapshot
         - Deploy from existing snapshot overwriting existing linode
         - Take note of the message shown in the event activity ("Backup restoration completed for" ...), specifically, take note of the time-stamp from Cloud Manager.
         - Verify start end job time-stamp in the Linode jobs queue from admin (I didn't actually do this - looked at the API response for events endpoint instead)
         - Note: this is more noticeable for larger backups. backups under 10gb the duration discrepancy is not so noticeable
    -   Another way to test this is by kicking off a linode migration:
        - Start a migration and note both the relative and absolute timestamps of the migration event on the events landing page.
        - Wait about until the migration completes (usually ~10 minutes for a cross-country migration of a nanode, just from anecdotal observation).
        - Use the Network tab of the browser dev tools to filter requests by "events" and find the last request that returned your event's data. Note the `created` timestamp is the same as the absolute timestamp on events landing, even though the event message has changed to reflect that the event has completed.
2. **How to verify changes?**
   - Repeat the above to check the accuracy of the timestamp for either a completed backup restore event or migration event.
   -  Use the Network tab of the browser dev tools to filter requests by "events" and find the last request that returned your event's data. Note the `created` timestamp and the `duration` in seconds. Verify this math looks correct for the completed timestamps you see for the event. 
   - Optionally, if you want to test failed events, follow the steps in #9184 to kick off a image creation event that will fail after 1-2 minutes. You should also see the updated timestamp for the time of the failure, not the creation. 

3. **How to run Unit or E2E tests?**
```
yarn test EventRow eventUtils
```

### DIFF
--- a/packages/manager/src/utilities/eventUtils.ts
+++ b/packages/manager/src/utilities/eventUtils.ts
@@ -1,4 +1,14 @@
 import { Event, EventAction } from '@linode/api-v4/lib/account';
+import { DateTime } from 'luxon';
+import { parseAPIDate } from './date';
+
+// Calculates the finished event timestamp from the created event timestamp by adding the duration;
+// if the event is not finished, uses the created event timestamp.
+export const getEventTimestamp = (event: Event): DateTime => {
+  return event.status === 'finished' && event?.duration
+    ? parseAPIDate(event.created).plus({ seconds: event.duration })
+    : parseAPIDate(event.created);
+};
 
 // Removes events we don't want to display.
 export const removeBlocklistedEvents = (
--- a/packages/manager/src/utilities/eventUtils.ts
+++ b/packages/manager/src/utilities/eventUtils.ts
@@ -2,10 +2,11 @@ import { Event, EventAction } from '@linode/api-v4/lib/account';
 import { DateTime } from 'luxon';
 import { parseAPIDate } from './date';
 
-// Calculates the finished event timestamp from the created event timestamp by adding the duration;
-// if the event is not finished, uses the created event timestamp.
+// Calculates the finished (or failed) event timestamp from the created event timestamp by adding the duration;
+// if the event is not finished or has not failed, uses the created event timestamp.
 export const getEventTimestamp = (event: Event): DateTime => {
-  return event?.status === 'finished' && event?.duration
+  return (event?.status === 'finished' || event?.status === 'failed') &&
+    event?.duration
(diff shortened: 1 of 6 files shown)


packages/manager/src/utilities/eventUtils.ts
```diff
@@ -1,4 +1,14 @@
 import { Event, EventAction } from '@linode/api-v4/lib/account';
+import { DateTime } from 'luxon';
+import { parseAPIDate } from './date';
+
+// Calculates the finished event timestamp from the created event timestamp by adding the duration;
+// if the event is not finished, uses the created event timestamp.
+export const getEventTimestamp = (event: Event): DateTime => {
+  return event.status === 'finished' && event?.duration
```
jaalah-akamai, line 8:
```suggestion
  return event?.status === 'finished' && event?.duration
```

bnussman:
Your PR has a changeset ✅

bnussman-akamai (approved):

hkhalil-akamai (approved):
Looks good! I like the `duration`-based approach

PR #9216, opened 2023-06-06
build: Ignore file casing in tsconfig

hana-linode:
## Description 📝
We recently renamed some files which caused TS to throw false positive errors. This PR adds an option to the tsconfig file to ignore file casing.

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screenshot 2023-06-06 at 4 11 31 PM | Screenshot 2023-06-06 at 4 11 43 PM |

## How to test 🧪
Run `yarn up` and you should not see any TS errors about file names

### DIFF
--- a/packages/manager/tsconfig.json
+++ b/packages/manager/tsconfig.json
@@ -34,6 +34,7 @@
     "noUnusedLocals": true,
     "strictNullChecks": true,
     "suppressImplicitAnyIndexErrors": true,
+    "forceConsistentCasingInFileNames": false,
 
     /* Completeness */
     "skipLibCheck": true,


cpathipa (approved):

bnussman-akamai (commented):
To be clear, I think these errors are happening because we use Typescript's `incremental` mode and I don't think it handled the rename well. 

I think deleting `rm packages/manager/tsconfig.tsbuildinfo` and re-running the dev server would be the fix here. 

We can still consider `forceConsistentCasingInFileNames` but I don't know what it does or if we even really need it

abailly-akamai (changes requested):
I don't think we should this. We'd be hiding errors that could be otherwise useful. As mentioned in the previous comments, this is the case of a file renaming job that left a few team members with local issues that can be addressed with a simple, one time job.

PR #9222, opened 2023-06-07
feat: [M3-6640] - Metadata beta updates

hana-linode:
## Description 📝
Metadata will be going into beta soon (shooting for 6/15 - 7/5). Rachna requested calling out that the feature is in beta in the Add User Data Accordion should be sufficient. We also don't want customers not tagged for the Metadata beta to see Metadata features

This PR:
- Adds a beta chip to the Add User Data Accordion
- Adds an additional check for the cloud-init compatible checkbox and cloud-init compatible icon in the images dropdown
  - This will basically serve as the customer tag check since the API will not return cloud-init in the image's capability if the customer doesn't have the customer tag once the beta starts
  - The Add User Data Accordion basically already uses a similar check so no update there is needed

## Preview 📷
image

## How to test 🧪
Right now, Metadata is enabled on dev even without the customer tag for ease of testing.

1. We can simulate what removing the customer tag would look like by pointing to prod instead of dev and ensuring that the Image capture/upload flows don't have the checkbox and that the Image drop downs don't have the cloud-init compatible icon.
2. Point to Dev or turn on the MSW
3. Go to Linode Create and select a cloud-init compatible image (e.g. `Ubuntu 22.10 - Cloud Init Special` or `metadata-test-distro`)
4. Select a region compatible with Metadata (that would currently be `London`)
5. Scroll down and you should see the Add User Data Accordion with the beta chip


mjac0bs (approved):
LGTM - observed the beta chip, image cloud-init checkboxes, and cloud-init incompatible image select icon when pointing to dev and confirmed none of that was visible when pointing to prod. Beta chip in the accordion looks good in light and dark mode at various screen sizes.

abailly-akamai (approved):
Looks good! Approving pending minor refactor

bnussman:
Your PR has a changeset ✅

PR #9248, opened 2023-06-12
change: [M3-6447] - Update Button storybook pages, rename StyledLinkButton

jaalah-akamai:
## Description 📝
Updated the Button storybook page to CSF 7.0 format

## Major Changes 🔄
- Fixed storybook issue where Button height was 52px (too large)
- Update storybook pages to CSF 7.0 format with 3 primary button types
- Renamed `tooltipGAEvent` to `tooltipAnalyticsEvent`
- Renamed `LinkStyledButton` to `StyledLinkButton`

> **Note**: I kept the content of the page in the MDX file, but we don't need all variations on the page. Those can just be toggled via controls

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screen Shot 2023-06-12 at 11 47 07 AM | Screen Shot 2023-06-12 at 11 42 01 AM |

## How to test 🧪
1. Run `yarn storybook`
2. Go to: http://localhost:6006/?path=/docs/components-button--documentation


jaalah-akamai:
I don't think we need a changeset for this

abailly-akamai (changes requested):
Let's use autodocs! We only really need mdx files for pure documentation if ever needed. See https://github.com/linode/manager/pull/9060/files for implementation details. It will ensure we get an interactive, auto generated props table, which is missing from this PR, and removes the need for a separate .mdx file, which is not portable.

- Add JSDocs to your component props (will generate a prop table). (You can still pass args in your `meta` as you did which will make the props table a playground
- Add those mdx descriptions above the components in your stories file as a JSDoc comment as well ([example](https://github.com/linode/manager/pull/9060/files#diff-8a1c68a0a6681de4da558634b4caedc4790b441cb9aedb01d22db4c2148bc61bR6)) and the component overview can be added right above the main export like [here](https://github.com/abailly-akamai/manager/blob/60a239ee25daab866e4625240592646adcef08f7/packages/manager/src/components/BarPercent/BarPercent.tsx#L24-L26). This will ensure our components are portable and self descriptive in case we move away from storybook for example.

jaalah-akamai:
@abailly-akamai Thanks for reminder, the page is looking much better!

Screen Shot 2023-06-13 at 11 24 13 AM

abailly-akamai (approved):
Nice thx for the changes!

abailly-akamai:
@jaalah forgot to mention, may be worth adding a story for the TextButton?

jaalah-akamai:
> @jaalah forgot to mention, may be worth adding a story for the TextButton?

Good idea! Yes 👍

PR #9249, opened 2023-06-12
fix: [M3-6686] - Linode Details crashing for Linodes that don't have plans

bnussman-akamai:
## Description 📝

- Fix Linode details crashing for Linodes with no plan
- For Linodes with no plan, we were requesting `/v4/linode/types/${id}` but id was empty string resulting in the query actually GETing all the types from the API

## How to test 🧪
- Test against an account with Linodes that don't have plans

### DIFF
--- a/packages/manager/src/features/Linodes/LinodesDetail/LinodeRescue/RescueDialog.tsx
+++ b/packages/manager/src/features/Linodes/LinodesDetail/LinodeRescue/RescueDialog.tsx
@@ -17,7 +17,10 @@ export const RescueDialog = (props: Props) => {
     linodeId ?? -1,
     linodeId !== undefined && open
   );
-  const { data: type } = useTypeQuery(linode?.type ?? '', linode !== undefined);
+  const { data: type } = useTypeQuery(
+    linode?.type ?? '',
+    Boolean(linode?.type)
+  );
 
   const isBareMetalInstance = type?.class === 'metal';
 
--- a/packages/manager/src/queries/types.ts
+++ b/packages/manager/src/queries/types.ts
@@ -54,6 +54,6 @@ export const useTypeQuery = (type: string, enabled = true) => {
     queryKey: specificTypesQueryKey(type),
     queryFn: () => getType(type),
     ...queryPresets.oneTimeFetch,
-    enabled: enabled && type !== '',
+    enabled: enabled && Boolean(type),
   });
 };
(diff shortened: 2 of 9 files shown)


packages/manager/src/features/Linodes/LinodesDetail/LinodeRescue/RescueDialog.tsx
```diff
@@ -17,7 +17,10 @@ export const RescueDialog = (props: Props) => {
     linodeId ?? -1,
     linodeId !== undefined && open
   );
-  const { data: type } = useTypeQuery(linode?.type ?? '', linode !== undefined);
```
bnussman-akamai, line 20: Basically, this was a sufficient check to enable the query. 

`linode !== undefined` only checks if the Linode object is defined... What we really want to check is if `linode.type` is defined (not null)

packages/manager/src/queries/types.ts
```diff
@@ -54,6 +54,6 @@ export const useTypeQuery = (type: string, enabled = true) => {
     queryKey: specificTypesQueryKey(type),
     queryFn: () => getType(type),
     ...queryPresets.oneTimeFetch,
-    enabled,
+    enabled: enabled && type !== '',
```
dwiley-akamai, line 57: Do we need to check for `type` being `null` as well/instead? That's what the linodes with no plans are coming back as having

dwiley-akamai (approved):
Confirmed the bug has been fixed 🚀

packages/manager/src/queries/types.ts
bnussman-akamai: Yeah, we could but Typescript won't allow a null or undefined value to be passed as `type` here. 

Should I still make the check?

```suggestion
    enabled: enabled && Boolean(type),
```

cpathipa (approved):

packages/manager/src/queries/types.ts
dwiley-akamai: Gotcha, I think it'd be good just for consistency/clarity

abailly-akamai (approved):
thx! This fixes it also shows that we may want to be more proactive about try/catch certain code blocks (at least closer to the HOC) and display an error to the user rather than crashing their browser. We also _may_ be able to spot things quicker if we log errors in the catch block. Not sure how to improve this best yet, but will put out an item to discuss this further

bnussman-akamai:
Yeah this should have been handled better.

Because of how the API is routed `/v4/linode/types/${id}` with `id` being an empty string, it is till a valid API call, although it returns unexpected data. 

I think the most optimal fix here would be to have the `api-v4` package validate that `typeId` is some truthy value for the `getType` function

PR #9256, opened 2023-06-13
refactor: [M3-6694] - Convert <App /> to functional component

hkhalil-akamai:
## Description 📝
Converts the `App` class component to a functional component, using hooks instead of containers where possible. This is useful for the RQ Events work since that query is awkward to deploy as a container.

## Major Changes 🔄
- Make `App` a functional component

## How to test 🧪
1. Ensure the Cloud Manager loads and works normally.
2. Ensure `<TheApplicationIsOneFire />` component is displayed when the app crashes (you can add a random `throw` statement to trigger a crash).
3. Ensure NewRelic and analytics scripts load and work as intended.
4. Ensure keyboard shortcuts to change theme/open go-to menu work.
5. Ensure event handlers work (e.g., the linodes cache is invalidated when an update event comes in)
6. Unit tests: `yarn test`
7. E2E tests: `yarn cy:e2e`


packages/manager/src/App.tsx
```diff
@@ -126,234 +131,178 @@ export class App extends React.Component<CombinedProps, State> {
...
+    return () => {
+      subscriptions.forEach((sub) => sub.unsubscribe());
+    };
+  }, [enqueueSnackbar]);
+
+  /**
+   * basically, if we get an "invalid oauth token"
+   * error from the API, just render nothing because the user is
+   * about to get shot off to login
+   */
```
jaalah-akamai, line 273:
```suggestion
  /**
   * in the event that we encounter an "invalid OAuth token" error from the API, 
   * we can simply refrain from rendering any content since the user will 
   * imminently be redirected to the login page.
   */
```

packages/manager/src/App.tsx
```diff
@@ -126,234 +131,178 @@ export class App extends React.Component<CombinedProps, State> {
...
-      return null;
-    }
+    document.addEventListener('keydown', keyboardListener);
+    return () => {
+      document.removeEventListener('keydown', keyboardListener);
+    };
+  }, [keyboardListener]);
+
+  React.useEffect(() => {
+    const subscriptions = [
```
jaalah-akamai, line 149: What do you think about slightly re-working this structure to something like:
```
React.useEffect(() => {
  const eventHandlers = [
    { 
      filter: ({ event }) => !event._initial && ['linode_migrate'].includes(event.action), 
      handler: handleMigrationEvent 
    }
    // Other event handlers...
  ];

  const subscriptions = eventHandlers.map(({ filter, handler }) =>
    events$
      .filter(filter)
      .subscribe(handler)
  );

  return () => {
    subscriptions.forEach(
      // Same code as before to unsubscribe from events
    )
  };
}, []);


const handleMigrationEvent = ({ event }) => {
  // Everything from L240-260
  // https://github.com/linode/manager/blob/bda9c8f536e46c323a9bc73e9e29887173f4c94a/packages/manager/src/App.tsx#L240-L260
};
```

packages/manager/package.json
bnussman-akamai, line 1: What are your thoughts on just using a class component error boundary. We did this somewhat recently with `PayPalErrorBoundary`. Usually I wouldn't be such a stickler about dependencies but yarn is so slow

bnussman-akamai (approved):
Nice work 🧼

bnussman-akamai:
Looks like the unit test needs to be updated. Might need to move away from enzyme

mjac0bs (approved):
Looks good with analytics. The only thing I didn't see working is the keyboard shortcut for toggling the app theme, but that's been weird on prod too so I wouldn't worry about it here. The 'go to' shortcut works.

There's a unit test that's failing, but after that's fixed then 🚢 .

PR #9258, opened 2023-06-14
refactor: [M3-6362,M3-6365] - Update `Components > ModeSelect, NavTabs`

mjac0bs:
## Description 📝
These were MUI Migration tickets, but there were no styles to migrate.

I looked into removing Reach UI from NavTabs in favor of the MUI Tabs but decided that effort should be its own tech story due to the amount of tab-related components we have and how widely we use tabs across the app. A ticket for this is now in our backlog: M3-6705.

## Major Changes 🔄
- Use named rather than default exports
- Update props names
- Remove React.FC

## How to test 🧪
Check that there have been no regressions in NavTabs or ModeSelect components in the app. 
NavTabs can be found on https://localhost:3000/profile/display and ModeSelect can be found on the Linode Disk Drawer from Linode Details > Storage > Add Disk.

### DIFF
--- a/packages/manager/src/components/ModeSelect/ModeSelect.tsx
+++ b/packages/manager/src/components/ModeSelect/ModeSelect.tsx
@@ -8,19 +8,14 @@ export interface Mode<modes> {
   mode: modes;
 }
 
-interface Props {
+interface ModeSelectProps {
   selected: string;
   modes: Mode<any>[];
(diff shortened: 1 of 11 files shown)


bnussman-akamai (approved):
Nice. Looking forward to replacing these with MUI's Tabs

abailly-akamai (approved):
Looks good!

PR #9271, opened 2023-06-16
refactor: Round dates on the filters for invoices and payments

jcallahan-akamai:
## Description 📝
On the Billing Activity table, requests to `/payments` and `/invoices` include a `date` filter depending on which time range the user has selected. 

E.g. if the user selected "30 days", Cloud requests invoices with a filter like this: `{ date: { "+gte": <currentDate - 30 days> } }`.

This negatively impacts RQ caching, because a new cache key is generated for every request (since the date is specified down to the _second_). 

**This PR rounds these dates to the start of the given day**. This means that RQ will cache results for an entire day.

**Additionally,** I removed the `date` filter when users select "All Time". Previously the unix epoch was used to keep the code simpler, but removing this altogether might be better from an API perspective.

## Preview 📷
Screenshot 2023-06-16 at 10 54 42 AM

## How to test 🧪
- Run the app and test the /account page. 
- Open RQ dev tools and observe the caching when you select different date ranges
- This is a refactor and performance improvement: there should be no behavioral differences between what's here and what's in production.

### DIFF
--- a/packages/manager/src/features/Billing/BillingPanels/BillingActivityPanel/BillingActivityPanel.tsx
+++ b/packages/manager/src/features/Billing/BillingPanels/BillingActivityPanel/BillingActivityPanel.tsx
@@ -39,7 +39,7 @@ import {
   useAllAccountInvoices,
   useAllAccountPayments,
 } from 'src/queries/accountBilling';
-import { isAfter, parseAPIDate } from 'src/utilities/date';
+import { parseAPIDate } from 'src/utilities/date';
 import formatDate from 'src/utilities/formatDate';
 import { getAll } from 'src/utilities/getAll';
 import { getTaxID } from '../../billingUtils';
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/Billing/BillingPanels/BillingActivityPanel/BillingActivityPanel.tsx
```diff
@@ -203,23 +203,22 @@ export const BillingActivityPanel = (props: Props) => {
     setSelectedTransactionDate,
   ] = React.useState<DateRange>(defaultDateRange);
 
+  const endDate =
+    selectedTransactionDate !== 'All Time'
+      ? getCutoffFromDateRange(selectedTransactionDate)
```
bnussman-akamai, line 208: Optional change but rather than doing logic here we could make `getCutoffFromDateRange` return `undefined` for `All Time`

bnussman-akamai (commented):
New cutoff date is working like i'd expect!

bnussman-akamai (approved):

packages/manager/src/features/Billing/BillingPanels/BillingActivityPanel/BillingActivityPanel.tsx
jcallahan-akamai: Yeah that makes sense, good idea. Just addressed this.

packages/manager/src/features/Billing/BillingPanels/BillingActivityPanel/BillingActivityPanel.tsx
```diff
@@ -399,7 +388,11 @@ export const BillingActivityPanel = (props: Props) => {
             </div>
           </div>
         </div>
-        <OrderBy data={filteredData} orderBy={'date'} order={'desc'}>
+        <OrderBy
+          data={selectedTransactionType === 'all' ? combinedData : filteredData}
+          orderBy={'date'}
+          order={'desc'}
+        >
```
bnussman-akamai, line 395: This could be handled in a follow up ticket but just realized this `<OrderBy />` isn't doing anything for us right now. We could remove it or maybe we could do some client-side ordering here?

bnussman-akamai (approved):
🧼

packages/manager/src/features/Billing/BillingPanels/BillingActivityPanel/BillingActivityPanel.tsx
jcallahan-akamai: Yeah, it might be nice to be able to order by Amount. I just created **M3-6715** for this.

jaalah-akamai (approved):
Nice to see you contributing again! 🥲

PR #9280, opened 2023-06-16
fix: [M3-6664] - Fix crash due to unescaped RegEx

hkhalil-akamai:
## Description 📝
Fixes a crash that occurs when viewing events that contain RegEx special characters.

## Todo:
- [x] Figure out how to reproduce
- [x] Add unit test

## How to test 🧪
1. Trigger an entity event containing RegEx special characters (a known bad string is `(?)`).
   - This PR adds an event to the MSW that contains special characters and should crash the app without this fix
2. Open the Notifications menu or navigate to `/events`.
3. Verify the app doesn't crash.

### DIFF
--- a/packages/manager/.changeset/pr-9280-fixed-1686949826559.md
+++ b/packages/manager/.changeset/pr-9280-fixed-1686949826559.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Fixed
+---
+
+Fixed crash caused by un-escaped RegEx ([#9280](https://github.com/linode/manager/pull/9280))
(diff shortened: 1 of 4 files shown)


hana-linode (commented):
I can't seem to find the event with special characters 🤔 

image

dwiley-akamai (commented):
I'm seeing the same as Hana -- I checked out `develop` and added the mocked data from this PR so I could observe the crash, but I didn't see the event anywhere and the app worked fine.

hkhalil-akamai:
@dwiley-akamai @hana-linode thanks for the catch! Updated the mocked event to make sure it appears in the Notification menu as well as `/events`.

hana-linode (approved):
Confirmed that the app doesn't crash with the fix!

mjac0bs (approved):

PR #9281, opened 2023-06-16
fix: [M3-6687] - Show accurate Notification Threshold values when toggling in Linode Settings

mjac0bs:
## Description 📝
Fixes unexpected behavior in the Notification Thresholds accordion within a Linode's Settings where default values were being displayed after toggling off -> on a setting, rather than the values a user saved.

Now, when a user makes and saves a change to a value and then toggles a setting, that saved threshold value will accurately display in the text field, rather than the default value.

## Preview 📷

| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/linode/manager/assets/114685994/fb0e0be0-8450-4ab0-97f3-4c52e04c37f7" /> | <video src="https://github.com/linode/manager/assets/114685994/590ee075-e469-4450-8703-20ab7fdbd92e" /> |

## How to test 🧪
**How to reproduce the issue (if applicable)?**
1. Have a Linode on hand or spin one up and and navigate to the Settings tab of the Linode Details page.
2. Under the Notification Threshold accordion, change the values of any one of the thresholds and click Save. 
3. Toggle off that threshold alert. Then toggle it back on.
4. Observe that the textfield shows the original default value, not the custom value you just saved for that threshold. 
5. Edit the threshold value to be the value you previously updated, before the default was restored.
6. Observe that the Save button is disabled, because as far as the form is concerned, now no changes to the values have been made.
7. Refresh the page and observe that the changed threshold value displays for that threshold setting.

**How to verify changes?**
1. Repeat the steps above, but observe that the textfield shows custom value you just saved for that threshold, not the default value, and the Save button is disabled, as expected.
2. Test different toggling and editing scenarios to make sure there are no regressions.

### DIFF
--- a/packages/manager/src/features/Linodes/LinodesDetail/LinodeSettings/LinodeSettingsAlertsPanel.tsx
+++ b/packages/manager/src/features/Linodes/LinodesDetail/LinodeSettings/LinodeSettingsAlertsPanel.tsx
@@ -92,7 +92,11 @@ export const LinodeSettingsAlertsPanel = (props: Props) => {
       ) =>
         formik.setFieldValue(
           'cpu',
-          checked ? 90 * (linode?.specs.vcpus ?? 1) : 0
+          checked
+            ? linode?.alerts.cpu
+              ? linode?.alerts.cpu
+              : 90 * (linode?.specs.vcpus ?? 1)
+            : 0
         ),
       onValueChange: (e: React.ChangeEvent<HTMLInputElement>) =>
         formik.setFieldValue('cpu', e.target.valueAsNumber),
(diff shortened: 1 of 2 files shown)


packages/manager/src/features/Linodes/LinodesDetail/LinodeSettings/LinodeSettingsAlertsPanel.tsx
```diff
@@ -112,7 +116,11 @@ export const LinodeSettingsAlertsPanel = (props: Props) => {
       onStateChange: (
         e: React.ChangeEvent<HTMLInputElement>,
         checked: boolean
-      ) => formik.setFieldValue('io', checked ? 10000 : 0),
+      ) =>
+        formik.setFieldValue(
+          'io',
+          checked ? (linode?.alerts.io ? linode?.alerts.io : 10000) : 0
```
bnussman-akamai, line 122: Maybe should just remove the default values? The API will always return something here. We'd just have to check that the linode is not `undefined` first.

```suggestion
          checked ? linode.alerts.io : 0
```

mjac0bs: This was a good thought. When I tried this for a threshold, such as `checked ? linode?.alerts.io : 0`, we end up in a situation where an alert can be toggled off, but then can't be toggled back on -- since the API returns `0` when an alert is disabled, and the toggle state is enabled when the value > `0`, the toggle seems perpetually disabled. 

https://github.com/linode/manager/assets/114685994/1f86f201-a871-47cc-a20e-509cfc94e375

bnussman-akamai (approved):
Thanks for the quick turnaround on this! 🚀

bnussman-akamai:
I have no idea what messed up banks bot on this PR

PR #9286, opened 2023-06-20
feat: [M3-6707] – Light/dark mode shortcut copy on the "My Settings" page

dwiley-akamai:
## Description 📝
For some historical context on the light/dark mode keyboard shortcut feature, refer to #7595 and #7606.

This PR exposes the keyboard shortcut in copy as opposed to having it be just an easter egg.

## Major Changes 🔄
- Instructional text/copy added to the "Theme" section re: keyboard shortcut

## How to test 🧪
Confirm that `Ctrl` + `Shift` + `D` works to change the theme in the app in Chrome, Firefox, and Safari.

Confirm that the copy in the Theme section of the "My Settings" page is accurate.

### DIFF
--- a/packages/manager/src/App.tsx
+++ b/packages/manager/src/App.tsx
@@ -70,7 +70,7 @@ const BaseApp = withFeatureFlagProvider(
     const theme = preferences?.theme;
     const keyboardListener = React.useCallback(
       (event: KeyboardEvent) => {
-        const letterForThemeShortcut = 'L';
+        const letterForThemeShortcut = 'D';
         const letterForGoToOpen = 'K';
         const modifierKey = isOSMac ? 'ctrlKey' : 'altKey';
         if (event[modifierKey] && event.shiftKey) {
@@ -83,9 +83,6 @@ const BaseApp = withFeatureFlagProvider(
               break;
             case letterForGoToOpen:
               setGoToOpen(!goToOpen);
-              break;
-            default:
-              console.log(event.key);
               break;
           }
         }
--- a/packages/manager/src/App.tsx
+++ b/packages/manager/src/App.tsx
@@ -71,7 +71,7 @@ const BaseApp = withFeatureFlagProvider(
     const keyboardListener = React.useCallback(
       (event: KeyboardEvent) => {
         const isOSMac = navigator.userAgent.includes('Mac');
-        const letterForThemeShortcut = 'D';
+        const letterForThemeShortcut = 'L';
         const letterForGoToOpen = 'K';
         const modifierKey = isOSMac ? 'ctrlKey' : 'altKey';
         if (event[modifierKey] && event.shiftKey) {
@@ -84,6 +84,9 @@ const BaseApp = withFeatureFlagProvider(
               break;
             case letterForGoToOpen:
               setGoToOpen(!goToOpen);
+              break;
+            default:
+              console.log(event.key);
               break;
           }
         }
--- a/packages/manager/src/App.tsx
+++ b/packages/manager/src/App.tsx
@@ -70,7 +70,6 @@ const BaseApp = withFeatureFlagProvider(
     const theme = preferences?.theme;
     const keyboardListener = React.useCallback(
       (event: KeyboardEvent) => {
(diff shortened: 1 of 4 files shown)


abailly-akamai:
@dwiley-akamai thanks! was wondering, is there a way for users to know about it? if not we should add a sentence below the toggle at /profile/settings

hkhalil-akamai:
There's a [comment in the code](https://github.com/linode/manager/pull/9286/files#diff-106b5b793d9c13dae5657b9f0bbb47f6e358158c5c6faf7adfe8aa26467dc55eR146-R149) that suggests this feature is supposed to be an easter-egg, so maybe users are expected to read the source code to find out about it?

abailly-akamai:
@dwiley-akamai @hkhalil-akamai sounds to me like a cool feature, not sure why we would hide it from users?

packages/manager/src/App.tsx
```diff
@@ -85,6 +84,9 @@ const BaseApp = withFeatureFlagProvider(
             case letterForGoToOpen:
               setGoToOpen(!goToOpen);
               break;
+            default:
+              console.log(event.key);
+              break;
```
dwiley-akamai, line 89: This is temporary to facilitate testing. It will be removed prior to merging.

packages/manager/src/features/Profile/Settings/Settings.tsx
```diff
@@ -161,3 +167,9 @@ export const ProfileSettings = () => {
     </>
   );
 };
+
+const ThemeKeyboardShortcut = isOSMac ? (
+  <Code>Ctrl + Shift + L</Code>
+) : (
+  <Code>Alt + Shift + L</Code>
+);
```
bnussman-akamai, line 175: I think I prefer this to just wrapping it all in one `<Code />`

```suggestion
const ThemeKeyboardShortcut = isOSMac ? (
  <span>
    <Code>Ctrl</Code> + <Code>Shift</Code> + <Code>L</Code>
  </span>
) : (
  <span>
    <Code>Alt</Code> + <Code>Shift</Code> + <Code>L</Code>
  </span>
);
```

bnussman-akamai (commented):
PR looks good but I've never had problems with `Ctrl` + `Shift` + `D` so it's hard for me support the breaking change. Curious to hear what others think

mjac0bs (approved):
This looked good on Chrome, Firefox, and Safari on Mac, as well as Chrome, Firefox, and Edge on Windows running in UTM. Observed the keyboard shortcut keys copy change for Windows, as well.

I am in support of making the breaking change -- I've only gotten the existing shortcut to work maybe one time ever? And this is a significant improvement. Plus, we never previously exposed the shortcut and now are exposing it with the breaking change, so if any users were using it successfully before, the new shortcut is visible. 

I was originally wondering if it would make more sense to make the shortcut `Ctrl + Shift + T` with `T` for "Theme"... but `L` is a home row key and the opposite of our existing shortcut, so I think keeping the `L` is fine.

bnussman-akamai:
@mjac0bs @dwiley-akamai Do you guys leave Webex open? If so, that's probably why `Ctrl` + `Shift` + `D` doesn't work for you

jaalah-akamai:
> @mjac0bs @dwiley-akamai Do you guys leave Webex open? If so, that's probably why `Ctrl` + `Shift` + `D` doesn't work for you

This is 100% the issue - works flawlessly once closed

mjac0bs:
Screenshot 2023-06-28 at 9 36 42 AM

So, the existing shortcut does work for me on both Windows and Firefox with Webex's "Share Content" shortcut unchecked. What if we use this PR to just expose the shortcut copy (especially since it's `Alt` and not `Ctrl` on Windows) and keep the keybinding itself as is?

dwiley-akamai:
Good find @bnussman-akamai, and I agree @mjac0bs with keeping the existing shortcut. I'll undo the changes aside from the copy.

bnussman-akamai (approved):
🧼

jaalah-akamai (approved):

PR #9288, opened 2023-06-21
style: [M3-6639] - Add outline to country flags that contain white

bnussman-akamai:
## Description 📝
- Adds an outline to some country flags so the contrast in light mode is better

## Major Changes 🔄
- Added outline to `jp`, `sg`, and `id` flags
- Added Storybook story for the `<Flag />` component 

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screenshot 2023-06-21 at 11 05 23 AM | Screenshot 2023-06-21 at 11 04 49 AM |

## New Storybook Story 🤤
Screenshot 2023-06-21 at 11 09 33 AM

## How to test 🧪
- Pull the PR
- Switch to **Light Mode**
- Go to the Linode Create page
- Open the Region Select component
- Verify that there is a subtitle outline around the `jp`, `sg`, and `id` flags
- Optionally, open storybook and verify you see the outline there too


packages/manager/src/components/Flag.tsx
```diff
@@ -1,21 +1,45 @@
...
 
+const COUNTRIES_TO_OUTLINE = ['jp', 'id', 'sg'];
+
 interface Props {
-  /** expects a iso code of a country - `us`, `ca`, etc... */
-  country: string;
+  country: Lowercase<Country>;
 }
 
+const useStyles = makeStyles()((theme) => ({
```
bnussman-akamai, line 16: I'm expecting feedback saying to used `styled`, but since I already have to use classNames for the flag classes, I figured `makeStyles` was fitting

packages/manager/src/components/Flag.tsx
```diff
@@ -1,21 +1,45 @@
...
+  root: {
+    fontSize: '1.5rem',
+    verticalAlign: 'top',
+    width: '1.41rem',
+  },
+  outline: {
+    outline: `1px solid ${theme.color.border3}`,
```
bnussman-akamai, line 23: I didn't pick this color, I just took what we us to have https://github.com/linode/manager/pull/8996/files#diff-fc0fd72cba5b8eafcc03067dc47013fe8e52342886c21da479bdfb817901b187L63

packages/manager/src/components/EnhancedSelect/variants/RegionSelect/RegionSelect.tsx
```diff
@@ -49,7 +49,7 @@ export const getRegionOptions = (regions: Region[]) => {
     groups[group].push({
       label: `${region.label} (${region.id})`,
       value: region.id,
-      flag: <Flag country={region.country} />,
+      flag: <Flag country={region.country as Lowercase<Country>} />,
```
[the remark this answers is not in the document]
jaalah-akamai: Should we just have country defined as `Country` in the `Regions` type? https://github.com/linode/manager/blob/develop/packages/api-v4/src/regions/types.ts#L24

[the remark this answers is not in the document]
bnussman-akamai: This crossed my mind. We could but we'd have to copy over all of the country data to the api-v4 package, which I'm not sure we should do.

packages/manager/src/components/Flag.tsx
jaalah-akamai: Haha 🤣  - Well... tss `makeStyles` was intended as stop-gap. In addition, I can get this down to less lines and make this "arguably" easier to follow:
```
export const Flag = (props: Props) => {
  const country = props.country.toLowerCase();

  return (
    <StyledFlag
      className={`fi fi-${COUNTRY_FLAG_OVERRIDES[country] ?? country} fi-xx`}
      hasOutline={COUNTRIES_TO_OUTLINE.includes(country)}
    />
  );
};

const StyledFlag = styled('div', {
  label: 'StyledFlag',
  shouldForwardProp: (prop) => isPropValid(['hasOutline'], prop),
})<{ hasOutline: boolean }>(({ theme, ...props }) => ({
  fontSize: '1.5rem',
  outline: props.hasOutline ? `1px solid ${theme.color.border3}` : 'none',
  verticalAlign: 'top',
  width: '1.41rem',
}));
```

bnussman-akamai: Okay yeah, that's pretty clean 😅 I'll make this change but gonna do a mini-rant below.

As much as I want to get on the styled train, I can't accept the fact that to make a completely separate component to style the intended component. It feels like an anti-pattern to me or something.

In my head I see this: 
`makeStyles` : Styles ➕ Component 🟰 Useable Component
`styled`: (Component A ➕ Styles) ➡️ Component B 🟰 Useable Component

I can't un-see this and it makes me not like styled. I feel like I might just need a different perspective on this to fully understand why `styled` is more generally preferred. Something about my mental model does not match that of others on this one 😖

abailly-akamai: We should have a discussion in the cafe about this, this is a healthy discussion to have because these patterns can feel ambiguous to some while logical to to others. It'Il probably will take the whole hour 😅. I frankly sit in between. My main concern is tech debt. I feel like styled components should maybe be limited to core components, but our problem is largely due to the fact we don't have a proper design system in place (we will in the near future!), which forces us to override far too much, and I honestly feel like no styling API will satisfy, as the problem is else.

In an ideal world we have styled components for core components (portable, re-usable, scoped) and sx/cx for occasional overrides. Their main benefit in ensuring style encapsulation, but the same can be said about makeStyles() if used properly. makestyles() aesthetically functions more like a SCSS traditional stylesheet and all the good and bad it allows. When i come to think of it, styled components should never have nested classes because that's when you start breaking specificity but they way we use them now does ...

Overusing styled component is a very valid concern, and I selfishly want to direct the conversation towards their longevity and migration ability. It's like asking what will take us the longest to refactor/migrate once we need to. To my understanding they'll be friendlier to port to a platform like Next.js for instance, but we'd probably need to build a new component library before doing so anyway (and server components will be part of the ecosystem too!). Super curious about @jaalah-akamai thoughts as well on this. Let's take this to the forum!

abailly-akamai: also.. is there a way that `shouldForwardProp: (prop) => isPropValid(['hasOutline'], prop)` could be a default of some sort?

bnussman-akamai: Yaya, agree. And all of what you said echoes my thoughts @abailly-akamai

To be clear, I have nothing against styled components or our choice to use them. I just was explaining why the mental model for me isn't quite clicking for me

jaalah-akamai: Ok we will dedicate some time to this subject in our cafe

packages/manager/src/components/EnhancedSelect/variants/RegionSelect/RegionSelect.tsx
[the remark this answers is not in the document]
abailly-akamai: I also feel like that's where it belongs (rather than defined in the manager package) but this could be a follow up discussion/task

abailly-akamai (approved):

packages/manager/src/components/EnhancedSelect/variants/RegionSelect/RegionSelect.tsx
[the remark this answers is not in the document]
bnussman-akamai: I agree. What I don't know is if would it be appropriate to list all 243 countries as a possible response from the API. I guess we'd have to

bnussman-akamai:
I agree with removing the border in dark mode. I can reach out to UX about border on flags in in light mode @hkhalil-akamai

hkhalil-akamai (approved):
Shadows look good 👍🏽

PR #9291, opened 2023-06-21
refactor: [M3-6359] - MUI v5 Migration - `Components > LongviewLineGraph`

hana-linode:
## Description 📝
Update styles and code patterns for the LongviewLineGraph component

## How to test 🧪
So, we don't have an account that has Longview setup and when I tried to run the curl command to set it up, I kept running into errors. I think the changes are minor enough that we can just check the code

### DIFF
--- a/packages/manager/src/components/LongviewLineGraph/LongviewLineGraph.tsx
+++ b/packages/manager/src/components/LongviewLineGraph/LongviewLineGraph.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
-import { compose } from 'recompose';
 import Divider from 'src/components/core/Divider';
-import { makeStyles } from '@mui/styles';
+import { makeStyles } from 'tss-react/mui';
 import { Theme } from '@mui/material/styles';
 import Typography from 'src/components/core/Typography';
 import { ErrorState } from 'src/components/ErrorState/ErrorState';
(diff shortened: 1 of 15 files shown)


packages/manager/src/components/LongviewLineGraph/LongviewLineGraph.tsx
```diff
@@ -11,7 +10,7 @@ import {
   LineGraphProps,
 } from 'src/components/LineGraph/LineGraph';
 
-const useStyles = makeStyles((theme: Theme) => ({
+const useStyles = makeStyles()((theme: Theme) => ({
```
hana-linode, line 13: I opted to just run the tss code mod instead of converting to a Styled component since there tends to be issues with specificity with Styled components and we don't really have a way to test longview

hana-linode, line 13: I opted to just run the tss code mod instead of converting to a Styled component since there tends to be issues with specificity with Styled components and we don't really have a way to test longview. Plus, we no longer support longview and we're probably going to remove it anyway in the future

bnussman-akamai (approved):

mjac0bs (approved):

cpathipa (approved):

PR #9296, opened 2023-06-22
fix: [M3-6638] - Object Storage > Create Access Key Drawer Layout Issue

cpathipa:
## Description 📝
**Fix Object Storage > Create Access Key Drawer Layout Issue**

| Before  | After   |
| ------- | ------- |
| image | image|
|image|image|

## How to test 🧪
1. **Navigate to Object storage and click create access key button**
2. **Validate drawer layout with no buckets.**
3. **Validate drawer layout with one or more buckets**

### DIFF
--- a/packages/manager/.changeset/pr-9296-fixed-1687442580014.md
+++ b/packages/manager/.changeset/pr-9296-fixed-1687442580014.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Fixed
+---
+
+Object Storage > Create Access Key Drawer Layout Issue ([#9296](https://github.com/linode/manager/pull/9296))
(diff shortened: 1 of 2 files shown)


bnussman-akamai:
I think it makes more sense to keep the drawer the same size and create a empty state in the table with some explanation to users on how they can create an Access Key, but that access key won't really be able to access any Bucket. 

I'm worried some users will be confused that they can create an Access Key that has no permissions

cpathipa:
> I think it makes more sense to keep the drawer the same size and create a empty state in the table with some explanation to users on how they can create an Access Key, but that access key won't really be able to access any Bucket.
> 
> I'm worried some users will be confused that they can create an Access Key that has no permissions

Thank you @bnussman-akamai! IMO, instead of showing empty table we could show some message to user. will sync with UX on this.

cpathipa:
Waiting for UX feedback.

cpathipa:
PR is ready for review, updated the copy from UX team.

abailly-akamai (approved):
Nice! Approving pending minor fix

cpathipa:
> Looks good!
> 
> Now I wonder if we should have a way for users to create full access tokens

As discussed with @bnussman-akamai  created a followup ticket to discuss this further if needed we could make it as feature in CM.

PR #9298, opened 2023-06-22
fix: Reduce Firewall create and update e2e test flakiness

jdamore-linode:
## Description 📝
This makes a few changes to the Firewall create and update e2e tests, as well as the Image capture test, to reduce flakiness.

Changes to address flake:
- Updates interactions involving Linode multi select fields to reflect recent UX improvements made to the component
- Adds a couple intercepts and waits for them to resolve during UI interactions

Other changes:
- Removes some UI interactions involving Firewall creation in update tests and replaces them with API calls
- Fixes a small accessibility issue with `LinodeSelectV2` involving its label `for` attribute not matching the input field's ID

## How to test 🧪
We can lean on the automated run for this, but you can run these tests locally to better confirm that flake is resolved:

`yarn && yarn start:manager:ci && yarn build`, and then
```bash
yarn cy:run -s "cypress/e2e/core/firewalls/update-firewall.spec.ts,cypress/e2e/core/firewalls/create-firewall.spec.ts"
```

To confirm the change involving `LinodeSelectV2`:
1. Navigate to a place on the app which uses it (Create Firewall drawer, for example)
2. Inspect the "Linodes" input field via your browser dev tools, and confirm that the related `<label />` element's `for` attribute matches the `id` attribute of the input

### DIFF
--- a/packages/manager/.changeset/pr-9298-fixed-1687455527847.md
+++ b/packages/manager/.changeset/pr-9298-fixed-1687455527847.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Fixed
+---
+
+Fix LinodeSelectV2 label association ([#9298](https://github.com/linode/manager/pull/9298))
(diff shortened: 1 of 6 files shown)


bnussman-akamai (approved):
Ran both create and update and got ✅

cpathipa (approved):

PR #9301, opened 2023-06-22
fix: [Marketplace] - Remove underscore in prometheus & grafana svg

hmorris3293:
We recently discovered that prometheus & grafana OCA was the only SVG using an underscore. It was causing issues with reporting so we figured the best course of action was to just rename that file.


packages/manager/public/assets/prometheusgrafana 2.svg
jaalah-akamai, line 1: I realize this was like this before, but can we remove the space in the filename?

jaalah-akamai (changes requested):
We may also need to update the filename here: https://github.com/linode/manager/blob/develop/packages/manager/src/features/OneClickApps/oneClickApps.ts#L1843

jaalah-akamai:
When you're all done - please add a changeset too
1. Run `yarn changeset` at project root
2. Select `manager > fix`
3. You can just use the title as the changeset `Remove underscore in prometheus & grafana svg`

Thanks for contributing @hmorris3293!

hana-linode (approved):

jaalah-akamai (approved):

PR #9306, opened 2023-06-23
fix: [M3-6720] - Restore DBaaS engine icons

jdamore-linode:
## Description 📝
This fixes the issue where DBaaS engine icons are no longer being shown in the engine select drop-down.

I think our Select component used to expect a callback which returned an icon element for this purpose, but now expects the icon element to be passed directly, and this one drop-down may have slipped through the cracks.

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-06-23 at 12 00 18 PM | Screenshot 2023-06-23 at 12 00 37 PM |

## How to test 🧪
1. Navigate to DBaaS create page
2. Click the "Database Engine" select
3. Confirm that icons are shown in drop-down


bnussman-akamai (approved):
Oops! This was probably my bad (see [this](https://github.com/linode/manager/pull/8996/files#diff-c6c9d2658ae5edce034d8c337a06f51ba0d990810ab8b95c57018fb1fec902c9)). 

I wonder if we can utilize Typescript types better so that this would have been caught

abailly-akamai (approved):
thank you!

PR #9294, opened 2023-06-21
refactor: [M3-6702] - React Query - Linodes - General clean up and refactors

bnussman-akamai:
## Description 📝
- The goal of this is to get rid of unnecessary Linode related hooks and contexts

## Major Changes 🔄
- Deleted `useEntities`, `useLinodes`, `useExtendedLinode`, `useExtendedLinodes`, `useLinodeActions`, `useReduxLoad` hooks
- Deleted `withLinodeActions`, `withLinodeDetailContext`, `LinodesDetail.container.tsx`
- Deleted lots of wrappers around Linode Details
- Made the Support Ticket Dialog pull Linodes from React Query
- Made search pull Linodes from React Query
- Switched out old hooks for React Query hooks

## Preview 📷
- Everything should look the same!

## How to test 🧪
- Test creating a support ticket that links a Linode (against a non-production environment please)
- Test the Linode Details page in general
- Test all Linodes features in general

PR #9318, opened 2023-06-27
refactor: Remove `withLoadingAndError` and clean up 2FA components

bnussman-akamai:
## Description 📝
- Delete `withLoadingAndError` HOC and clean up 2FA because this was the only place that consumed this component

## Major Changes 🔄
- Deletes `withLoadingAndError` high order component 🗑️
  - Doing this because it was only used in one place **and** this abstraction is unnecessary now that we have React Query, which tracks error and loading states 🎉
- Rewrites `DisableTwoFactorDialog` because it was a class component and it used `withLoadingAndError` 🧼
- Rewrites `ScratchCodeDialog` because it was a class component and an easy thing to clean up
- Removes `packages/manager/src/requestableContext.tsx` because I forgot to remove it in https://github.com/linode/manager/pull/9294

## Preview 📷
- No visual changes expected

## How to test 🧪
- Test 2FA flow of enabling and disabling 2FA


jaalah-akamai (approved):
👏

PR #9322, opened 2023-06-27
refactor: [M3-6794] - MUI v5 Migration - `Components > Box`

bnussman-akamai:
## Description 📝
- Style migration for our `<Box />` component

## Major Changes 🔄
- Moves Box from the `core` folder our our components folder
  - This makes the import look cleaner and deprecates the useless core folder 🗂️
- Uses a named export for the `<Box />` 🧼
- Adds a Storybook story for the `<Box />` 📝
  - This might be controversial but I see storybook as a tool for developers and I think we should expose as many components as possible

## Preview 📷
Screenshot 2023-06-27 at 12 06 42 PM

## How to test 🧪
- Do a sanity check of the entire app


packages/manager/src/components/Box.tsx
bnussman-akamai, line 1: I had to "extend" the box here to get the auto-doc args table to populate with props. I think this is fine because it gave me the opportunity to explain what the `<Box />` even is.

cpathipa (approved):

PR #9328, opened 2023-06-27
refactor: [M3-6798] - MUI v5 Migration - `Components > Typography`

bnussman-akamai:
## Description 📝
- Style migration for `<Typography />`

## Major Changes 🔄
- Move `packages/manager/src/components/core/Typography.ts` ➡️ `packages/manager/src/components/Typography.tsx`
- Use Storybook CSF 3 📝
- Use only named exports

Screenshot 2023-06-27 at 2 43 25 PM

## How to test 🧪
- Perform an overall check of the app for issues with text


packages/manager/src/components/Typography.tsx
bnussman-akamai, line 1: This is kind of 🤮 but I have to do this to get auto-docs working _I think_

bnussman-akamai: Okay wow, this actually causes some unit test failures

packages/manager/src/components/Typography.ts
bnussman-akamai, line 1: This is kind of 🤮 but it's one of the only methods I found that doesn't break any unit tests and still makes auto-doc somewhat work

jaalah-akamai: I think this is ok, we may want to have a wrapper for this eventually

jaalah-akamai (approved):

abailly-akamai (approved):
nice thanks for running with this one!

PR #9335, opened 2023-06-28
chore: [M3-6666] -  Clean up for consistent spelling of "canceled"

mjac0bs:
## Description 📝
There are two places in Cloud Manager ([image upload error toast](https://github.com/linode/manager/blob/develop/packages/manager/src/features/ToastNotifications/ToastNotifications.tsx#L140-L142), [entity transfer request status table column](https://github.com/linode/manager/blob/develop/packages/manager/src/features/EntityTransfers/RenderTransferRow.tsx#L170)) where we were performing string manipulation to edit an event message or status from API that returns "cancelled" rather than "canceled". Both spellings are correct -  "cancelled" is British-English and "canceled" is American English - but for consistency's sake, it would be better to use one spelling; "canceled" is already in use in all other places throughout the app.

API is now always returning the spelling "canceled" (see linked internal ticket), so this PR does some clean up to reflect those changes. 

## Major Changes 🔄
- 'Canceled' has a consistent spelling throughout the app without string manipulation needed.
- `EntityTransferStatus` types are updated to maintain parity with API. 

## Preview 📷
Should be no visual differences from prod, since we were already manually changing the spelling. Now we don't have to.

## How to test 🧪
- Check out this branch.
- `yarn up`
- Test this with the dev API.
2. **How to verify changes?**
- Upload an image in a valid region. After being redirected back to the image landing page, use the action menu to cancel the upload. Confirm that _no_ error toast displays with the "Upload canceled" message, which is the existing behavior in prod, where canceling an upload isn't considered an error and only the "image deleted" toast will display. 
(Note: There is an event message for a failed `image_upload` that does display when an upload is canceled, and it's not formatted particularly well due to the username tacked on the end. Thoughts welcome here; just didn't want to add more complexity to the already sketchy events message generator if no one is too bothered by this.)
- We can't make service transfers on dev, so turn on the MSW and observe the mocked entity transfer on the second page of the table with a status of 'canceled'. (http://localhost:3000/account/service-transfers?sent-transfers-table-page=2)
- Double check the API PR to make sure I didn't miss anything.
6. **How to run Unit or E2E tests?**
```
yarn cy:run -s "cypress/e2e/core/account/service-transfer.spec.ts"
```

### DIFF
--- a/packages/api-v4/.changeset/pr-9335-changed-1687980892035.md
+++ b/packages/api-v4/.changeset/pr-9335-changed-1687980892035.md
@@ -1,0 +1,5 @@
+---
+"@linode/api-v4": Changed
+---
+
+Use 'canceled' instead of 'cancelled' for EntityTransferStatus ([#9335](https://github.com/linode/manager/pull/9335))
--- a/packages/api-v4/src/entity-transfers/types.ts
+++ b/packages/api-v4/src/entity-transfers/types.ts
@@ -1,7 +1,7 @@
 export type EntityTransferStatus =
   | 'pending'
   | 'accepted'
-  | 'cancelled'
+  | 'canceled'
   | 'completed'
   | 'failed'
   | 'stale';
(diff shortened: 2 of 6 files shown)


abailly-akamai (approved):
Thanks for cleaning this up!

dwiley-akamai (approved):

PR #9337, opened 2023-06-28
chore: [M3-6803] - Use latest `@sentry/react` package and enable performance monitoring

bnussman-akamai:
## Description 📝
- Updates our Sentry package and makes sentry more useful

## Major Changes 🔄
- Uses `@sentry/react` rather than `@sentry/browser` because it is better for our use case 📦
- Enables Sentry's performance monitoring 🔎
  - This will duplicate some things New Relic does but it will give us some very nice information that we can all have access to seeing
  - This will give us better observability and debugability

## Preview 📷
Screenshot 2023-06-28 at 6 15 33 PM

## How to test 🧪
- Set `REACT_APP_SENTRY_URL` in your `packages/manager/.env` (ask me if you need the value)
- Throw an error in a component
- Verify you see the error in Sentry


packages/manager/src/App.tsx
```diff
@@ -34,7 +34,7 @@ import { tokenEventHandler } from './queries/tokens';
 import { volumeEventsHandler } from './queries/volumes';
 import { ApplicationState } from './store';
 import { getNextThemeValue } from './utilities/theme';
-import { ErrorBoundary } from './components/ErrorBoundary';
+import { ErrorBoundary } from '@sentry/react';
```
bnussman-akamai, line 37: Switching out for Sentry's `ErrorBoundry` in hopes we catch the most errors possible. It works the same as ours and will render the fallback as expected.

packages/manager/src/initSentry.ts
```diff
@@ -12,6 +17,7 @@ export const initSentry = () => {
       environment: import.meta.env.PROD ? 'production' : 'development',
       beforeSend,
       autoSessionTracking: false,
+      integrations: [new BrowserTracing()],
```
bnussman-akamai, line 20: Enables Sentry's performance feature 🎉

packages/manager/src/App.tsx
jaalah-akamai: Do we need to do this in the other 2 places we use our `ErrorBoundary`?

jaalah-akamai (approved):
🚀 📈

dwiley-akamai (approved):

packages/manager/src/App.tsx
bnussman-akamai: Good point. I'll just switch things over to the Sentry error boundy...

PR #9338, opened 2023-06-29
refactor: [M3-6299] - MUI v5 Migration - `Components > Checkbox` (part 2)

bnussman-akamai:
## Description 📝
- Style migration for the `<Checkbox />`
- Follow up to @dwiley-akamai's https://github.com/linode/manager/pull/8980

## Major Changes 🔄
- Use named exports 🧼
- Added CSF 3 Storybook Story 📝
  - Props are auto-documented now 🎉
- Removed `mdx` story
- Fixed flaws with the `<Checkbox />` component itself ☑️
- Deleted `components/code/Checkbox` and a few others 🗂️
- Renamed `CheckBox` to `Checkbox` for parity with MUI ✏️

Screenshot 2023-06-28 at 8 04 36 PM

## How to test 🧪
- Test checkboxes throughout the app
- Check the Storybook Story for checkboxes


packages/manager/src/components/Checkbox.tsx
```diff
@@ -0,0 +1,107 @@
...
+   * Styles applied to the `FormControlLabel`. Only works when `text` is defined.
+   */
+  sxFormLabel?: SxProps<Theme>;
+}
+
+/**
+ * ## Usage
+ *
+ * - Used when there are lists of options and the user may select any number of choices, including none, one, or many.
+ * - A stand-alone checkbox is used for a single option that the user can turn on or off (ie. accepting terms and conditions).
```
dwiley-akamai, line 34:
```suggestion
* - A standalone checkbox is used for a single option that the user can turn on or off (i.e., accepting terms and conditions).
```

dwiley-akamai (approved):
🚢

bnussman-akamai:
Ordering the props is probably a global changed so it's probably out of the scope of this ticket. @hana-linode 

With auto-doc it's unlikely we will get every args table perfect. I could put in some fixes but I ultimately think we shouldn't be putting too much time into perfecting storybook because at the end of the day it's just a tool for developers. 

@abailly-akamai Do you have any suggestions on how to prevent the args table from crashing? Do I need to manually specify an `argsType`?

jaalah-akamai:
I couldn't get the sorting to work properly either globally or locally: https://storybook.js.org/docs/react/essentials/controls#sorting-controls

I think editing `argTypes` for specific scenarios is fine. Right now it's very error prone not user-friendly. Updating the field to accept a string would suffice. We could always add a story for a more complex tooltip with a jsx element as opposed to thinking someone would manually input one (which they most likely won't)

```
toolTipText: {
  control: {
    type: 'text',
  },
}
```

To fix the `checked` issue, seems like adding the snippet below to the `default` story fixes the issue 
```
args: {
    checked: false,
  },
```

PR #9342, opened 2023-06-29
fix: Fix miscellaneous Cypress test flake

jdamore-linode:
## Description 📝
A couple small changes to slightly improve the resiliency of our tests.

- Most changes involve positioning intercepts and mocks before calls to `cy.visitWithLogin()`. Sometimes the desired API request is triggered before `cy.visitWithLogin()` has finished (since it waits for the page to finish loading before allowing the next command to execute, and some API requests are made in the meantime), so the desired API request is never intercepted and the test fails. This seems to happen more often on local runs than in CI.
- Fixes another potential cause of flake when interacting with a region select drop-down. I made the assertion more specific because the existing selector occasionally finds multiple matching elements which causes the test to fail.

## How to test 🧪
1. Wait for the automated run to finish
2. Confirm that there are no failures or flake caused by either of the two issues described above

### DIFF
--- a/packages/manager/cypress/e2e/core/account/default-payment-method.spec.ts
+++ b/packages/manager/cypress/e2e/core/account/default-payment-method.spec.ts
@@ -38,13 +38,10 @@ const gpayLastFourCcDefault = (ccDefault[1].data as CreditCardData).last_four;
 const ccLastFourGpayDefault = (gpayDefault[0].data as CreditCardData).last_four;
 
 describe('Default Payment Method', () => {
-  beforeEach(() => {
-    cy.visitWithLogin('/account/billing');
-  });
-
   it('makes google pay default', () => {
     mockGetPaymentMethods(ccDefault).as('getPaymentMethod');
     mockSetDefaultPaymentMethod(gpayIdCcDefault).as('changeDefault');
+    cy.visitWithLogin('/account/billing');
     cy.wait('@getPaymentMethod');
     cy.get(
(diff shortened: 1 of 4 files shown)


bnussman-akamai (approved):

mjac0bs (approved):
Neither of the flakes below look related to the issues you were targeting here. 🚀 

longview.spec.ts
`Timed out retrying after 300000ms: Expected not to find content: 'Waiting for data...' within the element: <div.MuiPaper-root.MuiPaper-elevation.MuiPaper-rounded.MuiPaper-elevation1.jss107.css-1j4ve6h> but continuously found it.`

smoke-linode-landing-table.spec.ts
`Timed out retrying after 80000ms: Expected to find element: :visible, but never found it. Queried from:
          > cy.get([data-qa-action-menu-item="Delete"])`
This test seems to have trouble with the action menu drop down occasionally, though. Anddd I just saw the other PR. :)

jdamore-linode:
> smoke-linode-landing-table.spec.ts `Timed out retrying after 80000ms: Expected to find element: :visible, but never found it. Queried from: > cy.get([data-qa-action-menu-item="Delete"])` This test seems to have trouble with the action menu drop down occasionally, though. Anddd I just saw the other PR. :)

Haha the `smoke-linode-landing-table.spec.ts` test flake while running against this PR was what got me looking at that test originally, but I opted to fix it separately since this PR had already gotten an approval by that point.

Haven't looked at the Longview flake yet but planning to soon!

PR #9348, opened 2023-06-29
fix: Fix Linode landing page delete smoke test flake

jdamore-linode:
## Description 📝
This fixes some flake in the Linode landing page test which deletes multiple Linodes in sequence. This test is flaky because it does not wait for the UI to update before proceeding with the next deletion, but when the UI does update it causes any action menus which are open to close. When this happens after Cypress has already clicked on the action menu, the test fails. 

## How to test 🧪
We can rely on the automated tests to confirm that this test passes without flake, but to test this locally you can run:

```bash
yarn cy:run -s "cypress/e2e/core/linodes/smoke-linode-landing-table.spec.ts"
```


bnussman-akamai (approved):

mjac0bs (approved):
Thanks for fixing, Joe!

Screenshot 2023-07-03 at 8 11 12 AM

PR #9349, opened 2023-06-30
fix: [M3-6627] - Fix `cy.defer()` TypeScript errors

jdamore-linode:
## Description 📝
This fixes the `cy.defer()` custom Cypress command signature and implementation to resolve TypeScript errors. The new implementation is a pretty thin wrapper around `cy.wrap()`, which accomplishes the same thing but isn't as ergonomic.

## How to test 🧪
1. Confirm that the tests continue to pass via the automated Cypress run
2. Check out this PR, and confirm there aren't any type errors in `cypress/support/commands.ts` at or around line 38.

### DIFF
--- a/packages/manager/cypress/e2e/core/account/service-transfer.spec.ts
+++ b/packages/manager/cypress/e2e/core/account/service-transfer.spec.ts
@@ -164,7 +164,7 @@ describe('Account service transfers', () => {
     cy.wait(['@getTransfers', '@getTransfers', '@getTransfers']);
 
     // Confirm that pending transfers are displayed in "Pending Service Transfers" panel.
-    cy.defer(getProfile()).then((profile: Profile) => {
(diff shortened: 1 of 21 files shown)


bnussman-akamai (approved):

jdamore-linode:
The StackScript failure was caused by Cypress timing out waiting for a Linode and an Image to be created which is not too concerning, but more concerning is the fact that using `cy.wrap()` seems to cause a lot of the context to be removed when Cypress reports an error waiting for a Promise to resolve (we've known about this a while with the before hook failure messages being kind of cryptic).

I'm going to circle back to this later this afternoon or early next week, but I'd like to find a way to make the errors more useful when something goes wrong while using `cy.wrap()`.

jaalah-akamai (approved):

PR #8717, opened 2023-01-12
M3-6068: Incorrect tab selection on Linode Create

cpathipa:
## Description 📝
Sets plan tab to "Dedicated"

## Preview 📷
image

**Remove this section or include a screenshot or screen recording of the change**

## How to test 🧪
Navigate to create linode
Validate the plan tab is set to default if no plan is selected.

PR #9351, opened 2023-06-30
test: [M3-6054] - Add Cypress test coverage for Linode Create via CLI dialog

jdamore-linode:
## Description 📝
This adds some test coverage to confirm the Linode Create via CLI dialog works as expected. I opted to add this as a separate test because I would like to eventually finish the e2e flow by executing the displayed command, but that'll be handled as a separate task later on.

## Major Changes 🔄
- Confirm that Linode create via CLI dialog shows expected cURL and Linode CLI snippets
- Confirm that Linode create via CLI dialog can be closed

## How to test 🧪
We can rely on the automated tests, but to run this spec locally you can use this command:

```bash
```


packages/manager/cypress/e2e/core/linodes/create-linode.spec.ts
```diff
@@ -32,4 +32,75 @@ describe('create linode', () => {
...
+      .should('be.visible')
+      .should('be.enabled')
+      .click();
+
+    ui.dialog
+      .findByTitle('Create Linode')
+      .should('be.visible')
+      .within(() => {
+        // Switch to cURL view if necessary.
+        cy.findByText('cURL').should('be.visible').click();
```
mjac0bs, line 76: Are there cases where the cURL tab isn't selected by default when the modal opens?

mjac0bs (approved):
🎉

packages/manager/cypress/e2e/core/linodes/create-linode.spec.ts
jdamore-linode: > Are there cases where the cURL tab isn't selected by default when the modal opens?

Not that I'm aware of! My thought here was that the initially selected tab is more of a design detail which could change, and I want the test to pass in either case. On the other hand, if the team thinks that this scenario would be more likely caused by a bug than an intentional design change then it would probably be wiser to remove this action and assert that cURL is the selected tab instead.

mjac0bs: This is a pretty minor detail so don't let it stop you from merging. :)

Just some follow up thoughts:

I'm didn't check what we're currently doing now with other e2es with tab panel content; if we're always assuming that initial active tab is a design detail and susceptible to change, then maybe it just makes sense to follow suit here. I do feel like I remember a bug that was had to fix in the create flow (plans table?) where the wrong tab was being selected, which made me think of tab selection here. (Oh, it was [this](https://github.com/linode/manager/pull/8717)).

> if the team thinks that this scenario would be more likely caused by a bug than an intentional design change

This might be a good question for UX and/or @cpathipa. My thought is that if our current practice app-wide is to default to the first tab as the active tab, then we should probably go with that assumption in our tests as well and treat divergence from that as a regression.

jaalah-akamai (approved):

packages/manager/cypress/e2e/core/linodes/create-linode.spec.ts
jdamore-linode: Thanks @mjac0bs !

> [...] what we're currently doing now with other e2es with tab panel content; if we're always assuming that initial active tab is a design detail and susceptible to change, then maybe it just makes sense to follow suit here

I didn't go back and search, but I'm almost certain we have instances of handling it both ways throughout the test suite -- some assume a given tab will already be selected, and others click on the tab they want even if it is already selected in reality. Helpful, I know 🙂.

> This might be a good question for UX and/or @cpathipa. My thought is that if our current practice app-wide is to default to the first tab as the active tab, then we should probably go with that assumption in our tests as well and treat divergence from that as a regression.

I'm inclined to agree. Of course open to hearing @cpathipa's thoughts, but in the meantime I've replaced the click with an assertion to confirm that the cURL tab is already selected upon opening the modal.

PR #9353, opened 2023-06-30
refactor: [M3-6790] – MUI v5 Migration - `Components > Divider`

dwiley-akamai:
## Description 📝
- Move `Divider.tsx` out of `/core`
- Use the `styled()` API in `Divider.tsx`
- Add some CSS in `AddNewMenu.tsx` to keep Create menu matching prod

## How to test 🧪
Look over the app and ensure that `<Divider />s appear as expected.


packages/manager/src/components/Divider.tsx
```diff
@@ -0,0 +1,33 @@
...
+  light?: boolean;
+  spacingTop?: number;
+  spacingBottom?: number;
+}
+
+const _Divider = (props: DividerProps) => {
+  return <StyledDivider {...props} />;
+};
+
+export default _Divider;
```
bnussman-akamai, line 17: Can we make this component use a named export?

```suggestion
```

packages/manager/src/components/Divider.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const StyledDivider = styled(Divider, {
+  label: 'StyledDivider',
+  shouldForwardProp: (prop) =>
+    isPropValid(['spacingTop', 'spacingBottom'], prop),
```
bnussman-akamai, line 22: Should `light` and `dark` also be filtered out? 

```suggestion
    isPropValid(['spacingTop', 'spacingBottom', 'light', 'dark'], prop),
```

bnussman-akamai (changes requested):

bnussman-akamai:
Also, can we add a Storybook Story just so the component is documented? This is all we'd need:

```ts
import React from 'react';
import { Meta, StoryObj } from '@storybook/react';
import { Divider } from './Divider';

const meta: Meta<typeof Divider> = {
  title: 'Components/Flag',
  component: Divider,
};

type Story = StoryObj<typeof Divider>;

export const Default: Story = {
  render: (args) => <Divider {...args} />,
};

export default meta;
```

packages/manager/src/components/Divider.tsx
dwiley-akamai: I don't mind adding them to the array, but I didn't see a console warning about them.

jaalah-akamai (approved):

PR #9355, opened 2023-06-30
fix: [M3-6177] - Only request logo once for invoice pdf

hana-linode:
## Description 📝
Previously, when downloading an invoice, we would make a request for the logo image for every page of the invoice potentially causing delays for large accounts. This PR fixes that by fetching the image only once and passing that to jsPDF's addImage function

## Preview 📷
image

## How to test 🧪
- Go to /account/billing
- In the browser dev tools, go to the network tab and filter by .png
- Download an invoice PDF and observe only 1 request made for the akamai logo
- Open the PDF and ensure that the logo is still on every page

### DIFF
--- a/packages/manager/src/features/Billing/PdfGenerator/PdfGenerator.ts
+++ b/packages/manager/src/features/Billing/PdfGenerator/PdfGenerator.ts
@@ -4,6 +4,7 @@ import {
   InvoiceItem,
   Payment,
 } from '@linode/api-v4/lib/account';
+import axios from 'axios';
 import jsPDF from 'jspdf';
 import { splitEvery } from 'ramda';
 import { ADDRESSES } from 'src/constants';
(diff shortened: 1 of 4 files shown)


jdamore-linode (approved):
Nice!

mjac0bs (approved):
🚀  Working nicely - I'm glad we are able to improve this, especially for large accounts!

packages/manager/src/features/Billing/PdfGenerator/PdfGenerator.ts
```diff
@@ -164,13 +165,25 @@ const addTitle = (doc: jsPDF, y: number, ...textStrings: Title[]) => {
...
+const getAkamaiLogo = () => {
+  return axios
+    .get(AkamaiLogo, { responseType: 'blob' })
+    .then((res) => {
+      return URL.createObjectURL(res.data);
+    })
+    .catch(() => {
+      return AkamaiLogo;
+    });
+};
```
bnussman-akamai, line 178: I think the same thing could be done with fetch, which removes the dependency on axios. But this is optional because the entire repo basically depends on Axios

```suggestion
const getAkamaiLogo = () => {
  return fetch(AkamaiLogo)
    .then(async (res) => {
      const blob = await res.blob();
      return URL.createObjectURL(blob);
    })
    .catch(() => {
      return AkamaiLogo;
    });
};
```

bnussman-akamai (approved):

PR #9361, opened 2023-07-03
feat: [M3-6722] – VPC endpoints, validation, & React Query queries

dwiley-akamai:
## Description 📝
Add initial VPC endpoints, validation, and RQ queries.

## Major Changes 🔄
- Add VPC endpoints
- Add VPC types
- Add `createVPCSchema` and `updateVPCSchema` for validation

## How to test 🧪
Reach out for detailed testing instructions if you want to work with the queries in alpha.

### DIFF
--- a/packages/api-v4/src/vpcs/types.ts
+++ b/packages/api-v4/src/vpcs/types.ts
@@ -1,0 +1,34 @@
+export interface VPC {
+  id: number;
+  label: string;
+  description: string;
+  region: string;
+  subnets: Subnet[];
+  created: string;
+  updated: string;
+}
+
+export interface CreateVPCPayload {
+  label: string;
+  description?: string;
+  region: string;
+  subnets?: SubnetPostObject[];
+}
+
+export interface UpdateVPCPayload {
+  label?: string;
+  description?: string;
+}
+
+export interface SubnetPostObject {
+  label: string;
+  ipv4: string;
+  ipv6: string;
+}
+
+export interface Subnet extends SubnetPostObject {
+  id: number;
+  linodes: number[];
+  created: string;
+  updated: string;
+}
--- a/packages/api-v4/src/vpcs/vpcs.ts
+++ b/packages/api-v4/src/vpcs/vpcs.ts
@@ -1,0 +1,81 @@
+import {
+  createVPCSchema,
+  updateVPCSchema,
+} from '@linode/validation/lib/vpcs.schema';
+import { BETA_API_ROOT as API_ROOT } from '../constants';
+import Request, {
+  setData,
+  setMethod,
+  setParams,
+  setURL,
+  setXFilter,
+} from '../request';
+import { Filter, ResourcePage as Page, Params } from '../types';
+import { CreateVPCPayload, UpdateVPCPayload, VPC } from './types';
+
+// VPC methods
+/**
+ * getVPCs
+ *
+ * Return a paginated list of VPCs on this account.
+ *
+ */
+export const getVPCs = (params?: Params, filter?: Filter) =>
+  Request<Page<VPC>>(
+    setURL(`${API_ROOT}/vpcs`),
+    setMethod('GET'),
+    setParams(params),
+    setXFilter(filter)
+  );
+
+/**
+ * getVPC
+ *
+ * Return details for a single specified VPC.
+ *
+ */
+export const getVPC = (vpcID: number) =>
+  Request<VPC>(
+    setURL(`${API_ROOT}/vpcs/${encodeURIComponent(vpcID)}`),
+    setMethod('GET')
+  );
+
+/**
+ * createVPC
+ *
+ * Create a new VPC in the specified region.
+ *
+ */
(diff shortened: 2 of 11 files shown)


packages/manager/src/App.tsx
```diff
@@ -50,6 +51,8 @@ const BaseApp = withFeatureFlagProvider(
     const { data: preferences } = usePreferences();
     const { mutateAsync: updateUserPreferences } = useMutatePreferences();
 
+    const { data: vpcData } = useVPCsQuery({}, {});
```
dwiley-akamai, line 54: Changes in this file are temporary, are only meant to facilitate testing, and will be removed prior to merging.

packages/validation/src/vpcs.schema.ts
```diff
@@ -0,0 +1,22 @@
...
+    .min(1, LABEL_MESSAGE)
+    .max(64, LABEL_MESSAGE)
+    .matches(
+      /[a-zA-Z0-9-]+/,
+      'Must include only ASCII letters, numbers, and dashes'
+    ),
+  description: string(),
+  region: string().required('Region is required'),
+  subnets: array().of(object()),
+});
```
jaalah-akamai, line 17: I think we need to check that the label 
- `Must not contain two dashes in a row` maybe using yup `.test()`
- `Must be unique among the customer’s VPCs` Although I can see us checking this elsewhere

packages/validation/src/vpcs.schema.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const updateVPCSchema = object({
+  label: string().notRequired().min(3, LABEL_MESSAGE).max(32, LABEL_MESSAGE),
+  description: string().notRequired(),
```
jaalah-akamai, line 21: Same question as above, just checking if any of these are truly not required?

packages/api-v4/src/vpcs/vpcs.ts
```diff
@@ -0,0 +1,81 @@
...
+    setXFilter(filter)
+  );
+
+/**
+ * getVPC
+ *
+ * Return details for a single specified VPC.
+ *
+ */
+export const getVPC = (vpcID: number) =>
```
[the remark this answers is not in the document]
dwiley-akamai: In general I think it's better to be explicit in the naming about what the parameter is, and I believe it's also standard naming convention in our JS client ([example](https://github.com/linode/manager/blob/580afef1b1b393f6d9d1cf2532db255e1379e427/packages/api-v4/src/firewalls/firewalls.ts#L44) and [example](https://github.com/linode/manager/blob/580afef1b1b393f6d9d1cf2532db255e1379e427/packages/api-v4/src/databases/databases.ts#L134))

packages/validation/src/vpcs.schema.ts
dwiley-akamai: I'll look into validating the first point, although the API spec mentions the conditions are the "same as VLAN labels today" so I matched [what we have for that currently](https://github.com/linode/manager/blob/580afef1b1b393f6d9d1cf2532db255e1379e427/packages/validation/src/linodes.schema.ts#L34-L40).

I don't think the second point can be validated/enforced within the schema itself. For other fields in the codebase that have to be unique from an API perspective, for example DBaaS labels, we don't have anything in the schema regarding that.

packages/api-v4/src/vpcs/types.ts
```diff
@@ -0,0 +1,34 @@
...
+export interface CreateVPCPayload {
+  label: string;
+  description?: string;
+  region: string;
+  subnets?: SubnetPostObject[];
+}
+
+export interface UpdateVPCPayload {
+  label?: string;
+  description?: string;
```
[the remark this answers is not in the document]
dwiley-akamai: Confirmed each is optional with the API team

packages/validation/src/vpcs.schema.ts
dwiley-akamai: Confirmed each is optional with the API team

packages/api-v4/src/vpcs/types.ts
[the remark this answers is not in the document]
hana-linode: Since it's an update, I would assume that either are optional

packages/validation/src/vpcs.schema.ts
```diff
@@ -0,0 +1,22 @@
...
+      /[a-zA-Z0-9-]+/,
+      'Must include only ASCII letters, numbers, and dashes'
+    ),
+  description: string(),
+  region: string().required('Region is required'),
+  subnets: array().of(object()),
+});
+
+export const updateVPCSchema = object({
+  label: string().notRequired().min(3, LABEL_MESSAGE).max(32, LABEL_MESSAGE),
```
hana-linode, line 20: Is there a reason for the label min to be 3 and max 32 when it's min 1 and max 64 in the POST schema? And should we be doing the same validation (only ASCII letters, numbers, and dashes)?

hana-linode (approved):
LGTM!

PR #9362, opened 2023-07-03
fix: [M3-6834]: Misleading MNTP percentage for accounts with no active services

mjac0bs:
## Description 📝
For accounts without any active Linode services, the MNTP percentage was displaying a misleading "You have used 100% of your Monthly Network Transfer Pool". This was due to `poolUsagePct` calculation: if `used` < `quota`, calculate the percentage; else, set the percentage to 100% because the `quota` must have been reached. 

In the case where usage was `0` and quota was `0`, this calculation resulted in 100%. Although technically a correct calculation, I think this seems misleading because it could cause customers to think they have reached a limit or are being billed for something they are not.

TODO:
- [x] Confirm UX copy with Product Marketing
- [x] Fix the link
- [x] Fix failing unit test

## Major Changes 🔄
- `poolUsagePct` evaluates to `0` if the data `used` is `0`
- When an account's usage is 0%, such as when there are no active services on the account, the MNTP text reads "You have used 0.00% of your Monthly Network Transfer Pool." and dialog shows a BarPercent at 0%.
- Fixes a console error for styled `BarPercent` whose DOM element was being passed the invalid prop `rounded` after the styled migration.
- Adds unit test file for `TransferDisplay`.

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screenshot 2023-07-03 at 10 14 50 AM |image |

## How to test 🧪

   - Use an account with no active services. This can also be mocked by updating the `accountTransferFactory` to 
   ``` 
    used: 0,
    quota: 0,
    billable: 0,
   ``` 
and turning on the MSW.

   - Go to the landing page and observe the MNTP is 100%, as in the Before screenshot above.
   - Check the browser console and note the console error when the MNTP dialog is opened. 
3. **How to verify changes?**
   - Go to the landing page and observe the MNTP is 0%, as in the After screenshot above.
   - Check the browser console and note there is no more console error when the MNTP dialog is opened. 
```
yarn test TransferDisplay
```

### DIFF
--- a/packages/manager/src/components/TransferDisplay/TransferDisplay.test.tsx
+++ b/packages/manager/src/components/TransferDisplay/TransferDisplay.test.tsx
@@ -1,8 +1,7 @@
+import { fireEvent } from '@testing-library/react';
 import React from 'react';
 import { renderWithTheme } from 'src/utilities/testHelpers';
-import { fireEvent } from '@testing-library/react';
 import { TransferDisplay } from './TransferDisplay';
-import { QueryClient } from 'react-query';
 
 const MockData = {
   used: 0,
@@ -23,9 +22,7 @@ jest.mock('../../queries/accountTransfer', () => {
 
 describe('TransferDisplay', () => {
   it('renders the MNTP button display text and opens the TransferDialog on click', async () => {
-    const { getByText, getByTestId } = renderWithTheme(<TransferDisplay />, {
-      queryClient: new QueryClient(),
-    });
+    const { getByText, getByTestId } = renderWithTheme(<TransferDisplay />);
 
     expect(
       getByText(transferDisplayButtonSubstring, { exact: false })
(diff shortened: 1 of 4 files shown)


abailly-akamai (approved):
Thanks for fixing! this looks good.

I was wondering about concerting with UX about this in general. because It's a bit odd to even display this modal at all if nothing's available. I wonder if there could be better messaging without the progress bar to just prompt the user with an incentive such as: "Create a resource to start checking your Monthly Network Transfer Pool" or show no modal at all.

packages/manager/src/components/TransferDisplay/TransferDisplay.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+  });
+
+  it('displays a percentage of 0.00% for no usage', async () => {
+    const { getByText } = renderWithTheme(<TransferDisplay />);
+
+    const usage = getByText(transferDisplayPercentageSubstring, {
+      exact: false,
+    }).innerHTML;
+    expect(usage).toMatch(/0.00%/);
+  });
```
bnussman-akamai, line 47: We could remove the `jest.mock` altogether if we use the MSW approach.

```suggestion
 it('displays a percentage of 0.00% for no usage', async () => {
    server.use(
      rest.get('*/account/transfer', (req, res, ctx) => {
        return res(ctx.json(MockData));
      })
    );
    const { findByText } = renderWithTheme(<TransferDisplay />);

    await findByText('0.00%', { exact: false });
  });
```

mjac0bs:
@abailly-akamai Good point, here's a current screenshot with a change for UX review:

Screenshot 2023-07-06 at 8 26 18 AM

jaalah-akamai (approved):

PR #9363, opened 2023-07-05
feat: [M3-6779] - AGLB api-v4 endpoints

bnussman-akamai:
## Description 📝
- Adds api-v4 endpoints and types for the AGLB

## Major Changes 🔄
- Adds all endpoints that should be needed for the AGLB project

## How to test 🧪
- No testing to be done, just review the endpoint, spelling, and typescript types
  - When reviewing, compare type to the API specification linked on the internal ticket

### DIFF
--- a/packages/api-v4/src/aglb/types.ts
+++ b/packages/api-v4/src/aglb/types.ts
@@ -9,11 +9,11 @@ export interface Loadbalancer {
 interface LoadbalancerPayload {
   name: string;
   regions: string[];
-  tags: string[];
+  tags?: string[];
 }
 
 export interface CreateLoadbalancerPayload extends LoadbalancerPayload {
-  entrypoints: EntrypointPayload[];
+  entrypoints?: EntrypointPayload[];
 }
 
 export interface UpdateLoadbalancerPayload extends LoadbalancerPayload {
--- a/packages/api-v4/src/aglb/types.ts
+++ b/packages/api-v4/src/aglb/types.ts
@@ -1,13 +1,13 @@
 export interface Loadbalancer {
   id: number;
   tags: string[];
-  name: string;
+  label: string;
   regions: string[];
   entrypoints: string[];
 }
 
 interface LoadbalancerPayload {
-  name: string;
+  label: string;
   regions: string[];
   tags?: string[];
 }
@@ -25,7 +25,7 @@ type Protocol = 'TCP' | 'HTTP' | 'HTTPS';
 type Policy = 'ROUND_ROBIN';
 
 export interface RoutePayload {
-  name: string;
+  label: string;
   rules: Rule[];
 }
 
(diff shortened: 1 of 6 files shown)


packages/api-v4/src/aglb/types.ts
```diff
@@ -0,0 +1,93 @@
...
+  tags: string[];
+  name: string;
+  regions: string[];
+  'entry-points': string[];
+}
+
+export interface CreateLoadbalancerPayload {
+  name: string;
+  regions: string[];
+  entrypoints: EntrypointPayload[];
```
jdamore-linode, line 12: I know this isn't on us, but wondering if we could get the API team to improve the consistency between `entry-points` on `Loadbalancer` and `entrypoints` on `CreateLoadbalancerPayload`.

The casing on these properties seems inconsistent on a broader level: some properties are `camelCase` (`CertificateTable.sniHostname`, for example), some are `snake_case` (which I think is generally the default?), and then there's this one `kebab-case` property. Do we think the API team might be willing to settle on one style and clean this up, or is there more to this?

jdamore-linode (approved):

packages/api-v4/src/aglb/types.ts
```diff
@@ -0,0 +1,88 @@
...
+  name: string;
+  regions: string[];
+  entrypoints: string[];
+}
+
+export interface CreateLoadbalancerPayload {
+  name: string;
+  regions: string[];
+  entrypoints: EntrypointPayload[];
+}
```
mjac0bs, line 13: Just confirming: we're excluding `tags` in this CreateLoadbalancePayload and the UpdateLoadbalancerPayload?

mjac0bs (approved):

packages/api-v4/src/aglb/types.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export interface Match {
+  path: string;
+}
+
+export interface ServiceTargetPayload {
+  name: string;
+  endpoints: Endpoint[];
+  ca_certificate: string;
+  load_bal_policy: Policy;
```
mjac0bs, line 70: API has this as `load_balancing_policy`, not sure if something has changed.

jaalah-akamai:
@bnussman-akamai this looks good pending your questions in the API spec, maybe we should add comments next to the things we're waiting to hear back from as to not forget about them 👍

jaalah-akamai (approved):

packages/api-v4/src/aglb/types.ts
bnussman-akamai: Added `tags`!

packages/api-v4/src/aglb/types.ts
bnussman-akamai: API spec had both... `load_balancing_policy ` is correct. Fixed!

PR #9372, opened 2023-07-06
test: [M3-6627] - Add region tests for Linodes

jdamore-linode:
## Description 📝
Adds end-to-end tests to our Cypress regions test suite to confirm that Linode create, update, and delete operations work end-to-end against all of our regions.

## Major Changes 🔄
- Add Linode create region tests
- Add Linode update (label, root password) region tests
- Add Linode delete (via details page) region tests

## How to test 🧪
At the moment, these tests can only be run locally. They're also slow because they create real resources against all of our regions. To speed things up, you can run the tests against only a single region:

(Note: You don't have to have Cloud running locally for this.)
```bash
CYPRESS_BASE_URL="https://cloud.linode.com" CY_TEST_SUITE="region" CY_TEST_REGION="us-east" yarn cy:run -s "cypress/e2e/region/linodes/*.spec.ts"
```

If you do want to run the tests against every region (which will be pretty slow), just omit the `CY_TEST_REGION` environment variable:

```bash
CYPRESS_BASE_URL="https://cloud.linode.com" CY_TEST_SUITE="region" yarn cy:run -s "cypress/e2e/region/linodes/*.spec.ts"
```

### DIFF
--- a/packages/manager/cypress/support/util/promises.ts
+++ b/packages/manager/cypress/support/util/promises.ts
@@ -7,32 +7,26 @@ import { timeout } from './backoff';
 /**
  * Resolves all of the given Promises in batches.
  *
- * @param promises - Promises to resolve.
+ * @param promiseGenerators - Functions that return Promises to resolve.
  * @param promisesPerPatch - How many Promises to attempt to resolve at once.
  * @param delayMs - Time in milliseconds to wait between batches.
  */
 export const resolveInBatches = async <T>(
-  promises: Promise<T>[],
+  promiseGenerators: (() => Promise<T>)[],
   promisesPerBatch: number = 3,
   delayMs: number = 1000): Promise<T[]> => {
     let results: T[] = [];
-    const unhandledPromises = [...promises];
-    console.log('a...');
+    const unhandledPromises = [...promiseGenerators];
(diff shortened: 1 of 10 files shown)


packages/manager/cypress/support/util/promises.ts
```diff
@@ -0,0 +1,32 @@
...
+import { timeout } from './backoff';
+
+/**
+ * Resolves all of the given Promises in batches.
+ *
+ * @param promiseGenerators - Functions that return Promises to resolve.
+ * @param promisesPerPatch - How many Promises to attempt to resolve at once.
+ * @param delayMs - Time in milliseconds to wait between batches.
+ */
+export const resolveInBatches = async <T>(
```
jdamore-linode, line 14: I wrote this because I was originally planning to create all of the Linodes needed for the update and delete tests upfront, and this util would've helped deal with rate limiting, but I ended up going in  another direction and didn't use this. I left it because I can see it coming in handy for other tests, though.

jaalah-akamai (approved):

cpathipa (approved):
@jdamore-linode We may need to update the description. I have to run CM locally to execute these tests. 
image

PR #9376, opened 2023-07-10
feat: [M3-6822] - Add initial routes, folders, pages, and tabs for AGLB

mjac0bs:
## Description 📝
Create basic files, folders, and routing for AGLB routes.

Initial routes in include:

Load Balancer Landing - `/loadbalancers`
Load Balancer Create - `/loadbalancers/create`
Load Balancer Details - `/loadbalancer/:id/:tab`
Route Landing - `/loadbalancers/routes`
Route Create - `/loadbalancers/routes/create`
Entry Point Landing - `/loadbalancers/entrypoints`
Entry Point Create - `/loadbalancers/entrypoints/create`

See self-review for a couple places I request feedback from the team.

## Preview 📷
Akamai Global Load Balancer Landing:
Screenshot 2023-07-10 at 12 27 04 PM

Load Balancer Details Landing:
Screenshot 2023-07-11 at 7 17 55 AM

## How to test 🧪
**How to setup test environment?**
Check out this branch.
**How to verify changes?**
Confirm that you can navigate to the following URLs:

Load Balancer Landing:
- http://localhost:3000/loadbalancers (url when the 'Load Balancers' tab is clicked) 
- http://localhost:3000/loadbalancers/routes (url when the 'Routes' tab is clicked) 
- http://localhost:3000/loadbalancers/entrypoints (url when the 'Service Targets' tab is clicked) 
- http://localhost:3000/loadbalancers/create (url when the 'Create Load Balancer' button is clicked from the Load Balancers tab) 
- http://localhost:3000/loadbalancers/routes/create (url when the 'Create Route' button is clicked from the Routes tab) 
- http://localhost:3000/loadbalancers/entrypoints/create (url when the 'Create Service Target' button is clicked from the Service Target tab) 

Load Balancer Details (use the URL bar for these; `1` is just an example id):
- http://localhost:3000/loadbalancer/1 (url when the 'Summary' tab is clicked) 
- http://localhost:3000/loadbalancer/1/entrypoints (url when the 'Entry Points' tab is clicked) 
- http://localhost:3000/loadbalancer/1/activity (url when the 'Activity Feed' tab is clicked) 
- http://localhost:3000/loadbalancer/settings (url when the 'Settings' tab is clicked) 

Also verify the structure of the folders/files makes sense. I followed other feature folders for this, but they weren't always consistent.

### DIFF
--- a/packages/manager/src/GoTo.tsx
+++ b/packages/manager/src/GoTo.tsx
@@ -84,7 +84,7 @@ const GoTo: React.FC<CombinedProps> = (props) => {
         href: '/volumes',
       },
       {
-        display: 'LoadBalancers',
+        display: 'Load Balancers',
         href: '/loadbalancers',
       },
       {
--- a/packages/manager/src/GoTo.tsx
+++ b/packages/manager/src/GoTo.tsx
@@ -84,6 +84,10 @@ const GoTo: React.FC<CombinedProps> = (props) => {
         href: '/volumes',
       },
       {
+        display: 'LoadBalancers',
+        href: '/loadbalancers',
+      },
+      {
         display: 'NodeBalancers',
         href: '/nodebalancers',
       },
--- a/packages/manager/src/GoTo.tsx
+++ b/packages/manager/src/GoTo.tsx
@@ -5,6 +5,7 @@ import EnhancedSelect, { Item } from 'src/components/EnhancedSelect/Select';
 import MUIDialog from 'src/components/core/Dialog';
 import { useHistory } from 'react-router-dom';
 import useAccountManagement from './hooks/useAccountManagement';
+import useFlags from './hooks/useFlags';
 
 const useStyles = makeStyles((theme: Theme) => ({
   paper: {
@@ -61,6 +62,7 @@ const GoTo: React.FC<CombinedProps> = (props) => {
   const classes = useStyles();
   const routerHistory = useHistory();
   const { _isManagedAccount, _hasAccountAccess } = useAccountManagement();
+  const flags = useFlags();
 
   const onSelect = (item: Item<string>) => {
     routerHistory.push(item.value);
@@ -84,6 +86,7 @@ const GoTo: React.FC<CombinedProps> = (props) => {
         href: '/volumes',
       },
       {
+        hide: !flags.aglb,
         display: 'Load Balancers',
         href: '/loadbalancers',
       },
--- a/packages/manager/src/features/LoadBalancers/EntryPoints/EntryPointCreate/EntryPointCreate.tsx
+++ b/packages/manager/src/features/LoadBalancers/EntryPoints/EntryPointCreate/EntryPointCreate.tsx
@@ -1,0 +1,11 @@
+import * as React from 'react';
+import { DocumentTitleSegment } from 'src/components/DocumentTitle/DocumentTitle';
+
+export const EntryPointCreate = () => {
+  return (
+    <>
+      <DocumentTitleSegment segment="Service Targets" />
+      TODO: AGLB M3-6817: Service Target Create
+    </>
+  );
+};
--- a/packages/manager/src/features/LoadBalancers/EntryPoints/EntryPointCreate/EntryPointCreate.tsx
+++ b/packages/manager/src/features/LoadBalancers/EntryPoints/EntryPointCreate/EntryPointCreate.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { DocumentTitleSegment } from 'src/components/DocumentTitle/DocumentTitle';
 
-export const EntryPointCreate = () => {
+const EntryPointCreate = () => {
   return (
     <>
       <DocumentTitleSegment segment="Service Targets" />
@@ -9,3 +9,5 @@ export const EntryPointCreate = () => {
     </>
   );
 };
+
+export default EntryPointCreate;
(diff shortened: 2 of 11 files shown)


packages/manager/src/GoTo.tsx
```diff
@@ -83,6 +83,10 @@ const GoTo: React.FC<CombinedProps> = (props) => {
         display: 'Volumes',
         href: '/volumes',
       },
+      {
+        display: 'Load Balancers',
+        href: '/loadbalancers',
+      },
```
mjac0bs, line 89: I added Load Balancers to our GoTo shortcut now, mostly for our convenience and because I didn't want it to be forgotten. Since it's an easter egg of a shortcut, I wouldn't expect it to get much use from anyone but us. If the team thinks it's better to hold off on including it in the list until later, I can remove and make a follow up ticket. 

Screenshot 2023-07-11 at 7 17 16 AM

packages/manager/src/features/LoadBalancers/EntryPoints/EntryPointCreate/EntryPointCreate.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
+import { DocumentTitleSegment } from 'src/components/DocumentTitle/DocumentTitle';
+
+const EntryPointCreate = () => {
+  return (
+    <>
+      <DocumentTitleSegment segment="Service Targets" />
+      TODO: AGLB M3-6817: Service Target Create
```
mjac0bs, line 8: I left many TODOs throughout these files with ticket numbers where applicable. The VS Code extensions [TODO Tree](https://marketplace.visualstudio.com/items?itemName=Gruntfuggly.todo-tree) and/or [TODO Highlight](https://marketplace.visualstudio.com/items?itemName=wayou.vscode-todo-highlight) seem like they will be useful, especially if we make a habit of formatting our todos consistently (e.g. `TODO: AGLB ...`) because you can add custom patterns to your config.

packages/manager/src/features/LoadBalancers/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+    <React.Suspense fallback={<SuspenseLoader />}>
+      <Switch>
+        <Route component={RouteCreate} path="/loadbalancers/routes/create" />
+        <Route
+          component={EntryPointCreate}
+          path="/loadbalancers/entrypoints/create"
+        />
+        <Route component={LoadBalancerCreate} path="/loadbalancers/create" />
+        <Route component={LoadBalancerDetail} path={`${path}/:id/:tab`} />
+        <Route component={LoadBalancerLanding} path="/loadbalancers/:tab?" />
```
mjac0bs, line 36: This is a place where I'm requesting any feedback or suggestions from the team:

The ticket specifies that Load Balancer details pages should have a path that includes the tab as a param. We do this other places throughout the app (Object Storage, Profile, NodeBalancers, Linodes). But in those places, the URL for the details page is not the same as the URL for the landing and/or tabs are not used in the URL of both pages. On a landing page with tabs:`/object-storage` vs details `/object-storage/buckets/:clusterId/:bucketName` or a landing page without tabs: `/nodebalancers` and details `/nodebalancers/:id`. 

The way we're doing this now, we can access the details page with `/loadbalancers/:id/:tab/` but not `/loadbalancers/:id`, because that will route to the landing page (`/loadbalancers/:tab?`). Results in different behavior than the rest of the app, where the `/summary` is not required to get to details pages.

abailly-akamai: I am not sure I understand too well this, but let's brainstorm here!

My initial thought would be to have two different URLs to differentiate between the landing and detail pages. for instance:

/loadbalancers/:tab (for landing)
/loadbalancer/:id/:tab (no "s" - for an instance of an LB)

Maybe it departs from our usual conventions and maybe I'm off cause I don't really understand the issue well, but would this be doable?

jaalah-akamai: I tend to agree with @abailly-akamai , drop the `s` and you won't need the `:tab` param for detail pages. When Banks returns we can have a larger discussion

mjac0bs: Yeah, I didn't explain very well, sorry about that. But yes, your suggestion made sense and is doable. It also breaks from convention (dropping the 's') in a more minor way that I first came up with (something like `loadbalancers/details/:id/:tab?`) which was what I wanted to avoid. 

See the changes in [36af7b2](https://github.com/linode/manager/pull/9376/commits/36af7b29a485593b2280b87126d8a50394a7f720), but I basically just needed to update the route in `MainContent.tsx` to allow us to route to the `LoadBalancer` component with both `/loadbalancers` and `/loadbalancer`. I'll update the PR description.

packages/manager/src/features/LoadBalancers/EntryPoints/EntryPointCreate/EntryPointCreate.tsx
cpathipa: Sounds good!

abailly-akamai (approved):
Looks great, nice work!

jaalah-akamai (approved):
🚀

cpathipa (approved):
Nice work! @mjac0bs!  Only thing I noticed is NaN in breadcrumb

image

mjac0bs:
@cpathipa We'll replace the id with the loadbalancer label name in the future.

As a note, I think I am seeing something slightly weird going on with the details page route and not making the active tab selected upon navigating to the page, but I'm going to merge this now so it's not blocking us and circle back to that on a details page ticket.

PR #9364, opened 2023-07-05
feat: [M3-6789] - Improve failed backup error messaging

mjac0bs:
## Description 📝
Error messaging for the backups service is minimal and does not provide users with much insight on the reason for a failure or whether they can address it.

Ultimately, we want to display more detailed error messages in Cloud Manager for failed backup restores and snapshot backups, which will require API to return these errors for Cloud to surface. This is ticketed and in the backlog: ARB-4001.

In the meantime, this PR addresses front-end improvements we can make to error messaging with the goal of (a) error prevention and (b) providing additional general context of limitations and requirements for the backup service.

We are updating generic error messages with a docs link to point users to potential reasons for the failure. We are also adding a warning notice in the linode create page to help prevent customers from encountering backup service failures due to unsupported disks being present in their configuration, which is a common reason backup restores fail. 

> **Note**: When testing, you may notice a bug where failed snapshot backups are returning incorrect `finished` timestamps (e.g. 2 seconds ago when the failure was 5 minutes ago). This is a known and ticketed API bug due to the way `duration` continues to accumulate for `linode_snapshot` events when the event is finished/failed. We're seeing it now due to recent changes (#9215) to update the timestamp for finished events, rather than use the created timestamp. 

TODO - Address UX feedback:
- [x] Fix the link color in dark mode on error toasts.

## Major Changes 🔄
- Updates notification menu/activity feed events for `backups_restore` and `linode_snapshot` failure events include link to doc.
- Updates error toasts for `backups_restore` and `linode_snapshot` failures to make them persistent, link to doc, and match the copy in the event failure message for consistency.
- Displays a warning notice above the backups section when a user enables Backups as an Add-on when creating or cloning a new linode with a custom image.
- Fixes a styling bug where the error notice on the Manual Snapshot section was displaying inline with textfield, rather than above.

## Preview 📷
| Before  | After   | Description |
| ------- | ------- | ------- |
| Screenshot 2023-07-04 at 10 05 33 PM | Screenshot 2023-07-04 at 10 06 00 PM | `linode_snapshot` error toast for failed manual snapshot backup |
| Screenshot 2023-07-04 at 10 02 22 PM | Screenshot 2023-07-04 at 10 03 02 PM | `linode_snapshot` event message for failed manual snapshot backup
| Screenshot 2023-07-05 at 10 53 23 AM | Screenshot 2023-07-05 at 10 48 28 AM | `backups_restore` event message for failure |
[...]

PR #9387, opened 2023-07-11
fix: [M3-6499] - Entity links in notification menu/events page

hana-linode:
## Description 📝
- Clicking on entity links in the notification menu no longer causes a hard refresh
  - Previously, the links were just `a` tags so they weren't connected with the react-router-dom. The fix was to use the `Link` component which required some refactoring to get `eventMessageGenerator` to return JSX instead of a string
- Fix backticked word display on the event page
- Fix fake entity links that shouldn't be links
- Remove extra space and quotation marks from support message link

## Preview 📷
| Before  | After   |
| ------- | ------- |
| event page before | event page after |

## How to test 🧪
```
yarn test eventMessageGenerator
```

- Check various events in the notification menu and event page
- Ensure that previous known regressions are not present e.g.
  - #9280
  - #9360


packages/manager/src/eventMessageGenerator.tsx
```diff
@@ -828,8 +830,7 @@ export default (e: Event): string => {
    * fails gracefully if the message we encounter a formatting error
    * */
   try {
-    const formattedMessage = applyLinking(e, messageWithUsername);
-    return applyBolding(e, formattedMessage);
+    return applyLinkingAndBolding(e, messageWithUsername, bold);
```
hana-linode, line 833: With these new changes, I don't think we need to wrap the code in a try/catch block since we're not doing any fancy regex anymore, but curious to hear the team's thoughts

packages/manager/src/eventMessageGenerator.tsx
```diff
@@ -899,27 +890,102 @@ export function applyLinking(event: Event, message: string) {
...
-}
+  return createMessageJSX(message, '', '', bold);
+};
+
+const createMessageJSX = (
+  message: string,
+  linkLabel: string,
+  linkTarget: string,
+  bold = false
+) => {
```
hana-linode, line 918: Sorry in advance for this ugly code 😅

This was the best solution I could come up with to turn a string into JSX without bringing in a new library/package

packages/manager/src/eventMessageGenerator.test.tsx
```diff
@@ -95,64 +99,68 @@ describe('Event message generation', () => {
...
-    it('should not replace entity label if label is inside backticks', () => {
+    it('should replace backticks with code tag', () => {
       const mockEvent = eventFactory.build({ entity });
       const message = 'created `foo`';
-      const result = applyLinking(mockEvent, message);
+      const { container } = render(applyLinkingAndBolding(mockEvent, message));
 
-      expect(result).toEqual('created `foo`');
+      // eslint-disable-next-line xss/no-mixed-html
+      expect(container.firstElementChild).toContainHTML('<code>foo</code>');
```
jaalah-akamai, line 141: Instead of `{container}` do search the DOM, can't we use `getByRole('code')`?

packages/manager/src/eventMessageGenerator.test.tsx
```diff
@@ -95,64 +99,68 @@ describe('Event message generation', () => {
...
-        }),
-      });
+      const mockEvent = eventFactory.build({ entity });
       const message = 'created entity Null label';
-      const result = applyLinking(mockEvent, message);
 
-      expect(result).toEqual('created entity Null label');
+      const { container } = render(applyLinkingAndBolding(mockEvent, message));
+
+      expect(container.firstElementChild).toHaveTextContent(message);
```
jaalah-akamai, line 163: Same here, let see if we can use a RTL query

packages/manager/src/eventMessageGenerator.tsx
```diff
@@ -899,27 +890,102 @@ export function applyLinking(event: Event, message: string) {
...
+          return (
+            <>
+              <code>{word.substring(1, word.length - 1)}</code>{' '}
+            </>
+          );
+        }
+        if (bold && wordsToBold.includes(word)) {
+          return (
+            <>
+              <strong key={word}>{word}</strong>{' '}
```
jaalah-akamai, line 952: We should add the index to the key in case we have similar words to avoid conflict. This is for every case as well

packages/manager/src/eventMessageGenerator.tsx
```diff
@@ -899,27 +890,102 @@ export function applyLinking(event: Event, message: string) {
...
+  return (
+    <Typography>
+      {words.map((word) => {
+        if (word === '``') {
+          return `${word} `;
+        }
+        if (word.startsWith('`')) {
+          return (
+            <>
+              <code>{word.substring(1, word.length - 1)}</code>{' '}
```
jaalah-akamai, line 945: I think we would need a key here

packages/manager/src/eventMessageGenerator.tsx
hkhalil-akamai: Parsing is notoriously one of the most difficult challenges in computer science so it's no surprise the code gets pretty gnarly here. While I think this works for the most part, I wonder how this could be made more readable, especially if we need to add more rules/functionality in the future.

It may be easier to process the message in 'passes', such as a bolding pass, linking pass, and <code> pass in three separate functions. This will allow being more specific in how the message is split up (e.g., for the <code> pass, instead of splitting word-by-word, the message could be split to focus on sections wrapped in the backtick).

packages/manager/src/eventMessageGenerator.tsx
```diff
@@ -899,27 +890,102 @@ export function applyLinking(event: Event, message: string) {
...
-  return newMessage;
-}
+  return createMessageJSX(message, '', '', bold);
+};
+
+const createMessageJSX = (
+  message: string,
+  linkLabel: string,
+  linkTarget: string,
+  bold = false
```
jaalah-akamai, line 917: I think it's a good habit that when function parameters are more than two, to this into an object so that we have some context for what's being passed in. If we ever need to extend this or make something optional in the future we'll have that flexibility

hkhalil-akamai (changes requested):
Great work at accomplishing a very technically complicated task.

Noticing some issues in testing:

1. The `formatEventWithAppendedText` function adds anchor tags to the messages in some cases (i.e., "Snapshot backup failed" from #9364), causing raw HTML to appear in the event message

    - Screenshot: Screenshot 2023-07-11 at 5 30 16 PM
    - It might be best to update that function to return `string | JSX` instead of inline HTML


2. There can sometimes be spaces inside of backticks (e.g., TXT domain records), causing the `<code>` formatting to break
    - Screenshot:  Screenshot 2023-07-11 at 5 32 43 PM
    - This might be complicated to fix using the current word-by-word approach. See my suggestion about using multiple passes as an alternative

3. There can sometimes be backticks within entity labels and even within other backticks
    - E.g., TXT records or support ticket titles can contain backticks
    - I'm not sure we can do much here besides ensuring starting backticks have a matching ending backtick, since the API does not sanitize user-specified labels to be free of Markdown (so it's impossible to distinguish a user-specified backtick from a backtick added by the API)

hkhalil-akamai:
Unrelated to this PR specifically, I'm not very keen on the way we format event messages in general between the API and the frontend. This PR makes a good attempt at improving that functionality but I think we need to fundamentally rethink how we do this.

My opinion on the best approach would be to generate the messages ourselves on the FE, using the event data like entity id, action, etc. This avoids needing to parse at all.

hkhalil-akamai:
Do we want to include formatting in the `EventsLanding` component? Previously, things like bolding and `<code>` formatting only appeared in the menu and not in the EventsLanding, although I'm not sure if that was intentional.

jaalah-akamai:
> My opinion on the best approach would be to generate the messages ourselves on the FE, using the event data like entity id, action, etc. This avoids needing to parse at all.

This isn't a bad idea if all the parts are coming from the api already. Probably a discussion for our cafe

hana-linode:
Closing this for now as these changes introduce more issues than it fixes and we want to go in a direction where we do not parse the messages. An epic (M3-6891) has been created to look into refactoring event messaging more holistically.

PR #9390, opened 2023-07-11
feat: [M3-6723] – VPC Subnet endpoints, validation, & React Query queries

dwiley-akamai:
## Description 📝
Add initial VPC subnet endpoints, validation, and RQ queries.

## Major Changes 🔄
- Add subnet endpoints
- Add subnet types
- Add `createSubnetSchema` and `modifySubnetSchema` for validation, plus a few constants and a utility function for testing IP validity

## How to test 🧪
Reach out for detailed testing instructions if you want to work with the queries in alpha.


packages/manager/src/App.tsx
dwiley-akamai, line 1: Like last time, all of the changes in this file are temporary for testing purposes and will be removed prior to merging.

You can uncomment out the lines and replace `vpcID` and `subnetID` with real IDs to check the data is coming through.

packages/manager/src/queries/vpcs.ts
```diff
@@ -57,8 +67,101 @@ export const useDeleteVPCMutation = (id: number) => {
...
+};
+
+export const useSubnetQuery = (vpcID: number, subnetID: number) => {
+  return useQuery<Subnet, APIError[]>(
+    [vpcQueryKey, 'vpc', vpcID, subnetQueryKey, 'subnet', subnetID],
+    () => getSubnet(vpcID, subnetID)
+  );
+};
+
+export const useCreateSubnetMutation = (vpcID: number) => {
```
dwiley-akamai, line 96: The invalidation in the `onSuccess` functions for `useCreateSubnetMutation`, `useUpdateSubnetMutation`, and `useDeleteSubnetMutation` can probably be streamlined/improved, but I'd rather optimize those later in the process in tickets where those queries are made use of as opposed to over-optimizing at this early stage.

packages/manager/src/App.tsx
```diff
@@ -58,6 +59,9 @@ const BaseApp = withFeatureFlagProvider(
 
     const { enqueueSnackbar } = useSnackbar();
 
+    // const { data: subnetData } = useSubnetQuery(vpcID, subnetID);
+    // console.log(subnetData);
```
dwiley-akamai, line 63: Like last time, all of the changes in this file are temporary for testing purposes and will be removed prior to merging.

You can uncomment out the lines and replace vpcID and subnetID with real IDs to check the data is coming through.

packages/validation/src/vpcs.schema.ts
```diff
@@ -1,21 +1,57 @@
...
       (value) => !value?.includes('--')
     )
     .required('Label is required')
     .min(1, LABEL_MESSAGE)
     .max(64, LABEL_MESSAGE)
-    .matches(
-      /[a-zA-Z0-9-]+/,
-      'Must include only ASCII letters, numbers, and dashes'
-    ),
+    .matches(/[a-zA-Z0-9-]+/, LABEL_REQUIREMENTS),
```
jaalah-akamai, line 54: Since the label validation is essentially the same for all the schemas, what do you think about re-using that? `const labelValidation = ...` then wherever it's used we can tack on `.required()` or `.notRequired()`?

jaalah-akamai: Can we also confirm that the regex we're using is sufficient? I think we may need to be more strict by doing `/^[a-zA-Z0-9-]+$/`.

packages/validation/src/vpcs.schema.ts
```diff
@@ -1,21 +1,57 @@
...
+  testMessage: 'Must not contain two dashes in a row',
+};
+
+const validateIP = (value?: string): boolean => {
+  if (!value) {
+    return false;
+  }
+
+  const addr = ipaddr.parse(value);
+  const type = addr.kind();
```
jaalah-akamai, line 18: I think we want a single try catch around all of this, even the `ipaddr.parse(value);`.  Parsing the IP address value could potentially throw an error

packages/validation/src/vpcs.schema.ts
hana-linode: I think the regex should be sufficient since we're using the same one in the Linodes schema https://github.com/linode/manager/blob/develop/packages/validation/src/linodes.schema.ts#L38

hana-linode (approved):

packages/validation/src/vpcs.schema.ts
dwiley-akamai: In the cases of `"test "`, `" test "`, and `" test"` I checked to see if the current VLAN validation would catch those and it did, so I think what we have currently should be sufficient

packages/validation/src/vpcs.schema.ts
```diff
@@ -14,75 +16,62 @@ const validateIP = (value?: string): boolean => {
...
     }
 
-    return true;
-  }
+    if (type === 'ipv6') {
+      try {
+        ipaddr.IPv6.parseCIDR(value);
+      } catch (err) {
+        return false;
+      }
```
jaalah-akamai, line 39: If we needed different error handling for these, this would be fine... but we can just have a single try/catch block now that we wrapped everything:
```
try {
  const addr = ipaddr.parse(value);
  if (addr.kind() === 'ipv4') {
    ipaddr.IPv4.isValid(value);
    ipaddr.IPv4.parseCIDR(value);
  } else if (addr.kind() === 'ipv6') {
    ipaddr.IPv6.parseCIDR(value);
  } else {
    return false;
  }

  return true;
} catch (err) {
  return false;
}
```

jdamore-linode (approved):

jaalah-akamai (approved):

PR #9395, opened 2023-07-13
chore: Fix broken Linode landing table top bar navigation test

jdamore-linode:
## Description 📝
Recent improvements to the Linode top bar navigation broke the Linode landing table tests which check those elements. This fixes that.

Because we replaced links with buttons for navigation, we can no longer test that the Linode Cloud Community button navigates the user to the expected place because it also opens in a separate tab, which Cypress is unable to interact with. Previously we asserted that the `href` attribute had an expected value but this is no longer available.

## How to test 🧪
We'll wait for the automated run to confirm that `smoke-linode-landing-table.spec.ts` passes.


carrillo-erik:
Awesome! I was able to validate the fix on my local and by running your PR on my local.

carrillo-erik (approved):

dwiley-akamai (approved):

PR #9402, opened 2023-07-13
refactor: [M3-6827] - Use RQ for `isLargeAccount`

hkhalil-akamai:
## Description 📝
Creates a new hook to derive `isLargeAccount` from React Query.

## How to test 🧪
1. Using MSW, set the number of Linodes to something greater than 1500 (the current `LARGE_ACCOUNT_THRESHOLD`).
2. Ensure `useIsLargeAccount` returns the `true` (e.g., Search uses API filtering instead of a client-side search)


packages/manager/src/hooks/useIsLargeAccount.ts
hkhalil-akamai, line 1: Decided to make this its own hook since there's no need to fetch all the data that `useAccountManagement` needs when you just want `isLargeAccount`.

packages/manager/src/components/AuthenticationWrapper/AuthenticationWrapper.tsx
```diff
@@ -80,8 +79,8 @@ export class AuthenticationWrapper extends React.Component<CombinedProps> {
         queryFn: getAccountSettings,
       }),
 
-      // Is this a large account? (should we use API or Redux-based search/pagination)
-      this.props.checkAccountSize(),
+      // Fetch first page of Linodes (TODO: remove once the Linodes RQ migration is complete)
+      this.props.getLinodesPage(),
```
hkhalil-akamai, line 83: The `checkAccountSize()` call originally requested the first 100 linodes. Doing this directly here now until this  component is RQ-ified.

packages/manager/src/components/AuthenticationWrapper/AuthenticationWrapper.tsx
```diff
@@ -170,17 +169,16 @@ const mapStateToProps: MapState<StateProps, {}> = (state) => ({
...
 }
 
 const mapDispatchToProps: MapDispatchToProps<DispatchProps, {}> = (
   dispatch: ThunkDispatch<ApplicationState, undefined, Action<any>>
 ) => ({
   initSession: () => dispatch(handleInitTokens()),
-  checkAccountSize: () => dispatch(checkAccountSize()),
-  requestLinodes: () => dispatch(requestLinodes({})),
+  getLinodesPage: () =>
+    dispatch(getLinodesPage({ params: { page_size: 100 } })),
```
bnussman-akamai, line 181: I think this will cause an extra fetch because this isn't stored in RQ. I think we want to basically prefetch "useLinodesQuery" instead of just calling `getLinodesPage`

dwiley-akamai (commented):
Looks good to me aside from the `getLinodesPage` point that Banks raised

packages/manager/src/components/AuthenticationWrapper/AuthenticationWrapper.tsx
hkhalil-akamai: `getLinodesPage` is defined in `linodes.requests.ts` and is used to initialize the Linodes Redux store. I think this is necessary until we completely replace the Linodes Redux store with RQ since there are still parts of the app that get Linodes from Redux (e.g., `LinodesLanding`).

hkhalil-akamai: For context of why I needed to add this line, [see here](https://github.com/linode/manager/pull/9402#discussion_r1263856574). Basically, `getLinodesPage` was being dispatched indirectly through `checkAccountSize`. Since `checkAccountSize` is no longer needed (as it's handled by the new `isLargeAccount` hook), I moved this call directly to the Auth wrapper so components that still use the Linodes Redux store can continue to work.

bnussman-akamai: Ohhh. I see! My fault!

What breaks if we don't fetch linodes for our Redux store? There might be stuff still using Redux that I overlooked or forgot about

bnussman-akamai (approved):

packages/manager/src/components/AuthenticationWrapper/AuthenticationWrapper.tsx
hkhalil-akamai: A few places according to VSCode. These will be refactored in a follow-up PR:
Screenshot 2023-07-17 at 5 33 20 PM

packages/manager/src/hooks/useAccountManagement.ts
dwiley-akamai, line 1: we should remove `_isLargeAccount` on L11 in this file as well

dwiley-akamai (approved):
Looks good, left a couple of small comments

bnussman-akamai:
I strongly agree with removing the underscore prefixes but that can be done at a later point if you want to merge!

PR #9406, opened 2023-07-14
refactor: [M3-6325] – MUI v5 Migration - `SRC > Components > DocumentTitle`

dwiley-akamai:
## Description 📝
Bring `DocumentTitle.tsx` up to speed with our current code conventions and restore the dynamic document title functionality that was accidentally lost after a [recent refactor](https://github.com/linode/manager/pull/9256).

## Major Changes 🔄
- Refactor class components and old patterns out of `DocumentTitle.tsx`
- Restore dynamic document title functionality 🎉 

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screenshot 2023-07-14 at 10 56 18 AM | Screenshot 2023-07-14 at 10 56 54 AM |

## How to test 🧪
Go through the app and ensure that the tab name gets updated with the appropriate section. For example, on the Linodes landing page, you should see `Linodes | Akamai Cloud Manager`.


packages/manager/src/App.tsx
```diff
@@ -43,244 +46,252 @@ import {
 // Ensure component's display name is 'App'
 export const App = () => <BaseApp />;
 
-const BaseApp = withFeatureFlagProvider(
-  withFeatureFlagConsumer(() => {
-    const history = useHistory();
+const BaseApp = withDocumentTitleProvider(
```
dwiley-akamai, line 49: The diff in this file is mostly indentation, but we're just wrapping this code with `withDocumentTitleProvider` so the dynamic document title functionality gets restored.

packages/manager/src/components/DocumentTitle/DocumentTitle.tsx
```diff
@@ -21,76 +22,69 @@ interface Props {
...
+      {(value: DocumentTitleSegmentsContext) => (
+        <InnerDocumentTitleSegment segment={segment} {...value} />
+      )}
+    </DocumentTitleSegmentsConsumer>
+  );
+};
+
+type Wrapper = (Component: React.ComponentType<{}>) => React.FC<unknown>;
+
+export const withDocumentTitleProvider: Wrapper = (
```
dwiley-akamai, line 60: This returns JSX so I believe we'll want to keep this as a wrapper.

hana-linode (approved):
Confirmed that document title has been restored! 🚀 

image

packages/manager/src/components/DocumentTitle/DocumentTitle.tsx
mjac0bs, line 1: Do you think it's reasonable to add a short comment in this file just explaining/giving an example what the Document Title is? I believe it got removed in that PR a while back because it wasn't totally clear what its use was, and I actually didn't know either when looking at other landing pages and doing some AGLB work.

mjac0bs (approved):
It's back! 🚢  

Just left one comment in hopes of making the purpose/use of the component more clear.

PR #9420, opened 2023-07-18
refactor: [M3-6342] - MUI v5 Migration - SRC > Component > GaugePercent

tyler-akamai:
## Description 📝
Ran Codemod against SRC > Component > GaugePercent

## Major Changes 🔄
- [x] Run codemod
- [x] Remove React.FC and rename interface 'Props' to '{ComponentName}Props'
- [x] Remove 'CombinedProps' 
- [x] Remove makeStyles for new styled API. We typically use 'Styled ComponentName) as our prefix
- [x] Export component inline
- [x] Any style attributes can be converted to 'sx prop' or 'styled API'
- [x] Move styled api calls to separate file (< 100 lines)

## Preview 📷
Before:
Screenshot 2023-07-18 at 5 00 59 PM

After:
Screenshot 2023-07-18 at 5 02 12 PM

## How to test 🧪
Go to the Longview section and make sure the functionality and the visual components are the same.

Screenshot 2023-07-18 at 5 04 23 PM

Note: You might need to setup a Longview client: 
1. Create a new Linode (ex. Debian 11, Debian 12 does not work)
2. Create a new Longview client (Click the "Add Client" button)
3. Copy the installation curl command
4. SSH into your Linode and paste the installation command

Reference: https://www.linode.com/docs/products/tools/longview/guides/troubleshooting/

### DIFF
--- a/packages/manager/src/components/GaugePercent/GaugePercent.stories.mdx
+++ b/packages/manager/src/components/GaugePercent/GaugePercent.stories.mdx
@@ -9,11 +9,11 @@ import { Typography } from 'src/components/Typography';
 - **Chart.js** is the charting tool we use for gauges
 - Best used to show percentages
 
-export const Template = (args) => <Gauge {...args} />;
+export const Template = (args) => <GaugePercent {...args} />;
 
 <Canvas>
   <Story
-    name="Gauge"
+    name="GaugePercent"
     args={{
       value: 50,
       max: 200,
@@ -79,4 +79,4 @@ export const Template = (args) => <Gauge {...args} />;
   </Story>
 </Canvas>
 
-<ArgsTable story="Gauge" sort="requiredFirst"/>
+<ArgsTable story="GaugePercent" sort="requiredFirst"/>
--- a/packages/manager/src/components/GaugePercent/GaugePercent.stories.mdx
+++ b/packages/manager/src/components/GaugePercent/GaugePercent.stories.mdx
@@ -1,5 +1,5 @@
 import { Canvas, Meta, Story, ArgsTable } from '@storybook/addon-docs';
-import Gauge from './GaugePercent';
+import { GaugePercent } from './GaugePercent';
 import { Typography } from 'src/components/Typography';
 
 <Meta title="Components/Analytics & Graphs" component={Gauge} />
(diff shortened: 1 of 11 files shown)


packages/manager/src/components/GaugePercent/GaugePercent.styles.ts
```diff
@@ -0,0 +1,40 @@
...
+  fontSize: innerTextFontSize || theme.spacing(2.5),
+  position: 'absolute',
+  textAlign: 'center',
+  top: `calc(${height}px + ${theme.spacing(1.25)})`,
+  width: width,
+}));
+
+export const StyledInnerTextDiv = styled('div', {
+  shouldForwardProp: (prop) => isPropValid(propKeys, prop),
+})<StyledGaugePercentProps>(({ theme, height = 300, width }) => ({
```
jaalah-akamai, line 25: `height = 300` may be redundant since we have this as a default where the prop is defined in the component

jaalah-akamai (approved):
Nice job - be sure to run `yarn changeset` and this is looking good

packages/manager/src/components/GaugePercent/GaugePercent.styles.ts
tyler-akamai: Fix:
Since the height is an optional parameter in GaugePercentProps, it was throwing a 'possible undefined' error in the string literal. So I just made the StyledGaugePercentProps interface pull from GaugePercentProps and make the width and height required. GaugePercentProps has the as optional, but if they are undefined, they are set within the component with a default value of 300.

packages/manager/src/components/GaugePercent/GaugePercent.stories.mdx
```diff
@@ -1,5 +1,5 @@
 import { Canvas, Meta, Story, ArgsTable } from '@storybook/addon-docs';
-import Gauge from './GaugePercent';
+import { GaugePercent } from './GaugePercent';
 import { Typography } from 'src/components/Typography';
 
 <Meta title="Components/Analytics & Graphs" component={Gauge} />
```
mjac0bs, line 5:
```suggestion
<Meta title="Components/Analytics & Graphs" component={GaugePercent} />
```

You're also going to need to change the name on L12: `export const Template = (args) => <GaugePercent {...args} />;` and I'd recommend L16 and L82, just for consistency with the actual component name.

mjac0bs (approved):
Congrats on your first PR! 🎉  Approving pending the issue below is addressed.

One small fix to make: good fix in the storybook file, but it's not quite working yet. You'll want to rename `Gauge` to `GaugePercent` where the component is being referenced after updating and renaming that import. Otherwise, the story is broken.

We can run storybook locally with `yarn storybook` in the the terminal to start up Storybook at localhost:6006 and check that the story is visible without errors.

Thanks for the instructions on how to set up a longview client!

PR #9427, opened 2023-07-19
fix: [M3-6728] - Make drawers full width on mobile viewports 

coliu-akamai:
## Description 📝
Ensure that drawers are full width on mobile viewports

## Major Changes 🔄
- change drawer container to have width of 100% when viewport is less than the theme's 'sm' breakpoint (which is 600px). Added an additional maxWidth css property because there were some cases where the drawer just didn't look great with just `width: 100%` (for if viewport is ~450-600px, will show screenshots in comments), but can remove this if needed

## Preview 📷
(the orange strip on the after is just the laptop background)
| Before  | After   |
| ------- | ------- |
| image | image |

## How to test 🧪
1. ensure that all drawers throughout site are full width on mobile viewports (caveat -- when the viewport is 450-600px, the drawer is not yet full width, will put up some screenshots) + ensure drawer functionality still works as expected


coliu-akamai:
Added maxWidth property -- with just` width: 100% ` and no maxWidth property, the drawer would look like the image below from viewport sizes ~450-600px. Adding the maxWidth property makes it so that the drawer is still not the viewport width from 445px-600px, but probably looks a bit better than having the drawer content not be aligned well (though in fairness, not too sure how many devices have a viewport width between 450-600px)

image

packages/manager/src/components/Drawer/Drawer.tsx
```diff
@@ -47,7 +47,8 @@ const useStyles = makeStyles()((theme: Theme) => ({
       padding: theme.spacing(4),
       width: 480,
     },
-    width: 300,
+    width: '100%',
+    maxWidth: 445,
```
abailly-akamai, line 51: I am not sure that's how you want to it, because your solution limits the max-width on all breakpoints?

Did you try to use the breakpoint API to make it full width only under a particular breakpoint?

coliu-akamai: ahh I missed that, thank you for catching! added a breakpoint api for maxwidth of 445 for only when viewport is below the 'sm' category

abailly-akamai: I meant: couldn't you just add:
```
 [theme.breakpoints.down('sm')]: {
     maxWidth: 100%,
  },
```
and leave the initial width rule alone? That would be cleaner and easier to understand

coliu-akamai: do you mean keep the original `width: 300` and then add ^? I just tested, and that led to the same behavior with the drawers not being full width on mobile viewports.

Or do you mean something like 
```
[theme.breakpoints.down('sm')]: {
     width: 100%,
  },
```
If so, I think that'd work in making the drawers have full width for mobile viewports, but just lead to the same as the screenshot above where the drawer contents aren't aligned. If we're good with that though, then will def change it

abailly-akamai: oops yes that's what I meant!

for the small issue in the screenshot above, you could maybe use a calc?

`width: calc(100% + XXpx)`, // or similar

coliu-akamai: yeah I can def try adding that as the width rule for that breakpoint :D 

I think it might end up making the original `width: 300` a bit unnecessary though, because there's an already existing breakpoint check that sets the drawer width to be 480 for any viewport width above the 'sm' breakpoint

abailly-akamai: make sense so maybe the width should just be 100% by default (sorry i did not check the whole stylesheet)

coliu-akamai: no worries, github didn't show it in the changes sadly ;-; And yup, I'd changed the width to 100% by default. I just added the maxWidth property with the breakpoint check in case we didn't want the issue of the screenshot in the comment happening, but am happy removing it too

bnussman-akamai (commented):
Changes look good but I feel like the code could be more readable. This isnt your fault at all but I'm finding the existing breakpoints logic to be hard to parse in my head 🧠. 

Here's my interpretation that has the same behavior, but I just find more readable. Not saying we have to do exactly this, but I'd like to hear your thoughts!
```ts
  paper: {
    '& .actionPanel': {
      display: 'flex',
      justifyContent: 'flex-end',
      marginTop: theme.spacing(),
    },
    '& .selectionCard': {
      flexBasis: '100%',
      maxWidth: '100%',
    },
    maxWidth: 480,
    padding: theme.spacing(4),
    [theme.breakpoints.down('sm')]: {
      maxWidth: 445,
      padding: theme.spacing(2),
    },
    [theme.breakpoints.down('xs')]: {
      maxWidth: '100%',
    },
  },
```

coliu-akamai:
> 

I like this a lot better too -- thank you, will try it out!! :D

coliu-akamai:
@bnussman-akamai made the changes, lmk your thoughts! Removed the check for the 'xs' breakpoint, since it looks like that breakpoint is just 0px. Had to add the width: 100% though, because it seemed like the default drawer width is way smaller than 480px

bnussman-akamai (commented):
Oops, I may have pushed you in a direction that introduced a bug 🦟

Our `<Drawer />` component has a `wide` prop, and it seems that wide drawers have been negatively impacted by this change. We want our wide drawer to remain wide. 

| Production | This PR |
|--------|-------|
|   Screenshot 2023-07-25 at 10 46 55 AM    |    Screenshot 2023-07-25 at 10 46 51 AM   |

coliu-akamai:
@bnussman-akamai made some changes that should hopefully fix the issue! I also noticed that when on a wide drawer, if the screen was 600-699px, parts of the wide drawer would be cut off, so changed some of the styling there too

bnussman-akamai:
Nice, looks fixed. Last thing: do you think we could make the wide drawer **stay** 100% width as I shrink the viewport?

https://github.com/linode/manager/assets/115251059/a5c232c2-6f7d-4b55-8156-8b9def4be798

coliu-akamai:
@bnussman-akamai how's this? :D

packages/manager/src/components/Drawer.tsx
```diff
@@ -92,7 +96,9 @@ export const Drawer = (props: Props) => {
         }
       }}
       anchor="right"
-      classes={{ paper: `${classes.paper} ${wide ? classes.wide : ''}` }}
+      classes={{
+        paper: `${classes.common} ${wide ? classes.wide : classes.default}`,
+      }}
```
bnussman-akamai, line 101: Drawer behavior looks great now! 🎉

Last bit of feedback: these classnames should be handled with a conditional classnames helper like `cx`. You can get the `cx` function from the useStyles hook like so:  `const { classes, cx } = useStyles();`

```suggestion
      classes={{
        paper: cx(classes.common, {
          [classes.default]: !wide,
          [classes.wide]: wide,
        }),
      }}
```

or 


```suggestion
      classes={{
        paper: cx(classes.common, wide ? classes.wide : classes.default),
      }}
```

bnussman-akamai (approved):

dwiley-akamai (approved):

cpathipa (approved):

PR #9428, opened 2023-07-19
chore: Improve Sentry's environment identification

bnussman-akamai:
## Description 📝

- Improves Sentry's recognition of the environment so that we have better identification of issues in the Sentry UI  🔎

## Major Changes 🔄
- Reports `dev`, `staging`, `production`, or `local` to Sentry as the environment 🎉
  - This is much more specific than what we use to have (just `production` and `development`)

## How to test 🧪
- Add Sentry DSN to your `packages/manager/.env` (ask me for value if needed)
```bash
REACT_APP_SENTRY_URL='xxx'
```
- Add `throw new Error('OMG!')` to a component so it errors and the app crashes
- Verify you see the error in Sentry under the `local` environment


abailly-akamai (approved):
nice 👍

jdamore-linode (approved):
Nice improvement!!

PR #9430, opened 2023-07-19
fix, style: [M3-6644] - Standardize styling of copy icon on OBJ bucket details page

mjac0bs:
## Description 📝
The original ask of this ticket was to fix the tooltip, which was being cut off on the left side at small screen sizes.

This ticket proposes a couple of other changes to the icon instead. The goal is to remove the single use of a custom implementation of a file copy icon and replace it with the copy icon we use throughout the rest of the app.

## Major Changes 🔄
- Swap out the custom styled file copy icon for a `CopyTooltip`component. 
   - We also get an  `aria-label` and tooltip text on hover (not just click).
- Move the placement of the tooltip to the right side of the breadcrumb, which is more consistent with how we display other copy icons. (e.g. right side of text fields for a linode's IP address and ssh access)
- UX approved the switch in icon type and placement, as well as requested the breadcrumb text size increase.

## Preview 📷

<details><summary>Videos 🎥 </summary>
The original issue:

https://github.com/linode/manager/assets/114685994/ab6ec6d5-2328-4496-a990-3b63cc9bdc05

The proposed fix:
| Before  | After   |
| ------- | ------- |
| ezgif com-video-to-gif | ezgif com-video-to-gif (1) |
</details>

## How to test 🧪
- Have OBJ enabled on an account and at least one bucket
 or
- Turn on the MSW and go to http://localhost:3000/object-storage/buckets/us-east-1/obj-bucket-0.
- In prod, adjust the browser's screen size and note that the tooltip text is cut off on the left when the copy button is clicked on smaller screens.
3. **How to verify changes?**
- Verify that the tooltip text is not getting cut off on any screen size on this branch.
- Verify that the copy tooltip is still copying the file path to the clipboard on this branch.


jaalah-akamai (approved):
Tested, this works!

coliu-akamai (approved):
looks good to me :D

Is the bottom white bar underneath the bucket name just part of the default styling / if the name is too long?
image

bnussman-akamai (changes requested):
Great changes but this seems to introduce a new issue. There are scroll bars around the bucket path

| Production | This PR |
|--------|-------|
|   Screenshot 2023-07-25 at 12 19 58 PM     |   Screenshot 2023-07-25 at 12 19 39 PM    |

mjac0bs:
@bnussman-akamai, @coliu-akamai - Good catch on this, thanks. For some reason, I wasn't seeing the overflow scroll bars on my external monitor (even when adjusting screen size), but I did see it in the browser on my laptop. Ultimately, it was caused by some changes to the padding I overlooked when I refactored the styles, causing `overflow` styling to apply when we didn't want it to. The fix is [L19 in 7530895](https://github.com/linode/manager/pull/9430/commits/7530895015bd21873b47c9c3e4ee412c1c238fa7#diff-edcf8a55b4c71b02867acec49c0db76e07a00ce6d2b81b287bfec25a498f9402R19).

Unrelated to these changes, but because it was bothering me, I also fixed/improved a text layout issue on the OBJ details page when the list of objects fails to load ([dc41911](https://github.com/linode/manager/pull/9430/commits/dc41911a2f5ba76eff08caa4401bd737994e9dc0)). (To test, you can block network requests to the request that includes something like `object-list?delimiter=%2F&prefix=` in the browser console, which will prevent a bucket's objects from loading.)

| Before |  After  | 
|---|---|
| Screenshot 2023-07-26 at 9 34 44 AM  | Screenshot 2023-07-26 at 9 33 37 AM  |

bnussman-akamai (approved):
Great fixes! 

It's crazy we had a one-off tooltip here in the first place.

PR #9432, opened 2023-07-20
fix: Fix StackScript test by using latest images available

jdamore-linode:
## Description 📝
Our StackScript test broke because one of the Images the test selects, `Ubuntu 22.10`, became deprecated and tripped up a Cypress selector.

This quickly fixes the test by updating the images used by the tests to their latest versions.

There's potential to fix this permanently by making an API request to fetch the same data that's shown in Cloud, but this happens so infrequently that today I opted to fix it quickly but plan to revisit this in the future.

## How to test 🧪
We can rely on the automated tests for this, but reviewers can use the following command to run the affected test locally:

```bash
yarn cy:run -s "cypress/e2e/core/stackscripts/create-stackscripts.spec.ts"
```

### DIFF
--- a/packages/manager/cypress/e2e/core/stackscripts/create-stackscripts.spec.ts
+++ b/packages/manager/cypress/e2e/core/stackscripts/create-stackscripts.spec.ts
@@ -11,6 +11,7 @@ import { createLinodeRequestFactory } from 'src/factories';
 import { createLinode, getLinodeDisks } from '@linode/api-v4/lib/linodes';
 import { createImage } from '@linode/api-v4/lib/images';
 import { chooseRegion } from 'support/util/regions';
+import { SimpleBackoffMethod } from 'support/util/backoff';
 
 // StackScript fixture paths.
 const stackscriptBasicPath = 'stackscripts/stackscript-basic.sh';
@@ -109,6 +110,7 @@ const createLinodeAndImage = async () => {
       label: randomLabel(),
       region: chooseRegion().id,
       root_pass: randomString(32),
+      type: 'g6-nanode-1',
     })
   );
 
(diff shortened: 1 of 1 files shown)


bnussman-akamai (approved):

mjac0bs (commented):
I tried running this locally a few times and was never able to get the second test to pass; it seems to consistently time out. In the video, the test just gets to the stackscript create page and never fills out the form. 

Screenshot 2023-07-20 at 1 16 51 PM

jdamore-linode:
@mjac0bs Yeah, that's the test timing out while creating an image to use for the test -- not really related to this fix and there isn't much I can do for that besides increase the timeout (which I'll do as soon as GitHub starts working again..).

One possibility is that your test account has internal regions enabled, and the test is using one which doesn't have Machine Images working yet. You can eliminate that possibility by forcing the test to run against a specific region like `us-east`:

```bash
CY_TEST_REGION=us-east yarn cy:run -s "cypress/e2e/core/stackscripts/create-stackscripts.spec.ts"
```

packages/manager/cypress/e2e/core/stackscripts/create-stackscripts.spec.ts
```diff
@@ -258,7 +264,7 @@ describe('Create stackscripts', () => {
    * - Confirms that private images can be selected when deploying with "Any/All" StackScripts.
    * - Confirms that a Linode can be deployed using StackScript with private image.
    */
-  it('creates a StackScript with Any/All target image', () => {
+  it.only('creates a StackScript with Any/All target image', () => {
```
bnussman-akamai, line 267: Do we still want the `only`?

bnussman-akamai (approved):

cpathipa (approved):

packages/manager/cypress/e2e/core/stackscripts/create-stackscripts.spec.ts
jdamore-linode: Shoot lol, thanks for catching this

coliu-akamai (approved):
commented out the .only on my local to see if they both passed, looks good!

carrillo-erik (approved):

carrillo-erik:
Both passing  ✅

Screenshot 2023-07-21 at 11 52 05 AM

PR #9434, opened 2023-07-20
refactor: [M3-6265] - MUI v5 Migration - `SRC > Features > Firewalls`

coliu-akamai:
## Description 📝
Migrates SRC > Features > Firewalls from MUI v4 to v5

## Major Changes 🔄
- Removed @mui/styles imports and converted most styling to styled components
- Removed React.FC
- removed default exports for functional components in favor of inline exports unless they were being lazy exported, then kept default exports
- renamed component props to be more specific, only for props that were exported
- removed index files that only had one export on them 

## How to test 🧪
- go to the /firewalls section locally and ensure all functionality works and that things look the same as prod
  - note: there seems to be a slight difference in width for the first table cell from the landing page (at least for my local), same as the discrepancy [here](https://github.com/linode/manager/pull/9403) on local compared to prod - likely isn't related to this specific PR


packages/manager/src/features/Firewalls/FirewallDetail/Devices/FirewallDeviceRow.tsx
```diff
@@ -4,9 +4,12 @@ import { Link } from 'react-router-dom';
 import { TableCell } from 'src/components/TableCell';
 import { TableRow } from 'src/components/TableRow';
 
-import ActionMenu, { Props as ActionProps } from './FirewallDeviceActionMenu';
+import {
+  FirewallDeviceActionMenu,
+  FirewallDeviceActionMenuProps as ActionProps,
```
jaalah-akamai, line 9: We can probably remove the `as ActionProps` and just use `FirewallDeviceActionMenuProps ` since we renamed `Props` to be more descriptive.

jaalah-akamai (approved):
Really nice cleanup! 🧼

coliu-akamai:
@jaalah-akamai Had a quick question regarding the FirewallRuleTable.styles.ts styled components -- I noticed that some of the components' props led to console warnings like this:

'instrument.ts:132 Warning: React does not recognize the `originalIndex` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `originalindex` instead. If you accidentally passed it from a parent component, remove it from the DOM element.'

In this case, should I change the prop names to be lowercase, or is it fine to leave as is, or should I try to find a different way to style?

jaalah-akamai:
> @jaalah-akamai Had a quick question regarding the FirewallRuleTable.styles.ts styled components -- I noticed that some of the components' props led to console warnings like this:
> 
> 'instrument.ts:132 Warning: React does not recognize the `originalIndex` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `originalindex` instead. If you accidentally passed it from a parent component, remove it from the DOM element.'
> 
> In this case, should I change the prop names to be lowercase, or is it fine to leave as is, or should I try to find a different way to style?

That means you have a styled component that's forwarding it's props to the underlying DOM element. You need import our utility `isPropValid` to omit them. See `Button.tsx` as an example of how to do that.

bnussman-akamai (approved):

packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleTable.styles.ts
```diff
@@ -0,0 +1,143 @@
...
+export const StyledUlBox = styled(Box, { label: 'StyledUlBox' })(
+  ({ theme }) => ({
+    backgroundColor: theme.bg.bgPaper,
+    borderBottom: `1px solid ${theme.borderColors.borderTable}`,
+    color: theme.textColors.tableStatic,
+    alignItems: 'center',
+    display: 'flex',
+    fontSize: '0.875rem',
+    justifyContent: 'center',
+    padding: '8px',
```
mjac0bs, line 72:
```suggestion
    padding: theme.spacing(1),
```

packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRuleTable.styles.ts
mjac0bs, line 1: Optional improvement: we could rely on [`theme.spacing`](https://mui.com/material-ui/customization/spacing/) in a couple more places (StyledUlBox, StyledFirewallTableButton) in this file to move away from hardcoded `px` values when we can.

mjac0bs (approved):
Looked good, didn't see any styling regressions in dark or light mode. Nice job with all these files.

packages/manager/src/features/Firewalls/FirewallLanding/CreateFirewallDrawer.tsx
```diff
@@ -37,162 +37,164 @@ const initialValues: CreateFirewallPayload = {
...
+              setStatus({ generalError: getErrorMap([], err).none });
+
+            setSubmitting(false);
+            handleFieldErrors(setErrors, err);
+            handleGeneralErrors(
+              mapErrorToStatus,
+              err,
+              'Error creating Firewall.'
+            );
+          });
```
cpathipa, line 107: Nice to have: We could use async/awaits 
```
try {
  await mutateAsync(payload);
  setSubmitting(false);
  onClose();
} catch (err) {
  const mapErrorToStatus = () =>
    setStatus({ generalError: getErrorMap([], err).none });

  setSubmitting(false);
  handleFieldErrors(setErrors, err);
  handleGeneralErrors(
    mapErrorToStatus,
    err,
    'Error creating Firewall.'
  );
}
```

coliu-akamai: ahh I already merged the PR in, but I can add this to my next PR!

PR #9421, opened 2023-07-18
refactor: [M3-6802] - Remove Linodes Redux store

hkhalil-akamai:
## Description 📝
Completes the RQ Linodes migration by removing the Linodes Redux store and refactoring the last few places that still used it.

## Major Changes 🔄
- Deleted Linode Redux store (`src/store/linodes`)
- Migrated `BackupDrawer` and `SelectPlanPanel` to RQ Linode hooks/HOCs
- Moved Linodes utilities to new `src/utilities/linodes.ts` file
- Slightly refactored `ConfigSelect` to display no options and error messages

## How to test 🧪
1. Ensure all actions on Linodes (i.e., create, rename, reboot, delete, migrate, etc.) continue to work as expected 
2. Ensure enabling/disabling backups (for single Linode, for all Linodes, auto-enroll) all continue to work as expected
3. Ensure `SelectPlanPanel` continues to work as expected
4. Ensure all other components in the App that fetch Linodes continue to work as expected, e.g.:
    - Setting up NodeBalancers for Linodes
    - Attaching volumes to Linodes
    - Adding Linodes to Firewalls
    - Creating Kubernetes clusters
5. Ensure unit tests and E2E test pass

PR #9435, opened 2023-07-20
chore: Only enable `useIsLargeAccount` when it's absolutely needed

bnussman-akamai:
## Description 📝

- Makes it so Cloud Manager will only try fetch Linodes when it absolutely needs to
  - After https://github.com/linode/manager/pull/9421 is merged, if you were to load the app to a non-linodes route, for the first time ever in Cloud Manager's history, Linodes will not be fetched! 🎉

## Major Changes 🔄
- Disable `useIsLargeAccount` until we absolutely need to know if the account is large or not

## Preview 📷

> Notice how there is one less fetch. After https://github.com/linode/manager/pull/9421 is also merged, there will be zero fetches to Linodes initially

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-07-20 at 5 13 09 PM | Screenshot 2023-07-20 at 5 12 43 PM |

## How to test 🧪
- Checkout this PR
- Configure Dev Tools to show API requests
- Observe that this PR does one less fetch to the Linodes endpoint when you initially load the app to a non-linode related roure

### DIFF
--- a/packages/manager/.changeset/pr-9435-tech-stories-1689887826129.md
+++ b/packages/manager/.changeset/pr-9435-tech-stories-1689887826129.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Tech Stories
+---
+
+Only enable useIsLargeAccount when it's absolutely needed ([#9435](https://github.com/linode/manager/pull/9435))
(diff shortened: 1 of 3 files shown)


dwiley-akamai (approved):
🚢

PR #9445, opened 2023-07-24
refactor: [M3-6934] MUI v5 Migration - `SRC > Features > Linodes` pt 1

coliu-akamai:
## Description 📝
Part of migrating SRC > Features > Firewalls from MUI v4 to v5 -- breaking this PR up since there will be a lot of files. This specific pr is for styling changes: running the codemod, and then converting styles to styled components / sx if possible.

There will be a later PR for fixing up imports/exports, etc later

## Major Changes 🔄
- removes any imports from MUI v4
- switch to styled components or use sx prop for styling 

## How to test 🧪
1. Go to the /linodes section locally and test to make sure that all functionality and styling works and looks as expected


packages/manager/src/features/Linodes/LinodesDetail/LinodeRebuild/RebuildFromStackScript.tsx
```diff
@@ -290,16 +266,11 @@ export const RebuildFromStackScript = (props: Props) => {
                 resetSelectedStackScript={resetStackScript}
                 selectedId={ss.id}
                 selectedUsername={ss.username}
-                updateFor={[classes, ss.id, errors]}
```
coliu-akamai, line 293: Since classes was just a Record<classes from the makeStyles, string> after the tss codemod (and just a ClassNameMap before), I think removing them from here should be ok? Seems like it was being checked in the renderGuard hoc

packages/manager/src/features/Linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
```diff
@@ -90,13 +90,25 @@ export const LinodeRow = (props: Props) => {
...
+
   return (
     <TableRow
       ariaLabel={label}
       className={classes.bodyRow}
       data-qa-linode={label}
       data-qa-loading
       key={id}
+      onMouseEnter={handleMouseEnter}
+      onMouseLeave={handleMouseLeave}
```
jaalah-akamai, line 111: Just giving some context on why I made this change: Instead of having the parent component control the state of it's child using very specific css selectors (see `LinodeRow.style.ts`), I believe the component itself should be responsible for it's own state changes. By passing down an `isHovered` prop, the child component can toggle the copy icon itself.

packages/manager/src/features/Linodes/LinodesLanding/IPAddress.tsx
```diff
@@ -133,36 +100,60 @@ export class IPAddress extends React.Component<Props & WithStyles<CSSClasses>> {
...
+          onMouseLeave: this.handleMouseLeave,
+        }
+      : undefined;
+
     return (
-      <div
-        className={`${classes.row} ${showAll && classes.multipleAddresses}`}
-        key={key}
+      <StyledRenderIPDiv
+        {...handlers}
```
jaalah-akamai, line 143: Same goes for this, see my comment [here](https://github.com/linode/manager/pull/9445/files#r1276629392). In this case, we're conditionally applying the handlers since we only need them if `showTooltipOnIpHover` is true

packages/manager/src/features/Linodes/LinodesCreate/TabbedContent/CommonTabbedContent.styles.ts
```diff
@@ -0,0 +1,14 @@
+import Grid from '@mui/material/Unstable_Grid2';
```
coliu-akamai, line 1: saw that a lot of files had the same styling for a component, so I created a separate 'common' styling file for the package -- would this be fine? or is there another approach?

packages/manager/src/features/Linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
```diff
@@ -132,25 +142,26 @@ export const LinodeRow = (props: Props) => {
             </>
           )
         ) : (
-          <div className={classes.maintenanceOuter}>
+          <div style={{ alignItems: 'center', display: 'flex' }}>
             <strong>Maintenance Scheduled</strong>
             <TooltipIcon
-              classes={{ tooltip: classes.maintenanceTooltip }}
+              sx={{ tooltip: { maxWidth: 300 } }} // ???
+              // classes={{ tooltip: classes.maintenanceTooltip }}
```
coliu-akamai, line 149: is this feasible? unsure if this is the correct way to get the same behavior

packages/manager/src/features/Linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
```diff
@@ -200,12 +209,12 @@ export const RenderFlag: React.FC<{
    * or if it has a pending mutation available. Mutations take
    * precedent over notifications
    */
-  const { classes, linodeNotifications, mutationAvailable } = props;
```
coliu-akamai, line 203: removed classes from the RenderFlag props, as it seemed it wasn't being used anyway (classes.flag was not a thing in the linode row styles anyway)

packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/LinodeNetworking.tsx
```diff
@@ -158,6 +127,8 @@ const LinodeNetworking = () => {
   };
 
   const renderIPRow = (ipDisplay: IPDisplay) => {
+    // TODO: in order to fully get rid of makeStyles for this file, may need to convert this to a functional component
```
coliu-akamai, line 130: Tried converting to styled componets/sx for components with classes.row and classes.copy, but this led to some styling behavior at first being lost. Possible solution to fully get rid of makestyles would likely be very similar to the stuff done in IPAddress.tsx/LinodeRow.tsx

packages/manager/src/features/Linodes/LinodesDetail/LinodeRebuild/RebuildFromStackScript.tsx
jaalah-akamai: Yes that's fine

packages/manager/src/features/Linodes/LinodesCreate/TabbedContent/CommonTabbedContent.styles.ts
jaalah-akamai: @coliu-akamai Yes! We need to do this

bnussman-akamai (approved):
Styles look good to me! 🎨

PR #9448, opened 2023-07-25
fix: Enable Sentry Tracing explicitly

bnussman-akamai:
## Description 📝

- Follow up to https://github.com/linode/manager/pull/9337 and https://github.com/linode/manager/pull/9428
- I just forgot to commit a `tracesSampleRate` which resulted in a feature being disabled when I expected it to be enabled

## How to test 🧪
- Test that we get basic performance metrics in Sentry once this is merged into `staging`

### DIFF
--- a/packages/manager/src/initSentry.ts
+++ b/packages/manager/src/initSentry.ts
@@ -12,6 +12,8 @@ import redactAccessToken from 'src/utilities/redactAccessToken';
 import packageJson from '../package.json';
 
 export const initSentry = () => {
+  const environment = getSentryEnvironment();
+
   if (SENTRY_URL) {
     init({
       allowUrls: [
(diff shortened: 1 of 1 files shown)


packages/manager/src/initSentry.ts
```diff
@@ -79,6 +82,7 @@ export const initSentry = () => {
       ],
       integrations: [new BrowserTracing()],
       release: packageJson.version,
+      tracesSampleRate: environment === 'production' ? 0.2 : 1,
```
bnussman-akamai, line 85: In production, we will only capture performance data from 20% of sessions. This is recommended practice and we can adjust as needed. 

https://docs.sentry.io/platforms/javascript/guides/react/performance

hana-linode (approved):

dwiley-akamai (approved):

cpathipa (approved):

PR #9451, opened 2023-07-26
refactor: [M3-6693] - Optimize Linode event handlers and query invalidation

bnussman-akamai:
## Description 📝

- By design, we made React Query over-invalidate Linode data to be safe as we transitioned from Redux to React Query. This caused Cloud Manager to **aggressively** make API calls, specifically when a event was in progress (for example, `linode_create`). 
- This PR's goal is
  - to invalidate **only** necessary data when Linode events are coming from the events polling
  - to optimize the `onSuccess` functions that run on Linode mutations

> **Note**
> The way I chose to write this code is very verbose and repetitive. I chose to write it this way because I think it is most clear and readable because it helps you see what exact invalidations happen after a respective event or request. This code is likely to change when we introduce a React Query "Factory" pattern into our workflow.

### Before (85 API requests during reboot) 👎🏼
https://github.com/linode/manager/assets/115251059/a7621dc3-dbbc-42ee-81ef-f2295dd02f6e

### After (only 33 API requests during reboot) 🎉
https://github.com/linode/manager/assets/115251059/619fe4a0-7900-4c7c-95c0-cf1b8d3be623

## How to test 🧪
- Test **all** Linode actions
  - Specifically test actions that take time to complete, like boot, reboot, create, delete, clone, migrate, resize, rescue, etc..


packages/manager/src/queries/linodes/events.ts
```diff
@@ -1,21 +1,145 @@
...
+ * For example, "linode_create", "linode_boot", "linode_resize", ...
+ */
 export const linodeEventsHandler = ({ event, queryClient }: EventWithStore) => {
-  // For now, invalidate any linode query. We can fine tune later.
-  queryClient.invalidateQueries([queryKey]);
+  const linodeId = event.entity?.id;
+
+  // Early return to cut down the number of invalidations.
+  // We early return if the event is "started" or "scheduled" beacuse we don't need to
+  // invaldate things when they are in progress and their percentage complete is activly updating.
```
jaalah-akamai, line 21:
```suggestion
  // invalidate things when they are in progress and their percentage complete is actively updating.
```

packages/manager/src/queries/linodes/events.ts
```diff
@@ -1,21 +1,145 @@
...
+  // If we didn't early return, the invalidations would happen every time an event's percentage updates.
+  if (!linodeId || ['scheduled', 'started'].includes(event.status)) {
+    return;
+  }
+
+  // Some Linode event are in indication that the reponse from /v4/account/notifications
```
jaalah-akamai, line 27:
```suggestion
  // Some Linode event are an indication that the response from /v4/account/notifications
```

jaalah-akamai (approved):

packages/manager/src/queries/linodes/linodes.ts
```diff
@@ -186,7 +196,10 @@ export const useBootLinodeMutation = (id: number) => {
     ({ config_id }) => linodeBoot(id, config_id),
     {
       onSuccess() {
-        queryClient.invalidateQueries([queryKey]);
+        queryClient.invalidateQueries([queryKey, 'paginated']);
+        queryClient.invalidateQueries([queryKey, 'all']);
+        queryClient.invalidateQueries([queryKey, 'infinite']);
+        queryClient.invalidateQueries([queryKey, 'linode', id, 'details']);
```
cpathipa, line 202: Since we are doing queryClient.invalidateQueries quite a few places. Any thoughts in having a util function that takes array of arrays and perform invalidation?

example:
```async function invalidateQueries(queryKeysArray) {
  try {
    await Promise.all(queryKeysArray.map(key => queryClient.invalidateQueries(key)));
    console.log('Queries invalidated successfully.');
  } catch (error) {
    console.error('Error while invalidating queries:', error);
  }
}

const queryKeysArray = [
  [queryKey, 'test', 'test1'],
  [queryKey, 'test01'],
  [queryKey, 'test02'],
  // Add more query keys as needed.
];


```

packages/manager/src/queries/linodes/linodes.ts
```diff
@@ -198,7 +211,10 @@ export const useRebootLinodeMutation = (id: number) => {
     ({ config_id }) => linodeReboot(id, config_id),
     {
       onSuccess() {
-        queryClient.invalidateQueries([queryKey]);
+        queryClient.invalidateQueries([queryKey, 'paginated']);
+        queryClient.invalidateQueries([queryKey, 'all']);
+        queryClient.invalidateQueries([queryKey, 'infinite']);
+        queryClient.invalidateQueries([queryKey, 'linode', id, 'details']);
```
cpathipa, line 217: If we go with above route we could lift some redundant code. or we could wrap it in a function  and reuse it.

packages/manager/src/queries/linodes/linodes.ts
```diff
@@ -208,7 +224,10 @@ export const useShutdownLinodeMutation = (id: number) => {
   const queryClient = useQueryClient();
   return useMutation<{}, APIError[]>(() => linodeShutdown(id), {
     onSuccess() {
-      queryClient.invalidateQueries([queryKey]);
+      queryClient.invalidateQueries([queryKey, 'paginated']);
+      queryClient.invalidateQueries([queryKey, 'all']);
+      queryClient.invalidateQueries([queryKey, 'infinite']);
+      queryClient.invalidateQueries([queryKey, 'linode', id, 'details']);
```
cpathipa, line 230: Same here.

packages/manager/src/queries/linodes/linodes.ts
```diff
@@ -218,22 +237,16 @@ export const useLinodeChangePasswordMutation = (id: number) =>
...
   const queryClient = useQueryClient();
   return useMutation<{}, APIError[], { region: string } | undefined>(
     (data) => scheduleOrQueueMigration(id, data),
     {
       onSuccess() {
-        queryClient.invalidateQueries([queryKey]);
+        queryClient.invalidateQueries([queryKey, 'paginated']);
+        queryClient.invalidateQueries([queryKey, 'all']);
+        queryClient.invalidateQueries([queryKey, 'infinite']);
+        queryClient.invalidateQueries([queryKey, 'linode', id, 'details']);
```
cpathipa, line 249: Same here.

cpathipa (approved):
LGTM, left comments for minor feedback, but it is totally optional.

PR #9452, opened 2023-07-26
refactor: [M3 6730] - LinodeSelect - Allow fully custom noOptionsText

jaalah-akamai:
## Description 📝
We want to allow for a fully custom `noOptionsText` without italics

## Major Changes 🔄
- Added two additional storybook stories for custom no options text and multi-select
- Allow for a fully custom `noOptionsText` without italics

The mult-select is just a visual with the `actions` tab used to show data received by event handler. Don't expect to uncheck/check additional options.

| Screenshots   | |
| ------- | ------- |
| Screenshot 2023-07-25 at 11 21 16 PM | Screenshot 2023-07-25 at 11 21 21 PM |

## How to test 🧪
1. Go to storybook page http://localhost:6006/?path=/story/components-linode-select--default
2. Observe you can customize the `noOptionsMessage` prop.

### DIFF
--- a/packages/manager/.changeset/pr-9452-tech-stories-1690342391207.md
+++ b/packages/manager/.changeset/pr-9452-tech-stories-1690342391207.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Tech Stories
+---
+
+LinodeSelect - Allow fully custom noOptionsText ([#9452](https://github.com/linode/manager/pull/9452))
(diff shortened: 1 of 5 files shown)


jdamore-linode (approved):
The two tests that failed in CI passed locally for me

bnussman-akamai:
@jdamore-linode @jaalah-akamai 

I think this might be one of those situations where the form tries to submit before the config is selected...

This endpoint will actually allow users to POST without a `config_id`, in which case, it will use the last booted config. But here we have some client side validation that is making the config_id required. 

Screenshot 2023-07-26 at 4 09 37 PM

cpathipa (approved):

bnussman-akamai (approved):

PR #9454, opened 2023-07-26
refactor: [M3-6943] - MUI v5 Migration - `SRC > Components > InputAdornment`

bnussman-akamai:
## Description 📝

- Adjusts `InputAdornment` to match the import/export structure of our other components

## Major Changes 🔄
- Use named exports 
- Document in Storybook 📝
- Move out of the `core` folder 🗂️

## Preview 📷
Screenshot 2023-07-26 at 6 54 58 PM

## How to test 🧪
- Verify that input adornments are unaffected by this change
  -  One place you can see them is in the "Make a Payment Drawer" (http://localhost:3000/account/billing/make-payment)


packages/manager/src/components/InputAdornment.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { InputAdornment } from './InputAdornment';
+
+const meta: Meta<typeof InputAdornment> = {
+  component: InputAdornment,
+  title: 'Components/TextField/InputAdornment',
+};
+
+type Story = StoryObj<typeof InputAdornment>;
+
+export const Default: Story = {
```
hana-linode, line 13: Do we need this standalone story when it's kind of already covered in the TextField story?

hana-linode (approved):

packages/manager/src/components/InputAdornment.stories.tsx
bnussman-akamai: I'm not sure, but we do get the benefit of the props being documented

mjac0bs: I thought about this too... I support keeping it for props' sake.

mjac0bs (approved):

PR #9456, opened 2023-07-27
fix: [M3-6898] - Fix markup defect in Help & Support landing search

carrillo-erik:
## Description 📝
Fixes the UI bug which was causing markup to appear in the `AlgoliaSearchBar` in the `/Support` landing page. Also makes it possible for text to be cleared via `isClearable` prop.

## Major Changes 🔄
- Change `isClearable={false}` to `isClearable={true}`

## Preview 📷
| Before  | After   |
| ------- | ------- |
|  before | after |

## How to test 🧪
1. Navigate to `/Support` landing page and type in a search term.
2. Use the mouse to click on one of the results, this will open a new tab.
3. Close the new tab.
4. Inspect the text inside the `AlgoliaSearchBar` and verify that the markup defect has been resolved.
5. Look for the 'X' icon and verify that it works as expected.

### DIFF
--- a/packages/manager/.changeset/pr-9456-fixed-1690469715549.md
+++ b/packages/manager/.changeset/pr-9456-fixed-1690469715549.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Fixed
+---
+
+Remove markup from search results. ([#9456](https://github.com/linode/manager/pull/9456))


cpathipa (approved):

mjac0bs (approved):
🚢 🔍

jaalah-akamai (approved):

PR #9457, opened 2023-07-27
refactor: Clean up exports of `src/hooks/`

bnussman-akamai:
## Description 📝
- Some hooks have **both** default and named exports. While this does not cause any problems, it is much cleaner to only export a hook once. By having less exports in general, `tsc` times may improve and by using named exports, our code is more traceable. 

## Major Changes 🔄
- Deletes the default export for many hooks in favor of using the named export
- Deletes unused `LinodeConfigSelectionDrawer.tsx` code

### Before 🚫
Screenshot 2023-07-27 at 11 26 42 AM

### After 🧼
Screenshot 2023-07-27 at 11 26 21 AM

## How to test 🧪
- These changes are relatively safe so...
  - Verify automated checks pass
  - Do a quick test that Cloud Manager generally works


jaalah-akamai (approved):

coliu-akamai (approved):
looks good!

dwiley-akamai (approved):

PR #9396, opened 2023-07-13
refactor: [M3-6875, M3-6719] - Upgrade remaining `LinodeSelect`s

hkhalil-akamai:
## Description 📝
Finishes migrating instances of `LinodeSelect` over to the V2 component introduced in #9193. Also fixes a bug causing the input field of `LinodeSelectV2` to reset while the user is typing.

## Major Changes 🔄
- Remove old `LinodeSelect` with associated files
- Allow for IP address to replace option label using `showIPAddressLabel`
- Filter linodes by region and options filter
- Move `✓` to the right and remove `x` for selected options. This is because there's always multiple intuitive ways to remove a selected option, and to allow for flags on the left.

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-07-19 at 10 37 44 AM | Screenshot 2023-07-19 at 10 37 22 AM |
| Screenshot 2023-07-19 at 10 38 59 AM | Screenshot 2023-07-19 at 10 47 40 AM |
| Screenshot 2023-07-19 at 10 47 59 AM | Screenshot 2023-07-19 at 10 47 48 AM |
| Screenshot 2023-07-19 at 10 58 41 AM | Screenshot 2023-07-19 at 10 58 27 AM |

## How to test 🧪
1. Checkout branch or view preview link
2. Check out areas we use LinodeSelectV2
    - http://localhost:3000/firewalls (create)
    - http://localhost:3000/volumes/create
    - http://localhost:3000/nodebalancers/create
4. Ensure that when a value/option is selected, you can still view other options within region
5. Ensure when a region has changed, linode select has been cleared

PR #9464, opened 2023-07-28
fix: Fix `<LinodeSelect />` bugs

bnussman-akamai:
## Description 📝
- This PR fixes some bugs introduced by the new <LinodeSelect /> component
  - https://github.com/linode/manager/pull/9396
  - https://github.com/linode/manager/pull/9193
- Improves NodeBalancer config node form experience

### 1️⃣ Use `useAllLinodesQuery` insted of `useLinodesInfiniteQuery`

- We must do this because of the following scenario... Imagine you have a `<LinodeSelect />` with the `value` prop set to the ID of a Linode that isn't returned in the first 25 results of the infinite query. This will result in the Select unable to show the label of the selected Linode because the infinite query does not yet have that Linode loaded into the cache.
- This bug would apply to the `<LinodeSelect />` used in the NodeBalancer Config Node section because the `value` would be defined when the page loads and the selected Linode might not be one of the first 25 Linodes loaded by the infinite query. 
- This problem could probably be solved in a few ways but using `useAllLinodesQuery` is probably our best option.

### 2️⃣ NodeBalancer Config Node not showing selected value

- When going to edit a NodeBalancer at `http://localhost:3000/nodebalancers/{id}/configurations`, you should see the Linode private IP that was configured when the NodeBalancer was configured. Currently, this is broken and it does not show you that a Linode is selected.

| Before | After |
|--------|-------|
|    Screenshot 2023-07-28 at 11 54 13 AM    |    Screenshot 2023-07-28 at 11 53 56 AM   |

### 3️⃣ NodeBalancer Config Node IP clearable not working

- We show the clearable button but it would not actually clear the underlying value. This is now fixed. 

| Before | After |
|--------|-------|
|   <video src="https://github.com/linode/manager/assets/115251059/6f3da51f-a4ce-49ca-aac9-2ec072cb7443" />   |     <video src="https://github.com/linode/manager/assets/115251059/89a6c58f-bbab-44e7-83ea-4a1df5a7e42d" />  |

## How to test 🧪
- Test that the bugs shown above are actually addressed by this PR without any adverse effects
- Verify e2e tests pass because they do provide some coverage of the <LinodeSelect /> component


jaalah-akamai (approved):
Nice changes - thanks for catching this

jdamore-linode (approved):

PR #9465, opened 2023-07-28
feat: [M3-6867] - Betas Landing

CodyFinn-Akamai:
## Description 📝
Adds the basic layout for the betas landing page and utility functions for working with `Beta` and AccountBeta objects.

## Major Changes 🔄
- Creates `features/betas/index.tsx`
- Adds utility functions for working with `Beta` and `AccountBeta` objects

## Preview 📷
image

## How to test 🧪
run `yarn test BetasLanding betaUtils`
Ensure that when you visit `/betas` the betas landing page is rendered


packages/manager/src/MainContent.tsx
```diff
@@ -355,6 +356,7 @@ const MainContent = (props: CombinedProps) => {
                             {showDatabases ? (
                               <Route component={Databases} path="/databases" />
                             ) : null}
+                            <Route component={Betas} path="/betas" />
```
mjac0bs, line 359: I haven't familiarized myself too much with this feature yet, but are we planning on leaving it un-feature-flagged?

jdamore-linode (dismissed):
@CodyFinn-Akamai Just pushed a couple integration tests to confirm that the Betas landing page is accessible (and present in the nav) when the feature is enabled, and is not accessible when the feature's disabled.

You can run the tests with this command:
```bash
yarn cy:run -s "cypress/e2e/core/account/betas.spec.ts"
```

packages/manager/src/featureFlags.ts
```diff
@@ -41,6 +41,7 @@ type OneClickApp = Record<string, string>;
 export interface Flags {
   aglb: boolean;
   apiMaintenance: APIMaintenance;
+  betas: boolean;
```
mjac0bs, line 44: Could we name this flag `selfServeBetas` to avoid potential confusion with any other betas? I'm also not seeing `betas` in LD -- does that still need to be configured?

mjac0bs (commented):
Tests and code of landing page looks good. Left a comment about the feature flag. Can we also add a changeset to this PR? (`yarn changeset`) Pulling in the latest `develop` will give you the ability to add our newest changeset type, 'Upcoming Features', which is what this would be best categorized as.

packages/manager/src/featureFlags.ts
CodyFinn-Akamai: Yeah I didn't add it to LD yet

mjac0bs: I think Banks did, you should be good now!

mjac0bs (dismissed):

bnussman-akamai (approved):

mjac0bs (approved):
Tests pass in CI and locally so I think we're good here. (Though I had to run e2e in debug mode and mouse over the side nav to avoid a timeout error when it couldn't find "Betas" because`<a.css-gl93uk-listItem> is not visible because it has CSS property: position: fixed and it's being covered by another element` -- am I recalling correctly that this is a known thing to occur, @jdamore-linode?)

bnussman-akamai (approved):

jdamore-linode:
> Tests pass in CI and locally so I think we're good here. (Though I had to run e2e in debug mode and mouse over the side nav to avoid a timeout error when it couldn't find "Betas" because`<a.css-gl93uk-listItem> is not visible because it has CSS property: position: fixed and it's being covered by another element` -- am I recalling correctly that this is a known thing to occur, @jdamore-linode?)

@mjac0bs Typically when we see this, it's because there's a banner that's been set up that's overlapping some element at the top of the page (and the solution is to dismiss the banner using the affected account from within Cypress).

This one's unusual because it's in a totally different place on the screen and probably has nothing to do with a banner. By any chance, were you running the test using `yarn up` or `yarn dev`?

I'm wondering if it's the dev tool icon just slightly overlapping the Betas nav item causing the issue:
 
Screenshot 2023-08-04 at 9 30 37 AM

CodyFinn-Akamai:
> > Tests pass in CI and locally so I think we're good here. (Though I had to run e2e in debug mode and mouse over the side nav to avoid a timeout error when it couldn't find "Betas" because`<a.css-gl93uk-listItem> is not visible because it has CSS property: position: fixed and it's being covered by another element` -- am I recalling correctly that this is a known thing to occur, @jdamore-linode?)
> 
> @mjac0bs Typically when we see this, it's because there's a banner that's been set up that's overlapping some element at the top of the page (and the solution is to dismiss the banner using the affected account from within Cypress).
> 
> This one's unusual because it's in a totally different place on the screen and probably has nothing to do with a banner. By any chance, were you running the test using `yarn up` or `yarn dev`?
> 
> I'm wondering if it's the dev tool icon just slightly overlapping the Betas nav item causing the issue:
> Screenshot 2023-08-04 at 9 30 37 AM

@jdamore-linode am I good to merge this then?

mjac0bs:
> were you running the test using `yarn up` or `yarn dev`

@jdamore-linode Yes, I was - good thought - though I ran again with `yarn start:manager:ci` to get rid of devtools and still saw the same error. Also seeing local failures on that first test for `load-balancer-navigation.spec.ts` and `vpc-navigation.spec.ts` -- it looks like the nav remains in collapsed state and doesn't expand so that the text is findable. Not sure what would be causing this locally, and it seems like it might just be something on my end if you or Cody aren't seeing it.

jdamore-linode:
> > were you running the test using `yarn up` or `yarn dev`
> 
> @jdamore-linode Yes, I was - good thought - though I ran again with `yarn start:manager:ci` to get rid of devtools and still saw the same error. Also seeing local failures on that first test for `load-balancer-navigation.spec.ts` and `vpc-navigation.spec.ts` -- it looks like the nav remains in collapsed state and doesn't expand so that the text is findable. Not sure what would be causing this locally, and it seems like it might just be something on my end if you or Cody aren't seeing it.

Ah yeah! I think I've seen that before -- I believe if you run `cy:debug` and then toggle the nav visibility from there, it will persist and the tests will start passing. They maybe shouldn't be this way, but the tests assume that the nav is in its expanded state

@CodyFinn-Akamai I think this is good to merge -- there might be room for improvement here, but no sense in holding up the merge since this should pass reliably in CI.

packages/manager/src/utilities/betaUtils.test.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  it('should return false if the beta does not have an enrolled field defined', () => {
+    const { activeBeta } = betas;
+    expect(wasCustomerEnrolled(activeBeta)).toBe(false);
+  });
+
+  it('should return false if the beta does not have an enrolled field defined', () => {
+    const { activeBeta } = betas;
+    expect(wasCustomerEnrolled(activeBeta)).toBe(false);
+  });
```
mjac0bs, line 137: Looks like a duplicate of the test case above. Not sure if you meant `activeAccountBeta`, something else, or if this just slipped in by accident?

mjac0bs (approved):
I pushed an update to the import statement that was causing typecheck and unit test to fail; both look good now, just a couple small bits of feedback on the betaUtils unit test. After that's addressed, 🚢 . (The unrelated e2e failure on domains is expected.)

bnussman-akamai (approved):

PR #9469, opened 2023-07-28
fix: Redirect the user back to `/linodes` when Linode is deleted from Linode Details

bnussman-akamai:
## Description 📝

- During the React Query for Linodes refactor, we missed a somewhat important redirect

## Major Changes 🔄
- Send the user back to `/linodes` when they delete a Linode from the Linodes Details page

### Before 😖

https://github.com/linode/manager/assets/115251059/dc1e670e-8b12-4950-b545-fc7a94027e4c

### After 🎉

https://github.com/linode/manager/assets/115251059/1494b5ff-7651-44f3-b1f4-d5e0521de62d

## How to test 🧪
- Delete a Linode from the Linodes Landing page and verify nothing is changed
- Delete a Linode from the Linodes Details page and verify that you get sent to `/linodes`

### DIFF
--- a/packages/manager/src/features/Linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
+++ b/packages/manager/src/features/Linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
@@ -238,6 +238,9 @@ const LinodeDetailHeader = () => {
         onClose={closeDialogs}
       />
       <DeleteLinodeDialog
+        onSuccess={() => {
+          history.push('/linodes');
+        }}
         linodeId={matchedLinodeId}
         onClose={closeDialogs}
         open={deleteDialogOpen}


packages/manager/src/features/Linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
```diff
@@ -240,6 +240,7 @@ const LinodeDetailHeader = () => {
       <DeleteLinodeDialog
         linodeId={matchedLinodeId}
         onClose={closeDialogs}
+        onSuccess={() => history.push('/linodes')}
```
bnussman-akamai, line 243: Is `history.push('/linodes')` what we want here? Or do we want `history.replace('/linodes')`?

dwiley-akamai: I would probably prefer `history.replace('/linodes')` since the previous `linodes/:linodeId` page would not be a useful thing for a user to navigate back to at that point.

dwiley-akamai (approved):

PR #9472, opened 2023-07-31
feat: Allow NodeBalancer creation without config

zliang-akamai:
## Description 📝
A configuration is not strictly required when invoking [the NodeBalancer creation API](https://www.linode.com/docs/api/nodebalancers/#nodebalancer-create), and only the region is required. I think it would be nicer if we can do the same thing in cloud manager. I think sometimes an user may want to create a nodebalancer and config it later, and this might be able to make their work easier.
 
```bash
curl -H "Content-Type: application/json" \
    -H "Authorization: Bearer $TOKEN" \
    -X POST -d '{"region": "us-central"}' https://api.linode.com/v4/nodebalancers 
```

I am not familiar with manager development, so please feel free to close the PR this change is not desired, or commit to this branch if there is anything else you guys would like to add.

## Major Changes 🔄
- Allow NodeBalancer creation without config

## Preview 📷

> **Note**: Use `<video src="" />` tag when including recordings in table

| Before  | After   |
| ------- | ------- |
| image | image |

## How to test 🧪
Not sure

### DIFF
--- a/packages/manager/src/features/NodeBalancers/NodeBalancerConfigPanel.tsx
+++ b/packages/manager/src/features/NodeBalancers/NodeBalancerConfigPanel.tsx
@@ -408,44 +408,37 @@ export const NodeBalancerConfigPanel = (
           </Grid>
         </Grid>
       </Grid>
-
-      {(forEdit || configIdx !== 0) && (
-        <React.Fragment>
-          <Grid xs={12}>
-            <Divider />
-          </Grid>
-          <Grid
-            alignItems="center"
-            container
-            justifyContent="flex-end"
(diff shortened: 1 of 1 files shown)


bnussman-akamai (approved):

zliang-akamai:
@bnussman I am not sure why the CI is failing... Should I fix it or it's something unrelated?

coliu-akamai (approved):
Looks good to me!

zliang-akamai:
@bnussman-akamai Thanks for generating the change set for this PR, sorry about didn't realize it was needed 😅

PR #9481, opened 2023-08-02
refactor: Replace VolumesActionsPanel with ActionsPanel

cpathipa:
## Description 📝
**This replaces VolumesActionsPanel with ActionsPanel since it's a variation of the ActionsPanel component.**

## Major Changes 🔄
- Change #1
- Change #2

## How to test
- Verify Volumes flow - create, update, clone and resize


carrillo-erik:
All of the `Volumes` flows works as expected and didn't notice any UI regressions.
My only reservation is that it appears that we duplicate a bit of code across the codebase.
The only `props` that I noticed that differ in each case are `label` and `disabled`.

jaalah-akamai (approved):

cpathipa:
Since it has some redundancy, Closing this PR for now.

PR #9485, opened 2023-08-02
feat: [M3-6729] - Add VPC column to Linodes landing table

coliu-akamai:
## Description 📝
Added vpc column to linode landing table so that users can see what vpc a linode is associated with 

## Major Changes 🔄
- add vpc column to linode landing table
- changed IP Address column to public iP address

- [ ] TODO: figure out sortable table header for vpc column ;-;

## Preview 📷
image

## How to test 🧪
1. Turn on mocks and navigate to linodes landing page to see the vpc column (you can see the vpc column without the mock as long as you have the vpc flag enabled, but you won't be able to see any mock data)


packages/manager/src/features/Linodes/LinodesLanding/SortableTableHead.tsx
```diff
@@ -135,6 +137,19 @@ const SortableTableHead = <T extends unknown>(props: CombinedProps<T>) => {
...
+                  <TableSortCell
+                    // TODO: VPC - figure out the orderBy (cannot use label, as that will order via linode label)
+                    active={isActive('vpc')}
+                    direction={order}
+                    handleClick={handleOrderChange}
+                    label="vpc"
+                  >
+                    VPC
+                  </TableSortCell>
+                </Hidden>
```
coliu-akamai, line 151: had several questions here actually, some of them as comments in the code ^
- Looking more into how to sort the VPC column, but at initial glance I don't think vpc is an option in the api (so the column currently doesn't sort), but using label would result in sorting by the linode's label. 
- is lgDown good for the Hidden, or do we want some other breakpoint/value? (smDown, etc)
- just to double check, if isVlan is true, we'd want to hide this column? (isVlan is currently being used as a check to hide columns like region, plan, public ip address)

packages/manager/src/features/Linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
```diff
@@ -199,6 +225,20 @@ export const LinodeRow = (props: Props) => {
   );
 };
 
+const findAssociatedVPC = (vpcs: VPC[], linodeId: number) => {
+  for (const vpc of vpcs) {
+    for (const subnet of vpc.subnets) {
+      if (subnet.linodes.includes(linodeId)) {
+        // a linode should have only one vpc associated with it (?) so we can
+        // short circuit once we find the associated vpc
```
coliu-akamai, line 233: Was seeing in the mock examples that linodes would often have multiple vpcs associated with them, but just wanted to confirm that in the actual live data, a linode will have at most one vpc associated with it (although this function may eventually be unnecessary if the api endpoints get updated)

packages/manager/src/features/Linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
```diff
@@ -52,6 +56,17 @@ export const LinodeRow = (props: Props) => {
...
+  // TODO: VPC - currently the only way to know what VPC a linode is associated with is to get all
+  // vpcs, and then loop through each vpc's subnets, and then through each of the subnets' linode IDs
+  // to see if that subnet contains this linode id. There is some discussion about an endpoint
+  // that will directly get vpcs associated with a linode ID, and if that happens, we can
+  // replace this query
+  // Can also put react query this in Linodes/index instead, so that getting all vpcs is only done once
+  // (but if react query caches the data from this call it should be ok either way?)
+  const { data: vpcs } = useVPCsQuery({}, {});
+  // query is paginated -- does not passing in any params mean that it will get all vpcs?
+  const vpc = findAssociatedVPC(vpcs?.data ?? [], id);
```
coliu-akamai, line 68: Had some questions + general thoughts here too, including about pagination

bnussman-akamai:
@jaalah-akamai Should we build this right now given that the API might add `/v4/linode/instances/:id/vpcs` for us to use for this purpose?

jaalah-akamai:
I should probably clarify - API will not be adding a new endpoint at this time, but they will correct the existing endpoint `/linode/instances/{linode id}/configs/{config id}/interfaces` to provide the VPC id which is needed to more easily determine the VPC label for that column

jaalah-akamai:
> @jaalah-akamai Should we build this right now given that the API might add `/v4/linode/instances/:id/vpcs` for us to use for this purpose?

I will also add this fix is pending API approval, so I'm for moving forward with the approach we know will work (even though it's a roundabout way of getting the VPC label).

bnussman-akamai:
So for every Linode row we'll have to do a `GET /v4/linode/instances/{linode id}/configs` and a `GET /v4/linode/instances/{linode id}/configs/{config id}/interfaces`? That isn't good.

coliu-akamai:
I think the API team's plan is to add the VPC id to the interface config return object, so just using `GET /v4/linode/instances/{linode id}/configs` should do the trick (and then I'd need to loop through every config that is returned to be able to loop through all their interfaces, and then just get the first interface with a purpose of `vpc`. From there I'd have to do a `GET /vpcs/<vpc_id>` so that I can also get the vpc's label.

tbh, I might still need to put this functionality into the [Linodes/index.tsx](https://github.com/linode/manager/pull/9485/files/8356a4dfc66d9511dc725f61fb48b5d9b1a4f206#diff-11fe479742b38927740f341d26d2f27a643d2c1b12bb8e94a0f158eccf2bd752) file in order to be able to have the table sort by the vpc label -- earlier, when I'd had a solution of getting the vpc id and label inside the [Linodes/LinodesLanding/LinodeRow/LinodeRow.tsx](https://github.com/linode/manager/pull/9485/files/8356a4dfc66d9511dc725f61fb48b5d9b1a4f206#diff-110dd10094fd4d3ebbb67f63f34ef939fedf03a722fe6dd3dbb60cd03f56db01) file only, the sorting didn't work. I can look into this more once the new api spec comes out on Tuesday though! 👀

bnussman-akamai:
Okay, understood @coliu-akamai.

I do want to bring some attention to making vpc label sortable on the client-side. I don't think we should let people order by vpc label. Practices like this have been blocking us from API paginating and sorting Linodes. I think the only way we should allow this is if the /linode/instances endpoint returns a `vpc_label` , which would mean the API can sort by it on the server-side.  @jaalah-akamai

dwiley-akamai (commented):
Left a couple of comments -- I would agree with not allowing for sorting by the VPC label in the Linodes landing table unless the API supports it.

coliu-akamai:
updated to reflect the new api changes -- have also made the vpc column not sortable

packages/api-v4/.changeset/pr-9485-upcoming-features-1691764196950.md
```diff
@@ -0,0 +1,5 @@
+---
+"@linode/api-v4": Upcoming Features
```
dwiley-akamai, line 2: I'd probably go with `Changed` for the category here

packages/manager/src/features/Linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
```diff
@@ -247,6 +254,21 @@ export const RenderFlag: React.FC<{
   return null;
 };
 
+const getVPCId = (configs: Config[]): number | undefined => {
+  for (const config of configs) {
+    for (const linodeInterface of config.interfaces) {
+      if (
+        linodeInterface.purpose === 'vpc' &&
+        (linodeInterface.vpc_id || linodeInterface.vpc_id === 0)
```
bnussman-akamai, line 259: Do we need this check `(linodeInterface.vpc_id || linodeInterface.vpc_id === 0)`? What does it do?

packages/manager/src/features/Linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
```diff
@@ -52,14 +55,18 @@ export const LinodeRow = (props: Props) => {
...
   } = props;
 
   const notificationContext = React.useContext(_notificationContext);
 
   const { data: notifications } = useNotificationsQuery();
 
+  // TODO: VPC - later if there is a way to directly get a linode's vpc, replace this
+  const { data: configs } = useAllLinodeConfigsQuery(id);
+  const vpcId = getVPCId(configs ?? []);
+  const { data: vpc } = useVPCQuery(vpcId ?? -1);
```
bnussman-akamai, line 66: We should only "enable" this query if `vpcId` is defined. This will require you too add an `enabled` parameter and pass it to the useQuery for `useVPCQuery` 

```suggestion
  const { data: vpc } = useVPCQuery(vpcId ?? -1, vpcId !== undefined);
```

This will make sure we don't fire-off any VPC API calls until we have determined if the Linode has a VPC

packages/manager/src/features/Linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
coliu-akamai: We don't really need the check for `linodeInterface.vpc_id` since if the purpose is `vpc` the `vpc_id` will definitely be present, otherwise the `vpc_id` will be null (or undefined as I marked it as optional in the api package). I think I just included it as an extra check that the vpc_id is actually a value -- I can take it out and then update the suggestion you had below to be  ` const { data: vpc } = useVPCQuery(vpcId ?? -1, vpcId !== undefined && vpcId !== null)`

Just wanted to double check, will there be any vpc's that have an id of 0? I was noticing that the mock service worker sometimes generated vpc's with id 0 and then they wouldn't show up because of 0 being a falsy value, so I included the check here so that the vpc's would appear (same with line 180 in this file)

bnussman-akamai: Code looks good now!

packages/manager/src/features/Linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
```diff
@@ -163,6 +177,17 @@ export const LinodeRow = (props: Props) => {
             <RegionIndicator region={region} />
           </TableCell>
         </Hidden>
+        {flags.vpc && (
+          <Hidden smDown>
+            <TableCell noWrap>
+              {vpcLabel && (vpcId === 0 || vpcId) && (
```
bnussman-akamai, line 183: We can simplify this!

```suggestion
              {vpcLabel && (
```

packages/manager/src/features/Linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
```diff
@@ -163,6 +177,17 @@ export const LinodeRow = (props: Props) => {
...
+                <Link tabIndex={0} to={`/vpc/${vpcId}`}>
+                  {vpcLabel}
```
bnussman-akamai, line 185: Should we have some kind of state for when there is no VPC attached? 

Screenshot 2023-08-14 at 1 45 44 PM
Screenshot 2023-08-14 at 1 52 21 PM

bnussman-akamai: Along the same lines: because we have to make a lot of API calls now, should we add a `<Skeleton />` as a placeholder while we load the data? We do this for NodeBalancer Ports and it looks nice.

Screenshot 2023-08-14 at 1 56 52 PM

bnussman-akamai (commented):
Looking good, just have a few small items before I approve ✅

packages/manager/src/features/Linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
coliu-akamai: @bnussman-akamai how's this? 👀

packages/manager/src/features/Linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
```diff
@@ -163,6 +180,21 @@ export const LinodeRow = (props: Props) => {
...
+          <Hidden smDown>
+            <TableCell noWrap>
+              {vpcLoading || configsLoading ? (
+                <Skeleton />
+              ) : vpcLabel ? (
+                <Link tabIndex={0} to={`/vpc/${vpcId}`}>
+                  {vpcLabel}
+                </Link>
+              ) : (
+                'None'
```
coliu-akamai, line 193: @bnussman-akamai added the skeleton / 'None'!

jdamore-linode:
Hey @coliu-akamai, just a heads up that there's an E2E test failure! It's nothing big, just a test expecting to see "IP Address" rather than "Public IP Address" in the Linodes landing page. The failure is in `manager/cypress/core/e2e/linodes/smoke-linode-landing-table.spec.ts` (and the specific test is `checks the table and action menu buttons/labels`).

You can fix that and verify your changes with this command while Cloud's running locally:

```bash
yarn cy:run -s "cypress/e2e/core/linodes/smoke-linode-landing-table.spec.ts"
```

Happy to help out if you have any trouble!

packages/manager/src/features/Linodes/LinodesLanding/LinodeRow/LinodeRow.tsx
coliu-akamai: Andrew approved! 🎉

coliu-akamai:
@jdamore-linode thanks for pointing that out! Those tests are now passing on local so hopefully they pass here too 🎉

bnussman-akamai (approved):
Nice job! 🚀

jdamore-linode (approved):
Great work, @coliu-akamai !

packages/validation/src/linodes.schema.ts
```diff
@@ -157,12 +157,21 @@ export const linodeInterfaceSchema = array()
         }),
       }),
       primary: boolean().notRequired(),
-      subnet: number().when('purpose', {
+      vpc_id: number().when('purpose', {
```
dwiley-akamai, line 160: My last comment regarding this was a bit off, and the API behavior for this field is a bit odd currently. It isn't required in the `POST` request (which makes sense), and including it in the request doesn't do anything (including it should probably be disallowed on the back-end 🤔). Its utility comes from its inclusion in the return object.

So we should actually pull this from `linodeInterfaceSchema` and update the changeset to only mention the `subnet` --> `subnet_id` piece.

dwiley-akamai (approved):
Good work on this! There's one cleanup item but I'm approving pending that being addressed.

PR #9487, opened 2023-08-03
fix: Capture less Sentry performance data in production

bnussman-akamai:
## Description 📝
- Decreases our Sentry performance capture rate from 20% to 10% in production

We used 100,000 transactions in 5 days with a capture rate of 20%. We need 100,000 transactions to last us 1 month so we need to use a percentage less than ~16%

## How to test 🧪
- Double check my math!

### DIFF
--- a/packages/manager/src/initSentry.ts
+++ b/packages/manager/src/initSentry.ts
@@ -82,7 +82,7 @@ export const initSentry = () => {
       ],
       integrations: [new BrowserTracing()],
       release: packageJson.version,
-      tracesSampleRate: environment === 'production' ? 0.2 : 1,
+      tracesSampleRate: environment === 'production' ? 0.1 : 1,
     });
   }
 };


packages/manager/.changeset/pr-9487-tech-stories-1691069597065.md
```diff
@@ -0,0 +1,5 @@
+---
+"@linode/manager": Tech Stories
+---
+
+Decreace Sentry capture rate ([#9487](https://github.com/linode/manager/pull/9487))
```
jdamore-linode, line 5:
```suggestion
Decrease Sentry capture rate ([#9487](https://github.com/linode/manager/pull/9487))
```
(We can always just fix this later today when we edit the changelog by hand, too)

bnussman-akamai:
Yikes, my fault. How does 2.5% sound? Should give us some wiggle room @jdamore-linode .

jdamore-linode:
> Yikes, my fault. How does 2.5% sound? Should give us some wiggle room @jdamore-linode .

That sounds like it would work to me!

jaalah-akamai (approved):

PR #9493, opened 2023-08-04
fix: IP Address copy icon regression

coliu-akamai:
## Description 📝
Due to styling changes in the IPAddress component (see this pr: https://github.com/linode/manager/pull/9445), the copy icon stopped showing up for Nodebalancers in both the Nodebalancer landing page and a Nodebalancer details page.

(Something interesting to note is that the Linodes landing page and the NodeBalancer landing pages have different behavior for the IP address copy icon. This PR just restores the NodeBalancer IP address behavior to what it used to be, but might be worth looking into trying to get them to have matching behavior)

## Major Changes 🔄
- Make IP address' copy icons show up for node balancer pages

## Preview 📷
**prod**:
image
image

**not prod**:
| Before  | After   |
| ------- | ------- |
| image | image |

| Before  | After   |
| ------- | ------- |
| image | image |

## How to test 🧪
Navigate to Nodebalancer's landing page or a single node balancer's summary page and confirm that IP address copy icon behavior is the same as prod


mjac0bs:
We should probably make sure this is branched from `[release-v1.99.0](https://github.com/linode/manager/tree/release-v1.99.0)` for this fix and then point the PR at the release branch to get this into Monday's release before the original bug makes it to prod, right, @jdamore-linode?

jdamore-linode:
> We should probably make sure this is branched from [`release-v1.99.0`](https://github.com/linode/manager/tree/release-v1.99.0) for this fix and then point the PR at the release branch to get this into Monday's release before the original bug makes it to prod, right, @jdamore-linode?

@mjac0bs Ah yeah, we probably should!

coliu-akamai:
ah thanks for the info -- let me close this PR and open up a new one then so that I can correctly branch off of release-v1.99.0!

PR #9495, opened 2023-08-04
fix: IP Address copy icon regression

coliu-akamai:
## Description 📝
Due to styling changes in the IPAddress component (see this pr: https://github.com/linode/manager/pull/9445), the copy icon stopped showing up for Nodebalancers in both the Nodebalancer landing page and a Nodebalancer details page.

(Something interesting to note is that the Linodes landing page and the NodeBalancer landing pages have different behavior for the IP address copy icon. This PR just restores the NodeBalancer IP address behavior to what it used to be, but might be worth looking into trying to get them to have matching behavior)

## Major Changes 🔄
- Make IP address' copy icons show up for node balancer pages

## Preview 📷
**prod**:
image
image

**not prod**:
| Before  | After   |
| ------- | ------- |
| image | image |

| Before  | After   |
| ------- | ------- |
| image | image |

## How to test 🧪
Navigate to Nodebalancer's landing page or a single node balancer's summary page and confirm that IP address copy icon behavior is the same as prod

### DIFF
--- a/packages/manager/src/features/NodeBalancers/NodeBalancerDetail/NodeBalancerSummary/SummaryPanel.tsx
+++ b/packages/manager/src/features/NodeBalancers/NodeBalancerDetail/NodeBalancerSummary/SummaryPanel.tsx
@@ -97,9 +97,11 @@ export const SummaryPanel = () => {
         <StyledSection>
           <StyledIPGrouping data-qa-ip>
             {nodebalancer?.ipv4 && (
-              <IPAddress ips={[nodebalancer?.ipv4]} showMore />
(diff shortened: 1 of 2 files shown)


mjac0bs (approved):
Good catch on this before it even was released, @coliu-akamai! Confirming that the original behavior of the copy icons for Nodebalancers is restored. 

You make a good point about the differing icon behavior between Linodes and Nodebalancers. (And I wonder if anywhere else?) This seems worth a follow up conversation with UX about after we get this fix merged.

bnussman-akamai (approved):

PR #9497, opened 2023-08-05
refactor: [M3-6922] - Add Autocomplete Component

jaalah-akamai:
## Description 📝
We want to enhance accessibility and streamline our use of third-party libraries by eliminating our reliance on `react-select`. This leverages a customized version of MUI Autocomplete which will enable us to gradually replace occurrences of `EnhancedSelect` across our entire codebase.

## Major Changes 🔄
- [x] Add generic Autocomplete that _should_ fit our needs. We can iterate on this component as we come across edge cases.
- [x] Added a complete storybook, which also includes an example of how we can transition our `Backups` and `Clone Linode` UI to an Autocomplete.
- [x] Moved `LinodeSelect.styles` to `Autocomplete.styles` since they made more sense there.
- [x] Added basic unit tests
- [x] Created a new hook for `Select All` feature

## TODO
- [x] Extend AutocompleteProps from MUI
- [x] Allow for flags or icons to be displayed (regions)
- [x] Clean up `renderOptions` by maybe being more explicit as to what can be passed in

## Preview 📷
| Single Select  | Multi-Select  |
| ---------- | ------- |
| Screenshot 2023-08-05 at 12 18 07 AM |  Screenshot 2023-08-08 at 10 45 11 PM 
| |Screenshot 2023-08-08 at 10 45 25 PM |
| |Screenshot 2023-08-08 at 10 45 00 PM |
| |<video src="https://github.com/linode/manager/assets/125309814/69abdf90-d5d7-4b23-a8cc-ea89bc16b364" />|

| Custom No Options Message  | Custom Render Options  |
| ------- | ------- |
| Screenshot 2023-08-05 at 12 17 54 AM | Screenshot 2023-08-05 at 12 17 45 AM  |

| Regions  |
| ------- | 
| Screenshot 2023-08-11 at 10 52 00 AM  |

| Separate Options List  | |
| ------- | ------- |
| Screenshot 2023-09-06 at 11 17 30 AM  | Screenshot 2023-09-06 at 11 17 34 AM |

## How to test 🧪
1. yarn storybook
2. http://localhost:6006/?path=/docs/components-autocomplete--documentation

Screenshot 2023-08-05 at 12 44 51 AM


packages/manager/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+import { CustomPopper, SelectedIcon } from './Autocomplete.styles';
+
+export interface OptionsType {
+  data?: any;
+  label: string;
+  value: string;
+}
+
+export interface AutocompleteProps {
```
abailly-akamai, line 19: You are overriding a lot of MUI `Autocomplete` props which I think we should avoid to prevent breaking changes when updating MUI. Rather we should extend the MUI component's props and only declare our custom props. (same with other interfaces below)

They will be automatically picked up by storybook

packages/manager/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,261 @@
...
+    onSelectionChange,
+    options,
+    optionsFilter,
+    placeholder,
+    renderOption,
+    renderOptionLabel,
+    sx,
+  } = props;
+
+  const [inputValue, setInputValue] = React.useState('');
```
abailly-akamai, line 136: Out of curiosity, can the value be an object if (for instance) there's an icon to be displayed with the label and in the selected state? I assume this could be also an iteration

packages/manager/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,261 @@
...
+          label={label}
+          loading={loading}
+          noMarginTop={noMarginTop}
+          placeholder={placeholder ?? 'Select an option'}
+          {...params}
+        />
+      )}
+      renderOption={(props, option: OptionsType, { selected }) => {
+        return (
+          <li {...props}>
```
abailly-akamai, line 199: Could we just pass the props we need here?

packages/manager/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,261 @@
...
+      popupIcon={<KeyboardArrowDownIcon />}
+      sx={sx}
+    />
+  );
+};
+
+interface DefaultNoOptionsMessageParams {
+  errorText: APIError[] | null | string | undefined;
+  filteredOptions: OptionsType[] | undefined;
+  loading: boolean | undefined;
```
abailly-akamai, line 242: This feels a bit odd, since it should just be a boolean (pretty low on this, not sure it matters in any way)

packages/manager/src/components/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+    label: 'Select a Fruit',
+    onSelectionChange: action('onSelectionChange'),
+    options: fruits,
+  },
+  component: Autocomplete,
+  decorators: [
+    (Story: StoryFn) => (
+      <div style={{ margin: '2em' }}>
+        <Story />
+      </div>
```
abailly-akamai, line 56: Maybe a min-height so we don't have to scroll the stories to see then options when the Autocomplete is open?

abailly-akamai (commented):
This is really nice and clean. Great work!

Added some comments to clean things up. 

Also, have we discussed the use of the "Autocomplete" VS "Select" nomenclature. I don't think it is a crucial discussion but I'd be curious to know why gearing towards the MUI naming convention rather than keeping our existing one

jaalah-akamai:
> This is really nice and clean. Great work!
> 
> Added some comments to clean things up.
> 
> Also, have we discussed the use of the "Autocomplete" VS "Select" nomenclature. I don't think it is a crucial discussion but I'd be curious to know why gearing towards the MUI naming convention rather than keeping our existing one

Yes, I thought about this too! The existing `EnhancedSelect` allows for typeahead interactions, so that falls within the `Autocomplete` nomenclature

packages/manager/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,261 @@
...
+     * We want to clear the input value when the value prop changes to null.
+     * This is for use cases where a user changes their region and the Linode
+     * they had selected is no longer available.
+     */
+    if (defaultValue === null) {
+      setInputValue('');
+    }
+  }, [defaultValue]);
+
+  const handleChange = (value: OptionsType | OptionsType[]) => {
```
tyler-akamai, line 153: Would it make sense to wrap this function in `React.useCallback(...)` to prevent unnecessary re-rendering?

```
const handleChange = React.useCallback((value: OptionsType | OptionsType[]) => {
  if (Array.isArray(value)) {
    // It's an array, so it's the multiple selection case
    if (!multiple) {
      return;
    }
    onSelectionChange(value);
  } else {
    // It's not an array, so it's the single selection case
    if (multiple) {
      return;
    }
    onSelectionChange(value);
  }
}, [multiple, onSelectionChange]);
```

packages/manager/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,261 @@
...
+import CloseIcon from '@mui/icons-material/Close';
+import KeyboardArrowDownIcon from '@mui/icons-material/KeyboardArrowDown';
+import MuiAutocomplete from '@mui/material/Autocomplete';
+import { SxProps } from '@mui/system';
+import React from 'react';
+
+import { Box } from 'src/components/Box';
+import { TextField } from 'src/components/TextField';
+
+import { CustomPopper, SelectedIcon } from './Autocomplete.styles';
```
tyler-akamai, line 11: I think the customPopper styling can be simplified:

Current code:
```
export const CustomPopper = (props: PopperProps) => {
  return (
    <Popper
      {...props}
      style={{
        ...(props.style ?? {}),
        ...(props.style?.width
          ? typeof props.style.width == 'string'
            ? { width: `calc(${props.style.width} + 2px)` }
            : { width: props.style.width + 2 }
          : {}),
        zIndex: 1,
      }}
      data-qa-autocomplete-popper
      modifiers={[{ enabled: false, name: 'preventOverflow' }]}
    />
  );
};
```

i dont think you need this conditional:
```
typeof props.style.width == 'string'
            ? { width: `calc(${props.style.width} + 2px)` }
            : { width: props.style.width + 2 }
```

I think the `calc(${props.style.width} + 2px)` will work whether the width is a string or number.

packages/manager/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,261 @@
...
+    helperText,
+    id,
+    label,
+    loading,
+    multiple,
+    noMarginTop,
+    noOptionsMessage,
+    onBlur,
+    onSelectionChange,
+    options,
```
tyler-akamai, line 128: Is it a good idea to provide a default value here to prevent optional chaining?

packages/manager/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,261 @@
...
+}
+
+/**
+ * Props for the Autocomplete component in single-select mode.
+ */
+export interface AutocompleteSingleSelectProps extends AutocompleteProps {
+  /** Default value for the Autocomplete. */
+  defaultValue?: OptionsType | null;
+  /** Enable single-select mode. */
+  multiple?: false;
```
tyler-akamai, line 88: Why is `multiple` for `AutocompleteSingleSelectProps` optional and `multiple` for `AutocompleteMultiSelectProps` required?

packages/manager/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+type DefaultNoOptionsMessageResult = JSX.Element | string;
+
+const getDefaultNoOptionsMessage = ({
+  errorText,
+  filteredOptions,
+  loading,
+}: DefaultNoOptionsMessageParams): DefaultNoOptionsMessageResult => {
+  if (errorText) {
+    return 'An error occurred while fetching your options';
```
tyler-akamai, line 253: Would it make sense to use an enum here?

jaalah-akamai:
Sorry @tyler-akamai been making major changes to this PR, I'm going to close out-of-date comments, but looking forward to your review when I put it up again!

packages/manager/src/components/Autocomplete/Autocomplete.tsx
jaalah-akamai: Yes I agree

packages/manager/src/components/Autocomplete/Autocomplete.tsx
jaalah-akamai: I'm not so sure if what you suggest would work, however I cleaned this up regardless. If you have time to help me test this aspect when this PR is ready I'd appreciate it!

packages/manager/src/features/Linodes/LinodeSelect/LinodeSelect.tsx
bnussman-akamai, line 1: So it the long-term goal to have `LinodeSelect` use this new `Autocomplete` or will `LinodeSelect` continue to import directly from MUI?

packages/manager/src/components/Autocomplete/Autocomplete.tsx
tyler-akamai: Just tested it, your right, it doesnt work like that haha

packages/manager/src/components/Autocomplete/Autocomplete.tsx
jaalah-akamai: Options is required, so I don't think we need a default unless I'm misunderstanding what you're saying

packages/manager/src/components/Autocomplete/Autocomplete.tsx
jaalah-akamai: We're type-checking that only list item properties are applied here, I'd rather keep it for flexibility.

packages/manager/src/components/Autocomplete/Autocomplete.tsx
jaalah-akamai: I'm assuming you mean like Regions selector? If so, I added an example of how to do that via the optional `data` property

packages/manager/src/components/Autocomplete/Autocomplete.tsx
jaalah-akamai: Consider this done!

jaalah-akamai:
There's still a lot more unit tests we can write for this. I may tackle writing them in a separate PR due to my time constraints or someone on the team can pick that up. Ready for review again! 🚀

packages/manager/src/components/Autocomplete/Autocomplete.tsx
tyler-akamai: if required then we dont need optional chaining on line 139, is that correct?

packages/manager/src/components/Autocomplete/Autocomplete.tsx
dwiley-akamai: To clarify, this should already be working for the selected option too? I might be missing something but in Storybook it doesn't seem like the flag gets displayed upon selecting an option.

dwiley-akamai (approved):
This is looking good based on my testing 🎉

We discussed on Friday during the group review/testing that some improvements can be made for the types (e.g. the one marked `TODO` in `Autocomplete.tsx`), but I think that could be done in a follow-up PR or iteratively as we implement this component in the app.

packages/manager/src/components/Autocomplete/Autocomplete.styles.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  color: theme.color.headline,
+  fontFamily: theme.font.bold,
+  fontSize: '1rem',
+  marginBottom: '9px',
+  position: 'relative',
+}));
+
+export const SelectedIcon = styled(DoneIcon, {
+  shouldForwardProp: (prop) => prop != 'visible',
```
hana-linode, line 34: Should we add a label here as well?

packages/manager/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,305 @@
...
+} from '@mui/material/Autocomplete';
+
+export interface OptionType<T = any> {
+  data?: T;
+  label: string;
+  value: string;
+}
+
+interface DefaultNoOptionsMessage {
+  errorText: APIError[] | null | string | undefined;
```
hana-linode, line 32:
```suggestion
  errorText?: APIError[] | null | string;
```

packages/manager/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,305 @@
...
+  const {
+    handleClearOptions,
+    handleSelectAll,
+    handleToggleOption,
+    selectedOptions,
+  } = useSelectAllOptions([]);
+
+  const [inputValue, setInputValue] = React.useState('');
+  const [selectAllActive, setSelectAllActive] = React.useState(false);
+  const selectAllText = selectAllActive ? 'Deselect All' : 'Select All';
```
hana-linode, line 132: Thoughts on the term `Unselect` vs `Deselect`?

packages/manager/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,305 @@
...
+  errorText,
+  options,
+}: DefaultNoOptionsMessage): NoOptionsMessage => {
+  if (errorText) {
+    return NoOptionsMessage.Error;
+  } else if (!options?.length) {
+    return NoOptionsMessage.NoOptions;
+  } else {
+    return NoOptionsMessage.NoResults;
+  }
```
hana-linode, line 304: We can unnest the ifs since each option returns out of the function
```suggestion
  if (errorText) {
    return NoOptionsMessage.Error;
  }
  if (!options?.length) {
    return NoOptionsMessage.NoOptions;
  }
  return NoOptionsMessage.NoResults;
```

packages/manager/src/components/Autocomplete/Autocomplete.tsx
jaalah-akamai: No preference, open to feedback

packages/manager/src/components/Autocomplete/Autocomplete.tsx
jaalah-akamai: I have no preference since it's interchangeable - open to feedback from others on both conventions

tyler-akamai: I would say the second one looks a little more readable.

packages/manager/src/components/Autocomplete/Autocomplete.tsx
tyler-akamai: `Dictionaries (Merriam-Webster and New Oxford American Dictionary) have deselect but not unselect. The NOAD defines deselect as “turn off (a selected feature) on a list of options on a computer menu”, which is what you want.

Unselected, on the other hand, is used to qualify something that has not been selected, not something that was selected and isn't anymore.` - https://english.stackexchange.com/questions/18465/unselect-or-deselect

I would probably go with deselect. It also seems more common across ux design.

jaalah-akamai:
Added a storybook example for having separate options as opposed to using `Tags`. 

@dwiley-akamai Adding flags to the value is not something we can do easily given the types we're restricted to. This may be one tradeoff unless we do a customized input.

bnussman-akamai:
Looks great but I have some general concerns about the typesafety of this component.

- No type safety on the `data`
- `renderInput` is required
- The `onSelectionChange` value is an array even though `multiple` is `false`

Screenshot 2023-09-06 at 12 50 39 PM

None of this stuff is easy but I propose we try to align our component more with MUI's implementation
- Allow options to be any type `T`
- Use `onChange` instead of a custom `onSelectionChange`
- And so on...

packages/manager/src/foundations/themes/light.ts
```diff
@@ -662,7 +667,6 @@ export const lightTheme: ThemeOptions = {
             backgroundColor: 'transparent',
             color: primaryColors.main,
           },
-          padding: 12,
```
jaalah-akamai, line 665: This was a bug that prevented the `IconButton` `size` prop from working properly. There should be no regressions since we default to the large size which has a padding of `12` by default. https://github.com/linode/manager/pull/9497/files#diff-f767369a01cd40921dc0da8dac44cf6777e3c155f3680aa2990be0da93c2dc48R659

jaalah-akamai:
Thanks @bnussman-akamai for the review and providing an alternative solution 🎉

packages/manager/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,170 @@
...
+  /** A required label for the Autocomplete to ensure accessibility. */
+  label: string;
+  /** Removes the top margin from the input label, if desired. */
+  noMarginTop?: boolean;
+  /** Callback function triggered when the input field loses focus. */
+  onBlur?: (e: React.FocusEvent) => void;
+  /** Placeholder text displayed in the input field. */
+  placeholder?: string;
+  /** Indicates whether the input is required, displaying an appropriate indicator. */
+  required?: boolean;
```
bnussman-akamai, line 39: I think we can remove these 3 props because `AutocompleteProps` already includes them!

bnussman-akamai (changes requested):

bnussman-akamai (approved):
I think this is a great starting point! 🎉 We can refine this component as we use it and see what changes it may need

PR #9517, opened 2023-08-08
refactor: [M3-6953] - Refactor Backups Drawer

bnussman-akamai:
## Description 📝
- Refactors the Backup Drawer to use modern patterns which will hopefully result in cleaner code and better maintainability 🧹 We're doing this because the Backup Drawer has become a serious pain-point due to its crustyness.

## Major Changes 🔄
- Removes Redux from this component completely 🗑️
  - This component use to manage its state with Redux. This was overkill and added extra complexity. 
- Only mounts this component where it is needed ☑️
  - This component use to be mounded at all times in `MainContent.tsx`. This is overkill because this component is only needed on two pages within the app. It was likely also hurting performance when it was mounted all the time. 
- Lots of refactoring and moving things around to make the code cleaner and more consumable 🧼

### Error Handling ⛔️
 
 > Shows error handling working as expected

https://github.com/linode/manager/assets/115251059/4097971f-b438-40af-b9ac-6722ea8ed785

 ### Normal Expected Use ✅
 
 > Shows the "happy-path" of how a user might use this

https://github.com/linode/manager/assets/115251059/b2e86a6e-fb4c-4bb2-b893-c2481addc179

## How to test 🧪
- Verify the Backups drawer works as expected
  - Behavior and UX should be mostly the same as before but please verify there are no changes that negatively impacted the UX
  - Verify you can enable backups
  - Verify you can enable backup auto-enrollment
  - Consider blocking a network request to verify the error state looks good
  
### How to make this component show up ❓
- Go to `http://localhost:3000/profile/settings?preferenceEditor=true` and set your preferences to `{}` and **Save**
- Deploy a Linode with backups disabled
- Make sure you are on an unrestricted and non-managed account
- You should see the Backups "Call to Action" banner on the Linodes Landing page 🎉


packages/manager/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     /* Language and Environment */
     "jsx": "react",
-    "lib": ["dom", "es2019"],
+    "lib": ["dom", "es2020"],
```
bnussman-akamai, line 5: Made this change so we can use `Promise.allSettled`. Should have no effect on the Vite bundle.

packages/manager/src/features/Backups/BackupDrawer.tsx
```diff
@@ -1,398 +1,202 @@
...
-    dismissSuccess,
-    enqueueSnackbar,
-  ]);
+    return linodesWithoutBackups.map((linode) => (
+      <BackupLinodeRow
+        error={
+          (enableBackupsResult?.find(
+            (result) =>
+              result.linode.id === linode.id && result.status === 'rejected'
+          ) as EnableBackupsRejectedResult | undefined)?.reason?.[0]?.reason
```
tyler-akamai, line 100: You can get rid of 'undefined' after EnableBackupsRejectedResult

tyler-akamai: I think this can actually be minimized to:

`as EnableBackupsRejectedResult)?.reason[0].reason`

since both 'reason: APIError[]' is required in 'EnableBackupsRejectedResult' and `reason` is required in `APIError`

packages/manager/src/features/Backups/BackupDrawer.tsx
```diff
@@ -1,398 +1,202 @@
...
 
-  const linodesWithoutBackups = (linodes: Linode[] | undefined) => {
-    return linodes?.filter((linode) => !linode.backups.enabled) ?? [];
-  };
+  const [shouldEnableAutoEnroll, setShouldEnableAutoEnroll] = React.useState(
+    true
+  );
 
-  const prevLinodesData = usePrevious(linodesData);
+  const {
```
tyler-akamai, line 68: Out of curiosity, I noticed that `useMutateAccountSettings()` returns Errors. Would this be something that we should add to `useEnableBackupsOnLinodesMutation()` since we are using it to render the table on line 96.

packages/manager/src/features/Backups/BackupDrawer.tsx
```diff
@@ -1,398 +1,202 @@
...
         {/* Don't show this if the setting is already active. */}
-        {!accountSettings.data?.backups_enabled && (
-          <Grid>
-            <AutoEnroll
-              enabled={autoEnroll}
-              error={autoEnrollError}
-              toggle={toggle}
-            />
-          </Grid>
+        {!accountSettings?.backups_enabled && (
```
tyler-akamai, line 164: Do we need optional chaining here?

packages/manager/src/features/Backups/BackupDrawer.tsx
bnussman-akamai: Because I chose to use `Promise.allSettled`, any errors that happen during the API calls will be included in the `data` value.

packages/manager/src/features/Backups/BackupDrawer.tsx
bnussman-akamai: This chunk of chunk of code is the part of this PR i'm lead proud because it's kind of hard to look at 😵‍💫

I think the `undefined` should be kept because `Array.find` always has a return type of `T | undefined`. We want to make sure we handle `undefined` accordingly if the `.find` happens to return `undefined`.

packages/manager/src/features/Backups/BackupLinodeRow.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useTypeQuery } from 'src/queries/types';
+
+interface Props {
+  error?: string;
+  linode: Linode;
+}
+
+export const BackupLinodeRow = (props: Props) => {
+  const { error, linode } = props;
+  const { data: type } = useTypeQuery(linode.type ?? '', Boolean(linode.type));
```
tyler-akamai, line 16: Out of curiosity, why does `useTypeQuery` return `LinodeType` but the 'Type' property within the `Linode` interface is of type `string`. Wouldn't it just easier to have the `Type` property within `Linode` be a `LinodeType`?

packages/manager/src/features/Backups/BackupLinodeRow.tsx
```diff
@@ -0,0 +1,41 @@
...
+              fontSize: 13,
+            })}
+            variant="body1"
+          >
+            {error}
+          </Typography>
+        )}
+      </TableCell>
+      <TableCell parentColumn="Plan">
+        {type?.label ?? linode.type ?? 'Unknown'}
```
tyler-akamai, line 34: and then down here, you wont have to have multiple checks.

packages/manager/src/features/Backups/BackupLinodeRow.tsx
bnussman-akamai: Yes 100% would be easier but this is API data. The API will respond with the following for a Linode:

```json
{
    "id": 36348428,
    "label": "beep-production-redis",
    "group": "",
    "status": "running",
    "created": "2022-05-15T22:24:07",
    "updated": "2023-04-02T16:03:02",
    "type": "g6-nanode-1", ⬅️ API gives us the type ID
    "ipv4": [],
    "image": "linode/alpine3.15",
    "region": "us-southeast",
    "specs": {
        "disk": 25600,
        "memory": 1024,
        "vcpus": 1,
        "gpus": 0,
        "transfer": 1000
    },
    "alerts": {
        "cpu": 90,
        "network_in": 10,
        "network_out": 10,
        "transfer_quota": 80,
        "io": 10000
    },
    "backups": {
        "enabled": false,
        "available": false,
        "schedule": {
            "day": null,
            "window": null
        },
        "last_successful": "2023-01-06T15:40:11"
    },
    "hypervisor": "kvm",
    "watchdog_enabled": true,
    "tags": []
}
```

so then we have to make another API call to `/v4/linode/types/](https://api.linode.com/v4/linode/types/g6-nanode-1` to get more information on the Linode's type

packages/manager/src/features/Backups/BackupLinodeRow.tsx
bnussman-akamai: Can you clarify what you mean? 

What I'm doing here is: Render the type's `label` if it is loaded from the API, otherwise fallback to the Linode's type ID, otherwise fall-back to `Unknown` if the Linode has no type id

packages/manager/src/features/Backups/BackupDrawer.tsx
tyler-akamai: Gotcha, I didn't know that.

I did have a hard time reading it at first haha.

What you could do to clean it up is to have a separate function:

```
const findRejectedBackupResult = (linodeId) => {
  return enableBackupsResult?.find(
    (result) => result.linode.id === linodeId && result.status === 'rejected'
  ) as EnableBackupsRejectedResult | undefined;
};

...

<BackupLinodeRow
  error={findRejectedBackupResult(linode.id)..reason?.[0]?.reason}
  key={linode.id}
  linode={linode}
/>
```

Up to you though, its not a big deal

packages/manager/src/features/Backups/BackupLinodeRow.tsx
tyler-akamai: Ah okay this makes more sense

tyler-akamai:
Great job!

cpathipa (approved):

PR #9530, opened 2023-08-09
feat: [M3-6945] – Support VPC and Subnet events

dwiley-akamai:
## Description 📝
Add the following VPC and Subnet events to `eventMessageGenerator.ts`:
- `vpc_create`
- `vpc_update`
- `vpc_delete`
- `subnet_create`
- `subnet_update`
- `subnet_delete`

~~WIP until API clarifies whether the event objects will be fleshed out for these events.~~

Additionally, I updated the routing from `/vpc` to `/vpcs` to be consistent with other entities like `/linodes`, `/databases`, etc.

## Preview 📷
Screenshots attached to the ticket

## How to test 🧪
- Confirm that VPC and subnet events appear as expected with proper links in Cloud
- Confirm all VPC routing functions as it should


abailly-akamai (approved):
👍

jdamore-linode:
Do these events need to be added to the API package at [`src/account/types.ts`](https://github.com/linode/manager/blob/develop/packages/api-v4/src/account/types.ts#L232)? I'm noticing a type error when attempting to add these to the [Cypress events smoke test](https://github.com/linode/manager/blob/develop/packages/manager/cypress/e2e/core/notificationsAndEvents/events.spec.ts) since these aren't present in the `EventAction` type.

packages/api-v4/src/account/types.ts
dwiley-akamai, line 1: Only the VPC and Subnet events are new here; I just alphabetized the types to find them more easily in the future.

coliu-akamai (commented):
I don't think I'd have found this if I didn't happen to have a VPC named 'te' but the following is happening for me 😅 

image
image


Also, on the VPC landing page, the Create VPC button still routes to `/vpc/create`

dwiley-akamai:
> I don't think I'd have found this if I didn't happen to have a VPC named 'te' but the following is happening for me 😅
> 
> image image
> 
> Also, on the VPC landing page, the Create VPC button still routes to `/vpc/create`

Unfortunately that's due to an existing issue with events logic, I also noticed it when testing on this PR and created M3-7056 for it.

Thanks, I will push up an adjustment for the Create VPC button shortly

coliu-akamai (approved):
🎉

hana-linode (approved):
LGTM, should we also rename the features folder from `VPC` to `VPCs`?

cliu-akamai (approved):

dwiley-akamai:
> LGTM, should we also rename the features folder from `VPC` to `VPCs`?

I agree with that. I'll open a follow-up PR doing that so the diff here doesn't get out of control

PR #9531, opened 2023-08-10
poc: See if unit tests are any faster on latest Node.js

bnussman-akamai:
## Description 📝
- I feel like unit tests got significantly slower when we upgraded to Node 18, but I might be wrong. I just want to see if this has any effect on the test speed 🏎️

## How to test 🧪
- Look at Github Actions results


bnussman-akamai:
OMG! It's slower 🐌

PR #9532, opened 2023-08-10
refactor: [M3-6281] - SRC > Features > Search CSS Migration

tyler-akamai:
## Description 📝
Migrated Features > Search to MUI v5

## Major Changes 🔄
1. Remove `React.FC` or `React.FunctionComponent` and rename interface `Props` to `{ComponentName}Props`. Interface can be applied to `props` parameter. IE: (props: MyProps)
2. Removed `CombinedProps` if it's directly referencing `Props` with nothing additional
3. Removed `makeStyles` for new `styled API`. We typically use `Styled{ComponentName}` as our prefix
4. `export` component inline, and remove `export default {ComponentName}` at bottom of file.
5. Eliminated unnecessary higher order component
6. Updated test from enzyme to react testing library

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-08-10 at 5 11 27 PM | Screenshot 2023-08-10 at 5 11 52 PM |
| Screenshot 2023-08-10 at 5 10 49 PM | Screenshot 2023-08-10 at 5 10 26 PM |

## How to test 🧪
1. Navigate to http://localhost:3000/search
2. Ensure the 'You searched for ...' icon is the same
3. Ensure the search bar is the same
4. Type something into the search bar. Ex. "label:"
7. Ensure the search result page looks the same


packages/manager/src/features/Search/SearchLanding.styles.ts
```diff
@@ -0,0 +1,55 @@
...
+  marginBottom: theme.spacing(),
+  [theme.breakpoints.down('md')]: {
+    marginLeft: theme.spacing(),
+  },
+}));
+
+export const StyledRootGrid = styled(Grid, {
+  label: 'StyledRootGrid',
+})({
+  '&.MuiGrid-container': {
```
jaalah-akamai, line 51: This classname may not be necessary, but we should keep `width` - You'll have to confirm this

packages/manager/src/features/Search/SearchLanding.tsx
```diff
@@ -314,9 +277,9 @@ export const SearchLanding: React.FC<CombinedProps> = (props) => {
         </Grid>
       )}
       {resultsEmpty && !loading && (
-        <Grid className={classes.emptyResultWrapper} data-qa-empty-state>
-          <div className={classes.emptyResult}>
-            <Error className={classes.errorIcon} />
+        <StyledGrid data-qa-empty-state>
+          <StyledDiv>
```
jaalah-akamai, line 281: `StyledDiv` could probably utilize the `Stack` component for future reference

packages/manager/src/features/Search/SearchLanding.styles.ts
tyler-akamai: Yeah, we can get rid of `&.MuiGrid-container`

packages/manager/src/features/Search/SearchLanding.tsx
tyler-akamai: Just changed it to Stack and added appropriate styling

packages/manager/src/features/Search/SearchLanding.tsx
```diff
@@ -231,7 +195,7 @@ export const SearchLanding: React.FC<CombinedProps> = (props) => {
     _privateImages,
     regions,
     nodebalancers,
-    searchableLinodes,
+    //searchableLinodes,
```
tyler-akamai, line 198: There is currently an infinite loop issue with this code. It seems like `seach(...)` in the `else`  statement changes updates `searchableLinodes` causing an infinite loop. Commenting out `searchableLinodes` in the dependency array solves the issue but we might want to create a new ticket to look more into this issue.

tyler-akamai: This was fixed with the merge

coliu-akamai:
Seems to be an issue on prod as well, but for some reason the first letter of a searched word or phrase with no results is always upside down 😆 
image
maybe we should put up a ticket for that

coliu-akamai (commented):
Good work! Saw a few changes compared to prod, am going to check if this is also happening on develop and not just your branch:

1) might only be on my end, but for the search results page, it seems like the description font is a bit smaller on local now than prod (tried to get similar sized images for comparison, but probably easier to verify yourself)

image image

2) search bar content looks different after a search (looking through your changes though I'm not sure what could have caused them, so let me check develop as well)
| Local  | Prod   |
| ------- | ------- |
| image | image |

coliu-akamai:
Just checked again -- crossed out the second thing since that was related to a PR. For the first one: I think the smaller size of the text is fine as that was also happening on develop; however, you might be missing a bit of padding for each row of the search results in comparison to develop and [this pr](https://github.com/linode/manager/pull/9534)

packages/manager/src/features/Search/ResultRow.styles.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export const StyledTableRow = styled(TableRow, {
+  label: 'StyledTableRow',
+})(({ theme }) => ({
+  '& td': {
+    paddingBottom: 4,
+    paddingTop: 4,
+  },
+  paddingBottom: '2px !important',
+  paddingTop: '0 !important',
```
coliu-akamai, line 54: try seeing if you're able to get rid of the !important but if not no worries

coliu-akamai (approved):
looks good! one small comment, but am approving

packages/manager/src/features/Search/ResultRow.styles.ts
tyler-akamai: Good catch, just fixed it.

bnussman-akamai (approved):

PR #9537, opened 2023-08-11
feat: [M3-6757] - Create VPC page

coliu-akamai:
## Description 📝
- Add VPC create page
- ** for v2 of validating subnets, go here: https://github.com/coliu-akamai/manager/pull/1

## Preview 📷
image
image
image

## How to test 🧪
` yarn test VPCCreate`
` yarn test subnets.test.ts`
- turn on mocks and navigate to /vpc/create to create a vpc OR use your dev account 9must have vpc abilities enabled)


packages/manager/src/features/VPC/VPCCreate/VPCCreate.tsx
```diff
@@ -1,13 +1,262 @@
...
+  const { data: regions } = useRegionsQuery();
+  const { isLoading, mutateAsync: createVPC } = useCreateVPCMutation();
+  const [subnetErrorsFromAPI, setSubnetErrorsFromAPI] = React.useState<
+    APIError[]
+  >();
+
+  // TODO: VPC - may eventually need to add global grants?
+  // customer tag? + to test, will need to comment out the account restriction
+  const disabled =
+    profile?.restricted || !account?.capabilities.includes('VPCs');
```
coliu-akamai, line 45: does this page need to check for account.capabilities?
++ for testing, will need to change this line as right now I think no account has VPC capabilities

packages/manager/src/features/VPC/VPCCreate/VPCCreate.tsx
```diff
@@ -1,13 +1,262 @@
...
+        'subnets',
+        validatedSubnets.map((subnet) => {
+          delete subnet.labelError;
+          delete subnet.ip.ipv4Error;
+          return { ...subnet };
+        })
+      );
+      return true;
+    }
+  };
```
coliu-akamai, line 85: This method is pretty similar to the validateIP method in DatabaseCreate -- I'm not the biggest fan of this validation outside of using the createVPCSchema, but was having trouble trying to parse through the errors returned from the schema. Playing around with this on my branch create-vpc-offshoot1, but it's messy rn

packages/manager/src/utilities/subnets.ts
```diff
@@ -0,0 +1,100 @@
+import { determineIPType, vpcsValidateIP } from "@linode/validation";
```
coliu-akamai, line 1: see comment on [the VPC Create page](https://github.com/linode/manager/pull/9537/files#diff-853e4046c348f0b47bba91c8c328c61a0314c90e8e25efc761b25d20a09db1b3) about validation -- testing things out on branch create-vpc-offshoot1 to see if I can get rid of validation that doesn't use the createVPCSchema

packages/validation/src/vpcs.schema.ts
```diff
@@ -14,6 +14,22 @@ const labelTestDetails = {
...
+      addr = parsed[0];
+    } else {
+      addr = ipaddr.parse(ip);
+    }
+    return addr.kind();
+  } catch (e) {
+    return undefined;
+  }
+};
+
```
coliu-akamai, line 32: moved this functionality outside of vpcsValidateIP and into its own function in order to be able to determine ip type for calculating the number of available ipv4s of a subnet mask

packages/manager/src/features/VPC/VPCCreate/SubnetNode.tsx
```diff
@@ -0,0 +1,112 @@
...
+  disabled?: boolean;
+  idx?: number;
+  // janky solution to enable SubnetNode to be an independent component or be part of MultipleSubnetInput
+  // (not the biggest fan tbh)
+  onChange: (
+    subnet: SubnetFieldState,
+    subnetIdx?: number,
+    remove?: boolean
+  ) => void;
+  removable?: boolean;
```
coliu-akamai, line 23: Trying to make SubnetNode its own component in order for it to have future use cases (like the edit/create Subnet drawers), but that led to packing the onChange method with some parameters that wouldn't be necessary if this file was combined with MultiplleSubnetInput.tsx and some other props (like removable) being included. The[ MultipleSubnetInput.tsx ](https://github.com/linode/manager/pull/9537/files#diff-e841c5d21392f61c38ba3e0c0db71760b8ef1e25f9534c429be008bba81e6503) functionality is otherwise along the same idea as the [MultipleIpInput](https://github.com/linode/manager/blob/develop/packages/manager/src/components/MultipleIPInput/MultipleIPInput.tsx) component

packages/manager/src/features/VPC/VPCCreate/MultipleSubnetInput.tsx
```diff
@@ -0,0 +1,68 @@
...
+        <Grid key={`subnet-${subnetIdx}`}>
+          {subnetIdx !== 0 && <Divider sx={{ marginTop: theme.spacing(3) }} />}
+          <SubnetNode
+            idx={subnetIdx}
+            subnet={subnet}
+            disabled={disabled}
+            // janky solution to enable SubnetNode to work on its own or be part of MultipleSubnetInput
+            onChange={(subnet, subnetIdx, removable) =>
+              handleSubnetChange(subnet, subnetIdx ?? 0, !!removable)
+            }
```
coliu-akamai, line 53: see comment at [SubnetNode.tsx](https://github.com/linode/manager/pull/9537/files#diff-4a4cf79cebe1d574186f8716a6fa0f20592c2881d4002d0870c7c6b7bdadd243)

packages/manager/src/features/VPC/VPCCreate/VPCCreate.tsx
dwiley-akamai: We shouldn't have to check for `account.capabilities` here. There's a ticket to handle that piece, I believe we'd just add some `isFeatureEnabled` logic in `VPC/index.ts` which would trickle down to this.

`disabled` here would just be based on whether the profile is restricted and whether they have a grant to create VPCs

dwiley-akamai (commented):
The UI is looking solid so far and I was able to create a VPC successfully 🎉

I still have a lot of code to comb through but I left some initial comments. Will be sure to dedicate some time tomorrow to looking at this more thoroughly.

packages/manager/src/features/VPC/VPCCreate/VPCCreate.tsx
coliu-akamai: ooh I had that before and then changed it for some reason ;-; let me go and add that back in!

packages/manager/src/cachedData/regions.json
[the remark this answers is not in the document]
coliu-akamai: Hm... honestly I'm a bit confused about why this file got added in -- looking at the difference, "Metadata" got taken out of the lines ..."Managed Databases","Premium Plans"..., but I'm not sure why the changes in this commit would have caused that. 

Do you know how these cached files get generated / what are they used for?

coliu-akamai:
Second version of validation here: https://github.com/coliu-akamai/manager/pull/1

packages/manager/src/features/VPC/VPCCreate/VPCCreate.tsx
```diff
@@ -1,11 +1,243 @@
...
+
+  const {
+    values,
+    errors,
+    setFieldValue,
+    setFieldError,
+    setSubmitting,
+    handleChange,
+    handleSubmit,
+  } = useFormik({
```
bnussman-akamai, line 124: I think it is best to try to keep the Formik values in the shape the API would expect. Is this possible?

```suggestion
  } = useFormik<CreateVPCPayload>({
```

coliu-akamai: Imo it's possible but potentially weird, due to the validation checks on subnets. In the other [validation version here](https://github.com/coliu-akamai/manager/pull/1), I at one point had the Formik values in the shape of the API data, but when I deleted subnets, the errors shown on the UI would shift and no longer correspond to the same subnets. One idea I had to get around that would be to clear subnet errors any time there's a change to the subnets (but to clear all subnet errors if I change a single subnet feels a bit strange)

packages/manager/src/cachedData/regions.json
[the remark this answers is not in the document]
dwiley-akamai: `yarn build` is probably why. Most of us run `yarn dev` to get our local instances up and running. Those aliases call different commands

The cached files are used for data that doesn't change frequently so that we can prevent some unneeded API requests. Some of them are also used in tests

hana-linode (commented):
Can we display general API errors as well? You can see this error by trying to create a VPC on the prod env

https://github.com/linode/manager/assets/115299789/41503561-b20e-4336-a2ff-90c53e9fc1da

packages/manager/src/features/VPC/VPCCreate/SubnetNode.tsx
```diff
@@ -0,0 +1,112 @@
...
+  subnet: SubnetFieldState;
+}
+
+const RESERVED_IP_NUMBER = 4;
+
+// TODO: VPC - currently only supports IPv4, must update when/if IPv6 is also supported
+export const SubnetNode = (props: Props) => {
+  const { disabled, idx, onChange, subnet, removable } = props;
+  const [availIps, setAvailIps] = React.useState<number | undefined>(undefined);
```
dwiley-akamai, line 32: I think throughout these files, the variable names are more readable if we capitalize the "IP" when it comes after the first word. What do you think? Granted, `availIPs` and `setAvailIPs` probably isn't the best example since lower-case L and upper-case I resemble each other 😅

A better example would be something like `calculateAvailableIPv4s`

packages/manager/src/features/VPC/VPCCreate/VPCCreate.tsx
```diff
@@ -1,11 +1,243 @@
...
+            />
+          </Paper>
+          <ActionsPanel
+            primaryButtonProps={{
+              'data-testid': 'submit',
+              label: 'Create VPC',
+              disabled: disabled,
+              loading: isLoading,
+              onClick: onCreateVPC,
+              // TODO: VPC - generate event on creation?
```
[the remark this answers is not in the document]
coliu-akamai: Good point, thanks! I'd had something like a toast notification in mind but wasn't sure if the API was behind the message for that

packages/manager/src/features/VPC/VPCCreate/SubnetNode.tsx
coliu-akamai: Agreed, will do!
in some functions/variables (I think especially for the offshoot branch, I camel-cased VPC and API as well - Vpc, Api) -- will update those as well!

packages/manager/src/features/VPC/VPCCreate/VPCCreate.tsx
```diff
@@ -1,11 +1,254 @@
...
+      <Grid>
+        {generalAPIError ? <Notice error text={generalAPIError} /> : null}
+        <form onSubmit={handleSubmit}>
+          <Paper>
+            <StyledHeaderTypography variant="h2">VPC</StyledHeaderTypography>
+            <StyledBodyTypography variant="body1">
+              A virtual private cloud (VPC) is an isolated network which allows
+              for control over how resources are networked and can communicate.
+              <Link to="#"> Learn more</Link>.
+              {/* TODO: VPC - learn more link here */}
```
tyler-akamai, line 189: Out of curiosity, do you know who is responsible for adding VPC documentation to the Linode docs?

tyler-akamai:
Do you use ESLint?

packages/manager/src/features/VPC/VPCCreate/VPCCreate.tsx
```diff
@@ -1,11 +1,254 @@
...
+  const { data: regions } = useRegionsQuery();
+  const { isLoading, mutateAsync: createVPC } = useCreateVPCMutation();
+  const [subnetErrorsFromAPI, setSubnetErrorsFromAPI] = React.useState<
+    APIError[]
+  >();
+  const [generalAPIError, setGeneralAPIError] = React.useState<
+    string | undefined
+  >();
+
+  const disabled = profile?.restricted && !grants?.global.add_vpcs;
```
tyler-akamai, line 48: When is the difference between profile restrictions and grant restrictions? Is it possible for `!grants?.global.add_vpcs` to be false and `profile?.restricted` to be true? If so, should this be an OR instead of an AND

tyler-akamai: From my understanding, I think `profile?.restricted` means that the user is restricted and the grants specify what sections is restricted. If so, if `!grants?.global.add_vpcs` is true, does that mean `profile?.restricted` is also true?

packages/manager/src/features/VPC/VPCCreate/SubnetNode.tsx
```diff
@@ -0,0 +1,108 @@
...
+        </Grid>
+        {isRemovable && !!idx && (
+          <Grid xs={1}>
+            <StyledButton onClick={removeSubnet}>
+              <Close data-testid={`delete-subnet-${idx}`} />
+            </StyledButton>
+          </Grid>
+        )}
+      </Grid>
+      <Grid xs={isRemovable ? 11 : 12}>
```
tyler-akamai, line 77: Would it make sense to add some helper text here that describes the format of the input? Will the customer know that it should be in cidr format? It might just help clarify. This can also be explained in the docs.

packages/manager/src/features/VPC/VPCCreate/VPCCreate.tsx
```diff
@@ -1,11 +1,254 @@
...
+              }
+              regions={regions ?? []}
+              isClearable
+              selectedID={values.region}
+            />
+            <TextField
+              disabled={disabled}
+              errorText={errors.label}
+              label="VPC label"
+              onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
```
tyler-akamai, line 205: This can be simplified to:
`onChange={handleChange}` like line 214

unless you like to specify the setFieldValue, if thats the case then maybe do the same with line 214 so everything is consistent.

tyler-akamai: However, Im not really sure how formik knows to update the description value on line 214 without manually calling setFieldValue. I did test `onChange={handleChange}` for both the description field and the label field. and it works. Just not sure how it knows to link each input to the right values option.

tyler-akamai: The only thing I can think of is this line:
`value={values.description}`
but i think thats moreso setting the value and not linking it to anything. 

But whatever, it works lol

packages/manager/src/features/VPC/VPCCreate/VPCCreate.tsx
```diff
@@ -1,11 +1,254 @@
...
+import { handleAPIErrors } from 'src/utilities/formikErrorUtils';
+import { SubnetFieldState, validateSubnets } from 'src/utilities/subnets';
+import { MultipleSubnetInput } from './MultipleSubnetInput';
+import scrollErrorIntoView from 'src/utilities/scrollErrorIntoView';
+import {
+  StyledBodyTypography,
+  StyledHeaderTypography,
+} from './VPCCreate.styles';
 
 const VPCCreate = () => {
```
tyler-akamai, line 34: Should we do a named export here instead of a default one at the bottom of the file?

coliu-akamai: VPCCreate is lazily imported in the index file, will keep this as is!

coliu-akamai:
> Do you use ESLint?

Believe so, think ESLint runs every time I commit files

packages/manager/src/features/VPC/VPCCreate/VPCCreate.tsx
coliu-akamai: ty! changed 214 to setFieldValue for consistency (esp. since that's needed for the subnets)

packages/manager/src/features/VPC/VPCCreate/SubnetNode.tsx
coliu-akamai: Errors for both this version and the other validation version will specify CIDR format is needed, but could be worth asking UX if helper text for the input is needed before any errors show up -- thoughts, @dwiley-akamai , @hana-linode?

TylerWJ: Gotcha, yeah the errors should be good then

packages/manager/src/features/VPC/VPCCreate/VPCCreate.tsx
dwiley-akamai: That's correct, `profile.restricted` indicates whether it's a restricted user, and the grants specify whether that ability is granted for the user or not. We typically check for both instead of just the grant to be explicit, but also in case an unexpected/incorrect combination was somehow returned.

packages/manager/src/features/VPC/VPCCreate/SubnetNode.tsx
dwiley-akamai: In other places, e.g. the "Manage Access Controls" drawer for Database instances, we have `192.0.2.1/32` as placeholder/guidance text so the expected format is clearer. Worth checking with UX in case a different example would be preferred here

coliu-akamai: added default value as per UX specifications

dwiley-akamai (commented):
Combed through the code more and did some testing; things look to be functioning well 👍 

Are you seeing the `sort-imports` suggestions on your end? If so, could you go through these files and do the "Fix all auto-fixable problems" option? We're trying to keep imports, props, etc. in alphabetical order as much as possible.

Screenshot 2023-08-21 at 1 23 51 PM

coliu-akamai:
@dwiley-akamai not sure why I'm not seeing the sort-imports and related suggestions, but I ran eslint on all the VPC related files which reorganized them all

packages/manager/src/features/VPC/VPCCreate/MultipleSubnetInput.tsx
```diff
@@ -0,0 +1,76 @@
...
+    } else {
+      newSubnets[subnetIdx] = subnet;
+    }
+    onChange(newSubnets);
+  };
+
+  return (
+    <Grid>
+      {subnets.map((subnet, subnetIdx) => (
+        <Grid key={`subnet-${subnetIdx}`}>
```
hana-linode, line 52: We want to avoid using index as a key since it's an anti-pattern

https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern/

packages/manager/src/features/VPC/VPCCreate/VPCCreate.tsx
```diff
@@ -1,11 +1,259 @@
...
+    setSubmitting,
+    values,
+  } = useFormik({
+    initialValues: {
+      description: '',
+      label: '',
+      region: '',
+      subnets: [
+        {
+          ip: { ipv4: DEFAULT_SUBNET_IPV4_VALUE, ipv4Error: '' },
```
hana-linode, line 139: I don't think we should set the subnet ip to any initial value since it's an optional field for creation. We can pass it as a placeholder prop though to the Textfield component

packages/manager/src/features/VPC/VPCCreate/VPCCreate.tsx
```diff
@@ -1,11 +1,259 @@
...
+        docsLabel="Getting Started"
+        docsLink="#" // TODO: VPC - add correct docs link
+        title="Create"
+      />
+      {disabled && (
+        <Notice
+          text={
+            "You don't have permissions to create a new VPC. Please contact an account administrator for details."
+          }
+          error={true}
```
hana-linode, line 178:
```suggestion
          error
```

packages/manager/src/features/VPC/VPCCreate/MultipleSubnetInput.tsx
dwiley-akamai: We could consider using `Math.random()` here -- maybe `Math.floor(Math.random() * 10000);`

dwiley-akamai (approved):
Great work! I'm in favor of rolling with the validation in this PR as opposed to the one on the other branch given the drawbacks of that other approach.

packages/manager/src/features/VPC/VPCCreate/VPCCreate.tsx
coliu-akamai: tbh I'm also more of a fan of the placeholder option, UX mentioned that we should offer a defaulted input instead of just ghost text 😅

hana-linode: What was their reasoning for defaulting the input over a placeholder?

packages/manager/src/features/VPC/VPCCreate/MultipleSubnetInput.tsx
coliu-akamai: Changing this is strangely causing a test to error, looking into it!

coliu-akamai: Ok so some strange things were happening when I started using Math.random(): I was able to get the test working, but then in the actual UI, I'd only be able to type one character at a time before losing focus on the subnet textfields. My guess is because each change to the state caused a rerender, which made the grids have different keys...

I also saw the bug that appears when using index as a key, due to the available IPs # being a state inside the SubnetNode component. It's a bit janky but to get around this I tried putting the availableIPs state into the formik field itself: it's now part of the SubnetFieldState type (see VPCCreate line 140, subnets.ts lines 7-13). Definitely see why using index as key is an anti-pattern, and it was super interesting to read about, I'm just not sure how to get around it with the issue above 😓

coliu-akamai:
@hana-linode nice catch, thanks! It's probably gonna be a change to line 114 on VPCCreate -- wasn't too sure what the api subnet error fields would look like before (but I've just now gotten the VPC capability and can actually create them instead of only relying on MSW :D  🎉 )

hana-linode (approved):
Awesome work! 🚀 🚀

coliu-akamai:
Unsure why the Banks-bot is saying this is not approved, but will be merging now since two approvals

PR #9538, opened 2023-08-11
Release v1.99.1 - staging → master

hana-linode:


cpathipa (approved):

jdamore-linode (approved):

jdamore-linode:
@hana-linode I completely overlooked the fact that PRs for long lived branches still run Cypress via GitHub Actions, and since that passed there's no reason to wait for the Jenkins run to finish as well (although it looks like it's about to finish too)

PR #9539, opened 2023-08-11
fix: Allow Cypress to make API requests against non-Prod environments

jdamore-linode:
There's more investigation and work that needs to be done here, but I'm opening this PR to unblock some developers on the Edge Experience team who are currently unable to run the tests.


jdamore-linode:
Closing -- Succeeded by #9631

PR #9540, opened 2023-08-11
refactor: [M3-7022] - Make `<ActionMenu />` use MUI instead of Reach UI

bnussman-akamai:
## Description 📝

- Reach UI duplicates MUI features so we might as well just use MUI. This will help us consolidate styles and packages.

## Major Changes 🔄
- Makes `ActionMenu` use MUI Menu instead of Reach UI 🎉
- Uninstalls all Reach UI packages except `@reach/tabs` because we still depend on it 😔

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-08-11 at 3 27 17 PM | Screenshot 2023-08-11 at 3 27 44 PM |

## How to test 🧪
- Verify `<ActionMenu />`'s still work
- Verify `<ActionMenu />` styles look alright and close to production
  - I tried my best to style it similarly to production, but if you have an suggestions to make it closer to production, push to my PR or leave a comment! 🎨
  - Please test both **light** and **dark** mode
- Verify `<ActionMenu />` is accessible


packages/manager/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -132,52 +74,96 @@ export const ActionMenu = React.memo((props: Props) => {
...
-                }
-              }}
-              data-qa-action-menu-item={a.title}
-              disabled={a.disabled}
-              key={idx}
-              valueText={a.title}
-            >
+        <KebabIcon />
+      </IconButton>
+      <Menu
```
jdamore-linode, line 106: This is a really subjective nitpick, so disregard if nobody agrees, but I wish the menu open transition was just a _little_ faster.

I was messing around with the `transitionDuration` prop and felt that a value between `175` and `225` felt slightly more snappy to me.

I hope it goes without saying that this is definitely not a priority from my point of view, just throwing it out there.

jdamore-linode (commented):
This is looking awesome! I've confirmed that interacting with the action menus via keyboard works well, that VoiceOver works well, and that touch interaction using the iOS simulator works well.

The only thing that is keeping me from approving is this behavior when opening/closing a menu when the body content is overflowing the viewport (i.e. a vertical scrollbar is present), where the scrollbar disappears and re-appears and causes the content to shift:

https://github.com/linode/manager/assets/97627410/d6d54260-8ae7-4b7d-be75-fbcd00f5611c

I've looked into this a bit but don't have a definitive solution, but I'm hoping that what I've found can help:

- This issue is exacerbated by a [CSS rule in `index.css`](https://github.com/linode/manager/blob/develop/packages/manager/src/index.css#L17) which forces the `body` padding to be 0. Without the `!important` keyword, MUI is able to apply 15px padding to the `body` to compensate for the scrollbar, preventing the content from shifting:

  https://github.com/linode/manager/assets/97627410/5d904a14-631e-4529-baab-be06c815abcc

  (I didn't test this change very thoroughly so I'm not sure if removing the `!important` keyword causes any other side effects or issues)

- Alternatively, we can disable the scroll locking behavior with the `<Menu />`'s `disableScrollLock` prop ([StackOverflow discussion here](https://stackoverflow.com/questions/69065717/material-ui-menu-component-locks-body-scrollbar)), but because the menu's position is `fixed`, scrolling with a menu open causes it to move along with the viewport. Changing the menu's position to `absolute` sort of resolved this, but introduced some other issues related to menu positioning (particularly when opening a menu at the bottom of the screen) so I'm not sure we want to go down this path

bnussman-akamai:
@jdamore-linode Looks like we may need to wait for https://github.com/mui/material-ui/pull/37773 to fix that scroll behavior 😣

packages/manager/src/components/ActionMenu/ActionMenu.tsx
bnussman-akamai: I agree 100%. The animation is nice but definitely makes the app feel a bit sluggish 🐌

jdamore-linode (approved):
I don't think the scrollbar thing is a big enough issue to stop this from being merged, especially knowing there's a good chance that MUI PR will allow us to set `disableScrollLock` to `true` in the future.

PR #9543, opened 2023-08-14
fix: Duplicate `key` error on Linodes Landing

bnussman-akamai:
## Description 📝

- Fixes the duplicate key error when on the Linodes Landing page (and NodeBalancer landing page)

## Major Changes 🔄
- Use **lodash** instead of **ramda** `tail()` function
- Clean up `IPAddress.tsx` while fixing the bug

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-08-14 at 11 18 33 AM | Screenshot 2023-08-14 at 11 18 19 AM |

## How to test 🧪
- Have at least 1 Linode 
- Go to `/linodes`
- Verify you don't see a console error about duplicate keys

### DIFF
--- a/packages/manager/.changeset/pr-9543-tech-stories-1692026704155.md
+++ b/packages/manager/.changeset/pr-9543-tech-stories-1692026704155.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Tech Stories
+---
+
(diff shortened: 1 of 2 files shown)


jdamore-linode (approved):
Nice! Confirmed that the error is gone on the Linodes landing page and NodeBalancer landing page.

PR #9544, opened 2023-08-14
feat: [M3-6690] - Beta details

CodyFinn-Akamai:
## Description 📝
Adds the `BetaDetails` and `BetaDetailsList` components to list out the betas relevant to a user.

## Preview 📷
image

## How to test 🧪
- Run `yarn test BetasLanding BetaDetailsList BetaDetails betaUtils`

### DIFF
--- a/packages/manager/.changeset/pr-9544-added-1692992268848.md
+++ b/packages/manager/.changeset/pr-9544-added-1692992268848.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Added
+---
+
+Showing beta details and allowing enrolling in a beta ([#9544](https://github.com/linode/manager/pull/9544))
--- a/packages/manager/.changeset/pr-9544-added-1692992268848.md
+++ b/packages/manager/.changeset/pr-9544-added-1692992268848.md
@@ -2,4 +2,4 @@
 "@linode/manager": Added
 ---
 
-Showing beta details and allowing enrolling in a beta ([#9544](https://github.com/linode/manager/pull/9544))
+Details for betas and ability to enroll in betas ([#9544](https://github.com/linode/manager/pull/9544))
--- a/packages/manager/src/features/Betas/BetaDetails.tsx
+++ b/packages/manager/src/features/Betas/BetaDetails.tsx
@@ -1,9 +1,11 @@
 import * as React from 'react';
+import Stack from '@mui/material/Stack';
 
 import { Beta } from '@linode/api-v4/lib/betas';
 import { AccountBeta } from '@linode/api-v4/lib/account';
-
-// import { Typography } from 'src/components/Typography';
+import { Typography } from 'src/components/Typography';
+import { DateTimeDisplay } from 'src/components/DateTimeDisplay';
+import { Button } from 'src/components/Button/Button';
 
 interface AccountBetaProps {
   beta: AccountBeta;
@@ -12,21 +14,42 @@ interface AccountBetaProps {
 interface BetaProps {
   beta: Beta;
 }
+
 function BetaDetails(props: BetaProps): React.ReactElement;
 function BetaDetails(props: AccountBetaProps) {
   const {
-    beta: { label, started, description, ended, more_info },
+    beta: { label, started, description, ended },
   } = props;
   const enrolled = props.beta?.enrolled;
+  const startDate = <DateTimeDisplay displayTime={false} value={started} />;
+  const endDate = ended ? (
+    <Typography>
+      End Date: <DateTimeDisplay displayTime={false} value={ended} />
+    </Typography>
+  ) : null;
+
   return (
-    <li>
-      <p>Label: {label}</p>
-      <p>Start Date: {started}</p>
-      <p>Description: {description}</p>
-      {ended ? <p>End Date: {ended}</p> : null}
-      {enrolled ? <p>Enrolled: {enrolled}</p> : null}
-      {more_info ? <p>More Info: {more_info}</p> : null}
-    </li>
+    <Stack
+      width="100%"
+      minHeight={66}
+      direction="row"
+      spacing={2}
+      justifyContent="space-between"
+    >
+      <Stack minHeight={66} spacing={2} width="100%">
+        <Stack direction="row" justifyContent="space-between">
+          <Typography>{label}</Typography>
+          <Stack direction="row" spacing="30px">
+            <Typography>Start Date: {startDate}</Typography>
+            {endDate}
+          </Stack>
+        </Stack>
+        <Typography>{description}</Typography>
+      </Stack>
+      <Stack minHeight={66} minWidth="111px">
+        {enrolled ? null : <Button buttonType="primary">More Info</Button>}
+      </Stack>
+    </Stack>
   );
 }
 
--- a/packages/manager/src/features/Betas/BetaDetails.tsx
+++ b/packages/manager/src/features/Betas/BetaDetails.tsx
@@ -60,7 +60,7 @@ function BetaDetails(props: AccountBetaProps) {
         {enrolled ? null : (
           <Button
             onClick={() => {
-              history.push(`/betas/${id}/signup`);
+              history.push('/betas/signup', { betaId: id });
             }}
             buttonType="primary"
           >
--- a/packages/manager/src/features/Betas/BetaDetails.tsx
+++ b/packages/manager/src/features/Betas/BetaDetails.tsx
@@ -9,19 +9,22 @@ import { DateTimeDisplay } from 'src/components/DateTimeDisplay';
 import { Link } from 'src/components/Link';
 import { Typography } from 'src/components/Typography';
 
-interface AccountBetaProps {
-  beta: AccountBeta;
+interface Props {
+  beta: AccountBeta | Beta;
 }
 
-interface BetaProps {
-  beta: Beta;
-}
-
-function BetaDetails(props: BetaProps): React.ReactElement;
-function BetaDetails(props: AccountBetaProps) {
+const BetaDetails = (props: Props) => {
   const history = useHistory();
+  let more_info = undefined;
+  let enrolled = undefined;
+  if ('more_info' in props.beta) {
+    more_info = props.beta.more_info;
+  }
+  if ('enrolled' in props.beta) {
+    enrolled = props.beta.enrolled;
+  }
   const {
-    beta: { description, ended, enrolled, id, label, more_info, started },
+    beta: { description, ended, id, label, started },
   } = props;
   const startDate = !enrolled ? (
     <Typography>
@@ -75,6 +78,6 @@ function BetaDetails(props: AccountBetaProps) {
       </Stack>
     </Stack>
   );
-}
+};
 
 export default BetaDetails;
--- a/packages/manager/src/features/Betas/BetaDetails.tsx
+++ b/packages/manager/src/features/Betas/BetaDetails.tsx
@@ -60,7 +60,7 @@ function BetaDetails(props: AccountBetaProps) {
         {enrolled ? null : (
           <Button
             onClick={() => {
-              history.push(`/betas/signup`, { betaId: id });
+              history.push(`/betas/${id}/signup`);
             }}
             buttonType="primary"
           >
--- a/packages/manager/src/features/Betas/BetaDetails.tsx
+++ b/packages/manager/src/features/Betas/BetaDetails.tsx
@@ -7,6 +7,7 @@ import { AccountBeta } from '@linode/api-v4/lib/account';
 import { Typography } from 'src/components/Typography';
 import { DateTimeDisplay } from 'src/components/DateTimeDisplay';
 import { Button } from 'src/components/Button/Button';
+import { Link } from 'src/components/Link';
 
 interface AccountBetaProps {
   beta: AccountBeta;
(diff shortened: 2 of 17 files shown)


packages/manager/src/features/Betas/BetaDetails.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+interface AccountBetaProps {
+  beta: AccountBeta;
+}
+
+interface BetaProps {
+  beta: Beta;
+}
+
+function BetaDetails(props: BetaProps): React.ReactElement;
```
dwiley-akamai, line 20: This seems like a new pattern, I see it's also in `betaUtils.ts`. What is the purpose/what are the benefits of it?

packages/manager/src/features/Betas/BetasLanding.tsx
```diff
@@ -1,28 +1,64 @@
...
+  const allBetasMerged = allBetas.reduce((acc, beta) => {
+    if (acc[beta.id]) {
+      acc[beta.id] = Object.assign(beta, acc[beta.id]);
+    } else {
+      acc[beta.id] = beta;
+    }
+    return acc;
+  }, {});
+
+  const { active, available, historical } = categorizeBetasByStatus(
```
bnussman-akamai, line 36: Can we delete things like categorizeBetasByStatus and use only API filters? If so, can we paginate each section?

packages/manager/src/features/Betas/BetaDetails.tsx
CodyFinn-Akamai: It can help keep type information consistent. For example If I use an intersection type then unless I explicitly determine that an object is a `Beta` or an `AccountBeta` will be `GenericBeta`. Mostly useful for keeping stuff straight during development via LSP. However I think using the intersection type here is fine so I will change that out.

packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -1405,21 +1405,38 @@ export const handlers = [
...
   }),
   rest.get('*/account/betas/:id', (req, res, ctx) => {
-    return res(ctx.json(accountBetaFactory.build({ id: req.params.id })));
+    if (req.params.id !== 'undefined') {
+      return res(ctx.json(accountBetaFactory.build({ id: req.params.id })));
+    }
+    return res(ctx.status(404));
   }),
-  rest.post('*/account/betas', (req, res, ctx) => {
+  rest.post('*/account/betas', (_req, res, ctx) => {
```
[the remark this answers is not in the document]
CodyFinn-Akamai: req works fine I added the underscore since it is not used and it makes the linter happier

packages/manager/src/features/Betas/BetasLanding.tsx
CodyFinn-Akamai: It is less requests to get them both in their entirety. Account betas are used to find both active and historical betas. And we will always show all available betas. the `/betas` endpoint only returns active betas. The only list at risk of getting very long is the historical but pagination will have to be a refinement

packages/manager/src/features/Betas/BetaDetails.tsx
CodyFinn-Akamai: Actually after trying to swap to the intersection type I was getting some type errors so I switched back to using the function overloading.

packages/manager/src/features/Betas/BetaDetailsList.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <Stack divider={<Divider spacingTop={20} spacingBottom={20} />}>
+        {betas.map((beta, index) => (
+          <BetaDetails beta={beta} key={`${index}-${beta.id}`} />
+        ))}
+      </Stack>
+    </Paper>
+  );
+};
+
+export default BetaDetailsList;
```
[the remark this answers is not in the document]
CodyFinn-Akamai: If I export on the same line then I have to change the imports. I can't export default on the same line as I declare the `const` can I?

packages/manager/src/features/Betas/BetaDetails.tsx
```diff
@@ -0,0 +1,80 @@
...
+          <Link external to={more_info}>
+            More Info
+          </Link>
+        ) : null}
+      </Stack>
+    </Stack>
+  );
+}
+
+export default BetaDetails;
```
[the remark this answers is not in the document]
CodyFinn-Akamai: If I export on the same line then I have to change the imports. I can't export default on the same line as I declare the `const` can I?

packages/manager/src/features/Betas/BetaDetails.tsx
```diff
@@ -0,0 +1,80 @@
...
+            buttonType="primary"
+          >
+            Sign Up
+          </Button>
+        )}
+        {more_info ? (
+          <Link external to={more_info}>
+            More Info
+          </Link>
+        ) : null}
```
dwiley-akamai, line 74: Based on the wireframes, we'll want "More Info" centered under the "Sign Up" button, with the external link icon to the right of the text

packages/manager/src/features/Betas/BetasLanding.tsx
```diff
@@ -1,30 +1,66 @@
...
-        </Paper>
+        <BetaDetailsList
+          betas={active}
+          title="Currently Enrolled Betas"
+          isLoading={areAccountBetasLoading}
+          errors={accountBetasErrors}
+        />
+        <BetaDetailsList
+          betas={available}
+          title="Available Betas"
```
dwiley-akamai, line 54: Should this be "Available & Upcoming Betas" based on the wireframes?

packages/manager/src/features/Betas/BetaDetails.tsx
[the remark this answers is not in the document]
dwiley-akamai: Correct, can't `export default` on the same line as the declaration. But it seems the `export default` is necessary due to the L20 pattern I asked about before so no worries.

packages/manager/src/features/Betas/BetaDetails.tsx
```diff
@@ -0,0 +1,80 @@
...
+            {endDate}
+          </Stack>
+        </Stack>
+        <Typography>{description}</Typography>
+      </Stack>
+      <Stack direction="column" minHeight={66} minWidth="111px" spacing={2}>
+        {enrolled ? null : (
+          <Button
+            onClick={() => {
+              history.push(`/betas/signup`, { betaId: id });
```
dwiley-akamai, line 63: This doesn't work for me, it just pushes to `/betas/signup` -- changing it to

```suggestion
              history.push(`/betas/${id}/signup`);
```

seems to work though (pushes to `betas/active_closed/signup` for example).

packages/manager/src/features/Betas/BetaDetailsList.tsx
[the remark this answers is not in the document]
dwiley-akamai: Looks like we can make the change in this file to `export const BetaDetailsList` since that pattern I asked about in the other file isn't used here and this file isn't being lazy loaded anywhere.

packages/manager/src/features/Betas/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { SuspenseLoader } from 'src/components/SuspenseLoader';
+
+const BetasLanding = React.lazy(() => import('./BetasLanding'));
+
+const BetaSignup = React.lazy(() => import('./BetaSignup'));
+
+const BetaRoutes = () => (
+  <React.Suspense fallback={<SuspenseLoader />}>
+    <Switch>
+      <Route component={BetaSignup} path="/betas/betaId:signup" />
```
dwiley-akamai, line 13: we'll need to change this to

```suggestion
      <Route component={BetaSignup} path="/betas/:betaId/signup" />
```

for the routing to work properly

dwiley-akamai (commented):
Nothing shows up on the Beta Signup page for me, I just get a loading spinner and then "Not Found" after a few seconds

packages/manager/src/features/Betas/BetaDetails.tsx
CodyFinn-Akamai: Kendall and I decided to keep it left aligned for now

bnussman-akamai (commented):
Should we show some kind of empty state?

> Screenshot 2023-08-25 at 4 40 23 PM

Instead of 3 independent loading spinners, I think it would look cleaner with just one. 
> Screenshot 2023-08-25 at 4 42 56 PM

I created a beta, and I see it coming from the API, but I don't see it in the UI... What am I missing?

> Screenshot 2023-08-25 at 4 48 45 PM

packages/manager/src/features/Betas/BetaDetailsList.tsx
```diff
@@ -0,0 +1,44 @@
...
+  props.isLoading ? <CircleProgress /> : null;
+
+export const BetaDetailsList = (props: Props) => {
+  const { title, betas, isLoading, errors } = props;
+  return (
+    <Paper>
+      <Typography variant="h2">{title}</Typography>
+      <Divider spacingBottom={20} spacingTop={20} />
+      <BetaError errors={errors} />
+      <BetaLoading isLoading={isLoading} />
```
bnussman-akamai, line 36: Unnecessary abstraction. 

```suggestion
      {isLoading && <CircleProgress />}
```

packages/manager/src/queries/accountBetas.ts
```diff
@@ -38,6 +38,7 @@ export const useCreateAccountBetaMutation = () => {
     {
       onSuccess() {
         queryClient.invalidateQueries([queryKey, 'paginated']);
+        queryClient.invalidateQueries(['regions', 'paginated']);
```
bnussman-akamai, line 41: Why do we need to invalidate regions?

bnussman-akamai (commented):
Signup functionality works now! 

Should the "More Info" page work?

https://github.com/linode/manager/assets/115251059/ca02a805-5264-411c-9eb5-d11917fa56c3

packages/manager/src/queries/accountBetas.ts
CodyFinn-Akamai: Yeah some new regions will be available for a beta period. So we need to invalidate the query to ensure that the user can see the region they signed up for a beta for. We don't have the information to do this on a more granular scale so for now at least we will invalidate the region query whenever a user signs up for a beta

packages/manager/src/features/Betas/BetaSignup.tsx
```diff
@@ -0,0 +1,92 @@
...
...e in quam.  Etiam vel tortor sodales tellus ultricies commodo.  Suspendisse potenti.  Aenean in sem ac leo mollis blandit.  Donec neque quam, dignissim in, mollis nec, sagittis eu, wisi.  Phasellus lacus.  Etiam laoreet quam sed arcu.  Phasellus at dui in ligula mollis ultricies.  Integer placerat tristique nisl.  Praesent augue.  Fusce commodo.  Vestibulum convallis, lorem a tempus semper, dui dui euismod elit, vitae placerat urna tortor vitae lacus.  Nullam libero mauris, consequat quis, varius et, dictum id, arcu.  Mauris mollis tincidunt felis.  Aliquam feugiat tellus ut neque.  Nulla facilisis, risus a rhoncus fermentum, tellus tellus lacinia purus, et dictum nunc justo sit amet elit. Aliquam erat volutpat.  Nunc eleifend leo vitae magna.  In id erat non orci commodo lobortis.  Proin neque massa, cursus ut, gravida ut, lobortis eget, lacus.  Sed diam.  Praesent fermentum tempor tellus.  Nullam tempus.  Mauris ac felis vel velit tristique imperdiet.  Donec at pede.  Etiam vel neque nec dui dignissim bibendum.  Vivamus id enim.  Phasellus neque orci, porta a, aliquet quis, semper a, massa.  Phasellus purus.  Pellentesque tristique imperdiet tortor.  Nam euismod tellus id erat.';
```
bnussman-akamai, line 18: Do we still need this placeholder copy?

packages/manager/src/features/Betas/BetaSignup.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const location = useLocation<{ betaId: string }>();
+  const history = useHistory();
+  const betaId = location?.state?.betaId;
```
bnussman-akamai, line 22: I think we should use path params just so we can deep link users to betas.

CodyFinn-Akamai: We don't have anything to display for a single beta. All the info is displayed on the landing page. Using the path params introduces the ability to navigate to a single beta via the breadcrumbs, but we don't have any design specs for a single beta details page.

packages/manager/.changeset/pr-9544-added-1692992268848.md
```diff
@@ -0,0 +1,5 @@
+---
+"@linode/manager": Added
+---
+
+Showing beta details and allowing enrolling in a beta ([#9544](https://github.com/linode/manager/pull/9544))
```
dwiley-akamai, line 5:
```suggestion
Details for betas and ability to enroll in betas ([#9544](https://github.com/linode/manager/pull/9544))
```

dwiley-akamai (approved):
Could you go through these files and sort the imports, etc. according to eslint's `perfectionist/sort-interfaces`? We have it enabled in the eslint config so it should pop up as a suggestion. You can hit all the instances in a file by hovering over one instance and going `Quick Fix` > `Fix all auto-fixable problems`

The rest LGTM, most of the remaining feedback/suggestions can be addressed in subsequent tickets, so approving pending the above

packages/manager/src/features/Betas/BetaSignup.tsx
CodyFinn-Akamai: Replaced with final copy

dwiley-akamai (approved):

bnussman-akamai (approved):
Basic functionality works! I feel like the general UI could be refined but approving based on functionality

PR #9545, opened 2023-08-14
test: [M3-6553] - Add e2e tests for Linode Rebuild

cliu-akamai:
## Description 📝
Add Cypress tests for rebuilding Linodes.

## Major Changes 🔄
- Add e2e tests to verify the flow of rebuilding a Linode.

## How to test 🧪
`yarn cy:run -s "cypress/e2e/core/linodes/rebuild-linode.spec.ts"`

### DIFF
--- a/packages/manager/cypress/e2e/core/linodes/rebuild-linode.spec.ts
+++ b/packages/manager/cypress/e2e/core/linodes/rebuild-linode.spec.ts
@@ -1,0 +1,201 @@
+import { createLinode } from 'support/api/linodes';
+import { containsClick, getClick } from 'support/helpers';
+import { apiMatcher } from 'support/util/intercepts';
+import { ui } from 'support/ui';
+import { randomString, randomLabel } from 'support/util/random';
+import { authenticate } from 'support/api/authentication';
+import { createStackScript } from '@linode/api-v4/lib';
+import { interceptGetStackScripts } from 'support/intercepts/stackscripts';
+import { createLinodeRequestFactory } from '@src/factories';
+import { cleanUp } from 'support/util/cleanup';
+
+const createStackScriptAndLinode = async (
+  stackScriptRequestPayload,
+  linodeRequestPayload
+) => {
+  return Promise.all([
+    createStackScript(stackScriptRequestPayload),
+    createLinode(linodeRequestPayload),
+  ]);
+};
+
+const checkPasswordComplexity = (rootPassword: string) => {
+  const weakPassword = '123';
+  const fairPassword = 'Akamai123';
+
+  // weak or fair root password cannot rebuild the linode
+  cy.get('[id="root-password"]').clear().type(weakPassword);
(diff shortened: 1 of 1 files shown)


packages/manager/cypress/e2e/core/linodes/rebuild-linode.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+      cy.get('[id="linode-label"]').type(linode.label);
+
+      // weak or fair root password cannot rebuild the linode
+      cy.get('[id="root-password"]').clear().type(weakPassword);
+      ui.button.findByTitle('Rebuild Linode').should('be.enabled').click();
+      cy.contains('Password does not meet complexity requirements.');
+
+      cy.get('[id="root-password"]').clear().type(fairPassword);
+      ui.button.findByTitle('Rebuild Linode').should('be.enabled').click();
+      cy.contains('Password does not meet complexity requirements.');
```
mjac0bs, line 62: Is this something worth creating a function for to avoid repetition? I can imagine that there could be other instances, even outside the rebuild flow, in which we'd also want to confirm that a weak or fair password results in an error notice for the user.

packages/manager/cypress/e2e/core/linodes/rebuild-linode.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+        containsClick('From Image').type('From Image{enter}');
+      });
+      cy.get('[data-qa-enhanced-select="Choose an image"]').within(() => {
+        containsClick('Choose an image').type(`${image}{enter}`);
+      });
+      cy.get('[id="linode-label"]').type(linode.label);
+      cy.get('[id="root-password"]').type(rootPassword);
+      ui.button.findByTitle('Rebuild Linode').should('be.enabled').click();
+      cy.wait('@linodeRebuild');
+      cy.contains('Linode busy.');
```
mjac0bs, line 212: This is a question for you and @jdamore-linode in general, but it applies here: in e2e tests, at what point do we test for invalid input + resulting errors in the form submission vs taking the happy path? For ex, these tests confirm that we can't rebuild a linode when it is busy and can't rebuild without a strong password, but we don't check that the form errors when a user attempts to rebuild without a stackscript selection or image selection.

mjac0bs (approved):
Thanks for adding test coverage! Tests passed locally. The unit test failure on this PR was fixed last week, and pulling in the latest develop would get that passing. 

Screenshot 2023-08-29 at 8 53 53 AM

Totally unrelated to this ticket, I noticed a bunch of UI styling inconsistencies when looking at the Rebuild dialog. I created M3-7064 so the team can address them in the future -- lots of issues with error notice padding, an inconsistently placed submit button, a missing error notice for lack of image selection when rebuilding from Image.

bnussman-akamai (approved):

PR #9547, opened 2023-08-15
fix: [M3-7030] - Creating a database redirects to new instance's details page

jdamore-linode:
## Description 📝
This fixes an issue where creating a DBaaS instance redirects the user to `/databases/:id` instead of `/databases/:engineType/:id`, causing them to see the DBaaS landing page instead of the new instance's details page.

Additionally, this updates the behavior of the landing page so that navigating to invalid paths, like `/databases/:id`, redirects back to `/databases`. I chose this behavior, rather than a _Not Found_ error, because the Images landing page also behaves this way and it was the closest parallel I could find to this DBaaS situation. Open to feedback here.

## Major Changes 🔄
- Creating a DBaaS instance redirects to `/databases/:engineType/:id` instead of `/databases/:id`
- Navigating to `/databases/:id` (or any other invalid path) redirects back to `/databases`
- Updated DBaaS create test for this change

## How to test 🧪
Observe results of the automated e2e tests to confirm that core DBaaS functionality remains in tact.

The following should also be confirmed:

- Navigating to `/databases/:id` redirects back to `/databases`
- Other DBaaS routes are not impacted (like `/databases/create`, `/databases/:engineType/:id`, etc.)

### DIFF
--- a/packages/manager/src/features/Databases/index.tsx
+++ b/packages/manager/src/features/Databases/index.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Route, Switch } from 'react-router-dom';
+import { Redirect, Route, Switch, useRouteMatch } from 'react-router-dom';
 
 import { DocumentTitleSegment } from 'src/components/DocumentTitle';
 import { SuspenseLoader } from 'src/components/SuspenseLoader';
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/Databases/index.tsx
```diff
@@ -9,16 +9,19 @@ const DatabaseDetail = React.lazy(() => import('./DatabaseDetail'));
...
-        <Route component={DatabaseCreate} path="/databases/create" />
+        <Route component={DatabaseLanding} exact path={path} />
+        <Route component={DatabaseCreate} path={`${path}/create`} />
         <Route
           component={DatabaseDetail}
-          path="/databases/:engine/:databaseId"
+          path={`${path}/:engine/:databaseId`}
         />
-        <Route component={DatabaseLanding} exact strict />
+        <Redirect to="/databases" />
```
jdamore-linode, line 24: I pretty much copied this pattern from the [Images routing](https://github.com/linode/manager/blob/develop/packages/manager/src/features/Images/index.tsx) since Images was the closest entity type I could find to DBaaS in terms of route structure, but I don't have a great understanding of how React Router works so would be very grateful for a second set of eyes here.

bnussman-akamai (approved):
Routing looks good, creation sent me to the correct page, and the e2e passed locally!

(Looks like Github webhooks are down)

hana-linode (approved):
Nice catch! 🔍 

If it's not too much effort, can we also fix the MSW redirect?
Screenshot 2023-08-15 at 3 24 49 PM

jdamore-linode:
> Nice catch! 🔍
> 
> If it's not too much effort, can we also fix the MSW redirect? [Screenshot 2023-08-15 at 3 24 49 PM](https://user-images.githubusercontent.com/115299789/260818733-5f500f56-1377-4257-b40b-d34ba1a459bf.jpg)

Thanks @hana-linode, I completely overlooked MSW! That should be fixed now (I think it may have been responding with a Linode object up til now)

jdamore-linode:
Going to merge this now because the 2 test failures (clone-volume.spec.ts and resize-volume.spec.ts) are unrelated to this PR.

PR #9550, opened 2023-08-16
fix: [M3:-7009] - Resizing Under the Reputation Score Provides Incorrect Error

tyler-akamai:
## Description 📝
When a customer wants to resize their linode to a plan that their reputation does not support, the page crashes. Instead, we should inform them to contact support.

This error is partly caused by the Axios intercepter which takes the error of type `string` and converts it into `TSX`. So I removed the catch.

I also added a Typography component that prints out the error message we receive from the api which is:
"Additional verification is required to add this service. Please contact support."

## Major Changes 🔄
- Removed the Axios interceptor
- Added an additional error message for "Additional verification" errors

## Preview 📷
| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/linode/manager/assets/139489745/58a93277-f71a-42dd-89ab-92a09c6f39e0" /> | <video src="https://github.com/linode/manager/assets/139489745/f38f4539-705e-4528-8c72-15a114cc352c" /> |

## How to test 🧪
1. Create a Linode 4GB.
2. Set Reputation to 20 in Admin Panel
3. Attempt to resize 16GB Dedicated.
4. When creating an Linode above rep. the customer is informed to contact support.


bnussman-akamai (commented):
The Axios interceptors are super sketchy because they break out type contract 😖. They make it so reason could be `string | JSX.Element` even though our type claims `reason` will be `string`

In general we need to find a better way to handle this through the app in my opinion. 

To this fix, could we just fix this by checking if `errorText` is a string before running `match`? This way, we can retain the `open a Support ticket` link.

Screenshot 2023-08-16 at 10 11 12 AM

tyler-akamai:
> The Axios interceptors are super sketchy because they break out type contract 😖. They make it so reason could be `string | JSX.Element` even though our type claims `reason` will be `string`
> 
> In general we need to find a better way to handle this through the app in my opinion.
> 
> To this fix, could we just fix this by checking if `errorText` is a string before running `match`? This way, we can retain the `open a Support ticket` link.
> 
> Screenshot 2023-08-16 at 10 11 12 AM

Good catch, just fixed it!

packages/manager/src/features/Linodes/LinodesDetail/LinodeResize/LinodeResize.tsx
```diff
@@ -313,6 +317,12 @@ const getError = (error: APIError[] | null) => {
       </Typography>
     );
   }
+  if (
+    typeof errorText === 'string' &&
+    errorText.match(/Additional verification/i)
+  ) {
+    return <VerificationError title="Verification Request" />;
+  }
```
bnussman-akamai, line 325: I think we can delete this because the Axios interceptor will automatically do this.

```suggestion
  if (
    typeof errorText === 'string' &&
    errorText.match(/Additional verification/i)
  ) {
    return <VerificationError title="Verification Request" />;
  }
```

tyler-akamai: Good catch!

hana-linode (approved):
🚀

packages/manager/src/features/Linodes/LinodesDetail/LinodeResize/LinodeResize.tsx
```diff
@@ -300,7 +300,10 @@ const getError = (error: APIError[] | null) => {
   }
 
   const errorText = error?.[0]?.reason;
-  if (errorText.match(/allocated more disk/i)) {
+  if (
+    typeof errorText === 'string' &&
```
bnussman-akamai, line 304: Maybe we should add a comment explaining why we have to check if the erroText is a string.

bnussman-akamai (approved):
🚀

PR #9552, opened 2023-08-16
fix: Reduce Cypress test flake, August 2023

jdamore-linode:
## Description 📝
Various small changes and fixes intended to reduce flakiness in Cloud Manager's Cypress tests.

This includes some improvements for Volumes tests, but I'm still observing a lot of (I think) API-related issues (like Volumes taking a long time to get into "active" status, lots of 500 responses when performing actions, and potentially not always sending events when a Volume goes from "creating" status to "active" status). I think these improvements will make our Volume tests better but I don't think they will solve the issues altogether.

## Major Changes 🔄
- Fixes an issue with the OBJ `shows error upon object upload failure` test where Cypress would occasionally trigger a mouse action on an element before Cloud has rendered the expected error, causing the tests to fail
- Fixes an issue with the OBJ `can upload, access, and delete objects` test where Cypress would occasionally fail when selecting the bucket's label because it's rendered in multiple places on the page
- Improves Volume clone and resize tests by waiting for the affected Volumes to be `active` before performing actions (see notes above)
- Refactors the DBaaS create tests to use more up-to-date patterns, adds assertions after navigating to the created instance's details page
- Restores some clean up in places where I had removed it, and improves it in other places, preventing accumulation of Images, Object Storage buckets, and LKE clusters on test accounts which has contributed to some recent failures
- Attempts to speed up the StackScript test by resizing the disk from 25GB to 1.5GB before using it to create an Image

## How to test 🧪
We can rely on the automated tests for this, but the tests can be run locally with this command:

```bash
yarn cy:run
```

Alternatively, to run only the modified tests, use:

```bash
yarn cy:run -s "cypress/e2e/core/account/default-payment-method.spec.ts,cypress/e2e/core/databases/create-database.spec.ts,cypress/e2e/core/objectStorage/object-storage-errors.spec.ts,cypress/e2e/core/objectStorage/object-storage.e2e.spec.ts,cypress/e2e/core/volumes/clone-volume.spec.ts,cypress/e2e/core/volumes/resize-volume.spec.ts"
```


packages/manager/cypress/e2e/core/images/machine-image-upload.spec.ts
```diff
@@ -134,13 +135,17 @@ const uploadImage = (label: string) => {
...
+  });
+
   /*
    * - Confirms update and delete UI flows using mock API data.
    * - Confirms that image label can be updated.
    * - Confirms that image description can be updated.
    * - Confirms that image can be deleted.
    */
-  it('updates and deletes a machine image', () => {
+  it.only('updates and deletes a machine image', () => {
```
bnussman-akamai, line 148: Do we want to keep this `.only()`?

packages/manager/cypress/e2e/core/volumes/clone-volume.spec.ts
```diff
@@ -14,6 +17,19 @@ const pageSizeOverride = {
...
 };
 
+/**
+ * Creates a Volume and waits for it to become active.
+ *
+ * @param volumeRequest - Volume create request payload.
+ *
+ * @returns Promise that resolves to created Volume.
+ */
+const createActiveVolume = async (volumeRequest: VolumeRequestPayload) => {
```
bnussman-akamai, line 27: What is the reason for this approach rather than waiting for the UI to show that the volume is active? Can we not lead on the events polling and the UI that is already in place?

packages/manager/cypress/e2e/core/images/machine-image-upload.spec.ts
jdamore-linode: Nope! Thanks for catching this!

packages/manager/cypress/e2e/core/volumes/clone-volume.spec.ts
jdamore-linode: Yeah, I actually attempted that first! I need to dig into this more and maybe write a ticket, but I was finding that when I tried waiting for "active" in the table row, the tests would occasionally time out because the landing page wasn't always updating to reflect that the volume has finished being created (and I reproduced this outside of Cypress, too).

From what I could tell, Cloud never received an event indicating that the Volume had finished being created, but I didn't dig into this super thoroughly once I ruled out the tests as the cause. I did have multiple Cloud tabs open, so before I go any further I want to rule out the possibility that, for example, I had one tab open and the event endpoint was polled and the event received in that tab, and meanwhile Cypress polled the event endpoint a few seconds later and the event wasn't present since it'd already been read, etc.

bnussman-akamai (approved):

packages/manager/cypress/e2e/core/volumes/resize-volume.spec.ts
```diff
@@ -14,6 +17,21 @@ const pageSizeOverride = {
...
+ * @returns Promise that resolves to created Volume.
+ */
+export const createActiveVolume = async (
+  volumeRequest: VolumeRequestPayload
+) => {
+  const volume = await createVolume(volumeRequest);
+  await pollVolumeStatus(volume.id, 'active', new SimpleBackoffMethod(10000));
+  return volume;
+};
+
```
coliu-akamai, line 34: super small nitpick - since this is already being exported, could we just have `clone-volume.spec.ts` import this function rather than write it again? or put this in its own file and have both files import it?

coliu-akamai (approved):
tests passed on local! 🚀

packages/manager/cypress/e2e/core/volumes/resize-volume.spec.ts
jdamore-linode: Ooh, nice catch Connie! I didn't actually mean to export it (muscle memory I guess 😅), but now that you mention it this might be better served as a util or something. I'll circle back to this tomorrow!

jdamore-linode: @coliu-akamai Just due to time constraints I opted to stop exporting this and otherwise leave everything as-is, but I'm planning to write a ticket (or multiple tickets?) to revisit all of these helpers and try to move them to a place that makes more sense.

Thanks again Connie!

PR #9558, opened 2023-08-17
fix: New LKE clusters get stuck provisioning when HA price is undefined

jdamore-linode:
## Description 📝
This fixes an issue where creating an LKE Cluster in environments where `REACT_APP_LKE_HIGH_AVAILABILITY_PRICE` is undefined results in a cluster with nodes stuck in "provisioning" state, and the only way to get them un-stuck is to enable high availability, which is irreversible. The issue is fixed by passing `false` for `control_plane.high_availability` in cases where the environment variable is undefined.

I don't think this is strictly a Cloud Manager issue because the API does not respond with a `400`, and the [API docs](https://www.linode.com/docs/api/linode-kubernetes-engine-lke/#kubernetes-cluster-create) state that `control_plane.high_availability` is treated as `false` by default. Regardless, explicitly passing `false` results in clusters that provision successfully, and leaving it absent seems to result in clusters that get stuck.

## Major Changes 🔄
- Pass `false` to `control_plane.high_availability` explicitly when the HA control plane prompt is not present

### To Reproduce the Issue
1. Check out `develop`, remove the `REACT_APP_LKE_HIGH_AVAILABILITY_PRICE` environment variable from your `.env` file if necessary, and build Cloud Manager
2. Create an LKE cluster, observe that you're redirected to the cluster's details page
3. Wait ~20 minutes and confirm that the Cluster's nodes are still in "provisioning" state

### To Verify These Changes
1. Check out this branch, remove the `REACT_APP_LKE_HIGH_AVAILABILITY_PRICE` environment variable from your `.env` file if necessary, and build Cloud Manager
2. Create an LKE cluster, observe that you're redirected to the cluster's details page
3. Confirm that the cluster finishes provisioning after a few minutes
5. Re-add the `REACT_APP_LKE_HIGH_AVAILABILITY_PRICE` environment variable, create an LKE cluster with HA enabled and an LKE cluster with HA disabled, and confirm that both clusters finish provisioning after a few minutes


jdamore-linode:
I don't think we need to worry about fitting this into the 1.100 release since this issue does not impact `cloud.linode.com`.

bnussman-akamai (approved):

TylerWJ:
It's still taking me <10min for the provisioning an HA cluster and its not done yet. It might be on my end but not too sure.

PR #9560, opened 2023-08-17
feat: [UIE-6576] - Dbaas total disk size and used disk size

corya-akamai:
## Description 📝
Adding two new fields to display (1) the total disk size available for the database (Selected size - OS overhead) and (2) the used disk space so the customer will now how much space is remaining in the DB. BE will be feature gated, so UI will fallback to current storage field if the API response does not contain the total_disk_space

## Major Changes 🔄
- Field total_disk_space_gb
- Field used_disk_space_gb
- database_low_disk_space_remaining event

## Preview 📷
Screenshot 2023-08-02 at 1 53 30 PM

## How to test 🧪
1. Use the mock service workers
2. Navigate to http://localhost:3000/databases/mysql/0
3. Verify Total Disk Size and Used fields are present. Modify packages > manager > src > factories > databases.ts > total_disk_size_gb to undefined to revert to previous view with the Storage field.
4. E2E test will be added from QA sujai


corya-akamai:
Reopening after rebasing https://github.com/linode/manager/pull/9638

PR #9565, opened 2023-08-18
refactor: [M3-6264] - MUI v5 Migration - SRC > Features > Events

tyler-akamai:
## Description 📝
MUI v5 CSS migration for SRC > Features > Events

## Major Changes 🔄
- Eliminated default exports
- Eliminated makeStyles
- Renamed props
- Created .styles files

## Preview 📷
| Before  | After   |
| --- | --- |
| Screenshot 2023-08-22 at 3 05 02 PM | Screenshot 2023-08-22 at 3 03 44 PM |

## How to test 🧪
1. Navigate to http://localhost:3000/events
2. Ensure the styling is consistent


packages/manager/src/features/Events/EventRow.tsx
```diff
@@ -10,38 +8,22 @@ import { Hidden } from 'src/components/Hidden';
...
-interface Props {
+interface EventRowProps {
   entityId?: number;
   event: ExtendedEvent;
 }
 
-type CombinedProps = Props;
-
-export const EventRow: React.FC<CombinedProps> = (props) => {
+export const EventRow = RenderGuard((props: EventRowProps) => {
```
coliu-akamai, line 26: We may be able to get rid of the RenderGuard -- I vaguely remember asking @bnussman-akamai about this before -- tagging Banks to double check!

packages/manager/src/features/Events/EventsRow.styles.ts
```diff
@@ -0,0 +1,23 @@
+import { styled } from '@mui/material/styles';
```
coliu-akamai, line 1: Could we rename this file to `EventRow.styles.ts` so that it matches the EventRow.tsx naming

packages/manager/src/features/Events/EventsRow.styles.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { TableRow } from 'src/components/TableRow';
+
+import { GravatarByUsername } from '../../components/GravatarByUsername';
+
+export const StyledGravatar = styled(GravatarByUsername, {
+  label: 'StyledGravatar',
+})({
+  height: 24,
+  width: 24,
```
coliu-akamai, line 11: could potentially take in `theme` as an argument to this styled component, and use `theme.spacing(3)` instead of 24 to avoid hardcoding values -- but either's probably fine

packages/manager/src/features/Events/EventsRow.styles.ts
```diff
@@ -0,0 +1,23 @@
...
+});
+
+export const StyledTableRow = styled(TableRow, {
+  label: 'StyledTableRow',
+})(({ theme }) => ({
+  row: {
+    '&:hover': {
+      backgroundColor:
+        theme.name === 'light' ? '#fbfbfb' : 'rgba(0, 0, 0, 0.1)',
+    },
```
coliu-akamai, line 21: Might have accidentally included the `row:` here (but also I didn't really notice any styling changes when hovering on the rows so maybe it doesn't matter?)

Also, is `#fbfbfb` one of the theme colors? If so we can use that instead of hardcoding

coliu-akamai (commented):
good work! just had a few comments

packages/manager/src/features/Events/EventsRow.styles.ts
tyler-akamai: Good catch!

packages/manager/src/features/Events/EventsRow.styles.ts
tyler-akamai: Yeah, the I got rid of the hover styling because the default hovering looks good and is consistent with the rest of TableRows.

coliu-akamai (approved):

cpathipa (approved):

packages/manager/src/features/Events/EventRow.tsx
bnussman-akamai: Yeah, I think we should remove it here!

PR #9567, opened 2023-08-18
fix: Make feature flag dev tools show reality

bnussman-akamai:
## Description 📝
- Makes it so the dev tools flag checkboxes actually represent which flags are on/off in real life
  - They have been very confusing and misleading in certain circumstances creating a confusing developer experience.
- Also, thoughts on reverting #9400?
  - I'd prefer to have my local app match exactly what launchdarkly has by default, does anyone else feel this way?

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-08-18 at 1 31 10 PM | Screenshot 2023-08-18 at 1 29 44 PM

## How to test 🧪
- Clear local storage and make sure the dev tools show feature flags accurately

### DIFF
--- a/packages/manager/src/dev-tools/FeatureFlagTool.tsx
+++ b/packages/manager/src/dev-tools/FeatureFlagTool.tsx
@@ -2,12 +2,12 @@ import Grid from '@mui/material/Unstable_Grid2';
 import * as React from 'react';
 import { useDispatch } from 'react-redux';
 
+import withFeatureFlagProvider from 'src/containers/withFeatureFlagProvider.container';
 import { FlagSet, Flags } from 'src/featureFlags';
 import { Dispatch } from 'src/hooks/types';
 import { useFlags } from 'src/hooks/useFlags';
 import { setMockFeatureFlags } from 'src/store/mockFeatureFlags';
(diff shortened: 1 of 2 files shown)


abailly-akamai (approved):
:1 nice thx for fixing but really against reverting https://github.com/linode/manager/pull/9400, the developer experience was improved and the tradeoff is you keeping track of your flags.

However, if important to you, we could add a "Don't store flags" checkbox to trigger logic in the useEffect to keep LDs defaults and prevent setting flags in local storage so we really only store that `noFlags` option in storage. This way we can both be happy :)

packages/manager/src/dev-tools/FeatureFlagTool.tsx
dwiley-akamai, line 1: While we're here we might as well get rid of L16 to more accurately reflect what's actually feature flagged currently

dwiley-akamai (approved):
LGTM -- I'm in favor of adding a "Don't store flags" toggle as suggested by Alban to accommodate both preferences.

bnussman-akamai:
Good idea about `Don't store flags`. I'm going to save that for another PR and merge this as-is.

PR #9568, opened 2023-08-18
feat: [M3-6969] - Add DC-specific pricing to Kubernetes HA

mjac0bs:
## Description 📝
High Availability (HA) pricing ($60.00/month per cluster for all DCs) is controlled through an environment variable in CI. This PR moves base pricing to a constant in the codebase and allows us to dynamically calculate HA pricing for new DCs.

## Major Changes 🔄
- With the feature flag on:
   - On the Kubernetes Create page, summary/pricing data and the Create Cluster button do not appear in the Checkout Summary bar until a Region selection has been made. This will avoid displaying inaccurate pricing in the checkout bar, since a user could enable HA before choosing a Region. 
   - Prices on enable HA Control Plane radio button on Kubernetes Create page reflect DC-specific HA prices.
      - When no Region is selected, no pricing information is shown.
   - Prices on enable Update to HA modal on Kubernetes Details page reflect DC-specific HA prices.
   - Prices on the Kube Summary section of the Kubernetes Details page reflect DC-specific HA prices.
- HA pricing relies on the pricing multiplier constant (`LKE_HA_PRICE`) rather than the `REACT_APP_LKE_HIGH_AVAILABILITY_PRICE` env var.
- Unit tests are updated and the HA selection that had been previously commented out in e2e tests was restored with the use of a constant for HA. 
- Updates Sao Paulo's price increase percentage. 

## Preview 📷
| Location  | Preview   |
| ------- | ------- |
| Create Cluster | <video src="https://github.com/linode/manager/assets/114685994/ba15ca1b-df84-4d19-baa5-7a38b6aba09b"/> |
| Create Cluster - No Region | Screenshot 2023-08-20 at 3 30 43 PM |
| Kube Cluster Specs | Screenshot 2023-08-20 at 3 32 16 PM |
| Kube Upgrade Cluster Dialog | Screenshot 2023-08-20 at 3 37 20 PM |

## How to test 🧪
- Check out code locally and enable the DC-Specific Pricing feature flag in the dev settings.

2. **How to verify changes?**
- Verify the e2es and unit tests pass.
- Test the following flows with both the flag on and off. Test Sao Paulo, Jakarta and any region without a price increase.
- When DC-specific pricing flag is *off*, the flows should match production. When *on*, the flows should show the major changes described above and visible in the preview section. 

Flows:
- http://localhost:3000/kubernetes/create
   - Without selecting a Region, toggle between HA selections. Confirm no pricing is displayed in the Checkout Summary.
   - Selecting a Region, confirm the pricing is displayed in the Checkout Summary. Test all DC-specific pricing regions.
   - Select a Plan and confirm the total is correct. (note this total)
   - *Do not* enable HA. (select 'No')
   - Create Cluster.
- http://localhost:3000/kubernetes/clusters/{id}/summary
   - Confirm the total price in the Kube Summary (top of the page) is accurate, matching the total from the Create page.
   - Click the Upgrade to HA button and confirm the copy and the HA price on the dialog is correct. 
   - Submit the form to upgrade to HA and confirm the total price in the Kube Summary has been updated to reflect the addition of HA.

```
yarn test HAControlPlane KubeCheckoutBar kubeUtils
```

```
yarn cy:run -s "cypress/e2e/core/kubernetes/smoke-lke-create.spec.ts cypress/e2e/core/kubernetes/lke-create.spec.ts"
```


packages/manager/src/utilities/pricing/dynamicPricing.ts
```diff
@@ -16,7 +16,7 @@ export interface DataCenterPricingOptions {
    * The `id` of the region we intended to get the price for.
    * @example us-east
    */
-  regionId: Region['id'];
+  regionId?: Region['id'] | undefined;
```
abailly-akamai, line 19: FYI: This is going to break existing or in progress work relying on a required prop. Besides, the `regionId` should always be required since the util is useless without it

abailly-akamai: I see @bnussman made the right change without making the prop optional 👍

packages/manager/src/features/Kubernetes/CreateCluster/HAControlPlane.tsx
```diff
@@ -49,16 +58,18 @@ export const HAControlPlane = (props: Props) => {
       >
         <Typography variant="inherit">HA Control Plane</Typography>
       </FormLabel>
-      <HACopy />
+      <HACopy isDCSpecificPricing={!!flags.dcSpecificPricing} />
```
bnussman-akamai, line 61: Does `isDCSpecificPricing` need to be a prop? Should we calculate this value in the component itself?

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeClusterSpecs.tsx
```diff
@@ -73,6 +78,14 @@ export const KubeClusterSpecs = (props: Props) => {
       pools ?? [],
       types ?? [],
       cluster.control_plane.high_availability
+        ? parseFloat(
+            getDCSpecificPrice({
```
bnussman-akamai, line 82: Hm. I wonder if there is anything we should do about `getDCSpecificPrice` returning a string. This works fine but feels wrong.

bnussman-akamai (approved):
Nice to see the env var go away 👋🏼

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeClusterSpecs.tsx
abailly-akamai: I had a similar situation - IMO I think this could be a case of consolidating price components and what type they expect once we get prices straight from the API, and once we have time to get to the nitty gritty

abailly-akamai (commented):
Code looks good to me!

However I wanted to point to a difference in the UX that bothers me:

Before, adding a cluster in the create flow would 
1. disable the add button and reset the field to 0 
2. add the subtotal in the panel

Now, in the new flow, we only get item 1 and it's a pretty odd experience because one would expect something else to happen> i think it is confusing to the user. I know we still have the "Select a Region and Plan to create a cluster." sentence in there but it's super easy to miss and wonder if something is broken. 

IMO it would make more sense to still add the items to the sidebar but not display the prices and the sentence for each item instead, OR at least adding a warning icon or something next to the empty sidebar sentence to make sure the user does not miss it.

Thoughts?

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeClusterSpecs.tsx
mjac0bs: Yeah, I'm okay with leaving this for now and revisiting to do some clean up here later. I'll make a note/ticket about this.

mjac0bs:
> it would make more sense to still add the items to the sidebar but not display the prices and the sentence for each item instead, OR at least adding a warning icon or something next to the empty sidebar sentence to make sure the user does not miss it.

@abailly-akamai - Yeah, this is fair. I've reached out to Matthew and Kendall for some feedback on this/overall UI changes when Region is not selected and are awaiting their input. I think either approach you suggested would work and improve the UX from what it is now. 

Here's my current thought: 

https://github.com/linode/manager/assets/114685994/fa838c13-5525-4550-ab38-4ae792cef4bb

mjac0bs:
@abailly-akamai This is ready for re-review with a UX sign off on these changes from Kendall.

Changes:

In the checkout summary, without a Region selection, the total price is replaced with a message indicate Region and Node Pool selections are needed. Until Region is selected, users can still add Pools, but they won’t see pricing for it under each pool.

Although the radio button label is long, this provides transparency and reiterates the need to make a Region selection. Removed the "Prices my vary based on Region" from the helper text above the options.
Screenshot 2023-08-21 at 3 29 25 PM

At UX request, added HA as a line item to the Cluster Summary -- this was slightly out of scope of this PR, but an improvement upon #9489, so this change is not feature flagged. Also updated the helper text copy slightly to more accurately instruct the user to add a Node Pool.
Screenshot 2023-08-21 at 3 30 01 PM

abailly-akamai (approved):
Looks great thx for the changes!

PR #9576, opened 2023-08-22
feat: [M3-7025] - AGLB Details - Certificates Tab

bnussman-akamai:
## Description 📝
- Adds the AGLB certificate tab to the AGLB details page 🎉
  - What I built out is similar to the mockup but I left out some table columns that would require a lot of client-side computation for now

## Preview 📷
Screenshot 2023-08-21 at 8 30 45 PM

## How to test 🧪
- Turn on the MSW
- Navigate to `http://localhost:3000/loadbalancers/1/certificates`
- Test basic functionality
  - Note that sorting and filtering won't work because the MSW does not handle API filters


abailly-akamai (approved):
LGTM 👍 just one comment to clarify something

Also, unrelated to this PR but the browser tab seems to only show the entity ID rather than `{entity}-id`

mjac0bs (commented):
Certificate tab looks on point with the mocks. Do we have a ticket for adding the action menu functionality? I see we have M3-7029 for the upload certification drawer, but I didn't see anything for an edit or delete modal -- this is true of the mocks, too, so maybe UX hasn't gotten to those UI details yet.

bnussman-akamai:
I will be creating new tickets soon! @mjac0bs

mjac0bs (approved):

PR #9592, opened 2023-08-24
fix: [M3-7055] - Metadata Image select dropdown fixes

hana-linode:
## Description 📝
We got feedback that the Image Select dropdown is less dense than it used to be, which increases the amount of scrolling needed to find a distribution. It is also alarming to see the cloud-init incompatible icon on most distribution images since they do not support it yet.

This PR addresses the above by making each select item dense again and also swaps the logic in how we display the cloud-init icon so that the icon only displays on compatible images instead.

## Major Changes 🔄
- Fix select item row height
- Cloud-init incompatible icon -> Cloud-init compatible icon
    - Only show icon for compatible images instead of incompatible images

## Preview 📷
| Before  | After   |
| ------- | ------- |
| before | image |

## How to test 🧪
Go to `/linodes/create` and check the distribution image dropdown


packages/manager/src/components/ImageSelect/ImageOption.tsx
```diff
@@ -93,7 +93,13 @@ const sxCloudInitTooltipIcon = {
     height: 20,
     width: 20,
   },
+  '&:hover': {
+    color: 'white',
```
mjac0bs, line 97: We have an edge-case issue of a seemingly vanishing icon in light mode when the user has selected a cloud-init supported distro and then goes back into the select menu and hovers over the icon. 

https://github.com/linode/manager/assets/114685994/97873d76-12a4-4f9a-8fe6-f270e94b41b6

mjac0bs (commented):
This looks much better overall! Just an edge-case to fix styling for.

> It is also alarming to see the cloud-init incompatible icon on most distribution images since they do not support it yet.

Fair, though wasn't the thought that they were going to support it by the time metadata was released?

cpathipa (approved):
Overall, LGTM! except for the hover icon issue.  Approving it with the pending fix for the hover icon.

hana-linode:
> Fair, though wasn't the thought that they were going to support it by the time metadata was released?

@mjac0bs Yeah, that was the original thought but now we don't know when all the distros will be supported

PR #9594, opened 2023-08-25
feat: [M3-6754] - [on hold] Create new react list component

coliu-akamai:
turns out I was working off the wrong designs 😢 
brb + ON HOLD until decision meeting about new component vs using existing multi select

## Description 📝
- created simple scrollable list 
- the majority of the logic would really be in the custom list items -- I have a version where all of this is one component, on [this branch](https://github.com/linode/manager/compare/develop...coliu-akamai:manager:feat-m3-6754) that I was first working on, before feeling like this wasn't as clean as an approach that I'd hoped.

## Major Changes 🔄
- Scrollable list component
- custom list items 

## Preview 📷
(this would be used for the unassign linodes drawer)
image

(this would be used for the assign linodes drawer)
image
image

## How to test 🧪
`yarn storybook` --> ScrollableList

### DIFF
--- a/packages/manager/.changeset/pr-9594-upcoming-features-1692971598640.md
+++ b/packages/manager/.changeset/pr-9594-upcoming-features-1692971598640.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Upcoming Features
+---
+
+Create scrollable list component and custom list items ([#9594](https://github.com/linode/manager/pull/9594))
--- a/packages/manager/src/components/ScrollableList/CustomListItems/CheckedLinodeListItem.tsx
+++ b/packages/manager/src/components/ScrollableList/CustomListItems/CheckedLinodeListItem.tsx
@@ -4,8 +4,9 @@ import * as React from 'react';
 
 import { Checkbox } from 'src/components/Checkbox';
 import { ListItem } from 'src/components/ListItem';
-import type { ListItemProps } from 'src/components/ListItem';
 import { ListItemText } from 'src/components/ListItemText';
+
+import type { ListItemProps } from 'src/components/ListItem';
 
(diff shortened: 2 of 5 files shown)


abailly-akamai (changes requested):
As per my comments, the API is a little too open IMO, and makes an "everything is possible" component, which I'd like to flag. I wonder about a component that accepts "deletableItems" & "selectableItems" props and contain that logic as well. 

Also, why are we not using the multiSelect component (or a variation of it at least) which achieves all this functionality? (ex: https://cloud.linode.com/firewalls/create)

coliu-akamai:
@abailly-akamai Thanks for the feedback, and good point about the multiSelect component -- I'll ask UX about using this rather than having a new list-like component, especially if the multiSelect already has all the functionality needed. Will work on the other feedback soon too!

coliu-akamai:
@abailly-akamai spoke with UX, one of the main reasons to not use the multi-select component was due to how the selected linodes appeared at the top -- if there are too many linodes selected, it could get cluttered. Depending on a linode label's length, multiple linodes might appear on the same line of the box at the top:

image

Might be worth looking into this component further to see if it's possible for each linode label tag to be on a separate line

abailly-akamai:
@coliu-akamai thanks for checking with UX - I think there's a disconnect between UX requirements and what we're trying to achieve here. We shouldn't have to build a whole new series of components an achieve existing functionality with a UI variation

coliu-akamai:
Closing based on 8/31 Frontend Cafe discussion as we won't be moving forward with this approach -- we will either go with [Jaalah's autocomplete](https://github.com/linode/manager/pull/9497) (which allows for multiselect) or the existing table component (similar to what's done in the Server Transfers page).

PR #9595, opened 2023-08-25
test: [M3-6551] - Add Community Stackscripts integration tests

cliu-akamai:
## Description 📝
Add integration tests for Community Stackscripts.

## Major Changes 🔄
- Add test to check StackScripts are listed with expected info.
- Add test to check Pagination works as expected.
- Add test to check Search works as expected
- Add test to check Deploy flow works as expected

## How to test 🧪
`yarn cy:run -s "cypress/e2e/core/stackscripts/smoke-community-stackscripts.spec.ts"`

### DIFF
--- a/packages/manager/cypress/e2e/core/stackscripts/smoke-community-stackscrips.spec.ts
+++ b/packages/manager/cypress/e2e/core/stackscripts/smoke-community-stackscrips.spec.ts
@@ -255,6 +255,7 @@ describe('Community Stackscripts integration tests', () => {
     const rootPassword = randomString(16);
     const image = 'AlmaLinux 9';
     const region = chooseRegion();
+    const linodeLabel = randomLabel();
 
     interceptGetStackScripts().as('getStackScripts');
     cy.visitWithLogin('/stackscripts/community');
@@ -277,7 +278,15 @@ describe('Community Stackscripts integration tests', () => {
         );
       });
 
-    cy.visitWithLogin('/stackscripts/community');
+    ui.nav.findItemByTitle('StackScripts').should('be.visible').click();
+
+    ui.tabList
+      .findTabByTitle('Community StackScripts')
+      .should('be.visible')
+      .click();
+
+    cy.url().should('endWith', '/stackscripts/community');
+
     cy.get(`[href="/stackscripts/${stackScriptId}"]`)
       .should('be.visible')
       .click();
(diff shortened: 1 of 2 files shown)


abailly-akamai (approved):
This is REALLY cool. thanks for writing this! 

I would wait for @jdamore-linode's approval before merging but you get my vote 👍

packages/manager/cypress/e2e/core/stackscripts/smoke-community-stackscrips.spec.ts
```diff
@@ -0,0 +1,363 @@
...
+    mockGetStackScripts(mockStackScripts).as('getStackScripts');
+    cy.visitWithLogin('/stackscripts/community');
+    cy.wait('@getStackScripts');
+
+    cy.get('[data-qa-stackscript-empty-msg="true"]').should('not.exist');
+    cy.findByText('Automate deployment scripts').should('not.exist');
+
+    const updatedTime = stackScript.updated
+      ? DateTime.fromISO(stackScript.updated)
+      : DateTime.fromISO('1970-01-01T00:00:00.000');
```
jdamore-linode, line 107: I'll circle back to this first thing tomorrow morning, but on my account specifically the tests are tripping up because this time does not equal what's shown on screen.

I did confirm it's working on other accounts, so I'm assuming this has to do with the timezone set in the Profile display settings at `/profile/display`

packages/manager/cypress/e2e/core/stackscripts/smoke-community-stackscrips.spec.ts
```diff
@@ -0,0 +1,363 @@
...
+    cy.wait('@getStackScripts');
+
+    cy.get('[data-qa-stackscript-empty-msg="true"]').should('not.exist');
+    cy.findByText('Automate deployment scripts').should('not.exist');
+
+    cy.get('tr').then((value) => {
+      const rowCount = Cypress.$(value).length - 1; // Remove the table title row
+
+      interceptGetStackScripts().as('getStackScripts1');
+      cy.scrollTo(0, 500);
```
jdamore-linode, line 183: Nice solution! I was curious how difficult it would be to test the infinite scrolling pagination on this page

jdamore-linode:
@cliu-akamai Even with those changes, it's still expecting to find the incorrect date via my test account -- I think the reason is that the timezone needs to be applied when we call `formatDate()` which I think is imported but isn't used in the spec (and actually you might be able to skip `DateTime.fromISO()` altogether since I think `formatDate()` accepts an ISO formatted string anyway).

Dealing with the timezone adjusted dates in the UI is always a pain, let me know if I can help out!

jdamore-linode (changes requested):
Hey Cassie! Just noticing there's still a failure here in CI and wanted to mark this with "Request changes" just so it doesn't get merged.

I'll catch up with you momentarily!

jdamore-linode (approved):

PR #9603, opened 2023-08-29
chore: Update Material UI and Emotion

bnussman-akamai:
## Description 📝
- Updates MUI and Emotion to `latest`
- Updating MUI fixes the issues where scrolling with a Menu open would not move the menu correctly. This allows us to re-enable scrolling while action menus are open.
  - https://github.com/linode/manager/pull/9540#issuecomment-1677665164
  - https://github.com/mui/material-ui/pull/37773

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Notice how you **can't** scroll 😣 | Notice how you **can** scroll and the action menu remains in the correct position 🎉 |
| <video src="https://github.com/linode/manager/assets/115251059/95ed7908-7c3e-40df-af90-908e511a7b44" /> | <video src="https://github.com/linode/manager/assets/115251059/5ed655de-e01b-4d67-bb04-52d901d343cd" /> |

## How to test 🧪
- Checkout the PR
- Run `yarn` for good measure
- Run `yarn dev` and verify the app works and looks normal

### DIFF
--- a/packages/manager/.changeset/pr-9603-tech-stories-1693333265033.md
+++ b/packages/manager/.changeset/pr-9603-tech-stories-1693333265033.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Tech Stories
+---
+
+Update Material UI and Emotion ([#9603](https://github.com/linode/manager/pull/9603))
--- a/packages/manager/package.json
+++ b/packages/manager/package.json
@@ -13,13 +13,13 @@
     "url": "https://github.com/Linode/manager.git"
   },
   "dependencies": {
-    "@emotion/react": "^11.10.4",
-    "@emotion/styled": "^11.10.4",
+    "@emotion/react": "^11.11.1",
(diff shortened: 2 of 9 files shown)


jdamore-linode (approved):
Nice! Can't believe how quickly that MUI fix got released.

A couple observations, but nothing that I think requires any action:

- The menus without scroll lock don't seem to play very nicely with MacOS scroll elasticity/bouncing, but I haven't looked too closely at it and I kind of doubt there's much we can do about it
- The user menu still locks scrolling -- is that intentional or would we want to use `disableScrollLock` here too now that we can? I don't have any opinion from a UX perspective

cpathipa (approved):

bnussman-akamai:
I'll be sure to follow up about the other Menus causing scroll lock @jdamore-linode.

The MacOS scroll bounce is unfortunate. I'll keep an eye out for fixes for that.

PR #9605, opened 2023-08-30
fix: Fix VPC landing page test import

jdamore-linode:
## Description 📝
This is a tiny change to fix the TypeScript error we're seeing when running `vpc-landing-page.spec.ts`. It looks like we did a find+replace of `vpc` to `vpcs` which also inadvertently touched the import.

## How to test 🧪
Run `yarn && yarn build && yarn start:manager:ci`, and then run:

```bash
yarn cy:run -s "cypress/e2e/core/vpc/vpc-landing-page.spec.ts"
```

### DIFF
--- a/packages/manager/cypress/e2e/core/vpc/vpc-landing-page.spec.ts
+++ b/packages/manager/cypress/e2e/core/vpc/vpc-landing-page.spec.ts
@@ -7,7 +7,7 @@ import {
   mockGetVPCs,
   mockDeleteVPC,
   mockUpdateVPC,
-} from 'support/intercepts/vpcs';
+} from 'support/intercepts/vpc';
 import { vpcFactory } from '@src/factories';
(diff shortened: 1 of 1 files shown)


coliu-akamai (commented):
Since we've renamed the vpc folder in src/features to be `src/features/VPCs` and updated routing to be `/vpcs`, would it make more sense to rename the file to `vpcs.ts` and change the name of the e2e vpc test folder to `vpcs` as well?

bnussman-akamai (approved):

coliu-akamai (approved):
🎉

jdamore-linode:
Going to merge this even though the tests haven't run. No job got kicked off so the alternative would be to push some dummy change to try to trigger a job, and I don't think it's really worth the effort when the scope of this fix is so small and clear, and there's no risk of introducing any user-facing issues here

PR #9608, opened 2023-08-30
feat: [M3-7019] - Create add Nodebalancer to Firewall drawer

tyler-akamai:
## Description 📝
Added the "Add NodeBalancer to Firewall Drawer"

## Major Changes 🔄
- Added "Add Nodebalancer to Firewall" Drawer
- Added a Notice to the top of the drawer - documentation link is still needed
- Updated the text at the top of the landing page for both Linodes and NodeBalancers
- Added toast notifications

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screenshot 2023-08-30 at 2 42 19 PM | <video src="https://github.com/linode/manager/assets/139489745/92ade8b7-fdfa-47ec-9cfd-df9038e62f66" /> |

## How to test 🧪
1. Navigate to the Firewalls landing page
3. Click on any of your active firewalls
4. Navigate to the Nodebalancers tab
5. Ensure when you add a Nodebalancer, a toast notification appears and the number next to 'Nodebalancer' also updates (don't worry about the table, thats my next ticket)
6. To test errors
   1. Create two test firewalls (firewall-test-1, firewall-test-2)
   2. Assign a nodebalancer and a linode to a firewall-test-1
   3. Try to assign the same linode and nodebalancer to firewal-test-2
   4. For both drawers an error notice should appear with the link to the nodebalancer/linode landing page
   5. Ensure that the toast notifications are correctly displayed
   6. Ensure that the errors are removed when closing and reopening the tab


jdamore-linode:
@tyler-akamai Pulling in the latest changes from `develop` should (🤞) fix the cause of the test failures on this PR

tyler-akamai:
> @tyler-akamai Pulling in the latest changes from `develop` should (🤞) fix the cause of the test failures on this PR

Awesome! Thank you for the update

packages/manager/src/features/NodeBalancers/NodeBalancerSelect.tsx
```diff
@@ -38,51 +92,125 @@ export const NodeBalancerSelect = (props: Props) => {
...
     '+order_by': 'label',
   });
-  const nodebalancers = data?.pages.flatMap((page) => page.data);
 
-  const options = nodebalancers?.map(({ id, label }) => ({ id, label }));
+  React.useEffect(() => {
+    if (hasNextPage && !isLoading) {
+      fetchNextPage();
+    }
+  }, [hasNextPage, isLoading, fetchNextPage]);
```
bnussman-akamai, line 100: This defeats the purpose of the infinite scroll. 

We should restore the `ListboxProps.onScroll` if we want infinite scrolling in this Select.

packages/manager/src/features/Firewalls/FirewallDetail/Devices/AddDeviceDrawer.tsx
bnussman-akamai, line 1: Instead of having this one drawer handle both Linodes and NodeBalancers, I would support breaking it into two different drawers. Completely up to you though!

packages/manager/src/features/NodeBalancers/NodeBalancerSelect.tsx
```diff
@@ -38,51 +92,125 @@ export const NodeBalancerSelect = (props: Props) => {
...
-      options={options ?? []}
-      value={selectedNodebalancer}
-    />
+    <>
+      <Notice variant={'warning'}>
+        Only the Firewall's inbound rules apply to NodeBalancers. Any existing
+        outbound rules won't be applied.
+        {/* add documentation link */}
+        <Link to="#"> Learn more.</Link>
+      </Notice>
```
bnussman-akamai, line 125: The notice should probably be in the Drawer component, not in the Select component. 

This way, our select component is more versatile.

tyler-akamai: Good catch!

packages/manager/src/features/NodeBalancers/NodeBalancerSelect.tsx
tyler-akamai: Yeah, the only reason I am using the useInfiniteNodebalancersQuery is because the normal useAllNodeBalancersQuery has a comment above it saying 'Please don't use'. To get around this I just used the infinite query and grabbed all the data at once. The Linode Select also selects all of them by using the useAllLinodeQuery. So I was trying to keep them consistent and also not use useAllNodeBalancersQuery.

but if its better to do infinite scrolling in this case, I can add it.

tyler-akamai: I did update it to add infinite scrolling though.

carrillo-erik (approved):

carrillo-erik:
Changes look good to me and I was able to successfully add NodeBalancers to an existing Firewall.
Screenshot 2023-09-05 at 4 24 45 PM

Screenshot 2023-09-05 at 4 25 20 PM


We do have conflicting changes in our PRs with regards to the `Autocomplete` component for this project. Your approach expands the current implementation which essentially is just another copy of `LinodeSelect`. In my opinion and in 
retrospect, the `LinodeSelect` component should have been the component we expanded. Of course this is probably all moot since we are going to adopt the new version of the `Autocomplete` component upon approval.

tyler-akamai:
> Changes look good to me and I was able to successfully add NodeBalancers to an existing Firewall. Screenshot 2023-09-05 at 4 24 45 PM
> 
> Screenshot 2023-09-05 at 4 25 20 PM
> 
> We do have conflicting changes in our PRs with regards to the `Autocomplete` component for this project. Your approach expands the current implementation which essentially is just another copy of `LinodeSelect`. In my opinion and in retrospect, the `LinodeSelect` component should have been the component we expanded. Of course this is probably all moot since we are going to adopt the new version of the `Autocomplete` component upon approval.

Yeah I kept going back and fourth between changing the LinodeSelect to DeviceSelect but found that there was a lot of conditional logic plus there was a lot of code specific to Linodes that we didnt need for nodebalancers. becuase of that I decided to just create nodebalancer select. But to your point, this will eventually, i believe, get replaced by the Autocomplete component.

packages/manager/src/features/Linodes/LinodeSelect/LinodeSelect.tsx
```diff
@@ -126,8 +126,10 @@ export const LinodeSelect = (
       }
       onChange={(_, value) =>
         multiple && Array.isArray(value)
-          ? onSelectionChange(value)
-          : !multiple && !Array.isArray(value) && onSelectionChange(value)
+          ? onSelectionChange(value as Linode[])
+          : !multiple &&
+            !Array.isArray(value) &&
+            onSelectionChange(value as Linode)
```
bnussman-akamai, line 132: What's going on with the types here? How is `value` getting the type `NodeBalancer`? We should fix this so that it is only `Linode`

packages/manager/src/features/Firewalls/FirewallDetail/Devices/AddNodebalancerDrawer.tsx
```diff
@@ -0,0 +1,181 @@
...
+        {/* add documentation link */}
+        <Link to="#"> Learn more.</Link>
+      </Notice>
+      <form
+        onSubmit={(e: React.ChangeEvent<HTMLFormElement>) => {
+          e.preventDefault();
+          handleSubmit();
+        }}
+      >
+        {errorMessage ? errorNotice(errorMessage) : null}
```
bnussman-akamai, line 147: I feel like this could be cleaned up. `errorNotice` shouldn't really need paramaters. 

```suggestion
        {errorMessage ? errorNotice(errorMessage) : null}
```

bnussman-akamai (commented):
I think we should make it so the drawer does not close if a general error occours.

https://github.com/linode/manager/assets/115251059/d5984355-fd0b-49b3-a8c7-10ac2c1398fd

packages/manager/src/features/Firewalls/FirewallDetail/Devices/AddLinodeDrawer.tsx
```diff
@@ -49,18 +50,35 @@ export const AddDeviceDrawer = (props: Props) => {
...
+
+    results.forEach((result, _) => {
+      if (result.status === 'fulfilled') {
+        // Assuming the response contains the device label, replace with the appropriate property if not.
+        const label = result.value.entity.label;
+        enqueueSnackbar(`${label} added successfully.`, { variant: 'success' });
+      } else {
+        hasError = true;
+        // Assuming the error object contains the device label, replace with the appropriate property if not.
+        const errorLabel = result.reason.label;
```
bnussman-akamai, line 73: This is not how API errors work. `result.reason.label` is `undefined`. API errors only return errors if errors have occurred, so you'll have to get the labels from somewhere else.

Screenshot 2023-09-07 at 9 37 00 AM

jaalah-akamai:
Yea let's hold on this until the newer Autocomplete is done instead of duplicating the `LinodeSelect`. Looking back we should not have added any data fetching within the `LinodeSelect` and that should have been at the parent level.

packages/manager/src/features/Firewalls/FirewallDetail/Devices/AddLinodeDrawer.tsx
tyler-akamai: Should be fixed, since the api firewall-nodebalancer changes are in production now, you have to create an error by assigning the same linode and nodebalancer to two different firewalls. As before, no matter what you did you got: `Adding firewall to a nodebalancer is not currently available`

packages/manager/src/features/Firewalls/FirewallDetail/Devices/AddLinodeDrawer.tsx
```diff
@@ -0,0 +1,233 @@
...
+  React.useEffect(() => {
+    if (linodeError) {
+      setLocalError('Could not load Linode Data');
+    }
+  }, [linodeError]);
+
+  const linodes = data?.filter(optionsFilter);
+
+  const options =
+    linodes?.map((linode) => ({ id: linode.id, label: linode.label })) || [];
```
bnussman-akamai, line 172: We don't need to do this map at all. Lets just pass `options={linodes}` to the `Autocomplete`.

```suggestion
  const options =
    linodes?.map((linode) => ({ id: linode.id, label: linode.label })) || [];
```

packages/manager/src/features/Firewalls/FirewallDetail/Devices/AddLinodeDrawer.tsx
```diff
@@ -0,0 +1,233 @@
...
+    );
+
+    const mappedLinodes: Linode[] = Array.isArray(result)
+      ? result
+      : result
+      ? [result]
+      : [];
+
+    setSelectedLinodes(mappedLinodes);
+  };
```
bnussman-akamai, line 187: This function seems overly complex. We should be able to make it as simple as 

```suggestion
  const onChange = (selectedLinodes: Linode[]) => {
    setSelectedLinodes(selectedLinodes);
  };
```

packages/manager/src/features/Firewalls/FirewallDetail/Devices/AddLinodeDrawer.tsx
```diff
@@ -0,0 +1,233 @@
...
+      title={`Add Linode to Firewall: ${firewall?.label}`}
+    >
+      <form
+        onSubmit={(e: React.ChangeEvent<HTMLFormElement>) => {
+          e.preventDefault();
+          handleSubmit();
+        }}
+      >
+        {localError ? errorNotice() : null}
+        <Autocomplete<{ id: number; label: string }, true>
```
bnussman-akamai, line 206: Let's avoid passing generics here. We can use the power of Typescripts inference to our advantage here. `Autocomplete` will infer it's type from what you pass as `options`. I think we should be passing `Linode[]` as the `options`.

packages/manager/src/features/Firewalls/FirewallDetail/Devices/AddLinodeDrawer.tsx
```diff
@@ -0,0 +1,233 @@
...
+          disabled={currentDevicesLoading || linodeIsLoading}
+          helperText={helperText}
+          isOptionEqualToValue={(option, value) => option.id == value.id}
```
bnussman-akamai, line 209: We probably don't need an `isOptionEqualToValue`. The autocomplete should be able to figure it out by itself

bnussman-akamai (approved):
> Looking back we should not have added any data fetching within the LinodeSelect and that should have been at the parent level.

I'm not sure I agree with this, but I don't want to block this ticket because of this because functionality looks good! 🎉 We can always come back and make changes later. 

<hr />

Can we update the table column label?
Screenshot 2023-09-11 at 4 05 47 PM

PR #9614, opened 2023-08-30
fix: Fix E2E failures caused by inadequate resource clean-up

jdamore-linode:
## Description 📝
Another test resource clean-up PR! This stems from the changes made in #9529. This should fix some of the failures we've recently seen in these specs:

- `smoke-create-image.spec.ts`
- `backup-linode.spec.ts`
- `longview.spec.ts` 

In all 3 cases, the test failures are due to the tests creating Linodes but failing to clean up existing Linodes first.

## How to test 🧪
We can rely on the automated test run to confirm that these 3 specs pass, indicating that these changes haven't broken the tests. However, manual testing may be required to confirm that the clean up actually functions as expected:

1. Create some test Linodes whose labels start with `cy-test-`
2. Navigate to the Linodes landing page in your browser, and stay there
3. Separately run one of the spec files above, and observe in your browser that the Linodes you created in step 1 get cleaned up automatically by the test
4. Repeat steps 1-3 for the other 2 spec files

### DIFF
--- a/packages/manager/cypress/e2e/core/images/smoke-create-image.spec.ts
+++ b/packages/manager/cypress/e2e/core/images/smoke-create-image.spec.ts
@@ -1,11 +1,13 @@
-import type { Image, Linode, Disk } from '@linode/api-v4/types';
+import type { Linode, Disk } from '@linode/api-v4/types';
 import { imageFactory } from 'src/factories/images';
+import { authenticate } from 'support/api/authentication';
 import { createLinode, deleteLinodeById } from 'support/api/linodes';
 import {
   mockCreateImage,
   mockGetCustomImages,
 } from 'support/intercepts/images';
 import { mockGetLinodeDisks } from 'support/intercepts/linodes';
+import { cleanUp } from 'support/util/cleanup';
 import { randomLabel, randomNumber, randomPhrase } from 'support/util/random';
 
 const diskLabel = 'Debian 10 Disk';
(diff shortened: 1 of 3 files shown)


bnussman-akamai (approved):

coliu-akamai (approved):
tests cleaned up the linodes before they ran 🎉 

One weird thing I noticed when running `backup-linode.spec.ts` was that although the test deletes the linode it created at the end, the linode still shows up for a bit in the landing page (but since it's already deleted, the delete modal won't work):
image

The already deleted linode eventually disappears from the landing page after a minute or so -- is this just the frontend lagging behind?

jdamore-linode:
> tests cleaned up the linodes before they ran 🎉
> 
> One weird thing I noticed when running `backup-linode.spec.ts` was that although the test deletes the linode it created at the end, the linode still shows up for a bit in the landing page (but since it's already deleted, the delete modal won't work): <screenshot>
> 
> The already deleted linode eventually disappears from the landing page after a minute or so -- is this just the frontend lagging behind?

Oh interesting! I tried a time or two to reproduce this and surprisingly wasn't able to, but I think I might have an idea of what's going on there because I think I've seen similar behavior on the Volumes landing page.

I think what may have happened was that the Cloud Manager page running via Cypress polled the API events endpoint and received the event for the Linode that was just deleted, and meanwhile a second or two later the Cloud Manager page running in your browser polled the same events endpoint, and since the delete event had already been consumed, it wasn't included in the response and so Cloud Manager in your browser basically had no way of knowing that that Linode was deleted. Admittedly I probably have the least solid understanding of how our events polling works, but I think this is a possibility given my (limited) understanding.

coliu-akamai:
I read in the docs that react query has a mechanism to clear data from its cache after a while and refetch the data -- maybe that's going on here? Haven't looked into it too deeply for here yet, but that seems like it could be it

PR #9618, opened 2023-08-31
feat: Basic Select Component

bnussman-akamai:
## Description 📝

- Adds a new Select component

## Why ❓

- So we don't have to always use `<EnhancedSelect />`
- So we don't have to use `<TextField select />` to achieve a basic select
  -  This will allow us to better refine the `<TextField />` implimentation
- Gives us a simple and accessible select component

## Preview 📷
Screenshot 2023-08-31 at 5 27 30 PM

## How to test 🧪
- Test the component in Storybook **or** on `http://localhost:3000/loadbalancers/0/certificates` with the MSW on

### DIFF
--- a/packages/manager/.changeset/pr-9618-added-1693517370679.md
+++ b/packages/manager/.changeset/pr-9618-added-1693517370679.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Added
+---
+
+New Basic Select Component ([#9618](https://github.com/linode/manager/pull/9618))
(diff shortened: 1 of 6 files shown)


abailly-akamai (approved):
I don't know how I feel about this. There should be more conversation about the need for this and yet introducing a new abstracted MUI component that (albeit being smaller) duplicates existing functionality. I don't see a ticket for this and we also need to concert with UI (this select styles also doesn't seem to match our design system).

What is the drawback of using enhanced Select? I understand the Textfield select pattern not being ideal, but we need to have some conversations before adding yet another select component.

I'd like to at least wait for @jaalah to be back before merging this all together.

PR #9623, opened 2023-09-01
fix: Payment Drawer Style Regression from MUI Update 

bnussman-akamai:
## Description 📝

- Caused by https://github.com/linode/manager/pull/9603

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-09-01 at 10 43 15 AM | Screenshot 2023-09-01 at 10 43 32 AM |

## How to test 🧪
- Make sure `/account/billing/make-payment` looks good 🎨

### DIFF
--- a/packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/PaymentDrawer.tsx
+++ b/packages/manager/src/features/Billing/BillingPanels/BillingSummary/PaymentDrawer/PaymentDrawer.tsx
@@ -1,8 +1,8 @@
 import { PaymentMethod } from '@linode/api-v4';
 import { makePayment } from '@linode/api-v4/lib/account';
 import { APIWarning } from '@linode/api-v4/lib/types';
+import Stack from '@mui/material/Stack';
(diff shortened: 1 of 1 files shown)


tyler-akamai:
nitpick: if you want, this can be simplified to:

`paymentMethods?: PaymentMethod[];`

on line 64: https://github.com/linode/manager/pull/9623/files#diff-8a21e23fcac02097f86824bdd156bc44d90851c4c159b1503c66cf78316ebacaL64

tyler-akamai:
For the warning component: https://github.com/linode/manager/pull/9623/files#diff-8a21e23fcac02097f86824bdd156bc44d90851c4c159b1503c66cf78316ebacaL381

Would it make sense for the api team to provide us error codes and use those codes instead of using the text in the error message since those are susceptible to change?

I know this would be a bigger project, but it might be better practice.

^ Again, you didnt touch this code, just curious.

tyler-akamai:
Looks great!

jdamore-linode (approved):

PR #9626, opened 2023-09-01
feat: [M3-7016] - Add Basic Filtering for Firewall Devices

tyler-akamai:
## Description 📝
Added device filtering for the Linode and Nodebalancer firewall tabs.

## Major Changes 🔄
- Added device filtering using the `DebouncedSearchTextField` component

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-09-11 at 6 16 05 PM | Screenshot 2023-09-11 at 6 13 56 PM |
| | Screenshot 2023-09-11 at 6 15 08 PM |

## How to test 🧪
1. Navigate to http://localhost:3000/firewalls
2. Click on any of the firewall
4. Navigate to the linodes tab
5. Make sure to add linodes to the firewall
6. Ensure the filter works properly
7. Navigate to the nodebalancers tab
8. Make sure to add nodebalancers to the firewall
9. Ensure the filter works properly


packages/manager/src/components/TextField.tsx
```diff
@@ -303,22 +303,24 @@ export const TextField = (props: TextFieldProps) => {
...
+            })}
+            data-qa-textfield-label={label}
+            htmlFor={validInputId}
+          >
+            {label}
+            {required ? (
+              <span className={classes.label}> (required)</span>
+            ) : optional ? (
+              <span className={classes.label}> (optional)</span>
+            ) : null}
```
cpathipa, line 321: Suggestion: 

```
const labelContent = required ? "(required)" : optional ? "(optional)" : null;
labelContent && <span className={classes.label}>{labelContent}</span>;

```

packages/manager/src/features/Firewalls/FirewallDetail/Devices/FirewallDeviceLanding.tsx
```diff
@@ -68,31 +86,49 @@ export const FirewallDeviceLanding = React.memo(
...
+            </Grid>
+            <Grid>
+              <Button
+                buttonType="primary"
+                data-testid="add-device-button"
+                disabled={disabled}
+                onClick={() => setDeviceDrawerOpen(true)}
+              >
+                Add {formattedType}s to Firewall
+              </Button>
```
cpathipa, line 122: We could use `ActionsPanel` here to avoid variations for primary actions button.

tyler-akamai: it seems like the ActionsPanel is also dependent on the primary actions button. Let me know if I misinterpreted your comment.

cpathipa: @tyler-akamai Yes, you are right. My intention is to use the ActionsPanel component to render the Primary button instead of having isolated references to the primary button, unless the ActionsPanel doesn't meet our requirements. This approach would make future maintenance easier.

abailly-akamai (commented):
@tyler-akamai i know this is a draft, but just heads up this PR is modifying two core components and you shouldn't have to do that. 

- Removing a required label prop isn't recommended
- you already have a `hideLabel` prop on the textfield component

tyler-akamai:
> @tyler-akamai i know this is a draft, but just heads up this PR is modifying two core components and you shouldn't have to do that.
> 
> * Removing a required label prop isn't recommended
> * you already have a `hideLabel` prop on the textfield component

yes, great catch.

packages/manager/src/features/Firewalls/FirewallDetail/Devices/FirewallDeviceLanding.tsx
tyler-akamai: I checked with the team on this. They said that for one off selections like "Add a Nodebalancer/Linode" we should use the `Button` component. We should only use the ActionsPanel component if there are multiple actions to select from, or we are in a modal or drawer.

tyler-akamai: I added a ActionsPanel storybook component and added this clarification to the documentation page.

packages/manager/src/components/TextField.tsx
tyler-akamai: Looking back, I think I might remove this.

I added a screenshot, I think the green looks a little cleaner than the red.

Screenshot 2023-09-11 at 6 49 32 PM

packages/manager/src/components/DebouncedSearchTextField/DebouncedSearchTextField.tsx
```diff
@@ -13,7 +13,6 @@ interface DebouncedSearchProps extends TextFieldProps {
   defaultValue?: string;
   hideLabel?: boolean;
   isSearching?: boolean;
-  label: string;
```
abailly-akamai, line 16: Per my previous comment, we don't want to remove a required prop. This change would break accessibility features. Instead, use the `hideLabel` prop available on `<DebouncedSearchTextField />`

packages/manager/src/features/Firewalls/FirewallDetail/Devices/FirewallDeviceLanding.tsx
```diff
@@ -73,35 +90,51 @@ export const FirewallDeviceLanding = React.memo(
...
+              <DebouncedSearchTextField
+                onSearch={(val) => {
+                  filter(val);
+                }}
+                expand={true}
+                hideLabel
+                label=""
+                placeholder={`Search ${formattedType}s`}
+                value={searchText}
+              />
```
abailly-akamai, line 118: you may want to set a lower debounceTime than the default (`250` instead of `400`)

packages/manager/src/features/Images/ImagesDrawer.tsx
```diff
@@ -397,14 +397,6 @@ export const ImagesDrawer = (props: CombinedProps) => {
...
           onClick: close,
         }}
-        updateFor={[
-          requirementsMet,
-          classes,
-          submitting,
-          mode,
-          label,
-          description,
-        ]}
```
abailly-akamai, line 407: Have you confirmed there's no side effect removing this renderGuard property?

abailly-akamai (commented):
Coming along! some more needed changes

packages/manager/src/components/DebouncedSearchTextField/DebouncedSearchTextField.tsx
tyler-akamai: yeah I removed it because DebouncedSearchProps extends TextFieldProps which also requires the label prop so I didnt think it was needed to state it again. But I can re-add it.

packages/manager/src/features/Images/ImagesDrawer.tsx
tyler-akamai: It was interfering with storybook and causing it to not read JS-Docs. So I checked with Banks and he said it most likely isnt critical. So then I tested out the image Drawer and it seems like its still working properly and theres no rendering issues. Open for additional thoughts.

packages/manager/src/components/DebouncedSearchTextField/DebouncedSearchTextField.tsx
abailly-akamai: How would a label be passed then?

packages/manager/src/features/Images/ImagesDrawer.tsx
tyler-akamai: It also seems like the image drawer is the only reference of many (for ActionsPanel) that utilizes the renderguard.

packages/manager/src/features/Firewalls/FirewallDetail/Devices/FirewallDeviceLanding.tsx
tyler-akamai: Okay! is there a reason its at 400? Should we change the default to something lower?

packages/manager/src/components/ActionsPanel/ActionsPanel.tsx
```diff
@@ -13,19 +12,22 @@ interface ActionButtonsProps extends ButtonProps {
...
   /**
    * secondary type actionable button custom aria descripton.
    */
   secondaryButtonProps?: ActionButtonsProps;
 }
 
-const ActionsPanelComponent = (props: ActionPanelProps) => {
+/**
+ * An `<ActionsPanel />` is a UI component that provides a primary and secondary action.
+ * It can also be used to render a single action within modals or drawers for consistency.
```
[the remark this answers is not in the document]
tyler-akamai: Just updated the comment!

packages/manager/src/features/Firewalls/FirewallDetail/Devices/FirewallDeviceLanding.tsx
abailly-akamai: We may want to discuss this for a later PR, but see how the experience feels with 400. I find it to feel laggy

packages/manager/src/components/DebouncedSearchTextField/DebouncedSearchTextField.tsx
abailly-akamai: Gotcha - in that case you'd have to move the label there for existing instances? You should be getting type errors, no?

abailly-akamai: i understand now, please carry on!

packages/manager/src/components/ActionsPanel/ActionsPanel.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+  args: {
+    primaryButtonProps,
+    secondaryButtonProps,
+  },
+  render: (args: ActionPanelProps) => {
+    return <ActionsPanel sx={{ justifyContent: 'flex-start' }} {...args} />;
+  },
+};
+
+export default meta;
```
abailly-akamai, line 31: thanks for adding!

carrillo-erik:
I'm experiencing a noticeable delay when adding or removing devices to the Firewall. I did not observe this in the other PR where the functionality to add devices was introduced.

https://github.com/linode/manager/assets/119514965/4977b083-3b41-4090-b402-2a7477ecd386

cpathipa:
@tyler-akamai Over all  LGTM, Othe than below observations:

The Linode firewall's table does not refresh successfully after adding Linodes until the page is manually refreshed.

image

Same Behavior when Remove action is performed

image

It would be nice to have descriptive tooltip message saying clear all or deselect all for X button. 

image

carrillo-erik:
```jsx 
    //   const devices =
    //   allDevices?.filter((device) => device.entity.type === type) || [];

    const [filteredDevices, setFilteredDevices] = React.useState<
      FirewallDevice[]
    >([]);

    React.useEffect(() => {
      setFilteredDevices(allDevices?.filter((device) => device.entity.type === type) || []);
    }, [allDevices, type]);

 const selectedDevice = filteredDevices?.find(
      (device) => device.id === selectedDeviceId
    );

    const filter = (value: string) => {
      setSearchText(value);
      const filtered = filteredDevices.filter((device) => {
        return device.entity.label.toLowerCase().includes(value.toLowerCase());
      });
      setFilteredDevices(filtered);
    };
```

https://github.com/linode/manager/assets/119514965/035590d0-be67-423f-9c32-535acb05cf76

tyler-akamai:
> ```js
>     //   const devices =
>     //   allDevices?.filter((device) => device.entity.type === type) || [];
> 
>     const [filteredDevices, setFilteredDevices] = React.useState<
>       FirewallDevice[]
>     >([]);
> 
>     React.useEffect(() => {
>       setFilteredDevices(allDevices?.filter((device) => device.entity.type === type) || []);
>     }, [allDevices, type]);
> 
>  const selectedDevice = filteredDevices?.find(
>       (device) => device.id === selectedDeviceId
>     );
> 
>     const filter = (value: string) => {
>       setSearchText(value);
>       const filtered = filteredDevices.filter((device) => {
>         return device.entity.label.toLowerCase().includes(value.toLowerCase());
>       });
>       setFilteredDevices(filtered);
>     };
> ```
> 
>  Screen.Recording.2023-09-12.at.3.49.44.PM.mov

Huge thank you!

carrillo-erik (approved):

PR #9645, opened 2023-09-07
feat: [M3-6993] – Add VPC data to Linode Detail header

dwiley-akamai:
## Description 📝
For Linodes assigned to a VPC, display the linkified VPC label, the subnets list, and the VPC IPv4 address in a "VPC" section in the Linode Detail header.

## To-Do:
- [x] Minor code cleanup
- [x] Unit tests

## Major Changes 🔄
For Linodes assigned to a VPC, display a "VPC" section in the Linode Detail header.

Also, add an `enabled` argument to the `useVPCsQuery()` so that the query doesn't run on Linode Detail pages if the VPC section shouldn't be displayed.

## Preview 📷
Screenshot 2023-09-07 at 2 17 54 PM

## How to test 🧪
Confirm that a Linode assigned to a VPC has the VPC section added to its Detail header.

Confirm that a Linode not assigned to a VPC does not have the VPC section in its Detail header.


packages/manager/src/features/Linodes/LinodeEntityDetail.styles.ts
```diff
@@ -215,3 +226,64 @@ export const StyledTableRow = styled(TableRow, { label: 'StyledTableRow' })({
     opacity: 1,
   },
 });
+
+// ---------------------------------------------------------------------
+// Hook
+// ---------------------------------------------------------------------
+
+export const useSxObjects = () => {
```
abailly-akamai, line 234: naming convention is a little off here. This is not technically a hook and it does not need to be a function.

All in all this makes a heavy use of sx which could be changed to use styled components

packages/manager/src/features/Linodes/LinodeEntityDetail.tsx
```diff
@@ -330,10 +348,45 @@ export const Body = React.memo((props: BodyProps) => {
...
+  const vpcsList = vpcData?.data ?? [];
+
+  const vpcLinodeIsAssignedTo = vpcsList.find((vpc) => {
+    const subnets = vpc.subnets;
+
+    return Boolean(subnets.find((subnet) => subnet.linodes.includes(linodeId)));
+  });
+
+  const { data: configs } = useAllLinodeConfigsQuery(linodeId);
+  let _configInterfaceWithVPC: Interface | undefined;
```
dwiley-akamai, line 373: Opted for this approach to hold the interface data as opposed to looping through two separate times as in the initial commit

packages/manager/src/features/Linodes/LinodeEntityDetail.tsx
```diff
@@ -416,6 +469,66 @@ export const Body = React.memo((props: BodyProps) => {
...
+                paddingLeft: '8px',
+              },
+            }}
+            container
+            direction="row"
+            spacing={2}
+          >
+            <StyledVPCGrid>
+              <StyledListItem>
+                <Box component="span" sx={sxLabel}>
```
hana-linode, line 502: Could this be a StyledLabelBox with the sxLabel css instead?

dwiley-akamai:
> Looks like there's a bit of border missing on the right side image

I haven't been able to get the border to extend all the way -- when I inspect that VPC subpanel, Dev Tools indicates there's a negative margin (❓):

Screenshot 2023-09-13 at 4 56 38 PM

@jaalah-akamai any insights into how to fix this issue?

jdamore-linode:
I had to abort the e2e test run because one of the runners was hanging, but all of the tests did run and pass -- feel free to treat it as a pass 👍

(cc @dwiley-akamai @jaalah-akamai )

jaalah-akamai (approved):

hana-linode (approved):

PR #9653, opened 2023-09-08
change: Allow Cypress file watching to be disabled using environment variable

jdamore-linode:
## Description 📝
Allows file watching and automatic reloading to be disabled via an environment variable (including in `.env`) when using the Cypress UI.

By default, Cypress's UI will automatically reload a test if any changes are made to the spec file which can be disruptive. This PR allows the behavior to optionally be disabled by setting the `CY_TEST_DISABLE_FILE_WATCHING` environment variable when running Cypress.

Developers can permanently set this behavior by adding it to their `.env` file.

## Major Changes 🔄
- Reads `CY_TEST_DISABLE_FILE_WATCHING` environment variable to determine whether or not to disable file watching UI
- Adds docs for `CY_TEST_DISABLE_FILE_WATCHING`

## How to test 🧪
1. Run `yarn cy:debug`
2. Run a test, and modify and save the corresponding test file; observe that Cypress reloads automatically
3. Close Cypress, and run it again with the environment variable: `CY_TEST_DISABLE_FILE_WATCHING=1 yarn cy:debug`
4. Repeat step 2, confirm that Cypress does not reload automatically

### DIFF
--- a/docs/development-guide/08-testing.md
+++ b/docs/development-guide/08-testing.md
@@ -203,13 +203,14 @@ These environment variables facilitate splitting the Cypress run between multipl
 | `CY_TEST_SPLIT_RUN_INDEX` | Numeric index for each Cypress test runner | `1`, `2`, etc. | Unset                      |
 | `CY_TEST_SPLIT_RUN_TOTAL` | Total number of runners for the tests      | `2`            | Unset                      |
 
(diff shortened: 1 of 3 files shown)


abailly-akamai (approved):
yessss - thank you 👍 🚢

PR #9654, opened 2023-09-08
feat: [M3-7159] - Update Enable Object Storage modal for beta DC-specific pricing

mjac0bs:
## Description 📝
This PR updates the copy in the Enable Object Storage modal to account for DC-specific pricing changes in the beta period. OBJ will remain free in Jakarta and Sao Paulo while in beta. After the beta ends, customers will be charged in these regions. 

## Major Changes 🔄
- Fixes a bug where the Close button ('X') did not close the modal
- Updates EnableObjectStorage modal to adhere to our coding standards (removed default export, renamed props).
- Creates and uses constants and a OBJ price increase map to determine regions with beta pricing.
   - Adds TODOs where this will be updated in a future ticket when DC-specific pricing is applied.
- Updates unit test coverage for EnableObjectStorageModal.

When the `dcSpecificPricing` feature flag is **enabled**:
- Adds copy to the Enable Object Storage modal in the buckets flow to inform users of free beta period
- Updates the copy of the Enable Object Storage modal in the access key flow to have more general messaging about pricing that applies to ALL regions, since no region selection is made
- Updates the "Learn more" link to direct to the OBJ section of https://linode.com/pricing rather than the OBJ docs page (https://www.linode.com/docs/products/storage/object-storage/#pricing)
   - Why? To help indicate that OBJ, like pricing for other products impacted by DC-specific pricing, can expect similar price changes in Jakarta and Sao Paulo once beta ends

## Preview 📷

| Description  | This Branch   |
| ------- | ------- |
| Prod / `objDCSpecificPricing` flag OFF | Screenshot 2023-09-19 at 9 12 58 PM |
|`objDCSpecificPricing` flag on with bucket in `us_east`  | Screenshot 2023-09-19 at 9 15 41 PM |
|`objDCSpecificPricing` flag on with bucket in `id_cgk`  | Screenshot 2023-09-19 at 9 17 28 PM |
|`objDCSpecificPricing` flag on with bucket in `br_gru`  | Screenshot 2023-09-19 at 9 18 22 PM |
|`objDCSpecificPricing` flag on no region selection (access keys) | Screenshot 2023-09-19 at 9 19 22 PM |

## How to test 🧪
- Check out this PR.
- Turn on the `dcSpecificPricing` feature flag.
- If Object Storage already enabled, go to /account/settings and cancel it
2. **How to verify changes?**
- With the feature flag on, confirm that the copy of the modal matches the copy in the screenshots above for each scenario. 
To mock clusters in Sao Paulo or Jakarta, update the `rest.get('*object-storage/clusters'` mocked GET request in `serverHandlers.ts` to include clusters in both regions. Then turn on the MSW.
```
  rest.get('*object-storage/clusters', (req, res, ctx) => {
    const jakartaCluster = objectStorageClusterFactory.build({
      id: `id-cgk-0` as any,
      region: 'id-cgk',
    });
    const saoPauloCluster = objectStorageClusterFactory.build({
      id: `br-gru-0` as any,
      region: 'br-gru',
    });
    const clusters = objectStorageClusterFactory.buildList(3);
    return res(
      ctx.json(makeResourcePage([jakartaCluster, saoPauloCluster, ...clusters]))
    );
  }),
  rest.get('*object-storage/keys', (req, res, ctx) => {
    return res(
      ctx.json(makeResourcePage(objectStorageKeyFactory.buildList(3)))
    );
  }),
```
- With the feature flag off, confirm that modal in the access key flow and bucket flow (all regions) matches prod.
- Confirm an unrelated bug fix: clicking the Close button now closes the modal rather than does nothing.
Verify unit test passes:
```
yarn test EnableObjectStorageModal
```

### DIFF
--- a/packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
@@ -129,6 +129,7 @@ export const CreateBucketDrawer = (props: Props) => {
           error={errorMap.cluster}
           onBlur={formik.handleBlur}
           onChange={(value) => formik.setFieldValue('cluster', value)}
+          required
           selectedCluster={formik.values.cluster}
         />
         {showAgreement ? (
@@ -142,6 +143,7 @@ export const CreateBucketDrawer = (props: Props) => {
         <ActionsPanel
           primaryButtonProps={{
             'data-testid': 'create-bucket-button',
+            disabled: !formik.values.cluster,
             label: 'Create Bucket',
             loading: isLoading,
             type: 'submit',
--- a/packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
@@ -18,6 +18,7 @@ import {
   useObjectStorageClusters,
 } from 'src/queries/objectStorage';
 import { useProfile } from 'src/queries/profile';
+import { useRegionsQuery } from 'src/queries/regions';
 import { sendCreateBucketEvent } from 'src/utilities/analytics';
 import { getErrorMap } from 'src/utilities/errorUtils';
 import { isEURegion } from 'src/utilities/formatRegion';
@@ -34,6 +35,7 @@ export const CreateBucketDrawer = (props: Props) => {
   const { data: profile } = useProfile();
   const { isOpen, onClose } = props;
   const isRestrictedUser = profile?.restricted;
+  const { data: regions } = useRegionsQuery();
   const { data: clusters } = useObjectStorageClusters();
   const { data: buckets } = useObjectStorageBuckets(clusters);
   const {
@@ -96,6 +98,12 @@ export const CreateBucketDrawer = (props: Props) => {
       isEURegion(formik.values.cluster)
   );
 
+  const clusterRegion =
+    regions &&
+    regions.filter((region) => {
+      return formik.values.cluster.includes(region.id);
+    });
+
   const errorMap = getErrorMap(['label', 'cluster'], error);
 
   return (
@@ -155,7 +163,7 @@ export const CreateBucketDrawer = (props: Props) => {
           handleSubmit={formik.handleSubmit}
           onClose={() => setIsEnableObjDialogOpen(false)}
           open={isEnableObjDialogOpen}
-          regionId={formik.values.cluster.slice(0, -2)}
+          regionId={clusterRegion?.[0]?.id}
         />
       </form>
     </Drawer>
--- a/packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
@@ -150,6 +150,7 @@ export const CreateBucketDrawer = (props: Props) => {
         />
 
         <EnableObjectStorageModal
+          regionId={formik.values.cluster.slice(0, -2)}
           handleSubmit={formik.handleSubmit}
           onClose={() => setIsEnableObjDialogOpen(false)}
           open={isEnableObjDialogOpen}
--- a/packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
@@ -22,7 +22,7 @@ import { sendCreateBucketEvent } from 'src/utilities/analytics';
 import { getErrorMap } from 'src/utilities/errorUtils';
 import { isEURegion } from 'src/utilities/formatRegion';
 
-import EnableObjectStorageModal from '../EnableObjectStorageModal';
+import { EnableObjectStorageModal } from '../EnableObjectStorageModal';
 import ClusterSelect from './ClusterSelect';
 
 interface Props {
@@ -150,10 +150,10 @@ export const CreateBucketDrawer = (props: Props) => {
         />
 
         <EnableObjectStorageModal
-          regionId={formik.values.cluster.slice(0, -2)}
           handleSubmit={formik.handleSubmit}
           onClose={() => setIsEnableObjDialogOpen(false)}
           open={isEnableObjDialogOpen}
+          regionId={formik.values.cluster.slice(0, -2)}
         />
       </form>
     </Drawer>
--- a/packages/manager/src/features/ObjectStorage/EnableObjectStorageModal.tsx
+++ b/packages/manager/src/features/ObjectStorage/EnableObjectStorageModal.tsx
@@ -8,11 +8,14 @@ import { TextTooltip } from 'src/components/TextTooltip';
 import { Typography } from 'src/components/Typography';
 import { useFlags } from 'src/hooks/useFlags';
 import {
+  ENABLE_OBJ_ACCESS_KEYS_MESSAGE,
   OBJ_STORAGE_PRICE,
   ObjStoragePriceObject,
 } from 'src/utilities/pricing/constants';
-import { getObjectStorageDCSpecificPrices } from 'src/utilities/pricing/dynamicPricing';
+import { objectStoragePriceIncreaseMap } from 'src/utilities/pricing/dynamicPricing';
 
+const OBJ_STORAGE_STORAGE_AMT = '250 GB';
+const OBJ_STORAGE_NETWORK_TRANSFER_AMT = '1 TB';
 export interface Props {
   handleSubmit: () => void;
   onClose: () => void;
@@ -25,12 +28,9 @@ export const EnableObjectStorageModal = (props: Props) => {
 
   const flags = useFlags();
   const objStoragePrices: ObjStoragePriceObject =
-    regionId &&
-    getObjectStorageDCSpecificPrices({
-      basePrice: OBJ_STORAGE_PRICE,
-      flags,
-      regionId,
-    });
+    regionId && flags.dcSpecificPricing
+      ? objectStoragePriceIncreaseMap[regionId] ?? OBJ_STORAGE_PRICE
+      : OBJ_STORAGE_PRICE;
 
   return (
     <ConfirmationDialog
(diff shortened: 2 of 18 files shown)


packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
```diff
@@ -153,6 +153,7 @@ export const CreateBucketDrawer = (props: Props) => {
           handleSubmit={formik.handleSubmit}
           onClose={() => setIsEnableObjDialogOpen(false)}
           open={isEnableObjDialogOpen}
+          regionId={formik.values.cluster.slice(0, -2)}
```
mjac0bs, line 156: Because `cluster` is something like `id-cgk-0`, we need to get rid of the last two characters in the string to have a valid region id.

packages/manager/src/features/ObjectStorage/EnableObjectStorageModal.tsx
```diff
@@ -1,57 +1,119 @@
...
-        <Link to="/account/settings">Account Settings</Link>.
-      </Typography>
-    </ConfirmationDialog>
-  );
-};
-
-export default React.memo(EnableObjectStorageModal);
+export const EnableObjectStorageModal = React.memo(
+  (props: EnableObjectStorageProps) => {
+    const { handleSubmit, onClose, open, regionId } = props;
```
mjac0bs, line 28: Just minor changes to bring this component up to speed with our coding standards. Got rid of default export and moved React.memo, renamed props interface since it's exported, and destructured props within the component because I think it looks cleaner.

packages/manager/src/features/ObjectStorage/EnableObjectStorageModal.tsx
```diff
@@ -1,57 +1,119 @@
...
+    const flags = useFlags();
+    const { data: regions } = useRegionsQuery();
+
+    const regionLabel =
+      regions?.find((r) => r.id === regionId)?.label ?? regionId;
+    const objStoragePrices: ObjStoragePriceObject = OBJ_STORAGE_PRICE;
+    // TODO: DC Pricing - M3-6973: Uncomment to calculate `objStoragePrices` based on region map rather than OBJ_STORAGE_PRICE constant
+    //   regionId && flags.dcSpecificPricing
+    //     ? objectStoragePriceIncreaseMap[regionId] ?? OBJ_STORAGE_PRICE
+    //     : OBJ_STORAGE_PRICE;
```
mjac0bs, line 39: In preparation for future OBJ pricing changes, I created new constants and a price map. Until beta ends, we'll just use the `OBJ_STORAGE_PRICE` constant to get our same account-wide OBJ prices as before.

packages/manager/src/features/ObjectStorage/EnableObjectStorageModal.tsx
```diff
@@ -1,57 +1,119 @@
...
+          {OBJ_STORAGE_STORAGE_AMT} of storage and{' '}
+          {OBJ_STORAGE_NETWORK_TRANSFER_AMT} of outbound data transfer. Beyond
+          that, it's{' '}
+          <strong>${objStoragePrices.storage_overage} per GB per month.</strong>{' '}
+          {!flags.dcSpecificPricing && (
+            <Link to="https://www.linode.com/docs/products/storage/object-storage/#pricing">
+              Learn more.
+            </Link>
+          )}
+        </Typography>
```
mjac0bs, line 77: This is our existing copy in prod; we're just now using `objectStoragePrices` rather than hardcoding the values directly in the text. The other minor change is removing the `subtitle` variant, which changes the text size slightly but makes it consistent throughout different versions of the modal.

packages/manager/src/features/ObjectStorage/EnableObjectStorageModal.tsx
```diff
@@ -1,57 +1,119 @@
...
+              },
+            }}
+            secondaryButtonProps={{
+              'data-testid': 'cancel',
+              label: 'Cancel',
+              onClick: onClose,
+            }}
+          />
+        )}
+        onClose={onClose}
```
mjac0bs, line 99: This was `onClose={close}` before, which was a bug preventing the modal from being closed when the close (X) button was clicked.

packages/manager/src/features/ObjectStorage/EnableObjectStorageModal.tsx
```diff
@@ -1,57 +1,119 @@
...
+        open={open}
+        title="Just to confirm..."
```
mjac0bs, line 101: This modal title doesn't make the most sense for the beta message with Jakarta and Sao Paulo, but UX has determined that it's okay for now.

packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
bnussman-akamai: We may want to make this more robust. Object Storage clusters are historically not known to be `id`ed consistently. For example, in the alpha environment, the Object Storage cluster is named `alpha` rather than `alpha-1`. 

It probably won't be a big deal if we merge as is, but we may want to consider making it more robust.

abailly-akamai: It wouldn't be such a huge lift to implement an include based on the region ID to make this a bit more reliable

packages/manager/src/utilities/pricing/dynamicPricing.ts
```diff
@@ -25,6 +25,20 @@ export const priceIncreaseMap = {
...
+    monthly: 0.0,
+    storage_overage: 0.0,
+    transfer_overage: 0.0,
+  },
+  'id-cgk': {
+    monthly: 0.0,
+    storage_overage: 0.0,
+    transfer_overage: 0.0,
+  },
+};
```
abailly-akamai, line 40: Out of curiosity, are those never going to come from the DB?

packages/manager/src/features/ObjectStorage/EnableObjectStorageModal.tsx
```diff
@@ -1,57 +1,119 @@
...
+    //     ? objectStoragePriceIncreaseMap[regionId] ?? OBJ_STORAGE_PRICE
+    //     : OBJ_STORAGE_PRICE;
+    const isObjBetaPricingRegion =
+      regionId && objectStoragePriceIncreaseMap.hasOwnProperty(regionId);
+
+    /**
+     * @param regionLabel The human-readable region for the bucket (e.g. Jakarta, ID)
+     * @returns Dynamic modal text dependent on the existence and selection of a region
+     */
+    const getRegionPricingText = (regionLabel: string) => {
```
abailly-akamai, line 47: nitpick: rename to `renderRegionPricingText` or just `regionPricingText`

packages/manager/src/features/ObjectStorage/EnableObjectStorageModal.tsx
```diff
@@ -1,57 +1,119 @@
...
+              label: 'Cancel',
+              onClick: onClose,
+            }}
+          />
+        )}
+        onClose={onClose}
+        open={open}
+        title="Just to confirm..."
+      >
+        {getRegionPricingText(regionLabel ?? 'this region')}
```
abailly-akamai, line 103: can it be undefined?

abailly-akamai (commented):
Everything looks good! left a few comments. Also:

We could get rid of this option by requiring the region select. 

Screenshot 2023-09-20 at 12 09 15

I know this copy is needed for the access keys flow, but here it seems superfluous?

packages/manager/src/features/ObjectStorage/EnableObjectStorageModal.tsx
mjac0bs: Yes, because the enable object storage modal in `AccessKeyDrawer.tsx` does not provide a region selection because there is currently no region associated with an access key upon creation.

packages/manager/src/features/ObjectStorage/EnableObjectStorageModal.tsx
mjac0bs: Addressed in [e4ff633](https://github.com/linode/manager/pull/9654/commits/e4ff63341c2f5ba55f91ac40392e9951f4e3f0dc).

packages/manager/src/utilities/pricing/dynamicPricing.ts
mjac0bs: When we get a `/prices` endpoint from API, I hope they can be included. 🤞🏼

packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
mjac0bs: > an include based on the region ID 

Can you clarify what you mean by this?

abailly-akamai:
```
const clusterRegion = regions.filter((region) => {
  return formik.values.cluster.includes(region.id);
});
```

I was thinking something along those lines but unsure if this would work long term either. Not too familiar with the cluster names. Based on what you and Banks said it may be a bit better still?

mjac0bs: Thanks, yeah, I think that is better ([bb485fc](https://github.com/linode/manager/pull/9654/commits/bb485fcbc1bf85be89f3d68001ca406b4588ca91)). If a cluster isn't a valid region now, we'll fall back on displaying the generic message, same as the access key drawer.

mjac0bs:
@abailly-akamai Feedback addressed! This is ready for another look when you're able. Joe should be pushing up some integration tests early tomorrow.

I did make region selection required for form submission:
image

abailly-akamai (approved):
Looks great! Took a second pass and it all looks good

- With flag off we get the same experience as production ✅ 
- Region select is now required ✅ 
- With flag on we get dynamic copy based on DC ✅ 
- We get a fallback in the case region we have no region ✅ 

Did you check with UX about the font size difference? Asking because it is a rather big diff. I'm with you we should have it consistent with the rest of our dialogs but I am also wondering if it was made big for a reason (important pricing notice or something).

| Production  | This branch   |
| ------- | ------- |
| Screenshot 2023-09-21 at 09 08 42 | Screenshot 2023-09-21 at 09 08 35 |

jdamore-linode (approved):
Looks great @mjac0bs!

Tested the changes manually and everything looked good (and I'm curious to hear what we find re: the dialog text size that Alban mentioned above). Pushed some integration tests which confirm the following:

- Confirms that Object Storage can be enabled using mock API data when DC pricing is enabled.
  - Confirms that pricing information link is present in enrollment dialog.
  - Confirms that cancellation explanation is present in enrollment dialog.
  - Confirms that free beta pricing is explained for regions with special price structures.
  - Confirms that regular pricing information is shown for regions with regular price structures.
  - Confirms that generic pricing information is shown when no region is selected.

- Confirms Object Storage enrollment dialog contents when DC-specific pricing is disabled.
   - Confirms expected dialog language is shown when enrolling by creating a bucket.
   - Confirms expected dialog language is shown when enrolling by creating an access key.
 - Confirms that "Create a Bucket" dialog cannot be submitted without selecting a region first.
 - Confirms that users can cancel Object Storage using mock API data.
  - Confirms that users are warned of Object Storage cancellation data loss.
  - Confirms that Account Settings page updates to reflect cancellation.

To run:

```bash
yarn cy:run -s "cypress/e2e/core/objectStorage/enable-object-storage.spec.ts"
```

jaalah-akamai (approved):
For the sake of not being redundant, I've tested this as well. 🚀

coliu-akamai (approved):
Tested all above scenarios, which worked as expected!

(wow this PR is extra tested 😆)

mjac0bs:
@jdamore-linode Tests are passing and looking good with coverage of changes. Thanks for including!

@abailly-akamai (cc @jdamore-linode) Confirmed with Matthew that a consistent, smaller font size for all versions of this modal is preferred, so I'm going to keep the change made in this PR.

Thanks for the feedback and testing everyone, going to merge this now.

Issue #9655, opened 2023-09-08
Compilation error under Angular 16: Cannot find module 'src/regions' or its corresponding type declarations.


kekonn:
Full error:

```
Error: node_modules/@linode/api-v4/lib/account/types.d.ts:2:24 - error TS2307: Cannot find module 'src/regions' or its corresponding type declarations.

2 import { Region } from 'src/regions';
                         ~~~~~~~~~~~~~
 ```

This is a regression of issue #7686 and was introduced in [this commit](https://github.com/linode/manager/commit/b5b4ad137c37a20b4afc1f030444493ec4f3c789).

bnussman-akamai:
Thank you for opening this issue! We released the fix in `@linode/api-v4@0.101.0`. Please reach out to us if you have anymore issues!

PR #9656, opened 2023-09-08
fix: [#9655] - Fix compilation errors under Angular 16

kekonn:
## Description 📝
This is a fix for issue #9655 .

## Fixes 🔄
- Change `src/regions` to `../regions` in `linode/types.ts`
- Change `src/regions` to `../regions` in `account/types.ts`


packages/api-v4/src/account/types.ts
```diff
@@ -1,5 +1,5 @@
 import { APIWarning } from '../types';
-import { Region } from 'src/regions';
+import { Region } from '../regions';
```
bnussman-akamai, line 2:
```suggestion
import type { Region } from '../regions';
```

packages/api-v4/src/linodes/types.ts
```diff
@@ -1,4 +1,4 @@
-import { Region } from 'src/regions';
+import { Region } from '../regions';
```
bnussman-akamai, line 1:
```suggestion
import type { Region } from '../regions';
```

bnussman-akamai (approved):
Great change! We should always be using relative imports where it makes sense.

coliu-akamai (approved):
Thanks for catching @kekonn!

@bnussman-akamai would we need a changeset for this PR?

bnussman-akamai:
Yes, good call @coliu-akamai

jdamore-linode (approved):
Also calling out that the e2es have passed -- the tests will kick off again and automatically fail when the changeset gets pushed, but I don't think re-running the tests will be necessary at that point

PR #9659, opened 2023-09-11
change: [poc] - Better subnet validation behavior on VPCCreate page

coliu-akamai:
## Description 📝
Previously when working on the VPCCreate page, I'd made two validation versions:
- The first one did not use the CreateVPCSchema validation for subnets (it validated subnets before using the schema) but had the desired UI -- this one got merged in
- The [second one that used the validation schema](https://github.com/coliu-akamai/manager/pull/1) had UI downsides, where you couldn't leave a subnet blank and have it be ignored

In the current reason, errors for specific subnets might sometimes appear as general subnet errors, because of how subnet validation happened... Wasn't the biggest fan of that behavior, so I looked into this some more. That's where this PR comes in -- it uses the CreateVPCSchema validation and processes any subnet specific errors returned by the API, while still having the same UI behavior of the  current version.

## Major Changes 🔄
- The logic for this PR is built off the logic of the [pr I linked above](https://github.com/coliu-akamai/manager/pull/1) -   I basically just added a map of the UI subnet indexes to the indexes of the subnets that get sent to the backend 😅 
- removed the now unused subnet validation logic 

## Preview 📷

| Before  | After   |
| ------- | ------- |
| image | image |

## How to test 🧪
`yarn test formikErrorUtils.test.ts`
1. Try creating VPCs with subnets and ensure that subnet-specific errors correspond with the correct subnet, and that general subnet errors still get handled the same


dwiley-akamai (approved):
Great work on this! 🚀 Left some minor comments on a few things.

Regarding general subnet errors: are there any aside from the max subnet limit that would get displayed from what you've seen?

coliu-akamai:
@dwiley-akamai I've also seen subnet errors for having two+ subnets with the same label or same IP range (these come in as `subnet.label` or `subnet.ipv4`)

packages/manager/src/features/VPCs/VPCCreate/VPCCreate.tsx
```diff
@@ -42,64 +41,73 @@ const VPCCreate = () => {
...
-          return { ...subnet };
-        })
-      );
-      return true;
+  const combineErrorsAndSubnets = (
+    errors: {},
+    visualToAPISubnetMapping: {}
+  ) => {
+    const combinedSubnets: SubnetFieldState[] = [];
+    for (let i = 0; i < values.subnets.length; i++) {
```
jaalah-akamai, line 82: Nit: I think we can simplify this by using a `.map` instead. That way we don't have to push the values to an array. We can also destructure the errors in the `if` block `const errorData = errors[apiSubnetIdx];`

jaalah-akamai (approved):
Approving because the validation looks good.

PR #9663, opened 2023-09-12
feat: [M3-7096] - Show Region `label` in addition to `id` on Invoices

bnussman-akamai:
## Description 📝

- Displays regions using their `label` and `id` in the DC specific pricing invoices. This ticket is a follow up to address some feedback from #9597. 

## Preview 📷

Screenshot 2023-09-12 at 10 39 26 AM

Screenshot 2023-09-12 at 10 45 17 AM

> [!important]
> The CSV does not contain the region `label` because I followed the current pattern of showing "pure" API data. (Notice how the dates are in the API's format)

Screenshot 2023-09-12 at 10 41 20 AM

## How to test 🧪
- Verify that you see Region `label`s and `id`s on the Invoices with the `DC Specific Pricing` feature flag on
- Verify you **do not** see the region column when the flag is **off**

### DIFF
--- a/packages/manager/src/features/Billing/InvoiceDetail/InvoiceTable.tsx
+++ b/packages/manager/src/features/Billing/InvoiceDetail/InvoiceTable.tsx
@@ -44,7 +44,7 @@ export const InvoiceTable = (props: Props) => {
   const { classes } = useStyles();
   const MIN_PAGE_SIZE = 25;
   const flags = useFlags();
-  const columns = flags.dcSpecificPricing ? 9 : 8;
+  const NUM_COLUMNS = flags.dcSpecificPricing ? 9 : 8;
 
   const {
     data: regions,
@@ -56,13 +56,13 @@ export const InvoiceTable = (props: Props) => {
 
   const renderTableContent = () => {
     if (loading || regionsLoading) {
-      return <TableRowLoading columns={columns} />;
+      return <TableRowLoading columns={NUM_COLUMNS} />;
     }
 
     if (regionsError) {
       return (
         <TableRowError
-          colSpan={columns}
+          colSpan={NUM_COLUMNS}
           message="Unable to retrieve regions for this invoice."
         />
       );
@@ -71,7 +71,7 @@ export const InvoiceTable = (props: Props) => {
     if (errors) {
       return (
         <TableRowError
-          colSpan={columns}
+          colSpan={NUM_COLUMNS}
           message="Unable to retrieve invoice items."
         />
       );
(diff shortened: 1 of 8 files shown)


jdamore-linode (approved):

packages/manager/src/features/Billing/InvoiceDetail/InvoiceTable.tsx
```diff
@@ -38,12 +40,123 @@ interface Props {
   loading: boolean;
 }
 
-const InvoiceTable = (props: Props) => {
+export const InvoiceTable = (props: Props) => {
   const { classes } = useStyles();
+  const MIN_PAGE_SIZE = 25;
   const flags = useFlags();
+  const columns = flags.dcSpecificPricing ? 9 : 8;
```
mjac0bs, line 47:
```suggestion
  const NUM_COLUMNS = flags.dcSpecificPricing ? 9 : 8;
```
Optional rename for clarity.

mjac0bs (commented):
Thanks for the updates with some good clean up and tests! 🧹 Left a couple tiny suggestions.

Also: can we update the MSW invoice items endpoint (`rest.get('*invoices/555/items',`) so we're seeing an accurate representation of what an overage line item in the global transfer pool?
It looks like that includes the substring "Transfer Overage" and we've currently got  "Outbound Transfer".
 
Screenshot 2023-09-12 at 1 58 43 PM

packages/manager/src/features/Billing/InvoiceDetail/InvoiceTable.tsx
```diff
@@ -133,7 +133,10 @@ export const InvoiceTable = (props: Props) => {
               ))}
               {count > MIN_PAGE_SIZE && (
                 <TableRow>
-                  <TableCell colSpan={columns}>
+                  <TableCell
+                    colSpan={columns}
+                    sx={{ paddingLeft: '0px !important' }}
```
mjac0bs, line 138: What was the crazy styling issue here?

bnussman-akamai: The pagination controls are inside of a table row for some reason, causing weird padding and a weird hover state. I just did this as a simple fix for now. You can see the bug in production.

Screenshot 2023-09-12 at 5 19 22 PM

mjac0bs: Oh, that is weird. In a table row for alignment purposes, I guess? Layout gets weird when the table cell is removed. Since this is a temp fix and we're trying to avoid `!important` it would be great if we can get a ticket in the backlog just to document it.

image

mjac0bs (approved):
LGTM, thanks! 🚢

packages/manager/src/features/Billing/PdfGenerator/PdfGenerator.ts
abailly-akamai, line 1: Would be nice to have tests for the PDF generation too. Could be a follow up but it wouldn't be too labor intensive using pdf-parse - will make a ticket

abailly-akamai (approved):

PR #9622, opened 2023-09-01
fix: Billing Summary UI Regression from MUI Update

bnussman-akamai:
## Description 📝

- Caused by https://github.com/linode/manager/pull/9603

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-09-01 at 10 24 48 AM | Screenshot 2023-09-01 at 10 25 22 AM |

## How to test 🧪
- Verify `/account/billing` looks good 🎨

PR #9667, opened 2023-09-13
fix: Billing Contact UI Regression from by MUI Update

bnussman-akamai:
## Description 📝

- Fixes alignment issue with Billing Contact information on the Account page (`http://localhost:3000/account/billing`)
- Caused by https://github.com/linode/manager/pull/9603 and related to 
  - https://github.com/linode/manager/pull/9623
  - https://github.com/linode/manager/pull/9622

## The Fix 🔨
- It seems that MUI need `<Grid>` to be a child of `<Grid container>` rather than a `<Box>` so it can correctly apply Grid styles

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-09-12 at 11 27 35 PM | Screenshot 2023-09-12 at 11 27 49 PM |

## How to test 🧪
- Go to `http://localhost:3000/account/billing` and check the alignment and responsiveness of the Billing Contact section

### DIFF
--- a/packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/ContactInformation.tsx
+++ b/packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/ContactInformation.tsx
@@ -1,4 +1,3 @@
-import { Box } from 'src/components/Box';
 import Grid from '@mui/material/Unstable_Grid2';
 import { styled } from '@mui/material/styles';
 import countryData from 'country-region-data';
@@ -104,7 +103,7 @@ const ContactInformation = (props: Props) => {
     (_country) => _country.countryShortCode === country
   )?.countryName;
 
-  const sxBox = {
+  const sxGrid = {
     flex: 1,
     maxWidth: '100%',
     position: 'relative',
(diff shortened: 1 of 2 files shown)


packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/ContactInformation.tsx
```diff
@@ -140,7 +139,7 @@ const ContactInformation = (props: Props) => {
             state ||
             zip ||
             country) && (
-            <Box sx={sxBox}>
+            <Grid sx={sxGrid}>
```
abailly-akamai, line 142: Now that's it's a Grid I don't think you need the extra CSS here. You can most likely remove `sxGrid` sx all together

abailly-akamai (approved):
thx!

packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/ContactInformation.tsx
cpathipa, line 1: Adding m={0} at line [133](https://github.com/linode/manager/pull/9667/files#diff-fc7b4cd30f7d5cebb8db7e8b5fdebbd66e3cde4c82cfa20e86845a473eed0659L133) would solve the issue and consistent with other boxes without having additional padding to subelements.

image


image

bnussman-akamai: We could do that, but do we really want to override MUI's Grid rules? Also, lifting  lift `sxGrid`/`sxBox` styles still breaks some UI

jaalah-akamai (approved):

PR #9668, opened 2023-09-13
feat: [M3-7033] - Update Firewall Landing Device Column

tyler-akamai:
## Description 📝
Updated Firewall Landing Device Column

## Major Changes 🔄
- Changes name of column from 'Linodes' to 'Devices'
- Populated the 'Devices' column with devices instead of linodes
-  Increased the number of visible devices in the column from 3 to 8. Further work/discussion is needed later. The original ticket has it spanning 3 rows but the logic turned out more difficult than expected.

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screenshot 2023-09-13 at 10 08 01 AM | Screenshot 2023-09-13 at 4 45 09 PM |

## How to test 🧪
1. Navigate to http://localhost:3000/firewalls
2. Ensure the column header is 'Devices' instead of 'Linodes'
3. Ensure additional nodes past the third row are not shown.
4. Ensure the links still bring you to the device landing pages


packages/manager/src/features/Firewalls/FirewallLanding/FirewallRow.tsx
```diff
@@ -113,23 +105,23 @@ const getLinodesCellString = (
...
         >
           {idx > 0 && `, `}
           {thisDevice.entity.label}
         </Link>
       ))}
-      {data.length > 3 && (
+      {data.length > 8 && (
         <span>
-          {`, `}plus {data.length - 3} more.
+          {`, `}+ {data.length - 8}
```
coliu-akamai, line 124: do we still need to include the 'more' after?

packages/manager/src/features/Firewalls/FirewallLanding/FirewallRow.tsx
```diff
@@ -113,23 +105,23 @@ const getLinodesCellString = (
 };
 
 export const getDeviceLinks = (data: FirewallDevice[]): JSX.Element => {
-  const firstThree = data.slice(0, 3);
+  const firstThree = data.slice(0, 8);
```
coliu-akamai, line 108:
```suggestion
  const firstEight = data.slice(0, 8);
```

coliu-akamai (commented):
Functionality works as expected! Had a couple questions/comments below. 

It feels like the devices content list might be a bit cramped at the top/bottom, especially as the row has to grow to fit all 8 devices and if the device names are super long (like my randomly generated 63 char device names haha) -- might be worth adding some padding on the top and bottom? For example:

image

coliu-akamai:
Also, here's an image of what the page might potentially look like with multiple firewalls that have a lot of devices with super long names:
image

Padding might be helpful to differentiate between rows / help declutter a bit? Maybe check with UX!

packages/manager/src/features/Firewalls/FirewallLanding/FirewallRow.tsx
tyler-akamai: Nope, the ui mockups dont include them anymore.

coliu-akamai (approved):
Works/looks as expected, good work! Also don't forget to update the description of this PR / how to test this, now that the behavior is different 

🎉

jaalah-akamai (approved):

jaalah-akamai:
Nice changes. Confirming everything looks good and that we walked through things together 🚀

PR #9669, opened 2023-09-13
chore: Remove `React.FC` from docs examples

bnussman-akamai:
## Description 📝

- Remove `React.FC` usage from our docs because we don't use `React.FC` anymore
  - https://medium.com/raccoons-group/why-you-probably-shouldnt-use-react-fc-to-type-your-react-components-37ca1243dd13

## How to test 🧪

- Check diff and make sure I didn't mess up any examples in the docs


mjac0bs (approved):
🧼 🚢

abailly-akamai (approved):
Nice thanks for the cleanup!

PR #9687, opened 2023-09-15
feat: [M3-6752] - Assign Linodes to Subnet drawer

coliu-akamai:
## Description 📝
Add new drawer for assigning linodes

TODOS
- [x] fix error handling
- [x] unit tests
- [x] write up findings on how to show that linodes need rebooting issue (ty Alban!!)
- [x] general cleanup (+ looking into further cleanup...)
- [x] unassign linodes that were already assigned 
- [x] downloading assigned linodes 

## Major Changes 🔄
- added Subnet Assign Linodes Drawer + relevant tests
- Had to update validation for the linodeInterfaceSchema (this expected an array of interface objects, but the endpoint that used this as validation only wanted a single interface)
- Add new component, RemovableSelectionsList, since this is used in both Un/assign linode flows

## Preview 📷
| selecting linode  | after selecting linode   |
| ------- | ------- |
| image | image |

image

## How to test 🧪
```
yarn test SubnetAssignLinodesDrawer
yarn test SubnetActionMenu
yarn test RemovableSelectionsList
```
Test the following flows: (****integration tests for these flows will be handled in a separate ticket/pr!)
- assign a linode with only one configuration (default) (confirm that linode is removed from the list of linodes you can assign)
  - test auto-assigning a VPC ipv4 vs not auto-assigning
- assign a linode with multiple configurations (confirm that linode is removed from the list of linodes you can assign)
  - test auto-assigning a VPC ipv4 vs not auto-assigning
- test that after assigning a linode, you can unassign it in the same drawer and that it appears back in the list of linodes that you can assign 
- test assigning a bad ipv4 - both nonsense one and a valid ip that is not in subnet range (should error)
- test: the button to assign a linode should be disabled if a linode is not selected
- test: if a linode has multiple configs but we have not selected one of these configs, the button to assign should be disabled
- test: assigning a linode with no configs (should error)
- double check to make sure that linodes now appear in the subnet you assigned them to in the subnet table
- assigning the same linode to multiple subnets (both within the same vpc and not)
  - note: there's a followup ticket (m3-7157) to make sure only the ips specific to the linode and subnet are shown, not all ips
- any flows that use the linodeInterfaceSchema still have the same functionality (creating a Linode, creating a Linode config, updating a linode config)
- downloading the CSV (no linodes assigned, 1+ linodes assigned, determine that CSV accurately reflects the linodes in the 'Linodes assigned to Subnet (x) list) (should have linode's ID, label, public ipv4 at minimum, may include more fields)
- dismissible banner warning for rebooting linodes appears if there's 1+ linode assigned to a VPC
- double check that you can't assign linodes to a subnet if you don't have the permissions for that specific vpc

++anything else you can think of that I haven't mentioned here

Honestly, wouldn't recommend using the MSW and instead using just your dev account as it's hard to see changes/confirm all the error handling on it


packages/manager/src/features/Users/UserPermissions.tsx
```diff
@@ -767,5 +767,5 @@ class UserPermissions extends React.Component<CombinedProps, State> {
 
 export default recompose<CombinedProps, Props>(
   withSnackbar,
-  withQueryClient
```
coliu-akamai, line 770: uhh... this should not be here, brb fixing

packages/api-v4/src/linodes/configs.ts
```diff
@@ -3,7 +3,7 @@ import {
   UpdateConfigInterfaceOrderSchema,
   UpdateConfigInterfaceSchema,
   UpdateLinodeConfigSchema,
-  linodeInterfaceSchema,
+  LinodeInterfaceSchema,
```
coliu-akamai, line 6: updated naming to be capitalized -- would this need a changeset?

packages/validation/src/linodes.schema.ts
```diff
@@ -128,78 +128,75 @@ const ipv6ConfigInterface = object().when('purpose', {
     }),
 });
 
-export const linodeInterfaceSchema = array()
-  .of(
-    object().shape({
-      purpose: mixed().oneOf(
-        ['public', 'vlan', 'vpc'],
-        'Purpose must be public, vlan, or vpc.'
+export const LinodeInterfaceSchema = object().shape({
```
coliu-akamai, line 131: updated capitalization + made `LinodeInterfaceSchema` only validate a single object, not an array of objects (that's now the job of `LinodeInterfacesSchema`)

Would this also need a changeset?

packages/manager/src/components/DownloadCSV/DownloadCSV.tsx
```diff
@@ -2,12 +2,14 @@ import { SxProps } from '@mui/system';
 import * as React from 'react';
 import { CSVLink } from 'react-csv';
 
+import DownloadIcon from 'src/assets/icons/lke-download.svg';
```
coliu-akamai, line 5: The changes in this file match @jaalah-akamai 's ticket for the Unassign Linode subnet drawer (copy-paste tbh), since they're super similar -- may have merge conflicts later

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.styles.ts
```diff
@@ -0,0 +1,62 @@
...
+  display: 'flex',
+  flexDirection: 'column',
+  height: '52px',
+  justifyContent: 'center',
+  maxWidth: '416px',
+  paddingLeft: theme.spacing(2),
+  width: '100%',
+}));
+
+export const SelectedOptionsHeader = styled('h4', {
```
coliu-akamai, line 28: Similarly, a lot of the styling in this file matches/copies @jaalah-akamai 's ticket as the lists in the UX mockup are the same -- we could move this styling to a common styling file or have both drawers import their styling from here later!

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.test.tsx
```diff
@@ -0,0 +1,80 @@
+import { Subnet } from '@linode/api-v4';
```
coliu-akamai, line 1: Unit tests here are pretty basic tbh, will need to be a lot more thorough in the integration tests (see all the cases to test for 😭)

packages/manager/src/features/VPCs/VPCDetail/VPCDetail.tsx
```diff
@@ -184,7 +188,15 @@ const VPCDetail = () => {
...
-      <VPCSubnetsTable vpcId={vpc.id} />
+      {numLinodes > 0 && (
+        <DismissibleBanner
+          preferenceKey={`reboot-linodes-warning-banner`}
+          variant="warning"
+        >
+          <Typography variant="body1">{REBOOT_LINODE_WARNING}</Typography>
+        </DismissibleBanner>
+      )}
+      <VPCSubnetsTable vpcId={vpc.id} vpcRegion={vpc.region} />
```
coliu-akamai, line 199: Adding vpc.region as a prop to the SubnetsTable because when determining which linodes we can assign to a subnet, we want them to be in the same region as the VPC

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
```diff
@@ -0,0 +1,447 @@
...
+
+  const { data: profile } = useProfile();
+  const { data: grants } = useGrants();
+  const vpcPermissions = grants?.vpc.find((v) => v.id === vpcId);
+
+  // there isn't a 'view VPC/Subnet' grant that does anything, so all VPCs get returned even for restricted users
+  // with permissions set to 'None'. Therefore, we're treating those as read_only as well
+  const userCannotAssignLinodes =
+    Boolean(profile?.restricted) &&
+    (vpcPermissions?.permissions === 'read_only' || grants?.vpc.length === 0);
```
coliu-akamai, line 101: I've noticed this logic pop up pretty frequently in all the drawers/dialogs -- may be worth moving to some parent component and then passing in the userCannotX variable to avoid repetition?

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
```diff
@@ -0,0 +1,447 @@
...
+  >([]);
+
+  // We only want to keep track the linodes we've assigned to a subnet while this drawer is open, so
+  // we need to store that information in local state instead of using the subnet's assigned linodes
+  // (which keeps track of all linodes assigned to a subnet, not just the ones currently being assigned).
+  // If we close the drawer and then reopen it, this value should be [].
+  const [
+    assignedLinodesAndConfigData,
+    setAssignedLinodesAndConfigData,
+  ] = React.useState<LinodeAndConfigData[]>([]);
```
coliu-akamai, line 89: Hm.. not too sure how well I can explain this in writing, so lmk if this is unclear/happy to hop on a call and discuss both the stuff commented above and bullets below:

- When determining which Linodes (and relevant config data) should be included in this array, I've relied on the endpoint calls for assigning/unassigning a linode: if there are no errors from assigning/unassigning, I add or remove the Linode from this array (see lines 170-190 and 213-218). (if the API throws an error, the linode will not get added/removed to this array)
- I think it'd be a lot safer to also validate against the linode ids of the passed in subnet, but I (believe I) noticed that even though I invalidate the relevant queries, the subnet prop being passed in doesn't get updated (I can see everything being updated on the details page in real time though), which means that the known linodes assigned to the subnet may not be up-to-date in the props
- So, is it worth querying for the relevant subnet in this drawer as well in order to make sure the information related to it is up-to-date, and then checking against that subnet's linodes? Or is the current check of just relying on the api erroring or not safe enough to determine whether the information in this array is accurate @abailly-akamai I think this was something we discussed, so tagging you here for more thoughts too 

thank you!! lmk if you'd like to hop on a call to discuss/explain further!

coliu-akamai (commented):
- The file for the SubnetAssignDrawer itself... is pretty big 😅 -- I'll be looking into potentially breaking it down (maybe moving the form portion into a different file? idk for sure yet). 
- There are quite a few similarities between this drawer and the Unassign Linodes one -- probably worth trying to cleanup/abstract more after both get merged in!
- After assigning a Linode to a subnet, if we go back to the Linode's page and delete it, the Linode will still appear to be assigned to the subnet (even though the Linode no longer exists) unless we refresh the page, because we don't invalidate any subnet/vpc related queries when deleting Linodes. We could potentially look into [this link](https://tanstack.com/query/v4/docs/react/guides/query-invalidation) (ty Dajahi!) -- could be worth following up on this in a separate ticket?

packages/api-v4/src/linodes/configs.ts
dwiley-akamai: I don't think the JS Client needs a changeset for this, but the Validation package does. It's obviously a super minor change, but technically it is a breaking change in the Validation package.

packages/validation/src/linodes.schema.ts
dwiley-akamai: Yes to a changeset here too for the same reason (technically a breaking change for consumers of the validation package)

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
```diff
@@ -0,0 +1,448 @@
...
+  configId: number;
+  configLabel: string;
+  interfaceId: number;
+};
+
+const REBOOT_LINODE_MESSAGE =
+  'Assigning a Linode to a subnet requires you to reboot the Linode to update its configuration.';
+const REGIONAL_LINODE_MESSAGE = `Select the Linodes you would like to assign to this subnet. Only Linodes in this VPC's region are displayed.`;
+const MULTIPLE_CONFIGURATIONS_MESSAGE =
+  'This Linode has multiple configurations. Select which configuration you would like added to the subnet.';
```
dwiley-akamai, line 65: I would be in favor of moving these copy constants and `REBOOT_LINODE_WARNING` in `VPCDetail.tsx` into a new `constants.ts` file within the `/src/features/VPCs` directory. We may be doing more of that across the codebase for constants as well.

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
dwiley-akamai: I think I saw @bnussman-akamai mention on a PR recently that we might create an epic for streamlining the permissions-related logic across the codebase

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
```diff
@@ -0,0 +1,448 @@
...
+          newErrors.none = 'No Linode selected';
+        } else if (configId === -1) {
+          newErrors.none = 'No configuration profile selected';
+        } // no else case: leave room for other errors unrelated to unselected linode/config
+      }
+      setAssignLinodesErrors(newErrors);
+    }
+  };
+
+  const onUnassignLinode = async (data: LinodeAndConfigData) => {
```
jaalah-akamai, line 211: Since the process of un-assigning linodes is almost identical between [our two implementations](https://github.com/linode/manager/pull/9703/files#diff-5ddea0d7689e7ab71d1d46e4ee263ea0ef5bcb7cf1015590705a7971025d4895R200-R227), I think we should turn this into a hook `useUnassignLinode`. The data passed in will be the three things we need to delete an interface.

If this is something you want me to take care of, just let me know and we can tag team this 👍

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
```diff
@@ -0,0 +1,448 @@
...
+        chosenIP: '',
+        selectedConfig: null,
+      });
+    } catch (errors) {
+      // if a linode/config is not selected, the error returned isn't very friendly: it just says 'Not found', so added friendlier errors here
+      const newErrors = getErrorMap(['ipv4.vpc'], errors);
+      if (newErrors.none) {
+        if (!selectedLinode) {
+          newErrors.none = 'No Linode selected';
+        } else if (configId === -1) {
```
dwiley-akamai, line 203: For Linodes with no configuration profile (`linodeConfigs.length === 0`), we end up with:

Screenshot 2023-09-20 at 11 25 24 AM

I think it'd be better if we can display a more specific error if an attempt is made to assign a linode with no config profiles to a subnet

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
```diff
@@ -0,0 +1,448 @@
...
+            }}
+            sx={{ marginBottom: '8px' }}
+            value={values.chosenIP}
+          />
+        )}
+        {linodeConfigs.length > 1 && (
+          <>
+            <FormHelperText sx={{ marginTop: `16px` }}>
+              {MULTIPLE_CONFIGURATIONS_MESSAGE}
+              {/* @TODO: VPC - add docs link */}
```
dwiley-akamai, line 360: `@TODO VPC: [...]` for formatting consistency when we go back and look for these to-do's

dwiley-akamai (commented):
Really great work on this so far! 🎉  This is just my initial pass over the PR, left a few comments and answered some of your questions.

Can we make sure we fix the `perfectionist/sort-imports` & `perfectionist/sort-objects` linter warnings throughout these files?

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
```diff
@@ -0,0 +1,448 @@
...
+        ? unassignLinodesErrors.map((apiError: APIError) => (
+            <Notice
+              key={apiError.reason}
+              spacingBottom={8}
+              text={apiError.reason}
+              variant="error"
+            />
+          ))
+        : null}
+      <SelectedOptionsHeader>{`Linodes Assigned to Subnet (${assignedLinodesAndConfigData.length})`}</SelectedOptionsHeader>
```
jaalah-akamai, line 397: Let's turn this into it's own re-useable component. Maybe something called `RemovableSelections` or some descriptive name that's also generic enough to be used for other products (not just unassign/assigning subnets)?

packages/validation/src/linodes.schema.ts
jaalah-akamai: Yes

packages/api-v4/src/linodes/configs.ts
jaalah-akamai: Yes

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
jaalah-akamai: Good explanation of the issue 👏 . I addressed one solution below!

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
coliu-akamai: looking into this! 👀

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
```diff
@@ -0,0 +1,448 @@
...
+              </IconButton>
+            </SelectedOptionsListItem>
+          ))}
+        </SelectedOptionsList>
+      ) : (
+        <StyledNoAssignedLinodesBox>
+          <StyledLabel>No Linodes have been assigned.</StyledLabel>
+        </StyledNoAssignedLinodesBox>
+      )}
+      <DownloadCSV
```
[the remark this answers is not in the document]
coliu-akamai: just to double check, which element are you referring to? The below, or the individual label or box?

```js
</StyledNoAssignedLinodesBox>
  <StyledLabel>No Linodes have been assigned.</StyledLabel>
</StyledNoAssignedLinodesBox>
```

packages/manager/src/cachedData/regions.json
```diff
@@ -1 +1 @@
...172.232.224.26,  172.232.224.27,  172.232.224.21,  172.232.224.24,  172.232.224.22,  172.232.224.20,  172.232.224.31,  172.232.224.28","ipv6": "2600:3c0c::f03c:93ff:feed:a90b,  2600:3c0c::f03c:93ff:feed:a9a5,  2600:3c0c::f03c:93ff:feed:a935,  2600:3c0c::f03c:93ff:feed:a930,  2600:3c0c::f03c:93ff:feed:a95c,  2600:3c0c::f03c:93ff:feed:a9ad,  2600:3c0c::f03c:93ff:feed:a9f2,  2600:3c0c::f03c:93ff:feed:a9ff,  2600:3c0c::f03c:93ff:feed:a9c8,  2600:3c0c::f03c:93ff:feed:a96b"}}, {"id": "br-gru","label": "Sao Paulo, BR","country": "br","capabilities": ["Linodes","NodeBalancers","Block Storage","Object Storage","Kubernetes","Cloud Firewall","Vlans","Premium Plans"],"status": "ok","resolvers": {"ipv4": "172.233.0.4, 172.233.0.9, 172.233.0.7, 172.233.0.12, 172.233.0.5, 172.233.0.13, 172.233.0.10, 172.233.0.6, 172.233.0.8, 172.233.0.11","ipv6": "2600:3c0d::f03c:93ff:fe3d:51cb, 2600:3c0d::f03c:93ff:fe3d:51a7, 2600:3c0d::f03c:93ff:fe3d:51a9, 2600:3c0d::f03c:93ff:fe3d:5119, 2600:3c0d::f03c:93ff:fe3d:51fe, 2600:3c0d::f03c:93ff:fe3d:517c, 2600:3c0d::f03c:93ff:fe3d:5144, 2600:3c0d::f03c:93ff:fe3d:5170, 2600:3c0d::f03c:93ff:fe3d:51cc, 2600:3c0d::f03c:93ff:fe3d:516c"}}],"page":1,"pages":1,"results":14}
```
coliu-akamai, line 1: not sure why this is appearing in the changed files section -- I couldn't find a difference between these files (used online tools to double check for differences)

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
```diff
@@ -0,0 +1,448 @@
...
+  const { data: linodes, refetch: getCSVData } = useAllLinodesQuery(
+    {},
+    {
+      region: vpcRegion,
+    }
+  );
+
+  // We need to filter to the linodes from this region that are not already
+  // assigned to this subnet
+  const findUnassignedLinodes = () => {
```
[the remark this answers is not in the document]
coliu-akamai: Update: sadly this updated the subnet props after we'd already checked for which linodes were assigned/unassigned (and seemed to cause a bunch of extra rerenders), so currently keeping the original solution. However, leaving this comment open in case we want to revisit

packages/manager/src/cachedData/regions.json
jaalah-akamai: You can remove this file - it's created when you `yarn build`

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
```diff
@@ -0,0 +1,391 @@
...
+    setAssignedLinodesAndConfigData,
+  ] = React.useState<LinodeAndConfigData[]>([]);
+  const [linodeConfigs, setLinodeConfigs] = React.useState<Config[]>([]);
+  const [autoAssignIPv4, setAutoAssignIPv4] = React.useState<boolean>(true);
+
+  const { data: profile } = useProfile();
+  const { data: grants } = useGrants();
+  const vpcPermissions = grants?.vpc.find((v) => v.id === vpcId);
+
+  // TODO VPC: this logic for vpc grants/perms appears a lot - commenting a todo here in case we want to move this logic to a parent component
```
dwiley-akamai, line 84: `@TODO VPC: [...]` so we don't miss it 🙏

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
dwiley-akamai: I am going to investigate this tomorrow

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
```diff
@@ -0,0 +1,391 @@
...
+  // there isn't a 'view VPC/Subnet' grant that does anything, so all VPCs get returned even for restricted users
+  // with permissions set to 'None'. Therefore, we're treating those as read_only as well
+  const userCannotAssignLinodes =
+    Boolean(profile?.restricted) &&
+    (vpcPermissions?.permissions === 'read_only' || grants?.vpc.length === 0);
+
+  const csvHeaders = [
+    { key: 'label', label: 'Linode Label' },
+    { key: 'ipv4', label: 'IPv4' },
+    { key: 'id', label: 'Linode ID' },
```
dwiley-akamai, line 94: minor: swapping L94 and L93 feels a little more natural to me when looking at the CSV

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
coliu-akamai: Ty! Jaalah has a possible solution [here](https://github.com/coliu-akamai/manager/compare/feat-m3-6752...jaalah-akamai:example-1234?expand=1), currently testing it!

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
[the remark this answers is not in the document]
coliu-akamai: resolving, currently testing [this solution](https://github.com/coliu-akamai/manager/compare/feat-m3-6752...jaalah-akamai:example-1234?expand=1)

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
coliu-akamai: 😅 ahh so sorry I always seem to forget one character 😅

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
```diff
@@ -0,0 +1,451 @@
...
+  const determineErrorMessage = (
+    selectedLinode: Linode | null,
+    configId: number,
+    errorMap: Record<string, string | undefined>
+  ) => {
+    if (!selectedLinode) {
+      return 'No Linode selected';
+    } else if (configId === -1) {
+      return linodeConfigs.length === 0
+        ? 'Selected Linode must have at least one configuration profile'
```
jaalah-akamai, line 206: @coliu-akamai This seems like the only error message we'll need since we disable the submit. If this is this correct, should we trim this `determineErrorMessage` down a bit?

packages/manager/src/cachedData/regions.json
jaalah-akamai, line 1: We can delete this file so it doesn't show in the PR

jaalah-akamai (approved):
This is looking really good.
- [x] Assign a Linode with default configuration (remove it from assignable Linodes).
- [x] Test auto-assigning VPC IPv4 vs. manual assignment.
- [x] Assign a Linode with multiple configurations (remove it from assignable Linodes).
- [x] Test auto-assigning VPC IPv4 vs. manual assignment.
- [x] Test unassigning a Linode in the same drawer (verify it reappears in the assignable list).
- [x] Test assigning invalid IPv4 addresses (both nonsense and out-of-subnet-range IPs should trigger errors).
- [x] Test deselecting a Linode after modifying auto-assign and inputting a valid VPC IPv4 (should trigger an error).
- [ ] Test attempting to assign a Linode without selecting a configuration (should trigger an error, especially with Linodes having multiple configs). **This is not possible?**
- [x] Verify that assigned Linodes appear in the respective subnet table.
- [x] Assign the same Linode to multiple subnets (within and outside the same VPC).
- [ ] Confirm that only relevant Linode and subnet-specific IPs are displayed (follow-up ticket: m3-7157).
- [x] Ensure LinodeInterfaceSchema functionality remains consistent (create Linode, Linode config, update Linode config).
- [x] Download CSV with no Linodes assigned and 1+ Linodes assigned, verifying accuracy (should include at least Linode ID, label, public IPv4, and possibly more fields).
- [x] Display dismissible banner warning for Linode reboots if 1+ Linodes are assigned to a VPC.
- [x] Verify permissions for assigning Linodes to subnets within a specific VPC.

| SS | Video |
| ----------- | ----------- |
|Screenshot 2023-09-22 at 11 42 09 AMScreenshot 2023-09-22 at 12 25 12 PM|<video src="https://github.com/linode/manager/assets/125309814/fb4c0b2b-eb6b-4758-8dff-a519dd6bd66e" />|

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
coliu-akamai: If you uncheck the automatic assigning of an IP and type in that box, you can submit without having chosen a linode 😭

coliu-akamai: Hmm.. I'm able to trigger all three errors mentioned there by doing the following:
- 'No linode selected' - triggered when I uncheck the 'auto assign an IPv4', type in something, and try to assign the linode
- 'Selected linode must have at least one...' - choose a linode without a configuration profile
- 'No configuration profile selected' - I select a linode with multiple configurations, but do not select a configuration profile from the dropdown
- Bonus: "Too many interfaces provided for this Linode's configuration" if I assign a linode too many times (it seems to be up to 3 interfaces per config)

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
```diff
@@ -279,6 +276,22 @@ export const SubnetAssignLinodesDrawer = (
     setValues,
   ]);
 
+  React.useEffect(() => {
+    if (
+      removedLinodeId.current !== -1 &&
+      !subnet?.linodes.includes(removedLinodeId.current) &&
+      !!assignedLinodesAndConfigData.find(
+        (data) => data.id === removedLinodeId.current
+      )
```
jaalah-akamai, line 285: Ok, so this checks if the Linode being removed is both valid (i.e., exists in assignedLinodesAndConfigData) and not assigned to the subnet... and if these conditions are met, it proceeds to remove the Linode from assignedLinodesAndConfigData, correct? If so, let's just add a comment to explain this conditional check or toss it into a const that is self explanatory.

```
const isLinodeToRemoveValid = ...

if (isLinodeToRemoveValid) {
   ...
}
```

coliu-akamai: yep, that's the logic basically + will do!

packages/manager/src/cachedData/regions.json
dwiley-akamai: I think it would get rebuilt later, but Git is picking this up now as having been deleted (instead of disregarding changes to the file for diff purposes). Might have to restore it?

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
jaalah-akamai: Confirmed! Thanks for steps

jaalah-akamai: Should we even enable the button if no linode/config is selected? Just something to think about

packages/manager/src/components/RemovableSelectionsList/RemovableSelectionsList.tsx
```diff
@@ -0,0 +1,134 @@
...
+  color: theme.color.headline,
+  fontFamily: theme.font.bold,
+  fontSize: '14px',
+  textTransform: 'initial',
+}));
+
+const SelectedOptionsList = styled(List, {
+  label: 'SelectedOptionsList',
+})(({ theme }) => ({
+  background: theme.bg.main,
```
dwiley-akamai, line 114: Would we want more of a contrast in dark mode?

Screenshot 2023-09-22 at 12 45 44 PM

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
coliu-akamai: Good point - Agreed that the button definitely shouldn't be enabled if a linode isn't selected -- will add that. 

I think with the config, the check for the button being disabled would probably be something like (linodeConfigs.length > 1 && !values.selectedConfig)? lemme test this

packages/manager/src/cachedData/regions.json
coliu-akamai: Tried deleting and pushing that, but it's leading to the PR checks not passing. Looking into what I can do

packages/manager/src/components/RemovableSelectionsList/RemovableSelectionsList.tsx
jaalah-akamai: Try `theme.name === 'light' ? theme.bg.main : theme.bg.app`

coliu-akamai: image

looks a lot better, thanks!! lemme push the stuff I have up now

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
coliu-akamai: Added the disabled, logic, could you verify the following:
 - test: the button to assign a linode _should_ be disabled if a linode is not selected
- test: if a linode has multiple configs but we have not selected one of these configs, the button to assign _should_ be disabled
- test: when selecting a linode with 1 or 0 configs, the assign linode button should _not_ be disabled (and then the linode with 0 configs should error, and the linode with 1 config may error depending on if the linode has been assigned too many times)

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
```diff
@@ -0,0 +1,463 @@
...
+
+  const handleAssignLinode = async () => {
+    const { chosenIP, selectedConfig, selectedLinode } = values;
+
+    const configId = getConfigId(linodeConfigs, selectedConfig);
+
+    const interfacePayload: InterfacePayload = {
+      ipam_address: null,
+      label: null,
+      purpose: 'vpc' as InterfacePurpose,
```
dwiley-akamai, line 138: I don't think the manual typing should be necessary for `purpose`

Screenshot 2023-09-22 at 5 06 11 PM

dwiley-akamai (approved):
Went down the functionality checklist and everything LGTM ✅ 🚀 

Going ahead and approving but I did leave a few small comments

packages/manager/src/components/RemovableSelectionsList/RemovableSelectionsList.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+import { Box } from 'src/components/Box';
+import { IconButton } from 'src/components/IconButton';
+import { List } from 'src/components/List';
+import { ListItem } from 'src/components/ListItem';
+
+export type RemovableItem = {
+  id: number;
+  label: string;
+} & { [key: string]: any };
```
abailly-akamai, line 13: Any way we can improve the types here? any should be your very last resport

packages/manager/src/components/RemovableSelectionsList/RemovableSelectionsList.tsx
```diff
@@ -0,0 +1,134 @@
...
+  textTransform: 'initial',
+}));
+
+const SelectedOptionsList = styled(List, {
+  label: 'SelectedOptionsList',
+})(({ theme }) => ({
+  background: theme.name === 'light' ? theme.bg.main : theme.bg.app,
+  maxHeight: '450px',
+  maxWidth: '416px',
+  overflow: 'auto',
```
abailly-akamai, line 117: These defaults are fine but I feel like `maxWidth` and `maxHeight` should be able to be overridden by props

packages/manager/src/queries/linodes/configs.ts
```diff
@@ -29,8 +29,8 @@ export const useAllLinodeConfigsQuery = (id: number, enabled = true) => {
   );
 };
 
-const configQueryKey = 'configs';
-const interfaceQueryKey = 'interfaces';
+export const configQueryKey = 'configs';
+export const interfaceQueryKey = 'interfaces';
```
abailly-akamai, line 33: nipick: capitalize those

packages/validation/src/linodes.schema.ts
```diff
@@ -128,78 +128,75 @@ const ipv6ConfigInterface = object().when('purpose', {
...
+      message: 'Must be a valid IPv4 range, e.g. 192.0.2.0/24.',
+      test: validateIP,
+    }),
+    otherwise: string().when('ipam_address', {
+      is: null,
+      then: string().nullable(),
+      otherwise: string().test({
+        name: testnameDisallowedBasedOnPurpose('VLAN'),
+        message: testmessageDisallowedBasedOnPurpose('vlan', 'ipam_address'),
+        test: (value) => typeof value === 'undefined' || value === '',
```
abailly-akamai, line 169: Did you confirm the behaviour with undefined or empty string values?

packages/validation/src/linodes.schema.ts
```diff
@@ -128,78 +128,75 @@ const ipv6ConfigInterface = object().when('purpose', {
...
-            test: (value) => typeof value === 'undefined' || value === '',
-          }),
-        }),
+    }),
+  }),
+  ipam_address: string().when('purpose', {
+    is: 'vlan',
+    then: string().notRequired().test({
+      name: 'validateIPAM',
+      message: 'Must be a valid IPv4 range, e.g. 192.0.2.0/24.',
```
abailly-akamai, line 160: Same here as screenshot shown below. the closing period is odd when providing an ip address example. I would get rid of it

Screenshot 2023-09-25 at 09 43 47

abailly-akamai (commented):
Looks great! very nice feature and coverage.

See my comments, also adding a few things here:

1. not sure if related to your PR but autocomplete seems to be complaining about the values passed to it
Screenshot 2023-09-25 at 09 46 35

2. Available IP address `124`? is that accurate, mocked? 
Screenshot 2023-09-25 at 09 48 46

packages/manager/src/features/VPCs/VPCDetail/VPCSubnetsTable.tsx
```diff
@@ -248,6 +270,15 @@ export const VPCSubnetsTable = (props: Props) => {
...
       />
+      {subnetAssignLinodesDrawerOpen && (
+        <SubnetAssignLinodesDrawer
+          onClose={() => setSubnetAssignLinodesDrawerOpen(false)}
+          open={subnetAssignLinodesDrawerOpen}
+          subnet={selectedSubnet}
+          vpcId={vpcId}
+          vpcRegion={vpcRegion}
+        />
+      )}
```
abailly-akamai, line 281: A couple comments about this table:

1. Accessibility - "expand row" is not descriptive enough. rather, "See details about linodes attached to {subnet label}" or similar
Screenshot 2023-09-25 at 09 50 00

2. Accessibility: It is hard in this to realize this row is a child row (dependable row) since it is on the same level (and they have the same aria label - the expandable table row should have a better descriptive tag
Screenshot 2023-09-25 at 09 51 19

packages/manager/src/components/RemovableSelectionsList/RemovableSelectionsList.tsx
```diff
@@ -0,0 +1,134 @@
...
+  justifyContent: 'space-between',
+  paddingBottom: 0,
+  paddingTop: 0,
+}));
+
+const StyledLabel = styled('span', { label: 'StyledLabel' })(({ theme }) => ({
+  color: theme.color.label,
+  fontFamily: theme.font.semiBold,
+  fontSize: '14px',
+}));
```
abailly-akamai, line 134: Great job with this component!

couple things:

1. Unless my system is showing scrollbar by defaults, there's no way to know there's more items once we reach the max height. I would concert with UX about this
Screenshot 2023-09-25 at 09 56 17

2. Should this be the same treatment as a row item?
Screenshot 2023-09-25 at 09 56 56

packages/manager/src/components/RemovableSelectionsList/RemovableSelectionsList.tsx
coliu-akamai: Hm... I'd at first tried unknown, but that had led to some type errors. I feel like the other keys could realistically be anything (as simple as string, number, boolean, but as complicated as whatever data/entity types we pass into here), so I'm not too sure how feasible it'd be to make this more specific

packages/validation/src/linodes.schema.ts
coliu-akamai: (i was typing a response to this and then accidentally closed the tab 😭)
Since the ipam_address for interfaces with purpose of vpc should always be null, not too sure how to test for undefined/"" here. I can look into the other flows. Also wanted to note, this logic isn't changed at all, just moved to a different constant to allow validation for both a single interface schema and an array of them (previously we only had validation for an array of schemas)

packages/manager/src/components/RemovableSelectionsList/RemovableSelectionsList.tsx
coliu-akamai: will create a new ticket to follow up on this!

coliu-akamai:
@abailly-akamai 
- hm.. I'm getting the warnings as well, and I'm wondering if that has to do with the fact that after selecting and assigning a linode (this triggers a rerender), we remove the assigned linode from the autocomplete's options, and the autocomplete triggers that warning before we rerender. If so, getting rid of the warning may be pretty complicated, will look more into it. Also interesting to note - I don't see this warning pop up at commit `a328b4c8d703237e3cd3d72c67f5271adc18e795`, but this version has its own issue (depends on the API erroring for determining what linodes to keep track of)
- for VPCs, available IP addresses = # of total IP addresses (128 in this case) - 4 (the reserved IPs for VPC stuff). 124 seems correct based on that
- Looking into the table accessibility issues as well -- may move that to another ticket, since it seems more related to the CollapsibleTable component

packages/manager/src/features/VPCs/VPCDetail/VPCSubnetsTable.tsx
jaalah-akamai: Excellent callout, my suggestion would be to start with the unique information: "{Subnet label} then descriptive text". So a SR user can more quickly get to the important information about the subnet since the other text will be repeated per row

abailly-akamai:
@coliu-akamai thx for confirming!

> Available IP address 124? is that accurate, mocked?

Let's change to "Number of available IP Address: {x}" to make it clearer cause it could look like a part of an ip address for instance

packages/manager/src/features/VPCs/VPCDetail/VPCSubnetsTable.tsx
coliu-akamai: Added the subnet's label to the aria-label for the first bullet (so if the subnet's label is subnet-0, the aria-label would be 'subnet-0 expand row'). Making a ticket [M3-7181] to follow up on bullet 2 (and circle back to bullet 1 if needed)

packages/manager/src/queries/linodes/configs.ts
coliu-akamai: forgot to add comment - discussed via slack, since the other query keys aren't capitalized, leaving as is

abailly-akamai (approved):
Looks great thanks for the changes and the follow up tasks - nice job! not an easy one 🚢

PR #9705, opened 2023-09-20
fix: [M3-7105] - Highlight.js Theme Switching

bnussman-akamai:
## Description 📝

- Fixes bug where Highlight.js theme would not switch
  - The the would not switch because both css files were loaded at the same time. The solution involves switching out the css when the theme is changed.

## Major Changes 🔄
- Made `LinodeThemeWrapper` a named export ✏️
- Moves Highlight.js logic from the `LinodeThemeWrapper` to the `HighlightedMarkdown` so that the css is only imported when `HighlightedMarkdown` is imported. (This helps us seperate concerns and defer loading until it is really needed)
- Removed unnecessary uses of `LinodeThemeWrapper` in some unit tests 🚫
- Removed weird `setTimeout` logic that did nothing 🚫

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-09-20 at 3 30 24 PM | Screenshot 2023-09-20 at 3 30 29 PM |

## How to test 🧪

> [!important]
> Please test locally **and** using the internal preview link! We need to verify this fix works on both **dev** and **production** builds.

- Check out this PR
- Spin up your dev server with `yarn dev`
- Go to a Kubernetes Cluster details page and open the Kubeconfig drawer
- Verify that toggling the theme works
- Verify that, upon refreshing the page, the correct theme is rendered
- Repeat similar steps above with the internal preview link to verify that this fix also works for production builds of the app

### DIFF
--- a/packages/manager/package.json
+++ b/packages/manager/package.json
@@ -271,7 +271,7 @@
       "^react-native$": "react-native-web",
       "ramda": "ramda/src/index.js",
       "^src/(.*)": "<rootDir>/src/$1",
-      "\\.css\\?raw$": "<rootDir>/src/components/NullComponent",
+      "(.*)\\.css\\?raw$": "$1.css",
       "@linode/api-v4/lib(.*)$": "<rootDir>/../api-v4/src/$1",
       "@linode/validation/lib(.*)$": "<rootDir>/../validation/src/$1",
       "@linode/api-v4": "<rootDir>/../api-v4/src/index.ts",
--- a/packages/manager/package.json
+++ b/packages/manager/package.json
@@ -271,6 +271,7 @@
       "^react-native$": "react-native-web",
       "ramda": "ramda/src/index.js",
       "^src/(.*)": "<rootDir>/src/$1",
+      "\\.css\\?raw$": "<rootDir>/src/components/NullComponent",
       "@linode/api-v4/lib(.*)$": "<rootDir>/../api-v4/src/$1",
       "@linode/validation/lib(.*)$": "<rootDir>/../validation/src/$1",
       "@linode/api-v4": "<rootDir>/../api-v4/src/index.ts",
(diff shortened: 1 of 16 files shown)


packages/manager/package.json
```diff
@@ -271,6 +271,7 @@
       "^react-native$": "react-native-web",
       "ramda": "ramda/src/index.js",
       "^src/(.*)": "<rootDir>/src/$1",
+      "(.*)\\.css\\?raw$": "$1.css",
```
bnussman-akamai, line 274: Vite supports importing `highlight.js/styles/a11y-dark.css?raw` to get the CSS as a raw string but jest does not by default. This allows our unit tests to understand `.css?raw` imports.

If we were to switch to Vitest, we'd likely wouldn't need this.

packages/manager/src/LinodeThemeWrapper.tsx
```diff
@@ -1,96 +1,30 @@
...
+        {children}
+      </ThemeProvider>
     </StyledEngineProvider>
   );
 };
-
-const toggleTheme = (value: ThemeChoice) => {
-  setTimeout(() => {
-    document.body.classList.remove('no-transition');
-  }, 500);
```
bnussman-akamai, line 92: This does nothing meaningful so I'm removing it. It should also help us on the switch to Vitest if we follow through with that. (It was causing issues if I recall correctly @jdamore-linode)

packages/manager/src/env.d.ts
```diff
@@ -40,3 +40,8 @@ declare module '*.svg' {
   const src: ComponentClass<any, any>;
   export default src;
 }
+
+declare module '*.css?raw' {
+  const src: string;
+  export default src;
+}
```
bnussman-akamai, line 47: This enables typescript support for `*.css?raw` imports. This is similar to how we allow Typescript to allow `.svg` imports

jaalah-akamai (approved):
- Confirmed that markdown is rendering properly in both light/dark mode and that the theme remains unchanged on refresh
- Confirmed the same with the internal preview link

PR #9709, opened 2023-09-21
feat: [M3-7121 & M3-7122] - Support VPC in Linode Add/Edit Config dialog

hana-linode:
## Description 📝
Add support for VPCs in the Linode Add Config and Linode Edit Config dialogs, as well as within the config row in the Linode Configurations table.

## Major Changes 🔄
- `VPCPanel` component used in the Add/Edit Linode Configuration dialogs
  - Tweaks in `LinodeConfigDialog.tsx`, `InterfaceSelect.tsx`, and `VPCPanel.tsx`
- Network Interfaces column in Linode Config table now supports VPCs
  - new `InterfaceListItem` component

## How to test 🧪
With a properly-tagged account or with the VPC feature flag on:

- For both the "Add Configuration" and "Edit Configuration" dialogs:
  - The section title should be "Networking" and there should be updated helper text that reflects the VPC option
  -  There should be a "Primary Interface (Default Route)" dropdown
  - You should be able to successfully create a config with the designated interfaces
  - Field errors should be surfaced (example: uncheck the "Auto-assign a VPC IPv4 [...]" checkbox, input an invalid VPC IPv4, and observe the field error)
  - The dialog should reset when closed and re-opened

- For the "Edit Configuration" dialog:
  - Ensure the "Primary Interface" field displays the correct primary interface (the interface designated as "primary" based on the API payload should be in the eth position indicated by the "Primary Interface" field)
  - Changing the primary interface gets reflected in the payload submitted to the API, and after a successful submission, opening the dialog again for that config should reflect the changes you just made

None of the above should be viewable/possible without the proper account tags or the VPC feature flag being on.

Because the `<VPCPanel />` component was changed to accommodate the Add/Edit Config use case, ensure there were no regressions to its behavior in the Linode Create flow.


packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx
```diff
@@ -853,23 +933,54 @@ export const LinodeConfigDialog = (props: Props) => {
...
+                      }}
                       handleChange={(newInterface: Interface) =>
                         handleInterfaceChange(idx, newInterface)
                       }
-                      ipamError={
-                        formik.errors[`interfaces[${idx}].ipam_address`]
-                      }
                       ipamAddress={thisInterface.ipam_address}
                       key={`eth${idx}-interface`}
                       label={thisInterface.label}
```
dwiley-akamai, line 976: (in the Edit dialog when it is first opened) -- working through a bug where the actual VLAN label isn't coming through reliably here, causing the VLAN dropdown to show "Create or select a VLAN" instead of having the existing VLAN show up as selected 🤔

dwiley-akamai: it's due to the `useEffect` logic in `InterfaceSelect.tsx` -- the VLAN `label` is being overwritten

dwiley-akamai:
Observed some instances of errors not surfacing, so the error mapping still needs tweaking

dwiley-akamai:
Going to go ahead and mark this as ready for review now. We are aware of two somewhat edge case errors that don't get surfaced currently for `ipv4.vpc` and `ipv4.nat_1_1`. We will try to get a fix for that into this PR but otherwise, it will be addressed in the coming days in a follow-up PR; this PR represents the last core piece of functionality for the full VPC flows in Cloud.

coliu-akamai:
If I uncheck the "Auto-assign a VPC IPv4 [...]" checkbox but do not input any IPv4, what's the expected behavior? Currently it creates (or edits) the config:

https://github.com/linode/manager/assets/139280159/5db1d57f-c5d8-42f1-a297-4409ac0ec1c0

coliu-akamai:
For spacing on smaller viewports, the eth0-2 fields get smaller than the primary field, and the margin between the eth-x and vpc field disappears. The tooltip also gets cutoff:

image

this could probably be moved to a later ticket if needed

coliu-akamai:
For some reason I'm no longer able to assign a linode to a subnet via the SubnetAssignLinode drawer -- will look into this one shortly 😖 
image

packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx
```diff
@@ -165,23 +168,51 @@ const interfacesToState = (interfaces?: Interface[]) => {
...
+    }) => ({
+      id,
+      ipam_address,
+      ipv4,
+      label,
+      primary,
+      purpose,
+      subnet_id,
+      vpc_id,
+    })
```
coliu-akamai, line 189: nitpick: I see what we're trying to do here -- maybe worth excluding the two properties and then just using rest? (I think these are the properties we want)
```suggestion
    ({
      ipv6, ip_ranges, ...rest
    }) => rest
```

coliu-akamai (commented):
great work!! left a few comments above. Did a first pass through the code, will be doing another one soon

dwiley-akamai:
> If I uncheck the "Auto-assign a VPC IPv4 [...]" checkbox but do not input any IPv4, what's the expected behavior? Currently it creates (or edits) the config:
> 
> Oh is this one of the errors that don't get surfaced correctly?
> 
>  Screen.Recording.2023-10-05.at.11.21.28.AM.mov 
> followup: when creating a Linode and leaving that blank, an error surfaces for a blank IPv4: image

The latter case is because we're able to easily see if the checkbox is unchecked and if there's a blank VPC IPv4 and set the error accordingly (L758-L773 in `LinodeCreateContainter.tsx`). There's only one pair of those fields in the Linode Create flow, whereas here we can have multiple pairs of those fields in the form (one pair per interface)

packages/api-v4/src/linodes/types.ts
```diff
@@ -266,7 +266,7 @@ export interface Disk {
 export type DiskStatus = 'ready' | 'not ready' | 'deleting';
 
 export interface LinodeConfigCreationData {
-  label: string;
+  label?: string;
```
abailly-akamai, line 269: Can you explain these this change? This props should remain required here and remain true to the API type. 

https://www.linode.com/docs/api/linode-instances/#configuration-profile-create

Screenshot 2023-10-06 at 09 47 08

If this is a case of an API request that can be undefined it should be handled in the component making the request (early return)

packages/api-v4/src/linodes/types.ts
```diff
@@ -165,8 +165,8 @@ export interface Interface {
   purpose: InterfacePurpose;
   ipam_address: string | null;
   primary?: boolean;
-  subnet_id?: number | null;
-  vpc_id?: number | null;
+  subnet_id?: number;
+  vpc_id?: number;
```
abailly-akamai, line 169: Please confirm these fields are indeed not required (can be undefined) VS being required and nullable.

abailly-akamai (changes requested):

dwiley-akamai:
For endpoints that take a list of interfaces, if one of the interfaces has an API error, that should be indicated in the `field` property -- e.g., `interfaces[0].ipv4.vpc` instead of `ipv4.vpc`. API has a bug ticket to address this and it should be done in the near future. In the meantime, I pushed up some code to surface those errors in a non-ideal capacity, just so they don't stay submerged until that fix is in. I'll make a follow-up ticket now for cleaning that up when the API fix is in.

With regards to the error when you uncheck the VPC IPv4 autoassign checkbox but don't input a VPC IPv4: I am going to make a separate ticket to handle that situation and will work on that today.

packages/api-v4/src/linodes/types.ts
hana-linode: @abailly-akamai This was a remnant of a previous implementation, will revert this change

packages/api-v4/src/linodes/types.ts
hana-linode: Same here, remnant of a previous implementation. Confirmed via the API Spec docs that these are required and nullable

packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx
dwiley-akamai: Going to keep this as-is for now since the linter will give an unused variable warning for the two unused ones currently, but once VPC is in GA and `ipv6` and `ip_ranges` are supported down the line we should be able to remove this bit entirely (just padding `interfaces` without doing any mapping).

coliu-akamai:
Functionality from above that I checked off still looks good + the error on the Subnet Assign Linodes Drawer is fixed and I saw the tickets for the other error surfacing issues 🚀 

One other thing I'm noticing is that when we edit a Linode's configuration details to include or delete a VPC related interface, the configuration will update, but other information won't until we refresh - something to do with query invalidation/the query cache, most likely?

https://github.com/linode/manager/assets/139280159/9150f54e-7507-4afc-a172-7eddb6d055c7

(Actually this reminds me about the issue for freshly deleted linodes still appearing in VPCs -- let me make a ticket for that so it doesn't get lost 😅)

coliu-akamai (approved):
awesome work Hana and Dajahi!

approving pending query caching issue is resolved (moved to a new ticket or handled in this PR)!

packages/manager/src/features/Linodes/LinodesCreate/VPCPanel.tsx
```diff
@@ -23,20 +25,23 @@ import scrollErrorIntoView from 'src/utilities/scrollErrorIntoView';
...
+  from: 'linodeConfig' | 'linodeCreate';
   handleSelectVPC: (vpcId: number) => void;
   handleSubnetChange: (subnetId: number) => void;
   handleVPCIPv4Change: (IPv4: string) => void;
+  nat_1_1Error?: string;
   region: string | undefined;
-  selectedSubnetId: number | undefined;
-  selectedVPCId: number | undefined;
+  selectedSubnetId?: null | number;
+  selectedVPCId?: null | number;
```
abailly-akamai, line 38: Can we confirm these are not required fields VS being required and nullable?

packages/manager/src/features/Linodes/LinodesCreate/VPCPanel.tsx
```diff
@@ -268,6 +306,15 @@ export const VPCPanel = (props: VPCPanelProps) => {
                   </Box>
                 }
               />
+              {assignPublicIPv4Address && nat_1_1Error && (
```
abailly-akamai, line 309: how can we test this error?

also, let's update this variable name so it's human readable, it's pretty unclear from reading the code

abailly-akamai (commented):
Nice work! things are working 

I tested both flows (edit and create config) as we will as the linode creation flow and the form behaviors work well. Left a couple comments to clarify and clean up a couple things.

Also, there is no test coverage added here although we are modifying core flows and adding/modifying many components. I know there must be upcoming e2e tests covering some of those cases but we should also have unit tests testing the rendering and some of the utilities added.

packages/manager/src/features/Linodes/LinodesCreate/VPCPanel.tsx
```diff
@@ -121,26 +137,47 @@ export const VPCPanel = (props: VPCPanelProps) => {
...
+    }
+
+    const copy =
+      vpcDropdownOptions.length <= 1
+        ? 'Allow Linode to communicate in an isolated environment.'
+        : 'Assign this Linode to an existing VPC.';
+
+    return (
+      <>
+        {copy} <Link to="">Learn more</Link>.
```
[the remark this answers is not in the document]
dwiley-akamai: The corresponding to-do is on L179, but I'll move it here to be more direct.

packages/manager/src/features/Linodes/LinodesCreate/VPCPanel.tsx
dwiley-akamai: If the linode doesn't have an unused public IP available and the "Assign a public IPv4 address for this Linode" checkbox is checked, the following error gets returned:

Screenshot 2023-10-03 at 2 24 44 PM

One way to get it to be returned is to have an interface with 2 VPC slots and checking off that checkbox for both of them. You need an extra tag to be able to get to that state.

I'll rename the variable to `publicIPv4Error` or something to that effect.

jaalah-akamai:
✅ For both the "Add Configuration" and "Edit Configuration" dialogs:

- [x] The section title should be "Networking" and there should be updated helper text that reflects the VPC option
- [x] There should be a "Primary Interface (Default Route)" dropdown
- [x] You should be able to successfully create a config with the designated interfaces
- [x] Field errors should be surfaced (example: uncheck the "Auto-assign a VPC IPv4 [...]" checkbox, input an invalid VPC IPv4, and observe the field error)
- [x] The dialog should reset when closed and re-opened

✅ For the "Edit Configuration" dialog:

- [x] Ensure the "Primary Interface" field displays the correct primary interface (the interface designated as "primary" based on the API payload should be in the eth position indicated by the "Primary Interface" field)
- [x] Changing the primary interface gets reflected in the payload submitted to the API, and after a successful submission, opening the dialog again for that config should reflect the changes you just made

jaalah-akamai (approved):

packages/manager/src/features/Linodes/LinodesCreate/VPCPanel.tsx
dwiley-akamai: Strictly speaking they are not optional for this component so I removed the `?` which suggests they are.

abailly-akamai (approved):
Thanks for addressing the feedback and adding coverage - looks great!

Functionality and UI are good as far as my testing went. ✅ 
E2E failure seems unrelated

PR #9723, opened 2023-09-26
feat: [M3-7160] - Add AGLB "Edit Certificate" drawer

mjac0bs:
## Description 📝
Add an edit drawer for AGLB certificates.

## Major Changes 🔄
- Creates the Edit Certificate drawer with dynamic fields depending on certificate type
- Adds the useLoadBalancerCertificateMutation
- Adds a unit test for `EditCertificateDrawer` component
- Updates the MSW `PUT` endpoint for `*/v4beta/aglb/:id/certificates/:certId` to return a `downstream` cert in addition to the existing `ca` certs.

## Preview 📷
| Service Target Certificates  | TLS Certificates   |
| ------- | ------- |
|  Screenshot 2023-10-05 at 3 30 54 PM | Screenshot 2023-10-05 at 3 30 38 PM |

## How to test 🧪
- Check out this PR.
- `yarn dev`
- Ensure AGLB feature flag is **on** and MSW is **enabled**.
2. **How to verify changes?**
- Go to http://localhost:3000/loadbalancers/0/certificates/downstream.
- Click on the Edit button in the action menu of the `tls-certificate`. Confirm that the edit drawer matches the mocks above for a TLS certificate.
- Click on the Edit button in the action menu of any other certificate (they're all of type `ca`). Confirm that the edit drawer matches the mocks above for a Service Target certificate.
- Confirm that a `PUT` request is sent by the MSW when the Update Certificate button is clicked.
- Confirm that no `PUT` request is sent by the MSW when the drawer is Canceled or closed out of.
- Confirm form validation:
   - If a user edits the `label` of either certificate type, the form submits without error.
   - If a user edits the `certificate` of a Service Target (`ca`) cert, the form submits without error.
   -  If a user edits the `certificate` of a TLS (`downstream`) cert and submits without a key, they should receive a validation error upon submission  indicating `Private Key is required` in that case.
Unit:
```
yarn test EditCertificateDrawer
```
E2E:
Will be completed in M3-7134, since there was an existing ticket created for this.

### DIFF
--- a/packages/api-v4/src/aglb/types.ts
+++ b/packages/api-v4/src/aglb/types.ts
@@ -141,6 +141,7 @@ type CertificateType = 'ca' | 'downstream';
 export interface Certificate {
   id: number;
   label: string;
+  certificate: string;
   type: CertificateType;
 }
 
--- a/packages/api-v4/src/aglb/types.ts
+++ b/packages/api-v4/src/aglb/types.ts
@@ -150,3 +150,10 @@ export interface CreateCertificatePayload {
   label: string;
   type: CertificateType;
 }
+
+export interface UpdateCertificatePayload {
+  key?: string;
+  certificate?: string;
+  label?: string;
+  type?: CertificateType;
+}
--- a/packages/manager/src/features/LoadBalancers/LoadBalancerDetail/Certificates/Certificates.tsx
+++ b/packages/manager/src/features/LoadBalancers/LoadBalancerDetail/Certificates/Certificates.tsx
@@ -22,6 +22,7 @@ import { useLoadBalancerCertificatesQuery } from 'src/queries/aglb/certificates'
 
 import { CreateCertificateDrawer } from './CreateCertificateDrawer';
 import { DeleteCertificateDialog } from './DeleteCertificateDialog';
+import { EditCertificateDrawer } from './EditCertificateDrawer';
 
 import type { Certificate, Filter } from '@linode/api-v4';
 
@@ -40,6 +41,7 @@ export const Certificates = () => {
   const history = useHistory();
 
   const isCreateDrawerOpen = location.pathname.endsWith('/create');
+  const [isEditDrawerOpen, setIsEditDrawerOpen] = useState(false);
   const [isDeleteDrawerOpen, setIsDeleteDrawerOpen] = useState(false);
 
   const [selectedCertificateId, setSelectedCertificateId] = useState<number>();
@@ -147,7 +149,7 @@ export const Certificates = () => {
               <TableCell actionCell>
                 <ActionMenu
                   actionsList={[
-                    { onClick: () => null, title: 'Edit' },
+                    { onClick: () => setIsEditDrawerOpen(true), title: 'Edit' },
                     {
                       onClick: () => onDeleteCertificate(certificate),
                       title: 'Delete',
@@ -174,6 +176,12 @@ export const Certificates = () => {
         loadbalancerId={id}
         open={isCreateDrawerOpen}
         type={certType}
+      />
+      <EditCertificateDrawer
+        certificateId={selectedCertificateId}
+        loadbalancerId={id}
+        onClose={() => setIsEditDrawerOpen(false)}
+        open={isEditDrawerOpen}
       />
       <DeleteCertificateDialog
         certificate={selectedCertificate}
--- a/packages/manager/src/features/LoadBalancers/LoadBalancerDetail/Certificates/Certificates.tsx
+++ b/packages/manager/src/features/LoadBalancers/LoadBalancerDetail/Certificates/Certificates.tsx
@@ -71,6 +71,11 @@ export const Certificates = () => {
     filter
   );
 
+  const onEditCertificate = (certificate: Certificate) => {
+    setIsEditDrawerOpen(true);
+    setSelectedCertificateId(certificate.id);
+  };
+
   const onDeleteCertificate = (certificate: Certificate) => {
     setIsDeleteDrawerOpen(true);
     setSelectedCertificateId(certificate.id);
@@ -149,7 +154,10 @@ export const Certificates = () => {
               <TableCell actionCell>
                 <ActionMenu
                   actionsList={[
-                    { onClick: () => setIsEditDrawerOpen(true), title: 'Edit' },
+                    {
+                      onClick: () => onEditCertificate(certificate),
+                      title: 'Edit',
+                    },
                     {
                       onClick: () => onDeleteCertificate(certificate),
                       title: 'Delete',
@@ -178,7 +186,7 @@ export const Certificates = () => {
         type={certType}
       />
       <EditCertificateDrawer
-        certificateId={selectedCertificateId}
+        certificate={selectedCertificate}
         loadbalancerId={id}
         onClose={() => setIsEditDrawerOpen(false)}
         open={isEditDrawerOpen}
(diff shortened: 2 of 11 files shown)


packages/manager/src/features/LoadBalancers/LoadBalancerDetail/Certificates/EditCertificateDrawer.tsx
```diff
@@ -0,0 +1,121 @@
...
+            labelTooltipText="TODO: AGLB"
+            multiline
+            name="certificate"
+            onChange={formik.handleChange}
+            trimmed
+            value={formik.values.certificate}
+          />
+          {certificate?.type === 'downstream' && (
+            <TextField
+              errorText={errorMap.key}
```
mjac0bs, line 100: @bnussman-akamai - Do we want input validation on the form if a user tries to submit a `downstream` cert without a key, or are we relying on the API to return an error, which we'll display here after a failed `PUT` request?

If we want validation, do you know of a similar place where we do that? Something that wasn't right: 
```
export const UpdateCertificateSchema = object({
   key: string().when(['type', 'certificate'], {
   is: (type: string, certificate: string) =>
   type === 'downstream' && certificate === '',
   then: string().required('Private Key is required'),
   }),
});
```

packages/manager/src/features/LoadBalancers/LoadBalancerDetail/Certificates/EditCertificateDrawer.tsx
```diff
@@ -0,0 +1,121 @@
...
+            errorText={errorMap.label}
+            label="Certificate Label"
+            name="label"
+            onChange={formik.handleChange}
+            value={formik.values.label}
+          />
+          <TextField
+            errorText={errorMap.certificate}
+            label={labelMap[certificate.type]}
+            labelTooltipText="TODO: AGLB"
```
mjac0bs, line 91: I'm not sure what the plan is for the copy in these tooltips, but given that we are not able to retrieve or display a certificate object's `certificate` or `key` fields for security reasons, we should probably mention that in this form so that the user does not think that their cert/key data is missing.

packages/manager/src/features/LoadBalancers/LoadBalancerDetail/Certificates/Certificates.tsx
```diff
@@ -40,6 +41,7 @@ export const Certificates = () => {
   const history = useHistory();
 
   const isCreateDrawerOpen = location.pathname.endsWith('/create');
+  const [isEditDrawerOpen, setIsEditDrawerOpen] = useState(false);
```
mjac0bs, line 44: @bnussman-akamai - Is this the desired way to determine the state of the drawer, or are we expecting to append `/{certificateId}/edit` to the URL?

bnussman-akamai: This would be cool but we don't have a great pattern or way to do it that I know of. The challenge here would be if a user navigated to `{certificateId}/edit` and the certificate isn't on the first page of the paginated data. We'd have to perform a fetch because we can't guarantee that the certificate is fetched if we paginate the certificates landing table.

mjac0bs: > We'd have to perform a fetch because we can't guarantee that the certificate is fetched if we paginate the certificates landing table.

Gotcha, yep. Will leave as is.

packages/manager/src/features/LoadBalancers/LoadBalancerDetail/Certificates/EditCertificateDrawer.tsx
bnussman-akamai: > Do we want input validation on the form if a user tries to submit a downstream cert without a key

Yeah, we probably should enforce that a key is provided for a `downstream` cert

For now I'd say just follow how `packages/manager/src/features/LoadBalancers/LoadBalancerDetail/Certificates/CreateCertificateDrawer.tsx` handles validation errors.

I'm building with a focus on API error handling because we're going to have some very complex forms and I want API errors to be the focus and source of truth.

mjac0bs: As it turns out, when you create a schema, it's most effective if you actually then *use* that schema. 🤦🏼‍♀️  I was missing that the first time around -- fixed those issues in [bb385ad](https://github.com/linode/manager/pull/9723/commits/bb385adbacbd6388f7a091a10ce31d0d96182827) and [3a896c4](https://github.com/linode/manager/pull/9723/commits/3a896c42c95bc18f4fbd55c70b0e0ce1acc33e6b).

Should be working as expected now:
image

bnussman-akamai (commented):
Looks great! My only concerns regard consistency between the Create Drawer and this new Edit Drawer. We may need to ask Daniel for some clarification.

- In the Create Service Target Cert drawer we say "TLS Certificate" but in the edit drawer we say "Service Certificate". I guess we should say "Service Certificate" in both places?
- We have placeholders in the Create drawer but not the edit drawer

mjac0bs:
@bnussman-akamai 

> * In the Create Service Target Cert drawer we say "TLS Certificate" but in the edit drawer we say "Service Certificate". I guess we should say "Service Certificate" in both places?

Good catch on this! 🔍 Asked Daniel for clarification in a comment on the mocks and reached out to let him know. 

> * We have placeholders in the Create drawer but not the edit drawer

Yeah, I'm not sure what's best here because we could use the same example placeholder as the Create drawer, but that might face the same problem as leaving it blank -- causing users to think that it is missing/incomplete, when really we're just not able to display the data. I also asked Daniel about this in the mocks.

packages/manager/src/features/LoadBalancers/LoadBalancerDetail/Certificates/CreateCertificateDrawer.tsx
```diff
@@ -91,7 +93,7 @@ export const CreateCertificateDrawer = (props: Props) => {
         />
         <TextField
           errorText={errorMap.certificate}
-          label="TLS Certificate"
+          label={labelMap[type]}
```
mjac0bs, line 96: > In the Create Service Target Cert drawer we say "TLS Certificate" but in the edit drawer we say "Service Certificate". I guess we should say "Service Certificate" in both places?

Daniel updated to mocks so that the Create drawer now labels the field as "Server Certificate" for ca/service target certificates and is "TLS Certificate" for downstream/TLS, so I made the change to dynamically render the label based on `type` in the Create drawer here.

packages/api-v4/src/aglb/types.ts
```diff
@@ -141,6 +141,7 @@ type CertificateType = 'ca' | 'downstream';
 export interface Certificate {
   id: number;
   label: string;
+  certificate: string;
```
mjac0bs, line 144: The API spec has been updated to indicate that `certificate` will now be returned in `GET /v4beta/aglb/{id}/certificates`.

bnussman-akamai (commented):
When we make the `PUT`, I think we may want to omit `certificate` and `key` from the body if they are unchanged. I worry this might make the API clear out the `certificate` and `key` if we send an empty string or perhaps cause a validation error.

Screenshot 2023-10-04 at 1 48 05 PM

bnussman-akamai (approved):
Great job! 🎉

cpathipa (approved):
Over all looks great! left couple comments for minor suggestions.
Validated 
-  Edit Certificate drawer with dynamic fields depending on certificate type
-  test coverage for EditCertificateDrawer

mjac0bs:
@bnussman-akamai , @cpathipa 

Thanks for the reviews! I'll be merging this shortly. In my last commit, I addressed UX feedback to make the cert field contain the actual cert value instead of placeholder text. We now check whether the user is updating the cert from its original value and, if they are, require them to provide a key (as per the API spec). (Validation seems to work correctly, as do request payloads; I think when I tried this initially, I was having unrelated validation issues.)

Creating follow up tickets to address additional UX feedback:
- display the cert id in the table

mjac0bs:
One relevant e2e failure fixed due to the typo correction in the Create Cert drawer; this test now passes:
image

PR #9725, opened 2023-09-27
fix: [M3-7038] - Add Firewall and Linode Configuration success toast notifications

tyler-akamai:
## Description 📝
The Firewall and Linode Configuration toast notifications did not have a success message, causing no toast notifications when an action was successful.

## Major Changes 🔄
- Added success notifications for
   - Firewall notifications (enable, disable, create, delete, device add, device remove)
   - Linode configurations (create, delete, update)
- Updated toast notification messages for consistency

## Preview 📷

<details>
<summary>Before</summary>
<video src="https://github.com/linode/manager/assets/139489745/3a840a5a-5444-47e0-8ea9-5411e4aa58c0" />
</details>

<details>
<summary>After</summary>

Screenshot 2023-09-27 at 10 30 38 AM

Screenshot 2023-09-27 at 10 29 51 AM

Screenshot 2023-09-27 at 10 30 05 AM

Screenshot 2023-09-27 at 10 31 52 AM

Screenshot 2023-09-27 at 10 32 09 AM

Screenshot 2023-09-27 at 10 32 41 AM

Screenshot 2023-09-27 at 10 30 22 AM

Screenshot 2023-09-27 at 10 36 05 AM

</details>

## How to test 🧪

*** Note: The toast notifications take a couple seconds to pop up.

Navigate to the firewall landing page (/firewalls) and ensure:
- [ ] Creating a Firewall triggers a successful toast notification
- [ ] Disabling a Firewall triggers a successful toast notification
- [ ] Enabling a Firewall triggers a successful toast notification
- [ ] Deleting a Firewall triggers a successful toast notification

Navigate to the linodes tab within any firewall details page (/firewalls/[firewall_ID]/linodes) and ensure:
- [ ] Adding a Linode triggers a successful toast notification
- [ ] Deleting a Linode triggers a successful toast notification

Navigate to the config tab within any linode details page (/linodes/[linode_id]/configurations) and ensure:
- [ ] Adding a Linode configuration triggers a successful toast notification
- [ ] Deleting a Linode configuration triggers a successful toast notification
- [ ] Editing a Linode configuration triggers a successful toast notification


abailly-akamai (commented):
Thanks for working on this!

Do you know why we're getting those delays (VS creating or deleting a Linode for instance)? Pretty important IMO because the toast ends up being a bit distracting.  It happens a few seconds afterwards (that was my experience at least, sometimes close to 5/6 seconds) and it's a bit odd since we can clearly see the UI update (ex row gets deleted), then the notification happens much later. 

Considering the UI updates + the notification menu event notification, I wonder about the worth of this update if the delay is that long

bnussman-akamai (commented):
I feel like `linode_config*`, `firewall_enable`, `firewall_disable`, `firewall_create`, `firewall_device_add`, `firewall_device_remove` toasts should just be done on the initial request, not based on events polling. I'd have to look at the API code and see what actually takes place when some of these events are created but I feel like most of the events I listed above don't need to wait on an event to show, This is mostly my opinion. Anyone else have thoughts?

abailly-akamai:
@bnussman-akamai ah yeah that makes sense, they're based on event polling. Agree 💯

tyler-akamai:
Awesome, will change that over now.

packages/manager/src/features/Firewalls/FirewallLanding/FirewallDialog.tsx
```diff
@@ -54,8 +57,17 @@ export const FirewallDialog = React.memo((props: Props) => {
...
   const onSubmit = async () => {
-    await requestMap[mode]();
-    onClose();
+    try {
+      await requestMap[mode]();
+      enqueueSnackbar(`Successfully ${mode}d ${label}`, {
+        variant: 'success',
+      });
+      onClose();
+    } catch (e) {
```
abailly-akamai, line 66: I noticed you're not doing anything with the actual error. Is this intentional? Do we want to provide better information to the user rather than a generic message?

abailly-akamai (commented):
Thanks for making those changes! Feels much better.

for consistency, in the firewall flows, do we also want to add toasts when adding inbound/outbound rules? /firewalls/{id}/rules

packages/manager/src/features/Firewalls/FirewallLanding/FirewallDialog.tsx
tyler-akamai: Yeah I was unsure if API errors belong in toast notifications? My concern is that they might contain too much text.

tyler-akamai: What are your thoughts?

packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRulesLanding.tsx
```diff
@@ -191,6 +195,9 @@ const FirewallRulesLanding = (props: Props) => {
         // Reset editor state.
         inboundDispatch({ rules: _rules.inbound ?? [], type: 'RESET' });
         outboundDispatch({ rules: _rules.outbound ?? [], type: 'RESET' });
+        enqueueSnackbar('Successfully updated Firewall rules', {
```
tyler-akamai, line 198: The firewalls section is a little odd. Instead of having an api call for each action (edit, delete, add, clone), the user first executes an action and then when they hit 'Save Changes', the api is called. This makes my job a little easier with the Toast Notifications since they are all routed through the function `applyChanges`

tyler-akamai: @abailly-akamai

packages/manager/src/features/Firewalls/FirewallLanding/FirewallDialog.tsx
tyler-akamai: Also, I added toast notifications to the inbound/outbound rules, I tagged you in a comment explaining the code.

abailly-akamai: > What are your thoughts? Because I do agree it would be helpful for the user to have a more detailed error message.

I would mirror what we are already doing everywhere else.

tyler-akamai: Just added the caught error to the error map. This error then gets passed into the <ConfirmationDialog /> which prints it out at the top.

packages/manager/src/features/Firewalls/FirewallDetail/Devices/RemoveDeviceDialog.tsx
```diff
@@ -37,7 +40,16 @@ export const RemoveDeviceDialog = React.memo((props: Props) => {
...
-    await mutateAsync();
+    try {
+      await mutateAsync();
+      enqueueSnackbar(`Successfully removed ${device?.entity.label}`, {
+        variant: 'success',
+      });
+    } catch {
+      enqueueSnackbar(`Failed to remove ${device?.entity.label}`, {
+        variant: 'error',
+      });
```
bnussman-akamai, line 51: I don't think we need an error toast here. For 99.9% of our Dialogs, we use the `error` prop, which displays the error directly in the dialog. Therefore, we also want to make sure the Dialog does not close if there is an error.

packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/DeleteConfigDialog.tsx
bnussman-akamai, line 1: Same feedback here. The error is already shown in the dialog by `error={error?.[0].reason}`. We don't need an error toast here, just a success toast.

packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx
```diff
@@ -342,6 +349,11 @@ export const LinodeConfigDialog = (props: Props) => {
             err.field = 'devices';
           }
         });
+
+        const actionType = isCreate ? 'create' : 'update';
+        enqueueSnackbar(`Failed to ${actionType} ${configData.label}`, {
+          variant: 'error',
+        });
```
bnussman-akamai, line 356: Instead of passing `isCreate` as a parameter, can we just derive the value inside the function? I think using less parameters makes the code a bit cleaner.  


Maybe something like:
```ts
const isCreate = Boolean(config);
```

packages/manager/src/features/Firewalls/FirewallDetail/Devices/AddDeviceDrawer.tsx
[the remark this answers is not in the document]
tyler-akamai: Good point, this is fixed in the firewall-nodebalancer branch, so I will revert the file.

packages/manager/src/features/Firewalls/FirewallLanding/CreateFirewallDrawer.tsx
```diff
@@ -50,6 +51,8 @@ export const CreateFirewallDrawer = React.memo(
 
     const { mutateAsync } = useCreateFirewall();
 
+    const { enqueueSnackbar } = useSnackbar();
```
tyler-akamai, line 54: This also will conflict with the new firewall-nodebalancer changes

packages/manager/src/features/Firewalls/FirewallDetail/Rules/FirewallRulesLanding.tsx
```diff
@@ -218,6 +225,9 @@ const FirewallRulesLanding = (props: Props) => {
             });
           }
         }
+        enqueueSnackbar('Failed to update Firewall rules', {
```
tyler-akamai, line 228: Kept this error because there is no "Save Changes" dialog

packages/manager/src/features/Firewalls/FirewallLanding/FirewallDialog.tsx
```diff
@@ -54,8 +57,15 @@ export const FirewallDialog = React.memo((props: Props) => {
...
+      errorMap[mode] = e;
+    }
```
tyler-akamai, line 68: Removed the error toast notification and passed the error into the errorMap

packages/manager/src/features/Firewalls/FirewallLanding/FirewallDialog.tsx
tyler-akamai: removed unnecessary error toast notification.

packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx
tyler-akamai: refer to line 323

abailly-akamai (commented):
Creating a Firewall triggers a successful toast notification 🚫 
Disabling a Firewall triggers a successful toast notification ✅ 
Enabling a Firewall triggers a successful toast notification ✅
Deleting a Firewall triggers a successful toast notification ✅

Adding a Linode triggers a successful toast notification 🚫 
Deleting a Linode triggers a successful toast notification ✅

Adding a Linode configuration triggers a successful toast notification ✅ 
Deleting a Linode configuration triggers a successful toast notification ✅ 
Editing a Linode configuration triggers a successful toast notification ✅

packages/manager/src/features/Firewalls/FirewallDetail/Devices/RemoveDeviceDialog.tsx
```diff
@@ -37,17 +40,21 @@ export const RemoveDeviceDialog = React.memo((props: Props) => {
...
+      // Since the linode was removed as a device, invalidate the linode-specific firewall query
+      queryClient.invalidateQueries([
+        linodesQueryKey,
+        'linode',
+        linodeId,
+        'firewalls',
+      ]);
 
-    onClose();
+      onClose();
```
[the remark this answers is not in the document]
bnussman-akamai: Could we clean this up still? Should we put `queryClient.invalidateQueries` in the `then`? Should we put `onClose` in the `then` too? Should we revert to the await syntax?

Right now, it's unclear to me why `enqueueSnackbar` is in the `then` but `queryClient.invalidateQueries` and `onClose` are not

packages/manager/src/features/Firewalls/FirewallLanding/FirewallDialog.tsx
```diff
@@ -54,8 +57,15 @@ export const FirewallDialog = React.memo((props: Props) => {
...
-    await requestMap[mode]();
-    onClose();
+    try {
+      await requestMap[mode]();
+      enqueueSnackbar(`Successfully ${mode}d ${label}`, {
+        variant: 'success',
+      });
+      onClose();
+    } catch (e) {
+      errorMap[mode] = e;
```
bnussman-akamai, line 67: We don't need to update `errorMap`. It's values are already defined and will update during render if errors come down from React Query

```suggestion
```

packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/DeleteConfigDialog.tsx
```diff
@@ -25,7 +25,9 @@ export const DeleteConfigDialog = (props: Props) => {
 
   const onDelete = async () => {
     await mutateAsync();
-    enqueueSnackbar('Successfully deleted config', { variant: 'success' });
+    enqueueSnackbar(`Successfully deleted ${config?.label}`, {
```
bnussman-akamai, line 28: I wonder if we should say the entity name: `Successfully deleted Configuration ${config?.label}`

packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx
```diff
@@ -342,6 +350,10 @@ export const LinodeConfigDialog = (props: Props) => {
             err.field = 'devices';
           }
         });
+
+        enqueueSnackbar(`Failed to ${actionType} ${configData.label}`, {
+          variant: 'error',
+        });
```
bnussman-akamai, line 356: Do we want both inline error and a toast error here? Maybe we just don't need an error toast for this dialog?


https://github.com/linode/manager/assets/115251059/3ea57430-9ce2-45f1-a599-8556d55675b2

packages/manager/src/features/Firewalls/FirewallDetail/Devices/RemoveDeviceDialog.tsx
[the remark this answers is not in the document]
tyler-akamai: `queryClient.invalidateQueries` and `onClose` are in the `then` statement.

I can go back to the `await` syntax though to make it more clear.

packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx
tyler-akamai: Good catch!

tyler-akamai:
> Creating a Firewall triggers a successful toast notification 🚫 Disabling a Firewall triggers a successful toast notification ✅ Enabling a Firewall triggers a successful toast notification ✅ Deleting a Firewall triggers a successful toast notification ✅
> 
> Adding a Linode triggers a successful toast notification 🚫 Deleting a Linode triggers a successful toast notification ✅
> 
> Adding a Linode configuration triggers a successful toast notification ✅ Deleting a Linode configuration triggers a successful toast notification ✅ Editing a Linode configuration triggers a successful toast notification ✅

The two tests that failed are corrected in the nodebalancer-firewall feature branch. Banks left a comment pointing out that the changes conflict with the feature branch. 

So the functionality will be added shortly once the feature branch is merged.

abailly-akamai (approved):
Approving based on latest comments about expected cases to be handled in a different PR

packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx
bnussman-akamai, line 1: Looks like we might need to update an e2e test

Screenshot 2023-10-06 at 1 43 22 PM

packages/manager/cypress/e2e/core/linodes/clone-linode.spec.ts
```diff
@@ -116,7 +116,7 @@ describe('clone linode', () => {
 
       ui.toast.assertMessage(`Your Linode ${newLinodeLabel} is being created.`);
       ui.toast.assertMessage(
-        `Linode ${linode.label} has been cloned successfully to ${newLinodeLabel}.`
+        `Linode ${linode.label} successfully cloned to ${newLinodeLabel}.`
```
tyler-akamai, line 119: I had to update the E2E tests to match the toast notification messages.

bnussman-akamai (approved):
Looks good! Thanks for resolving the e2e failures! 🎉

PR #9731, opened 2023-09-28
fix: [M3-7072] - Long drawer titles overlapping close icon

mjac0bs:
## Description 📝
This PR fixes an edge case where very long `Drawer` titles (in this case, due to long firewall names) overlap the `Drawer` close (X) icon.

Note: we fixed a similar issue before with Dialogs, not Drawers.

## Major Changes 🔄
- Adds right margin to the drawer title to prevent overlap with close icon
- Reorders props to resolve eslint perfectionist warnings
- Fixes a typo that appears in the component's Storybook documentation

## Preview 📷

| Before | After |
| ---- | ---- |
| longdrawertitle_prod | Screenshot 2023-10-04 at 6 56 05 AM |

## How to test 🧪
- Create a firewall with a long label (max is 32 characters) at https://cloud.linode.com/firewalls/create with word breaks that get as close as possible to the close button. (e.g. `test-this-re-l-y-l-o-n-gf-ir-e-w`)
- Check out this PR.
- `yarn dev`
- Go to https://cloud.linode.com/firewalls and visit the details page for your new firewall with a really long name.
- Select the `Linodes` tab and click the button to `Add Linodes to Firewall`.
- Observe the AddDeviceDrawer that opens and has a long title that overlaps the close icon. 
3. **How to verify changes?**
- Go to http://localhost:3000/firewalls and visit the details page for your new firewall with a really long name.
- Select the `Linodes` tab and click the button to `Add Linodes to Firewall`.
- Observe the AddDeviceDrawer that opens and has a long title that *does not* overlap the close icon.
- Check a few other drawers throughout the app to ensure there have been no unintended styling regressions from eslint reordering by comparing to prod. 
This was a styling regression.

### DIFF
--- a/packages/manager/src/components/Drawer.tsx
+++ b/packages/manager/src/components/Drawer.tsx
@@ -132,7 +132,6 @@ export const Drawer = (props: Props) => {
         <Grid>
           <IconButton
             sx={{
-              position: 'absolute',
               right: '-12px',
               top: '-12px',
             }}
--- a/packages/manager/src/components/Drawer.tsx
+++ b/packages/manager/src/components/Drawer.tsx
@@ -15,7 +15,7 @@ interface Props extends DrawerProps {
    */
   title: string;
   /**
-   * Increaces the Drawers width from 480px to 700px on desktop-sized viewports
+   * Increases the Drawers width from 480px to 700px on desktop-sized viewports
    * @default false
    */
   wide?: boolean;
--- a/packages/manager/src/components/Drawer.tsx
+++ b/packages/manager/src/components/Drawer.tsx
@@ -62,6 +62,7 @@ const useStyles = makeStyles()((theme: Theme) => ({
     },
   },
   title: {
+    // paddingRight: theme.spacing(4),
     wordBreak: 'break-word',
   },
   wide: {
@@ -132,6 +133,7 @@ export const Drawer = (props: Props) => {
         <Grid>
           <IconButton
             sx={{
+              // position: 'absolute',
               right: '-12px',
               top: '-12px',
             }}
--- a/packages/manager/src/components/Drawer.tsx
+++ b/packages/manager/src/components/Drawer.tsx
@@ -62,7 +62,7 @@ const useStyles = makeStyles()((theme: Theme) => ({
     },
   },
   title: {
-    // paddingRight: theme.spacing(4),
+    paddingRight: theme.spacing(4),
     wordBreak: 'break-word',
   },
   wide: {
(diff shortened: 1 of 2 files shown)


packages/manager/src/components/Drawer.tsx
```diff
@@ -132,7 +132,6 @@ export const Drawer = (props: Props) => {
         <Grid>
           <IconButton
             sx={{
-              position: 'absolute',
```
mjac0bs, line 135: I removed this property because with a position of `absolute`: ["the element is removed from the normal document flow, and no space is created for the element in the page layout"](https://developer.mozilla.org/en-US/docs/Web/CSS/position), which is why the drawer title was overlapping the icon button -- as far as it was concerned, it didn't exist. By removing that positioning, we inherit `position: relative` and the icon button element is part of the document flow again. 

I don't see a reason to keep `position: absolute` here, but if we wanted to keep it, another way to address this issue would be to add some `paddingRight` to the `title`.

packages/manager/src/components/Drawer.tsx
```diff
@@ -90,18 +90,18 @@ export const Drawer = (props: Props) => {
...
           [classes.default]: !wide,
           [classes.wide]: wide,
         }),
       }}
+      onClose={(event, reason) => {
+        if (onClose && reason !== 'backdropClick') {
+          onClose(event, reason);
+        }
+      }}
+      anchor="right"
```
mjac0bs, line 104: Fixing eslinting warnings.

hana-linode (approved):
Just recently noticed this issue too, thanks for putting up the fix! Confirmed that long drawer titles are no longer overlapping the close icon. 🎉

bnussman-akamai (changes requested):
Looks like some extra spacing is being added below all Drawer headers now. 

| `develop` | This PR |
|--------|-------|
|     Screenshot 2023-09-29 at 12 04 15 PM   |   Screenshot 2023-09-29 at 12 04 32 PM    |

packages/manager/src/components/Drawer.tsx
mjac0bs: Absolute positioning is useful for the icon button because it allows us to position it within the grid slightly offset to the upper right of the title, and does not take the icon button's spacing into account in the document layout. Spacing within icon button affects the bottom margin of the drawer header and was leading to additional spacing between the header and the rest of the drawer when the icon button was included in the document flow (`position: relative` used).

`position: relative`
Screenshot 2023-10-03 at 9 01 40 PM

`position: absolute`
image

mjac0bs:
> Looks like some extra spacing is being added below all Drawer headers now.

@bnussman-akamai Should be addressed now.

I ended up reverting the change to remove absolute padding and added a right margin to the title to account for the icon button within the drawer header.

| Prod | This Branch |
| --- | --- |
| editvolume_prod | editvolume_thisbranch |

bnussman-akamai (approved):
Looks good throughout the app! 🎉

PR #9743, opened 2023-10-02
fix: [M3-7226] - Make Reboot Linodes dismissible banner on VPC Details page unique

coliu-akamai:
previous PR here: https://github.com/linode/manager/pull/9742
Switched to point to develop

## Description 📝
Previously, the dismissible banner on the VPC details page was global, so dismissing it once dismissed this for every VPC. Added a unique identifier to the preference key so that the banner is unique for each VPC

image

## How to test 🧪
- Using the dev environment:
  - Create at least two VPCs
  - Assign at least one linode to each of the VPCs
  - Confirm that dismissing the banner that pops up on one of the VPC details page (image above) does not dismiss the banner on the other VPC's detail page page
- Using MSW:
  - Close the dismissible banner on one of the mock VPC's detail pages
  - Confirm the banner remains closed on that detail page
  - Navigate to another VPC's detail page and confirm the banner still appears

### DIFF
--- a/packages/manager/.changeset/pr-9743-upcoming-features-1696347149626.md
+++ b/packages/manager/.changeset/pr-9743-upcoming-features-1696347149626.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Upcoming Features
+---
+
+Make Reboot Linodes dismissible banner on VPC Details page unique for each VPC ([#9743](https://github.com/linode/manager/pull/9743))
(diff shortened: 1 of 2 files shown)


hana-linode (approved):
Confirmed that dismissing the banner only dismisses it for the VPC instead of globally ✅

dwiley-akamai (approved):
Confirmed unique as opposed to global dismissals ✅

PR #9745, opened 2023-10-02
chore: [M3-7224] - Disable Sentry Performance Tracing

bnussman-akamai:
## Description 📝
- We're disabling Sentry performance tracing because we are running into plan limits and this Sentry feature isn't crucial to our workflow.

(We still use Sentry for logging errors)

## How to test 🧪

> [!NOTE]
> These steps just confirm error reporting is still working as expected

- Check out this PR
- Make sure you have `REACT_APP_SENTRY_URL` defined in your env
- Throw an error in a component
- Go to that component and make Cloud Manager crash 
- Verify you see the error in sentry

Screenshot 2023-10-02 at 2 39 13 PM

Screenshot 2023-10-02 at 2 38 19 PM

### DIFF
--- a/packages/manager/src/initSentry.ts
+++ b/packages/manager/src/initSentry.ts
@@ -78,7 +78,7 @@ export const initSentry = () => {
       /**
        * Uncomment the 3 lines below to enable Sentry's "Performance" feature.
        * We're disabling it October 2nd 2023 because we are running into plan limits
-       * and this Sentry feature isn't crutial to our workflow.
+       * and this Sentry feature isn't crucial to our workflow.
        */
       // enableTracing: true,
       // integrations: [new BrowserTracing()],
--- a/packages/manager/src/initSentry.ts
+++ b/packages/manager/src/initSentry.ts
@@ -1,9 +1,4 @@
-import {
-  BrowserOptions,
-  BrowserTracing,
-  Event as SentryEvent,
-  init,
-} from '@sentry/react';
+import { BrowserOptions, Event as SentryEvent, init } from '@sentry/react';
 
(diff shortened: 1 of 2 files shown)


packages/manager/src/initSentry.ts
```diff
@@ -80,9 +74,15 @@ export const initSentry = () => {
         // This is apparently a benign error: https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded
         'ResizeObserver loop limit exceeded',
       ],
-      integrations: [new BrowserTracing()],
       release: packageJson.version,
-      tracesSampleRate: environment === 'production' ? 0.025 : 1,
+      /**
+       * Uncomment the 3 lines below to enable Sentry's "Performance" feature.
+       * We're disabling it October 2nd 2023 because we are running into plan limits
+       * and this Sentry feature isn't crutial to our workflow.
```
mjac0bs, line 81: Thanks for the clear code comment here. I also think it's worth adding this info to the PR description for future quick reference of anyone looking back at this PR. 

Also: can we track this in Jira with the creation of a M3 ticket?

mjac0bs, line 81:
```suggestion
       * and this Sentry feature isn't crucial to our workflow.
```

mjac0bs (approved):
✅  Triggered error and verified in Sentry:

image

I left a couple of non-blocking comments.

packages/manager/src/initSentry.ts
bnussman-akamai: Created M3-7224

jdamore-linode (approved):
@bnussman-akamai Thanks for taking care of this, and for the detailed documentation!

PR #9757, opened 2023-10-04
refactor: Refactor Cypress test setup, cache API account request

jdamore-linode:
## Description 📝
This is a first pass at fixing some tests that have been flaky recently by restoring the account request mock using real API data. One Jenkins test account is consistently taking 30+ seconds to make requests to the `/account` endpoint which is causing random failures and making it hard to pinpoint flake caused by more specific issues.

I plan to follow up with a more comprehensive PR containing more targeted fixes for the rescue, rebuild, and resize test flake we've been seeing (I originally intended for that to be _this_ PR), but want to prioritize this fix since it'll go a long way towards getting our Jenkins queue under control.

## Major Changes 🔄
- Add missing clean up to some domains tests to resolve resource accumulation issues
- Restore account request mocking by using real account response data rather than factory data
  - This should improve performance without causing subtle and hard-to-troubleshoot issues like what cropped up during the DBaaS EOS changes
- Refactor code to set up Cypress tests (adding commands, setting up request tracking, etc.), moving much of that code into a new `support/setup` dir
- Merge OAuth tests into a single spec

## How to test 🧪
This is targeting a CI-specific issue, so we will have to rely on the Jenkins run to see if this makes any impact. Specifically, one runner in particular is consistently taking 45+ minutes even when all of the tests pass; we can validate that this PR resolves that issue by confirming that all 4 runners run in a reasonable length of time (~20-25 mins)


packages/manager/cypress/support/plugins/disable-chrome-new-headless.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Attempts to resolve hanging tests by forcing Chrome's old headless implementation.
+ *
+ * See also: https://github.com/cypress-io/cypress/issues/27264
+ */
+export const disableChromeNewHeadless: CypressPlugin = (on, _config) => {
+  on('before:browser:launch', (browser, launchOptions) => {
+    if (browser.name === 'chrome' && browser.isHeadless) {
+      launchOptions.args.push('--headless=old');
```
bnussman-akamai, line 11: Maybe we should try `--disable-dev-shm-usage` too? Just guessing at this point 😵‍💫

jdamore-linode: Giving that a try too! Although the most recent run is hanging because these changes aren't actually getting applied 🤦‍♂️. It looks like we can only have one type of hook per config, so since another plugin also uses `before:browser:launch`, it gets overridden. Hoping to have this sorted out soon!

bnussman-akamai: Ohhh, I see! ++

jdamore-linode:
Since the most recent commit, we've had two runs against this branch that have finished without hanging (runs #5 and #9). 5 had a failure from a test that's been flaky lately which is OK.

Runs 6, 7, and 8 can be disregarded -- I kicked those off to test an unrelated pipeline change.

Going to continue running this to build more confidence that the hanging issue is actually resolved, but I like what I'm seeing so far.

bnussman-akamai (commented):
`yarn cy:debug` worked for me but I can't run the test headless. Is this just an issue on my end?


https://github.com/linode/manager/assets/115251059/2b2d43e2-499b-4418-80d5-508cebfbdf6b

jdamore-linode:
> `yarn cy:debug` worked for me but I can't run the test headless. Is this just an issue on my end?
> 
>  Screen.Recording.2023-10-06.at.9.47.47.AM.mov

Looking into this now! I've never seen that before -- are you connected to the VPN by any chance? I just noticed that navigating to baker is suddenly giving me `ERR_CONNECTION_CLOSED` and I'm wondering if there may be some larger issue with the VPN that might be causing this too.

bnussman-akamai:
`yarn cy:run` against develop works fine 😟 Let me debug some more

jdamore-linode:
> `yarn cy:run` against develop works fine 😟 Let me debug some more

Thanks @bnussman-akamai! I'll keep trying to reproduce as well -- I'd definitely feel better fixing (or at least understanding) this before going forward.

bnussman-akamai:
Turns out `launchOptions.args.push('--headless=old');` is causing my chrome to crash. `DEBUG=cypress:* yarn cy:run` was super helpful. Not sure why my chrome is crashing with that launch flag. I updated my chrome and now it works 🎉

bnussman-akamai (approved):

dwiley-akamai (commented):
Baker is timing out for me as well and I got this error when I tried `yarn cy:run` locally:

Screenshot 2023-10-06 at 10 51 18 AM

Is this easily resolved? My `.env` file and `MANAGER_OAUTH` are in good shape from what I can tell (no issues with Cloud locally, etc.)

bnussman-akamai:
Yeah, sounds like it! @jdamore-linode

dwiley-akamai (commented):
> @dwiley-akamai I would wager that your `MANAGER_OAUTH` PAT is expired/revoked/invalid/etc. I'm pretty sure that `MANAGER_OAUTH` is only consumed by Cypress, so I don't think you would see any issues stemming from having a bad `MANAGER_OAUTH` value during regular development.

Yup, that was it (I actually forgot to change it after you updated it a while back for me), thanks! I was able to run `yarn cy:run` locally and got failures for the following:

```
account/betas.spec.ts 
account/security-questions.spec.ts 
account/service-transfer.spec.ts
general/smoke-deep-link.spec.ts 
kubernetes/smoke-lke-create.spec.ts
linodes/backup-linode.spec.ts 
linodes/migrate-linode.spec.ts 
volumes/create-volume.spec.ts
```

It also took a total of 1:21:17 to run. Are those tests flaky and the failures expected, and is that run time normal?

Still no luck on the Baker front but if others were able to get in and confirmed the runners look good, and the above results are normal, I can go ahead and approve.

jdamore-linode:
> > @dwiley-akamai I would wager that your `MANAGER_OAUTH` PAT is expired/revoked/invalid/etc. I'm pretty sure that `MANAGER_OAUTH` is only consumed by Cypress, so I don't think you would see any issues stemming from having a bad `MANAGER_OAUTH` value during regular development.
> 
> Yup, that was it (I actually forgot to change it after you updated it a while back for me), thanks! I was able to run `yarn cy:run` locally and got failures for the following:
> 
> ```
> account/betas.spec.ts 
> account/security-questions.spec.ts 
> account/service-transfer.spec.ts
> general/smoke-deep-link.spec.ts 
> kubernetes/smoke-lke-create.spec.ts
> linodes/backup-linode.spec.ts 
> linodes/migrate-linode.spec.ts 
> volumes/create-volume.spec.ts
> ```
> 
> It also took a total of 1:21:17 to run. Are those tests flaky and the failures expected, and is that run time normal?
> 
> Still no luck on the Baker front but if others were able to get in and confirmed the runners look good, and the above results are normal, I can go ahead and approve.

Wow! An hour and a half isn't _too_ farfetched with that many specs failing, but I'd definitely like to get to the bottom of what's causing all those failures. I suppose it could be a combination of the following:

- Running the tests with `yarn dev` instead of `yarn build && yarn start:manager:ci`
  - For some reason Vite's dev server does not play nicely with Cypress and causes the tests to run a lot slower than normal
  - This is also probably responsible for the `betas.spec.ts` failure since the dev tools button overlaps the betas nav item and causes Cypress to fail when trying to navigate to the betas landing page
- Banners that need to be dismissed from within Cypress via `yarn cy:debug` obscuring some navigation elements
- Lots of existing resources on the test account that might cause Cypress to fail when it expects new resources to appear in a landing page
- Managed being enabled? (I know that `backup-linode.spec.ts` fails when Managed is enabled)
- Other account-specific issues, like tags that might impact the behavior of certain tests? (No specific ideas come to mind here)

dwiley-akamai:
@jdamore-linode Used `yarn build && yarn start:manager:ci` instead and that shaved a lot of time off, this time it completed in 45:39. Had these failures:

```
account/security-questions.spec.ts 
account/service-transfer.spec.ts
linodes/backup-linode.spec.ts
objectStorage/object-storage.e2e.spec.ts
```

Managed is enabled on the account so that accounts for `backup-linode.spec.ts` at least.

I can see Baker now so I'll take a look when the current run finishes 👍

jdamore-linode:
We've got 10 runs in a row now that have finished without hanging -- at this point I'm pretty satisfied that `--headless=old` resolves or at least seriously improves the situation with our stuck builds!

dwiley-akamai (approved):

PR #9762, opened 2023-10-05
fix [UIE-6942]: dbaas menu

corya-akamai:
## Description 📝
DBaaS menu

## Major Changes 🔄
- Correctly display menu

## How to test 🧪
1. Read/write access


packages/manager/.changeset/pr-9762-fixed-1696961275313.md
```diff
@@ -0,0 +1,5 @@
+---
+"@linode/manager": Fixed
+---
+
+Display databases menu ([#9762](https://github.com/linode/manager/pull/9762))
```
dwiley-akamai, line 5: Would something like this be more accurate?

```suggestion
Display Database menu items in side nav and Create dropdown for permissioned users ([#9762](https://github.com/linode/manager/pull/9762))
```

dwiley-akamai (commented):
Will test this thoroughly once testing steps are outlined in the PR description 👍

coliu-akamai (approved):
Confirmed unit tests for: ✅ 
- MainContent.test.tsx
- PrimaryNav.test.tsx
- AddNewMenu.test.tsx

Confirmed following flows: ✅ 
- Without any restrictions, all database pages are visible
- Restricted user without /account access but with the Database global permission, the Database pages are visible
- Restricted user without /account access but with Read or Read-Write access on DBs, the Database pages are visible

Approving based on the PR's description/how to test flows 🎉 


----------
When testing, I noticed something that might be out of scope for this ticket -- seems like we're missing the following error notices for the Database create page if a user doesn't have permission to create them:
image
I tested an account that did not have the necessary permissions, but I didn't see this notice and the form wasn't disabled. Only when I filled out the form and then clicked create database would I get an error:
image

Similar behavior for only having read access for databases -- the flows to edit the database (access controls, root password, maintenance, etc) were not disabled and were missing the 'you can't edit this database'-esque notices.

PR #9660, opened 2023-09-11
feat: [M3-7063] - Provide error handling when DC-specific pricing can…

cpathipa:
…not be determined

## Description 📝
**Brief description explaining the purpose of the changes**

## Major Changes 🔄
**List highlighting major changes**
- Change #1
- Change #2

## Preview 📷
**Include a screenshot or screen recording of the change**

> **Note**: Use `<video src="" />` tag when including recordings in table

| Before  | After   |
| ------- | ------- |
| Content | Content |

## How to test 🧪
1. **How to setup test environment?**
2. **How to reproduce the issue (if applicable)?**
3. **How to verify changes?**
4. **How to run Unit or E2E tests?**

## Commit message and pull request title format standards

> **Note**: Remove this section before opening the pull request

**Make sure your PR title and commit message on squash and merge are as shown below**

`<commit type>: [JIRA-ticket-number] - <description>`

**Commit Types:**
- `feat`: New feature for the user (not a part of the code, or ci, ...).
- `fix`: Bugfix for the user (not a fix to build something, ...).
- `refactor`: Restructuring existing code without changing its external behavior or visual UI. Typically to improve readability, maintainability, and performance.
- `style`: General styling changes. Does not change any functionality.
- `build`: Changes to the application build.
- `chore`: Other changes that do not impact production code.
- `ci`: Changes to the CI pipeline.
- `docs`: Changes to the documentation docs.
- `perf`: Performance changes.
- `test`: New tests or changes to existing tests. Does not change the production code.
- `change`: Only use for something that doesn’t fit in any other category.

**Example:** `feat: [M3-1234] - Allow user to view their login history`

PR #9765, opened 2023-10-05
test: [M3-7071] - Add Cypress integration tests for NodeBalancer Create DC-specific pricing

cliu-akamai:
## Description 📝
Add new cypress tests for NodeBalancer creation tiered pricing.

## Major Changes 🔄
- Add e2e tests to verify the DC pricing information in NodeBalancer creation flow.

## How to test 🧪
```
yarn cy:run -s "cypress/e2e/core/nodebalancers/smoke-create-nodebal.spec.ts"
```

### DIFF
--- a/packages/manager/cypress/e2e/core/nodebalancers/smoke-create-nodebal.spec.ts
+++ b/packages/manager/cypress/e2e/core/nodebalancers/smoke-create-nodebal.spec.ts
@@ -11,6 +11,12 @@ import {
 import { apiMatcher } from 'support/util/intercepts';
 import { randomLabel } from 'support/util/random';
 import { chooseRegion, getRegionById } from 'support/util/regions';
+import { dcPricingRegionNotice } from 'support/constants/dc-specific-pricing';
+import {
+  mockAppendFeatureFlags,
+  mockGetFeatureFlagClientstream,
+} from 'support/intercepts/feature-flags';
+import { makeFeatureFlagData } from 'support/util/feature-flags';
 import { ui } from 'support/ui';
 import { cleanUp } from 'support/util/cleanup';
 import { authenticate } from 'support/api/authentication';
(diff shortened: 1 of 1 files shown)


abailly-akamai (commented):
This test is currently not passing for me

Screenshot 2023-10-09 at 11 12 14

packages/manager/cypress/e2e/core/nodebalancers/smoke-create-nodebal.spec.ts
```diff
@@ -19,13 +25,48 @@ const deployNodeBalancer = () => {
...
+    // Confirms that the price will not display when the region is not selected
+    cy.get('[data-qa-summary="true"]').within(() => {
+      cy.findByText('/month').should('not.exist');
+    });
+
+    // Confirms that the price will show up when the region is selected
+    containsClick(selectRegionString).type(`${regionName}{enter}`);
+    cy.get('[data-qa-summary="true"]').within(() => {
+      cy.findByText(`$10.00/month`).should('be.visible');
+    });
```
mjac0bs, line 48: This test is also failing for me, and we seem to get stuck here. From watching the recording, it seems like something is happening with the select field that is preventing Cypress from scrolling down to the summary section and finding the price. I don't understand why this would be the case in the DC pricing flow, when `containsClick(selectRegionString).type(`${regionName}{enter}`);` is used in the non-feature-flagged flow and does not have the same issue. 😅 Maybe @jdamore-linode has some pointers about what's missing here. 

As Alban noted in the screenshot, the error is coming from a persisting `span` from the Select that prevents us from getting to the element we need:

```
     CypressError: Timed out retrying after 80050ms: `cy.click()` failed because this element:

`<span id="aria-selection-event">option ...</span>`

is being covered by another element:

`<div class="MuiBox-root css-k008qs">...</div>`
```

cliu-akamai: Change to use ```cy.get(`[value="${regionName}"]`).click().type(`${newRegion.label}{enter}`);``` instead

mjac0bs: I don't fully remember what this code looked like before the rebase, but seems like the test just wasn't using the correct region name, in which case: thanks for fixing!

mjac0bs (approved):
Confirmed this passed locally and covers DC-pricing changes for NodeBalancers. Thanks for the updates, @cliu-akamai! 🚢 

image

hana-linode (commented):
Tests are failing for me 🤔 

image

abailly-akamai (commented):
Tests are passing for me ✅ 

It would be nice to test the submission of multiple nodes and multiple configs since this form is complex. It will help us when refactoring it from a class component.

mjac0bs:
> It would be nice to test the submission of multiple nodes and multiple configs since this form is complex. It will help us when refactoring it from a class component.

@abailly-akamai Fair, but this sounds like a good candidate for a follow up ticket to me, because it extends beyond DC-specific pricing. I think we can merge this as is, and get something in the backlog for further NodeBalancer Create test coverage.

mjac0bs:
@cliu-akamai - Thanks for your work on this! I'm going to go ahead and merge since this test passes in CI and locally for at least two of us. (Not quite sure what @hana-linode was experiencing, but can't replicate it.) I'll then merge `develop` into #9660 and make the adjustment we need in that PR to avoid a test failure.

PR #9770, opened 2023-10-06
change: [SD-449] - update support error links

dgreen9:
* added ticket page link to errors containing the string "contact support"

## Description 📝

This PR updates `SupportError.tsx` to split out on "contact support" found with error messages returned via API, allowing links to the support ticket page of Cloud to be inserted. This is needed after language updates to some error messages were modified, using this wording rather than "open a support ticket".

### DIFF
--- a/packages/manager/src/components/SupportError.tsx
+++ b/packages/manager/src/components/SupportError.tsx
@@ -12,7 +12,9 @@ interface Props {
 export const SupportError = (props: Props) => {
   const theme = useTheme();
   const { errors } = props;
-  const errorMsg = errors[0].reason.split(/(open a support ticket)/i);
+  const errorMsg = errors[0].reason.split(
+    /(open a support ticket|contact support)/i
+  );
 
   return (
     <Typography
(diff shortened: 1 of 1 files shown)


packages/manager/src/components/SupportError.tsx
```diff
@@ -35,6 +38,17 @@ export const SupportError = (props: Props) => {
...
+          return (
+            <SupportLink
+              text={
+                substring.match(/Contact.*/)
+                  ? 'Contact support'
+                  : 'contact support'
+              }
+              key={`${substring}-${idx}`}
+            />
+          );
```
mjac0bs, line 51:
```suggestion
```

Rather than adding another conditional for the new substring, I think we can improve the existing SupportLink and do:

```
  if (openTicket) {
    return (
      <SupportLink
        text={
              substring.match(/^[A-Z]/)
              ? capitalize(openTicket[0])
              : openTicket[0]
        }
        key={`${substring}-${idx}`}
      />
    );
  } else {
    return substring;
  }
```

Either 'contact Support' or 'open a Support ticket' will be returned in `openTicket` from `.match()`, and then we can use regex to check if the first character is capital, use `capitalize` (`import { capitalize } from 'src/utilities/capitalize';`) or otherwise use the substring as is.

mjac0bs (approved):
Thanks for the updates. 🚀 

The test cases I tried all linked "contact Support" copy to the support ticket modal, and I tested that there were no regressions in linking "open a support ticket" copy.

I ran e2es locally and received no related failures.

bnussman-akamai (approved):
Test cases worked as expected

PR #9776, opened 2023-10-10
fix: [M3-7256] - Added spacing for Hively external link

jaalah-akamai:
## Description 📝
Feedback Smiley Ratings should be easily clickable, however it overlaps with the "How did I do?" link

## Changes  🔄
- `margin` is overwritten due to being in a `Stack`, so switching to `padding` which is the more better approach.
- Added Storybook

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screenshot 2023-10-10 at 12 31 40 PM | Screenshot 2023-10-10 at 12 36 54 PM |

Screenshot 2023-10-10 at 12 37 26 PM

### Prerequisites
- Run storybook `yarn storybook`

### Reproduction steps
- Go to http://localhost:6006/?path=/docs/components-hively--documentation

### Verification steps 
- Observe that there's no more overlapping happening in component

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/Support/Hively.stories.tsx
+++ b/packages/manager/src/features/Support/Hively.stories.tsx
@@ -1,0 +1,24 @@
+import React from 'react';
+
+import { Hively } from './Hively';
+
+import type { Meta, StoryObj } from '@storybook/react';
+
+const meta: Meta<typeof Hively> = {
+  args: {
+    linodeUsername: 'linodeuser',
+    replyId: '123',
+    ticketId: '123',
+  },
+  component: Hively,
+  title: 'Components/Hively',
+};
+
+export default meta;
+
+type Story = StoryObj<typeof Hively>;
+
+export const Default: Story = {
+  args: {},
+  render: (args) => <Hively {...args} />,
+};
--- a/packages/manager/src/features/Support/Hively.tsx
+++ b/packages/manager/src/features/Support/Hively.tsx
@@ -51,7 +51,11 @@ export const Hively = (props: Props) => {
     <>
       <Divider />
       <Stack alignItems="center" direction="row" pl={1} spacing={1.5}>
-        <Typography pr={2.5}>
+        <Typography
+          sx={{
+            paddingRight: 2.5,
+          }}
+        >
           <Link external to={href + '3'}>
             How did I do?
           </Link>
--- a/packages/manager/src/features/Support/Hively.tsx
+++ b/packages/manager/src/features/Support/Hively.tsx
@@ -51,7 +51,7 @@ export const Hively = (props: Props) => {
     <>
       <Divider />
       <Stack alignItems="center" direction="row" pl={1} spacing={1.5}>
-        <Typography pr={3}>
+        <Typography pr={2.5}>
           <Link external to={href + '3'}>
             How did I do?
           </Link>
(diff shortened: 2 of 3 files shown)


packages/manager/src/features/Support/Hively.tsx
```diff
@@ -48,7 +51,7 @@ export const Hively = (props: Props) => {
     <>
       <Divider />
       <Stack alignItems="center" direction="row" pl={1} spacing={1.5}>
-        <Typography mr={3}>
+        <Typography pr={2.5}>
```
mjac0bs, line 54: Am I recalling correctly from our cafe discussions that we preferred `marginRight` and `paddingRight` over Tailwind-like prop styles (`mr` and `pr`), so as to reduce the number of kinds of styling we have in the code base?

packages/manager/src/features/Support/Hively.stories.tsx
mjac0bs, line 1: Thanks for this!

mjac0bs (commented):
Confirmed the component looks good in Storybook. 

I also tested this in Cloud by removing the conditional on [L188 of `ExpandableTicketPanel.tsx`](https://github.com/linode/manager/blob/develop/packages/manager/src/features/Support/ExpandableTicketPanel.tsx#L188). I was seeing the bottom of the external link icon get cut off on desktop viewports. (see the first ticket in the screenshot below)

Does adding a little bit of marginTop and marginBottom (see the second ticket in the screenshot below for example, that's `5px`) resolve the issue sufficiently?)

Screenshot 2023-10-10 at 12 10 46 PM

packages/manager/src/features/Support/Hively.tsx
bnussman-akamai: I think we decided to use `sx` over styling via props in general, although I really enjoy the inline props styling 😖

jaalah-akamai:
@mjac0bs What browser / version are you using? I'm on `Chrome - Version 117.0.5938.149 (Official Build) (arm64)`
Screenshot 2023-10-10 at 4 47 06 PM

mjac0bs:
> @mjac0bs What browser / version are you using? I'm on `Chrome - Version 117.0.5938.149 (Official Build) (arm64)` !

Hmm, the same, it appears: `Version 117.0.5938.149 (Official Build) (arm64)`.

bnussman-akamai (commented):
The fix looks good but I'm starting to think that the issue is caused by the `<Link>` component. The external icon just floats beside the `<a>`. Maybe the `<Link>` component needs to be updated to that It has some margin to account for the floating icon?

Screenshot 2023-10-11 at 10 43 28 AM

jaalah-akamai:
@bnussman-akamai that would require a refactor of `Link`. I will create a separate ticket / PR for this.

mjac0bs (approved):
Approving this after talking offline. The clipping issue is inconsistent and can only be seen in some cases when a browser window is a certain width and zoom (less than 100%), which seems like an unlikely edge case. Experimenting with margin and padding also didn't seem to produce a fix. It would be nice to look into this a little more in the follow up ticket -- maybe Link issues would solve this -- but main issue resolved in this PR looks good.

PR #9677, opened 2023-09-14
feat: [M3-7114] - Allow user to choose resize migration type when resizing Linode

jaalah-akamai:
## Description 📝
This work will introduce `warm` migrations which combine the benefits of `cold` and `live` migrations to greatly reduce customer downtime when resizing Linodes.

> **Important**
The notification and event copy/messaging for `warm` migrations will be updated with API work, but you test the UI flow, payload, feature flags and everything else. I will have to update the cypress tests later with actual copy (9/27)

## Major Changes 🔄

- [x] Add feature flag `unifiedMigrations`

**Resize Flow:**
- [x] Add an optional `migration_type` POST field for the `linode/instances/.../resize` endpoint that should be `warm` or `cold`. Backend defaults to `cold`, but in the UI we are defaulting to `warm` unless the user chooses otherwise.
- [x] Add a radio button group for selecting `warm` or `cold` migrations in the `LinodeResize` flow.
    - [x] If the Linode is powered off, we disable controls (even though warm is selected by default) and proceed with a `cold` migration.
- [x] Update & add cypress E2E tests for changes

**Linode Landing:**
- [x] As a user the flag icon is ambiguous and doesn't give much meaning, therefore we want to remove the icon altogether and provide and updated status where it makes sense. For resizes, we want to change the wording from `running` to `migrating` with a green dot icon

**Other:**
- [x]  Fixed the animations for `<Notice />` and fixed a bug in our `fadeIn` keyframes. _Pesky colons!_
    - [x]  In addition, the padding for the `<Notice />` didn't seem right or match `<DismissibleBanner />` which seemed to be the intention, so this has been fixed.

## TODO:
- [ ] Waiting on APIv4 dev work with event/notification messaging to update Cypress tests
- [x] Waiting on UX copy for messaging

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screenshot 2023-09-13 at 9 30 01 PM | Screenshot 2023-09-13 at 9 29 52 PM |

|Event Notification|
| ------- | 
|Screenshot 2023-09-15 at 11 56 01 AM|

| Enabled  | Disabled   |
| ------- | ------- |
| Screenshot 2023-09-28 at 9 29 14 AM | Screenshot 2023-09-28 at 9 27 39 AM |

| Payload  |
| ------- |
|Screenshot 2023-09-13 at 3 49 32 PM|

| Linode Landing  |
| ------- |
[...]

PR #9778, opened 2023-10-10
feat: [M3-7254] - Add AGLB Routes - Rule Edit Drawer

bnussman-akamai:
## Description 📝
- Allows users to edit AGLB Route rules ✏️

## Preview 📷

Screenshot 2023-10-10 at 1 21 49 PM

### Prerequisites
- Turn on the MSW

### Verification steps 
- Go to `http://localhost:3000/loadbalancers/0/routes`
- Test the general functionality of **editing a rule** 
- Verify you can still **add a rule**
- Verify you can still **re-order rules**

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/components/Notice/Notice.styles.ts
+++ b/packages/manager/src/components/Notice/Notice.styles.ts
@@ -28,6 +28,7 @@ export const useStyles = makeStyles<
     [`& .${classes.noticeText}`]: {
       fontFamily: theme.font.normal,
     },
+    backgroundColor: theme.bg.bgPaper,
   },
   info: {
     [`&.${classes.important}`]: {
@@ -56,11 +57,7 @@ export const useStyles = makeStyles<
     [`& .${classes.error}`]: {
       borderLeftColor: theme.color.red,
     },
-    [`& .${classes.important}`]: {
-      backgroundColor: theme.bg.bgPaper,
-    },
     alignItems: 'center',
-    backgroundColor: theme.bg.bgPaper,
     borderRadius: 1,
     display: 'flex',
     fontSize: '1rem',
--- a/packages/manager/src/features/LoadBalancers/LoadBalancerDetail/RulesTable.tsx
+++ b/packages/manager/src/features/LoadBalancers/LoadBalancerDetail/RulesTable.tsx
@@ -28,6 +28,7 @@ import type { MatchField, Route } from '@linode/api-v4';
 
 interface Props {
   loadbalancerId: number;
+  onEditRule: (ruleIndex: number) => void;
   route: Route;
 }
 
@@ -42,7 +43,7 @@ const matchFieldMap: Record<MatchField, string> = {
 const screenReaderMessage =
   'Some screen readers may require you to enter focus mode to interact with Loadbalancer rule list items. In focus mode, press spacebar to begin a drag or tab to access item actions.';
 
-export const RulesTable = ({ loadbalancerId, route }: Props) => {
+export const RulesTable = ({ loadbalancerId, route, onEditRule }: Props) => {
   const { label, protocol, rules } = route;
   const theme = useTheme();
   const { enqueueSnackbar } = useSnackbar();
@@ -146,10 +147,7 @@ export const RulesTable = ({ loadbalancerId, route }: Props) => {
                     >
                       {(provided) => (
                         <li
-                          aria-label={
-                            rule.match_condition.hostname ??
-                            `Rule ${rule.match_condition.hostname}`
-                          }
+                          aria-label={`Rule ${index}`}
                           aria-roledescription={screenReaderMessage}
                           aria-selected={false}
                           ref={provided.innerRef}
@@ -158,7 +156,6 @@ export const RulesTable = ({ loadbalancerId, route }: Props) => {
                           {...provided.dragHandleProps}
                         >
                           <StyledRuleBox
-                            aria-label={`Rule ${rule.match_condition.hostname}`}
                             key={index}
                             sx={{ backgroundColor: theme.bg.bgPaper }}
                           >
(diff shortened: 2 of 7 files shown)


packages/manager/src/components/Notice/Notice.styles.ts
```diff
@@ -56,11 +57,7 @@ export const useStyles = makeStyles<
     [`& .${classes.error}`]: {
       borderLeftColor: theme.color.red,
     },
-    [`& .${classes.important}`]: {
-      backgroundColor: theme.bg.bgPaper,
-    },
     alignItems: 'center',
-    backgroundColor: theme.bg.bgPaper,
```
bnussman-akamai, line 63: This caused a UI regression. I don't think we want notices to have a background unless they are `important`

Screenshot 2023-10-10 at 2 17 36 PM

packages/manager/src/features/LoadBalancers/LoadBalancerDetail/RulesTable.tsx
```diff
@@ -146,10 +147,7 @@ export const RulesTable = ({ loadbalancerId, route }: Props) => {
                     >
                       {(provided) => (
                         <li
-                          aria-label={
-                            rule.match_condition.hostname ??
-                            `Rule ${rule.match_condition.hostname}`
-                          }
+                          aria-label={`Rule ${index}`}
```
bnussman-akamai, line 150: Is this change okay for accessibility? Rules don't have a solid way to uniquely identify them so I think `index` is best here

packages/manager/src/features/LoadBalancers/LoadBalancerDetail/RulesTable.tsx
```diff
@@ -253,7 +250,7 @@ export const RulesTable = ({ loadbalancerId, route }: Props) => {
                                 }}
                               >
                                 {rule.match_condition
-                                  .session_stickiness_cookie &&
+                                  .session_stickiness_cookie ||
```
bnussman-akamai, line 253: Fixes incorrect logic

packages/manager/src/features/LoadBalancers/LoadBalancerDetail/RulesTable.tsx
mjac0bs: Since order matters and hostname isn't unique, then I think the change makes sense.

mjac0bs (commented):
Overall, adding, editing, and reordering rules looks good. The drawer works in add and edit mode. The UI updates accordingly, populates the form with data in edit mode, and displays the updates in the table after the request is submitted.

A few questions (sorry if these have already been addressed in past PRs -- would appreciate a reminder):

Related directly to this PR:
1. Is the API spec for the request payload for `PUT /v4beta/aglb/{id}/routes/{id}` out of date? I noticed some slight discrepancies. 
- API spec includes route `label` in the request; our request payload in Cloud *does not*.
- API spec *does not* include service target `label` in the request; our payload in Cloud *does*.

Screenshot 2023-10-11 at 12 19 42 PM


Not directly related to this PR:
1. Have we talked about expanding the width of this tooltip? It's a lot of text and a little hard to read with how thin it is.

Screenshot 2023-10-11 at 12 25 33 PM

3. Should we add some logic to *not* display the tooltip on rules where there are no service targets? Currently, we display a blank tooltip popover, which is a little weird.

Screenshot 2023-10-11 at 12 25 16 PM

bnussman-akamai:
@mjac0bs 

1. We do have `label` in the PUT request type but we don't need it in the actual requets. All Linode API PUTs assume all fields are optional. I will fix the service target type though.
2. I wanted to expand the tooltip but at the time of its creation it wasn't possible. If it's possible now, I'll do it
3. I'll see if I can add something for the empty state or just hide the tooltip

mjac0bs (approved):
@bnussman-akamai 
Approving, pending the typo I mentioned is addressed and follow up tickets are created.

> 
> 1. We do have `label` in the PUT request type but we don't need it in the actual requets. All Linode API PUTs assume all fields are optional. 

Oops, yeah, you're right! 

> You are correct about the fact service target should not have `label` as a type but I hesitate to fix it as part of this PR because it will add a significant amount of code.

That's fair and a good call about not wanting to expand the scope of this PR. Can we make a follow up ticket for this?

> 2. I wanted to expand the tooltip but at the time of its creation it wasn't possible. @abailly-akamai said I should be able to widen it after [feat: [M3-7114] - Allow user to choose resize migration type when resizing Linode #9677](https://github.com/linode/manager/pull/9677) but I don't think this is true because [feat: [M3-7114] - Allow user to choose resize migration type when resizing Linode #9677](https://github.com/linode/manager/pull/9677) didn't touch the TextField component

Good context -- thank you for looking into and linking. I felt like this had been discussed before, but didn't remember exactly where. TBH, I don't understand how this styled component is working, but I see it in Storybook... [a template literal tacked on after the `styled` function](https://github.com/linode/manager/pull/9677/files#diff-a52698471b09154e66903f4b71f1dc9a8da0a5d376e20890c430007de76114e4R181-R185)? Didn't know you could do that.

It looks like [`TextfField` has a `tooltipClasses` prop](https://design.linode.com/?path=/docs/components-textfield--documentation) and we've [used that in the past to pass in a minWidth](https://github.com/linode/manager/blob/develop/packages/manager/src/features/Databases/DatabaseCreate/DatabaseCreate.tsx#L128-L134)... but from what I can tell, it seems like this is not actually working on the label field Database Create form. 

Anyway, this is minor and not worth holding up the PR over or expanding the scope too much. (Unless @jaalah-akamai can help do this cleanly.) A follow up ticket would be nice to address at some point in the future since it sounds like we're in consensus that wider is better.

> 3. I'll see if I can add something for the empty state or just hide the tooltip

Empty state of "None" looks good now; thanks for the add.

cpathipa (approved):
Overall LGTM! 
- [x] Verified the functionality of editing a rule
- [x] Verified add  rule
- [x] Verified re-order rules

PR #9784, opened 2023-10-11
test: Resolve warnings and errors in Jest tests

jdamore-linode:
## Description 📝
This PR aims to clean up our Jest tests a bit by resolving some of the errors and warnings that get logged to the console when running Jest. Specifically, this aims to fix issues within the tests themselves -- warnings and errors that need to be solved by modifying the source files themselves are out of scope here.

I plan to circle back tomorrow with a self review and comparison showing how the changes impact the Jest output.

## How to test 🧪
We can rely on GitHub Actions for this, but you can also run the tests locally:

```bash
yarn test
```

### DIFF
--- a/packages/manager/src/features/Kubernetes/ClusterList/ClusterActionMenu.test.tsx
+++ b/packages/manager/src/features/Kubernetes/ClusterList/ClusterActionMenu.test.tsx
@@ -9,6 +9,16 @@ import { ClusterActionMenu } from './ClusterActionMenu';
 
 const mockGetKubeConfig = jest.spyOn<any, any>(kube, 'getKubeConfig');
 
+// Mock snackbar to prevent "Can't perform a React state update on an unmounted component" warning.
+jest.mock('notistack', () => ({
+  ...jest.requireActual('notistack'),
+  useSnackbar: () => {
+    return {
+      enqueueSnackbar: jest.fn(),
+    };
+  },
+}));
+
 const props = {
   closeSnackbar: jest.fn(),
   clusterId: 123456,
--- a/packages/manager/src/features/Longview/LongviewLanding/LongviewPlans.test.tsx
+++ b/packages/manager/src/features/Longview/LongviewLanding/LongviewPlans.test.tsx
@@ -77,7 +77,7 @@ describe('LongviewPlans', () => {
   it('renders all columns for all plan types', async () => {
     renderWithTheme(<LongviewPlans {...props} />);
 
-    testRow(
+    await testRow(
       LONGVIEW_FREE_ID,
       'Longview Free',
       '10',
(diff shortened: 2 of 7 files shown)


bnussman-akamai (approved):
Nice! Lots of warnings/errors are gone compared to `develop`

packages/manager/src/features/Kubernetes/ClusterList/ClusterActionMenu.test.tsx
jdamore-linode, line 1: A toast notification appears when an attempt by the user to download their kubeconfig fails for any reason. This failure always occurs from within this Jest test since we're not mocking the download request, and we can't make real Linode API requests via Jest.

Since the app attempts to show a toast notification, Jest logs an error because Notistack is not actually set up and available from within the context of the Jest test. This change resolves the Jest error by mocking Notistack's `useSnackbar` export. However, it may be a better solution to mock the HTTP request to prevent the toast notification in the first place.

| Before | After |
|--------|-------|
| ClusterActionMenu BEFORE | ClusterActionMenu AFTER |

packages/manager/src/features/Longview/LongviewLanding/LongviewPlans.test.tsx
jdamore-linode, line 1: Even though `testRow` is async, it indirectly calls `act()` which cannot be executed concurrently. This change fixes the problem by `await`ing each call to `testRow()` to ensure that `act()` calls do not overlap.

| Before | After |
|--------|-------|
|     LongviewPlans BEFORE   |   LongviewPlans AFTER    |

packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.test.tsx
jdamore-linode, line 1: Similar to `BucketLanding.test.tsx`, we're testing an error condition so the Jest error is expected, and we can safely suppress it to reduce log noise. However, there is also a Formik warning being logged; I think resolving that will require a change in the form itself.

| Before | After |
|--------|-------|
|   CreateBucketDrawer BEFORE     |    CreateBucketDrawer AFTER   |

packages/manager/src/features/components/PlansPanel/PlanSelection.test.tsx
jdamore-linode, line 1: We're getting a DOM validation error here because we're not wrapping the plan selection table rows in a table. Using the `wrapWithTableBody` helper resolves the errors. This is a rare case where the DOM validation error is caused by the test; other DOM validation errors that get logged by React are actual issues with our component markup.

| Before | After |
|--------|-------|
|   PlanSelection BEFORE     |   PlanSelection AFTER    |

tyler-akamai:
Confirmed the tests now pass, great job!

Screenshot 2023-10-16 at 5 33 06 PM

also thank you for the self review!

tyler-akamai (approved):

PR #9813, opened 2023-10-18
feat: [M3-6973] - Add DC-specific pricing Object Storage overages

mjac0bs:
## Description 📝
Object storage has hardcoded prices throughout the codebase. The service is a flat $5/month. Beyond the first 250GB storage and 1TB outbound transfer, it's and $0.02/GB/month. We need to update this due to pricing changes for DC-specific overage rates and clarification that outbound transfer is *not* added to DC-specific pools, only the global pool.

The only time a user sees the Enable Object Storage notice is the first time they create a bucket or access key and must opt-in to the service. 
- For this reason, we are adding region-specific overage pricing to the Create Bucket drawer.

Note: Access keys are not currently tied to regions.

## Changes  🔄
- Overage prices on the Create Bucket drawer reflect accurate prices for dynamic region pricing.
- Copy updates are made in the Enable Object Storage modal to match the mocks for enabling via both bucket and key.
- Cleaned up `dcSpecificPricing` feature flag in the Enable OBJ Modal and related tests, since initial DC-specific pricing work has been released, to avoid extra conditional logic. DC-specific pricing behavior is now the default.
- Updated tests and added a new unit test for a new component.

## Preview 📷
| Description | Prod | This Branch |
| ---- | ---- | ---- |
| Enable Object Modal | Screenshot 2023-10-18 at 11 26 56 AM Screenshot 2023-10-18 at 11 16 30 AM Screenshot 2023-10-18 at 11 17 49 AM | Screenshot 2023-10-18 at 11 16 54 AM Screenshot 2023-10-18 at 11 17 33 AM |
| Create Bucket Drawer | Screenshot 2023-10-18 at 11 14 25 AM | Screenshot 2023-10-18 at 11 02 18 AM Screenshot 2023-10-18 at 11 02 26 AM  Screenshot 2023-10-18 at 11 02 34 AM|
| Create Bucket Drawer Tooltip | N/A | Screenshot 2023-10-18 at 11 02 44 AM Screenshot 2023-10-18 at 11 02 56 AM Screenshot 2023-10-18 at 1 07 51 PM|

### Prerequisites
- Check out this PR.
- Turn on the `objDCSpecificPricing` feature flag.
- If Object Storage already enabled, go to /account/settings and cancel it.

### Verification steps 
- With the feature flags on:
   - Confirm that the copy of the modal matches the copy in the screenshots above for each scenario, which matches the mocks attached to the internal ticket.
   - [ ] Create your first access key and enable OBJ.
   - [ ] Create your first bucket and enable OBJ.
   - Confirm that overage pricing displays in the Create Bucket drawer, matches the mocks attached to the internal ticket, and updates dynamically depending on cluster region selection.
   - [ ] Create Bucket in Sao Paulo.
   - [ ] Create Bucket in Jakarta.
   - [ ] Create Bucket in any other region (will use base pricing).

Verify unit tests pass:
```
yarn test EnableObjectStorageModal OveragePricing CreateBucketDrawer
```
Verify Cypress test passes:
```
yarn cy:run -s "cypress/e2e/core/objectStorage/enable-object-storage.spec.ts"
```
   
*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/cypress/e2e/core/objectStorage/enable-object-storage.spec.ts
+++ b/packages/manager/cypress/e2e/core/objectStorage/enable-object-storage.spec.ts
@@ -32,8 +32,7 @@ import { mockGetAccessKeys } from 'support/intercepts/object-storage';
 // under different circumstances.
 const objNotes = {
   // When enabling OBJ using a region with a regular pricing structure, when OBJ DC-specific pricing is disabled.
-  regularPricing:
-    "Linode Object Storage costs a flat rate of $5/month, and includes 250 GB of storage and 1 TB of outbound data transfer. Beyond that, it's $0.02 per GB per month.",
+  regularPricing: /Linode Object Storage costs a flat rate of \$5\/month, and includes 250 GB of storage and 1 TB of outbound data transfer. Beyond that, it.*s \$0.02 per GB per month./,
 
   // When enabling OBJ using a region with special pricing during the free beta period (OBJ DC-specific pricing is disabled).
   dcSpecificBetaPricing: /Object Storage for .* is currently in beta\. During the beta period, Object Storage in these regions is free\. After the beta period, customers will be notified before charges for this service begin./,
--- a/packages/manager/cypress/e2e/core/objectStorage/enable-object-storage.spec.ts
+++ b/packages/manager/cypress/e2e/core/objectStorage/enable-object-storage.spec.ts
@@ -296,6 +296,13 @@ describe('Object Storage enrollment', () => {
     cy.findByText('Enable Object Storage').should('not.exist');
   });
 
+  /*
+   * - Confirms that Object Storage can be enabled using mock API data.
+   * - Confirms that pricing information link is present in enrollment dialog.
+   * - Confirms that cancellation explanation is present in enrollment dialog.
+   * - Confirms that DC-specific overage pricing is explained for regions in the Create Bucket drawer.
+   * - Confirms that consistent pricing information is shown for all regions in the enable modal.
+   */
   // TODO: DC Pricing - M3-7073: Remove feature flag mocks once feature flags are cleaned up.
   it('can enroll in Object Storage with OBJ DC-specific pricing', () => {
     const mockAccountSettings = accountSettingsFactory.build({
--- a/packages/manager/cypress/e2e/core/objectStorage/enable-object-storage.spec.ts
+++ b/packages/manager/cypress/e2e/core/objectStorage/enable-object-storage.spec.ts
@@ -31,24 +31,26 @@ import { mockGetAccessKeys } from 'support/intercepts/object-storage';
 // Various messages, notes, and warnings that may be shown when enabling Object Storage
 // under different circumstances.
 const objNotes = {
-  // When enabling OBJ using a region with a regular pricing structure, or when
-  // DC-specific pricing is disabled.
+  // When enabling OBJ using a region with a regular pricing structure, when OBJ DC-specific pricing is disabled.
   regularPricing:
     "Linode Object Storage costs a flat rate of $5/month, and includes 250 GB of storage and 1 TB of outbound data transfer. Beyond that, it's $0.02 per GB per month.",
 
-  // When enabling OBJ using a region with special pricing during the free beta period.
+  // When enabling OBJ using a region with special pricing during the free beta period (OBJ DC-specific pricing is disabled).
   dcSpecificBetaPricing: /Object Storage for .* is currently in beta\. During the beta period, Object Storage in these regions is free\. After the beta period, customers will be notified before charges for this service begin./,
 
-  // When enabling OBJ without having selected a region.
+  // When enabling OBJ without having selected a region, when OBJ DC-specific pricing is disabled.
   dcPricingGenericExplanation:
     'Pricing for monthly rate and overage costs will depend on the data center you select for deployment.',
 
-  // Link to further DC-specific pricing information when DC-specific pricing is enabled.
+  // When enabling OBJ, in both the Access Key flow and Create Bucket flow, when OBJ DC-specific pricing is enabled.
+  objDCPricing:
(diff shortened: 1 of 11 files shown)


packages/manager/cypress/e2e/core/objectStorage/enable-object-storage.spec.ts
```diff
@@ -149,7 +150,7 @@ describe('Object Storage enrollment', () => {
     // Confirm dialog contents shows the expected information for regions
     // with special pricing during beta period, then cancel.
     ui.dialog
-      .findByTitle('Just to confirm...')
+      .findByTitle('Enable Object Storage')
```
mjac0bs, line 153: As I commented in another spot, I did not hide this title change behind a feature flag because it's a general improvement, so these tests needed updating in multiple spots.

packages/manager/cypress/e2e/core/objectStorage/enable-object-storage.spec.ts
```diff
@@ -251,7 +252,7 @@ describe('Object Storage enrollment', () => {
 
         // Click "Enable Object Storage".
         ui.button
-          .findByTitle('Enable Object Storage')
+          .findByAttribute('data-qa-enable-obj', 'true')
```
mjac0bs, line 255: Now that the modal title and primary action button match, this selector needed to be changed to be more specific, or the test would fail.

packages/manager/cypress/e2e/core/objectStorage/enable-object-storage.spec.ts
```diff
@@ -417,14 +486,70 @@ describe('Object Storage enrollment', () => {
...
+    // Click through the "Access Keys" dialog which displays the new access key.
     ui.dialog
-      .findByTitle('Just to confirm...')
+      .findByTitle('Access Keys')
       .should('be.visible')
       .within(() => {
-        // Confirm that pricing and cancellation explanations are present, but
-        // DC-specific information is hidden.
-        confirmModalContents();
+        ui.button
```
mjac0bs: The diff looks weird here, but I think it is just because I deleted the test that checked that old modal copy displayed when the `dcSpecificPricing` feature flag was off - now that this code has been released, we won't be turning it off.

packages/manager/src/features/ObjectStorage/EnableObjectStorageModal.tsx
```diff
@@ -45,36 +41,50 @@ export const EnableObjectStorageModal = React.memo(
...
+              transfer will be added to your global network transfer pool.
+            </StyledTypography>
+          </>
+        );
+      }
+
+      if (!regionId) {
+        return (
+          <StyledTypography>{ENABLE_OBJ_ACCESS_KEYS_MESSAGE}</StyledTypography>
+        );
```
abailly-akamai, line 65: do we still need this? now that the create button is disabled without a selected region I wonder if that's a plausible case?

packages/manager/src/utilities/pricing/dynamicPricing.ts
```diff
@@ -26,17 +27,14 @@ export const priceIncreaseMap = {
...
-    transfer_overage: 0.0,
+    storage_overage: 0.028,
+    transfer_overage: 0.007,
   },
   'id-cgk': {
-    monthly: 0.0,
-    storage_overage: 0.0,
-    transfer_overage: 0.0,
+    storage_overage: 0.024,
+    transfer_overage: 0.015,
```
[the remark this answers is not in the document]
abailly-akamai: Do we have any confidence those numbers won't need frequent updates? I am assuming so, but I'd like to point out we could make them part of the feature flag data (instead of making it a boolean) if confidence was low

abailly-akamai (approved):
Great work - just left a couple questions

matches mocks ✅ 
Created your first access key and enable OBJ. ✅ 
Created your first bucket and enable OBJ. ✅ 
Overage pricing display ✅ 
Create Bucket in Jakarta. ✅ 
Create Bucket in any other region ✅ 
Unit test ✅ 
e2e ✅

bnussman-akamai (commented):
This is looking great so far! (Much better than the original proposals for obi pricing UX)

Kind of unrelated to the pricing stuff but it feels weird that `Label` isn't marked as `(required)` but `Region` is `(required)`

Screenshot 2023-10-20 at 11 13 08 AM

packages/manager/src/utilities/pricing/dynamicPricing.ts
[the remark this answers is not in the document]
mjac0bs: I would not expect this map to need frequent updates once prices are set and this is released, and the future goal is to get them from API. I did check with Lynn to confirm that these are for sure accurate, since a question was raised about Sao Paulo's transfer overage being lower than Sao Paulo's earlier in the week. Received confirmation that prices are final.

mjac0bs:
> Kind of unrelated to the pricing stuff but it feels weird that `Label` isn't marked as `(required)` but `Region` is `(required)`

@bnussman-akamai Yeah, agreed. We made Region `(required)` in the first OBJ pricing release to provide user feedback because we disabled the create button unless a region is selected. But now the mismatch is weird, so addressed this in [fe7479e](https://github.com/linode/manager/pull/9813/commits/fe7479ee28187291cb79c81547be2e68c980e4d4). I left the label on the Access Key flow as is, with the thought that of course the only form field in a form would be required; it's implied.

Screenshot 2023-10-20 at 9 17 04 AM

Worth noting in this PR that the only reason we're passing `regionId` to the Enable modal now is because we need a way to differentiate between whether we're in the access key flow or the bucket flow for a small copy difference: `To create your first [access key / bucket ]...`. I kept the `regionId` prop as the differentiator since it exists for buckets and not access keys. I could rename this prop (and revert form `(required)` UX + disabled button), but I think the earlier feedback can be beneficial to the user.

bnussman-akamai (approved):
Great work as always @mjac0bs

mjac0bs:
The CI e2e failure is once again due to the 504 issue on that 4th runner, so that subset of tests didn't run. I ran them locally using the handy command Joe shared with us, and all passed, so I am merging this now:

image

PR #9817, opened 2023-10-19
test: [M3-6428] - Migrate from Jest to Vitest

jdamore-linode:
## Description 📝
Replaces Jest with Vitest and attempts to get all of our tests passing. See also PR #8951 -- a lot of the test changes in this PR stem from that.

**Trouble tests and other things to call attention to:**
- Made a change to `confirmObjectStorage()` in `ObjectStorage/utilities.ts` to convert the function to async/await syntax, but have not tested the change in Cloud
  - I think this is only called when enabling Object Storage by creating an access key
- `RescueDialog.test.tsx`, `SearchLanding.test.tsx`, `CreateAPITokenDrawer.test.tsx` are each failing intermittently, but seem to pass when run by themselves
- We get a lot of failures when running single threaded

I've kept pretty good documentation on each of the significant changes made to the tests, and plan to follow up with a self review that describes each significant change.

## Changes  🔄
- The majority of our tests pass by replacing `jest` with `vi`
- Others required some minor refactoring, mostly involving module mocking and other calls to `vi.mock`, moving component rendering to inside of tests, and making changes involving async operations

## How to test 🧪
`yarn test`

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset


bnussman-akamai (approved):
I know this is marked as a work in progress but I'm leaving my approval now.

I think this change is the right direction for us and has an overwhelming number of advantages over our jest setup

jdamore-linode:
> I know this is marked as a work in progress but I'm leaving my approval now.
> 
> I think this change is the right direction for us and has an overwhelming number of advantages over our jest setup

Thanks, @bnussman-akamai! Another advantage I discovered earlier today is memory usage is _dramatically_ improved with Vitest.

Try running `yarn test --logHeapUsage` on `develop` and you'll see the memory usage slowly creep up for each test until it's well over a gig. Running the same command on the Vitest branch shows each test hovering around ~50-150MB (minus a few exceptions) with no apparent upward creep.

package.json
```diff
@@ -32,7 +32,7 @@
     "start:manager": "yarn workspace linode-manager start",
     "start:manager:ci": "yarn workspace linode-manager start:ci",
     "clean": "rm -rf node_modules && rm -rf packages/@linode/api-v4/node_modules && rm -rf packages/manager/node_modules && rm -rf packages/@linode/validation/node_modules",
-    "test": "yarn workspace linode-manager test --maxWorkers=4",
+    "test": "yarn workspace linode-manager test",
```
dwiley-akamai, line 35: Would we no longer have multiple workers running with the switch to Vitest?

dwiley-akamai (commented):
Going to circle back to review this more in-depth in the coming days, but this was my initial look at it. I ran `yarn test` and didn't get any failures 🎉  and it seems to run the test suite quite quickly

Screenshot 2023-10-27 at 12 09 46 PM

package.json
jdamore-linode: @dwiley-akamai Nope! This just removes the upper limit of having 4 workers. If I remember correctly, Vitest (and maybe Jest too) will default to the number of available cores minus 1, but we added this argument to work around an issue we were having in CI, I believe.

I originally removed it to see if CI would pass without it, but I may restore it to see if having it improves performance at all. And if it doesn't, I'll remove it again 😅

packages/manager/src/features/Account/AutoBackups.test.tsx
jdamore-linode, line 1: There is no Vitest equivalent to `jest.setTimeout`, so in this case we just pass the desired timeout to the test as the third parameter when calling `it(...)`.

packages/manager/package.json
```diff
@@ -191,11 +190,8 @@
     "eslint-plugin-xss": "^0.1.10",
     "factory.ts": "^0.5.1",
     "glob": "^10.3.1",
-    "jest": "~26.4.2",
-    "jest-axe": "^3.3.0",
-    "jest-junit": "^10.0.0",
-    "jest-sonar-reporter": "^2.0.0",
-    "jest_workaround": "^0.1.10",
+    "jest-axe": "^8.0.0",
```
bnussman-akamai, line 193: I think we only have one test using `jest-axe`. If we're not getting much value from it, I'd be in favor of removing. We can worry about this later to keep this PR focused.

packages/api-v4/vitest.config.ts
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    globals: true,
```
bnussman-akamai, line 5: Because api-v4 is easily changed, I'd be in favor of no globals. I find globals to be annoying. I'm okay with leaving as is to match manager though.

bnussman-akamai (approved):
🚀

coliu-akamai:
Coming over from frontend cafe 😆 -- will take a look tomorrow! 👀

coliu-akamai (approved):
Looks good, thanks Joe!! 🎉  Appreciate the explanations on this and the previous pr, was helpful for understanding the changes made

I did get the intermittent test failures you mentioned for SearchLanding.test.tsx and CreateAPITokenDrawer.test.tsx when running `yarn test` - any chance you know why they happen?

dwiley-akamai (commented):
I need to finish looking through the diff but I ran the tests a couple more times and got a handful of failures each time:

Screenshot 2023-11-06 at 4 43 45 PM

Screenshot 2023-11-06 at 4 49 38 PM

dwiley-akamai (approved):

PR #9827, opened 2023-10-23
change: [M3-7182] - Update RemovableSelectionsList default maximum height to cut off the last list item

coliu-akamai:
## Description 📝
- Add dropshadow and update maxHeight to list to indicate scrolling is necessary
- Fix dropshadow UI bug on Safari for market place apps

#### some background
- See [this comment here](https://github.com/linode/manager/pull/9687#discussion_r1335967787) for some background behind this issue

## Preview 📷

| Before  | After   |
| ------- | ------- |
|image |image |

### Prerequisites
- checkout this branch
- yarn storybook --> RemovableSelectionsList

- Marketplace fix: navigate to the marketplace apps on **safari**

### Verification steps 
- For the removable selections list
  - verify that the default example now cuts off the 11th item in the list. 
  - verify that there is a drop shadow if the list can be scrolled
  - Alternatively, if you want to test this with the SubnetAssignLinodesDrawer, make sure you're on the dev environment with the vpc tags, and you can assign 11+ linodes to a subnet

- for the Marketplace apps, verify that the drop shadow appears above the individual apps (it should now mirror chrome's behavior)

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/.changeset/pr-9827-upcoming-features-1698091273479.md
+++ b/packages/manager/.changeset/pr-9827-upcoming-features-1698091273479.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Upcoming Features
+---
+
+Update `RemovableSelectionsList` default maximum height to cut off last list item and indicate scrolling ([#9827](https://github.com/linode/manager/pull/9827))
--- a/packages/manager/src/components/RemovableSelectionsList/RemovableSelectionsList.style.ts
+++ b/packages/manager/src/components/RemovableSelectionsList/RemovableSelectionsList.style.ts
@@ -1,0 +1,90 @@
+import { styled } from '@mui/material/styles';
+
+import { Box } from 'src/components/Box';
+import { List } from 'src/components/List';
+import { ListItem } from 'src/components/ListItem';
+import { omittedProps } from 'src/utilities/omittedProps';
+
+import type { RemovableSelectionsListProps } from './RemovableSelectionsList';
+
+export const StyledNoAssignedLinodesBox = styled(Box, {
+  label: 'StyledNoAssignedLinodesBox',
+  shouldForwardProp: omittedProps(['maxWidth']),
+})(({ maxWidth, theme }) => ({
+  background: theme.name === 'light' ? theme.bg.main : theme.bg.app,
+  display: 'flex',
+  flexDirection: 'column',
+  height: '52px',
+  justifyContent: 'center',
+  maxWidth: maxWidth ? `${maxWidth}px` : '416px',
+  paddingLeft: theme.spacing(2),
+  width: '100%',
+}));
+
+export const SelectedOptionsHeader = styled('h4', {
+  label: 'SelectedOptionsHeader',
+})(({ theme }) => ({
+  color: theme.color.headline,
+  fontFamily: theme.font.bold,
+  fontSize: '14px',
+  textTransform: 'initial',
+}));
+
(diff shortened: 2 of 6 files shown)


abailly-akamai (commented):
Looks good! Did we consider adding a inner shadow such as the marketplace list one?

It would prob add visual consistency with to those scrollable containers

Screenshot 2023-10-24 at 09 22 36

coliu-akamai:
@abailly-akamai thanks Alban! I'd checked with Andrew, and as of right now we do not plan to add the drop shadow. I'll update the internal ticket to make note of that!

abailly-akamai:
@coliu-akamai what was the reasoning? It's good to remember we may all have different experiences based on our browser settings

tyler-akamai:
Confirmed the list does now get cut off:
Screenshot 2023-10-24 at 9 55 39 AM

tyler-akamai:
> @coliu-akamai what was the reasoning? It's good to remember we may all have different experiences based on our browser settings

It might be worth taking a screenshot with and without the shadow, and then showing it to Andrew.

coliu-akamai:
@abailly-akamai let me double check, but iirc, we saw behavior with the drop shadow disappearing behind the list items in the marketplace apps when scrolling that was weird and potentially even made it worth revisiting the shadow there. However, I'm currently not seeing that on my browser (chrome) for marketplace and the lists, so I'll check with Andrew again and get back to you.

@tyler-akamai @abailly-akamai I'm experimenting with the drop-shadow for the react list, below (if we go with the drop shadow, we may need to darken the color due to the background for this list being darker than the marketplace app's background -- currently using theme.color.boxShadow). 

Am also currently looking into a way to get the drop shadow to only show up if the list needs a scrollbar:

image
image

coliu-akamai:
@abailly-akamai to follow up on this, spoke with Andrew - it is Safari that has the weird behavior with box shadows (note how the market place apps are above the boxshadow when scrolling). 

image

Andrew noted that this may be how Safari treats boxshadows, and that it might be better to use a gradient instead --I'm going to look more into this issue soon.

Regarding the box shadow on the RemovableSelectionsList - got the go ahead to add it for when the list actually needs to scroll (but otherwise hide it if we do not yet need to scroll the list). Just pushed up that change!

abailly-akamai:
@coliu-akamai great thanks for the update! i wonder if the safari bug has to do with a simple z-index issue.

tyler-akamai:
- [x] Shadow appears with enough items to scroll
Screenshot 2023-10-25 at 10 02 53 AM

- [ ] Shadow does not appear when there is not enough items to scroll
Screenshot 2023-10-25 at 10 03 29 AM
Screenshot 2023-10-25 at 10 04 16 AM

When there is 10 items the shadow appears even if its not scrollable. But as soon as you go down to 9, the shadow disappears. This is a very small detail but just throwing it out there.

Also, I love the Reset List button, makes testing really easy.

coliu-akamai:
Thanks @tyler-akamai! Actually gonna make a few changes to make the drop shadow more dynamic, will lyk when those changes are pushed!

bnussman-akamai (approved):
Safari Marketplace UI is fixed, VPC changes look as described

tyler-akamai (approved):

abailly-akamai (approved):
Looks great! thanks for fixes.

confirmed max-height and dynamic list height logic ✅ 
Confirmed drop shadow fixes in both places ✅

PR #9839, opened 2023-10-24
change: [M3-7250] – Properly surface VPC interface errors

dwiley-akamai:
## Description 📝
There was an API bug for API endpoints that take a list of interfaces where, if there was an error for one of the interfaces, the `field` of the error didn't indicate which interface index it applied to. This presented difficulties for surfacing the errors in the UI. Temporary code was added in https://github.com/linode/manager/pull/9709 to surface the errors in a general way just so that they wouldn't stay submerged, but since the API fix has been merged in, that temporary code is being removed so the errors surface and map to the appropriate interfaces.

## Changes  🔄
- Remove temporary code in `LinodeConfigDialog.tsx`
- Move `publicIPv4Error` outside of `<Box />` in `VPCPanel.tsx` to avoid odd formatting of error in the Add/Edit Linode Config dialog

## Preview 📷
(for the `publicIPv4Error` error change -- this error is only really possible to get back in the Linode Config flow, not the Linode Create flow)
| Before  | After   |
| ------- | ------- |
| Screenshot 2023-10-24 at 4 22 12 PM | Screenshot 2023-10-24 at 4 21 57 PM |

### Prerequisites
- Account with proper VPC tags
- Point at `dev` environment

### Verification steps 
Try adding a Linode config or editing an existing config:
- Choose a VPC interface, uncheck the "Auto-assign a VPC IPv4 address for this Linode in the VPC" checkbox, and input a reserved IP in the "VPC IPv4" field (example: if the subnet's range is 10.0.4.0/24, you can input 10.0.4.0 in that field).
  - Observe: when you submit the form, you should see a field error: "The provided IP is reserved"
- Try adding two VPC interfaces and check the "Assign a public IPv4 address for this Linode" checkbox for both of them
  - Observe: the second one should get a "Linode has no public IP to use for 1:1 NAT" error

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/Linodes/LinodesCreate/VPCPanel.tsx
+++ b/packages/manager/src/features/Linodes/LinodesCreate/VPCPanel.tsx
@@ -304,16 +304,16 @@ export const VPCPanel = (props: VPCPanelProps) => {
                   </Box>
                 }
               />
-              {assignPublicIPv4Address && publicIPv4Error && (
-                <Typography
-                  sx={(theme) => ({
-                    color: theme.color.red,
-                  })}
-                >
-                  {publicIPv4Error}
-                </Typography>
-              )}
             </Box>
+            {assignPublicIPv4Address && publicIPv4Error && (
+              <Typography
+                sx={(theme) => ({
+                  color: theme.color.red,
+                })}
+              >
+                {publicIPv4Error}
+              </Typography>
+            )}
           </Stack>
         )}
       </Stack>
--- a/packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx
+++ b/packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx
@@ -425,20 +425,9 @@ export const LinodeConfigDialog = (props: Props) => {
         });
       };
 
-      // @TODO VPC: Remove this override and surface the field errors appropriately
-      // once API fixes interface index bug for ipv4.vpc & ipv4.nat_1_1 errors
-      const overrideFieldForIPv4 = (error: APIError[]) => {
-        error.forEach((err) => {
-          if (err.field && ['ipv4.nat_1_1', 'ipv4.vpc'].includes(err.field)) {
-            err.field = 'interfaces';
-          }
-        });
-      };
-
       formik.setSubmitting(false);
 
       overrideFieldForDevices(error);
-      overrideFieldForIPv4(error);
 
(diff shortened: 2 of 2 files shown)


mjac0bs (commented):
I observed both errors following the provided testing instructions. ✅ 

One clarification question:
> Try adding two VPC interfaces and check the "Assign a public IPv4 address for this Linode" checkbox for both of them

This means two distinct VPC interfaces, correct? And user is allowed to configure the same VPC on multiple network interfaces, and should not expect to see that error?

Screenshot 2023-10-25 at 8 39 04 AM

dwiley-akamai:
> One clarification question:
> 
> > Try adding two VPC interfaces and check the "Assign a public IPv4 address for this Linode" checkbox for both of them
> 
> This means two distinct VPC interfaces, correct? A user is allowed to configure the same VPC on multiple network interfaces, and should not expect to see that error?

I tried assigning the same VPC and subnet for each interface and observed the error 🤔

In verifying this, though, I did come across a bug earlier: trying to submit an initial combo that yields the error, and then changing the selected VPC and subnet resulted in a successful submission despite that interface slot's "Assign a public IPv4 address for this Linode" checkbox being checked. That may have been what you were actually running into?

I believe this was happening because `handleVPCLabelChange()` and `handleSubnetChange()` weren't accounting for `ipv4.nat_1_1` in the object, so it was getting excluded from the payload on the final submission. This should be fixed in the next commit

mjac0bs (approved):
> trying to submit an initial combo that yields the error, and then changing the selected VPC and subnet resulted in a successful submission despite that interface slot's "Assign a public IPv4 address for this Linode" checkbox being checked. That may have been what you were actually running into?

Hmm, must have been? I thought the first attempt that I made was with assinging the same VPC and subnet to `eth0` and `eth1`, and did not see the expected error, at which point I went on to try (and see it successfully) with two different VPCs and subnets. In any case, I retested with the latest changes and am consistently seeing the expected `Linode has no public IP to use for 1:1 NAT` error when creating/editing an interface with two VPC interfaces and checked "Assign a public IPv4 address for this Linode", regardless of whether the VPCs were the same, different, or changed and then resubmitted.

hana-linode (approved):
Confirmed both errors surface ✅ 

| Inputting a reserved IP  | Two VPC interfaces with assign IPv4  |
| ------- | ------- |
| Screenshot 2023-10-30 at 1 36 11 PM | Screenshot 2023-10-30 at 1 35 53 PM |

PR #9845, opened 2023-10-26
change: [M3-7289] – Payload refinements in "Assign Linodes" flow

dwiley-akamai:
## Description 📝
Currently, VPC interfaces added via the Assign Linodes flow are not marked as `primary` and are not 1:1 NAT'ed. This results in a sub-optimal and possibly non-functional configuration.

## Changes  🔄
- Add `primary: true` and `nat_1_1: 'any'` to the interface payload
- If the config being modified only had an implicit public interface previously (i.e., the `interfaces` array comes back from the API as empty), explicitly add a public interface in eth0

### Prerequisites
- An account with the proper VPC tags
- Pointed at the `dev` environment

### Verification steps 
With the Network tab of Dev Tools open...
- Create a new Linode with no VPC or VLAN assigned
  - Go to a VPC in the same region as the newly-created linode with a subnet and enter the "Assign Linodes" flow. Select the newly-created linode and assign it. Observe: two requests to `/interfaces`, the first for adding an explicit public interface, and the second for the VPC interface (confirm it has `primary: true` and `nat_1_1: 'any'`)

For other possible permutations (e.g., a linode with a VLAN already assigned), confirm that going through the "Assign Linodes" flow results in a VPC interface being added with `primary: true` and `nat_1_1: 'any'` in the payload. No explicit public interfaces should be in the payload outside of the scenario detailed above.

*Check all that apply*

- [X] 👀 Doing a self review
- [X] ➕ Adding a changeset


packages/manager/src/features/Linodes/LinodesCreate/LinodeCreate.tsx
```diff
@@ -943,7 +943,7 @@ export class LinodeCreate extends React.PureComponent<
   ];
 }
 
-const defaultPublicInterface: InterfacePayload = {
+export const defaultPublicInterface: InterfacePayload = {
```
dwiley-akamai, line 946: I briefly considered moving this to `src/constants.ts`, but it felt out of place there and I didn't see a better spot so I just exported it from here for now.

hana-linode:
Confirmed that two requests are made to `/interfaces`, one for adding a public interface and another for the vpc interface. 

Should we add the checkbox for the public IPv4 address similar to the flow in Linodes Create?

image

dwiley-akamai:
@hana-linode I think we just take care of that with `nat_1_1: 'any'` so that we help the user towards a functional configuration when they go through the "Assign Linodes" flow, but we can revisit its inclusion within the flow depending on user feedback

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
```diff
@@ -144,16 +145,34 @@ export const SubnetAssignLinodesDrawer = (
...
     const configId = getConfigId(linodeConfigs, selectedConfig);
 
+    const configToBeModified = linodeConfigs.find(
+      (config) => config.id === configId
+    );
+
     const interfacePayload: InterfacePayload = {
       ipam_address: null,
+      ipv4: {
+        nat_1_1: 'any',
```
abailly-akamai, line 155: Shouldn't it be: `!autoAssignIPv4 ? 'any' : undefined,` ?


Am seeing this inconsistency with the [LinodeCreate vpcInterfaceData](https://github.com/linode/manager/blob/develop/packages/manager/src/features/Linodes/LinodesCreate/LinodeCreate.tsx#L801-L814).

abailly-akamai: In the light of that am thinking it would be better to make a `getVPCInterfacePayload` util to have consistent naming convention and payload logic instead of repeating those declaration with potential variations

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
```diff
@@ -144,16 +145,34 @@ export const SubnetAssignLinodesDrawer = (
...
       purpose: 'vpc',
       subnet_id: subnet?.id,
       vpc_id: vpcId,
-      ...(!autoAssignIPv4 && { ipv4: { vpc: chosenIP } }),
     };
 
     try {
+      // If the config has only an implicit public interface, make it explicit and
+      // add it in eth0
+      if (configToBeModified?.interfaces.length === 0) {
```
abailly-akamai, line 168: Are we guaranteed to have an empty `interfaces` array or can `interfaces` here be null or undefined?

dwiley-akamai: `interfaces` [is a non-nullable property of the config profile return object](https://www.linode.com/docs/api/linode-instances/#configuration-profile-view)

packages/manager/src/features/VPCs/VPCDetail/SubnetAssignLinodesDrawer.tsx
dwiley-akamai: The difference is intentional and requested by stakeholders. A public IPv4 assignment checkbox is not currently part of this flow to abstract away some complexity.

abailly-akamai: Sounds good - I would encourage to document your code a little better to that effect because it's not something someone without the VPC context can guess.

abailly-akamai (approved):
Looks good! Approving after confirmation of a few questions. ✅

coliu-akamai (approved):
Looks good, thanks Dajahi!

- confirmed two network requests sent in the first case described
- confirmed one network request sent if linode already had vlan assigned to it

packages/manager/cypress/e2e/core/databases/update-database.spec.ts
mjac0bs, line 1: Looks like an unrelated linter change slipped in.

mjac0bs (approved):
✅  Two requests to `/interfaces` with expected payload when no VLAN or VPC on initial create:

https://github.com/linode/manager/assets/114685994/5f3e36d9-8476-4005-86c4-04fe8545b40d


✅  One request to `/interfaces` with expected payload when there's a VLAN on initial create:

https://github.com/linode/manager/assets/114685994/ec2bacec-2e5e-4b7f-ab8c-48064befc671

hana-linode (approved):
✅ Two requests to /interfaces, the first for adding an explicit public interface, and the second for the VPC interface (confirm it has primary: true and nat_1_1: 'any')

✅ One request to /interfaces when there’s already a VLAN

PR #9868, opened 2023-11-02
change: [M3-7350] Display `Network Interfaces`/`Networking` section unconditionally & conditionally display "unavailable in region" helper text

dwiley-akamai:
## Description 📝
Display the `Network Interfaces`/`Networking` section in the Linode Config dialog unconditionally instead of only showing it if the region supports VLANs.

Conditionally render "VPC is not available in this Linode's region" / "VLAN is not available in this Linode's region" if the user selects the interface's purpose as VPC or VLAN, respectively, and that product is not available in the Linode's region. 

## To-Do:
- [X] Add unit test

## Changes  🔄
- Display the `Network Interfaces`/`Networking` section in the Linode Config dialog unconditionally
- Conditionally render "VPC is not available in this Linode's region" / "VLAN is not available in this Linode's region"
- Add unit test for `InterfaceSelect`

## Preview 📷
Screenshot 2023-11-02 at 5 42 15 PM

### Prerequisites
- An account with the proper VPC tags

### Verification steps
- If using real data: go to the Add/Edit Config dialog of a Linode in a region that doesn't have VPC, select VPC as the purpose for one of the eth slots, and observe the helper text below the dropdown. (I haven't found a region that doesn't have VLANs enabled where you can create a linode)
- If using mock data: pick a region in your `regions.json` file & remove `Vlans` from its `capabilities` array --> update the `region` property in the response for `rest.get('*/linode/instances/:id')` in `serverHandlers.ts` to the `id` of the region you changed in the previous step --> do the same steps in Cloud as you would for real data above (but you can verify for both VPC and VLAN here)

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


coliu-akamai (approved):
Awesome, thanks Dajahi! 🎉 

✅ confirmed unit test passes
✅ confirmed  "VPC is not available in this Linode's region" / "VLAN is not available in this Linode's region" is conditionally rendered
✅ Confirmed Networking section is displayed

PR #9870, opened 2023-11-03
feat: [M3-7374] - Add AGLB Configuration Create Flow

bnussman-akamai:
## Description 📝
- Allows users to create AGLB configurations 
- Allows users to remove routes from configurations rather than just outright deleting the route

## Preview 📷

Screenshot 2023-11-03 at 1 07 01 PM

### Prerequisites
- Turn on the MSW
- Go to `http://localhost:3000/loadbalancers/0/configurations`

### Verification steps 

> [!important]
> Because we are testing with the MSW, some actions may have unexpected results

- Verify you can still **edit** a configuration
- Verify you can **create** a configuration

```
yarn cy:run -s "cypress/e2e/core/loadBalancers/load-balancer-configurations.spec.ts"
```

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/validation/src/loadbalancers.schema.ts
```diff
@@ -155,36 +155,40 @@ export const UpdateRouteSchema = object({
...
-    }
-  ),
-  host: string().test(
-    'host-or-ip',
-    'Either host or IP must be provided.',
-    function (value) {
-      const { ip } = this.parent;
-      return !!value || !!ip;
-    }
-  ),
```
bnussman-akamai, line 187: I'm not sure where we got this info, I don't think it is correct

packages/manager/src/features/LoadBalancers/LoadBalancerDetail/Configurations/ConfigurationForm.tsx
bnussman-akamai, line 1: I made this form useable for both editing and creating.

packages/manager/src/features/LoadBalancers/LoadBalancerDetail/Configurations/ConfigurationAccordionHeader.tsx
bnussman-akamai, line 1: Separated this into its own component

packages/validation/src/loadbalancers.schema.ts
mjac0bs: This is a bullet point under `POST /v4beta/aglb` in the API spec.

- When passing data to create endpoints for service targets, you pass either “ip” with a value and “host” remain empty or vice versa. USE THE “FORBIDDEN” VALIDATION CHECK FOR THIS. REJECT IF BOTH, CAN ONLY USE ONE OR THE OTHER.

But this is the bullet point under `POST /v4beta/aglb/{id}/routes` and `POST /v4beta/aglb/{id}/service-targets`...

- When passing data to create endpoints for service targets, “ip” and “port” are mandatory and host is optional.

...can we clarify? 😅

cpathipa:
Not sure if this is related to this PR, but I see the app is crashing when adding a route.

https://github.com/linode/manager/assets/119517080/398b1791-af2d-4221-a312-f37a65d57b90

bnussman-akamai:
@cpathipa Fixed in d26312f. The `queryKey` of the infinite query and the paginated query were conflicting

packages/manager/src/features/LoadBalancers/LoadBalancerDetail/Configurations/UnusedConfigurationNotice.tsx
```diff
@@ -0,0 +1,76 @@
...
+            variant: 'success',
+          }
+        );
+      })
+      .catch((error: APIError[]) => {
+        enqueueSnackbar(error[0].reason, {
+          variant: 'error',
+        });
+      });
+  };
```
cpathipa, line 58: We could use `await` to replace callbacks here.
```
const onAttach = async () => {
  if (!loadbalancer) {
    return;
  }

  const existingConfigs = loadbalancer?.configurations.map(
    (config) => config.id
  );

  try {
    await updateLoadbalancer({
      configuration_ids: [...existingConfigs, configurationId],
    });
    enqueueSnackbar(
      `Successfully enabled configuration ${configurationId}.`,
      {
        variant: 'success',
      }
    );
  } catch (error) {
    enqueueSnackbar(error[0].reason, {
      variant: 'error',
    });
  }
};

```

cpathipa (approved):
Overall looks good!  verified edit / create  a configuration. Left minor non blocking feedback.
Approving, pending confirmation on passing data to endpoints (IP and host).

packages/validation/src/loadbalancers.schema.ts
bnussman-akamai: Looking at the API code, it allows both to be passed. We can revist later if this changes.

packages/manager/src/features/LoadBalancers/LoadBalancerDetail/Configurations/ApplyCertificatesDrawer.tsx
```diff
@@ -38,7 +38,6 @@ export const ApplyCertificatesDrawer = (props: Props) => {
       onAdd(values.certificates);
       onClose();
     },
-    validateOnChange: false,
```
mjac0bs, line 41: Was this intentional? It seems like it may be causing the following behavior where clicking the Add Another button triggers form validation. It's a weird experience if the user has not yet filled anything in and then sees their whole form turn red. 

https://github.com/linode/manager/assets/114685994/b33d62f1-4122-4eff-8fcc-e4568c4a1c07

packages/validation/src/loadbalancers.schema.ts
```diff
@@ -155,36 +155,42 @@ export const UpdateRouteSchema = object({
 });
 
 export const UpdateConfigurationSchema = object({
-  label: string().min(1),
-  port: number(),
+  label: string().min(1, 'Label must not be empty.'),
+  port: number()
+    .min(0, 'Port must be greater than 0.')
```
mjac0bs, line 160:
```suggestion
    .min(1, 'Port must be greater than 0.')
```
Currently, this allows a port number of 0.

packages/validation/src/loadbalancers.schema.ts
```diff
@@ -155,36 +155,42 @@ export const UpdateRouteSchema = object({
...
+  }),
+  route_ids: array().of(number()),
+});
+
+export const CreateConfigurationSchema = object({
+  label: string()
+    .min(1, 'Label must not be empty.')
+    .required('Label is required.'),
+  port: number()
+    .min(0, 'Port must be greater than 0.')
```
mjac0bs, line 177:
```suggestion
    .min(1, 'Port must be greater than 0.')
```
See comment above.

packages/validation/src/loadbalancers.schema.ts
```diff
@@ -155,36 +155,42 @@ export const UpdateRouteSchema = object({
 });
 
 export const UpdateConfigurationSchema = object({
-  label: string().min(1),
-  port: number(),
+  label: string().min(1, 'Label must not be empty.'),
+  port: number()
+    .min(0, 'Port must be greater than 0.')
+    .max(65_535, 'Port must be less than 65535.')
```
mjac0bs, line 161: This validation works... but was the `_` intentional or a typo?

```suggestion
    .max(65_535, 'Port must be less than 65535.')
```

packages/validation/src/loadbalancers.schema.ts
```diff
@@ -155,36 +155,42 @@ export const UpdateRouteSchema = object({
...
+  route_ids: array().of(number()),
+});
+
+export const CreateConfigurationSchema = object({
+  label: string()
+    .min(1, 'Label must not be empty.')
+    .required('Label is required.'),
+  port: number()
+    .min(0, 'Port must be greater than 0.')
+    .max(65_535, 'Port must be less than 65535.')
```
mjac0bs, line 178: See comment above.

mjac0bs (commented):
This is looking good from what I've tested when adding and editing configs.

Wanted to confirm:
- Validation on change in the Apply Certificate drawer (see comment/screencap)
- Validation on ports

packages/validation/src/loadbalancers.schema.ts
```diff
@@ -155,36 +155,42 @@ export const UpdateRouteSchema = object({
 });
 
 export const UpdateConfigurationSchema = object({
-  label: string().min(1),
-  port: number(),
+  label: string().min(1, 'Label must not be empty.'),
+  port: number()
+    .min(1, 'Port must be greater than 0.')
+    .max(65535, 'Port must be less than 65535.')
```
mjac0bs, line 161: Sorry, I should have caught this when I caught the `min`:

Should `port` be _less than or equal to_ 65535 or _less than_ 65535?

If it's less than, then we should have 
```suggestion
    .max(65534, 'Port must be less than 65535.')
```
across the board for this field validation.

mjac0bs (approved):
This all looks good! Thanks for the cert drawer revert - I think that method of validation is less alarming for a user. One minor thing: I left a comment on the `max` for `port` that I missed earlier. 😅

PR #9871, opened 2023-11-03
test: [M3-6960] - Add regression test for downloading invoice of maximum size

cliu-akamai:
## Description 📝
Add regression tests for downloading the invoice of maximum size.

## Major Changes 🔄
- Add regression tests to download the invoice with maximum size.
- Add `fs` library to `devDependency`.
- Add new setup plugins `downloadProcess` to clean up the download folder.

## How to test 🧪
```
yarn cy:run -s "cypress/e2e/core/billing/billing-invoices.spec.ts"
```

### DIFF
--- a/packages/manager/cypress/e2e/core/billing/billing-invoices.spec.ts
+++ b/packages/manager/cypress/e2e/core/billing/billing-invoices.spec.ts
@@ -376,8 +376,6 @@ describe('Account invoices', () => {
         displayTime: false,
       });
 
-      console.log('Invoice date: ' + invoiceDate);
-
       // Confirms that it can download the PDF from billing landing page
       mockGetInvoices([mockInvoice]).as('getInvoices');
       mockGetInvoiceItems(mockInvoice, mockInvoiceItems);
@@ -410,12 +408,6 @@ describe('Account invoices', () => {
         timezone,
         displayTime: true,
       });
-
-      console.log('Invoice date: ' + invoiceDate);
-      console.log(
-        'Invoice date with time: ' +
-          invoiceDateWithTime.replace(' ', 'T').replace(':', '_')
-      );
 
       // Confirms that it can download the CSV / PDF from invoice details page
       mockGetInvoice(mockInvoice).as('getInvoice');
--- a/packages/manager/cypress/e2e/core/billing/billing-invoices.spec.ts
+++ b/packages/manager/cypress/e2e/core/billing/billing-invoices.spec.ts
@@ -18,7 +18,7 @@ import { randomItem, randomLabel, randomNumber } from 'support/util/random';
 import { chooseRegion, getRegionById } from 'support/util/regions';
 import { readDownload, cleanUpDownloadFiles } from 'support/util/downloads';
 import {
-  maximumInvoiceSize,
+  maximumInvoiceItems,
   invoicePdfName,
   invoiceCsvName,
 } from 'support/constants/accounts';
@@ -367,8 +367,7 @@ describe('Account invoices', () => {
    */
   it('can download an invoice with maximum size correctly from billing landing page', () => {
     const mockInvoice = invoiceFactory.build();
-    const mockInvoiceItems = invoiceItemFactory.buildList(maximumInvoiceSize);
-    // mockInvoice.date = "2022-08-01T03:00:12";
+    const mockInvoiceItems = invoiceItemFactory.buildList(maximumInvoiceItems);
 
     cy.defer(getProfile()).then((profile: Profile) => {
       const timezone = profile.timezone;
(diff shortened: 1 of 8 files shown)


packages/manager/package.json
```diff
@@ -191,6 +191,7 @@
     "eslint-plugin-testing-library": "^3.1.2",
     "eslint-plugin-xss": "^0.1.10",
     "factory.ts": "^0.5.1",
+    "fs": "^0.0.1-security",
```
jdamore-linode, line 194: I don't think we should have to add any packages to use the `fs` module (and it looks like the NPM [fs package](https://www.npmjs.com/package/fs) is basically a stub).

Did you get an error when you tried to use it? If so, I'm wondering if we need to include `node` in the [Cypress TypeScript config's](https://github.com/linode/manager/blob/develop/packages/manager/cypress/tsconfig.json#L23) `types` field.

packages/manager/cypress/support/plugins/download-processor.ts
```diff
@@ -0,0 +1,23 @@
...
+        try {
+          readdirSync(folderPath).forEach((file) => {
+            rmSync(`${folderPath}/${file}`, { recursive: true });
+          });
+          resolve(null);
+        } catch (err) {
+          reject(err);
+        }
+      });
+    },
```
jdamore-linode, line 21: This makes a lot of sense since the downloaded invoice comes to a whopping 85MB on my machine -- that would add up really quickly.

From my experimenting, it looks like the downloaded invoice has the same filename every time thanks to your mocks, and I think Cypress's download behavior automatically overwrites any existing files so I'm not sure we have to worry about these invoices accumulating. Is there another  situation where this might be required?

packages/manager/package.json
cliu-akamai: No, I don't have any issue when using it. Did you?

packages/manager/cypress/support/plugins/download-processor.ts
cliu-akamai: Also, it might be helpful for other tests that will download items.

packages/manager/cypress/e2e/core/billing/billing-invoices.spec.ts
```diff
@@ -401,4 +416,127 @@ describe('Account invoices', () => {
...
     });
   });
+
+  /*
+   * - Confirms that invoice with maximum size can be downloaded correctly.
+   */
+  it('can download an invoice with maximum size correctly from billing landing page', () => {
+    const mockInvoice = invoiceFactory.build();
+    const mockInvoiceItems = invoiceItemFactory.buildList(maximumInvoiceSize);
+    // mockInvoice.date = "2022-08-01T03:00:12";
```
carrillo-erik, line 426: Do we need to keep this comment here?

packages/manager/package.json
carrillo-erik: As an alternative there is also the [fs-extra](https://www.npmjs.com/package/fs-extra) package which adds extra functionality.

carrillo-erik:
I was able to run the tests locally without issues. All 6 tests are passing ✅

Screenshot 2023-11-08 at 3 25 15 PM

packages/manager/cypress/e2e/core/billing/billing-invoices.spec.ts
cliu-akamai: Good catch, we don't.

mjac0bs (approved):
I had no issues running the billing-invoice.spec.ts locally; it passed consistently. 

I also had a look at the last e2e run. 
`smoke-domain-download-zone-file.spec.ts` passes locally, but failed in CI.
`billing-invoice.spec.ts` still fails in CI.
`clone-linode.spec.ts` was failing last week and I think this would be fixed if you merged the latest `develop` in.

It seems like the two download file specs could be failing due to what Joe mentioned about the lack of a download directory?
> It looks like this is also behind the 2 test failures in CI -- it's basically tripping up here because the downloads directory doesn't exist when this hook is called. That can be fixed pretty easily if we decide this clean up is required!

I'm approving pending that's addressed.

jaalah-akamai (approved):
Tests passed for me - @cliu-akamai if you accept suggested changes, let's just make sure things are good throughout the app 👍

packages/manager/cypress/e2e/core/billing/billing-invoices.spec.ts
```diff
@@ -346,4 +361,126 @@ describe('Account invoices', () => {
...
+      const invoiceDateWithTime = formatDate(mockInvoice.date, {
+        timezone,
+        displayTime: true,
+      });
+
+      console.log('Invoice date: ' + invoiceDate);
+      console.log(
+        'Invoice date with time: ' +
+          invoiceDateWithTime.replace(' ', 'T').replace(':', '_')
+      );
```
coliu-akamai, line 418: same here

coliu-akamai (approved):
confirmed that tests locally passed on my end! ✅ 
saw the latest ci run is failing -- unfortunately I can't see what's causing them, so approving this pr pending that they're unrelated to the changes here!

packages/manager/cypress/e2e/core/billing/billing-invoices.spec.ts
cliu-akamai: No we don't. I will remove them.

jdamore-linode:
> @jdamore-linode @cliu-akamai just circling back on this PR - is this good to go?

@jaalah-akamai I'll try to circle back to this soon, but I think there may be changes we'll want to make to this before merging (cc @cliu-akamai -- I'll keep you posted too). For now I'm going to add the Wait to Merge label.

PR #9878, opened 2023-11-06
upcoming: [M3-7365] - New endpoints, query, types and mock data for Sold Out Plans

coliu-akamai:
## Description 📝
 - Adds api-v4 endpoints and types for Sold Out Plans
- Adds react queries and mock data for Sold Out Plans

## How to test 🧪
This should have no impact (yet) on customer facing codebase.

Verification steps
- Review endpoints, spelling, and types -- cross reference endpoints and types with these two endpoints: [one](https://www.linode.com/docs/api/regions/#regions-availability-list) and [two](https://www.linode.com/docs/api/regions/#region-availability-view)
- If you'd like to test the MSW, you can use the react queries in some file and check the network calls. You can also cross reference the shape of the data returned by the MSW with the data returned from the actual queries
- Verify `/regions/:regionId` endpoint still works as expected (changed `regionID` to `regionId` for the input name)

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/api-v4/src/regions/regions.ts
+++ b/packages/api-v4/src/regions/regions.ts
@@ -1,6 +1,6 @@
 import { API_ROOT } from '../constants';
-import Request, { setMethod, setParams, setURL } from '../request';
-import { Params, ResourcePage as Page } from '../types';
+import Request, { setMethod, setParams, setURL, setXFilter } from '../request';
+import { Filter, Params, ResourcePage as Page } from '../types';
 import { Region, RegionAvailability } from './types';
 
 /**
@@ -39,22 +39,27 @@ export const getRegion = (regionId: string) =>
 export { Region };
 
 /**
+ * getRegionAvailabilities
  *
- *
+ * Returns the availability status for all Linode plans for all regions.
  */
-export const getRegionAvailabilities = (params?: Params) =>
+export const getRegionAvailabilities = (params?: Params, filter?: Filter) =>
   Request<Page<RegionAvailability>>(
     setURL(`${API_ROOT}/regions/availability`),
     setMethod('GET'),
-    setParams(params)
+    setParams(params),
+    setXFilter(filter)
   );
 
 /**
+ * getRegionAvailability
  *
+ * Return the availability status of Linode plans for the given region.
  *
+ * @param regionId { string } The region to get the availabilities for
  */
 export const getRegionAvailability = (regionId: string) =>
-  Request<RegionAvailability>(
+  Request<RegionAvailability[]>(
     setURL(`${API_ROOT}/regions/${encodeURIComponent(regionId)}/availability`),
     setMethod('GET')
   );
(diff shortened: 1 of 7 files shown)


packages/api-v4/src/regions/regions.ts
```diff
@@ -27,13 +27,39 @@ export const getRegions = (params?: Params) =>
  *
  * Return detailed information about a particular region.
  *
- * @param regionID { string } The region to be retrieved.
+ * @param regionId { string } The region to be retrieved.
```
coliu-akamai, line 30: changing this to `regionId` instead of `regionID` as per one of the discussions in our FrontendCafe, since I was working in this file anyway

packages/manager/src/queries/regions.ts
```diff
@@ -17,3 +29,33 @@ export const useRegionsQuery = () =>
...
+// Region Availability queries
+
+export const useRegionsAvailabilitiesQuery = (
+  params: Params,
+  filter: Filter,
+  enabled: boolean = true
+) => {
+  return useQuery<ResourcePage<RegionAvailability>, APIError[]>(
+    [queryKey, 'paginated', params, filter],
+    () => getRegionAvailabilities(params, filter),
```
bnussman-akamai, line 42: Depending on how large we expect this response to be and how we intend to use it, we may want to perform a `getAll` here. I believe we do this for regions.

bnussman-akamai (approved):

packages/manager/src/queries/regions.ts
coliu-akamai: Good point, thanks Banks! iirc the actual data was about 325 region availability objects and we'd probably want to have get it all in one go rather than having to send multiple requests -- I'll look into using getAll here.

I noticed that the Regions query uses placeholder data at first if necessary - do you think we'd need that for this query too, if we used the getAll? cc @abailly-akamai

bnussman-akamai: We shouldn't need placeholder data unless it turns out the API is super slow.

coliu-akamai: Hm.. On my laptop/network at least, the query (first one) takes about 100-150ms to resolve:

image

abailly-akamai: Yeah, get all and no placeholder data is fine - agreed with the approach, i don't thing this endpoint will be probelmatic

packages/manager/src/factories/regions.ts
```diff
@@ -34,3 +40,13 @@ export const regionWithDynamicPricingFactory = Factory.Sync.makeFactory<Region>(
     status: 'ok',
   }
 );
+
+export const regionAvailabilityFactory = Factory.Sync.makeFactory<RegionAvailability>(
+  {
+    available: false,
+    plan: Factory.each((id) =>
+      pickRandom([`g7-premium-${id}`, `g1-gpu-rtx6000-${id}`])
```
abailly-akamai, line 48: you need to add TODOs cause the response is going to change

abailly-akamai (approved):
Looks great, we're getting data!

PR #9887, opened 2023-11-08
fix: [M3-7331] - VPC and StackScript details overflow

coliu-akamai:
## Description + notes on some changes 📝
- Fixes the overflow bug on the VPC and StackScript details page
- Fixes cutoff searchbar on VPC details page 
  - note: placeholder text so far still gets larger when the screen gets smaller (consistent across all searchbars) -- currently forcing a default width of the searchbar on the VPC details page as a result + Also updated the margin for the searchbar and Create Subnet button in the VPCSubnetsTable, due to adding the flex-wrap property so that the search bar and button don't squash each other on smaller screens. this is an alternate fix compared to trying to decrease the placeholder text font (think I just need to target the correct mui classes, but was having trouble with it -- will still look into it)

## Preview 📷

| Before  | After   |
| ------- | ------- |
| image | image |

| Before  | After   |
| ------- | ------- |
| image | image |

### Prerequisites
- With VPC admin tags, using the prod env (or dev env), make sure you have a VPC. Give it a long, one word description
- Create a stackscript and also give it a long description

### Reproduction steps
- Notice how the description overflows for both VPC and StackScript
- Notice how the searchbar placeholder text gets cutoff

### Verification steps 
- Confirm description text is now broken up for VPC and StackScript description for a very long single word
- Confirm VPC and StackScript description work normally without a single very long word
- Confirm search bar text no longer gets cutoff on small screens

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/components/StackScript/StackScript.tsx
+++ b/packages/manager/src/components/StackScript/StackScript.tsx
@@ -49,9 +49,9 @@ const useStyles = makeStyles()((theme: Theme) => ({
     marginTop: theme.spacing(1),
   },
   description: {
-    overflowWrap: 'break-word',
+    overflowWrap: 'anywhere',
     whiteSpace: 'pre-wrap',
-    wordBreak: 'break-word',
+    wordBreak: 'normal',
   },
   descriptionText: {
     marginBottom: theme.spacing(2),
--- a/packages/manager/src/components/StackScript/StackScript.tsx
+++ b/packages/manager/src/components/StackScript/StackScript.tsx
@@ -49,6 +49,7 @@ const useStyles = makeStyles()((theme: Theme) => ({
     marginTop: theme.spacing(1),
   },
   description: {
+    overflowWrap: 'break-word',
     whiteSpace: 'pre-wrap',
     wordBreak: 'break-word',
   },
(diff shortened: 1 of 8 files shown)


packages/manager/src/components/StackScript/StackScript.tsx
carrillo-erik, line 1: Just some code clean up not necessarily related to you changes. We seem to have both, default and named, exports in this file. My suggestion would be to stick with the named export, however, this would require a few minor  additional changes.

`export const StackScript = React.memo((props: StackScriptProps) => {...}`
and remove the `export default` at the end of this file.

coliu-akamai: done!

mjac0bs (approved):
Thanks for fixing both of these! Confirmed that the description field now wraps rather than overflows on StackScript details and VPC details. Search bar placeholder text is no longer cut off.
| StackScripts ✅  | VPC ✅  |
| --- | --- |
|Screenshot 2023-11-14 at 12 58 19 PM | Screenshot 2023-11-14 at 12 58 48 PM|

I think extending the width of the search bar is fine. It seems like there's some default styling for the MUI TextField component that is setting that font size to `1rem` at smaller viewports and that would need to get overridden to go that styling route, but I couldn't make it happy either.

bnussman-akamai (approved):

PR #9900, opened 2023-11-14
fix: [M3-7400] - Linode Network Transfer History - Back button incorrectly disabled

bnussman-akamai:
## Description 📝
 
Fixes back button disabled state on the Linode Details network tab

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-11-14 at 9 51 10 AM | Screenshot 2023-11-14 at 9 57 07 AM |

### Prerequisites
- You must have a Linode that has > 1 month of network data to show on the graph

### Reproduction steps
- Go to http://localhost:3000/linodes/:id/networking
- Observe that the back button appears to be disabled even though clicking it goes back a month

### Verification steps 
- Verify the button appears to be enabled when there is past network history to be viewed

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/NetworkingSummaryPanel/TransferHistory.tsx
+++ b/packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/NetworkingSummaryPanel/TransferHistory.tsx
@@ -1,12 +1,13 @@
 import { Stats } from '@linode/api-v4/lib/linodes';
 import ArrowBackIosIcon from '@mui/icons-material/ArrowBackIos';
-import { Theme, styled, useTheme } from '@mui/material/styles';
+import ArrowForwardIosIcon from '@mui/icons-material/ArrowForwardIos';
+import { IconButton } from '@mui/material';
+import { styled, useTheme } from '@mui/material/styles';
 import { DateTime, Interval } from 'luxon';
 import * as React from 'react';
 
 import PendingIcon from 'src/assets/icons/pending.svg';
 import { Box } from 'src/components/Box';
-import { StyledLinkButton } from 'src/components/Button/StyledLinkButton';
 import { CircleProgress } from 'src/components/CircleProgress';
 import { ErrorState } from 'src/components/ErrorState/ErrorState';
 import { LineGraph } from 'src/components/LineGraph/LineGraph';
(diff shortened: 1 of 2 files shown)


packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/NetworkingSummaryPanel/TransferHistory.tsx
```diff
@@ -200,63 +201,45 @@ export const TransferHistory = React.memo((props: Props) => {
...
+          </IconButton>
         </Box>
       </Box>
       {renderStatsGraph()}
     </div>
   );
 });
 
-const sxArrowIconDisabled = (theme: Theme) => ({
-  cursor: 'not-allowed',
```
mjac0bs, line 245: I'm okay with losing this `cursor` style because the disabled arrow seems clear enough without the  🚫 cursor and I think we only change the cursor on disabled buttons in a handful of places, but wanted to point out it is a change in case it wasn't intended or anyone has other thoughts.

mjac0bs (approved):
Nice fix and good catch with the disabled-but-still-clickable back button in prod. 🔍 

With a linode older than 30 days, I confirmed the back button was enabled as expected until I reached the month of its creation. Styles looked good in both light and dark mode.

packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/NetworkingSummaryPanel/TransferHistory.tsx
cpathipa: I'm in line with showing a 🚫 cursor on actionable items. IMO, it adds an additional layer of visual clarity and facilitates international understanding, indicating that the actionable item is disabled for some reason. This is unless we choose the design option of removing the 🚫. I think this could be a good topic for our cafe meeting to improve consistency.

bnussman-akamai: My fault for not calling out the change in the PR description. **We did lose the `not-allowed` cursor.**

I liked the cursor effect, but I hate to write one-off styles for it... That's why I was okay with removing it. Open to re-adding it

mjac0bs: Since we're not consistent about it across the app already, I think it's not a huge deal to remove it here. But I do think @cpathipa has a good point that this could be worth discussing with UX in cafe to see if we want to apply this cursor treatment more consistently. If there's a consensus of 'yes' that we should, another tech story ticket can revisit this and other places.

PR #9903, opened 2023-11-14
upcoming: [M3-7401] - AGLB Load Balancer and Configuration Refinements

bnussman-akamai:
## Description 📝

- The main goal of the PR is to get AGLB Configuration creation working with the real API. A handful of other small refinements are included

## Changes  🔄
- Adds `hostname` to AGLB landing table
- Make accordion open upon successful config creation
- Removed unneeded `UnusedConfigurationNotice` logic and components
  - The API handled making the association between a config and a load balancer so we don't need to worry about this
- Improves React Query invalidations
- Makes the _Port_ TextField `type="number"` so that a integer gets passed to the API
  - This fixes Configuration creation with the real API
- Adds skeleton loader and error state to Load Balancer Landing ports column
- Improves loading state of Load Balancer Details
- Removes hardcoded health status components
  - This won't be available for alpha. I could have commented this out instead of deleting, but it needed to be made into a component anyway. A ticket will be created when we build a "beta" epic.

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-11-14 at 1 12 55 PM   | <video src="https://github.com/linode/manager/assets/115251059/d80e3bb6-9ac8-414b-ac6c-e6bad4b20e3e" /> |

### Prerequisites
- You should be able to test with the `dev-test-aglb` account in dev

### Verification steps 
- Verify each change listed in the `Changes` section
- The main things to check
  - You can create a Configuration using the real API
  - Check for general regressions, all other changes are minor

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


mjac0bs (commented):
I had a handful of questions when looking at this, mostly because I wasn't sure the exact status of what we are expecting to work and what we're still anticipating errors for. 😅 

Verified:
⚠️  Adds hostname to AGLB landing table - it displays, but see notes below
❌  Make accordion open upon successful config creation
✅ Removed unneeded UnusedConfigurationNotice logic and components
✅ Makes the Port TextField type="number" so that a integer gets passed to the API
✅ Adds skeleton loader and error state to Load Balancer Landing ports column

Questions:
- Should we be expecting a blank `hostname` in the loadbalancer summary right now? (API is not returning this yet, but plans to?)
   - Also, I'm seeing a console error `<div> cannot appear as a descendant of <p>` that seems like it traces back to the `IPAddress` component for hostname
image
- Should we be expecting an Oh Snap crash when trying to reach the Certificates page? (or trying to Add a Certificate in the configuration flow)
- Should we be expecting an Oh Snap crash when trying to `Add a Rule` to a Route?
- Should we be expecting to see the route added to the table in the configuration flow when we select `Add Route` > Create New <protocol> Route? It's only visible in the table when I `Add an Existing Route`.
- Do we need to delete existing configurations to be able to create new ones? When I tried to create a new config, I received the error shown below, so I couldn't confirm successful config creation.

https://github.com/linode/manager/assets/114685994/b17c07d9-a07d-4d37-b4b0-1e12d1adbe34

bnussman-akamai:
> Should we be expecting a blank hostname in the loadbalancer summary right now? 

Yes, it was not populated in the db

> Should we be expecting an Oh Snap crash when trying to reach the Certificates page? 

Yes, the API is in progress of fixing the API shape which should resolve this.

> Should we be expecting an Oh Snap crash when trying to Add a Rule to a Route?

Not anymore. The API just pushed a fix for the shape of the Service Targets response, which should fix this crash.

> Should we be expecting to see the route added to the table in the configuration flow when we select Add Route > Create New Route? It's only visible in the table when I Add an Existing Route.

Looking into this 👀

> Do we need to delete existing configurations to be able to create new ones? When I tried to create a new config, I received the error shown below, so I couldn't confirm successful config creation.

You are probably  getting that error because the name of the configuration is not unique.

packages/manager/src/features/LoadBalancers/LoadBalancerDetail/LoadBalancerDetail.tsx
```diff
@@ -108,7 +118,7 @@ export const LoadBalancerDetail = () => {
         docsLabel="Docs"
         docsLink="" // TODO: AGLB - Add docs link
       />
-      <Tabs index={tabIndex === -1 ? 0 : tabIndex}>
+      <Tabs index={tabIndex === -1 ? 0 : tabIndex} onChange={() => null}>
```
bnussman-akamai, line 121: We don't need an `onChange` because the Tabs are Links, which cause the browser to redirect and we derive `tabIndex` from the URL. I had to do this because ReachUI console.errors when there is no `onChange`

mjac0bs (commented):
I'm back for another round! Thanks for updating the PR description with the notes on the health status removal. And for fixing the console errors!

> Should we be expecting an Oh Snap crash when trying to reach the Certificates page? 

> Yes, the API is in progress of fixing the API shape which should resolve this.

✅ There is no longer a crash when visiting the certificates page.

⚠️ Is this work still in progress? In case it isn't and should be working now, this is what I'm seeing for both the Service Target and TLS Certificates:

Screenshot 2023-11-16 at 2 53 42 PM

> Should we be expecting an Oh Snap crash when trying to Add a Rule to a Route?

> Not anymore. The API just pushed a fix for the shape of the Service Targets response, which should fix this crash.

✅ There is no longer a crash when adding a rule to a route. 
❌  I can add a rule to a route. (I see the following error message.)

https://github.com/linode/manager/assets/114685994/0fc17c57-c296-4977-98bb-ac7aea4f5459

> Should we be expecting to see the route added to the table in the configuration flow when we select Add Route > Create New Route? It's only visible in the table when I Add an Existing Route.

> Looking into this 👀

> Edit: It looks like the API is not behaving as expected. I will monitor and report this to Anthony

✅ I can see the route added to the table in the configuration flow when we select Add Route > Create New Route.
⚠️ Is it a known issue that adding an existing route to a configuration results in adding ALL existing routes? And then, in order to delete one route, all routes must be deleted? (See video.)

https://github.com/linode/manager/assets/114685994/d9b76607-b8e0-47d1-af4b-64aefefc7219

> Do we need to delete existing configurations to be able to create new ones? When I tried to create a new config, I received the error shown below, so I couldn't confirm successful config creation.

> You are probably  getting that error because the name of the configuration is not unique or something is going wrong on the backend.

> Edit: It looks like HTTPS configurations aren't createable. I'll work on this with Anthony. HTTP and TCP should work

✅ I can create a new configuration now of any protocol type.


Found a 🐛 in the Add Rule drawer that I thought I was just imagining in weeks prior, but no, I can repro it.
- When a user fills out some form fields and then goes to click on "Add a Service Target", they have to click the link twice. The first time has no effect. If the user clicks "Add a Service Target" before interacting other parts of the form, one link click works as expected.

https://github.com/linode/manager/assets/114685994/2a4698a9-125c-41c0-8b36-7ef0273a6816

bnussman-akamai:
>  Is it a known issue that adding an existing route to a configuration results in adding ALL existing routes?

Yes. This is a known issue caused by the API not properly filtering.

Looking into the `Add a Service Target` button issue. This one is savage! No idea what's going on

mjac0bs (approved):
Thanks for addressing all those points. What's expected to work in the PR description does work, so approving this! 🚢 

As discussed offline, the issue with the `Add a Service Target` button has to do with the `LinkButton` component (which isn't entirely deprecated but mostly is?). We've switched to `StyledLinkButton`, but with a `type=submit`, we'd be submitting the form at that point and so this will probably need some reworking, to be addressed in a follow up ticket.

cpathipa (approved):
Over all LGTM! verified on creating Configurations. 
- [x] You can create a Configuration using the real API
- [x] Check for general regressions, all other changes are minor

packages/validation/src/loadbalancers.schema.ts
mjac0bs, line 1: Can we add a changeset for validation?

PR #9914, opened 2023-11-16
feat: [M3-7345] – Unrecommended Linode configuration indicators on VPC Detail page

dwiley-akamai:
## Description 📝
Assigning a linode to a VPC but not having the VPC interface be `primary` results in situations where network connectivity issues can arise (see ticket for more details). This PR adds warning indicators on the VPC Detail page when such configurations exist: a warning icon at the `SubnetLinodeRow` level

## Changes  🔄
- Interactive tooltip w/ warning icon added to SubnetLinodeRow when a linode has an unrecommended configuration
- `hasUnrecommendedConfiguration` utility function
- Test updates

## Preview 📷
<details>
<summary>Warning icon</summary>

Screenshot 2023-11-21 at 4 04 26 PM

</details>

### Prerequisites
- An account with proper VPC tags
- A VPC with:
  - a subnet within it
  - at least one linode assigned to it, with the linode having:
    - a public interface in eth0, designated as the primary interface (you can do this in the Linode Edit Config dialog)
    - a VPC interface in eth1 

### Verification steps 
With the prereqs described above, observe:
- The warning icon being present for the linode with an unrecommended configuration

If you go to the linode with an unrecommended configuration and adjust it by making the VPC interface (eth1) the primary interface, returning to the VPC Detail page should show that the warning icon is no longer present for that linode.

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/components/Notice/Notice.tsx
```diff
@@ -79,7 +79,7 @@ export interface NoticeProps extends Grid2Props {
 - Appear within the page or modal
 - Might be triggered by user action
 - Typically used to alert the user to a new service, limited availability, or a potential consequence of the action being taken
-- Consider using a [Dismissible Banner](/docs/components-notifications-dismissible-banners--beta-banner) if it&rsquo;s not critical information
+- Consider using a [Dismissible Banner](/docs/components-notifications-dismissible-banners--beta-banner) if it’s not critical information
```
dwiley-akamai, line 82: This fixes this issue at design.linode.com:

Screenshot 2023-11-15 at 9 16 44 PM

packages/manager/src/features/VPCs/VPCDetail/VPCDetail.tsx
```diff
@@ -42,6 +47,11 @@ const VPCDetail = () => {
   const [deleteVPCDialogOpen, setDeleteVPCDialogOpen] = React.useState(false);
   const [showFullDescription, setShowFullDescription] = React.useState(false);
 
+  const [
+    unrecommendedConfigurationPresent,
+    setUnrecommendedConfigurationPresent,
+  ] = React.useState(false);
```
dwiley-akamai, line 53: Grabbing all the configs in this component could get expensive and messy, so this gets passed down and is used to decide whether the dismissible banner gets displayed or not.

coliu-akamai:
Should the warning icon and dismissible banner supposed to show up for all linode statuses, including when the linode is inactive?

Also, when I switch the linode from a good config profile one to an unrecommended one, I need to reload the page in order to get the icon and banner to show up (when the linode is booting/active. If the linode's inactive, then the banner/icon don't show up after a page reload) -- see video: 

Screen going white/loading is me reloading the page. (also deleting this linode now)
<video src="https://github.com/linode/manager/assets/139280159/725eaa50-9f1f-4ab1-ae5d-b3b32e438fd6" />

dwiley-akamai:
> Should the warning icon and dismissible banner show up for all linode statuses, including when the linode is inactive, or just when it's booting/online?

Confirmed with UX last Friday that we don't want these indicators displayed when the linode is inactive 👍 

> When I switch the linode from a good config profile one to an unrecommended one, I need to reload the page in order to get the icon and banner to show up (when the linode is booting/active. If the linode's inactive, then the banner/icon don't show up after a page reload) -- see video:

What was happening here is: the linode is off in order to change the config --> config is changed, query gets refetched and `active` shows as `false` for the interfaces because the linode is off --> turn the linode back on, query doesn't get refetched. This should be fixed now

hana-linode (commented):
It seems a bit odd to me for the warning text to be separate (i.e. in a notice) from the warning icon. What about removing the notice and adding the text as a tooltip using the `TooltipIcon` component? 

https://design.linode.com/?path=/docs/components-tooltip-tooltip-icon--documentation

image

packages/manager/src/features/ObjectStorage/ObjectUploader/ObjectUploader.tsx
dwiley-akamai, line 1: autoformatting from the linter

coliu-akamai:
thanks for confirming @dwiley-akamai! Still getting the warning indicator to show up for inactive linodes. This happens if the linode had the warning indicator while online -- it stays after the linode is shut down. Might need to invalidate config queries as the linode is shutting down as well then?
image

packages/manager/src/features/VPCs/utils.ts
```diff
@@ -26,3 +26,24 @@ export const getSubnetInterfaceFromConfigs = (
 
   return undefined;
 };
+
+export const hasUnrecommendedConfiguration = (configs: Config[]) => {
```
coliu-akamai, line 30: could we also add some tests for this utils function?

coliu-akamai (commented):
✅ SubnetLinodeRow.test.tsx tests pass
✅ Warning icon + tooltip show up as expected when Linode is online - see comment above for offline linode

Regarding the warning icon status, how does it work for linodes with multiple configurations? Is it a similar situation to the Reboot Needed status, where it might get super confusing for linodes w multiple configs? 👀

dwiley-akamai:
> Regarding the warning icon status, how does it work for linodes with multiple configurations? Is it a similar situation to the Reboot Needed status, where it might get super confusing for linodes w multiple configs? 👀

It's a similar situation unfortunately: I created a linode with an unrecommended config profile and a recommended config profile, assigning the former to one subnet and the latter to another subnet. I booted it with the unrecommended config profile and the icon displays for both of the linode's appearances in the table:

Screenshot 2023-11-21 at 5 48 46 PM

coliu-akamai:
@dwiley-akamai thanks for confirming! Is UX good with this behavior? Otherwise maybe you could pass the subnet ID into the `hasUnrecommendedConfiguration` method and then use that to determine that whether or not the warning indicator should be displayed?

dwiley-akamai:
> @dwiley-akamai thanks for confirming! Is UX good with this behavior? Otherwise maybe you could pass the subnet ID into the `hasUnrecommendedConfiguration` method and then use that to determine that whether or not the warning indicator should be displayed?

What did you have in mind with the subnet ID? I'm not sure just the subnet ID is enough on its own. I wrote this out and the logic seems to work:

<details>
<summary>Revised hasUnrecommendedConfiguration() and new helper function</summary>

```typescript
export const hasUnrecommendedConfiguration = (
  configs: Config[],
  subnet: Subnet | undefined
) => {
  const subnetAssignedLinodeData = subnet?.linodes;

  for (const config of configs) {
    const configInterfaces = config.interfaces;

    /*
     If there is a VPC interface marked as active but not primary, we want to display a
     message re: it not being a recommended configuration.

     Rationale: when the VPC interface is not the primary interface, it can communicate
     to other VMs within the same subnet, but not to VMs in a different subnet
     within the same VPC.
    */

    return configInterfaces.some(
      (_interface) =>
        _interface.active &&
        _interface.purpose === 'vpc' &&
        !_interface.primary &&
        interfaceInSubnet(_interface.id, subnetAssignedLinodeData)
    );
  }

  return false;
};

export const interfaceInSubnet = (
  interfaceId: number,
  subnetAssignedLinodeData: SubnetAssignedLinodeData[] | undefined
) => {
  return subnetAssignedLinodeData?.some((subnetAssignedLinodeDatum) => {
    const assignedInterfaces = subnetAssignedLinodeDatum.interfaces;
    return assignedInterfaces.some(
      (assignedInterface) =>
        assignedInterface.id === interfaceId && assignedInterface.active
    );
  });
};
```
</details>

In the scenario I described in my previous comment, this logic has the warning displayed in the subnet for the linode whose VPC interface has the unrecommended config, and not in the other subnet (same linode but whose VPC interface has the recommended config).

coliu-akamai:
@dwiley-akamai I wasn't clear, apologies. I meant pass in the subnet_id **in addition to the configs you were already passing in, which would allow you do do what you wrote above ^, to check if an interface was associated with the subnet_id you passed in. Glad to hear that works ^!

coliu-akamai:
@dwiley-akamai Wednesday before Thanksgiving shenanigans 😅 - I just looked over what you wrote above again, and realized you had a different approach. I had more in mind that you could compare the `interface.subnet_id` to the passed in `subnet_id`, since interfaces should also have a subnet_id field if they're assigned to a vpc. You might not need the helper function `interfaceInSubnet` if so - you could compare directly in at line 53 if `_interface.subnet_id === subnet_id`

Will be re-reviewing this shortly!

packages/manager/src/features/VPCs/utils.ts
```diff
@@ -26,3 +26,46 @@ export const getSubnetInterfaceFromConfigs = (
 
   return undefined;
 };
+
+export const hasUnrecommendedConfiguration = (
```
coliu-akamai, line 30: Seems like this function short circuits and might lead to the warning icon not displaying when it should be for linodes with multiple configs. I set up a linode with two configs, and made the second config have the active VPC interface. Both configs have unrecommended setups
- first config leads to the correct warning indicator as expected (no indicator because its VPC interface is not active)
- the linode using the second config doesn't have a warning indicator even though its VPC interface is active and not primary
- we never get to check the second config bc we always return some boolean value based on the first config's interfaces (line 48)

| 2nd Config's interfaces  | but no warning indicator |
| ------- | ------- |
| image | image |

maybe we could rewrite it to be something like this - I used the subnetId in this case instead of passing in the entire subnet object 
- (lightly tested this - one unit test for `hasUnrecommendedConfiguration` fails with this, but it seems to be due to mismatching subnet IDs. changing the subnet id in the test at line 103 in `utils.test.ts` to the subnet_id of the mock interface got the test to pass, which I think is expected)

```js
export const hasUnrecommendedConfiguration2 = (
  configs: Config[],
  subnetId: number
) => {
  for (const config of configs) {
    const configInterfaces = config.interfaces;

    /*
     same comments
    */

    if (
      configInterfaces.some((_interface) => _interface.subnet_id === subnetId)
    ) {
      return configInterfaces.some(
        (_interface) =>
          _interface.active &&
          _interface.purpose === 'vpc' &&
          !_interface.primary &&
          _interface.subnet_id === subnetId // this check may be redundant due to the if condition
      );
    }
  }

  return false;
};
```
(May need to revisit this logic if the ability for a single config to have multiple vpc interfaces is revisited)

dwiley-akamai:
@coliu-akamai Thanks for the thorough testing and review -- with the most recent changes I think the expected indicators should be displayed

coliu-akamai (approved):
Tested ✅ 
- 1 config linode unrecommended config
- 1 config linode recommended config 
- 2 configs linode ✅ 
    - first one in use and unrecommended, second one recommended but not in use (warning indicator for first)
    - first one not in use and unrecommended, second one in use and recommended (no warning indicators)
    - first one in use and recommended, second one unrecommended but not in use (no warning indicators)
    - first one not in use and recommended, second on in use and unrecommended (warning indicator for second)
    - both unrecommended and first one in use  (warning indicator for first)
    - both unrecommended and second one in use  (warning indicator for second)
    - both recommended, first one in use  (no warning indicator)
    - both recommended, second one in use (no warning indicator)

The indicators worked as expected -- thanks Dajahi! 🚀

coliu-akamai:
Ah - one small thing I was thinking about in terms of accessibility for the warning indicator: is there a way to use the keyboard to click on the `Configuration Profiles` link that appears in the tooltip? I'm able to use the keyboard to get the tooltip to show up and have the contents be read in the voiceover, but couldn't figure out how to click on the link

hana-linode (commented):
Confirmed that the warning icon is visible in the table for Linodes with an unrecommended configuration 👍 

Should we also add a warning notice to the Linode's details page about the unrecommended configuration?

hana-linode (approved):

dwiley-akamai:
Created M3-7486 for investigating the accessibility question

PR #9929, opened 2023-11-22
refactor: [M3-7437] - Update Profile Queries

bnussman-akamai:
> [!important]
> Typecheck, Unit Tests, and E2E test will fail and Cloud Manager will not function yet
> Things will start slowly working again as we continue refactor 

## Description 📝

- Adds new package to handle typesafe query keys 🔑📦
- Refactors all profile related queries and mutations to use new format and query key ♻️

## How to test 🧪

- **Nothing is going to work yet** (things should start working after we update account queries, which I will do next)
- Check the diff and verify the changes are solid

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/queries/images.ts
```diff
@@ -55,7 +55,7 @@ export const useCreateImageMutation = () => {
       onSuccess() {
         queryClient.invalidateQueries(`${queryKey}-list`);
         // If a restricted user creates an entity, we must make sure grants are up to date.
-        queryClient.invalidateQueries([PROFILE_QUERY_KEY, 'grants']);
+        queryClient.invalidateQueries(profileQueries.grants.queryKey);
```
bnussman-akamai, line 58: Because we use `profileQueries` as a source of truth, we don't have to worry about accidentally messing up query keys.

packages/manager/src/queries/profile.ts
```diff
@@ -33,115 +32,146 @@ import {
...
+
 export const getGrantData = (queryClient: QueryClient) =>
-  queryClient.getQueryData<Grants>([queryKey, 'grants']);
+  queryClient.getQueryData<Grants>(profileQueries.grants.queryKey);
 
 export const useSMSOptOutMutation = () => {
   const queryClient = useQueryClient();
-  return useMutation<{}, APIError[]>(smsOptOut, {
+  return useMutation<{}, APIError[]>({
+    mutationFn: smsOptOut,
```
bnussman-akamai, line 117: Notice how we now pass a single object to `useMutation` and `useQuery` instead of many parameters. This will be a breaking change in v5 so we're going to start doing it now as we refactor. (https://tanstack.com/query/v5/docs/react/guides/migrating-to-v5#breaking-changes)

packages/manager/src/queries/securityQuestions.ts
```diff
@@ -1,77 +1,31 @@
...
-    (data) => {
-      return updateSecurityQuestions(data);
+  >({
+    mutationFn: updateSecurityQuestions,
+    onSuccess() {
+      queryClient.invalidateQueries(profileQueries.securityQuestions.queryKey);
     },
-    {
-      onSuccess: (response) => {
-        queryClient.setQueryData<SecurityQuestionsData | undefined>(
```
bnussman-akamai, line 31: This is way too complicated. Let's just invalidate.

bnussman-akamai:
The query key factory isn't what I hoped it would be. Going to reassess

packages/manager/src/queries/securityQuestions.ts
```diff
@@ -1,77 +1,31 @@
...
     SecurityQuestionsPayload,
     APIError[],
     SecurityQuestionsPayload
-  >(
-    (data) => {
-      return updateSecurityQuestions(data);
+  >({
+    mutationFn: updateSecurityQuestions,
+    onSuccess() {
+      queryClient.invalidateQueries(profileQueries.securityQuestions.queryKey);
```
cpathipa, line 28: Not sure if it has an explicit queryKey property defined in the configuration.

packages/manager/src/queries/profile.ts
```diff
@@ -152,34 +181,33 @@ export const sshKeyEventHandler = (event: EventWithStore) => {
...
 };
 
 export const useDisableTwoFactorMutation = () => {
   const queryClient = useQueryClient();
-  return useMutation<{}, APIError[]>(disableTwoFactor, {
+  return useMutation<{}, APIError[]>({
+    mutationFn: disableTwoFactor,
     onSuccess() {
-      queryClient.invalidateQueries([queryKey]);
+      queryClient.invalidateQueries(profileQueries.info.queryKey);
```
cpathipa, line 210: Not sure if it has an explicit queryKey property defined in the configuration.

cpathipa (commented):
Looks like some of the `profileQueries` don't have an explicit `queryKey` property defined.

packages/manager/src/queries/securityQuestions.ts
bnussman-akamai: The query key `profileQueries.securityQuestions.queryKey` isn't explicitly defined, but when `getQueryKeys` in run, it generates the query key based on the objects shape!

PR #9938, opened 2023-11-27
upcoming: [M3-7301] - Replace Linode Network Transfer History chart with Recharts

hana-linode:
## Description 📝
This PR is part of a bigger effort to replace chart.js with the more modern Recharts. As a first step, replace the Linode Network Transfer History chart.

## Preview 📷
| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/linode/manager/assets/115299789/3376b759-9062-4e5f-a1ea-db283660eb5b" /> | <video src="https://github.com/linode/manager/assets/115299789/0bc10953-bda9-4290-be3f-470311a9b4de" /> |

### Prerequisites
- Have a Linode that's been running for a while

### Verification steps 
- Go to the Network tab of a Linode that's been running for a while
- Confirm the updated UI and that there are no regressions in the graph, units, etc

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/NetworkingSummaryPanel/NetworkTransferHistoryChart.tsx
+++ b/packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/NetworkingSummaryPanel/NetworkTransferHistoryChart.tsx
@@ -4,11 +4,13 @@ import {
   Area,
   AreaChart,
   CartesianGrid,
+  ResponsiveContainer,
   Tooltip,
   XAxis,
   YAxis,
 } from 'recharts';
 
+import { Box } from 'src/components/Box';
 import { NetworkUnit } from 'src/features/Longview/shared/utilities';
 import { roundTo } from 'src/utilities/roundTo';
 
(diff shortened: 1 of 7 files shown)


packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/NetworkingSummaryPanel/NetworkTransferHistoryChart.tsx
```diff
@@ -0,0 +1,83 @@
...
+    return DateTime.fromMillis(t, { zone: timezone }).toFormat(
+      'LLL dd, yyyy, h:mm a'
+    );
+  };
+
+  const tooltipValueFormatter = (value: number) =>
+    `${roundTo(value)} ${unit}/s`;
+
+  return (
+    <Box marginLeft={-4}>
```
hana-linode, line 42: For some reason, the generated graph viewport is not centered so this left margin makes it look centered

image

jdamore-linode: I'm guessing Recharts does this to leave room for the Y axis label, so I think this solution makes a lot of sense in our case

bnussman-akamai (commented):
This is awesome. I am very eager to get Chart.js out of here 🗑️


Can we improve the dark mode experience a bit? 

Screenshot 2023-11-29 at 10 50 45 AM

jdamore-linode (commented):
Nice work @hana-linode, this is a nice improvement over Chart.js!

Functionality wise I don't see any issues; I played around with it using Chrome and Firefox (and tested using VoiceOver on both) and didn't catch any issues.

Besides the dark theme issues Banks pointed out, the only definite accessibility issue I’m seeing is the green "Public Outbound Traffic" text on the tooltip. In Prod this is black and bold -- the green doesn't pass WebAIM's contrast checker.

Screenshot 2023-11-29 at 9 50 34 AM
Screenshot 2023-11-29 at 10 50 14 AM

The only other thing that might be worth raising attention to is the animations. They look nice but are also kind of slow and easily triggered (by clicking the prev and next buttons, and by resizing the window, etc.). I'm wondering if it would be worthwhile to disable the animations using the `prefers-reduced-motion` media query (I'm not aware of anywhere else where we do this in Cloud, but I think it would be a nice thing to do).

packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/NetworkingSummaryPanel/NetworkTransferHistoryChart.tsx
```diff
@@ -0,0 +1,83 @@
...
+      <ResponsiveContainer height={190} width="100%">
+        <AreaChart
+          margin={{
+            bottom: 5,
```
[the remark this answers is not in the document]
hana-linode: Good point, none of these margins are needed since we have the Box margin. Removed!

bnussman-akamai:
Dark mode looks much better now, but would it be possible for the tooltip to have a dark gray background with white text instead of a while background? I think that would look more fitting with our theme.

Hopefully we can do this without adding a ton of extra code. If we need to, we can define a custom component to use as the tooltip (https://github.com/bnussman/managernext/blob/main/src/linodes/Stats.tsx#L121-L132)

carrillo-erik:
Nice, I really like the refreshed look of the chart. I tested the changes in mobile view and only noticed a small visual regression, however, it appears that the dates displayed to the user are also different.

| OLD | NEW |
| ------- | ------- |
| bb22 | aa11 |

jdamore-linode (approved):
Looks good @hana-linode!

hana-linode:
@carrillo-erik Updated x-axis intervals. The dates will be slightly different but there's not much we can do without increased code complexity. The information is still correct, recharts is just choosing to display a different date

bnussman-akamai (approved):
This is awesome! 

I'd love to see the NodeBalancer graphs done next!

PR #9939, opened 2023-11-28
M3-7315 Add tests assgining VPCs during Linode creation flow

cliu-akamai:
## Description 📝
Add regression tests for assigning VPCs during Linode creation flow.

## Major Changes 🔄
- Add regression tests to assign a VPC when creating a Linode.

## How to test 🧪
```
```


packages/manager/cypress/e2e/core/linodes/create-linode.spec.ts
```diff
@@ -54,8 +66,39 @@ const mockRegions: Region[] = [
...
+    },
+  ];
+  const mockVLANs: VLAN[] = VLANFactory.buildList(2);
+  const mockVPCs: VPC[] = vpcFactory.buildList(5, {
+    region: 'us-southeast',
+  });
+
   before(() => {
     cleanUp('linodes');
+    mockConfig.interfaces.splice(2, 1);
```
jdamore-linode, line 101: Can we create the `mockConfig` object with the expected interfaces rather than modifying them after the fact?

packages/manager/cypress/e2e/core/linodes/create-linode.spec.ts
```diff
@@ -54,8 +66,39 @@ const mockRegions: Region[] = [
...
+      created: '2020-08-21T17:26:14',
+      updated: '2020-08-21T17:26:31',
+      filesystem: 'swap',
+      size: 512,
+    },
+  ];
+  const mockVLANs: VLAN[] = VLANFactory.buildList(2);
+  const mockVPCs: VPC[] = vpcFactory.buildList(5, {
+    region: 'us-southeast',
+  });
```
jdamore-linode, line 97: Can we move these to the VPC assignment test since they're not shared by any of the other tests?

packages/manager/cypress/e2e/core/linodes/create-linode.spec.ts
```diff
@@ -318,4 +361,84 @@ describe('create linode', () => {
...
+    ui.regionSelect.findItemByRegionLabel(region.label).click();
+    fbtClick('Shared CPU');
+    getClick(`[id="${dcPricingMockLinodeTypes[0].id}"]`);
+
+    // the "VPC" section is present, and the VPC in the same region of
+    // the linode can be selected.
+    getVisible('[data-testid="vpc-panel"]').within(() => {
+      containsVisible('Assign this Linode to an existing VPC.');
+      // VPC in different region should not be available.
+      containsVisible('VPC is not available in the selected region.');
```
coliu-akamai, line 423: This text shows up when the region we've selected doesn't support VPCs. Since we're testing assigning a Linode to a VPC during creation, can we mock a region that supports VPCs and confirms this text doesn't show up? Then, could we test the user flow for assigning a VPC? It looks like this:

https://github.com/linode/manager/assets/139280159/d3d76da6-ec5d-4b93-82f1-28083dc6f68c

coliu-akamai (commented):
Thanks Cassie! Left a comment below. Could we also add a changeset for this PR?

packages/manager/cypress/e2e/core/linodes/create-linode.spec.ts
```diff
@@ -318,4 +361,84 @@ describe('create linode', () => {
...
+    });
```
jdamore-linode, line 424: I think we'll need to do more here -- it's good that we're confirming the state of the UI when VPC isn't available, but we primarily want this test to be able to select a VPC and proceed with the Linode creation.

I think this might require some changes and additions to the mocks, specifically:

- `mockGetVPCs` should be passed an array of VPC objects which contains the `mockVPC` object you created -- this should allow the mock to appear in the list of VPCs when attempting to select it
- We'll probably need to [mock the regions request](https://github.com/linode/manager/blob/develop/packages/manager/cypress/support/intercepts/regions.ts#L16) so that `us-southeast` has the `VPCs` capability (we do [something similar in the VPC create test](https://github.com/linode/manager/blob/develop/packages/manager/cypress/e2e/core/vpc/vpc-create.spec.ts#L53-L55)).

packages/manager/cypress/e2e/core/linodes/create-linode.spec.ts
```diff
@@ -318,4 +361,84 @@ describe('create linode', () => {
...
+      '@getFeatureFlags',
+      '@getVPCs',
+    ]);
+
+    mockGetVLANs(mockVLANs);
+    mockGetVPC(mockVPC).as('getVPC');
+    mockCreateLinode(mockLinode).as('linodeCreated');
+    mockGetLinodeConfigs(mockLinode.id, [mockConfig]).as('getLinodeConfigs');
+    mockGetLinodeDisks(mockLinode.id, mockDisks).as('getDisks');
+    mockGetLinodeVolumes(mockLinode.id, []).as('getVolumes');
```
jdamore-linode, line 407: We'll probably want to move these calls before `cy.visitWithLogin()` -- putting them here _usually_ works, but it can be flaky because it's possible for Cloud to make these HTTP requests before Cypress sets up the intercepts.

jdamore-linode (changes requested):
Great start @cliu-akamai! I left some feedback -- feel free to reach out if you have any questions!

packages/manager/cypress/e2e/core/linodes/create-linode.spec.ts
```diff
@@ -321,4 +336,181 @@ describe('create linode', () => {
     fbtVisible(linodeLabel);
     cy.contains('RUNNING', { timeout: 300000 }).should('be.visible');
   });
+
+  it('assigns a VPC not enabled in the region to the linode during create flow', () => {
```
coliu-akamai, line 340:
```suggestion
  it('prevents a VPC from being assigned in a region that doesn't support VPCs during the Linode Create flow', () => {
```

packages/manager/cypress/e2e/core/linodes/create-linode.spec.ts
```diff
@@ -321,4 +336,181 @@ describe('create linode', () => {
...
+    ui.regionSelect.find().click().type(`${region.label} {enter}`);
+    fbtClick('Shared CPU');
+    getClick(`[id="${dcPricingMockLinodeTypes[0].id}"]`);
+
+    // the "VPC" section is present
+    getVisible('[data-testid="vpc-panel"]').within(() => {
+      containsVisible(
+        'Allow Linode to communicate in an isolated environment.'
+      );
+      // VPC in different region should not be available.
```
coliu-akamai, line 377:
```suggestion
      // Helper text appears if VPC is not available in selected region
```

coliu-akamai (approved):
awesome, thanks Cassie!

It may or may not be out of scope for this ticket - should we also have a test for creating a VPC in the Linode Create flow and assigning it to this Linode? I can create a new ticket for it if needed @jdamore-linode

jaalah-akamai:
I had 3 tests fail for me:
Screenshot 2024-01-17 at 10 14 22 AM

packages/manager/cypress/e2e/core/linodes/create-linode.spec.ts
```diff
@@ -318,4 +333,181 @@ describe('create linode', () => {
...
+
+    getClick('#linode-label').clear().type(linodeLabel);
+    cy.get('#root-password').type(rootpass);
+    getClick('[data-qa-deploy-linode]');
+    cy.wait('@linodeCreated').its('response.statusCode').should('eq', 200);
+    fbtVisible(linodeLabel);
+    cy.contains('RUNNING', { timeout: 300000 }).should('be.visible');
+
+    fbtClick('Configurations');
+    cy.wait(['@getLinodeConfigs', '@getVPC', '@getDisks', '@getVolumes']);
```
hkhalil-akamai, line 503: Seeing this error repeatedly across runs -- I think it originates from this line.
Screenshot 2024-01-22 at 4 42 18 PM

hkhalil-akamai (changes requested):
Tests overall look good. Highlighting one small issues causing failures for me locally.

packages/manager/cypress/e2e/core/linodes/create-linode.spec.ts
hkhalil-akamai: Might be related to this 404? Do we need to add a mock for this route?
Screenshot 2024-01-22 at 4 44 34 PM

jdamore-linode (approved):
Thanks @hkhalil-akamai! Just pushed a couple commits that should resolve the merge conflict and fix this failure.

@cliu-akamai Nice work! Approved pending Cypress test results -- thanks for addressing the feedback and sorry for the slow follow up on my end!

bnussman-akamai (approved):

hkhalil-akamai (approved):
Thanks for addressing feedback!

PR #9944, opened 2023-11-29
feat: [M3-7412] - Add new Parent/Child api-v4 endpoints

jaalah-akamai:
## Description 📝
New Parent/Child @linode/api-v4 endpoints:
- `getChildAccounts`
- `getChildAccount`
- `createChildAccountPersonalAccessToken`

## Changes  🔄
- Fixed our `request.tsx` to not override headers that are explicitly passed in via `setHeaders()`.
- New endpoints use a destructured syntax so we can explicitly state which properties we're interested in, without having to define unused arguments, i.e.: `const { data: linodes } = useLinodesQuery({}, {}, shouldCheckLinodeCount);` 

## Preview 📷
Screenshot 2023-11-30 at 9 11 26 PM

## How to test 🧪
**Test for Request.test.tsx**
1. Checkout branch
2. Run `yarn test src/request.test.tsx`
3. Observe passing tests

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


jaalah-akamai:
Actually I will be added a unit test to `Request.test.tsx`, but everything else is ready for review - I'll get that done in a bit

packages/api-v4/src/types.ts
```diff
@@ -31,6 +33,12 @@ export interface Params {
   page_size?: number;
 }
 
+export interface RequestConfig {
```
bnussman-akamai, line 36: RequestConfig makes sense. I wonder if `RequestOptions` or `PaginatedOptions` or `PaginatedRequestOptions` would be any better?

jaalah-akamai: Open to suggestions! I like `RequestOptions`

packages/manager/src/factories/grants.ts
```diff
@@ -41,6 +41,7 @@ export const grantsFactory = Factory.Sync.makeFactory<Grants>({
     add_volumes: true,
     add_vpcs: true,
     cancel_account: false,
+    child_account_access: null,
```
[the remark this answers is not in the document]
jaalah-akamai: Just got confirmation - it's a boolean, I'll update here

packages/manager/src/queries/account.ts
```diff
@@ -31,3 +41,45 @@ export const useMutateAccount = () => {
...
+export const useChildAccount = ({ euuid }: ChildAccountPayload) => {
+  const { data: grants } = useGrants();
+  return useQuery<Account, APIError[]>(
+    [queryKey, 'childAccounts', 'childAccount', euuid],
+    () => getChildAccount({ euuid }),
+    { enabled: Boolean(grants?.global?.child_account_access) }
+  );
+};
+
+export const useCreateChildAccountPersonalAccessTokenMutation = ({
```
jaalah-akamai, line 67: @bnussman-akamai I wanted to get your opinion on this! Although this is a POST request, we're not updating/modifying any data on the server so I changed it to use `useQuery`. There's no endpoint to GET these tokens once they're issued and their short-lived.

jaalah-akamai: Speaking of which I could probably drop the `Mutation` part of this name too

packages/api-v4/.changeset/pr-9944-upcoming-features-1701281121547.md
```diff
@@ -0,0 +1,5 @@
+---
+"@linode/api-v4": Upcoming Features
+---
+
+Add new endpoints for Parent/Child ([#9944](https://github.com/linode/manager/pull/9944))
```
dwiley-akamai, line 5:
```suggestion
Add new endpoints for Parent/Child account switching ([#9944](https://github.com/linode/manager/pull/9944))
```

packages/api-v4/src/account/account.ts
```diff
@@ -149,3 +152,53 @@ export const signAgreement = (data: Partial<Agreements>) => {
...
+/**
+ * createChildAccountPersonalAccessToken
+ *
+ * This endpoint will allow Parent Account Users with the "child_account_access" grant to
+ * create an ephemeral token for their proxy user on a child account, using the euuid of
+ * that child account. As noted in previous sections, this Token will inherit the
+ * permissions of the Proxy User, and the token itself will not be subject to additional
+ * restrictions.
+ *
+ * setHeaders() will be used for creating tokens from within the proxy account.
```
mjac0bs, line 191: Won't we also need `setHeaders` logic within `getChildAccounts` to make the GET request to populate the Switch Account drawer from the proxy user?

packages/manager/.changeset/pr-9944-upcoming-features-1701281085057.md
```diff
@@ -0,0 +1,5 @@
+---
+"@linode/manager": Upcoming Features
+---
+
+Add new queries, grants for Parent/Child ([#9944](https://github.com/linode/manager/pull/9944))
```
dwiley-akamai, line 5:
```suggestion
Add new grants and RQ queries for Parent/Child account switching ([#9944](https://github.com/linode/manager/pull/9944))
```

packages/api-v4/src/types.ts
```diff
@@ -31,6 +33,12 @@ export interface Params {
   page_size?: number;
 }
 
+export interface RequestOptions {
+  params?: Params;
+  filter?: Filter;
+  headers?: AxiosRequestHeaders;
```
bnussman-akamai, line 39: Thinking out loud as someone who strongly dislikes Axios: I wonder if we should just accept an object instead of `AxiosRequestHeaders`. It would decrease the reliance on Axios in case we ever remove it

packages/manager/src/queries/account.ts
bnussman-akamai: Hmm, this is a hard one. Judging by the name `createChildAccountPersonalAccessToken`, it sounds like we **are** in fact updating/modifying on the server. 

Is the goal here to store a child token in the React Query cache for often re-use? If so, I think a query makes sense here. If not, then I think a mutation would be better.

dwiley-akamai: I think it may depend on when and how the POST request would be made. If it would be made only upon the user selecting a child account to switch to, `useMutation` feels more correct to me ([declarative and no shared state](https://tkdodo.eu/blog/mastering-mutations-in-react-query#differences-to-usequery) for these ephemeral tokens)

jaalah-akamai:
> Is the goal here to store a child token in the React Query cache for often re-use
Exactly

packages/api-v4/src/types.ts
jaalah-akamai: Yea we can do this pretty easily 👍

jaalah-akamai:
@bnussman-akamai We're going to update this to a mutation due to a recent ask of needing to revoke tokens 👍

packages/manager/src/queries/account.ts
```diff
@@ -31,3 +41,49 @@ export const useMutateAccount = () => {
...
+  euuid,
+  headers,
+}: ChildAccountPayload) =>
+  useMutation<Token, APIError[], ChildAccountPayload>(() =>
+    createChildAccountPersonalAccessToken({ euuid, headers })
+  );
```
jaalah-akamai, line 89: The way this will work is from the Parent, we'll create an ephemeral token for the child account in order to log in. While in the Proxy account, we'll be able revoke this the typical by hitting `/profile/tokens/{tokenId}`.

jaalah-akamai:
Unit test added 🚀 This should be good now

mjac0bs (approved):
Thanks for the unit test. CI looks good; that one e2e failure on rescue-linode is unrelated. 

Let's 🚢 !

packages/api-v4/src/account/account.ts
```diff
@@ -115,7 +118,7 @@ export const getAccountAgreements = () =>
  *
  */
 export const getAccountAvailabilities = (params?: Params, filter?: Filter) =>
-  Request<Page<AccountAvailability>>(
```
bnussman-akamai, line 118: ++

packages/api-v4/src/account/account.ts
```diff
@@ -18,8 +19,10 @@ import {
   CancelAccountPayload,
   Agreements,
   RegionalNetworkUtilization,
+  ChildAccountPayload,
 } from './types';
-import { Filter, ResourcePage as Page, Params } from '../types';
+import { Filter, ResourcePage, Params, RequestOptions } from '../types';
+import { Token } from '../profile';
```
bnussman-akamai, line 25: Optional

```suggestion
import type { Filter, ResourcePage, Params, RequestOptions } from '../types';
import type { Token } from '../profile';
```

bnussman-akamai (approved):

PR #9972, opened 2023-12-06
test: Update notification language in test to correspond with API change

jdamore-linode:
## Description 📝
Fixes the `resize-linode.spec.ts` test by updating the notice string to match the language used by the API since the release of unified migrations.

## Changes  🔄
- Replaces a notification string

## How to test 🧪
We can rely on CI for this, but to run the test locally you can use `yarn && yarn build && yarn start:manager:ci`, and then:

```bash
yarn cy:run -s "cypress/e2e/core/linodes/resize-linode.spec.ts"
```
(Only the first test is modified by this PR, and it's the only test that's failing, so you don't have to wait for the entire spec to run since these are some of our most time consuming tests.)

*Check all that apply*

- [ ] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/.changeset/pr-9972-tests-1701888586591.md
+++ b/packages/manager/.changeset/pr-9972-tests-1701888586591.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Tests
+---
+
+Fix warm resize test by updating notice text ([#9972](https://github.com/linode/manager/pull/9972))


packages/manager/cypress/e2e/core/linodes/resize-linode.spec.ts
```diff
@@ -38,7 +38,7 @@ describe('resize linode', () => {
 
       // TODO: Unified Migration: [M3-7115] - Replace with copy from API '../notifications.py'
```
bnussman-akamai, line 39: Not related to this, but I noticed we kind of leaked some API code info here. Not a huge deal because it's super basic, but I think we generally try to avoid this.

jdamore-linode: This is a good callout, thanks @bnussman-akamai -- I read these comments but didn't think twice about the code reference there. I know there's another reference to this code elsewhere in this spec, and it might be worthwhile doing a quick search to find any other API code references that might be lurking around.

dwiley-akamai (approved):
LGTM 🚢

PR #9976, opened 2023-12-07
upcoming: [M3-7423] - Create Bucket - Able to choose a single Compute Region ID eg ('us-east')

cpathipa:
## Description 📝
OBJ Multi-Cluster - Create Bucket  flow - Able to choose a single Compute Region ID eg ('us-east') instead of cluster ID. 

| Before  | After   |
| ------- | ------- |
| image|image|

### Prerequisites
- Turn on the OBJ Multi Cluster feature flag
- Enable MSW
- Navigate to http://localhost:3000/object-storage/buckets/create.

### Verification steps 
(How to verify changes)
- Verify there is no regression in existing functionality (Create Bucket flow with feature flag disabled and disable MSW).
- Verify the new GUI as per the mockup.
- Verify Create Bucket payload and response wrt to API spec.

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/features/ObjectStorage/BucketLanding/BucketRegions.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface Props {
+  disabled?: boolean;
+  error?: string;
+  onBlur: (e: any) => void;
+  onChange: (value: string) => void;
+  required?: boolean;
+  selectedRegion: string;
+}
+
+export const BucketRegions: React.FC<Props> = (props) => {
```
dwiley-akamai, line 15:
```suggestion
export const BucketRegions = (props: Props) => {
```

packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawerObjMultiCluster.tsx
```diff
@@ -0,0 +1,188 @@
...
+  isOpen: boolean;
+  onClose: () => void;
+}
+
+export const CreateBucketDrawerObjMultiCluster = (props: Props) => {
+  const { data: profile } = useProfile();
+  const { isOpen, onClose } = props;
+  const isRestrictedUser = profile?.restricted;
+  const { data: regions } = useRegionsQuery();
+  // TODO: OBJ-MULTI-CLUSTER - clusters will likely to be replaced with regions and will be taken care in future tickets.
```
dwiley-akamai, line 41:
```suggestion
  // @TODO OBJ Multicluster - clusters will likely to be replaced with regions and will be taken care in future tickets.
```

packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawerObjMultiCluster.tsx
```diff
@@ -0,0 +1,188 @@
...
+    </Drawer>
+  );
+};
+
+const StyledEUAgreementCheckbox = styled(EUAgreementCheckbox, {
+  label: 'StyledEUAgreementCheckbox',
+})(({ theme }) => ({
+  marginButton: theme.spacing(3),
+  marginTop: theme.spacing(3),
+}));
```
dwiley-akamai, line 188: let's create a `CreateBucketDrawerObjMultiCluster.styles.ts` file per code conventions: https://github.com/linode/manager/blob/cc9020e96525ffa9bbe564ee8fc03d0bf0150705/docs/development-guide/02-component-structure.md?plain=1#L35

(could also consider renaming this file something like `OMC_CreateBucketDrawer.tsx` -- "OMC" for OBJ Multi-Cluster. Just to keep the file names shorter and we can use the OMC prefix whenever the temporary duplicate files need to be created)

dwiley-akamai (commented):
Feature flag off & MSW disabled: existing functionality works as expected ✅

Feature flag on & MSW enabled: payload includes `region` instead of `cluster` and submission goes through ✅

Formatting of region options matches wireframes ✅
(did you want to update the copy for the Create Bucket drawer as part of this PR too?)

packages/manager/src/utilities/formatRegion.ts
```diff
@@ -34,7 +34,7 @@ export const getSelectedRegion = (
   regions: Region[],
   selectedRegionId: string
 ): Region | undefined => {
-  return regions.find((thisRegion) => selectedRegionId === thisRegion.id);
+  return regions.find((thisRegion) => selectedRegionId === thisRegion?.id);
```
abailly-akamai, line 37: We don't want to loosen this, rather require a region at the component level

abailly-akamai (commented):
Looks good!

I tested the flow and did not find issues with payload and UI. Agreed with comments already added and left some other. 

Also, you have duplicated keys console errors when opening the drawer with MSW enabled. It's previous to the PR and not affecting production code, but can be a bit misleading when testing so it would be nice to fix

Screenshot 2023-12-15 at 10 10 27

packages/api-v4/src/object-storage/types.ts
```diff
@@ -36,6 +37,7 @@ export interface ObjectStorageDeleteBucketRequestPayload {
 }
 
 export interface ObjectStorageBucket {
+  region?: string;
```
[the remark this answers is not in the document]
cpathipa: I agree. This is to support backward compatibility until we fully roll out the feature. Once it is fully rolled out, we will replace 'region' with 'cluster.' (The 'Region' field won't be available in the response until API changes are deployed.)

abailly-akamai (approved):
Thanks for addressing all the changes.

- Code looks good ✅ 
- Confirmed seeing the right regions in the UI ✅ 
- Validation works as expected ✅

packages/validation/src/buckets.schema.ts
```diff
@@ -1,14 +1,24 @@
...
+    cluster: string().when('region', {
+      is: (region: string) => !region || region.length === 0,
+      then: string().required('Cluster is required.'),
+    }),
+    region: string().when('cluster', {
+      is: (cluster: string) => !cluster || cluster.length === 0,
+      then: string().required('Region is required.'),
+    }),
+  },
+  [['cluster', 'region']]
```
mjac0bs, line 20: One question about required fields and form validation: didn't we, at one point in time, add the "(required)" field to the Region selection since the cluster/region is also required to enable the button? I definitely remember this being a discussion that came up during DC pricing, and I thought we implemented it, but there's no required label on Region now, even in prod. Was thinking of this as I looked at the validation for these fields, which I don't think we'll ever see with the disabled button.

mjac0bs (approved):
Thanks Chandra, this is looking good! Left one question related to required field and validation, slightly tangential to this PR.

- Test are passing.
- Create Bucket region dropdown now shows regions instead of clusters.
- For regions with DC specific pricing, the overage costs are still correctly calculated.
- Existing Create Bucket functionality remains unchanged when feature flag is off. 
- Mocked POST buckets submits and now includes the `region` rather than cluster. 

Confirmed new POST:
Screenshot 2023-12-21 at 10 02 42 AM
Screenshot 2023-12-21 at 10 02 47 AM

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
jdamore-linode, line 1: The prefix on this filename (and the `.styles.ts` file) really makes it stick out -- is this a new convention we're adopting for project work going forward? If not, is there something else we can name these files that fits with our current file naming conventions?

jdamore-linode (approved):

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
cpathipa: @jdamore-linode OMC_CreateBucketDrawer is a temporary name. Once the feature is fully integrated into production, we will replace the existing CreateBucketDrawer with OMC_CreateBucketDrawer, and the prefix 'OMC' will be removed.

jdamore-linode: Thanks for the explanation @cpathipa! Sounds good to me 👍

packages/validation/src/buckets.schema.ts
cpathipa: @mjac0bs I see what you meant, to avoid confusion now it shows required label with disabled button. Addressed in the commit  7cbc121.
image

mjac0bs: @cpathipa Thank you! The RegionSelectAutocomplete PR a couple months back might have caused a regression preventing the required label from showing. It looks good now in the Create Bucket flow. It also now appears as expected in the Image Upload flow, where we can see it's currently missing in prod although the [RegionSelect component has the required prop](https://github.com/linode/manager/blob/ef6a9c628f1517f6b7dfbd362d76545f6ad98761/packages/manager/src/features/Images/ImageUpload.tsx#L279). Spot checked other instances of RegionSelect and Autocomplete and didn't see any unintended side effects. 

| Prod - missing required label | This Branch - has required label |
| --- | --- |
| Screenshot 2024-01-03 at 9 51 21 AM | Screenshot 2024-01-03 at 9 51 13 AM |

PR #9987, opened 2023-12-11
feat: [M3-7526] - Add setHeader to GET /profile endpoint

tyler-akamai:
## Description 📝
In order for the proxy account to access the username of the parent account, we need to use the auth token of the parent to collect the parent's profile info on behalf of the proxy account. We will have the parent account's token available to us for this use after storing it in local storage when the parent logs in as the proxy user via the "Switch Account" button.

## Changes  🔄
- Added setHeaders to the /profile endpoint
- Added an empty object to all useProfile instances

## How to test 🧪
- Add the code from the commit named [use this to test PR](https://github.com/linode/manager/pull/9987/commits/33ccb670d37f85f3e2afa6b331f39c56f04396aa) - make sure to remove the `undefined` param from the `useProfile` call
- Make sure to add a PAT from another account to your .env file (this could be your personal account or a test account)
- Ensure both usernames are logged similar to the screenshot below
Screenshot 2023-12-13 at 5 35 25 PM

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/api-v4/src/account/account.ts
+++ b/packages/api-v4/src/account/account.ts
@@ -31,7 +31,7 @@ import type { Token } from '../profile';
  * including contact and billing info.
  *
  */
-export const getAccountInfo = ({ headers }: RequestOptions) => {
+export const getAccountInfo = ({ headers }: RequestOptions = {}) => {
   return Request<Account>(
     setURL(`${API_ROOT}/account`),
     setMethod('GET'),
--- a/packages/api-v4/src/account/account.ts
+++ b/packages/api-v4/src/account/account.ts
@@ -31,8 +31,12 @@ import type { Token } from '../profile';
  * including contact and billing info.
  *
  */
-export const getAccountInfo = () => {
-  return Request<Account>(setURL(`${API_ROOT}/account`), setMethod('GET'));
+export const getAccountInfo = ({ headers }: RequestOptions) => {
+  return Request<Account>(
+    setURL(`${API_ROOT}/account`),
+    setMethod('GET'),
+    setHeaders(headers)
+  );
 };
 
 /**
--- a/packages/api-v4/src/account/account.ts
+++ b/packages/api-v4/src/account/account.ts
@@ -31,12 +31,8 @@ import type { Token } from '../profile';
  * including contact and billing info.
  *
  */
-export const getAccountInfo = ({ headers }: RequestOptions = {}) => {
-  return Request<Account>(
-    setURL(`${API_ROOT}/account`),
-    setMethod('GET'),
-    setHeaders(headers)
-  );
+export const getAccountInfo = () => {
+  return Request<Account>(setURL(`${API_ROOT}/account`), setMethod('GET'));
 };
 
 /**
--- a/packages/manager/src/containers/account.container.ts
+++ b/packages/manager/src/containers/account.container.ts
@@ -13,7 +13,7 @@ export const withAccount = <Props>(
   Component: React.ComponentType<Props & WithAccountProps>
 ) => {
   return (props: Props) => {
-    const account = useAccount();
+    const account = useAccount({});
 
     return React.createElement(Component, {
       ...props,
--- a/packages/manager/src/containers/account.container.ts
+++ b/packages/manager/src/containers/account.container.ts
@@ -13,7 +13,7 @@ export const withAccount = <Props>(
   Component: React.ComponentType<Props & WithAccountProps>
 ) => {
   return (props: Props) => {
-    const account = useAccount({});
+    const account = useAccount();
 
     return React.createElement(Component, {
       ...props,
--- a/packages/manager/src/queries/account.ts
+++ b/packages/manager/src/queries/account.ts
@@ -23,14 +23,24 @@ import type {
 
 export const queryKey = 'account';
 
-export const useAccount = () => {
+export const useAccount = ({ headers }: RequestOptions) => {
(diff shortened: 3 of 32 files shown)


packages/manager/src/queries/account.ts
```diff
@@ -23,14 +23,19 @@ import type {
...
-    ...queryPresets.noRetry,
-    enabled: !profile?.restricted,
-  });
+  return useQuery<Account, APIError[]>(
+    queryKey,
+    () => getAccountInfo({ headers }),
+    {
+      ...queryPresets.oneTimeFetch,
+      ...queryPresets.noRetry,
+      enabled: !profile?.restricted || hasExplicitAuthToken,
```
tyler-akamai, line 36: If an auth token is passed in (ex. from a proxy account, which be either restricted or unrestricted) enable the query.

packages/api-v4/src/account/account.ts
```diff
@@ -31,8 +31,12 @@ import type { Token } from '../profile';
  * including contact and billing info.
  *
  */
-export const getAccountInfo = () => {
-  return Request<Account>(setURL(`${API_ROOT}/account`), setMethod('GET'));
+export const getAccountInfo = ({ headers }: RequestOptions) => {
```
tyler-akamai, line 34: This is where I added the headers param.

Even though we are taking headers, i decided to still use RequestOptions for consistency across other queries

packages/manager/src/containers/account.container.ts
```diff
@@ -13,7 +13,7 @@ export const withAccount = <Props>(
   Component: React.ComponentType<Props & WithAccountProps>
 ) => {
   return (props: Props) => {
-    const account = useAccount();
+    const account = useAccount({});
```
tyler-akamai, line 16: Then I had to go through and pass in an empty obj since it now accepts RequestOptions

mjac0bs:
Cypress CI failures look relevant here and can be reproduced with a local run:

Screenshot 2023-12-12 at 9 12 10 AM

packages/api-v4/src/account/account.ts
jdamore-linode: Is there any possibility of making the headers param optional (and defaulting to `{}`)?

Outside of Manager and Parent/Child, I would guess that explicitly defined headers would not be needed most of the time this API is used, and requiring this to be passed is a breaking change (the Cypress tests are one example).

cc @mjac0bs in case you have any thoughts or additional insight

jdamore-linode:
Just a heads up that the E2E test failures all seem to stem from HTTP 500 responses from the API during resource clean up and aren't caused by these changes

packages/manager/src/queries/profile.ts
```diff
@@ -37,13 +37,20 @@ import { Grants } from '../../../api-v4/lib';
...
 
+import type { RequestOptions } from '@linode/api-v4';
+
 export const queryKey = 'profile';
 
-export const useProfile = (givenProfile?: Profile) =>
-  useQuery<Profile, APIError[]>(queryKey, getProfile, {
+export const useProfile = (
+  givenProfile?: Profile,
+  { headers }: RequestOptions = {}
```
mjac0bs, line 46:
```suggestion
  { headers }: RequestOptions = {}
   givenProfile?: Profile,
```
Prevents us from having to pass in `undefined` when calling `useProfile` with headers (e.g. [here](https://github.com/linode/manager/pull/9987/commits/33ccb670d37f85f3e2afa6b331f39c56f04396aa#diff-bbc007980e447b77254be68318934de3f8109be20948e1aa9f9321a534a517d0R65))

mjac0bs (approved):
Nice, I confirmed that I see another request to `/profile` on behalf of the account we're passing a token in for, and their username is available:

Screenshot 2023-12-13 at 3 12 11 PM

and there's our RQ query key:
Screenshot 2023-12-13 at 3 19 12 PM

Thanks for your work on this! If you get a chance to update the internal ticket based on our discussion, that would be great.

Edit: The tests that failed in CI are `billing/smoke-billing-activity.spec.ts` (passes locally) and `account/sms-verification.spec.ts`, which seems to be failing consistently locally when looking for the phone number to display after it's verified, and it doesn't. This test passed in `develop`, so not quite sure what's happening there.
Screenshot 2023-12-13 at 7 47 18 PM

packages/manager/src/queries/profile.ts
```diff
@@ -37,13 +37,20 @@ import { Grants } from '../../../api-v4/lib';
...
 import { queryPresets } from './base';
 
+import type { RequestOptions } from '@linode/api-v4';
+
 export const queryKey = 'profile';
 
-export const useProfile = (givenProfile?: Profile) =>
-  useQuery<Profile, APIError[]>(queryKey, getProfile, {
+export const useProfile = (
+  givenProfile?: Profile,
```
bnussman-akamai, line 45: While we're here. I think we can get rid of this `givenProfile`. It's not used

```suggestion
```

jaalah-akamai:
@tyler-akamai You may need to update this branch to avoid all the file noise we're seeing in `Files Changed`

jaalah-akamai:
This looks good, while you're at it, can update the defaults for these too? https://github.com/linode/manager/blob/develop/packages/api-v4/src/account/account.ts#L162 `headers = {}`

jaalah-akamai (approved):
This looks good 👍

PR #9973, opened 2023-12-06
ci: [M3-7550] - Add Lint To Github Actions 

bnussman-akamai:
## Description 📝

- Adds a Lint step in Github Actions for Cloud Manager's, api-v4, and validation ✅
  - This honestly should have been done years ago
  - For those of us who don't believe in pre-commit hooks, we can now commit without worrying about merging bad code 🎉 
  - Another added benefit is that we'll more easily know if something about our eslint configuration breaks, which thankfully hasn't happened, but could 

## How to test 🧪

- View Github Actions below 👀⬇️
  - You should see 3 new lint checks
 
## As an Author I have considered 🤔

- [x] 👀 Doing a self review
- [x] ❔ Our [contribution guidelines](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md)
- [x] 🤏 Splitting feature into small PRs
- [x] ➕ Adding a changeset
- [x] 🧪 Providing/Improving test coverage
- [x] 🔐 Removing all sensitive information from the code and PR description
- [x] 🚩 Using a feature flag to protect the release
- [x] 👣 Providing comprehensive reproduction steps
- [x] 📑 Providing or updating our documentation
- [x] 🕛 Scheduling a pair reviewing session
- [x] 📱 Providing mobile support
- [x] ♿  Providing accessibility support

PR #9988, opened 2023-12-11
ci: Speed Up Code Coverage Github Actions Jobs

bnussman-akamai:
## Description 📝

Slightly speeds up our new code coverage Github Actions jobs by not building a production Cloud Manager bundle 🏎️

We should see an improvement of about 30-60 seconds ⏲️ Nothing groundbreaking, but slightly better. 

## How to test 🧪

- Check the GitHub Actions below this PR and verify that the Code Coverage jobs work as expected

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -25,8 +25,11 @@ jobs:
       - name: Install Dependencies
         run: yarn --frozen-lockfile
 
-      - name: Run build
-        run: yarn build
+      - name: Build @linode/validation
+        run: yarn build:validation
+
+      - name: Build @linode/api-v4
+        run: yarn build:sdk
 
       - name: Run Base Branch Coverage
         run: yarn coverage:summary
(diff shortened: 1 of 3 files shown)


abailly-akamai (approved):
This works! small improvement, wondering what prompted you to want to improve the speed here. This action is non blocking and we usually want to wait for the e2e suite to finish before merging anyway, which takes far more time.

If we really want to improve speed here, we could upload the base coverage pct to another bucket on every merge to `develop` and download it on each run so we don't have to run the unit suite on the ref branch. Would cut the job time by almost half but unless I am missing a particular use case I dunno if worth the effort?

bnussman-akamai:
No significant reason, I just noticed we could skip the Cloud Manager build when I did https://github.com/linode/manager/pull/9973

> If we really want to improve speed here, we could upload the base coverage pct to another bucket on every merge to develop and download it on each run so we don't have to run the unit suite on the ref branch. Would cut the job time by almost half but unless I am missing a particular use case I dunno if worth the effort?

I thought about this, but I appreciate the simplicity of how it is now and the fact that it can dynamically get the coverage of the base_ref

mjac0bs (approved):
Confirmed the action works to produce the coverage report and we see a small increase in speed:

Action run on this branch:
Screenshot 2023-12-11 at 9 45 18 PM

Action run on another PR branch:
Screenshot 2023-12-11 at 9 46 34 PM

PR #9994, opened 2023-12-12
refactor: [M3-7562] - Remove Radio.mdx and Textfield.mdx files

coliu-akamai:
## Description 📝
- Remove Radio.mdx and TextField.mdx files + moved documentation comments in those files to relevant component files
- Add tests 

### Verification steps 
- http://localhost:6006/?path=/docs/components-radio--documentation
- http://localhost:6006/?path=/docs/components-textfield--documentation
- yarn test Radio.test.tsx
- yarn test TextField.test.tsx

*Check all that apply*

- [x] 👀 Doing a self review
- [ ] ➕ Adding a changeset


packages/manager/src/components/TextField.tsx
```diff
@@ -176,6 +176,51 @@ export type TextFieldProps = BaseProps &
...
+- Placeholder text is the text that users see before they interact with a field. It should be a useful guide to the input type and format <br/> Don’t make the user guess what format they should use for the field. Tell this information up front.
+
+### Best Practices
+
+- A single column form with input fields stacked sequentially is the easiest to understand and leads to the highest success rate. Input fields in multiple columns can be overlooked or add unnecessary visual clutter.
+- Grouping related inputs (ex. mailing address) under a subhead or rule can add meaning and make the form feel more manageable.
+- Avoid breaking a single form into multiple “papers” unless those sections are truly independent of each other.
+- Consider sizing the input field to the data being entered (ex. the field for a street address should be wider than the field for a zip code). Balance this goal with the visual benefits of fields of the same length. A somewhat outsized input that aligns with the fields above and below it might be the best choice.
+
+## Textfield errors
```
coliu-akamai, line 204: This documentation section corresponds to the error textfield story - had to add it here instead of above the error textfield story to get it to show up in the storybook documentation section. We removed showing all stories in the documentation (see [here](https://github.com/linode/manager/pull/9963#discussion_r1417645755)). 


| Before  | Current (see comment above for changes)   |
| ------- | ------- |
| image | image |

packages/manager/src/components/TextField.tsx
```diff
@@ -73,6 +73,10 @@ interface BaseProps {
    * className to apply to the underlying TextField component
    */
   className?: string;
+  /**
```
coliu-akamai, line 76: The textfield component is currently in the root of the `components` package - thoughts on moving it into a separate TextField folder within `components`? (this will lead to a way bigger file diff - like +80 files due to all the import changes though 😅 )

carrillo-erik: Thanks for considering this. In the past we had `components/core/` and we migrated away from that to just having `components/`.  I don't seem to recall if we have guidelines to determine how things should be organized. At the moment it appears to be a mixture of individual folders and files. I would prefer that each component is contained within a designated folder containing tests and storybook files.

This might be good to discuss during the Cafe.

carrillo-erik (approved):

coliu-akamai:
thanks @carrillo-erik!

packages/manager/src/components/TextField.tsx
abailly-akamai: I think this could be done separately if we wanted to. Considering the large January release in front of us and the PR to wrap up MUI migrations I would leave it alone for now

abailly-akamai (approved):
Looks good! Thanks @coliu-akamai 

Only thing I noticed is the Radio `size` prop not doing anything. Approving tho we may wanna look into that. 

Also, don't hesitate to get a little deeper with testing props sometimes. Especially the ones that are going to change the appearance of the component. This is usually the kind of visual regression we miss when doing a upgrade or migration.

abailly-akamai:
@coliu-akamai also restarting your e2e suite since once of the run failed to complete the job

coliu-akamai:
@abailly-akamai looking into the size issue, seems like size as a prop just doesn't have any effect on the radio appearance at all, even though the svg fontsize changes 🤔 gonna look into this some more!

image
image

coliu-akamai:
update: Alban and I investigated, and the issue is due to `MuiSvgIcon` style overrides at the theme layer (line 1040 of `light.ts`). Opened up a new ticket (M3-7565) to further describe the issue and lay out some potential ways to handle this

PR #9996, opened 2023-12-12
feat: [M3-7535] – Update OBJ types to support Multicluster changes

dwiley-akamai:
## Description 📝
Update OBJ types to support upcoming changes for Multicluster as per API spec (see ticket for more details).

## Changes  🔄
- `regions` property added to `ObjectStorageKey`
- refactor: new `ScopeRequest` and `ScopeObject` that extend `Scope`
- `regions` property added to `ObjectStorageKeyRequest` and `UpdateObjectStorageKeyRequest`
  - `label` made optional for `UpdateObjectStorageKeyRequest`
- `region` added to `ObjectStorageBucketRequestPayload`
  - `cluster` made optional
- `region` added to `ObjectStorageBucket`

## How to test 🧪
The main things are confirming the changes match what is prescribed by the API spec and that no functionality has been adversely impacted.

### Verification steps 
- The following are populated without app errors/crashing:
    - OBJ landing page (Buckets and Access Keys tabs)
    - Bucket Detail drawer
    - Edit Access Key drawer
- Able to create Buckets
- Able to create Access Keys

- [X] 👀 Doing a self review
- [X] ➕ Adding a changeset


packages/api-v4/src/object-storage/types.ts
```diff
@@ -1,33 +1,55 @@
...
 }
 
+export interface ScopeRequest extends Scope {
+  cluster?: string;
+  region?: string;
+}
+
+export interface ScopeObject extends Scope {
+  cluster: string;
+  region?: string; // @TODO OBJ Multicluster: Remove optional indicator when API changes get released to prod
```
dwiley-akamai, line 29: The API spec states that both `cluster` and `region` will always be returned in objects in the `bucket_access` field. For TypeScript purposes in the interim, I have this as an optional property.

tyler-akamai:
Verified:

The following are populated without app errors/crashing:
   - [ ] OBJ landing page (Buckets and Access Keys tabs)
   - [ ] Bucket Detail drawer
- [ ] Edit Access Key drawer
- [ ] Able to create Buckets
- [ ] Able to create Access Keys

packages/api-v4/src/object-storage/types.ts
```diff
@@ -36,12 +58,13 @@ export interface ObjectStorageDeleteBucketRequestPayload {
...
 export interface ObjectStorageBucket {
-  label: string;
-  created: string;
   cluster: string;
+  created: string;
   hostname: string;
-  size: number; // Size of bucket in bytes
+  label: string;
   objects: number;
+  region: string;
```
cpathipa, line 66: I think we should make this optional for now. The `regions` field will not be available in the API response until API changes are deployed. Otherwise, this may lead to type errors. Same with `label` and `created`

dwiley-akamai: Would we get type errors if `region` isn't used in the code or is safe accessed/optional chained?

`label` and `created` are existing fields, what would cause type errors for those?

packages/api-v4/src/object-storage/types.ts
hana-linode: Do we need to create a new ScopeObject interface? What about doing something like

```typescript
export interface Scope {
  bucket_name: string;
  permissions: AccessType;
  cluster: string;
  region?: string; // @TODO OBJ Multicluster: Remove optional indicator when API changes get released to prod
}

export interface ScopeRequest extends Omit<Scope, 'cluster'> {
  cluster?: string;
  region?: string;
}
```

packages/api-v4/src/object-storage/types.ts
```diff
@@ -1,33 +1,55 @@
+export interface RegionS3EndpointAndID {
```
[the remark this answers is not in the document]
dwiley-akamai: In this case I was thinking it'd be better to be more explicit about what is included in the object since what it includes didn't strike me as very intuitive at first glance. Can definitely change it if there's consensus for that though

hana-linode (approved):
Approving pending merge conflicts are addressed

- The following are populated without app errors/crashing:
    - OBJ landing page (Buckets and Access Keys tabs) ✅ 
    - Bucket Detail drawer ✅ 
    - Edit Access Key drawer ✅ 
- Able to create Buckets ✅ 
- Able to create Access Keys ✅

packages/api-v4/src/object-storage/types.ts
cpathipa: think I encountered some type errors a couple of weeks back when I pulled this into my working branch. However, I'm not seeing them anymore.

cpathipa (approved):
Changes LGTM!

- [x] Verified type changes according to the API spec.
- [x] Able to create Bucket.
- [x] Able to create Accesskey.

PR #10005, opened 2023-12-15
change: [M3-7454] - Add child access user permissions for parent accounts

tyler-akamai:
## Description 📝
As an admin on a parent account, when I create new non-admin users, I need the ability to grant full account access to child accounts. To facilitate this, a new permissions toggle labeled "Enable child account access" will be introduced during the creation of non-admin parent account users. Additionally, the interface will be enhanced by incorporating a "General permissions" headline, and the "Full account access" option will be relocated within the Paper section for improved clarity.

Also, I named the ticket category `change` because although the 'Enable Child Access' toggle button is a future feature, there are a couple UI changes that will take place as soon as this ticket is merged.

## Changes  🔄
- New "Enable child account access" toggle during the creation of non-admin parent account users.
- Introduce a new headline titled "General permissions" to provide clarity and organization.
- Relocate the "Full account access" option within the Paper section for improved visibility and alignment with the overall interface design.
- Remove the borders between each permission toggle
- Invalidate query when permissions change
- Improved mobile view

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screenshot 2023-12-19 at 1 52 10 PM | Screenshot 2023-12-19 at 1 47 06 PM |
| Screenshot 2023-12-19 at 1 52 19 PM | Screenshot 2023-12-19 at 1 49 30 PM |

### Prerequisites
- Use the Mock Service Worker
- Ensure Parent/Child Account Feature Flag is on

### Verification steps 
(How to verify changes)
- Navigate to http://localhost:3000/account/users/User/permissions
- Ensure the toggle works and displays the permissions
- Ensure there is an 'Enable child account access' toggle
- Ensure there is _no_ toggle for http://localhost:3000/account/users/ChildUser/permissions and http://localhost:3000/account/users/ParentCompany_a1b2c3d4e5/permissions because they are child and proxy accounts
- Ensure there is no 'Enable child account access' toggle button when the feature flag is off

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/features/Users/UserDetail.tsx
```diff
@@ -252,6 +252,7 @@ export const UserDetail = () => {
             <UserPermissions
               clearNewUser={clearNewUser}
               currentUser={profile?.username}
+              queryClient={queryClient}
```
tyler-akamai, line 255: Since UserPermissions is a class component, we can't call hooks within it and we have to pass in the query client in order to invalidate the data.

packages/manager/src/features/Users/UserPermissions.tsx
```diff
@@ -138,6 +149,33 @@ class UserPermissions extends React.Component<CombinedProps, State> {
     }
   };
 
+  checkAndEnableChildAccountAccess = async () => {
```
tyler-akamai, line 152: This function is used to enable and disable the Child account access toggle:

Enables when:
1. Current user is a parent
2. Selected user is _not_ a child account
3. Selected user is _not_ a proxy account

packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -508,6 +509,12 @@ const childAccountUser = accountUserFactory.build({
   user_type: 'child',
   username: 'ChildUser',
 });
+const accountUser = accountUserFactory.build({
```
tyler-akamai, line 512: This user represents a user within a parent account.

packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -1216,10 +1224,20 @@ export const handlers = [
...
     return res(ctx.json(parentAccountUser));
   }),
+  rest.get(`*/account/users/${accountUser.username}`, (req, res, ctx) => {
+    return res(ctx.json(accountUser));
+  }),
   rest.get('*/account/users/:user', (req, res, ctx) => {
     // Parent/Child: switch the `user_type` depending on what account view you need to mock.
     return res(ctx.json(accountUserFactory.build({ user_type: 'parent' })));
   }),
+  rest.put(`*/account/users/${accountUser.username}`, (req, res, ctx) => {
```
tyler-akamai, line 1234: I had to add in some mock requests in order for the permission toggle to work

packages/manager/src/features/Users/UserPermissions.tsx
```diff
@@ -456,25 +498,27 @@ class UserPermissions extends React.Component<CombinedProps, State> {
...
-        <Divider />
       </Grid>
     );
   };
 
   renderGlobalPerms = () => {
     const { grants, isSavingGlobal } = this.state;
+    if (
+      this.state.childAccountAccessEnabled &&
+      !this.globalBooleanPerms.includes('child_account_access')
```
tyler-akamai, line 509: I had to add this line:
`!this.globalBooleanPerms.includes('child_account_access'`
strictly for the MSW. Because for some reason the MSW keeps adding additional toggle buttons whenever we click toggle. This however does not happen when not using MSW. So we can add a note to either remove this line when the API is ready or look more into why these duplicates are occurring. Either way, this shouldnt be a blocker.

packages/manager/src/features/Users/UserPermissions.tsx
```diff
@@ -4,6 +4,7 @@ import {
   GrantType,
```
tyler-akamai, line 4: I know that there are talks about refactoring the LinodeCreate flow, I think this should also be refactored soon. Its a class component with almost 800 lines of code and its something that we will always have to come back to.

packages/manager/src/mocks/serverHandlers.ts
mjac0bs: We've got so many accounts users to keep track of; does naming this something like "parentAccountNonAdminUser" make sense?

packages/manager/src/features/Users/UserPermissions.tsx
```diff
@@ -138,6 +149,33 @@ class UserPermissions extends React.Component<CombinedProps, State> {
...
+    const selectedUsername = this.props.username;
+    if (currentUsername && selectedUsername) {
+      try {
+        const currentUser = await getUser(currentUsername);
+        const selectedUser = await getUser(selectedUsername);
+
+        const isParentAccount = currentUser.user_type === 'parent';
+        const isSelectedAccountValid =
+          selectedUser.user_type !== 'child' &&
+          selectedUser.user_type !== 'proxy';
```
mjac0bs, line 163: In thinking about this more: do we need this check for a selected user?

- A parent account isn't going to have a child or a proxy user on their account in reality -- I know our mocks do to prevent us from having to add these different users depending on user_type of the current account. (We can change the current user type on [L1232](https://github.com/linode/manager/pull/10005/files#diff-c4904f695ece84849bc2372518313ff64475b5598ab35b45708266fa74e2e6e8R1232) of the mocks.)
- A child, proxy, or regular user who is the current user already wouldn't see this toggle because `isParentAccount` will be false. 
- The only other type of user would be an admin parent user, and admin users on accounts don't see these perms because they already have full account access.

Have I missed any edge cases?

mjac0bs (commented):
Thanks Tyler!

I was able to confirm the UI enhancements and I see the new toggle displaying for a parent account's user and it is toggleable. I've left some feedback in line.

packages/manager/src/mocks/serverHandlers.ts
jaalah-akamai: Yes more specific the better for this project

packages/manager/src/features/Users/UserPermissions.tsx
```diff
@@ -607,23 +651,7 @@ class UserPermissions extends React.Component<CombinedProps, State> {
...
-    return (
-      <Paper
-        sx={(theme) => ({
-          marginTop: theme.spacing(2),
-          padding: theme.spacing(3),
-        })}
-      >
-        <Typography data-qa-unrestricted-msg>
-          This user has unrestricted access to the account.
-        </Typography>
```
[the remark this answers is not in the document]
jaalah-akamai: Yea we want the save button because this is kind of related to `M3-7458` where all perms get set to `none` if a user toggles that switch (which is an unintended).

packages/manager/src/features/Users/UserPermissions.tsx
jaalah-akamai: I was going to say the same thing, in addition this wouldn't cover the scenario where the `user_type` is `null` for non-parent/child accounts. So I think we're fine with removing that check.

jaalah-akamai: Technically there could be two admin parent users and they would be able to see the toggle for each other.

packages/manager/src/features/Users/UserPermissions.tsx
```diff
@@ -138,6 +149,33 @@ class UserPermissions extends React.Component<CombinedProps, State> {
...
+        const isSelectedAccountValid =
+          selectedUser.user_type !== 'child' &&
+          selectedUser.user_type !== 'proxy';
+        const isFeatureFlagOn = this.props.flags.parentChildAccountAccess;
+
+        if (isParentAccount && isSelectedAccountValid && isFeatureFlagOn) {
+          this.setState({ childAccountAccessEnabled: true });
+        } else {
+          this.setState({ childAccountAccessEnabled: false });
+        }
```
jaalah-akamai, line 170: We can simplify this to
```
this.setState({
   childAccountAccessEnabled: isParentAccount && isFeatureFlagOn,
});
```

packages/manager/src/features/Users/UserPermissions.tsx
```diff
@@ -138,6 +149,33 @@ class UserPermissions extends React.Component<CombinedProps, State> {
...
+    if (currentUsername && selectedUsername) {
+      try {
+        const currentUser = await getUser(currentUsername);
+        const selectedUser = await getUser(selectedUsername);
+
+        const isParentAccount = currentUser.user_type === 'parent';
+        const isSelectedAccountValid =
+          selectedUser.user_type !== 'child' &&
+          selectedUser.user_type !== 'proxy';
+        const isFeatureFlagOn = this.props.flags.parentChildAccountAccess;
```
jaalah-akamai, line 164: Nit: We can apply some destructuring here if we wanted
```
const { currentUser, flags } = this.props;
```

jaalah-akamai (commented):
Nice work overall - This will be 😗 👌 after some minor cleanup 🚀

packages/manager/src/features/Users/UserPermissions.tsx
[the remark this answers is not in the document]
tyler-akamai: Good catch, 

The only issue I see with adding a 'Save' Button here is that instead of relying on the API to populate the permissions, we have to add logic on the frontend side to initialize temporary Global and Specific Grants (temporary because we have to wait for the user to click 'Save') and also go through each Specific Grant and add the account instances.

For example, in the image below, we would have to loop through all of the Linode instances to add it to the Linode Grant section, same with the Firewalls, NodeBalancers, Domains, Images, Volumes, etc.

Screenshot 2023-12-22 at 12 03 07 PM

In comparison to how we do it now, all of this information is returned in one `grants` api call. This is because the UI state matches the API state.

One potential option if we can still utilize the API. Meaning from the viewpoint of the API, every change the user does in the UI, is actually sending requests to the UI. But if the user does not click 'Save' and navigates to another page, we just send one additional request which returns the account back to the previous state. 

Couple issues with this though is that:
1. Its not intuitive, and tbh I dont think its great design (bigger issue)
2. Its still going to require a decent bit of additional logic to store the previous state, monitor history and send additional requests (smaller issue)

I am interested in hearing your thoughts @mjac0bs @jaalah-akamai.

packages/manager/src/features/Users/UserPermissions.tsx
tyler-akamai: Yeah that all makes sense

packages/manager/src/features/Users/UserPermissions.tsx
```diff
@@ -138,6 +149,33 @@ class UserPermissions extends React.Component<CombinedProps, State> {
...
+
+        if (isParentAccount && isSelectedAccountValid && isFeatureFlagOn) {
+          this.setState({ childAccountAccessEnabled: true });
+        } else {
+          this.setState({ childAccountAccessEnabled: false });
+        }
+      } catch (error) {
+        this.setState((prevState) => ({
+          errors: [...(prevState.errors || []), error],
```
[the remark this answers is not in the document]
tyler-akamai: Good catch, just fixed:
Screenshot 2023-12-22 at 2 28 50 PM

packages/manager/src/features/Users/UserPermissions.tsx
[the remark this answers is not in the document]
jaalah-akamai: We talked about the different approaches here offline and will move forward with tackling this in `M3-7458`.

mjac0bs (approved):
✅  I only see the "Enable child account access" grant when viewing the perms of a restricted parent user. 
✅ Enable child account access perm is hidden when feature flag is off.
✅ UI improvements look good - "general" permissions, no divider, layout.
✅ Mobile view looks good.

#10009, which we merged yesterday, adds some test coverage for User Permissions flows that we didn't have previously. The one thing left to do here is merge in develop and fix the e2es that will fail due to some of the UI improvements in this PR. Approving pending those updates are made. Nice work, Tyler!

jaalah-akamai:
Thanks for quick updates - tested this via instructions and this is looking good

jaalah-akamai (approved):

PR #10010, opened 2023-12-19
feat: [M3-7168] - Increment Subnet IPv4 value when recommending new subnet range

coliu-akamai:
## Description 📝
When creating new subnets via both the Create VPC flow and the Create Subnet drawer, we will now be incrementing the subnet range so that the default value is not a static 10.0.4.0/24 value.

Talked with Daniel about this behavior 👍 - 
- For both the VPC Create flows and Subnet Create Drawer, the recommended IP will start at 10.0.0.0/24 or 10.0.1.0/24 and go up to 10.0.1.0/24, 10.0.2.0/24, etc. 10.0.x.0/24 ips will be the only ones recommended. 
  - (There was a previous version that depended more on the user's already existing IPs for the Subnet Create Drawer >> checkout and run [this commit](https://github.com/linode/manager/pull/10010/commits/207c2d4d1efe99d474d0b324877917bfe414af59) - this led to pretty frequent recommendation of "bad IPs" (ips that are technically valid but have overlapping ranges with existing ip addresses, leading to the backend throwing an error), so I switched it to current version 😢)
  - In terms of enough subnets being recommended: since a person can normally have at most 10 subnets to a VPC (special exceptions can occur), for the most part, we should not run out of subnets to recommend 
  - note that due to user input, there's still a chance of a 'bad ip' being recommended (ex. if I have an IP of 10.0.0.0/8, then any 10.0.x.0/24 recommendation is already covered by 10.0.0.0/8...)
- If we want a version where there's more robust ip-recommending, it's going to be a lot more complicated. Since these are just ip recommendations, not what people actually have to choose, this version is a bit better than the previous behavior of just recommending 10.0.4.0/24 though!

## Changes  🔄
- Created new helper function to recommend an IPv4 for a subnet (ty to @bnussman for the inspiration in AGLB create flow!) + add relevant tests
- update SubnetCreateDrawer and VPCCreate flows
- added a getAllSubnets request to get around pagination

## Preview 📷

| VPC Create Flow  | Subnet Create Drawer   |
| ------- | ------- |
| <video src="https://github.com/linode/manager/assets/139280159/e44b49e9-4985-40e3-bc85-1b57a524c0bf" /> | <video src="https://github.com/linode/manager/assets/139280159/354d2df1-8afb-4d94-b9a2-e91ef1050bfa" /> |

### Prerequisites
NOTE - there is a separate bug regarding SubnetCreateDrawer erroring when we try to create more than 10 subnets - see internal ticket for details

### Verification steps 
(How to verify changes)
- yarn test subnets.test.ts
- Verify that when creating subnets on VPC Create flows (drawer and page) and Subnet Create Drawer, a new subnet IP that is not in the already existing list of IPs is recommended (note - if you go past 256 subnets though, there will be duplicates)

*Check all that apply*

- [ ] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/VPCs/VPCDetail/SubnetCreateDrawer.tsx
+++ b/packages/manager/src/features/VPCs/VPCDetail/SubnetCreateDrawer.tsx
@@ -6,7 +6,7 @@ import { ActionsPanel } from 'src/components/ActionsPanel/ActionsPanel';
 import { Drawer } from 'src/components/Drawer';
 import { Notice } from 'src/components/Notice/Notice';
 import { useGrants, useProfile } from 'src/queries/profile';
-import { useAllSubnetsQuery, useCreateSubnetMutation } from 'src/queries/vpcs';
+import { useCreateSubnetMutation, useVPCQuery } from 'src/queries/vpcs';
 import { getErrorMap } from 'src/utilities/errorUtils';
 import {
   DEFAULT_SUBNET_IPV4_VALUE,
@@ -27,13 +27,13 @@ export const SubnetCreateDrawer = (props: Props) => {
 
   const { data: profile } = useProfile();
   const { data: grants } = useGrants();
-  const { data: allSubnets } = useAllSubnetsQuery(vpcId);
+  const { data: vpc } = useVPCQuery(vpcId, open);
 
   const userCannotAddSubnet = profile?.restricted && !grants?.global.add_vpcs;
 
   const recommendedIPv4 = getRecommendedSubnetIPv4(
     DEFAULT_SUBNET_IPV4_VALUE,
-    allSubnets?.map((subnet) => subnet.ipv4 ?? '') ?? []
+    vpc?.subnets?.map((subnet) => subnet.ipv4 ?? '') ?? []
   );
 
   const [errorMap, setErrorMap] = React.useState<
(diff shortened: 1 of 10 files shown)


packages/manager/src/features/VPCs/VPCDetail/SubnetCreateDrawer.tsx
```diff
@@ -26,9 +27,15 @@ export const SubnetCreateDrawer = (props: Props) => {
 
   const { data: profile } = useProfile();
   const { data: grants } = useGrants();
+  const { data: allSubnets } = useAllSubnetsQuery(vpcId);
```
coliu-akamai, line 30: rather than a getAll subnets of a VPC, we could directly get the VPC since subnets are part of the VPC's return object if that's a better approach! (we just wouldn't need any of the vpc's other information besides its subnets)

bnussman-akamai: Yes, I think using `useVPCQuery` to get all of the subnets would be much better because it would be one less things to fetch and work about invalidating. I pushed this change in 644f882aa7263eab8c265f9c6d6465ed30f8da66

hana-linode (approved):
Unit test passed and confirmed that the subnet range is increased when adding a new subnet in the Create VPC and Create Subnet flow

bnussman-akamai (approved):

jaalah-akamai (approved):

dwiley-akamai (approved):
Incrementation in VPC Create flow and Create Subnet drawer ✅
Code review ✅

PR #10011, opened 2023-12-19
change: [M3-7461] - Update toast notifications for UserPermissions

tyler-akamai:
## Description 📝
Updated toast notifications in user permissions.

## Changes  🔄
- Updated toast notifications

## Preview 📷
| | 
|---|
| Screenshot 2023-12-22 at 2 46 49 PM |
| Screenshot 2023-12-22 at 2 47 23 PM |
| Screenshot 2023-12-22 at 2 47 37 PM |

### Verification steps 
- Ensure that saving full account access, global (soon to be renamed "general"), and entity specific permissions triggers a toast notification. ('User permissions successfully saved.' or '[Entity-specific/general] user permissions successfully saved.')

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/.changeset/pr-10011-changed-1703028750129.md
+++ b/packages/manager/.changeset/pr-10011-changed-1703028750129.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Changed
+---
+
+Update toast notifications for UserPermissions ([#10011](https://github.com/linode/manager/pull/10011))
--- a/packages/manager/src/features/Users/UserPermissions.tsx
+++ b/packages/manager/src/features/Users/UserPermissions.tsx
@@ -268,6 +268,9 @@ class UserPermissions extends React.Component<CombinedProps, State> {
           this.getUserGrants();
           // refresh the data on /account/users so it is accurate
           this.props.queryClient.invalidateQueries('account-users');
+          this.props.enqueueSnackbar('User permissions successfully saved.', {
+            variant: 'success',
+          });
         })
         .catch((errResponse) => {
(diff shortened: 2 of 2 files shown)


mjac0bs (commented):
The other spot that applies here is the Full Account Access toggle. Can we make a toast appear when that permission is changed as well?

I'm wondering if UX is concerned at all about the loss of granularity in the toast message for the global-now-general and entity-specific permissions. (For example: could this change cause a user to think that all their permissions have been saved when they made changes to both general and entity-specific permissions, but only clicked the save button on entity-specific permissions?) I reached out to Jan from UX about this. If we want to preserve the granularity, a solution could be to have `General user permissions successfully saved.` and `Entity-specific user permissions successfully saved.`

Edit after reaching out to UX: When the mocks were created, UX was requesting a toast because the Full Account Access perm lacked one. For our existing General/Specific perm toasts, let's update them to what I've suggested above.

bnussman-akamai (approved):

tyler-akamai:
> The other spot that applies here is the Full Account Access toggle. Can we make a toast appear when that permission is changed as well?

Awesome, for the Full Account Access toggle, I just added a toast notification that says 'User permissions successfully saved.' Im open for suggestions

> For our existing General/Specific perm toasts, let's update them to what I've suggested above.
Done!

mjac0bs (approved):
Thank you for making those updates!

✅  Verified the correct toast messages appear when the three types of user permissions are changed: full account access, global, and specific.

PR #10024, opened 2024-01-02
feat: [M3-7529] – Support VPC in Personal Access Token drawer

dwiley-akamai:
## Description 📝
Coming soon

## Changes  🔄
Coming soon

## Preview 📷
Coming soon

### Prerequisites
- ...
- ...

### Reproduction steps
- ...
- ...

### Verification steps 
- ...
- ...

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/ObjectStorage/AccessKeyLanding/AccessCell.tsx
+++ b/packages/manager/src/features/ObjectStorage/AccessKeyLanding/AccessCell.tsx
@@ -4,6 +4,7 @@ import * as React from 'react';
 
 import Check from 'src/assets/icons/monitor-ok.svg';
 import { Radio } from 'src/components/Radio/Radio';
+import { Tooltip } from 'src/components/Tooltip';
 
 interface RadioButton extends HTMLInputElement {
   name: string;
@@ -15,11 +16,20 @@ interface AccessCellProps {
   onChange: (e: React.SyntheticEvent<RadioButton>) => void;
   scope: string;
   scopeDisplay: string;
+  tooltipText?: string;
   viewOnly: boolean;
 }
 
 export const AccessCell = React.memo((props: AccessCellProps) => {
-  const { active, disabled, onChange, scope, scopeDisplay, viewOnly } = props;
+  const {
+    active,
+    disabled,
+    onChange,
+    scope,
+    scopeDisplay,
+    tooltipText,
+    viewOnly,
+  } = props;
 
   if (viewOnly) {
     if (!active) {
@@ -36,7 +46,7 @@ export const AccessCell = React.memo((props: AccessCellProps) => {
     );
   }
 
-  return (
+  const radioBtn = (
     <Radio
       inputProps={{
         'aria-label': `${scope} for ${scopeDisplay}`,
@@ -48,6 +58,14 @@ export const AccessCell = React.memo((props: AccessCellProps) => {
       onChange={onChange}
       value={scope}
     />
+  );
+
+  return tooltipText ? (
+    <Tooltip placement="top" title={tooltipText}>
+      <span>{radioBtn}</span>
+    </Tooltip>
+  ) : (
+    radioBtn
   );
 });
 
--- a/packages/manager/src/features/Profile/APITokens/CreateAPITokenDrawer.tsx
+++ b/packages/manager/src/features/Profile/APITokens/CreateAPITokenDrawer.tsx
@@ -16,6 +16,7 @@ import { TableRow } from 'src/components/TableRow';
 import { TextField } from 'src/components/TextField';
 import { ISO_DATETIME_NO_TZ_FORMAT } from 'src/constants';
 import { AccessCell } from 'src/features/ObjectStorage/AccessKeyLanding/AccessCell';
+import { VPC_READ_ONLY_HELPER_TEXT } from 'src/features/VPCs/constants';
 import { useFlags } from 'src/hooks/useFlags';
 import { useAccountUser } from 'src/queries/accountUsers';
 import { useProfile } from 'src/queries/profile';
@@ -258,6 +259,9 @@ export const CreateAPITokenDrawer = (props: Props) => {
               if (!basePermNameMap[scopeTup[0]]) {
                 return null;
               }
+
+              const scopeIsForVPC = scopeTup[0] === 'vpc';
+
               return (
                 <TableRow
                   data-qa-row={basePermNameMap[scopeTup[0]]}
@@ -281,8 +285,11 @@ export const CreateAPITokenDrawer = (props: Props) => {
                     parentColumn="Read Only"
                   >
                     <AccessCell
+                      tooltipText={
+                        scopeIsForVPC ? VPC_READ_ONLY_HELPER_TEXT : undefined
+                      }
                       active={scopeTup[1] === 1}
-                      disabled={scopeTup[0] === 'vpc'} // "Read Only" is not a valid scope for VPC
+                      disabled={scopeIsForVPC} // "Read Only" is not a valid scope for VPC
                       onChange={handleScopeChange}
                       scope="1"
                       scopeDisplay={scopeTup[0]}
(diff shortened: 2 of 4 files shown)


packages/manager/src/features/ObjectStorage/AccessKeyLanding/AccessCell.tsx
```diff
@@ -49,6 +59,14 @@ export const AccessCell = React.memo((props: AccessCellProps) => {
       value={scope}
     />
   );
+
+  return tooltipText ? (
+    <Tooltip placement="top" title={tooltipText}>
+      <span>{radioBtn}</span>
```
dwiley-akamai, line 65: By default, tooltips do not activate on disabled elements. The workaround for this is to add a wrapper such as a `<span>`. More details [here](https://mui.com/material-ui/react-tooltip/#disabled-elements).

packages/manager/src/features/Profile/APITokens/CreateAPITokenDrawer.tsx
```diff
@@ -122,8 +136,19 @@ export const CreateAPITokenDrawer = (props: Props) => {
...
+        const basePermNameMapCopy = Object.assign({}, basePermNameMap);
+        delete basePermNameMapCopy.vpc;
+        setBasePermNameMap(basePermNameMapCopy);
+      }
     }
-  }, [open]);
+
+    return () => {
+      setBasePermNameMap(basePermNameMap);
+    };
```
dwiley-akamai, line 150: This cleanup function seems to be needed to prevent edge case bugs related to when, between drawer openings, the feature flag is toggled. Without this, during testing, I was observing situations where the scope for VPC was displaying when it shouldn't have been, or not showing up when it should've been.

packages/manager/src/features/Profile/APITokens/CreateAPITokenDrawer.tsx
```diff
@@ -102,6 +106,16 @@ export const CreateAPITokenDrawer = (props: Props) => {
     mutateAsync: createPersonalAccessToken,
   } = useCreatePersonalAccessTokenMutation();
 
+  const [basePermNameMap, setBasePermNameMap] = React.useState(
+    _basePermNameMap
+  );
```
bnussman-akamai, line 111: Rather than having state for this, could we derive the value?

Would it work if we had something like `getPermsMap` that would take `basePermNameMap` and omit values as needed? 

```ts
const basePermNameMap = getPermsMap(flags.vpc ? [] : ['vpc']);
```

dwiley-akamai: It can be derived. Avoiding the edge case I referenced previously would require having `useRef` in the mix, but I don't think that edge case is valid at this time: with open beta, API is returning `VPCs` in the `account.capabilities` array for all users. That actually makes all of the conditional logic for the PAT unnecessary, but having some conditional logic just to be on the safe side until VPC enters GA wouldn't be bad.

So we could do a simplified approach like

```
  // @TODO VPC: once VPC enters GA, remove _basePermNameMap logic and references.
  // Just use the basePermNameMap import directly w/o any manipulation.
  const basePermNameMap = getPermsNameMap(_basePermNameMap, {
    name: 'vpc',
    shouldBeIncluded: showVPCs,
  });
```

where the VPC perm would get removed if `showVPCs` was false. This would be vulnerable to the edge case, but that edge case would only come into play now if open beta got rolled back and we were testing with our DevTool (or the LD feature flag was toggled at the exact moment a user was interacting with this flow).

packages/manager/src/features/Profile/APITokens/ViewAPITokenDrawer.tsx
```diff
@@ -30,8 +35,32 @@ export const ViewAPITokenDrawer = (props: Props) => {
...
+      // If VPCs should not be shown, do not display the VPC scope
+      const basePermNameMapCopy = Object.assign({}, basePermNameMap);
+      delete basePermNameMapCopy.vpc;
+      setBasePermNameMap(basePermNameMapCopy);
+    }
+
+    return () => {
+      setBasePermNameMap(basePermNameMap);
+    };
+  }, [open, showVPCs]);
```
jaalah-akamai, line 62: We actually took a slightly different approach which I'm interested in hearing your thoughts. We actually removed them from the list of perms until it no longer needs to be feature flagged: https://github.com/linode/manager/pull/10040/files

dwiley-akamai: I noticed that and gave that approach a try earlier in the week, commenting out `'vpc'` in `basePerms` and `vpc: 'VPCs'` in `basePermNameMap` in `utils.ts` and adding the below (+ commenting out the current approach) in `CreateAPITokenDrawer.tsx`:

```
  if (showVPCs) {
    const vpcScopeIndex = allPermissions.findIndex(
      ([scope]) => scope === 'vpc'
    );

    if (vpcScopeIndex === -1) {
      allPermissions.push(['vpc', 0]);
    }

    basePermNameMap.vpc = 'VPCs';
  }
```

It resulted in some weird behavior though where the scope for VPC doesn't appear until another scope is clicked on:

https://github.com/linode/manager/assets/114682940/7bd9b91e-2ff0-4cdd-807b-053fe02d7711

Since GA isn't far off and because API is returning `VPCs` in `account.capabilities` for all users, I'm leaning towards the simplified approach I described in my previous comment

dwiley-akamai:
Looking into the unit test failures for `CreateAPITokenDrawer` & `ViewAPITokenDrawer`

dwiley-akamai:
> Looking into the unit test failures for `CreateAPITokenDrawer` & `ViewAPITokenDrawer`

Spent some time looking into these and haven't reached a solution -- the failed assertions basically say the VPC line/element cannot be found, but based on the flags/mock data and logic in the underlying components, that shouldn't be the case. Could I get a double-check on those in case I happened to overlook something? @bnussman-akamai @jaalah-akamai

packages/manager/src/features/Profile/APITokens/utils.ts
```diff
@@ -193,3 +195,22 @@ export const isWayInTheFuture = (time: string) => {
...
+ */
+export const getPermsNameMap = (
+  basePermNameMap: Record<string, string>,
+  perm: { name: string; shouldBeIncluded: boolean }
+) => {
+  if (basePermNameMap[perm.name] && !perm.shouldBeIncluded) {
+    delete basePermNameMap[perm.name];
+  }
+
+  return basePermNameMap;
```
jdamore-linode, line 215: This should fix the unit test failures:

```suggestion
  const modifiedBasePerms = {...basePermNameMap};
  if (modifiedBasePerms[perm.name] && !perm.shouldBeIncluded) {
    delete modifiedBasePerms[perm.name];
  }

  return modifiedBasePerms;
```

When the `ViewAPITokenDrawer` and `CreateAPITokenDrawer` components call this function, they both pass a reference to the `basePermNameMap` constant that's defined and exported at the top of this module.

Because the first several tests in `CreateAPITokenDrawer.test.tsx` and `ViewAPITokenDrawer.test.tsx` do not have VPC enabled, when `getPermsNameMap` gets called it's mutating the `basePermNameMap` object by deleting the `vpc` property. When subsequent tests import `basePermNameMap`, the `vpc` property has already been removed and as a result is excluded from the `getPermsNameMap` function's output and so the VPC permission is not rendered. You can test this by moving your tests to the top of their `.test` file; as long as they run before other tests where the `vpc` flag would be disabled, they'll pass.

I think you could fix this a couple ways but the suggestion above does it by ensuring that `getPermsNameMap` returns a new perm map rather than mutating the given map.

dwiley-akamai: 👀 That makes sense, thanks for taking a look and for the explanation!

bnussman-akamai:
Changes are looking good. Should we fix `Child Account Access` to use a similar pattern? Maybe `getPermsNameMap` could take an array of things to conditionally hide? I'm setting the bug where if I make a selection, `Child Account Access` pops up randomly.

jaalah-akamai:
> Changes are looking good. Should we fix Child Account Access to use a similar pattern? Maybe getPermsNameMap could take an array of things to conditionally hide? I'm seeing the bug where if I make a selection, Child Account Access pops up randomly.

I'll create a ticket and update this separately 👍

jaalah-akamai (approved):

bnussman-akamai (approved):

PR #10028, opened 2024-01-03
feat: [M3-7589] - Improve Accessibility of Button Component

jaalah-akamai:
## Description 📝
We aim to enhance the accessibility and inclusivity of our disabled buttons, particularly when it comes to effectively communicating context-specific messages explaining the reasons behind a button's disabled state.

UX Approved ✅

## Changes  🔄
- The tooltip icon has been integrated into the `Button` component, serving as an end adornment rather than residing independently outside of it.
- Using `aria-disabled` instead of applying the `disabled` attribute. This will maintain focus while still effectively conveying to Screen Readers that it exists but is currently disabled.
- Updated the `StyledActionButton.ts` to account for changes
- Majority of tests dealing with `Button` had to be changed from `.toBeDisabled();` to look for `aria-disabled` attribute.

> [!note]
The `disabled` attribute is still utilized for `loading` states

## Preview 📷
| UX Change |
| ------- |
|<video src="https://github.com/linode/manager/assets/125309814/ea453f02-872c-45fd-b89a-e7144e27842a" />|

| A11Y: Prod | A11Y: Dev |
| ------- | ------- |
|<video src="https://github.com/linode/manager/assets/125309814/30c435bd-4c25-4b1f-bcf5-0b0f7d6b54eb" />|<video src="https://github.com/linode/manager/assets/125309814/7076756c-b74f-40f6-bced-8784fdf4bd1d" />|
| Screenshot 2024-01-04 at 4 15 51 PM|

| Before  | After   |
| ------- | ------- |
|Screenshot 2024-01-03 at 4 10 56 PM| Screenshot 2024-01-03 at 4 11 53 PM|
|Screenshot 2024-01-03 at 4 12 21 PM|Screenshot 2024-01-03 at 4 13 14 PM|
| Screenshot 2024-01-03 at 3 53 26 PM | Screenshot 2024-01-03 at 1 40 33 PM |
| Hover: Screenshot 2024-01-03 at 3 54 07 PM | Hover: Screenshot 2024-01-03 at 4 06 12 PM |
Screenshot 2024-01-03 at 3 53 59 PM | Screenshot 2024-01-03 at 4 06 55 PM

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-01-03 at 5 09 33 PM | Screenshot 2024-01-03 at 5 09 20 PM|

| Variant  |   |
| ------- | ------- |
| Primary | Screenshot 2024-01-03 at 1 41 00 PM |
| Secondary |  Screenshot 2024-01-03 at 1 40 39 PM Screenshot 2024-01-03 at 1 41 05 PM
| Outlined | Screenshot 2024-01-03 at 1 40 46 PM Screenshot 2024-01-03 at 1 41 11 PM

### Prerequisites
- Log into a restricted user account with read_only perms

### Reproduction steps
- Go to /firewalls and observe ActionButton have been updated
- Then run `yarn storybook`
    - Go to http://localhost:6006/?path=/story/components-button--default 

### Verification steps 
- Check all variants and states (hover, focus, active)
- Verify voice over is able to read tooltip text, but that `aria-disabled="true"`
- Verify you can tab to disabled button
- Verify analytics events still work and are being fired

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


jaalah-akamai:
May be some broken E2E tests I'll need to fix with this

packages/manager/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -9,6 +9,7 @@ import { convertToKebabCase } from 'src/utilities/convertToKebobCase';
 
 export interface Action {
   disabled?: boolean;
+  id?: string;
```
jaalah-akamai, line 12: Optional because I'm just using it for react keys and test-id's.

packages/manager/src/components/InlineMenuAction/InlineMenuAction.tsx
```diff
@@ -53,6 +54,7 @@ export const InlineMenuAction = (props: InlineMenuActionProps) => {
       onClick={onClick}
       tooltipAnalyticsEvent={tooltipAnalyticsEvent}
       tooltipText={tooltip}
+      {...rest}
```
jaalah-akamai, line 57: Added to pass data-attributes to button component.

jdamore-linode:
> Need to update unit tests and some broken E2E tests

For the e2e tests, I'm thinking it may be a good idea to override the Chai `enabled` and `disabled` assertions to account for `aria-disabled` so that we don't have to explicitly check for it in each test.

Adding this to line 26 of `cypress/support/e2e.ts` seems to resolve a lot of the failures, but I'm seeing TypeScript issues reported in my editor -- I'm guessing Cypress is using a different version of Chai than what's in our `node_modules` dir but I didn't investigate:

```typescript
chai.use(function (chai, utils) {
  utils.overwriteProperty(chai.Assertion.prototype, 'disabled', function() {
    return function() {
      const obj = utils.flag(this, 'object');
      const isDisabled = Cypress.$(obj).is(':disabled');
      const isAriaDisabled = (Cypress.$(obj).attr('aria-disabled') === 'true');

      this.assert(
        isDisabled || isAriaDisabled,
        'expected #{this} to be disabled',
        'expected #{this} not to be disabled',
      );
    }
  });

  utils.overwriteProperty(chai.Assertion.prototype, 'enabled', function() {
    return function() {
      const obj = utils.flag(this, 'object');
      const isDisabled = Cypress.$(obj).is(':disabled');
      const isAriaDisabled = (Cypress.$(obj).attr('aria-disabled') === 'true');

      this.assert(
        !isDisabled && !isAriaDisabled,
        'expected #{this} to be disabled',
        'expected #{this} not to be disabled',
      );
    }
  });
});
```

(Something similar may be possible for the unit tests too)

jdamore-linode (changes requested):
This is really cool, nice work @jaalah-akamai! Observing one issue but other than that this is looking good to me. Tested keyboard navigation, tested VoiceOver on Chrome, but haven't tested mobile yet.

Wrote the comment below earlier, but later confirmed it's an issue -- can be reproduced on the OAuth page (and probably other places):

1. Navigate to `/profile/clients`
2. Click "Edit" next to any of the listed OAuth clients
3. Observe that the "Save Changes" button visually appears disabled, has the `aria-disabled` attribute, etc.
4. Observe that clicking "Save Changes" submits the form even when the button appears disabled

packages/manager/src/components/Button/Button.tsx
```diff
@@ -117,34 +120,42 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
...
+        aria-disabled={disabled}
+        buttonType={buttonType}
+        className={className}
+        color={color}
+        compactX={compactX}
+        compactY={compactY}
+        disableRipple={disabled}
+        disabled={loading}
+        loading={loading}
+        onClick={disabled ? undefined : rest.onClick}
```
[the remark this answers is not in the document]
jaalah-akamai: Ahh yes that should do it - thanks @jdamore-linode

packages/manager/src/components/Button/Button.test.tsx
abailly-akamai, line 1: Can we actually add coverage for the accessibility enhancements this PR brings up?

`loading` => disabled attribute, tabIndex -1, no click even handler
`disabled` => aria-disabled, tanIndex 0, click handlers
`tooltipText` => tooltip content on hover

packages/manager/src/components/Button/Button.tsx
```diff
@@ -117,34 +120,49 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
...
+        aria-disabled={disabled}
+        buttonType={buttonType}
+        className={className}
+        color={color}
+        compactX={compactX}
+        compactY={compactY}
+        disableRipple={disabled}
+        disabled={loading}
+        loading={loading}
+        onClick={disabled ? (e) => e.preventDefault() : rest.onClick}
```
abailly-akamai, line 147: This would have to be confirmed, but i think we want to handle `onKeyPress` similarly.

This may add another level of accessibility compliance.

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/DeleteKubernetesClusterDialog.tsx
```diff
@@ -54,7 +54,7 @@ export const DeleteKubernetesClusterDialog = (props: Props) => {
     >
       {error ? <Notice text={error?.[0].reason} variant="error" /> : null}
       <Notice variant="warning">
-        <Typography style={{ fontSize: '0.875rem' }}>
+        <Typography component="div" style={{ fontSize: '0.875rem' }}>
```
abailly-akamai, line 57: Might as well use `sx` here instead of style since we're touching it.

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/DeleteKubernetesClusterDialog.tsx
```diff
@@ -54,7 +54,7 @@ export const DeleteKubernetesClusterDialog = (props: Props) => {
...
           <strong>Warning:</strong>
           <ul style={{ margin: '5px 0px 0px', paddingLeft: '15px' }}>
```
abailly-akamai, line 59: nitpick: unrelated to PR but this should use our JSX `<List>` and `<ListItem>`

abailly-akamai (commented):
Great PR! happy to see this coming. Left a few comments to improve a few things.

Also, something slightly bothers me about the nomenclature. We are using the `disabled?` prop to now describe a "only visually disabled" element. This creates a shift  in the way we are describing a button state and the way any test out there (jest, ci) is treating a `disabled` assertion. 

Tho I am not convinced myself, I would argue that we may want to rename `disabled` to what is really is (`ariaDisabled` or `visuallyDisabled`) so that it matches assertions and nomenclature in general.

packages/manager/src/components/Button/Button.tsx
jdamore-linode: Just double checked this and wasn't able to trigger a focused `disabled` button using keyboard 👍 (a test to make sure this stays true would be really good though)

jdamore-linode (approved):
Nice! Approved pending the last few e2e test fixes.

Also circled back and confirmed the tooltip works as expected on mobile. I had to press and hold on iOS to get the tooltip to appear which took me a second to figure out, curious if others have seen this interaction before.

The last few e2e failures are all in `resize-linode.spec.ts`, and can be fixed by doing a find and replace of `cy.get('[data-qa-resize="true"]').click();` with `cy.get('[data-qa-resize="true"]').should('be.enabled').click();` (there are 4 or 5 instances). Before, Cypress would implicitly wait for a button to become enabled before clicking it, but this isn't true anymore with these changes so we just have to explicitly add the `enabled` assertion.

packages/manager/src/components/Button/Button.tsx
abailly-akamai: It may be an edge case when someone adds a custom onKeyDown handler, but worth covering IMO

jdamore-linode: Ohh, I misunderstood -- I totally see what you're saying now. Great catch and agreed

jaalah-akamai:
Thanks for the reviews @jdamore-linode and @abailly-akamai 
| Tests |
| ----|
|Screenshot 2024-01-05 at 11 19 13 AM|
|Screenshot 2024-01-05 at 11 33 17 AM|

jaalah-akamai:
> Tho I am not convinced myself, I would argue that we may want to rename disabled to what is really is (ariaDisabled or visuallyDisabled) so that it matches assertions and nomenclature in general.

Yea this is a fair point... let me see what I can do.

abailly-akamai (approved):
> Tho I am not convinced myself, I would argue that we may want to rename disabled to what is really is (ariaDisabled or visuallyDisabled) so that it matches assertions and nomenclature in general.
> 
> Yea this is a fair point... let me see what I can do.
> 
> Edit: This would require a significant amount of changes if we went down this route - whereas keeping it as disabled we keep the intent of what the developer wants to do... which is to make it disabled.

Yeah that's fine. It's not something i feel strong about.

packages/manager/src/components/Button/Button.tsx
```diff
@@ -141,6 +141,7 @@ export const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
         color={color}
         compactX={compactX}
         compactY={compactY}
+        data-testid="Button"
```
jdamore-linode: I think this may be the cause of the unit test failures -- I'm guessing we have instances of `<Button />` throughout the app with their own `data-testid`s that this is overriding

jaalah-akamai: Haha yea

PR #10029, opened 2024-01-03
refactor: [M3-7588] - Remove `classnames` and `@types/classnames`

bnussman-akamai:
## Description 📝

- Removes `classnames` (and `@types/classnames`) 🧹
  - `tss-react` contains the same helper (cx)
  - Streamlines how we handle conditional classnames 
  - I could be wrong, and the difference would be minimal, but using cx over classnames may improve performance because I think it uses the emotion cache?
  - I'm really just trying to find ways to continue to lighten up our dependencies

## How to test 🧪

- Check for UI regressions in touched components 👁️

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/components/PrimaryNav/NavItem.tsx
```diff
@@ -51,17 +53,14 @@ export const NavItem = React.memo((props: Props) => {
     <React.Fragment>
       {href ? (
         <Link
-          className={classNames({
-            [linkClasses(href)]: true,
-            listItemCollapsed: isCollapsed,
```
bnussman-akamai, line 56: `listItemCollapsed` isn't a thing so I removed it

dwiley-akamai (commented):
I need to finish looking through the app but the Selects, gravatars, and notifications/events are looking good so far

abailly-akamai (approved):
nice cleanup! this was an old lib that has been there from almost day one i think 👍 

visually tested all components affected against prod and found no regressions ✅

jaalah-akamai (approved):
Yea that's true - essentially a 1:1 swap here 🧹

dwiley-akamai (approved):
No regressions observed during my visual audit of the app ✅
Code review ✅

PR #10040, opened 2024-01-08
fix: Profile Query Keys, User Permissions loading state, and read_only PAT creation

bnussman-akamai:
## Description 📝
- Fixes profile query keys causing some mutations to not invalidate the correct query key
- Relates to https://github.com/linode/manager/pull/9987
- I left `useProfile` alone because the parent/child project needs `headers` functionality. Other profile query keys were reverted to their old format
- This bug is a perfect example of why we need a query key factory 🏭😖
- Also fixes:
   - An unrelated bug with User Permission loading state
   - An unrelated bug with `read_only` PAT resulting in an error: `Invalid OAuth scope: child_account:read_only`

## How to test 🧪

Query keys:
- Check out this PR
- Test profile related features such as...
  - Revoking a trusted device
  - Enabling / disabling 2FA
  - Creating, updating, and deleting SSH keys
- Verify the changes to `useGrants` does not break anything 
- Do you best to verify all query keys "match up"

Loading state:
- Have a restricted user on your account.
- Go to http://localhost:3000/account/users and select the "User Permissions" for that restricted user
- Confirm that the loading spinner displays while the queries load and then the Full Account Access is **disabled and toggled to false**; we should not see the page load with the Full Account Access toggled to true and enabled and then later change to false (the bug)

`read_only` PAT error:
- Go to http://localhost:3000/profile/tokens
- Create a Personal Access Token and select all to 'Read Only'
- Confirm that you can create a new PAT with Read Only permissions

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/Profile/APITokens/CreateAPITokenDrawer.tsx
+++ b/packages/manager/src/features/Profile/APITokens/CreateAPITokenDrawer.tsx
@@ -168,6 +168,18 @@ export const CreateAPITokenDrawer = (props: Props) => {
   const filteredPermissions = allPermissions.filter(
     (scopeTup) => basePermNameMap[scopeTup[0]] !== 'Child Account Access'
   );
+  // TODO: Parent/Child - remove this conditional once code is in prod.
+  // Note: We couldn't include 'child_account' in our list of permissions in utils
+  // because it needs to be feature-flagged. Therefore, we're manually adding it here.
+  if (flags.parentChildAccountAccess && user?.user_type !== null) {
+    const childAccountIndex = allPermissions.findIndex(
+      ([scope]) => scope === 'child_account'
+    );
+    if (childAccountIndex === -1) {
+      allPermissions.push(['child_account', 0]);
+    }
+    basePermNameMap.child_account = 'Child Account Access';
+  }
 
   return (
     <Drawer onClose={onClose} open={open} title="Add Personal Access Token">
--- a/packages/manager/src/features/Users/UserPermissions.tsx
+++ b/packages/manager/src/features/Users/UserPermissions.tsx
@@ -19,6 +19,7 @@ import { compose as recompose } from 'recompose';
 
 import { ActionsPanel } from 'src/components/ActionsPanel/ActionsPanel';
 import { Box } from 'src/components/Box';
+import { CircleProgress } from 'src/components/CircleProgress';
 // import { Button } from 'src/components/Button/Button';
 import { DocumentTitleSegment } from 'src/components/DocumentTitle';
 import { Item } from 'src/components/EnhancedSelect/Select';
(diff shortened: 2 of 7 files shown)


packages/manager/src/queries/profile.ts
```diff
@@ -60,28 +60,25 @@ export const updateProfileData = (
   newData: Partial<Profile>,
   queryClient: QueryClient
 ): void => {
-  queryClient.setQueryData([queryKey, undefined], (oldData: Profile) => ({
+  queryClient.setQueryData([queryKey, {}], (oldData: Profile) => ({
```
bnussman-akamai, line 63: `useProfile` will be `[queryKey, {}]`, **not** `[queryKey, undefined]`. This query key was incorrect

packages/manager/src/features/Users/UserPermissions.tsx
```diff
@@ -113,12 +114,13 @@ class UserPermissions extends React.Component<CombinedProps, State> {
...
 
   render() {
+    const { loading } = this.state;
     const { username } = this.props;
 
     return (
       <React.Fragment>
         <DocumentTitleSegment segment={`${username} - Permissions`} />
-        {this.renderBody()}
+        {loading ? <CircleProgress /> : this.renderBody()}
```
mjac0bs, line 123: This addresses a regression in the User Permissions page introduced in [here](https://github.com/linode/manager/pull/10005/files#diff-ba683f131f5e31776732082d7516053cb18f029960abb101f31027b8c575fd6eL101-L107) where eliminating the loading state check is causing the User Permissions page for a restricted user to display as Full Account Access set to true and then update to false once the data loads.

dwiley-akamai (approved):
Fix for profile-related features such as CRUD operations for SSH Keys ✅
Fix for "User Permissions" page loading state ✅

abailly-akamai:
@bnussman @mjac0bs the `two-factor-auth` e2e failure seems legit?

bnussman-akamai:
@abailly-akamai I believe I've pushed a fix. Parent child folks, please double check my work. I'm not super familiar with how `useProfile` should work anymore 😵‍💫

abailly-akamai (approved):
Confirm the `two-factor-auth` is now passing locally and both fixes (query and loading state) are good  as far as my testing went (authentication methods, SSH keys, trusted device)

packages/manager/src/features/Profile/APITokens/CreateAPITokenDrawer.tsx
```diff
@@ -168,6 +168,18 @@ export const CreateAPITokenDrawer = (props: Props) => {
...
+  // because it needs to be feature-flagged. Therefore, we're manually adding it here.
+  if (flags.parentChildAccountAccess && user?.user_type !== null) {
+    const childAccountIndex = allPermissions.findIndex(
+      ([scope]) => scope === 'child_account'
+    );
+    if (childAccountIndex === -1) {
+      allPermissions.push(['child_account', 0]);
+    }
+    basePermNameMap.child_account = 'Child Account Access';
+  }
```
mjac0bs, line 182: This was a bug caused by the fact that we can't yet add `child_account` to the scopes until it is returned by API, so we have to add it in the drawers for now.

image

jaalah-akamai (approved):

jdamore-linode (approved):

mjac0bs (approved):
`useProfile` looks good. SSH key CRUD, 2FA, and trusted devices was working as expected. Didn't see any unintended issues with the other queries.

PR #10045, opened 2024-01-09
upcoming: [M3-7463] - Disable Billing access user permission for child users

mjac0bs:
## Description 📝
As a child user, the UI of the permissions page will be the same as global or non-parent/child, except for Billing Access. Billing Access will be set to "Read Only" with the option for unrestricted child users to switch it to "None" for other child users. "Read-Write" will always be disabled. In other words, child account users will have no permission greater than "Read Only" billing access.

**Why are we enforcing this in the UI, rather than relying on the `account_access` grant?**
For an unrestricted (admin) child account user's `account_access` (which determines Billing Access), the API will return `read_write` in order for the user to have access to the rest of the user permissions, including managing those of their proxy user. Therefore, we need to make an additional check of the `user_type` to ensure on the front-end that we actually restrict billing access for `child` users.

## Changes  🔄
- The Billing Access default setting is configured to "Read Only" for child users.
- Child users have the option to switch Billing Access to "None."
- The "Read-Write" option is consistently disabled and cannot be selected by child users.
- Adds test coverage for the above to `user-permissions.spec.ts`.

## Preview 📷
Screenshot 2024-01-08 at 4 19 02 PM

### Prerequisites
- Check out this PR and `yarn dev`.
- In devtools, make sure the feature flag is on and enable the MSW.
- Go to `serverHandlers.ts` and edit the `account/users/:user` request to the following in order to mock a **child** user:
```
  rest.get('*/account/users/:user', (req, res, ctx) => {
  return res(ctx.json(accountUserFactory.build({ user_type: 'child' })));
  }),
```

### Verification steps 
(How to verify changes)
- Go to http://localhost:3000/account/users and navigate to the User Permissions for a mock restricted user.
- Confirm the Changes described above.
- Toggle off the feature flag and confirm that Read-Write billing access is *not* disabled and *is* the default selection.
- Play around with the Billing Access options and make sure there are no regressions in the selected options (e.g. no more than one option can be selected at the same time).
- Verify test passes:
```
```

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/Users/UserPermissions.tsx
+++ b/packages/manager/src/features/Users/UserPermissions.tsx
@@ -12,7 +12,7 @@ import { APIError } from '@linode/api-v4/lib/types';
 import { Paper } from '@mui/material';
 import Grid from '@mui/material/Unstable_Grid2';
 import { WithSnackbarProps, withSnackbar } from 'notistack';
-import { compose, flatten, lensPath, omit, set, view } from 'ramda';
+import { compose, flatten, lensPath, omit, set } from 'ramda';
 import * as React from 'react';
 import { QueryClient } from 'react-query';
 import { compose as recompose } from 'recompose';
@@ -407,13 +407,8 @@ class UserPermissions extends React.Component<CombinedProps, State> {
           <SelectionCard
             checked={
               grants.global.account_access === 'read_only' ||
-              Boolean(
-                this.state.isAccountAccessRestricted &&
-                  view<State, string>(
-                    lensPath(['grants', 'global', 'account_access']),
-                    this.state
-                  )
-              )
+              (this.state.isAccountAccessRestricted &&
+                Boolean(this.state.grants?.global.account_access))
             }
             data-qa-billing-access="Read Only"
             heading="Read Only"
--- a/packages/manager/src/features/Users/UserPermissions.tsx
+++ b/packages/manager/src/features/Users/UserPermissions.tsx
@@ -407,11 +407,13 @@ class UserPermissions extends React.Component<CombinedProps, State> {
           <SelectionCard
             checked={
               grants.global.account_access === 'read_only' ||
-              (this.state.isAccountAccessRestricted &&
-                view(
-                  lensPath(['grants', 'global', 'account_access']),
-                  this.state
-                ))
+              Boolean(
+                this.state.isAccountAccessRestricted &&
+                  view<State, string>(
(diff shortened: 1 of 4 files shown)


packages/manager/src/features/Users/UserPermissions.tsx
```diff
@@ -394,18 +403,29 @@ class UserPermissions extends React.Component<CombinedProps, State> {
...
           <SelectionCard
-            checked={grants.global.account_access === 'read_only'}
+            checked={
+              grants.global.account_access === 'read_only' ||
+              (this.state.isAccountAccessRestricted &&
+                view(
+                  lensPath(['grants', 'global', 'account_access']),
+                  this.state
+                ))
+            }
```
mjac0bs, line 413: Selection defaults to the Read Only card when the `account_access` grant is read only (for most users) OR account access is restricted (parent/child feature flag is on + is a `child` user) and Billing Access has not been selected as `None` (the other remaining option). 

Without the final check of the `account_access` grant in state, the card was not deselecting when the None card was clicked.

packages/manager/src/features/Users/UserPermissions.tsx
```diff
@@ -394,18 +405,31 @@ class UserPermissions extends React.Component<CombinedProps, State> {
...
           <SelectionCard
-            checked={grants.global.account_access === 'read_only'}
+            checked={
+              grants.global.account_access === 'read_only' ||
+              Boolean(
+                this.state.isAccountAccessRestricted &&
+                  view<State, string>(
+                    lensPath(['grants', 'global', 'account_access']),
+                    this.state
+                  )
```
bnussman-akamai, line 415: Would it be possible to un-ramda this? 

I don't really understand Ramda. Is this the same as doing `this.state.grants?.global.account_access`?

mjac0bs: Okay, yeah, I think we're good here with [a115b60](https://github.com/linode/manager/pull/10045/commits/a115b604664c7c60423ccb8cc78e1b20a4e13f39).

jdamore-linode (approved):
Nice work @mjac0bs!

mjac0bs:
The CI e2e failures are irrelevant to this PR and I connected with Banks offline regarding the above comment - we can't derive since user data isn't already stored in state.

Going to go ahead and merge. 🚢

PR #10046, opened 2024-01-09
changed: [Marketplace] Deprecate 7 Marketplace apps

hmorris3293:
## Description 📝
This PR removes the following apps from the Marketplace in cloud manager

Mist.io
Percona PMM
Bitninja
MagicSpam
ARK
TF2
Terraria


abailly-akamai (commented):
@hmorris3293 if this is a true deprecation with no plan to bring it back, let's delete those entries rather than commenting, and remove their associated metadata in https://github.com/linode/manager/blob/develop/packages/manager/src/features/OneClickApps/oneClickApps.ts

We also will need a changeset (see https://linode.github.io/manager/CONTRIBUTING.html) and update the one click app cache (I can help you with both once the cleanup mentioned above has been done. thx!

abailly-akamai (approved):

jaalah-akamai (approved):
Thanks @abailly-akamai for fixing test 👍

PR #10064, opened 2024-01-12
upcoming: [M3-7430] - Implement Account Switching Functionality

jaalah-akamai:
## Description 📝
Presently, there's no way for resellers to seamlessly log into indirect customer accounts. To overcome this obstacle, we aim to introduce a new feature. This feature allows users within the reseller account to choose an indirect customer account, obtaining the EUUID (Entity Unique User ID) in the process. Subsequently, this EUUID becomes instrumental in making a necessary endpoint call, triggering the generation of an ephemeral token allowing for the reseller to authenticate into a proxy account within the main indirect reseller's account. The primary goal of this token is to streamline the transition, providing reseller users with a frictionless experience when logging into the proxy account.

## Changes  🔄
- Moved the rendering of child accounts into a new component called `ChildAccountList.tsx`
- Implemented functionality / logic to:
    - Request ephemeral token to log into the proxy account of a specific indirect customer account.
    - Set both reseller and proxy token in local storage to allow for account switching, utilizing the original `/authentication/` token as the active account.
    - Account for loading and error states if requests fail or tokens expire.
- Added unit test for `SwitchAccountButton.tsx`

All PAT in this demo have been revoked. 👌 

| Fetch Proxy Token, Set Storage, Switch Accounts |
| ------- | 
|<video src="https://github.com/linode/manager/assets/125309814/f75e99e6-af65-48ee-b950-3be16c5ad9ae" />|

>[!note]
We will most likely be removing the toast when some of the errors are shown (as in not to show both a notice _and_ a toast). While this is still in draft, I wanted to get some UX input.

| Description  | Preview  |
| ------- | ------- |
| Loading of indirect customer accounts | Screenshot 2024-01-12 at 11 31 06 AM |
| Error in loading indirect customer accounts | Screenshot 2024-01-11 at 12 52 02 PM |
| No indirect customer accounts | Screenshot 2024-01-12 at 3 38 18 PM |
| Error switching to indirect customer accounts | <video src="https://github.com/linode/manager/assets/125309814/b570f698-ff59-4133-af04-99a35f377f8f" /> |
| - Success toast when logging into proxy account <br />- Error if parent token is expired  | <video src="https://github.com/linode/manager/assets/125309814/7147fa91-3ce3-4006-99b1-09bed44081f2" /> |

## How to test 🧪

>[!Important]
You won't see any actual UI changes when performing the switch account feature until we're off mocks.

### Prerequisites
- Add `REACT_APP_PROXY_PAT` to your `.env` file with a temporary PAT that you've created (could be the same account for testing)
- Enable `Parent / Child Account` feature flag
- Turn on Mock Service Worker (MSW)
- Open dev tools > Application > and clear local storage
- Ensure that `{ user_type: 'parent' }` in [serverHandlers.ts](https://github.com/linode/manager/blob/develop/packages/manager/src/mocks/serverHandlers.ts#L1267)

### Reproduction & Verification steps

⭐ **Parent View**
- Open dev tools > Application > and make note of payload under `authentication/token`
- Open user menu in Cloud Manager and click/press "Switch Account"
- Select any child account in the list
- Observe that tokens have been created for parent/proxy accounts and that:
    - Original token has been stored under `authentication/parent_token/token`
    - `authentication/token` and `authentication/proxy_token/token` contain the same token indicating we've switched to a proxy account.

⭐ **Proxy View: Switching Into Another Proxy**
- Open dev tools > Application > and make note of token value under `authentication/token`
- Open user menu in Cloud Manager and click/press "Switch Account"
- Select any child account in the list
- Observe that:
    - `authentication/parent_token/token` remains the same
    - `authentication/token` and `authentication/proxy_token/token` have new token values
   
⭐ **Proxy View: Switching Back to Parent**
- Open dev tools > Application > and make note of token value under `authentication/token` (should still mirror proxy token)
- Update user type `{ user_type: 'proxy' }` in [serverHandlers.ts](https://github.com/linode/manager/blob/develop/packages/manager/src/mocks/serverHandlers.ts#L1267)
- Open user menu in Cloud Manager and click/press "Switch Account"
- Select "switch back to your account"
- Observe that `authentication/token` token value matches `authentication/parent_token/token`

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/features/Account/SwitchAccountDrawer.tsx
```diff
@@ -1,104 +1,201 @@
...
+        // ================================================================
+        // throw new Error(
+        //   `There was an error switching to ${companyName}. Please try again.`
+        // );
+        // ================================================================
+
+        const proxyToken = await getProxyToken({ euuid, headers });
+
+        // We don't need to worry about this if we're a proxy user.
+        if (!isProxyUser) {
```
jcallahan-akamai, line 97: Minor/nit: the readability of this sequence could be improved slightly IMO– I was confused how we could get a proxy token AS another proxy user. (It became clearer once I saw how the `headers` argument was used). 

Conceptually, maybe the `getProxyToken` function itself should decide what the headers should be based on whether it's a parent/sibling account? (Instead of taking in arbitrary headers).

jaalah-akamai: Great suggestion - I like that idea! 👍

packages/manager/src/features/Account/SwitchAccountDrawer.tsx
```diff
@@ -1,104 +1,213 @@
...
+          headers:
+            userType === 'proxy'
+              ? {
+                  Authorization: `Bearer ${token}`,
+                }
+              : undefined,
+        });
+      } catch (error) {
+        setIsProxyTokenError(error as APIError[]);
+        throw error;
```
jaalah-akamai, line 71: This will cause the try block in `handleSwitchToChildAccount` to error.

packages/manager/src/features/Account/SwitchAccountDrawer.tsx
```diff
@@ -1,104 +1,198 @@
...
+            prefix: 'authentication/parent_token',
+            token: parentToken,
+          });
+        }
+
+        const proxyToken = await getProxyToken({
+          euuid,
+          token: isProxyUser
+            ? currentTokenWithBearer
+            : getStorage('authentication/parent_token/token'),
```
jaalah-akamai: For 'proxy' userType, use the stored parent token in the request. Otherwise, if 'parent' use current token.

packages/manager/src/features/Account/SwitchAccountDrawer.tsx
```diff
@@ -1,104 +1,198 @@
...
+            prefix: 'authentication/parent_token',
+            token: parentToken,
+          });
+        }
+
+        const proxyToken = await getProxyToken({
+          euuid,
+          token: isProxyUser
+            ? getStorage('authentication/parent_token/token')
+            : currentTokenWithBearer,
```
jaalah-akamai, line 115: For 'proxy' userType, use the stored parent token in the request. Otherwise, if 'parent' use current token.

packages/manager/src/features/Account/SwitchAccountDrawer.tsx
```diff
@@ -1,104 +1,198 @@
...
-        }}
-        key={`child-account-link-button-${idx}`}
-      >
-        {childAccount.company}
-      </StyledChildAccountLinkButton>
-    ));
-  }, [childAccounts, error, isLoading]);
+    updateCurrentTokenBasedOnUserType({ userType: 'parent' });
+    handleClose();
+  }, [handleClose, isProxyUser]);
```
jaalah-akamai, line 161: Switching back to parent is simpler. Validate parent token in local storage and swap to it.

packages/manager/src/features/Account/utils.ts
```diff
@@ -0,0 +1,77 @@
...
+    // new Date(mockExpiredTime).toISOString()
+    // ================================================================
+  ) {
+    return false;
+  }
+  return true;
+};
+
+/**
+ * Set token information in the local storage.
```
mjac0bs, line 38:
```suggestion
 * Set token information in the local storage.
 * This allows us to store a token for later use, such as switching between parent and proxy accounts.
```

I think we should document this more clearly.

packages/manager/src/features/Account/SwitchAccountButton.test.tsx
mjac0bs, line 1: Thanks for adding this!

mjac0bs (commented):
Here's some more feedback; I will give this an additional pass through to make sure that all loading and error states are addressed once this stuff is addressed.

Noteworthy:
- Without a page reload, I don't think we can actually account switch, even if we were off mocks. Updating local storage isn't enough without a refresh. Is that not the case?
- We lost a feature flag in `canSwitchBetweenParentOrProxyAccount` and so there's still a regression in the user menu.

Looks like there's a failure on `SwitchAccountDrawer.test.tsx` that probably came about when `ChildAccountList` was pulled out into its own component:

image

packages/manager/src/features/TopMenu/UserMenu/UserMenu.tsx
```diff
@@ -50,35 +49,49 @@ const profileLinks: MenuLink[] = [
...
-  // For proxy accounts: configure request headers using a parent's token to fetch the parent's username from /profile.
-  const headers =
-    flags.parentChildAccountAccess && user?.user_type === 'proxy'
-      ? new AxiosHeaders({ Authorization: authentication.token.get() }) // TODO: Parent/Child - M3-7430: replace this token with the parent token in local storage.
+  const hasGrant = (grant: GlobalGrantTypes) =>
+    grants?.global?.[grant] ?? false;
+  const isRestrictedUser = profile?.restricted ?? false;
+  const hasAccountAccess = !isRestrictedUser || hasGrant('account_access');
+  const hasReadWriteAccountAccess = hasGrant('account_access') === 'read_write';
+  const user_type = user?.user_type;
```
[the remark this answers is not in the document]
jaalah-akamai: How did snake_case get in there? 🙄 😬  It's an old habit, but in this case it's not used too many times and it's already readable so I can save some space here 👍

carrillo-erik:
I've verified the switch account flow as described in the reproduction steps. I was able to observe the updates to the `authentication/token`,  `authentication/parent_token/token`, and `authentication/proxy_token/token` upon switching accounts. Also, after changing the `user_type` I verified that `authentication/token` matched the `authentication/parent_token/token`.

Regarding the note on the `error` states, I'm all for removing the toast in favor of displaying an error notice.

jaalah-akamai:
Toasts should be removed for error states 👍

packages/manager/src/features/Account/utils.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const userToken = getStorage(`${storageKeyPrefix}/token`, false);
+  const userScope = getStorage(`${storageKeyPrefix}/scopes`, false);
+  const userExpiry = getStorage(`${storageKeyPrefix}/expire`, false);
+
+  if (userToken) {
+    setStorage('authentication/token', userToken);
+    setStorage('authentication/scopes', userScope);
+    setStorage('authentication/expire', userExpiry);
+  }
```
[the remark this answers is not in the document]
mjac0bs: I gave that a try back in my [old POC branch](https://github.com/mjac0bs/manager/commit/9e667e2c60c7143505db7f3b2914f79720815f36) to try account switching with real accounts off mocks and see what would happen, and wasn't able to successfully account switch so not sure about this. But I've forgotten exactly why it was that we needed the refresh for the new token to take affect, so if you want to leave as is until we can test for certain in the dev environment, that's fine.

packages/manager/src/features/Account/SwitchAccountDrawer.tsx
```diff
@@ -1,104 +1,207 @@
...
+        // throw new Error(
+        //   `Account switching failed. Try again.`
+        // );
+        // ================================================================
+
+        // We don't need to worry about this if we're a proxy user.
+        if (!isProxyUser) {
+          const parentToken = {
+            expiry: storage.authentication.expire.get(),
+            scopes: storage.authentication.scopes.get(),
```
mjac0bs, line 109:
```suggestion
            expiry: getStorage('authenication/expire'),
            scopes: getStorage('authenication/scopes'),
```
Would prevent us from having to import `storage` in this file, since we're already using `getStorage()` elsewhere.

mjac0bs (approved):
Thanks for addressing all the little pieces of feedback! 🙌🏼 

Confirmed:
- [x] When feature flag is off, no regressions in user menu.
- [x] When feature flag is on with `user_type: null`, no regressions in user menu.
- [x] When feature flag is on with `user_type: child`, no regressions in user menu.
- [x] When feature flag is on with `user_type: parent`, user menu displays the parent's company name with a clickable Switch Account button, which opens the Switch Account drawer. Clicking a child account swaps the proxy token into local storage active token as saves the current token as the local storage parent token.
- [x] When feature flag is on with `user_type: parent`, user menu displays the parent's company name with a clickable Switch Account button, which opens the Switch Account drawer. Drawer allows user to select another proxy account or switch back to parent account (swapping parent token into local storage active token).
- [x] Switch Account drawer gracefully handles an error when switching to a child account (shows error notice in drawer).
- [x] Switch Account drawer gracefully handles loading state and error loading child accounts (shows error copy and Try Again button in drawer, which tries to load accounts again when clicked).
- [x] Switch Account drawer gracefully handles error with expired parent token when switching back to parent account from proxy (shows error notice in drawer).
- [x] Same behavior as above for the Switch Account button on the Account Landing page.

Notes:
- I don't think we have any loading state when clicking a child account and completing the account switch. May be worth adding as a follow up.

This LGTM. I'm still not certain the actual switching will work as expected, as I commented, but we know token swapping is. Looking forward to re-confirming all this with actual API data.

jaalah-akamai:
> I don't think we have any loading state when clicking a child account and completing the account switch. May be worth adding as a follow up.

carrillo-erik (approved):

PR #10065, opened 2024-01-15
test: [M3-7482] - Add cypress tests to add users on "Users & Grants" page

cliu-akamai:
## Description 📝
Add regression tests for adding users on `Users & Grants` page.

## Major Changes 🔄
- Add regression tests to add users on `Users & Grants` page.

## How to test 🧪
```
```

### DIFF
--- a/packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
+++ b/packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
@@ -4,6 +4,7 @@ import { accountUserFactory } from '@src/factories/accountUsers';
 import { grantsFactory } from '@src/factories/grants';
 import { userPermissionsGrants } from 'support/constants/user-permissions';
 import {
+  interceptAddUser,
   mockGetUser,
   mockGetUserGrants,
   mockGetUsers,
@@ -18,7 +19,7 @@ import { mockGetProfile } from 'support/intercepts/profile';
 import { ui } from 'support/ui';
 import { shuffleArray } from 'support/util/arrays';
 import { makeFeatureFlagData } from 'support/util/feature-flags';
-import { randomLabel } from 'support/util/random';
+import { randomHex, randomLabel } from 'support/util/random';
 
 // Message shown when user has unrestricted account acess.
 const unrestrictedAccessMessage =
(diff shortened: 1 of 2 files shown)


bnussman-akamai (approved):

packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
```diff
@@ -558,4 +559,246 @@ describe('User permission management', () => {
...
+    };
+
+    // TODO: Parent/Child - M3-7559 clean up when feature is live in prod and feature flag is removed.
+    mockAppendFeatureFlags({
+      parentChildAccountAccess: makeFeatureFlagData(false),
+    }).as('getFeatureFlags');
+    mockGetFeatureFlagClientstream().as('getClientStream');
+
+    mockGetUsers([mockUser]).as('getUsers');
+    mockGetUser(mockUser).as('getUser');
```
mjac0bs, line 704:
```suggestion
    mockGetUser(mockUser);
```
It doesn't look like we're `wait`ing on this anywhere.

packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
```diff
@@ -558,4 +559,246 @@ describe('User permission management', () => {
...
+    });
+
+    const mockUserGrantsUpdated = grantsFactory.build();
+    const mockUserGrants = {
+      ...mockUserGrantsUpdated,
+      global: undefined,
+    };
+
+    mockGetUsers([mockUser]).as('getUsers');
+    mockGetUser(mockUser).as('getUser');
```
mjac0bs, line 583:
```suggestion
    mockGetUser(mockUser);
```
It doesn't look like we're `wait`ing on this anywhere.

packages/manager/cypress/support/intercepts/account.ts
```diff
@@ -91,6 +91,15 @@ export const mockGetUser = (user: User): Cypress.Chainable<null> => {
...
 
+/**
+ * Intercepts POST request to add an account user.
+ *
+ * @returns Cypress chainable.
+ */
+export const interceptAddUser = (): Cypress.Chainable<null> => {
+  return cy.intercept('POST', apiMatcher('account/users'));
+};
+
```
jdamore-linode, line 102: I think we need to replace this with a mock util, e.g. `mockAddUser`, which mocks a response containing the created user object.

Because this is only intercepting the request, this test actually creates new users when it runs. We want to make sure this response is mocked so we don't leave behind any leftover users here -- your assertions on the request body `restricted` property should continue to work 👍

packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
```diff
@@ -18,7 +19,7 @@ import { mockGetProfile } from 'support/intercepts/profile';
 import { ui } from 'support/ui';
 import { shuffleArray } from 'support/util/arrays';
 import { makeFeatureFlagData } from 'support/util/feature-flags';
-import { randomLabel } from 'support/util/random';
+import { randomHex, randomLabel } from 'support/util/random';
```
jdamore-linode, line 22:
```suggestion
import { randomLabel } from 'support/util/random';
```

packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
```diff
@@ -558,4 +559,246 @@ describe('User permission management', () => {
           .click();
       });
   });
+
+  it('can add full-access users with no redirect occurs', () => {
```
jdamore-linode, line 563:
```suggestion
  it('can add users with full access', () => {
```
No need to mention the redirect in the test title, similar case for the test below. Only suggesting we change it because if we ever change the redirect behavior, we'll have to change the test titles too and that'll mess up our metrics for the test (test length, flakiness, etc.)

packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
```diff
@@ -558,4 +559,246 @@ describe('User permission management', () => {
...
+    cy.wait('@getUsers');
+
+    // the new user is displayed in the user list
+    cy.findByText(newUser.username).should('be.visible');
+
+    // no redirect occurs
+    cy.url().should('endWith', '/users');
+  });
+
+  it('can add non full-access users with redirecting to the new user\'s "User Permission" page', () => {
```
jdamore-linode, line 678:
```suggestion
  it('can add users with limited access', () => {
```
(Similar to above)

jdamore-linode (changes requested):
Nice work Cassie! I think we'll want to replace the intercept util with a mock util, and I made a couple other really minor suggestions, but otherwise this is looking really good!

packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
jdamore-linode: @mjac0bs Thanks! `restricted` was what I was going for but for some reason `limited` came to mind and my brain was like "yeah, sounds right 👍😎". Updated my suggestion

mjac0bs (approved):
Thanks for the changes; looks good to me! 🚢

jdamore-linode (approved):
Looks great, thanks @cliu-akamai!

PR #10080, opened 2024-01-18
upcoming: [M3-7659] - Add `user_type` to `/profile` endpoint

jaalah-akamai:
## Description 📝
We're updating our `/profile` endpoint to return the user role associated with Parent/Child Accounts by adding a new `user_type` field. Previous we had to get this information from the `/users` or `/users/:user` endpoint. This will allow us to cut down on the amount of queries we need in order to determine the user role.

## Changes  🔄
- Add `user_type` to `type Profile` in APIv4
- You'll see with these changes we were able to remove `useAccountUser` query hooks
- For MSW, you'll have to manually change the `user_type` for `*/profile` similar to how we do for `*/account/users/:user`. 

### Prerequisites
- One account with full access
- One account with restricted access

### Reproduction steps
- Go to http://localhost:3000/account/billing

### Verification steps 
1. Log into any restricted account with parent/child feature flag **disabled**, observe the `Edit` and `Add Payment Method` buttons are disabled.
2. Turn parent/child feature flag on with Mock Service Workers
    - Observe the `Edit` and `Add Payment Method` buttons are enabled (by default we're in a `parent` state)
    - Then change `parent` to `child` [on this line](https://github.com/linode/manager/blob/9ad5df4411d3c21a2d78645b2e73e29f105aed5d/packages/manager/src/mocks/serverHandlers.ts#L554)
    - Observe the `Edit` and `Add Payment Method` buttons are disabled.
    - Now change MSW from `child` to `proxy` and observe the `Edit` and `Add Payment Method` buttons are enabled.

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


jaalah-akamai:
Moving with an approach to add `user_type` to /profile endpoint instead.

mjac0bs (approved):
We'll need to upate `restricted-user-billing.spec.ts` for the child accounts.
Screenshot 2024-01-19 at 11 51 03 AM

Approving pending that is done; confirmed that by updating the mock `profile` endpoint, the `user_type` is returned and correctly disables the buttons for child users only. 🚢 

Screenshot 2024-01-19 at 11 47 06 AM

packages/api-v4/src/profile/types.ts
```diff
@@ -1,3 +1,5 @@
+import { UserType } from '../account';
```
bnussman-akamai, line 1: Optional: 

```suggestion
import type { UserType } from '../account';
```

packages/manager/src/features/Billing/BillingDetail.tsx
```diff
@@ -77,13 +82,15 @@ export const BillingDetail = () => {
...
           taxId={account.tax_id}
+          userType={userType}
           zip={account.zip}
         />
         <PaymentInformation
           error={paymentMethodsError}
           isAkamaiCustomer={account?.billing_source === 'akamai'}
           loading={paymentMethodsLoading}
           paymentMethods={paymentMethods}
+          userType={userType}
```
bnussman-akamai, line 93: Can we just move the `useProfile` into these components instead of passing `userType` via props? 

I find that that has benefits in readability, maintainability, and even performance at times.

jaalah-akamai: Yea absolutely

carrillo-erik (approved):

bnussman-akamai (approved):

PR #10084, opened 2024-01-19
feat: [M3-7665] - RegionMultiSelect Component

cpathipa:
## Description 📝
This Allows to select multiple regions and renders the list of selected regions

## Changes  🔄
- RemovableSelectionsList - Added an optional prop to render custom label component.

## Preview 📷

| Region Multi select  | Renders selected list   |
| ------- | ------- |
|image|image|

## How to test 🧪
- Run story book verify the functionality of Region Multi select component.

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset


packages/manager/src/components/RegionSelect/RegionMultiSelect.tsx
```diff
@@ -0,0 +1,204 @@
...
+    () =>
+      getRegionOptions({
+        accountAvailabilityData: accountAvailability,
+        currentCapability,
+        regions,
+      }),
+    [accountAvailability, currentCapability, regions]
+  );
+
+  const removeOption = (optionToRemove: RegionSelectOption) => {
```
jaalah-akamai, line 99:
```suggestion
  const handleRemoveOption = (optionToRemove: RegionSelectOption) => {
```

packages/manager/src/components/RegionSelect/RegionMultiSelect.tsx
```diff
@@ -0,0 +1,204 @@
...
+          onBlur={onBlur}
+          options={options}
+          placeholder={placeholder ?? 'Select Regions'}
+          renderTags={() => null}
+          value={selectedRegions}
+        />
+      </StyledAutocompleteContainer>
+      {selectedRegions.length > 0 && RenderSelectedRegionsList && (
+        <RenderSelectedRegionsList
+          onRemove={removeOption}
```
jaalah-akamai, line 198:
```suggestion
          onRemove={handleRemoveOption}
```

jaalah-akamai:
The dark mode background colors/shading is a little weird. Do we also want to have a border under `Select All` and `Deselect All` like in the original Autocomplete example?

packages/manager/src/components/RegionSelect/RegionMultiSelect.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+    </Box>
+  );
+};
+
+const SelectedRegionsList = ({
+  onRemove,
+  selectedRegions,
+}: SelectedRegionsProps) => {
+  const handleRemove = (data: RemovableItem) => {
+    const { id: _id, ...dataWithoutId } = data;
```
abailly-akamai, line 49: This is confusing - why are you doing this?

packages/manager/src/components/RegionSelect/RegionMultiSelect.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+      noDataText=""
+      onRemove={handleRemove}
+    />
+  );
+};
+
+export const Default: StoryObj<RegionMultiSelectProps> = {
+  render: (args) => {
+    const SelectWrapper = () => {
+      const [selectedRegions, setSelectedRegions] = useState<string[]>([]);
```
abailly-akamai, line 69: I know it's a story, but we can be a little better at naming things here. ex: 
```suggestion
      const [selectedRegionsIds, setSelectedRegionsIds] = useState<Region['id'][]>([]);
```
for better readability

packages/manager/src/components/RegionSelect/RegionMultiSelect.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+      const handleSelectionChange = (selectedIds: string[]) => {
+        debugger;
```
abailly-akamai, line 72: remove debugger

packages/manager/src/components/RegionSelect/RegionMultiSelect.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+        currentCapability="Block Storage"
+        handleSelection={mockHandleSelection}
+        regions={regions}
+        selectedIds={[]}
+      />
+    );
+
+    screen.getByRole('combobox', { name: 'Regions' });
+  });
+});
```
abailly-akamai, line 29: we need better coverage here and test interaction as well as checking a region properly gets added to the list underneath

packages/manager/src/components/RegionSelect/RegionMultiSelect.tsx
```diff
@@ -0,0 +1,204 @@
...
+    return 0;
+  }
+  if (a.label > b.label) {
+    return 1;
+  }
+  if (a.label < b.label) {
+    return -1;
+  }
+  return 0;
+};
```
abailly-akamai, line 37: Screenshot 2024-01-22 at 10 23 36

Would be nice to keep the same ordering as in the RegionSelect itself. You could extract the sorting alg from `getRegionOptions` as its own util and use it in both places.

packages/manager/src/components/RegionSelect/RegionMultiSelect.tsx
```diff
@@ -0,0 +1,204 @@
...
+            return option?.data?.region;
+          }}
+          isOptionEqualToValue={(
+            option: RegionSelectOption,
+            value: RegionSelectOption
+          ) => option.value === value.value}
+          onChange={(_, selectedOption) =>
+            handleRegionChange(selectedOption as RegionSelectOption[])
+          }
+          renderOption={(props, option, { selected }) => {
```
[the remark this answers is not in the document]
abailly-akamai: seconding this, if possible

packages/manager/src/components/RegionSelect/RegionMultiSelect.tsx
```diff
@@ -0,0 +1,204 @@
...
+            if (!option.data) {
+              // Render options like "Select All / Deselect All "
+              return (
+                <StyledListItem {...props}>
+                  <Box
+                    sx={{
+                      alignItems: 'center',
+                      // display: 'flex',
```
abailly-akamai, line 135: cleanup

packages/manager/src/components/RegionSelect/RegionSelect.utils.ts
```diff
@@ -151,3 +152,21 @@ export const getRegionOptionAvailability = ({
...
+    .map((selectedRegionId) =>
+      getSelectedRegionById({
+        accountAvailabilityData,
+        currentCapability,
+        regions,
+        selectedRegionId,
+      })
+    )
+    .filter((region): region is RegionSelectOption => !!region);
+};
```
abailly-akamai, line 172: Be consistent with JSDocs comment and making sure you describe your utils

packages/manager/src/components/RegionSelect/RegionSelect.utils.ts
```diff
@@ -151,3 +152,21 @@ export const getRegionOptionAvailability = ({
...
+  return selectedRegionIds
+    .map((selectedRegionId) =>
+      getSelectedRegionById({
+        accountAvailabilityData,
+        currentCapability,
+        regions,
+        selectedRegionId,
+      })
+    )
+    .filter((region): region is RegionSelectOption => !!region);
```
abailly-akamai, line 171: can we avoid casting here?

abailly-akamai (commented):
Thanks for breaking this down in its own component! 

The component tested in Storybook looks good 👍 

Left a couple questions and some comments to improve code 

Also, I am not a fan of this UI pattern in general. When selecting a region, the user has no idea a list is behind the dropdown, and the checkmark next to the region is faint. There's a risk the user does not know the selection worked. I'd like to bring this to UX cause we may want to re-thing this pattern. CC @jaalah-akamai 

Screenshot 2024-01-22 at 10 28 40

packages/manager/src/components/RegionSelect/RegionMultiSelect.stories.tsx
cpathipa: Added this because the 'RemovableItem' structure consists of an 'id' that is not present in RegionSelectOption. So, I thought it best to eliminate it. However, I now think we don't need it here, so I have removed it.

packages/manager/src/components/RegionSelect/RegionMultiSelect.tsx
cpathipa: Actually, based on user session feedback, it seems that the order of selected regions should match that of the buckets table, which displays regions in alphabetical order. However, I like the idea of maintaining the same ordering as in RegionSelect and have kept it as the default behavior. This comes with the flexibility of rendering the selected list order through the `sortRegionOptions` prop.

packages/manager/src/components/RegionSelect/RegionSelect.utils.ts
cpathipa: Sure, Added descriptive comments to the utils 135a8dd

packages/manager/src/components/RegionSelect/RegionSelect.utils.ts
cpathipa: I think we could do something like this 
```}: GetSelectedRegionsByIdsArgs): (RegionSelectOption | undefined)[] => {
  return selectedRegionIds
    .map((selectedRegionId) =>
      getSelectedRegionById({
        accountAvailabilityData,
        currentCapability,
        regions,
        selectedRegionId,
      })
    )
    .filter(Boolean);```

However, this approach might cause us to update all the references to `(RegionSelectOption | undefined)[]`. Also, we may need to refer to it using optional chaining.

**Comment by cpathipa** [comment]
@jaalah-akamai @jaalah-akamai Great feedback.

- Renamed - `RenderSelectedRegionsList`  to `SelectedRegionsList`
- Renamed - `removeOption` to `handleRemoveOption`
- improved readability in story book variable naming - `selectedRegionsIds`, `setSelectedRegionsIds`
- Removed debugger and commented code and described the util function.
- Added UX concern of selected region list and checkmark selection item to our next cafe.
-  Added border under Select All and Deselect All.
- Created a followup ticket M3-7675 for coverage and to make renderOption as reusable for `RegionSelect` and `RegionMultiSelect`.

File: packages/manager/src/components/RegionSelect/RegionSelect.utils.ts (hunk 11 above)
Comment (abailly-akamai) [reply packages/manager/src/components/RegionSelect/RegionSelect.utils.ts:176]: Why not using a `flatMap`?

```
export const getSelectedRegionsByIds = ({
  accountAvailabilityData,
  currentCapability,
  regions,
  selectedRegionIds,
}: GetSelectedRegionsByIdsArgs): RegionSelectOption[] => {
  return selectedRegionIds.flatMap((selectedRegionId) => {
    const region = getSelectedRegionById({
      accountAvailabilityData,
      currentCapability,
      regions,
      selectedRegionId,
    });
    return region ? [region] : [];
  });
};
```

File: packages/manager/src/components/RegionSelect/RegionMultiSelect.tsx [hunk 12]
```diff
@@ -0,0 +1,160 @@
...
+      </StyledAutocompleteContainer>
+      {selectedRegions.length > 0 && SelectedRegionsList && (
+        <SelectedRegionsList
+          selectedRegions={
+            sortRegionOptions
+              ? [...selectedRegions].sort(sortRegionOptions)
+              : selectedRegions
+          }
+          onRemove={handleRemoveOption}
+        />
```
abailly-akamai, line 156: Still a purely alphabetical order - Not wrong! just wanted to point it out since it differs from the order in the select

Screenshot 2024-01-23 at 09 06 50

packages/manager/src/components/RegionSelect/RegionMultiSelect.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+  );
+};
+
+const SelectedRegionsList = ({
+  onRemove,
+  selectedRegions,
+}: SelectedRegionsProps) => {
+  const handleRemove = (data: RemovableItem) => {
+    onRemove((data as unknown) as RegionSelectOption);
+  };
```
abailly-akamai, line 55: No need to cast here I think

```suggestion
 const handleRemove = (item: RemovableItem) => {
    onRemove(item.data);
 };
```

abailly-akamai (commented):
Thanks for the changes! - a couple more comments to clean things up and this should be good to go

packages/manager/src/components/RegionSelect/RegionSelect.utils.ts
cpathipa: IMO, the current approach of using `map` followed by `filter` is more straightforward and readable for this specific scenario. It clearly shows the intent of transforming each ID into a region and then filtering out any invalid or non-existent regions. The type guard in the filter (`region` is `RegionSelectOption`) is a clean way to ensure type safety and clarity in the code, making it evident that the final array should only contain valid `RegionSelectOption` objects. Unless there's a specific reason or benefit we are aiming with `flatMap`. I think using `flatMap` directly isn't really suitable in this scenario unless `getSelectedRegionById` return an array.

abailly-akamai: ok. but to be clear, this only provides compile-time type safety, it doesn't guarantee runtime type safety.

packages/manager/src/components/RegionSelect/RegionMultiSelect.tsx
cpathipa: Sure thing, added a follow-up ticket (M3-7675) to revisit the order of rendering selected regions. At this point, the component supports the flexibility of using a custom component to render the list and sort order.

packages/manager/src/components/RegionSelect/RegionSelect.utils.ts
cpathipa: Yeah, you are right. Considering the context and requirements, I think we are good with compile-time type safety. Also, we have control over the structure of the data as long as `getSelectedRegionById` returns either a `RegionSelectOption` type object or a falsy value (like null or undefined).

bnussman-akamai (approved):
One day, we may want to combine RegionSelect and RegionMultiSelect into one component, but this is good for now!

PR #10093, opened 2024-01-22
test: [M3-7483] - Add test to delete users on "Users & Grants" page

cliu-akamai:
## Description 📝
Add regression tests for deleting users on `Users & Grants` page.

## Major Changes 🔄
- Add regression tests to delete users on `Users & Grants` page.

## How to test 🧪
```
```

### DIFF
--- a/packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
+++ b/packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
@@ -10,6 +10,7 @@ import {
   mockGetUsers,
   mockUpdateUser,
   mockUpdateUserGrants,
+  mockDeleteUser,
 } from 'support/intercepts/account';
 import {
   mockAppendFeatureFlags,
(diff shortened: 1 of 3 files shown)


packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
```diff
@@ -904,4 +905,95 @@ describe('User permission management', () => {
...
+      });
+
+    // the "Confirm Deletion" dialog opens
+    ui.dialog.findByTitle('Confirm Deletion').within(() => {
+      ui.button.findByTitle('Delete').should('be.visible').click();
+    });
+    cy.wait(['@deleteUser', '@getUsers']);
+
+    // the user is deleted
+    cy.findByText(additionalUser.username).should('not.exist');
```
jdamore-linode, line 997:
```suggestion
    ui.toast.assertMessage(`User ${additionalUser.username} has been deleted successfully.`);
    cy.findByText(additionalUser.username).should('not.exist');
```
This wasn't a ticket requirement, but it would be good to have the toast notification covered too!

packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
```diff
@@ -904,4 +905,95 @@ describe('User permission management', () => {
...
+
+    // the "Confirm Deletion" dialog opens
+    ui.dialog.findByTitle('Confirm Deletion').within(() => {
+      ui.button.findByTitle('Cancel').should('be.visible').click();
+    });
+    // click the "Cancel" button will do nothing
+    cy.findByText(mockUser.username).should('be.visible');
+    cy.findByText(additionalUser.username).should('be.visible');
+
+    // clickl the "x" button will dismiss the dialog and do nothing
```
jdamore-linode, line 961:
```suggestion
    // clicking the "x" button will dismiss the dialog and do nothing
```

jdamore-linode (approved):
Great work @cliu-akamai!

packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
```diff
@@ -904,4 +905,95 @@ describe('User permission management', () => {
...
+          .should('be.visible')
+          .should('be.enabled')
+          .click();
+      });
+
+    // the "Confirm Deletion" dialog opens
+    ui.dialog.findByTitle('Confirm Deletion').within(() => {
+      ui.button.findByTitle('Cancel').should('be.visible').click();
+    });
+    // click the "Cancel" button will do nothing
```
mjac0bs, line 957:
```suggestion
    // clicking the "Cancel" button will dismiss the dialog and do nothing
```

packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
```diff
@@ -904,4 +905,95 @@ describe('User permission management', () => {
     // redirects to the new user's "User Permissions" page
     cy.url().should('endWith', `/users/${newUser.username}/permissions`);
   });
+
+  it('can delete users', () => {
```
mjac0bs, line 909: @jdamore-linode Do we need to mock the parentChildAccountSwitching feature flag to false here, as called out in the ticket AC? Wasn't sure if you had anything specific in mind, but I wouldn't expect any changes with Parent/Child and regular user deletion (unless we were a Child trying to delete a Proxy user, but that's covered in another ticket).

packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
mjac0bs: Agreed!

mjac0bs (approved):
Can we add a Test changeset with `yarn changeset` please?

Other than the couple of typos to fix and Joe's additional suggestion of confirming the toast, this passes as expected and covers the expected behavior - thank you!

image

packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
jdamore-linode: @mjac0bs Nice catch! I'd say if we don't expect this to be impacted then we can skip the feature flag mock 👍 we can add it later if we decide we need it, but I don't really see that happening

packages/manager/cypress/e2e/core/account/user-permissions.spec.ts
dwiley-akamai: I agree with this small adjustment; it'll make it align with the comment for the next assertion too.

dwiley-akamai (approved):
Test passes ✅
Code review ✅

PR #10094, opened 2024-01-22
test: Fix Domains landing page empty state test flake

jdamore-linode:
## Description 📝
This fixes a flaky test by changing a `should('not.be.visible')` assertion to `should('not.exist')`. This test passes most of the time because the `should('not.be.visible')` assertion passes during a brief moment when the drawer close animation occurs (i.e. the element is still present in the DOM, but not visible). However, every so often this assertion does not occur quickly enough and the element is no longer present in the DOM, and the assertion fails.

This was initially planned to be included in a larger PR containing more fixes, but that's taking longer than I had planned and there's no reason to hold this back in the meantime.

## Changes  🔄
- Fix Domains landing page empty state test flake

## How to test 🧪
We can rely on CI, but this test can also be run locally using this command. In its current state, the test seems to fail more often in CI than locally; I was only able to trigger a failure twice out of 100 runs.

```bash
yarn cy:run -s "cypress/e2e/core/domains/domains-empty-landing-page.spec.ts"
```

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

---

### DIFF
--- a/packages/manager/.changeset/pr-10094-tests-1705959873279.md
+++ b/packages/manager/.changeset/pr-10094-tests-1705959873279.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Tests
+---
+
+Improve Domains landing page empty state test flakiness ([#10094](https://github.com/linode/manager/pull/10094))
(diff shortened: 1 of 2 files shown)


hkhalil-akamai (approved):
Great find!

bnussman-akamai (approved):
Nice! Thank you!

I almost opened a PR making the same change, but changing `not.be.visible` to `not.exist` was going to be a complete guess from me. Thanks for explaining the difference

PR #10095, opened 2024-01-22
feat: [M3-7490] - Improve NodeBalancer Restricted User Experience

jaalah-akamai:
## Description 📝
To prevent unauthorized access to specific flows and provide clearer guidance, we aim to restrict entry to users without the required permissions while providing new notices. 

Tooltips specifically designed for disabled buttons will persist within action menus. These tooltips prove valuable in scenarios where a restricted user has permission to delete their personal resources but lacks the permission to delete those belonging to the admin user. In such cases, we refrain from displaying notices on the page.

For restricted users:
- Error Notices:
    - Drawers - Error notifications **_without_** an icon:
    - Create Pages - Error notifications **_with_** an icon
- Warning Notices w/ Icon:
   - Detail Pages
- Disabled Create Button
   - Empty State Landing Pages
   - Landing 

## Changes  🔄
- Added `useIsResourceRestricted` hook that verifies if a user with restricted access can edit specific cloud resources (Linodes, NodeBalancers, Volumes, etc.) Admin-created resources may be restricted, but those created by the restricted user are editable.
- Changed `entityNameMap` to `grantTypeMap` and added `as const` in order to generate a union of the mapped names to provide type-safety to `useIsResourceRestricted` and allow for capitalization/plurality in the names.
- Added `StyledTagButton` to replace our one-off usages of adding tags.
- Added `getRestrictedResourceText` utility to help generate resource-specific notices.
- Added `isRestrictedGlobalGrantType` util which can be used to throughout Cloud Manager (+15 occurrences we can update in the future)

## Preview 📷

>[!note]
We changed the banner slightly from the screenshots. The text will now say `You don't have permissions to ${action} ${resourceType}. Please contact your account administrator to request the necessary permissions.`

| Page  | Before | After |
| ------- | ------- | ------- |
| Empty State Landing |  Screenshot 2024-01-26 at 9 48 38 AM| Screenshot 2024-01-26 at 9 48 15 AM |
| Landing: Unable to delete admin NodeBalancer | Screenshot 2024-01-26 at 9 50 34 AM |Screenshot 2024-01-26 at 9 49 50 AM Screenshot 2024-01-26 at 9 43 44 AM|

| Page  | Before | After |
| ------- | ------- | ------- |
| Landing Screenshot 2024-01-23 at 9 05 46 PM | No Restrictions Screenshot 2024-01-23 at 8 43 22 PM |  No Notice, Can Add Linodes, Can Add NodeBalancers Screenshot 2024-01-23 at 8 45 26 PM |
| Create Page | cloud linode com_nodebalancers_create (1) | localhost_3000_nodebalancers_create|
| Configurations | Screenshot 2024-01-22 at 1 22 29 PM | Screenshot 2024-01-22 at 1 29 01 PM |
| Settings | Screenshot 2024-01-22 at 1 37 43 PM | Screenshot 2024-01-22 at 1 37 15 PM |
| Summary: Tags | Screenshot 2024-01-22 at 10 51 17 AM | Screenshot 2024-01-22 at 10 49 29 AM |

| Before | After |
| ------- | ------- |
| Screenshot 2024-01-22 at 10 09 07 AM | Screenshot 2024-01-22 at 10 08 53 AM |

### Prerequisites
- Log into two accounts side by side:
   - An unrestricted admin user account: full access
   - A restricted user account (use Incognito for this)
       - Start with `Read Only` for everything
       - After we'll be toggling permissions

### Reproduction steps
**Restricted User Perm: "Read Only"**
- Empty State Landing:
    - Start with no NodeBalancers and perms set to "Read Only" on restricted account
    - Observe as restricted user, you **_cannot_** create a NodeBalancer
    - Change perm to include "Can add Linodes" and "Can add NodeBalancers"
    - Observe as restricted user, you **_can_** create a NodeBalancer
    - Reset settings for restricted user back to "Read Only"
- Create:
    - As admin user, create a Nodebalancer 
    - Observe as restricted user, you shouldn't be able to reach this page except by changing URL
    - Observe that when changing URL, all fields and actions are disabled (including firewalls)
- Landing:  
    - Observe as restricted user, notice shows and you cannot delete or create NodeBalancers
- Detail:
    - Observe as restricted user, notice shows
    - Observe on Summary tab, you cannot create tags
    - Observe on Configurations tab, all fields and actions are disabled
    - Observe on Settings tab, all fields and actions are disabled

**Restricted User Perm: "Read Only" w/ Can add Linodes, Can add NodeBalancers**
- Create:
    - Observe as restricted user, you should be able to create a NodeBalancer
    - Observe that there's no notices
- Landing:  
    - Observe as restricted user, you should be able to create a NodeBalancer
    - Observe there's no notices
    - Observe you **_cannot_** delete the admin user's NodeBalancer
    - Observe you **_can_** delete your own NodeBalancer
- Detail:
    - Observe no notices show for your own NodeBalancer, but that they do for admin user's NodeBalancer
    - Observe you can edit your own NodeBalancer, but that you cannot edit the admin user's NodeBalancer

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/components/ActionMenu/ActionMenu.tsx
```diff
@@ -69,14 +70,15 @@ export const ActionMenu = React.memo((props: ActionMenuProps) => {
...
-    '& .MuiSvgIcon-root': {
-      fill: '#fff',
-      height: '20px',
-      width: '20px',
-    },
     '& :hover': {
       color: '#4d99f1',
     },
+    '&& .MuiSvgIcon-root': {
+      fill: theme.color.disabledText,
```
jaalah-akamai, line 77: We only display the tooltip for when things are disabled

packages/manager/src/components/Button/StyledTagButton.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import { Button } from './Button';
+
+/**
+ * A button for Tags. Eventually this treatment will go away,
+ * but the sake of the MUI migration we need to keep it around for now, and as a styled component in order to get rid of
+ * spreading excessive styles for everywhere this is used.
+ *
+ */
+export const StyledTagButton = styled(Button, {
```
jaalah-akamai, line 13: We can use this in other places as well, but I didn't want to expand the scope of this PR too much.

packages/manager/src/components/SelectFirewallPanel/SelectFirewallPanel.tsx
```diff
@@ -78,7 +81,7 @@ export const SelectFirewallPanel = (props: Props) => {
           value={selectedFirewall}
         />
         <StyledLinkButtonBox>
-          <LinkButton onClick={handleCreateFirewallClick}>
+          <LinkButton isDisabled={disabled} onClick={handleCreateFirewallClick}>
```
jaalah-akamai, line 84: To disable this link in create flows for restricted users, probably should've been passed in to begin with.

packages/manager/src/features/Account/constants.ts
```diff
@@ -1,2 +1,15 @@
 export const BUSINESS_PARTNER = 'business partner';
 export const ADMINISTRATOR = 'administrator';
+
+export const grantTypeMap = {
```
jaalah-akamai, line 4: This was originally located within `/src/features/Users/`, but I found it more fitting in its current location. I am flexible about relocating it. In either case, I included as const to generate a union of the newly mapped names called `type GrantTypeMap`, incorporating both capitalization and plurality.

packages/manager/src/features/Account/types.ts
```diff
@@ -0,0 +1,6 @@
+import { grantTypeMap } from 'src/features/Account/constants';
+
+// A useful type for getting the values of an object
+export type ObjectValues<T> = T[keyof T];
+
+export type GrantTypeMap = ObjectValues<typeof grantTypeMap>;
```
jaalah-akamai, line 6: See note above in constants file.

packages/manager/src/features/NodeBalancers/NodeBalancerDetail/NodeBalancerConfigurations.tsx
```diff
@@ -87,6 +88,7 @@ const StyledConfigsButton = styled(Button, {
 }));
 
 interface Props {
+  grants: Grants | undefined;
```
jaalah-akamai, line 91: Passing in `grants` due to the fact this is a class component and it was available from it's parent.

packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
```diff
@@ -123,7 +123,6 @@ export const CreateBucketDrawer = (props: Props) => {
         {isRestrictedUser && (
           <Notice
             data-qa-permissions-notice
-            important
```
jaalah-akamai, line 126: Don't want the icon version for drawers anymore.

packages/manager/src/features/Users/UserPermissionsEntitySection.tsx
```diff
@@ -21,19 +21,7 @@ import { createDisplayPage } from 'src/components/Paginate';
...
-  stackscript: 'StackScripts',
-  image: 'Images',
-  volume: 'Volumes',
-  nodebalancer: 'NodeBalancers',
-  domain: 'Domains',
-  longview: 'Longview Clients',
-  firewall: 'Firewalls',
-  database: 'Databases',
-  vpc: 'VPCs',
-};
```
jaalah-akamai, line 36: Moved to `/src/features/Account/constants`

bnussman-akamai:
I'm not a huge fan of how we force a restricted user to look at a big yellow banner on landing pages. I think it would be nice if we had a more subtile way to indicate they have read-only permissions for example.

For the load balancer empty state, should we disable the "Create" button with a tooltip instead of putting a yellow banner at the top?

I still feel like "Access is restricted" is too generic. I think it would be nice to let the user know they have "read only permission" or that they "do not have permission to create"... 

Those are just my opinions on the UX aspect of this, will continue to review regardless.

jaalah-akamai:
@bnussman-akamai I'll circle back to UX about the wording, but if you can review the functionality that would be awesome!

packages/manager/src/features/NodeBalancers/NodeBalancerCreate.tsx
```diff
@@ -126,8 +127,11 @@ const NodeBalancerCreate = () => {
   const theme = useTheme<Theme>();
   const matchesSmDown = useMediaQuery(theme.breakpoints.down('md'));
 
+  // NodeBalancer creation is possible for restricted accounts,
+  // contingent on the ability to add Linodes.
   const disabled =
-    Boolean(profile?.restricted) && !grants?.global.add_nodebalancers;
+    Boolean(profile?.restricted) &&
+    !(grants?.global.add_nodebalancers && grants?.global.add_linodes);
```
bnussman-akamai, line 134: I confirmed I was able to create a NodeBalancer as a restricted user **without** `add_linodes` permission. 

I understand why we might want to take  `add_linodes` into account for UX reasons, but I'd prefer if we just check `add_nodebalancers` and not  `add_linodes` so we are in parity with how the API works.


https://github.com/linode/manager/assets/115251059/d0cc6aca-afe8-40c6-ae3d-4165275d5f88

packages/manager/src/features/NodeBalancers/NodeBalancerDetail/NodeBalancerConfigurations.tsx
```diff
@@ -540,6 +545,18 @@ class NodeBalancerConfigurations extends React.Component<CombinedProps, State> {
     return true;
   };
 
+  isRestrictedUser = () => {
```
bnussman-akamai, line 548: This function name is a bit misleading. Might want to change it to something like

```suggestion
  isNodeBalancerReadOnly = () => {
```

packages/manager/src/hooks/useIsResourceRestricted.ts
```diff
@@ -0,0 +1,36 @@
...
+        (grant: Grant) => grant.id === id && grant.permissions === grantLevel
+      );
+      setIsRestricted(hasRestriction);
+    } else {
+      setIsRestricted(false);
+    }
+  }, [grants, id, grantType, grantLevel]);
+
+  return isRestricted;
+};
```
bnussman-akamai, line 36: I believe this can be written without needed a `useEffect` and state. I think this change works the same, but I didn't fully test it.

```suggestion
export const useIsResourceRestricted = ({
  grantLevel = 'read_only',
  grantType,
  id,
}: {
  grantLevel?: GrantLevel;
  grantType: GrantType;
  grants?: Grants;
  id: number | undefined;
}) => {
  const { data: grants } = useGrants();

  if (!grants) {
    return false;
  }

  return grants[grantType].some(
    (grant) => grant.id === id && grant.permissions === grantLevel
  );
}
```

packages/manager/src/hooks/useIsResourceRestricted.ts
```diff
@@ -0,0 +1,36 @@
...
+import { useGrants } from 'src/queries/profile';
+
+import type { Grant, GrantLevel, GrantType, Grants } from '@linode/api-v4';
+
+/**
+ * This hook verifies if a user with restricted access can edit specific cloud resources (Linodes, NodeBalancers, Volumes, etc.).
+ * Admin-created resources may be restricted, but those created by the restricted user are editable.
+ */
+export const useIsResourceRestricted = ({
+  grantLevel = 'read_only',
```
bnussman-akamai, line 12: I wonder if we should remove this default of `read_only`.  

When this hook is used, it looks a bit ambiguous as to what it does

```typescript
const isRestricted = useIsResourceRestricted({
   grantType: 'nodebalancer',
   id: nodeBalancerId,
});
```

I'm wondering if changing the hook's name and making `grantLevel` required would help.

```typescript
const isNodeBalancerReadOnly = useDoesUserHaveGrant({
   grantLevel: 'read_only',
   grantType: 'nodebalancer',
   id: nodeBalancerId,
});
```

packages/manager/src/features/Account/utils.ts
```diff
@@ -1,6 +1,20 @@
...
+interface GetRestrictedResourceText {
+  action?: ActionType;
+  resourceType: GrantTypeMap;
+}
+export const getRestrictedResourceText = ({
+  action = 'edit',
+  resourceType,
+}: GetRestrictedResourceText) => {
+  return `You don't have permissions to ${action} ${resourceType}. Please contact your account administrator to request the necessary permissions.`;
+};
```
bnussman-akamai, line 17: This is what I see as a restricted user with read-only permission on one of my NodeBalancers.

Screenshot 2024-01-26 at 10 50 48 AM

I think we'll need to change this message to say 

```
You don't have permissions to edit this NodeBalancer
``` 

instead of 

```
You don't have permissions to edit NodeBalancers.
```

bnussman-akamai (commented):
General functionality looks good! Have a few comments/questions

packages/manager/src/features/NodeBalancers/NodeBalancerCreate.tsx
hkhalil-akamai: +1 this. Any reason why the Add Linodes permission should be required for adding/modifying NodeBalancers?

packages/manager/src/components/Button/StyledTagButton.ts
abailly-akamai: 👍 that being said we should both add a unit test and add/modify a story if it is meant to stay

packages/manager/src/features/Account/constants.ts
abailly-akamai: nitpick: enum?

packages/manager/src/features/Account/types.ts
abailly-akamai: same comment with enum - I feel like it's a good usage for this type of case

packages/manager/src/features/NodeBalancers/NodeBalancerCreate.tsx
abailly-akamai: if that's the case we should document this in code more thouroughly

packages/manager/src/features/NodeBalancers/NodeBalancerCreate.tsx
```diff
@@ -518,6 +523,7 @@ const NodeBalancerCreate = () => {
               <Link to={FIREWALL_GET_STARTED_LINK}>Learn more</Link>.
             </Typography>
           }
+          disabled={disabled}
```
abailly-akamai, line 526: The intent is a bit unclear here. what's a "disabled" notice? If we mean the call to action, it should be more verbose. Also we can have several call to actions in a notice?

This should be added to our coverage as well once improved

packages/manager/src/features/NodeBalancers/NodeBalancerDetail/NodeBalancerConfigurations.tsx
```diff
@@ -628,6 +645,8 @@ class NodeBalancerConfigurations extends React.Component<CombinedProps, State> {
       ? parseInt(expandedConfigId, 10) === config.id
       : false;
 
+    const isRestricted = this.isRestrictedUser();
+
     const L = {
```
abailly-akamai, line 650: Not related to this PR, but `const L` 👎

packages/manager/src/features/NodeBalancers/NodeBalancersLanding/NodeBalancersLanding.tsx
```diff
@@ -35,6 +36,14 @@ export const NodeBalancersLanding = () => {
...
   const history = useHistory();
   const pagination = usePagination(1, preferenceKey);
+  const { data: grants } = useGrants();
+  const { data: profile } = useProfile();
+
+  // NodeBalancer creation is possible for restricted accounts,
+  // contingent on the ability to add Linodes.
+  const isRestricted =
+    Boolean(profile?.restricted) &&
+    !(grants?.global.add_nodebalancers && grants?.global.add_linodes);
```
abailly-akamai, line 46: Consider making the code more DRY with a util since we are reusing this logic

packages/manager/src/features/NodeBalancers/NodeBalancersLanding/NodeBalancersLandingEmptyState.tsx
```diff
@@ -7,9 +7,19 @@ import { DocumentTitleSegment } from 'src/components/DocumentTitle';
...
 export const NodeBalancerLandingEmptyState = () => {
   const history = useHistory();
+  const { data: grants } = useGrants();
+  const { data: profile } = useProfile();
+
+  // NodeBalancer creation is possible for restricted accounts,
+  // contingent on the ability to add Linodes.
+  const isRestricted =
+    Boolean(profile?.restricted) &&
+    !(grants?.global.add_nodebalancers && grants?.global.add_linodes);
```
abailly-akamai, line 22: and here

abailly-akamai (commented):
Nice work! Added comments to clean things up and some questions. I'll give this a more thorough review once code improvements from reviewers have been addressed 👍

packages/manager/src/hooks/useIsResourceRestricted.ts
jaalah-akamai: Yea it depends on if we want to react to grants changing or not, let me test 👍

packages/manager/src/hooks/useIsResourceRestricted.ts
jaalah-akamai: I agree with `grantLevel`, I'll think about the name. I almost like a blend of the two:

packages/manager/src/features/NodeBalancers/NodeBalancerCreate.tsx
jaalah-akamai: Ahh yea I was under the impression that linodes with read_only perms couldn't be added to NodeBalancers, I guess that's not the case.

packages/manager/src/features/Account/utils.ts
jaalah-akamai: Yup good catch 👁️

packages/manager/src/features/NodeBalancers/NodeBalancerDetail/NodeBalancerConfigurations.tsx
bnussman-akamai: ramda 😡

packages/manager/src/features/Account/utils.ts
```diff
@@ -1,6 +1,38 @@
...
+  globalGrantType,
+  grants,
+  profile,
+}: {
+  globalGrantType: GlobalGrantTypes;
+  grants: Grants | undefined;
+  profile: Profile | undefined;
+}) => {
+  return Boolean(profile?.restricted) && !grants?.global[globalGrantType];
+};
```
abailly-akamai, line 35: I know it's clear from the name however 
1. can we add return types here?
2. i think it's always super useful to complement those utils with JSDocs

abailly-akamai (approved):
Reads so much better, thanks for the changes!

Feature behaves as described in the PR and looks great from my vantage point ✅

jaalah-akamai:
> Agree with this new component and thanks for creating a test/story as per @abailly-akamai. Should we open a ticket for the other places?

PR #10083, opened 2024-01-19
upcoming: [M3-7647] - Update logic for child_account PAT

jaalah-akamai:
## Description 📝
Used our approach from https://github.com/linode/manager/pull/10024 to cleanup Parent/Child scopes.

Nothing should change in terms of what we see from the original PR: https://github.com/linode/manager/pull/9992

## Changes  🔄
- Changed `getPermsNameMap` to `filterPermsNameMap` and now it accepts an array of perms
- Enabled `child_account` perms and removed our old approach
- Update tests

### Prerequisites
(How to setup test environment)
- Check out this PR and `yarn dev`.
- In the dev tools:
   - Make sure the Parent/Child Account Switching feature flag is on.
   - Turn mocks on.

### Verification steps 
(How to verify changes)
- Go to http://localhost:3000/profile/tokens and click the Create a Personal Access Token button.
- Observe that "Child Account Access" is visible as a new scope and all scopes default to None.
- On the tokens landing page, click the View Scope button for any mock PAT.
- Observe that the "Child Account Access" scope is visible in the table.
- Toggle the feature flag off and observe that the Child Account Access token is not visible in the Create or View drawers and all scopes default to Read/Write.
- Toggle the feature flag back on.
- Go to `serverHandler.ts` and edit the following request to change the user_type to `null`, `child`, or `proxy`:
```
  rest.get('*/account/users/:user', (req, res, ctx) => {
    // Parent/Child: switch the `user_type` depending on what account view you need to mock.
    return res(ctx.json(accountUserFactory.build({ user_type: 'parent' })));
  }),
```
- Observe that the Child Account Access token is not visible in the Create or View drawers.
- Verify tests pass:
```
yarn test utils CreateAPITokenDrawer ViewAPITokenDrawer

```

## As an Author I have considered 🤔

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ❔ Our [contribution guidelines](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md)
- [ ] 🤏 Splitting feature into small PRs
- [x] ➕ Adding a changeset
- [x] 🧪 Providing/Improving test coverage
- [x] 🔐 Removing all sensitive information from the code and PR description
- [x] 🚩 Using a feature flag to protect the release
- [x] 👣 Providing comprehensive reproduction steps
- [ ] 📑 Providing or updating our documentation
- [ ] 🕛 Scheduling a pair reviewing session
- [ ] 📱 Providing mobile support
[...]

PR #10102, opened 2024-01-24
upcoming: [M3-7660] - Cleanup files to use `profile` to get `user_type`

jaalah-akamai:
## Description 📝
Now that `user_type` is returned from the `/account/profile` endpoint, we can save ourselves some query calls.

## Changes  🔄
- Updated tests and files that didn't need to call `/account/users/`

## Preview 📷
Shouldn't be any visual changes

### Prerequisites
- Run `yarn test` ensure all tests pass
- Run `cyrun cypress/e2e/core/account` ensure all tests pass

### Verification steps 
- Verify user menu still works as intended as with MSW
- Verify we can see `Child Account Access` by [following steps here](https://github.com/linode/manager/pull/10083)

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/Account/AccountLanding.tsx
+++ b/packages/manager/src/features/Account/AccountLanding.tsx
@@ -54,6 +54,8 @@ const AccountLanding = () => {
   const accountAccessGrant = grants?.global?.account_access;
   const readOnlyAccountAccess = accountAccessGrant === 'read_only';
   const isAkamaiAccount = account?.billing_source === 'akamai';
+  const isProxyUser = profile?.user_type === 'proxy';
+  const isParentUser = profile?.user_type === 'parent';
 
   const tabs = [
     {
@@ -115,8 +117,7 @@ const AccountLanding = () => {
 
   const isBillingTabSelected = location.pathname.match(/billing/);
   const canSwitchBetweenParentOrProxyAccount =
-    flags.parentChildAccountAccess &&
-    (profile?.user_type === 'parent' || profile?.user_type === 'proxy');
+    flags.parentChildAccountAccess && (isParentUser || isProxyUser);
 
   const landingHeaderProps: LandingHeaderProps = {
     breadcrumbProps: {
@@ -172,7 +173,7 @@ const AccountLanding = () => {
         </React.Suspense>
       </Tabs>
       <SwitchAccountDrawer
-        isProxyUser={profile?.user_type === 'proxy'}
+        isProxyUser={isProxyUser}
         onClose={() => setIsDrawerOpen(false)}
         open={isDrawerOpen}
       />
--- a/packages/manager/src/features/Account/AccountLanding.tsx
+++ b/packages/manager/src/features/Account/AccountLanding.tsx
@@ -13,7 +13,6 @@ import { TabPanels } from 'src/components/Tabs/TabPanels';
 import { Tabs } from 'src/components/Tabs/Tabs';
 import { useFlags } from 'src/hooks/useFlags';
 import { useAccount } from 'src/queries/account';
-import { useAccountUser } from 'src/queries/accountUsers';
 import { useGrants, useProfile } from 'src/queries/profile';
 
 import AccountLogins from './AccountLogins';
@@ -48,7 +47,6 @@ const AccountLanding = () => {
   const { data: account } = useAccount();
   const { data: grants } = useGrants();
   const { data: profile } = useProfile();
-  const { data: user } = useAccountUser(profile?.username ?? '');
 
   const flags = useFlags();
   const [isDrawerOpen, setIsDrawerOpen] = React.useState<boolean>(false);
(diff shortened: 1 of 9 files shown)


packages/manager/src/features/Account/SwitchAccountDrawer.test.tsx
```diff
@@ -33,7 +33,7 @@ describe('SwitchAccountDrawer', () => {
   it('should include a link to switch back to the parent account if the active user is a proxy user', async () => {
     server.use(
       rest.get('*/account/users/*', (req, res, ctx) => {
-        return res(ctx.json(accountUserFactory.build({ user_type: 'proxy' })));
+        return res(ctx.json(profileFactory.build({ user_type: 'proxy' })));
```
bnussman-akamai, line 36: I believe this should be reverted. We want to make sure we respond with `accountUserFactory` data for the `*/account/users/*` endpoint

bnussman-akamai (changes requested):

packages/manager/src/features/Account/SwitchAccountDrawer.test.tsx
jaalah-akamai: Actually I had to update the endpoint, but good catch

packages/manager/src/features/Account/SwitchAccounts/ChildAccountList.test.tsx
```diff
@@ -18,7 +18,7 @@ const props = {
 it('should display a list of child accounts', async () => {
   server.use(
     rest.get('*/account/users/*', (req, res, ctx) => {
-      return res(ctx.json(accountUserFactory.build({ user_type: 'parent' })));
+      return res(ctx.json(profileFactory.build({ user_type: 'parent' })));
```
bnussman-akamai, line 21: Same thing here!

packages/manager/src/features/Account/AccountLanding.tsx
```diff
@@ -174,7 +172,7 @@ const AccountLanding = () => {
         </React.Suspense>
       </Tabs>
       <SwitchAccountDrawer
-        isProxyUser={user?.user_type === 'proxy'}
+        isProxyUser={profile?.user_type === 'proxy'}
```
cpathipa, line 175: nitpicky :Can we move the check `profile?.user_type === 'proxy'` to a constant and reference it, since we are using it in couple of places?

packages/manager/src/features/TopMenu/UserMenu/UserMenu.tsx
```diff
@@ -67,13 +65,13 @@ export const UserMenu = React.memo(() => {
   const hasAccountAccess = !isRestrictedUser || hasGrant('account_access');
   const hasReadWriteAccountAccess = hasGrant('account_access') === 'read_write';
   const hasParentChildAccountAccess = Boolean(flags.parentChildAccountAccess);
-  const isParentUser = user?.user_type === 'parent';
-  const isProxyUser = user?.user_type === 'proxy';
+  const isParentUser = profile?.user_type === 'parent';
+  const isProxyUser = profile?.user_type === 'proxy';
```
cpathipa, line 69: Looks like we have similar checks for user_type at the feature level. Any thoughts on having utility functions for user types? 

Something like this
`export const isParentUser = (profile) => profile?.user_type === 'parent';`
`export const isProxyUser = (profile) => profile?.user_type === 'proxy';`

jaalah-akamai: Moving it to a const was a good idea in that other file - I think we have too many utils and this is pretty clear as to what it's doing. I infer the type from the other user types if there were `child`

packages/manager/src/features/Account/SwitchAccountDrawer.test.tsx
```diff
@@ -32,8 +32,8 @@ describe('SwitchAccountDrawer', () => {
 
   it('should include a link to switch back to the parent account if the active user is a proxy user', async () => {
     server.use(
-      rest.get('*/account/users/*', (req, res, ctx) => {
-        return res(ctx.json(accountUserFactory.build({ user_type: 'proxy' })));
+      rest.get('*/account/profile/*', (req, res, ctx) => {
```
bnussman-akamai, line 35:
```suggestion
      rest.get('*/profile', (req, res, ctx) => {
```

bnussman-akamai (changes requested):
One more change and should be good to go

jaalah-akamai:
Weird that those don't cause tests to fail 🤔

mjac0bs (approved):
Pushed a fix for `ChildAccountList.test.tsx` that was *not* failing when I ran it locally, but was failing in CI. Unit tests are passing as of that last run.

LGTM now. Confirmed that the user menu and Switch Account button (in both menu and Account Landing) is only visible for the expected parent and proxy users. Confirmed that the Child Account Access PAT is visible only for parent users and that there are no regressions when creating a PAT.

bnussman-akamai (approved):

jaalah-akamai:
✅  Confirmed E2E tests pass with latest develop changes

PR #10110, opened 2024-01-25
test: [M3-7516] - Cypress tests for Parent → Child account switching flows

jdamore-linode:
## Description 📝
Adds Cypress integration tests to cover Parent/Child flows involving account switching from a Parent account to a Child account.

## Changes  🔄
- Adds tests to cover Parent/Child account switching flows:
  - Switch from Parent to Child using Account Billing page
  - Switch from Parent to Child using user menu
  - Error flows upon fetching child accounts and switching accounts
- Adds various utilities to support tests
  - Intercept/mock utils for child account API endpoints
  - Util to assert local storage values

## How to test 🧪
We can rely on automated tests for this, but you can also run the test locally:

### Prerequisites
Build and serve Cloud:

```bash
yarn && yarn build && yarn start:manager:ci
```

### Verification steps 

```bash
yarn cy:run -s "cypress/e2e/core/parentChild/account-switching.spec.ts"
```

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/cypress/e2e/core/parentChild/account-switching.spec.ts
+++ b/packages/manager/cypress/e2e/core/parentChild/account-switching.spec.ts
@@ -1,0 +1,137 @@
+import {
+  accountFactory,
+  appTokenFactory,
+  profileFactory,
+} from '@src/factories';
+import { accountUserFactory } from '@src/factories/accountUsers';
+import { DateTime } from 'luxon';
+import {
+  mockCreateChildAccountToken,
+  mockGetAccount,
+  mockGetChildAccounts,
+  mockGetUser,
+} from 'support/intercepts/account';
+import {
+  mockAppendFeatureFlags,
+  mockGetFeatureFlagClientstream,
+} from 'support/intercepts/feature-flags';
+import { mockAllApiRequests } from 'support/intercepts/general';
+import { mockGetProfile } from 'support/intercepts/profile';
+import { ui } from 'support/ui';
+import { makeFeatureFlagData } from 'support/util/feature-flags';
+import { assertLocalStorageValue } from 'support/util/local-storage';
+import { randomLabel, randomNumber, randomString } from 'support/util/random';
+
+describe('Parent/Child account switching', () => {
+  it('can switch from Parent account to Child account', () => {
+    const mockParentAccount = accountFactory.build({
+      company: 'Parent Company',
+    });
+    const mockParentProfile = profileFactory.build({
+      username: randomLabel(),
+      user_type: 'parent',
+    });
+    const mockParentUser = accountUserFactory.build({
+      username: mockParentProfile.username,
+      user_type: 'parent',
+    });
+
+    const mockChildAccount = accountFactory.build({
+      company: 'Child Company',
+    });
+
+    const mockChildAccountToken = appTokenFactory.build({
+      id: randomNumber(),
+      created: DateTime.now().toISO(),
+      expiry: DateTime.now().plus({ hours: 1 }).toISO(),
+      label: `${mockChildAccount.company}_proxy`,
+      scopes: '*',
+      token: randomString(32),
+      website: undefined,
+      thumbnail_url: undefined,
+    });
+
+    mockAppendFeatureFlags({
+      parentChildAccountAccess: makeFeatureFlagData(true),
+    });
+    mockGetFeatureFlagClientstream();
+    mockGetProfile(mockParentProfile);
+    mockGetAccount(mockParentAccount);
+    mockGetChildAccounts([mockChildAccount]);
+    mockGetUser(mockParentUser);
+
+    cy.visitWithLogin('/');
+
+    // Confirm that Parent account username and company name are shown in user
+    // menu button, then click the button.
+    ui.userMenuButton
+      .find()
+      .should('be.visible')
+      .within(() => {
+        cy.findByText(mockParentProfile.username).should('be.visible');
+        cy.findByText(mockParentAccount.company).should('be.visible');
+      })
+      .click();
+
+    // Click "Switch Account" button in user menu.
+    ui.userMenu
+      .find()
+      .should('be.visible')
+      .within(() => {
+        ui.button
+          .findByTitle('Switch Account')
+          .should('be.visible')
+          .should('be.enabled')
+          .click();
+      });
+
+    // Click mock company name in "Switch Account" drawer.
+    mockCreateChildAccountToken(mockChildAccount, mockChildAccountToken).as(
+      'switchAccount'
+    );
+    ui.drawer
+      .findByTitle('Switch Account')
+      .should('be.visible')
+      .within(() => {
+        cy.findByText(mockChildAccount.company).should('be.visible').click();
+      });
+
+    cy.wait('@switchAccount');
+
+    // Confirm that Cloud Manager updates local storage authentication values.
+    assertLocalStorageValue(
+      'authentication/token',
+      mockChildAccountToken.token
+    );
+    assertLocalStorageValue(
+      'authentication/expire',
+      mockChildAccountToken.expiry
+    );
+    assertLocalStorageValue(
+      'authentication/scopes',
+      mockChildAccountToken.scopes
+    );
+
+    // From this point forward, we will not have a valid test account token stored in local storage,
+    // so all non-intercepted API requests will respond with a 401 status code and we will get booted to login.
+    // We'll mitigate this by broadly mocking ALL API-v4 requests, then applying more specific mocks to the
+    // individual requests as needed.
+    mockAllApiRequests();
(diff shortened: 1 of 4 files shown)


packages/manager/cypress/e2e/core/parentChild/account-switching.spec.ts
```diff
@@ -0,0 +1,289 @@
...
+      // We'll mitigate this by broadly mocking ALL API-v4 requests, then applying more specific mocks to the
+      // individual requests as needed.
+      mockAllApiRequests();
+      mockGetAccount(mockChildAccount);
+      mockGetProfile(mockParentProfile);
+      mockGetUser(mockParentUser);
+
+      // TODO Remove the call to `cy.reload()` once Cloud Manager automatically updates itself upon account switching.
+      // TODO Add assertions for toast upon account switch.
+      cy.reload();
```
jdamore-linode, line 157: cc @jaalah-akamai @mjac0bs -- I think this relates to this [discussion in the account switching PR](https://github.com/linode/manager/pull/10064#discussion_r1456565849).

Cloud definitely isn't responding as expected upon updating the Local Storage auth values, so a page reload is currently necessary on the Cypress (e.g. user) side to trigger the change. I'm guessing if we want to avoid an automatic page reload, it'll involve at least resetting Cloud's Axios interceptors, resetting React Query, and maybe doing something with our Redux state?

Either way, this test can be used to develop/troubleshoot this functionality without having to wait for anything from the API -- just remove these calls to `cy.reload()`

packages/manager/cypress/e2e/core/parentChild/account-switching.spec.ts
```diff
@@ -0,0 +1,289 @@
...
+ */
+const assertAuthLocalStorage = (
+  token: string,
+  expiry: string,
+  scopes: string
+) => {
+  assertLocalStorageValue('authentication/token', token);
+  assertLocalStorageValue('authentication/expire', expiry);
+  assertLocalStorageValue('authentication/scopes', scopes);
+};
```
jdamore-linode, line 60: Question for @jaalah-akamai or @mjac0bs: does the token value returned by the API already include the `Bearer ` prefix, or do we have logic somewhere that prepends it before saving it in Local Storage?

mjac0bs: I don't believe it includes the `Bearer` prefix. It looks like the token first gets set when a [session begins](https://github.com/linode/manager/blob/develop/packages/manager/src/layouts/OAuth.tsx#L97-L105) and `Bearer` comes from [`token_type`](https://github.com/linode/manager/blob/develop/packages/manager/src/layouts/OAuth.tsx#L133-L139), which is how it ends up with `Bearer` in local storage.

And from the API spec:

image


On a related bearer-in-token-related note, I'm not sure that we're supposed to have `Bearer` when calling `getChildAccountPersonalAccessToken` [here](https://github.com/linode/manager/blob/develop/packages/manager/src/features/Account/SwitchAccountDrawer.tsx#L65) in the switch account drawer. Maybe it was a oversight or I'm missing something and @jaalah-akamai can clarify.

packages/manager/cypress/e2e/core/parentChild/account-switching.spec.ts
```diff
@@ -0,0 +1,289 @@
...
+      );
+
+      // From this point forward, we will not have a valid test account token stored in local storage,
+      // so all non-intercepted API requests will respond with a 401 status code and we will get booted to login.
+      // We'll mitigate this by broadly mocking ALL API-v4 requests, then applying more specific mocks to the
+      // individual requests as needed.
+      mockAllApiRequests();
+      mockGetAccount(mockChildAccount);
+      mockGetProfile(mockParentProfile);
+      mockGetUser(mockParentUser);
```
jdamore-linode, line 153: Also hoping for an extra set of eyes here -- do these mocks accurately reflect what the account, profile, and user API request will look like in this situation?

packages/manager/cypress/e2e/core/parentChild/account-switching.spec.ts
jaalah-akamai: > On a related bearer-in-token-related note, I'm not sure that we're supposed to have Bearer when calling getChildAccountPersonalAccessToken [here](https://github.com/linode/manager/blob/develop/packages/manager/src/features/Account/SwitchAccountDrawer.tsx#L65) in the switch account drawer. Maybe it was a oversight or I'm missing something and @jaalah-akamai can clarify.

This is true - left over probably from one of my iterations ✅

packages/manager/cypress/e2e/core/parentChild/account-switching.spec.ts
```diff
@@ -0,0 +1,291 @@
...
+
+      cy.wait('@switchAccount');
+
+      // Confirm that Cloud Manager updates local storage authentication values.
+      // Satisfy TypeScript using non-null assertions since we know what the mock data contains.
+      assertAuthLocalStorage(
+        mockChildAccountToken.token!,
+        mockChildAccountToken.expiry!,
+        mockChildAccountToken.scopes
+      );
```
jaalah-akamai, line 145: We also want to ensure that parent tokens are being created here too

jaalah-akamai (approved):
This looks good overall - nice tests 🙏

packages/manager/cypress/support/util/local-storage.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * @file Utilities to access and validate Local Storage data.
+ */
+
+/**
+ * Asserts that a local storage item has a given value.
+ *
+ * @param key - Local storage item key.
+ * @param value - Local stroage item value to assert.
```
dwiley-akamai, line 9:
```suggestion
 * @param value - Local storage item value to assert.
```

dwiley-akamai (approved):
Remote and local tests passed ✅
Changeset ✅

packages/manager/cypress/e2e/core/parentChild/account-switching.spec.ts
mjac0bs: I think that mockGetUser is now left over from when we previously had to call `*/users/:user` to retrieve the `user_type` and it can now be removed.

Otherwise, this seems correct. We'll have the child account once logged in as a proxy user and we'll also need to fetch the parent's username from `*/profile`.

packages/manager/cypress/e2e/core/parentChild/account-switching.spec.ts
```diff
@@ -0,0 +1,289 @@
...
+  thumbnail_url: undefined,
+});
+
+const mockErrorMessage = 'An unknown error has occurred.';
+
+describe('Parent/Child account switching', () => {
+  /*
+   * Tests to confirm that Parent account users can switch to Child accounts as expected.
+   */
+  describe('From Parent to Child', () => {
```
mjac0bs, line 97: I'm wondering here if it would be best for consistency to rename mentions of "child" to "proxy" in places like this throughout these tests because if we're being literal, a parent user can't switch to a `child` account, they switch to a `proxy` account which is their allocated user on the child's account.
```suggestion
  describe('From Parent to Proxy', () => {
```

packages/manager/cypress/e2e/core/parentChild/account-switching.spec.ts
mjac0bs: Good point, the Switch Account drawer does that [here](https://github.com/linode/manager/pull/10064/files#diff-ef77afc013835684663f6da8ba90d1beda21ccc418965d2195fbec75c7b3bea3R107-R116).

mjac0bs (approved):
Thanks, Joe - looks good overall.

Left a comment about removing some extra mockGetUser calls now that #10102 is merged and a thought about naming.

packages/manager/cypress/e2e/core/parentChild/account-switching.spec.ts
jdamore-linode: Thanks @jaalah-akamai and @mjac0bs -- sorry for the slow follow up here.

If I'm understanding your explanation @mjac0bs, it sounds like we're supposed to have the `Bearer ` prefix in local storage but the token that gets stored during the account switching flow does not contain the prefix.

Is there a ticket for this? I can update this test to expect the `Bearer ` prefix but it will fail in the meantime.

packages/manager/cypress/e2e/core/parentChild/account-switching.spec.ts
jdamore-linode: Going to hold off on this for now -- the Cypress tests deal with a real token, so asserting its value will cause it to be logged and recorded.

PR #10122, opened 2024-01-29
change: [M3-7540] - Improve tags experience

hkhalil-akamai:
## Description 📝
**Update: it appears that the React 18 upgrade resolved the underlying bug! 🪄 Keeping this PR open without the workaround since it includes some improvements to the tag UX.**

~Fixes a bug in the `TagCell` component's overflow detection logic that caused it to sometimes render incorrectly. This bug only appeared in production builds of Cloud Manager and was caused by a parent `Grid` component briefly rendering as `display: none`, breaking the overflow detection in `TagCell`.~

~To fix this issue, this PR introduces a somewhat convoluted workaround: upon mount, `TagCell` locates its nearest ancestor in the DOM with `display: none` and then attaches a [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) (via the newly introduced `useMutationObserver` hook) to that element, re-computing overflow when it is finally displayed.~

This PR also includes improvements to the display of the `TagCell` component and `AddTag` input on small screens.

## Changes  🔄
* ~Fix overflow detection using `useMutationObserver` on parent element~
* Increase width of `AddTag` input on small screens
* Always render "Add a tag" and ellipses buttons to the right of tags
* Upgrade deprecated `Select` component to `Autocomplete` (this also fixes a momentary flicker when adding a tag)
* Replace `TagsPanel` component with `TagCell` (these components shared a lot of logic)

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Not detecting overflow Screenshot 2024-01-29 at 4 55 00 PM | Correctly detecting overflow Screenshot 2024-01-29 at 4 55 16 PM |
| `AddTag` unusably small on small displays: Screenshot 2024-01-29 at 4 56 21 PM | `AddTag` takes up entire width: Screenshot 2024-01-29 at 4 56 26 PM |
| "Add a tag" and ellipses buttons appear to the left of existing tags on small displays: Screenshot 2024-01-29 at 4 55 56 PM | Buttons now appear to the right of new tags, consistent with the "shadow" effect: Screenshot 2024-01-29 at 4 56 11 PM |

### Prerequisites
As far as I am aware, this bug is only reproducible in production builds of Cloud Manager. For reproducing or testing, use either:

* https://cloud.linode.com / [Preview link for this PR](https://linode-cloud-pr-10122.website-us-east-1.linodeobjects.com/)
* `yarn build && yarn workspace linode-manager serve build`

### Reproduction steps
(For overflow detection bug)
- Add several tags to a Linode until they overflow in the Linode details footer
- Navigate back to the Linodes landing page (*/linodes*)
- Refresh the page
- Return to the details page for the Linode
- Observe that overflow detection fails, causing tags to clip instead of displaying the "shadow" effect and ellipses button

(For style changes on small screens)
- In the details page for a Linode with overflowing tags, reduce the window size until the mobile layout is activated
- Observe that the "Add a tag" and ellipses buttons appear to the left of the tags, inconsistent with how they're displayed on larger screens and unintuitive with the "shadow" UX metaphor indicating an overflow
- Click on the the "Add a tag" button
- Observe the tag input field is unusably small, with placeholder and tag labels clipped

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/components/TagCell/TagCell.tsx
```diff
@@ -37,21 +39,33 @@ const checkOverflow = (el: any) => {
...
-    [tags]
-  );
+  const [elRef, setElRef] = React.useState<HTMLDivElement | null>(null);
+
+  // In production, a parent element sometimes is briefly
+  // set to display: none, breaking overflow detection.
+  // On mount, find the parent and listen for changes.
+  const renderParent = React.useMemo(() => {
+    let parent: HTMLElement | null = elRef;
+    while (parent != null && parent.style.display != 'none') {
```
hkhalil-akamai, line 51: While testing in devtools, I discovered that an MUI Grid component 3 levels above the tags container was briefly rendering as "display: none" causing the overflow detection to break. I am still not sure why this occurs and strangely still, only in production builds.

packages/manager/src/components/TagCell/TagCell.tsx
```diff
@@ -37,21 +39,33 @@ const checkOverflow = (el: any) => {
...
+
+  // Dependencies are not used explicitly for overflow detection but
+  // changes in these values may indicate the overflow state has changed.
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const hasOverflow = React.useMemo(() => !!elRef && checkOverflow(elRef), [
+    windowDimensions,
+    lastMutations,
+    tags,
+    elRef,
+  ]);
```
hkhalil-akamai, line 68: This is a complicated interaction but can be summarized as:

1. Locates the nearest parent element with `display: none`
2. Attaches a [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) via `useMutationObserver` that listens to changes to the parent element's style attribute
3. Recomputes `hasOverflow` whenever:
   1. `windowDimensions` changes (this was not present previously but is necessary since sometimes resizing the window causes tags to overflow even if no new tags are added)
   2. The parent element is mutated (i.e., style attribute changes; we are particularly interested in the display `display` property)
   3. Tags are added or deleted

packages/manager/src/features/Linodes/LinodeEntityDetailFooter.tsx
```diff
@@ -153,12 +153,7 @@ export const LinodeEntityDetailFooter = React.memo((props: FooterProps) => {
       >
         <TagCell
           sx={{
-            [theme.breakpoints.down('lg')]: {
-              '& > button': {
-                marginRight: theme.spacing(0.5),
-              },
-              flexDirection: 'row-reverse',
```
hkhalil-akamai, line 160: I'm not sure why these styles were included before specifically for small screens, but it is arguably more intuitive to render the ellipses button next to the shadow effect when tags are overflowing, and the "Add a tag button" at the very right, in a LTR context.

|Before|After|
|------|-----|
| Screenshot 2024-01-29 at 4 55 56 PM | Screenshot 2024-01-29 at 4 56 11 PM |

packages/manager/src/features/Linodes/LinodeEntityDetailFooter.tsx
```diff
@@ -153,12 +153,7 @@ export const LinodeEntityDetailFooter = React.memo((props: FooterProps) => {
...
-            },
+            width: '100%',
```
hkhalil-akamai, line 156: |Before|After|
|------|-----|
| Screenshot 2024-01-29 at 4 56 21 PM | Screenshot 2024-01-29 at 4 56 26 PM |

packages/manager/src/components/TagCell/TagCell.tsx
hana-linode: Agreed. Do we know which parent Grid component is briefly rendering `display: none`? I think if we can address the parent component issue, we can avoid this logic.

jaalah-akamai: I agree with the sentiment here. In addition, I wonder if the button shrinking to two lines compounds the issue: can we add a `flexShrink: 0` to [StyledTagButton](https://github.com/linode/manager/blob/develop/packages/manager/src/components/Button/StyledTagButton.ts#L16)?

hkhalil-akamai:
@abailly-akamai @hana-linode @jaalah-akamai

In a happy turn of events, it appears that @bnussman-akamai's work to upgrade to React 18 has resolved the underlying bug causing the overflow detection to fail 🎉. I am no longer able to reproduce the bug in develop under all previous conditions that I was aware of.

However, I'm keeping this PR open (but removing the workaround code) since it includes some improvements to the tags experience:

* Fixing the broken display of `AddTag` in small screens
* Listening to changes in window size, which can cause the tags to overflow
* Replacing the deprecated `Select` component (which was causing some flickering) with `Autocomplete`
* Tweaked the interaction when deleting tags (previously, all tags would be set to disabled; now only the tag that is being deleted is set to disabled)

bnussman-akamai (commented):
Nice clean up! I'm seeing only two unexpected things

- In the tag drawer, the Autocomplete is opening randomly after I add a tag
- Somehow I'm able to make a tag look more transparent than the others

https://github.com/linode/manager/assets/115251059/d4c1391c-00c2-43b1-b57a-877825b18ec9

abailly-akamai (commented):
Thanks for tackling this, this is a bit of a UX/UI mine field.

Here are a few things I noticed (pretty sure 2 & 3 are already happening without your refactor but since your PR mentions improving the experience I thought I would point them out).

1. The drawer uses `TagPanel`, which uses the legacy select. It results in a different user experience (ex: after adding a tag, it still displayes in the select which is odd, among other thins) - we should consolidate and use Autocomplete there as well

Screenshot 2024-02-29 at 12 12 19

https://github.com/linode/manager/assets/130582365/194ef2c7-af4a-4541-be3a-c3aebfef67a7

2. At a certain breakpoint (witnessed this around 1280), the tag select shifts around to the other side of the page, which is pretty jarring

https://github.com/linode/manager/assets/130582365/38a1b643-a770-446b-899f-fbcc9c7a2b50



3. With a tag panel in the sidebar ex: /nodebalancers/{id}
the select overflows the sidebar when lowering the horizontal viewport
Screenshot 2024-02-29 at 12 02 37

hkhalil-akamai:
> * In the tag drawer, the Autocomplete is opening randomly after I add a tag

Ah, good catch. Seems like this was caused by differences in how Chrome vs Safari trigger onblur.

> * Somehow I'm able to make a tag look more transparent than the others

This is a new indication for a tag that is in progress of being deleted. Since multiple tags could be deleted at the same time, sometimes this would cause a race condition leading to inconsistent states. Fixed with a new `useAtomic` hook.

hkhalil-akamai:
@abailly-akamai Thanks for the feedback. Took it a step further and completely removed `TagsPanel` in favor of `TagCell`, since I noticed most of the logic was being duplicated ([`aa3c9`](https://github.com/linode/manager/pull/10122/commits/aa3c9cd79cf52ea070756ed48ae471bce3ec1cb6)).

packages/manager/src/components/Button/StyledTagButton.ts
```diff
@@ -12,9 +12,14 @@ import { Button } from './Button';
  */
 export const StyledTagButton = styled(Button, {
   label: 'StyledTagButton',
-})(({ theme, ...props }) => ({
+})<{ panel?: boolean }>(({ theme, ...props }) => ({
```
abailly-akamai, line 15: You need to prevent this prop from being forwarded to the dom

Screenshot 2024-03-05 at 09 16 53

packages/manager/src/components/TagCell/TagCell.tsx
```diff
@@ -34,122 +36,132 @@ const checkOverflow = (el: any) => {
...
 
-  const handleAddTag = async (tag: string) => {
-    await updateTags([...tags, tag]);
-  };
+  const windowDimensions = useWindowDimensions();
+
+  const [hasOverflow, setHasOverflow] = React.useState(false);
+  React.useLayoutEffect(() => {
+    setHasOverflow(!!elRef && checkOverflow(elRef));
+  }, [windowDimensions, tags, elRef]);
```
abailly-akamai, line 53: What are you solving with this? 

1. This feels like it can be solved with CSS alone (happy to take a look with you)
2. If this _has_ to stay:
  - This seems expensive, especially without debouncing
  - This is lacking comments

packages/manager/src/components/TagCell/TagCell.tsx
```diff
@@ -34,122 +36,132 @@ const checkOverflow = (el: any) => {
...
+
+  const updateTagsAtomic = useAtomic(tags, props.updateTags);
```
abailly-akamai, line 55: I am not sold on this:

- The intent is not super clear. Why are you implementing this. Race conditions on deleting tags quickly?
- This feels overly complex. The old UI was disabling all tags (no other tags can be interacted with) while deleting one. This is sufficient and removes the need for introducing a complex util
- `useAtomic` is an odd naming convention that seems to wrongly correlate to jotai

abailly-akamai (commented):
Thanks for addressing the UI issues @hkhalil-akamai - the improved UI looks great!

I've left comments about the nitty gritty of your PR, specifically around the new hook you introduced and introducing extra complexity. Happy to pair up to discuss those further

packages/manager/src/components/TagCell/TagCell.tsx
hkhalil-akamai: Note that this is a slight change to the existing overflow detection code (the main addition is detecting overflows caused by resizing the window) -- this is what causes the display of the ellipses button/shadow effect when tags overflow in the details footer.

A pure CSS solution would be preferable but I'm having trouble finding such a solution. It seems like what we have currently is the [most frequently recommended solution](https://stackoverflow.com/a/9541579).

packages/manager/src/components/TagCell/TagCell.tsx
bnussman-akamai: I think I'd prefer keeping things super simple and disabling all tags when one is deleting. I respect the idea of `useAtomic `, I just can't help but think that we can find a more straightforward way. 

Maybe simpler solution to this problem would be to perform optimistic UI updates when updating tags? This might take some refactoring, but I could see this working nicely.

abailly-akamai: > Thanks for this well-considered feedback. As I was working on the hook, I was concerned that it may be difficult to understand its purpose and behavior. That's why included a detailed TSDoc as well as unit tests that explain and demonstrate its usage.

Yes, thanks for adding a well tested and documented util! But to my point if you have to do all this may point to an overly complex implementation. 

> And while I agree that it introduces some complexity, I also believe that that complexity is manageable and justified since we may need this utility in other places in the app (e.g., if we introduce UI to asynchronously update a resource).

This is a case or premature optimization IMO. This sentence contradicts itself (_justified since we may_). You also make assumption about the user experience here. I don't think we ever got feedback that this was a bad UI/experience? (not that things can't be improved, but the cost of refactor is an important consideration here).

> I disagree with the final point about correlating useAtomic with jotai. [Atomicity](https://en.wikipedia.org/wiki/Atomicity_(database_systems)) is a general concept in programming and doesn't belong to any particular library. I recognize that it is difficult to capture the purpose of the hook in a succinct name so I'm open to alternative suggestions.

You're right! TIL - something more human readable would be nice but after reading about the concept nomenclature I don't have a better suggestion. Hopefully it is clear to most without a CS degree like me 😆 

@hkhalil-akamai Am frankly not trying to ruin this appreciated effort, just trying to to weight the benefits VS added complexity. I'd like others to chime so my opinion on the matter isn't the only one being considered, I have no issue being wrong if the consensus goes towards your implementation! Just want to flag it because it is warranted. Perhaps bringing this PR to the cafe is the next step?

packages/manager/src/components/TagCell/TagCell.tsx
abailly-akamai: Gotcha - i think we could at the very least debounce as an improvement?

hkhalil-akamai: `useWindowDimensions` defaults to a [300ms debounce](https://github.com/linode/manager/blob/f08356f003411798f52d8a4b5edd272fb2e18712/packages/manager/src/hooks/useWindowDimensions.ts#L19)

packages/manager/src/components/TagCell/TagCell.tsx
hkhalil-akamai: Capturing our discussion from cafe here: we have decided to remove `useAtomic` and return to the original pattern of disabling all tags while updates are being performed.

Other ideas discussed include optimistic updates and presenting tags in a multi-select. These may be worth investigating in the future, especially if we receive user feedback, but this PR already has too much history so I'll leave that experimentation for a later time.

bnussman-akamai (commented):
Looking good. Can we add a loading state to this action?

https://github.com/linode/manager/assets/115251059/4f203ea5-d5f1-474d-a58f-538cd62de963

PR #10103, opened 2024-01-24
upcoming: [M3-7530] - Restrict proxy users from updating username/email

jaalah-akamai:
## Description 📝
As a parent user logged into a proxy account, I should not be able to change the email address or username set by the provisioning API.  

## Changes  🔄
- Fixed alignment of input fields and buttons at `/profile/display` page.
- Added unit tests for disabling `Username` and `Email`
- As proxy user: added tooltip text to buttons to disabled buttons for context

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screenshot 2024-01-23 at 10 41 23 PM | Screenshot 2024-01-23 at 10 51 06 PM |

### Prerequisites
- Turn on MSW and Parent/Child feature flag
- Set `user_type` to `proxy` in [serverHandlers](https://github.com/linode/manager/blob/develop/packages/manager/src/mocks/serverHandlers.ts#L554)

### Reproduction steps
- Go to http://localhost:3000/profile/display

### Verification steps 
- Observe disabled fields
- Change `user_type` to something else, observe fields are editable

## As an Author I have considered 🤔

*Check all that apply*
- [x] 👀 Doing a self review
- [x] ❔ Our [contribution guidelines](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md)
- [x] 🤏 Splitting feature into small PRs
- [x] ➕ Adding a changeset
- [x] 🧪 Providing/Improving test coverage
- [x] 🔐 Removing all sensitive information from the code and PR description
- [x] 🚩 Using a feature flag to protect the release
- [x] 👣 Providing comprehensive reproduction steps
- [x] 📑 Providing or updating our documentation
- [x] 🕛 Scheduling a pair reviewing session
- [x] 📱 Providing mobile support
- [x] ♿  Providing accessibility support

PR #10139, opened 2024-02-01
test: [M3-7698] - Add test to check proxy user disabled username/email field

cliu-akamai:
## Description 📝
Add regression tests for checking proxy user disabled username/email field

## Major Changes 🔄
- Add regression tests to check proxy user disabled username/email field
## How to test 🧪
```
yarn cy:run -s "cypress/e2e/core/account/change-username.spec.ts"
```

### DIFF
--- a/packages/manager/cypress/e2e/core/account/change-username.spec.ts
+++ b/packages/manager/cypress/e2e/core/account/change-username.spec.ts
@@ -1,3 +1,17 @@
+import { profileFactory } from '@src/factories';
+import { accountUserFactory } from '@src/factories/accountUsers';
+import { grantsFactory } from '@src/factories/grants';
+import {
+  mockGetUser,
+  mockGetUserGrants,
+  mockGetUsers,
+} from 'support/intercepts/account';
+import {
+  mockAppendFeatureFlags,
+  mockGetFeatureFlagClientstream,
+} from 'support/intercepts/feature-flags';
+import { makeFeatureFlagData } from 'support/util/feature-flags';
+import { mockGetProfile } from 'support/intercepts/profile';
 import { getProfile } from 'support/api/account';
 import { interceptGetProfile } from 'support/intercepts/profile';
 import {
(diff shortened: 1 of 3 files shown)


jdamore-linode (commented):
Nice work @cliu-akamai!

A couple suggestions:

- In addition to confirming the buttons are disabled, we also want to confirm that a tooltip appears explaining why the user can't update the username/email:
  Screenshot 2024-02-01 at 1 39 18 PM
  This is pretty easy to do -- you just have to click the button and then use our new [tooltip UI helper](https://github.com/linode/manager/blob/develop/packages/manager/cypress/support/ui/tooltip.ts#L8):
  ```typescript
  // Click the button first, then confirm the tooltip is shown:
  ui.tooltip.findByText('This account type cannot update this field.').should('be.visible');
  ```
  (We use this in `billing/restricted-user-billing.spec.ts` if you want to see a working example)

- Since this page is under `/profile/display`, I'm not sure we should put this test under `user-permissions.spec.ts`. If it isn't too much work, maybe this would make more sense if it were moved to the `change-username.spec.ts` file? (Even that doesn't seem like a perfect fit)

jdamore-linode (approved):
Thanks @cliu-akamai! The only other thing is the missing changeset (you can run `yarn changeset` to generate and commit one), but approving this in the meantime. Nice work!

packages/manager/cypress/e2e/core/account/change-username.spec.ts
```diff
@@ -96,4 +110,57 @@ describe('username', () => {
...
+    });
+
+    const mockRestrictedProxyProfile = profileFactory.build({
+      username: 'restricted-proxy-user',
+      user_type: 'proxy',
+    });
+
+    const mockUserGrants = grantsFactory.build({
+      global: { account_access: 'read_write' },
+    });
```
mjac0bs, line 128:
```suggestion
    const mockRestrictedProxyProfile = profileFactory.build({
      username: 'restricted-proxy-user',
      user_type: 'proxy',
      restricted: true
    });
```

I don't think we need to mock the account users endpoint here at all, nor grants. We [disable the buttons and display the tooltips](https://github.com/linode/manager/blob/develop/packages/manager/src/features/Profile/DisplaySettings/DisplaySettings.tsx#L107-L114) based on whether the user is restricted or a proxy user. We can use `restricted` and `user_type` from `*/profile` for that.

packages/manager/cypress/e2e/core/account/change-username.spec.ts
```diff
@@ -96,4 +110,57 @@ describe('username', () => {
       cy.findByText('Username updated successfully.').should('be.visible');
     });
   });
+
+  it('disables username/email fields for proxy user', () => {
```
mjac0bs, line 114: If we want to cover all test cases here, it would be good to test the following:

- disables username and email fields for restricted proxy user (what we currently have)
- disables username and email fields for unrestricted proxy user (`restricted: false` in `*/profile`)
- disables username and email fields for a regular restricted user (`user_type: null` and `restricted: true` in `*/profile`)

I also wonder a little about the usefulness of some [unit](https://github.com/linode/manager/pull/10103/files#diff-b2a93542dd7bc1466f1b288b00d9fca34e0dfa33ee3f5fa478eba32bd2c4c05c) and Cypress test overlap. We're testing some of the same things. At what point does duplication become detrimental? (That's more of a larger question - cc @jdamore-linode.)

mjac0bs (changes requested):
Thanks, Cassie! I left some thoughts in comments.

packages/manager/cypress/e2e/core/account/change-username.spec.ts
cliu-akamai: @mjac0bs Not sure if it's a bug, the email field is not disabled for the regular restricted user.

jaalah-akamai: Nope that's correct, restricted users can still edit email - just not proxy users

mjac0bs (approved):
Thanks for the updates. Looks good locally and in CI! The test failure in CI is the billing-contact.spec which is fixed in develop now, so you should be good to go here.

image

PR #10141, opened 2024-02-01
change: [M3-7703] - Hide error message for $0 regions

hana-linode:
## Description 📝
Hide error message for $0 regions since that is not technically an error. Continue to display the error icon otherwise.

## Changes  🔄
- Error message tooltip no longer displayed for $0 region prices in Linode and Kubernetes Create
- Plus and minus and Add buttons are no longer disabled in Kubernetes Create Plan table if the price is $0

## Preview 📷
| Linode Create Before  | Linode Create After   |
| ------- | ------- |
| image | image |

| Kubernetes Create Before  | Kubernetes Create After   |
| ------- | ------- |
| <video src="https://github.com/linode/manager/assets/115299789/bafdc8b2-2582-436a-b25f-8d1869ba1df6" /> | <video src="https://github.com/linode/manager/assets/115299789/426c7b66-0022-449d-8ab1-c71a364889a1" /> |

### Reproduction steps
- On another branch/prod, go to Linode/Kubernetes Create
- Select the Madrid region
- Notice the error icon for both the `$0` hourly price and `$--.--` monthly price
- Notice the disabled plus, minus, add buttons in the Kubernetes Create Plan table

### Verification steps 
- Go to Linode/Kubernetes Create and select the Madrid region
- You should no longer see an error icon for the `$0` hourly price. You should continue to see an error icon for the `$--.--` monthly price
- The plus, minus, add buttons should no longer be disabled in the Kubernetes Create Plan table

```
yarn test KubernetesPlanSelection
yarn test PlanSelection
```

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/Kubernetes/KubernetesPlansPanel/KubernetesPlanSelection.test.tsx
+++ b/packages/manager/src/features/Kubernetes/KubernetesPlansPanel/KubernetesPlanSelection.test.tsx
@@ -72,6 +72,41 @@ describe('KubernetesPlanSelection (table, desktop view)', () => {
     expect(queryByText(regionMonthlyPrice)).toBeInTheDocument();
   });
 
+  it('should not display an error message for $0 regions', () => {
+    const propsWithRegionZeroPrice = {
+      ...props,
+      type: extendedTypeFactory.build({
+        region_prices: [
+          {
+            hourly: 0,
+            id: 'id-cgk',
+            monthly: 0,
+          },
+        ],
+      }),
+    };
+    const { container } = renderWithTheme(
+      wrapWithTableBody(
+        <KubernetesPlanSelection
+          {...propsWithRegionZeroPrice}
+          selectedRegionId={'id-cgk'}
+        />
+      )
+    );
+
+    const monthlyTableCell = container.querySelector('[data-qa-monthly]');
+    const hourlyTableCell = container.querySelector('[data-qa-hourly]');
+    expect(monthlyTableCell).toHaveTextContent('$0');
+    // error tooltip button should not display
+    expect(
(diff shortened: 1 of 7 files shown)


packages/manager/src/features/Kubernetes/KubernetesPlansPanel/KubernetesPlanSelection.test.tsx
hkhalil-akamai, line 1: Should we also add a test (here and in `PlanSelection`) for when a region is selected but the price is null?

hkhalil-akamai (approved):
LGTM. Left one suggestion for a possible test.

carrillo-erik (approved):

mjac0bs (commented):
Thanks for these changes, Hana!

- Confirmed that $0 is a valid price that does not display an error icon/tooltip in either Linodes Create or Kube Create in both desktop and mobile view. 
- Verified I can create a kube cluster in Madrid.

One minor issue I see is that I think we'll need to change the logic in `CheckoutBar.tsx` because right now the `priceHelperText` is displaying when the user selects a DC with $0 pricing and "No" for HA. A user would expect to see a total price of $0 in that case. And, presumably, if the DC costs $0, then HA should also cost $0, if they select "Yes" for HA - we're showing them a total of $60.

Screenshot 2024-02-01 at 3 52 30 PM
Screenshot 2024-02-01 at 3 52 48 PM

hana-linode:
@mjac0bs Addressed the $0 view. Updating the HA price from $60 will be a bigger lift due to hardcoded values and coupled code. I created M3-7732 to address that since it's a bit out of scope for this ticket

PR #10152, opened 2024-02-06
upcoming: [M3-7716] - Add session expiry confirmation dialog for proxy users

jaalah-akamai:
## Description 📝
When a proxy user wants to switch back to their parent account, the parent account's authentication token must still be valid in order for them to switch back to the parent's account or log into any other proxy accounts displayed in the drawer.

If the parent token is not valid, a proxy user should be given adequate notice that they must re-log in to the parent account. We also will disabled the "Switch Account" buttons if this is the case.

## Changes  🔄
- Updated the `useDialogContext.ts` hook API. The original API was kind of rigid, whereas now we can have dynamic state that we can extend if we need to. The original references to `open` and `closed` and API still work, but I will clean that up after this PR.

```
useDialogContext({
    continueSession: false,
    isOpen: false,
});
```
- Added `SwitchAccountSessionDialog.tsx` for when parent session expires
- Added `useParentTokenManagement.tsx` for checking expiration and showing dialog. We still may need to poll the expiry or at least check every time the user menu opens or a user lands on the account page.

## Preview 📷
| Screenshots   |
| ------- |
| localhost_3000_account_billing (2)|
|localhost_3000_account_billing|
|localhost_3000_account_billing (1)|

### Prerequisites
- Turn on MSW and Parent/Child feature flag
- By default you should be a `parent` user, if not, update: [serverHandlers](https://github.com/linode/manager/blob/develop/packages/manager/src/mocks/serverHandlers.ts#L554)

### Reproduction steps
- Open user menu
- Click Switch Account button
- Select a child company

### Verification steps 
- Observe parent/proxy tokens are created in local storage
- Set `user_type` to `proxy` in [serverHandlers](https://github.com/linode/manager/blob/develop/packages/manager/src/mocks/serverHandlers.ts#L554)
- Observe no dialog appears since we have an active parent token

### Prerequisites
- Clear application cookies
- Turn on MSW and Parent/Child feature flag
- Set `user_type` to `proxy` in [serverHandlers](https://github.com/linode/manager/blob/develop/packages/manager/src/mocks/serverHandlers.ts#L554)

### Reproduction steps
- Since there is no parent token defined, it's invalid and you should see the dialog

### Verification steps 
- Close the dialog and observe that it should not appear again
- Observe that both "Switch Account" buttons in user menu and account page are disabled

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/MainContent.tsx
+++ b/packages/manager/src/MainContent.tsx
@@ -29,6 +29,7 @@ import { isFeatureEnabled } from 'src/utilities/accountCapabilities';
 
 import { ENABLE_MAINTENANCE_MODE } from './constants';
 import { complianceUpdateContext } from './context/complianceUpdateContext';
+import { switchAccountSessionContext } from './context/switchAccountSessionContext';
 import { FlagSet } from './featureFlags';
 import { useIsACLBEnabled } from './features/LoadBalancers/utils';
 import { useGlobalErrors } from './hooks/useGlobalErrors';
@@ -190,7 +191,12 @@ export const MainContent = () => {
   const contextValue = useNotificationContext();
 
   const ComplianceUpdateProvider = complianceUpdateContext.Provider;
-  const complianceUpdateContextValue = useDialogContext();
+  const complianceUpdateContextValue = useDialogContext({});
+
+  const SwitchAccountSessionProvider = switchAccountSessionContext.Provider;
+  const switchAccountSessionContextValue = useDialogContext({
+    continueSession: false,
+  });
 
   const [menuIsOpen, toggleMenu] = React.useState<boolean>(false);
   const {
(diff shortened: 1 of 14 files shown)


packages/manager/src/MainContent.tsx
```diff
@@ -294,100 +301,105 @@ export const MainContent = () => {
...
-                  desktopMenuIsOpen ||
-                  (desktopMenuIsOpen && desktopMenuIsOpen === true),
-              })}
-            >
-              <TopMenu
-                desktopMenuToggle={desktopMenuToggle}
-                isSideMenuOpen={!desktopMenuIsOpen}
-                openSideMenu={() => toggleMenu(true)}
-                username={username}
+      <SwitchAccountSessionProvider value={switchAccountSessionContextValue}>
```
jaalah-akamai, line 304: Only change here was adding the Provider wrapper

packages/manager/src/context/useDialogContext.ts
```diff
@@ -1,26 +1,51 @@
...
+  initialState: UseDialogContextOptions = {}
+): DialogContextProps => {
+  const [state, setState] = useState({ ...defaultContext, ...initialState });
+
+  // TODO: We no longer need the open and close functions after we update other references
+  const open = useCallback(
+    () => setState((prevState) => ({ ...prevState, isOpen: true })),
+    []
+  );
+  const close = useCallback(
```
jaalah-akamai, line 35: Added TODO since `open` and `close` are no longer necessary, but I didn't want to break existing API.

packages/manager/src/context/useDialogContext.ts
```diff
@@ -1,26 +1,51 @@
...
 
-export const defaultContext = {
+export type UseDialogContextOptions = {
+  [key: string]: boolean;
+};
+
+export const defaultContext: DialogContextProps = {
   close: () => void 0,
   isOpen: false,
   open: () => void 0,
```
jaalah-akamai, line 21: We'll be able to remove these 3 as well in upcoming PR

packages/manager/src/features/Account/SwitchAccounts/useParentTokenManagement.tsx
```diff
@@ -0,0 +1,32 @@
...
+    }
+  }, [isProxyUser]);
+
+  React.useEffect(() => {
+    if (isParentTokenExpired && sessionContext.continueSession === false) {
+      sessionContext.updateState({ continueSession: true, isOpen: true });
+    }
+  }, [isParentTokenExpired, sessionContext]);
+
+  return { isParentTokenExpired };
```
jaalah-akamai, line 31: Open to input/thoughts here

packages/manager/src/features/Account/SwitchAccounts/useParentTokenManagement.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  const checkParentToken = async () => {
+    const isExpired = !isParentTokenValid();
+    setIsParentTokenExpired(isExpired);
+  };
+
+  React.useEffect(() => {
+    if (isProxyUser) {
+      checkParentToken();
+    }
```
hana-linode, line 22: Can we inline this? The function name `checkParentToken` is a bit misleading since it's also setting a value

```suggestion
    if (isProxyUser) {
      const isExpired = !isParentTokenValid();
      setIsParentTokenExpired(isExpired);
    }
```

hana-linode (commented):
If a user is able to continue working on an account with an expired parent token, I think it would make more sense to only display this dialog when they try to switch accounts?

image

jaalah-akamai:
> If a user is able to continue working on an account with an expired parent token, I think it would make more sense to only display this dialog when they try to switch accounts?

The goal was to prevent users from attempting to switch accounts once the token expired - so the `Switch Account` button would be disabled at expiry. I'll spend some more time thinking about the UX, but that was the intention.

hana-linode:
@jaalah What if we don't disable the Switch Account button at expiry but just have it open the session expiry confirmation dialog instead? That way the dialog doesn't display when the user isn't trying to switch accounts but still prevents users from switching accounts

jaalah-akamai:
@hana-linode 🙌 


https://github.com/linode/manager/assets/125309814/89b8e234-67a5-4721-9a9a-abab533e02e7

jaalah-akamai:
I'm keeping things in context because we're most likely going to implement a way to refresh parent tokens as expiry nears an end, so I'll swap the dialog and save me some work.

hana-linode:
Liking this flow a lot better. I was thinking we can keep the Switch Account buttons enabled even after the user hits close

https://github.com/linode/manager/assets/115299789/78ce7a91-e089-48b1-926f-ac0e33f1c833

jaalah-akamai:
@hana-linode @mjac0bs ✅ thanks for feedback!

packages/manager/src/features/Account/AccountLanding.tsx
```diff
@@ -137,7 +153,10 @@ const AccountLanding = () => {
     }
     landingHeaderProps.disabledCreateButton = readOnlyAccountAccess;
     landingHeaderProps.extraActions = canSwitchBetweenParentOrProxyAccount ? (
-      <SwitchAccountButton onClick={() => setIsDrawerOpen(true)} />
+      <SwitchAccountButton
+        onClick={handleAccountSwitch}
+        tooltipText={PARENT_SESSION_EXPIRED}
```
mjac0bs, line 158: Do we need this tooltip text at all now that we're keeping the Switch Account button enabled?

packages/manager/src/features/Account/SwitchAccounts/useParentTokenManagement.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+import { isParentTokenValid } from 'src/features/Account/utils';
+
+export const useParentTokenManagement = ({
```
mjac0bs, line 5: This could use a function comment.

packages/manager/src/features/TopMenu/UserMenu/UserMenu.tsx
```diff
@@ -255,7 +270,8 @@ export const UserMenu = React.memo(() => {
           {canSwitchBetweenParentOrProxyAccount && (
             <SwitchAccountButton
               buttonType="outlined"
-              onClick={() => setIsDrawerOpen(true)}
+              onClick={handleAccountSwitch}
+              tooltipText={PARENT_SESSION_EXPIRED}
```
mjac0bs, line 274: Same comment as the other Switch Account button about the existence of this tooltip.

packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -553,7 +553,7 @@ export const handlers = [
     const profile = profileFactory.build({
       restricted: false,
       // Parent/Child: switch the `user_type` depending on what account view you need to mock.
-      user_type: 'parent',
+      user_type: 'proxy',
```
mjac0bs, line 556: Let's revert this before merging just to keep things consistent.

packages/manager/src/features/Account/utils.ts
```diff
@@ -41,16 +62,33 @@ export const getAccessRestrictedText = (userType: UserType | null) => {
...
 };
 
+/**
+ * Determine whether the user has restricted access to a specific resource.
+ *
+ * @example
+ * // If account access does not equal 'read_write', the user has restricted access.
+ * isRestrictedGlobalGrantType({
+ *   globalGrantType: 'account_access',
+ *   allowableGrantLevel: 'read_write',
```
mjac0bs, line 72:
```suggestion
 *   permittedGrantLevel: 'read_write',
```

packages/manager/src/features/Account/utils.ts
```diff
@@ -17,6 +18,26 @@ interface GetRestrictedResourceText {
...
+  permittedGrantLevel: GrantLevel;
+}
+
+interface NonAccountAccessGrant extends GrantsProfileSchema {
+  globalGrantType: Exclude<GlobalGrantTypes, 'account_access'>;
+  permittedGrantLevel?: GrantLevel;
+}
+
+// Discriminating union to determine the type of global grant
+export type IsRestrictedGlobalGrantType =
```
mjac0bs, line 37:
```suggestion
export type RestrictedGlobalGrantType =
```
Do we need the `Is`?

packages/manager/src/features/Account/utils.ts
```diff
@@ -41,16 +62,33 @@ export const getAccessRestrictedText = (userType: UserType | null) => {
...
   grants,
+  permittedGrantLevel,
   profile,
-}: {
-  globalGrantType: GlobalGrantTypes;
-  grants: Grants | undefined;
-  profile: Profile | undefined;
-}): boolean => {
-  return Boolean(profile?.restricted) && !grants?.global[globalGrantType];
+}: IsRestrictedGlobalGrantType): boolean => {
```
mjac0bs, line 83:
```suggestion
}: RestrictedGlobalGrantType): boolean => {
```
Related to suggestion above.

mjac0bs (approved):

hana-linode (approved):

PR #10160, opened 2024-02-07
refactor: [M3-7582] - Remove Enzyme

bnussman-akamai:
## Description 📝

Removing Enzyme in favor or React Testing Library has been in the works for years now. This PR's goal is to finally get Enzyme out of here 🗑️. We're really close, so I'm just going to do it all in one sweep with this PR. 

Also, removing Enzyme will make updating to React 18 a bit easier.

## How to test 🧪

- Verify unit tests pass in Github Actions
- Optional: check out this PR and run some of the unit tests that have been touched
- Verify that we're not losing any meaningful test coverage 

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset


packages/manager/package.json
```diff
@@ -199,9 +195,11 @@
     "mocha-junit-reporter": "^2.2.1",
     "msw": "~1.3.2",
     "prettier": "~2.2.1",
+    "react-test-renderer": "16.14.0",
```
bnussman-akamai, line 198: We have to explicitly define this now that enzyme does not include it. 

If we don't install this, we'll get the following:
Screenshot 2024-02-07 at 3 35 29 PM

packages/manager/src/components/Pagey/Pagey.test.tsx
bnussman-akamai, line 1: Thankfully, we don't use "Pagey" anywhere so I was able to remove all of this!

packages/manager/package.json
bnussman-akamai: My hope is that when we upgrade React Testing Library, we can remove this explicit definition.

packages/manager/package.json
```diff
@@ -172,8 +171,6 @@
     "cypress-real-events": "^1.11.0",
     "cypress-vite": "^1.5.0",
     "dotenv": "^16.0.3",
-    "enzyme": "^3.10.0",
-    "enzyme-adapter-react-16": "^1.14.0",
```
hkhalil-akamai, line 176: ✌️

hkhalil-akamai (approved):
Great work rewriting all of these tests! I glanced over all changes and ran tests and coverage locally and everything looks good.

PR #10163, opened 2024-02-08
refactor: [M3-7749] - Update `luxon` to latest

bnussman-akamai:
## Description 📝

Updates Luxon to latest 📦

## Changes  🔄
- Luxon now returns `null` for invalid dates instead of throwing an error ⛔️
- Removes `chartjs-adapter-luxon` because it is not doing anything meaningful 🗑️📦

## How to test 🧪

- Verify none of the code changes will cause breaking UI changes
- Verify unit tests and cypress tests both pass

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/features/Billing/BillingPanels/BillingActivityPanel/BillingActivityPanel.test.tsx
```diff
@@ -168,6 +168,11 @@ describe('paymentToActivityFeedItem', () => {
         { zone: 'utc' }
       );
       const testDateISO = testDate.toISO();
+
+      if (!testDateISO) {
```
bnussman-akamai, line 172: Because `toISO` will now return `null` instead of throwing, we need to handle the `null` case. 

In most places, we can just put a simple check. This is fine because in most cases, the dates are just being used for testing.

packages/manager/src/features/Linodes/CloneLanding/utilities.ts
```diff
@@ -304,7 +304,7 @@ export const getEstimatedCloneTime = (
     minutes: estimatedTimeInMinutes,
     seconds: estimatedTimeInMinutes > 0 ? 1 : 0, // in case less than 1 min
   });
-  let humanizedEstimate = then.toRelative(now.toObject());
+  let humanizedEstimate = then.toRelative({ base: now });
```
jaalah-akamai, line 307: What's the difference here?

jaalah-akamai (approved):

packages/manager/src/features/Linodes/CloneLanding/utilities.ts
bnussman-akamai: The type for `toRelative` changed, but these should do the same thing. `{ base: now }` probably isn't needed because Luxon defaults to "now" anyway, but I went ahead and kept it to be explicit. 

This function is unit tested so I feel good about this change.

abailly-akamai (approved):
Everything running well locally and tests passing ✅ 

I can't wait for [temporal](https://tc39.es/proposal-temporal/docs/) to make it to ECMAScript and finally have a browser native date API. Luxon is fine but if I am not wrong doesn't handle tree shaking well, resulting in a fatter bundle. Not a big deal and far better than `moment.js` still

bnussman-akamai:
Ohh, thanks for sharing @abailly-akamai. I wasn't aware of Temporal 😲

I can't speak for the performance of Luxon, but I've come to appreciate the utility it gives us

PR #10167, opened 2024-02-08
test: [M3-7452] - Add Cypress test to check empty state in Images landing page

cliu-akamai:
## Description 📝
Add regression tests to check Images landing page empty state.

## Major Changes 🔄
- Add regression tests to check the empty state in Images landing page.

## How to test 🧪
```
yarn cy:run -s "cypress/e2e/core/images/images-empty-landing-page.spec.ts"
```

### DIFF
--- a/packages/manager/cypress/e2e/core/images/images-empty-landing-page.spec.ts
+++ b/packages/manager/cypress/e2e/core/images/images-empty-landing-page.spec.ts
@@ -1,8 +1,6 @@
 import { ui } from 'support/ui';
 import { mockGetAllImages } from 'support/intercepts/images';
-import { authenticate } from 'support/api/authentication';
 
-authenticate();
 describe('Images empty landing page', () => {
   /**
    * - Confirms Images landing page empty state is shown when no Images are present:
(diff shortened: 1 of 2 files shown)


packages/manager/cypress/e2e/core/images/images-empty-landing-page.spec.ts
```diff
@@ -0,0 +1,51 @@
+import { ui } from 'support/ui';
+import { mockGetAllImages } from 'support/intercepts/images';
+import { authenticate } from 'support/api/authentication';
+
+authenticate();
```
mjac0bs, line 5: Do we need this call to `authenticate()`?  Based on my understanding, we don't since we're mocking the request to GET images, not using a real API request, and since other similar empty landing page specs don't use it.

mjac0bs (approved):
Looks good pending one minor comment and we'll need to generate a test changeset. 🚢

packages/manager/cypress/e2e/core/images/images-empty-landing-page.spec.ts
jdamore-linode: @mjac0bs That's right -- `authenticate()` is only necessary when we're making real API requests from inside the tests, so it's not needed here 👍

jdamore-linode (approved):
Nice work @cliu-akamai! Approved pending the things Mariah mentioned:

- Remove call to `authenticate()` and related import
- Add a changeset

PR #10168, opened 2024-02-08
PoC - React Query Factory

hkhalil-akamai:
## Description 📝
Introduces a new utility `createQueryStore` that enables automatic, type-safe management of React query keys.

In this PR, I demonstrate using `createQueryStore` to manage profile queries.

## Changes  🔄
- ...
- ...

## Preview 📷

| Before  | After   |
| ------- | ------- |
| 📷 | 📷 |

### Prerequisites
- ...
- ...

### Reproduction steps
- ...
- ...

### Verification steps 
(How to verify changes)
- ...
- ...

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

---

**Commit Types:**

---

### DIFF
--- a/packages/manager/src/features/Profile/AuthenticationSettings/PhoneVerification/PhoneVerification.tsx
+++ b/packages/manager/src/features/Profile/AuthenticationSettings/PhoneVerification/PhoneVerification.tsx
@@ -12,7 +12,7 @@ import { LinkButton } from 'src/components/LinkButton';
 import { TextField } from 'src/components/TextField';
 import { Typography } from 'src/components/Typography';
 import {
-  queryKey,
+  profileQueryStore,
   updateProfileData,
   useProfile,
   useSendPhoneVerificationCodeMutation,
@@ -98,7 +98,7 @@ export const PhoneVerification = ({
       );
     } else {
       // Cloud Manager does not know about the country, so lets refetch the user's phone number so we know it's displaying correctly
-      queryClient.invalidateQueries(queryKey);
+      profileQueryStore.invalidateQueries(queryClient);
     }
 
     // reset form states
--- a/packages/manager/src/utilities/createQueryStore.ts
+++ b/packages/manager/src/utilities/createQueryStore.ts
@@ -1,0 +1,105 @@
+import { APIError } from '@linode/api-v4';
+import {
+  InvalidateOptions,
+  InvalidateQueryFilters,
+  QueryClient,
+  QueryFunction,
+  QueryKey,
+  SetDataOptions,
+  UseQueryOptions,
+  UseQueryResult,
+  useQuery,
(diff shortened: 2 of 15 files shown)


packages/manager/src/utilities/createQueryStore.ts
```diff
@@ -0,0 +1,117 @@
...
+          getQueryData: (queryClient, filters) =>
+            queryClient.getQueryData(queryKey, filters),
+          setQueryData: (queryClient, updater, options) =>
+            queryClient.setQueryData(queryKey, updater, options),
+          useQuery: (options) => useQuery(queryKey, config.queryFn!, options),
+        }
+      : {}),
+    ...(config.children
+      ? Object.keys(config.children).reduce((acc, child) => {
+          const childQueryKey = [...queryKey, child[0]];
```
hkhalil-akamai, line 92:
```suggestion
          const childQueryKey = [...queryKey, child];
```

packages/manager/src/features/Profile/AuthenticationSettings/PhoneVerification/PhoneVerification.tsx
```diff
@@ -98,7 +98,7 @@ export const PhoneVerification = ({
       );
     } else {
       // Cloud Manager does not know about the country, so lets refetch the user's phone number so we know it's displaying correctly
-      queryClient.invalidateQueries(queryKey);
+      profileQueryStore.invalidateQueries(queryClient);
```
bnussman-akamai, line 101: I wonder if it would make sense to make `profileQueryStore` a hook so that we don't have to pass `queryClient` every time?

```typescript
const profileQueryStore = useProfileQueryStore();

profileQueryStore.invalidateQueries()
```

bnussman-akamai (commented):
Overall this is really neat. The type-safety seems on point too. 

My main concern with this is the amount of abstraction taking place. It delivers a lot of value, but looks and feels significantly different than vanilla React Query and has a slightly different mental model. I really value keeping things as simple and "vanilla" because I find it more maintainable and newcomer friendly. 

I think this is great, but I'm currently on the fence just for the abstraction reason. I might just be thinking this way because I was terrified of how over-abstracted Cloud Manager was a few years back. Everything was so convoluted and extra. We've come a long way since. 

I'm curious to hear what others think of this solution compared to just a more basic query key factory.

hkhalil-akamai:
I can see the danger with over-abstraction. My main concern is that if there's a new breaking change to RQ we'd have to reimplement the wrapper. On the other hand, this approach gives us the strongest type-safety, particularly for methods like `setQueryData`

hkhalil-akamai:
Closed in favor of [@bnussman/query-key](https://github.com/bnussman/query-key).

PR #10170, opened 2024-02-09
feat: [M3-7645] – Support VPC IPv4 Ranges in Add/Edit Linode Config dialog

dwiley-akamai:
## Description 📝
Add support for VPC IPv4 ranges in Add/Edit Linode Config dialog

## Changes  🔄
- Remove some `vpcEnabled`/`isFeatureEnabled` logic in `LinodeConfigDialog.tsx`
- Logic to include `ip_ranges` in config payload, display individual IP range errors

## Preview 📷
<details>
<summary>Starting state, before any ranges have been assigned</summary>

Screenshot 2024-02-09 at 1 41 29 PM
</details>

### Verification steps 
For a Linode already assigned to a VPC, go to Linode Details > Configurations > Edit and confirm that the info displayed for the VPC interface is correct (particularly the IP ranges -- existing IP ranges should be displayed, but if there are none, you should just see the "Assign additional IPv4 ranges" section with no input fields to start).

Try adding and deleting IP ranges and ensure the payload (PUT request to `linode/instances/{linode_id}/configs/{config_id}`) and UI matches what you'd expect.

Note: The CIDR notation (e.g., `/24`) of the range needs to be narrower (i.e., a larger number) than the subnet IP range. For example, if the subnet IP range has `/16`, the CIDR notation for the ranges could be `/24`.

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx
```diff
@@ -971,30 +958,44 @@ export const LinodeConfigDialog = (props: Props) => {
...
+                    `interfaces[${idx}].ip_ranges[${index}]`
+                  ]?.includes('is invalid')
+                    ? 'Invalid IP range'
+                    : formik.errors[`interfaces[${idx}].ip_ranges[${index}]`];
+
+                  return {
+                    address: ip_range,
+                    error: errorString,
+                  };
+                });
```
dwiley-akamai, line 997: Because this is a deeply nested error that we're trying to surface, it required a bit of less-than-ideal logic and mapping.

I am aware of one unfortunate shortcoming–if you have multiple IP ranges but one of them (not the last one in the sequence) gets an error when you click "Save Changes", clicking the "X" to remove that IP range will result in the error moving onto the range that moves up to occupy that index. Sadly, I could not see a way around this, as I tried having something like `ipRangeIndicesWithErrors` in component state or as a ref to track the indices so that the errors could be cleared from inside `handleInterfaceChange` or `handleChange`, but that was not fruitful.

bnussman-akamai (approved):
The Linode Config dialog is working as I'd expect! Error handling looks good too 🎉 

Screenshot 2024-02-12 at 11 46 21 AM


There is some weirdness with this specific case, but I don't think we need to address this right now. 

https://github.com/linode/manager/assets/115251059/3b11cc58-d36e-478b-9b81-279671cb53a7

cpathipa (approved):
LGTM! confirming on "Assign additional IPv4 ranges"

abailly-akamai (approved):
UI and code looks good, behavior works as expected ✅ 

The layout with a large right col is a bit odd. Maybe it can be improved in the future

Screenshot 2024-02-12 at 11 56 01

packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx
```diff
@@ -667,12 +659,9 @@ export const LinodeConfigDialog = (props: Props) => {
...
   const networkInterfacesHelperText = (
     <Typography>
-      Configure the network that a selected interface will connect to (
-      {vpcEnabled
-        ? '"Public Internet", VLAN, or VPC'
-        : 'either "Public Internet" or a VLAN'}
-      ) . Each Linode can have up to three Network Interfaces. For more
-      information, see our{' '}
+      Configure the network that a selected interface will connect to
+      &quot;Public Internet&quot;, VLAN, or VPC. Each Linode can have up to
```
hkhalil-akamai, line 663: Do we need the `&quot;`?
```suggestion
      "Public Internet", VLAN, or VPC. Each Linode can have up to
```

jaalah-akamai (approved):

hana-linode (approved):
VPC IPv4 Configs LGTM other than the mentioned error cases

hkhalil-akamai (approved):

packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx
dwiley-akamai: ESLint recommended escaping them

Screenshot 2024-02-12 at 12 34 48 PM

jdamore-linode (approved):

PR #10179, opened 2024-02-12
upcoming: [M3-7632] - Use infinite query for child accounts

jaalah-akamai:
## Description 📝
Considering parent accounts may have a large number of children, for better performance we're loading and displaying the number of child accounts by pages, which will update upon scrolling.

## Changes  🔄
- Updated MSW to account for changes
- Added new: `useChildAccountsInfiniteQuery`

## Preview 📷
Screenshot 2024-02-12 at 1 12 08 PM

### Prerequisites
- Turn on MSW

### Verification steps 
(How to verify changes)
- Open user menu and click "Switch Account"
- Observe first 25 accounts load and rest are fetched upon scroll

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/queries/account.ts
mjac0bs, line 1: Do we need to keep `useChildAccounts` around now that we have `useChildAccountsInfiniteQuery`?

mjac0bs (approved):
Updates to mocks make sense and confirmed that the requests for each page of accounts loaded as a parent/proxy user scrolls the list of accounts in the Switch Account drawer. ✅ 

https://github.com/linode/manager/assets/114685994/d26a5bfc-e7ab-4abc-8865-f8bd97acf187

packages/manager/src/queries/account.ts
```diff
@@ -69,6 +74,42 @@ export const useChildAccounts = ({
...
+        params: {
+          page: pageParam,
+          page_size: 25,
+        },
+      }),
+    {
+      enabled:
+        (Boolean(profile?.user_type === 'parent') && !profile?.restricted) ||
+        Boolean(grants?.global?.child_account_access) ||
+        hasExplicitAuthToken,
```
bnussman-akamai, line 101: Optional: Move `enabled` into a const to help readability

bnussman-akamai (approved):
Nice! Love seeing us optimize our data fetching 🎉

packages/manager/src/queries/account.ts
jaalah-akamai: If we find we have no use for it, we can remove it after next week 👍

packages/manager/src/queries/account.ts
```diff
@@ -69,6 +74,42 @@ export const useChildAccounts = ({
...
+  filter,
+  headers,
+  params,
+}: RequestOptions) => {
+  const { data: profile } = useProfile();
+  const { data: grants } = useGrants();
+  const hasExplicitAuthToken = Boolean(headers?.Authorization);
+
+  return useInfiniteQuery<ResourcePage<Account>, APIError[]>(
+    [queryKey, 'childAccounts', 'paginated', params, filter],
```
bnussman-akamai, line 87: We should probably make the query key different than the paginated query to prevent them overwriting each other

We also should consider putting `headers` in the query key. 

```suggestion
    [queryKey, 'childAccounts', 'infinite', params, filter],
```

jaalah-akamai: @bnussman-akamai yes definitely missed this - good catch

PR #10204, opened 2024-02-16
test: [M3-7497] - Add tests for child user verification banner

cliu-akamai:
## Description 📝
Add regression tests for checking child user verification banner

## Major Changes 🔄
- Add regression tests to check child user verification banner

## How to test 🧪
```
yarn cy:run -s "cypress/e2e/core/account/user-verification-banner.spec.ts"
```

### DIFF
--- a/packages/manager/cypress/e2e/core/account/user-verification-banner.spec.ts
+++ b/packages/manager/cypress/e2e/core/account/user-verification-banner.spec.ts
@@ -1,0 +1,165 @@
+import { profileFactory, securityQuestionsFactory } from '@src/factories';
+import { accountUserFactory } from '@src/factories/accountUsers';
+import { grantsFactory } from '@src/factories/grants';
+import {
+  mockGetUser,
+  mockGetUserGrants,
+  mockGetUsers,
+} from 'support/intercepts/account';
+import { mockGetSecurityQuestions } from 'support/intercepts/profile';
+import {
+  mockAppendFeatureFlags,
+  mockGetFeatureFlagClientstream,
+} from 'support/intercepts/feature-flags';
+import { ui } from 'support/ui';
+import { mockGetProfile } from 'support/intercepts/profile';
+import { makeFeatureFlagData } from 'support/util/feature-flags';
+
+describe('User verification banner', () => {
+  /*
+   * - Confirms that a banner is present the child users do not have a phone number or security questions.
+   * - Confirms that the "Add Verification Details" button redirects the user to /profile/auth.
+   */
+  it('can show up when a child user has not associated a phone number or set up security questions for their account', () => {
+    const mockChildProfile = profileFactory.build({
+      username: 'child-user',
+      user_type: 'child',
+      verified_phone_number: null,
+    });
+
+    const mockChildUser = accountUserFactory.build({
+      restricted: false,
+      username: 'child-user',
+      user_type: 'child',
+      verified_phone_number: null,
+    });
+
+    const mockRestrictedProxyUser = accountUserFactory.build({
+      restricted: true,
+      user_type: 'proxy',
+      username: 'restricted-proxy-user',
+      verified_phone_number: null,
+    });
+
+    const mockUserGrants = grantsFactory.build({
+      global: { account_access: 'read_write' },
+    });
+
+    // TODO: Parent/Child - M3-7559 clean up when feature is live in prod and feature flag is removed.
+    mockAppendFeatureFlags({
+      parentChildAccountAccess: makeFeatureFlagData(true),
+    }).as('getFeatureFlags');
+    mockGetFeatureFlagClientstream().as('getClientStream');
+
+    mockGetUsers([mockRestrictedProxyUser]).as('getUsers');
+    mockGetUser(mockChildUser);
+    mockGetUserGrants(mockChildUser.username, mockUserGrants);
+    mockGetProfile(mockChildProfile);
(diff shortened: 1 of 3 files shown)


jaalah-akamai (changes requested):
@cliu-akamai I think we need one more test to ensure that the banner shows when we have security questions but not a phone number. Otherwise, this is looking good.

mjac0bs (commented):
Other than Jaalah's ask, we'll also need a changeset before this is good to go.

carrillo-erik:
@cliu-akamai I was able to run this locally and after running the tests I did not see the test failing. I re-started the end-to-end test in the Jenkins UI and hopefully that resolves the issue. Everything else looks good.

carrillo-erik (approved):

packages/manager/cypress/e2e/core/account/user-verification-banner.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+  mockGetFeatureFlagClientstream,
+} from 'support/intercepts/feature-flags';
+import { ui } from 'support/ui';
+import { mockGetProfile } from 'support/intercepts/profile';
+import { makeFeatureFlagData } from 'support/util/feature-flags';
+import { verificationBannerNotice } from 'support/constants/user';
+
+describe('User verification banner', () => {
+  /*
+   * - Confirms that a banner is present the child users do not have a phone number or security questions.
```
mjac0bs, line 21:
```suggestion
   * - Confirms that a banner is present when child users do not have a phone number or security questions.
```

packages/manager/cypress/e2e/core/account/user-verification-banner.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+    ui.button
+      .findByTitle('Add verification details')
+      .should('be.visible')
+      .should('be.enabled')
+      .click();
+    cy.url().should('endWith', `/profile/auth`);
+  });
+
+  /*
+   * - Confirms that a banner is present the child users set up security questions but not have a phone number.
```
mjac0bs, line 94:
```suggestion
   * - Confirms that a banner is present when child users set up security questions but not a phone number.
```

packages/manager/cypress/e2e/core/account/user-verification-banner.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+      .should('be.visible')
+      .should('be.enabled')
+      .click();
+    cy.url().should('endWith', `/profile/auth`);
+  });
+
+  /*
+   * - Confirms that a banner is not shown when the child user sets up security questions.
+   */
+  it('does not show up when a child user adds a phone number or set up security questions', () => {
```
mjac0bs, line 180:
```suggestion
  it('does not show up when a child user adds a phone number and sets up security questions', () => {
```
Phone number *and* questions need to be set up in order for the banner to go away. We're testing this correctly! Description is just inaccurate.

packages/manager/cypress/e2e/core/account/user-verification-banner.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+    ui.button
+      .findByTitle('Add verification details')
+      .should('be.visible')
+      .should('be.enabled')
+      .click();
+    cy.url().should('endWith', `/profile/auth`);
+  });
+
+  /*
+   * - Confirms that a banner is not shown when the child user sets up security questions.
```
mjac0bs, line 178:
```suggestion
   * - Confirms that a banner is not shown when the child user sets up both phone number and security questions.
```
Same comment as above.

packages/manager/cypress/e2e/core/account/user-verification-banner.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+      global: { account_access: 'read_write' },
+    });
+
+    // TODO: Parent/Child - M3-7559 clean up when feature is live in prod and feature flag is removed.
+    mockAppendFeatureFlags({
+      parentChildAccountAccess: makeFeatureFlagData(true),
+    }).as('getFeatureFlags');
+    mockGetFeatureFlagClientstream().as('getClientStream');
+
+    mockGetUsers([mockRestrictedProxyUser]).as('getUsers');
```
mjac0bs, line 55: Nit: if we're not using the aliases in a `cy.wait` call, we could remove them here. (Related: I don't know if not waiting for feature flags could potentially cause any flakiness, which is why we generally do it, or if we are confident it will happen fast enough that it's not very necessary in the first place.)

```suggestion
    mockAppendFeatureFlags({
      parentChildAccountAccess: makeFeatureFlagData(true),
    });
    mockGetFeatureFlagClientstream();

    mockGetUsers([mockRestrictedProxyUser]);
```

mjac0bs (approved):
Thanks for the changes! For completeness, we _could_ test if the banner appears with only questions defined. 

Confirmed this spec runs locally and tests: banner with only phone number defined, banner with neither phone number or question defined, and no banner when both phone number and questions are defined.

Screenshot 2024-03-01 at 7 31 55 AM

Approving pending we clean up a few comments/test descriptions with typos or a slightly inaccurate description of the test. It would also be good to cover that final test.

I'm going to dismiss Jaalah's requested changes since his feedback has been addressed. Rerequesting your reviewers is another thing you can do in the future after addressing feedback - it's helpful for reviewers, especially when someone blocks changes on your PR.

PR #10211, opened 2024-02-20
test: [M3-7434]: Add a Cypress test for OBJ Multicluster Bucket create flow

jdamore-linode:
## Description 📝
Adds a Cypress test to cover the OBJ Bucket creation flow when OBJ Multicluster is enabled. Using mock API data, it confirms that a bucket can be created, that Cloud uses a region ID rather than a cluster ID in its payload, and that the landing page updates to reflect the new bucket.

## Changes  🔄
- Add a test for Bucket creation when OBJ Multicluster feature is enabled
- Update OBJ create mock util to be more consistent with other utils
- Add UI helper to find an autocomplete popper
- Add mock util for OBJ create error

## How to test 🧪
We can rely on CI for this, or `yarn cy:debug` and search for `object-storage.smoke.spec.ts`

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/cypress/e2e/core/objectStorage/object-storage.smoke.spec.ts
+++ b/packages/manager/cypress/e2e/core/objectStorage/object-storage.smoke.spec.ts
@@ -4,6 +4,7 @@
 
 import 'cypress-file-upload';
 import { objectStorageBucketFactory } from 'src/factories/objectStorage';
+import { mockGetRegions } from 'support/intercepts/regions';
 import {
   mockCreateBucket,
   mockDeleteBucket,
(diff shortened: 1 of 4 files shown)


packages/manager/cypress/e2e/core/objectStorage/object-storage.smoke.spec.ts
```diff
@@ -13,11 +14,137 @@ import {
...
+      expect(body.region).to.eq(mockRegionWithObj.id);
+    });
+
+    cy.findByText(mockBucket.label)
+      .should('be.visible')
+      .closest('tr')
+      .within(() => {
+        // TODO Confirm that bucket region is shown in landing page.
+        cy.findByText(mockBucket.hostname).should('be.visible');
+        // cy.findByText(mockRegionWithObj.label).should('be.visible');
```
jdamore-linode, line 144: @cpathipa (cc @dwiley-akamai) is there a ticket to update the landing page so it uses the bucket's `region` when the feature flag is enabled, or would that be part of the M3-7536 ticket? At the moment, nothing is rendered in the "Region" column when the mocked bucket objects don't contain a `cluster` property.

cpathipa: @jdamore-linode Good catch! looks like we missed it, created a ticket M3-7798 to update the buckets landing to use regions instead of clusters.

cpathipa (approved):
LGTM!

dwiley-akamai (approved):
Test passed remotely and locally ✅

PR #10219, opened 2024-02-22
change: [M3-7802] - Remove API caching

bnussman-akamai:
## Description 📝

- Removes all API caching from Cloud Manager
- Creates `Kernel` factory
- Updates `Kernel` type to match what the API actually returns
- Fixes a handful of tests that were built under the assumption that regions would always be loaded into memory

## Why ❓ 
- Caching large amounts of API data increases our bundle size 📦
- The API has gotten faster 💨
- Allows you to run `yarn build` without an internet connection because we no longer need to fetch from the API 🛜
- Simplifies things because there is one less moving part to worry about and maintain 🧹

## What will be removed and why ℹ️ 
- `src/cachedData/typesLegacy.json` because it was unused
- `src/cachedData/kernels.json` because it was only used for unit testing (I replaced it with a factory 🏭)
- `src/cachedData/marketplace.json` because the API now responds much faster (~300ms) (use to take many seconds ⏲️ )
- `src/cachedData/regions.json` because it was an over-optimization, not gitignored, and was misused as mock data - see https://github.com/linode/manager/pull/10191#discussion_r1488642973 (this was recently corrected by @abailly-akamai but I don't want it to happen again in the future) 🚓

## How to test 🧪

- Check out this PR or use the preview link
- Verify region data still populates correctly throughout the app
- Verify marketplace apps load in a reasonable amount of time
- Verify Kernel related features still work (this is found in a Linode's configuration)
- Verify that all automated checks pass

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset


abailly-akamai (approved):
Great! this is a nice cleanup 🧹 

As a result this API update also improved the stackscripts/community tab which was still painfully slow. 🎉 

Noticed no regression ✅ 
Approving pending no e2e complaints ✅

packages/manager/src/features/Kubernetes/ClusterList/KubernetesClusterRow.test.tsx
```diff
@@ -23,13 +25,23 @@ describe('ClusterRow component', () => {
...
+  it('renders a TableRow with label, and region', async () => {
+    server.use(
+      rest.get('*/regions', (req, res, ctx) => {
+        const regions = regionFactory.buildList(1, {
+          id: 'us-central',
+          label: 'Fake Region, NC',
+        });
+        return res(ctx.json(makeResourcePage(regions)));
+      })
+    );
```
hkhalil-akamai, line 37: I'm not sure if I fully understand why these mocks had to be added. Is the mock from `serverHandlers.ts` incorrect (I believe it comes from `regionsData`)? If so, should they be changed/removed from that file?

hkhalil-akamai (approved):
Great clean up 🧼. Approved with one question.

packages/manager/src/features/Kubernetes/ClusterList/KubernetesClusterRow.test.tsx
bnussman: The mocks in `serverHandlers.ts` are the default, but in my opinion, it's best practice to explicitly define any API data that the component depends on in the test itself. This way the unit test will always pass, even if we make changes in `serverHandlers.ts` , which is something we do frequently.

cpathipa (approved):
Thank you for the cleanup @bnussman 
Verified and no regression found.

PR #10226, opened 2024-02-23
fix: [M3-7796] - Correct Default VPC Scope and Adjust PAT Creation Behavior

mjac0bs:
## Description 📝
There is an issue with the current Personal Access Token (PAT) creation process. When using the "Select All" option, it incorrectly selects the VPC scope even when it is disabled. This behavior could lead to unintentional granting of access and needs to be corrected to improve the usability and security of the platform.

## Changes  🔄
- Adds a check and early return in `handleSelectAllScopes` function to explicitly handle the VPC case where `Read Only` is checked, which should instead be `None` for that scope.
- Modifies `allScopesAreTheSame` util to optionally pass in excluded scopes, allowing the correct "Select All" radio button to be checked when some scopes have a different default.
- Adds test coverage to `CreateAPITokenDrawer.test.tsx` and APIToken `utils.test.ts`.

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screenshot 2024-02-20 at 4 32 11 PM | localhost_3000_profile_tokens (4) |
| Screenshot 2024-02-20 at 4 32 50 PM | Screenshot 2024-02-23 at 9 17 41 AM |

### Prerequisites
- Check out this PR and `yarn dev`.

### Reproduction steps
- Go to https://cloud.linode.com/profile/tokens.
- Click the "Select All" read only option and observe that the disabled VPC read only radio button is selected.
- Observe that when the PAT is created, its VPC scope will be Read Only. (This can be seen either in the request body or the View PAT drawer.)

### Verification steps
(How to verify changes)
- Go to https://localhost:3000/profile/tokens.
- Click the "Select All" read only option and verify that radio button is checked even though not all the scopes in that column are read only.
- Verify that the disabled VPC read only radio button is *not* selected. Instead, `None` remains selected.
- Verify that when the PAT is created, its VPC scope will be None. This can be seen either in the request body (the VPC scope will not be included) or the View PAT drawer.

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


bnussman-akamai (approved):

packages/manager/src/features/Profile/APITokens/CreateAPITokenDrawer.tsx
```diff
@@ -165,7 +170,13 @@ export const CreateAPITokenDrawer = (props: Props) => {
   ): void => {
     const value = +e.currentTarget.value;
     const newScopes = form.values.scopes.map(
-      (scope): Permission => [scope[0], value]
+      (scope): Permission => {
+        // VPC does not support Read Only access; default to None.
+        if (scope[0] === 'vpc' && value === 1) {
+          return [scope[0], 0];
+        }
+        return [scope[0], value];
```
bnussman-akamai, line 178: Optional: Not sure exactly how to word this or if it makes any sense for this, but it could be nice if this chunk of code and `excludedScopesFromSelectAll` could pull from one source of truth array/object.

mjac0bs: Does [b15337e](https://github.com/linode/manager/pull/10226/commits/b15337efa4842b702afc1d362cc49c8383c9edd9) make sense?

bnussman-akamai: Yes, I think so!

abailly-akamai (approved):
Thanks for the fix - confirmed using the testing instructions and verifying payload ✅ 

❓ **Question** 

Screenshot 2024-02-29 at 11 47 13

hana-linode (approved):
Confirmed that the disabled VPC Read Only radio button is not selected when selecting all and defaults to None instead. View Scopes drawer is displaying None as expected ✅

mjac0bs:
@abailly-akamai - I went with yes because otherwise it was an odd experience for the user to click the Select All > Read Only and *not* see their action reflected in a checked radio button. All are "Read Only" except any that can't be and have other defaults.

PR #10228, opened 2024-02-23
feat: [M3-7789] – Disable 512GB plans

dwiley-akamai:
## Description 📝
We are temporarily disabling the 512 GB plans for both Dedicated CPU and Premium CPU plan types. This approach aims to direct users towards accessible offerings and ensure clarity in our plan options. To achieve this, we introduce a "placeholder" 512 GB entry in the Linode plans table. This entry is visually distinct (greyed out).

We are also moving away from the "Sold Out" chip in favor of a tooltip informing the user of limited availability + a dismissible banner with a docs link the user can navigate to in order to learn more about plans and availability.

## Changes  🔄
- **Disable 512 GB Plan Entries:** In the Cloud Manager's Linode plans table, the 512 GB plans for Dedicated CPU and Premium CPU types are disabled, making them easily identifiable as not available.
- **Tooltip/Message Clarification:** A tooltip or message is presented when users hover over or engage with the disabled 512 GB plan entry, providing an explanation for its unavailability.
- **Placeholder Entry Creation:** A "placeholder" 512GB plan entry is introduced to the Linode plans table with the specific purpose of being disabled. This entry is designed to be non-selectable and will not initiate any plan.
- **Feature Flag Management:** The activation or deactivation of this disabled option is managed through a feature flag, ensuring flexibility in its implementation.

## Target release date 🗓️
03/18/2024

## Preview 📷
<details>
<summary>Edit: 3/14/24 revised styling of dismissible availability banner</summary>
Bold text and reduced padding around the notice to better align with existing notices

Screenshot 2024-03-14 at 11 01 05 AM
</details>

<details>
<summary>Linode Create</summary>

Dedicated tab:

Screenshot 2024-03-11 at 5 43 26 PM

Premium and GPU tabs:

If Premium/GPU plans are not supported in the selected region...
Screenshot 2024-03-11 at 5 44 24 PM

Screenshot 2024-03-11 at 5 46 56 PM

If Premium/GPU plans are supported in the selected region...
- and half or less of the plans are unavailable...
  - Screenshot 2024-03-11 at 5 45 04 PM

  - Screenshot 2024-03-11 at 5 47 45 PM

- and more than half of the plans are unavailable... (notice the slight change in wording in the dismissible banner, and the lack of help icons in the table)
  - Screenshot 2024-03-11 at 5 56 43 PM

</details>

Tabs in **Linode Resize**, **LKE Create**, and **LKE "Add a Node Pool" drawer** should match what you see for Linode Create (given the same region is selected). The Dedicated tab for **Database Create** should also match up.

The docs link in the dismissible banner should correspond to the plan tab (Dedicated tab --> Dedicated docs link, etc.).

### Prerequisites
- Pull down branch or preview link

### Verification steps for disabled 512 GB plans
- verify Linode Create flow for Dedicated, Premium, and GPU plans
  - with the feature flag for "Gecko" on, select an Edge region and ensure that the dismissible banner from this PR re: plans and availability _is not_ visible (Edge regions only support up to 128GB plans so there's no reason to display the dismissible banner in that situation)
- verify Kubernetes Create flow and "Add a Node Pool" drawer for Dedicated & Premium plans
- verify Databases Create flow for Dedicated plans

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/features/components/PlansPanel/utils.ts
```diff
@@ -194,6 +199,54 @@ export const planTabInfoContent = {
...
+  // For DBaaS
+  const dbaasDedicated512GbIndex = types.findIndex(
+    (type) => type.label === 'DBaaS - Dedicated 512GB'
+  );
+
+  if (isInDatabasesFlow) {
+    if (dbaasDedicated512GbIndex !== -1) {
+      types[dbaasDedicated512GbIndex] = DBAAS_DEDICATED_512_GB_PLAN;
+    } else {
+      types.push(DBAAS_DEDICATED_512_GB_PLAN);
```
dwiley-akamai, line 229: The TS error is coming from this here, it doesn't like this object being pushed to that array

jaalah-akamai: Consider this fixed  ✅

cpathipa (commented):
Small screen is rendering` LA` chip instead of `Sold out` is this expected ?
image

cpathipa (commented):
Tooltip message is not close enough to the `sold out `chip.

image

cpathipa (commented):
Looks like in Premium `sold out` chip is missing for 512gb plan in Linode and  LKE plans table.
Steps: 

- Select the region where premium plans are not available.
- Navigate to Premium tab and notice `sold out` chip missing.

image

jdamore-linode:
Not specific to the changes in this PR, but the pattern of adding a tooltip to the chip in a disabled plan row (or card, on small screens) doesn't work on touch devices, because the chip cannot be interacted with and therefore the tooltip cannot be displayed:

https://github.com/linode/manager/assets/97627410/9e411d9b-d0c8-438a-bf57-1904e4e37f2f

Less crucially, it looks like passing a React node, rather than a string, to the `Tooltip` component causes it to lose its `aria-label` (and makes it unable to render a value for its `data-qa-` attribute, which we can figure out a way around):

*String:*
Screenshot 2024-02-27 at 1 16 09 PM

*Node:*
Screenshot 2024-02-27 at 1 16 18 PM

I don't _think_ there are any accessibility implications here since as far as I can tell there's no way to actually interact with these disabled plan rows in a way that would cause the ARIA label to be announced, but wanted to call it out in case somebody with more knowledge than me has a better understanding of the implications.

packages/manager/src/features/components/PlansPanel/PlanContainer.tsx
```diff
@@ -105,7 +105,7 @@ export const PlanContainer = (props: Props) => {
           disabledClasses={disabledClasses}
           idx={id}
           isCreate={isCreate}
-          isPlanSoldOut={disabled ? false : isPlanSoldOut} // no need to add sold out chip if the whole panel is disabled (meaning that the plan isn't available for the selected region)
+          isPlanSoldOut={disabled ? false : isPlanSoldOut} // no need to add "Sold Out" chip if the whole panel is disabled (meaning that the plan isn't available for the selected region)
```
cpathipa, line 108: If it's not a hard requirement to hide the chip regardless of whether the panel is interactive or not, can we show the chip to maintain consistency?

jdamore-linode:
@jaalah-akamai Pushed a handful of unit tests to confirm that the 512 GB plan selections get disabled -- there's room for more coverage (confirming other plans don't get disabled, confirming it works for "Premium" 512 GB plans in addition to "Dedicated" and that the tooltip message reflects that, Cypress coverage, etc.) and I may work on that some tomorrow, but hoping this helps out some in the meantime

carrillo-erik:
1. The `Sold Out` chip is hard to see in Light Mode. 
2. I am also experiencing the issue mentioned by @cpathipa in which the chip switches from `Sold Out` to `LA` when going from Desktop view to Mobile view.
3. I also ran into the issue @jdamore-linode identified where the chip component does not show the tooltip content in mobile view because the user cannot interact with it.

mobile-view-chip
desktop-view-chip

packages/manager/src/features/Kubernetes/KubernetesPlansPanel/KubernetesPlanContainer.tsx
```diff
@@ -72,7 +72,7 @@ export const KubernetesPlanContainer = (
           disabled={disabled}
           getTypeCount={getTypeCount}
           idx={id}
-          isPlanSoldOut={disabled ? false : isPlanSoldOut} // no need to add sold out chip if the whole panel is disabled (meaning that the plan isn't available for the selected region)
+          isPlanSoldOut={disabled ? false : isPlanSoldOut} // no need to add "Sold Out" chip if the whole panel is disabled (meaning that the plan isn't available for the selected region)
```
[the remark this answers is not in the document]
jaalah-akamai: There are DCs where plans are not available and that's sufficient from a product perspective. In other words, you can't sell out of plans that aren't available.

jaalah-akamai:
@carrillo-erik @cpathipa Unfortunately we had some shifting product decisions today that will solve or make most of these irrelevant. I'm going to move this back into WIP, and any reviews that are still relevant I'll tag you all in.

@jdamore-linode Good call, I have a solution in mind for that.

dwiley-akamai:
The most recent commit brings us to a point of parity when the feature flag is off, preventing a scenario where all plans on a tab are disabled with no explanation:

| Prod (GPU tab) | This branch (GPU branch, `disableLargestGbPlans` flag off) |
| Screenshot 2024-03-12 at 4 41 30 PM | Screenshot 2024-03-12 at 4 41 47 PM |

jdamore-linode:
I'm still having a hard time with the tooltip on touch devices -- I _can_ get it to appear now, but I can't get it to appear consistently. And when it does appear, it sometimes stays indefinitely, and other times seemingly dismisses itself after a second or so:

https://github.com/linode/manager/assets/97627410/039d3f62-4b20-4728-88ab-e40754f66ee6

dwiley-akamai:
> I'm still having a hard time with the tooltip on touch devices -- I _can_ get it to appear now, but I can't get it to appear consistently. And when it does appear, it sometimes stays indefinitely, and other times seemingly dismisses itself after a second or so:

Thanks for checking that @jdamore-linode. Is this something you're only seeing in this section of the app or is it a more general issue with tooltips in the app?

I'm also seeing that `account/user-profile.spec.ts` is failing, looking into that but do you happen to have any insights on it?

jdamore-linode:
> Is this something you're only seeing in this section of the app or is it a more general issue with tooltips in the app?

@dwiley-akamai Other tooltips appear to be working fine*, it looks like it's specific to the tooltip inside the 512 GB plan row.

*(I am seeing a little bit of weirdness with other tooltips where sometimes tapping the icon reveals the tooltip without the icon turning blue, but I'm seeing that in Prod too so definitely not related to this PR)

> I'm also seeing that `account/user-profile.spec.ts` is failing, looking into that but do you happen to have any insights on it?

It seems as though the `data-qa-tooltip` attribute is not being applied to tooltips so Cypress if failing to find it:

Screenshot 2024-03-13 at 4 05 59 PM

This seems to be the cause for most of the test failures, although the `lke-update.spec.ts` failure might be something else (at a glance it looks like _only_ the 512 GB plan is getting rendered in the "Add a Node Pool" drawer) and the `update-firewall.spec.ts` failure looks like it might be a fluke but I'll look more closely if it fails again.

carrillo-erik:
I'm experiencing an issue with the tooltip in mobile view. The tooltip text is not showing in the create Linode flow or the create LKE flow. I tried to activate the tooltip and I can see it change colors upon clicking it, however, the text does not display. Please look at screen recording below.

https://github.com/linode/manager/assets/119514965/1b992b5d-2402-4ff3-b134-0e6db195fb5f

packages/manager/src/features/components/PlansPanel/constants.ts
```diff
@@ -0,0 +1,195 @@
+import type { PlanSelectionType } from './types';
+import type { ExtendedType } from 'src/utilities/extendType';
+
+export const LIMITED_AVAILABILITY_TEXT = 'This plan has limited availability.';
```
carrillo-erik, line 4: Don't we usually have a delimiter on these type of strings? I think this should be `This-plan-has-limited-availability.`

packages/manager/src/components/Tooltip.tsx
```diff
@@ -3,11 +3,30 @@ import React from 'react';
...
+
+interface TooltipNonDataAttribute extends TooltipProps {
+  title: string;
+}
+
+// Discriminated union to enforce `data-qa-tooltip` if title is a ReactNode
+export type EnhancedTooltipProps =
+  | TooltipDataAttribute
+  | TooltipNonDataAttribute;
+
```
bnussman-akamai, line 19: Do we really need all of this?

In my opinion, I don't think we should be enforcing that Toolips have `data-qa-tooltip` under any circumstance.

packages/manager/src/features/Account/Maintenance/MaintenanceTableRow.tsx
```diff
@@ -72,7 +72,7 @@ export const MaintenanceTableRow = (props: AccountMaintenance) => {
       <Hidden lgDown>
         <TableCell>
           {isTruncated ? (
-            <Tooltip title={<HighlightedMarkdown textOrMarkdown={reason} />}>
+            <Tooltip title={reason}>
```
bnussman-akamai, line 75: Should this be reverted?

packages/manager/src/components/Tooltip.tsx
jdamore-linode: Agree -- I get how this came about, but the `data-qa-*` and `data-testid` attributes should be inconsequential to the app itself (i.e. we should be able to remove _all_ of these attributes throughout the app without it causing any build issues or changes in behavior to the app itself)

dwiley-akamai:
> @dwiley-akamai we also need to fix the treatment of this banner, it's inconsistent with others:
> 
> * should be bolded text
> * shouldn't have top and bottom padding
> 
> **New banner** Screenshot 2024-03-14 at 09 16 14
> 
> **Existing banners** Screenshot 2024-03-14 at 09 34 57

We're matching the approved UX wireframes on this, but I will double-check with Lauren on these two points

packages/manager/src/features/components/PlansPanel/constants.ts
dwiley-akamai: I don't believe so, is there an example in the codebase?

carrillo-erik: Actually, I confused this with how we set the `data-testid` strings. Those we use a delimiter.

carrillo-erik:
> It seems as though the `data-qa-tooltip` attribute is not being applied to tooltips so Cypress is failing to find it:
> 
> Screenshot 2024-03-13 at 4 05 59 PM
> This seems to be the cause for most of the test failures, although the `lke-update.spec.ts` failure might be something else (at a glance it looks like _only_ the 512 GB plan is getting rendered in the "Add a Node Pool" drawer) and the `update-firewall.spec.ts` failure looks like it might be a fluke but I'll look more closely if it fails again.

@jdamore-linode
I wonder if this has to do with how the string for `data-qa-tooltip` is being set. My understanding is that `data-*` attributes values should consist of delimited strings which is consistent to how we set values for `data-testid`. It could be that it's reading each word in the string as a separate value for the `data-qa-tooltip`. Will need further research on this matter.

Screenshot 2024-03-14 at 9 47 23 AM

packages/manager/src/components/Tooltip.tsx
jaalah-akamai: This may break some tests. However `title` should almost always be a string now since we determined `ReactNode` (which was used for links) was inaccessible.

dwiley-akamai: @jaalah-akamai Locally I undid these and the related `dataAttrs` changes throughout the diff, the currently failing E2E tests pass aside from `lke-update.spec.ts` (possibly unrelated) and there are no TS errors. Would it be objectionable to reintroduce the discriminated union and associated changes as part of a separate ticket?

packages/manager/cypress/e2e/core/kubernetes/lke-update.spec.ts
dwiley-akamai, line 1: Per the video of the run, the test was failing because the Dedicated tab consisted of just the disabled 512 GB plan, so this code brings the test to the Shared CPU tab instead so it could proceed as it typically does.

carrillo-erik: Ok, I was just running the test locally and noticed that as well. It would get stuck trying to click the `+` symbol but it was disabled since the 512GB plan is not available.

Screenshot 2024-03-14 at 2 58 26 PM

jaalah-akamai (approved):

bnussman-akamai:
Functionality is looking good so far. Going to keep reviewing!

UX feedback: I think the `Learn more about plans and availability.` dismissible banner is kind of extra. I feel like that should just be included as normal copy. I don't think we should be warning our users to "Learn more"

bnussman-akamai (approved):

jaalah-akamai:
> Functionality is looking good so far. Going to keep reviewing!
> 
> UX feedback: I think the `Learn more about plans and availability.` dismissible banner is kind of extra. I feel like that should just be included as normal copy. I don't think we should be warning our users to "Learn more"

This is something the team is aware of and will follow up in a future release if they feel it needs to be changed - thanks for the feedback though

abailly-akamai (approved):
Approving pending as a follow up

- improvement of the tooltip accessibility/behavior for cards
- Improved coverage (e2e & new units)
- naming conventions (`mostClassPlansAreLimitedAvailability`, `isMajorityLimitedAvailabilityPlans` `types` & `_types`, `plans` & ` _plans`, `plansForThisLinodeTypeClass` & `_plansForThisLinodeTypeClass` etc)

Also agree the reasoning behind making the banner dismissible is dubious, not to mention inconsistent with other banners in the plan panel

PR #10229, opened 2024-02-23
feat: [M3-7801] - Reintroduce NVMe Volume Upgrades

bnussman-akamai:
## Description 📝

- Reintroduce NVMe Volume Upgrades because we still have one region (Fremont, CA 🇺🇸) with HDD storage 💾 
- We had some code left-over, but I had to reimplement some missing pieces. I tried made the UX much better while I was at it! 💻 

## Target release date 🗓️

3/5/2024

## Preview 📷

https://github.com/linode/manager/assets/115251059/d35cc9b3-8e02-4074-8bae-b3cb25cfc757

## How to test 🧪

- For now, I recommend just viewing the Cypress test I wrote and checking my logic throughout (use `yarn cy:debug`)
- I will follow up if there is a better way to test this complex flow.

- [ ] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/features/Linodes/LinodesDetail/LinodesDetailHeader/UpgradeVolumesDialog.tsx
```diff
@@ -1,71 +1,65 @@
...
 import { Typography } from 'src/components/Typography';
 import { VolumeUpgradeCopy } from 'src/features/Volumes/UpgradeVolumeDialog';
-import { queryKey } from 'src/queries/accountNotifications';
+import { getUpgradeableVolumeIds } from 'src/features/Volumes/utils';
+import { useNotificationsQuery } from 'src/queries/accountNotifications';
+import { useLinodeVolumesQuery } from 'src/queries/volumes';
 import { useVolumesMigrateMutation } from 'src/queries/volumesMigrations';
 import { getAPIErrorOrDefault } from 'src/utilities/errorUtils';
 
-import { ExtendedLinode } from '../types';
```
hkhalil-akamai, line 17: Nice 🧼

We should be able to delete `types.ts` now

packages/manager/src/features/Volumes/VolumeTableRow.tsx
```diff
@@ -85,25 +106,36 @@ export const VolumeTableRow = React.memo((props: Props) => {
           }}
         >
           {volume.label}
+          {isEligibleForUpgradeToNVMe && (
+            <Chip
+              clickable
+              label="UPGRADE TO NVMe"
```
hkhalil-akamai, line 112: Curious about the capitalization here; if we're doing all caps, would we want to spell it as 'NVME' (even though it is ordinarily 'NVMe')?

hkhalil-akamai: Related, is there precedent for making something like this all caps as opposed to "Upgrade to NVMe"?

packages/manager/src/features/Volumes/VolumeTableRow.tsx
```diff
@@ -85,25 +106,36 @@ export const VolumeTableRow = React.memo((props: Props) => {
...
+              onClick={handleUpgrade}
+              size="small"
+            />
+          )}
+          {isNVMeUpgradeImminent && !isVolumeMigrating && (
+            <Chip color="default" label="UPGRADE PENDING" size="small" />
```
hkhalil-akamai, line 118: Same comment as above

packages/manager/src/features/Volumes/utils.test.ts
hkhalil-akamai, line 1: Thanks for adding unit tests!

packages/manager/cypress/e2e/core/volumes/upgrade-volume.spec.ts
hkhalil-akamai, line 1: Thanks for including this E2E!

hkhalil-akamai (approved):
Great work!

packages/manager/src/features/Linodes/LinodesDetail/LinodesDetailHeader/UpgradeVolumesDialog.tsx
bnussman-akamai: Ah, good catch! That's great

packages/manager/src/features/Linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
```diff
@@ -225,6 +228,7 @@ const LinodeDetailHeader = () => {
       <HostMaintenance linodeStatus={linode?.status ?? 'running'} />
       <MutationNotification linodeId={matchedLinodeId} />
       <Notifications />
+      <VolumesUpgradeBanner linodeId={linode.id} />
```
jaalah-akamai, line 231: Do we need optional chaining here like we're doing elsewhere?

jaalah-akamai (approved):
Thanks for all the unit/cypress tests

packages/manager/src/features/Linodes/LinodesDetail/LinodesDetailHeader/LinodeDetailHeader.tsx
bnussman-akamai: We don't need it because `linode` will always be defined when this is rendered. 

I cleaned up the other optional chains in 076a6e379a

mjac0bs (approved):
Super smooth testing instructions, thank you for that!

Confirmed that the whole migration can happen without a refresh to the page. Also confirmed this is the case for unattached vs attached volumes, as well as on the Linode Details > Storage tab.


Screenshot 2024-02-27 at 7 52 58 AM

Very minor UX thing (maybe related to the recent Chips PR?): on mobile screens, the chips can get a little too close to the volume label and could do with some margin.
Screenshot 2024-02-27 at 7 45 42 AM
Screenshot 2024-02-27 at 7 56 57 AM

packages/manager/src/features/Volumes/utils.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  it('should return null if the API does not return a percentage', () => {
+    const event = eventFactory.build({
+      percent_complete: null,
+      status: 'started',
+    });
+    expect(getEventProgress(event)).toBe(null);
+  });
+
+  it('should return a formatted percenatge if the API returns a percentage and the event is "started"', () => {
```
jdamore-linode, line 51:
```suggestion
  it('should return a formatted percentage if the API returns a percentage and the event is "started"', () => {
```

packages/manager/src/features/Volumes/utils.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    const event = eventFactory.build({
+      percent_complete: 25,
+      status: 'started',
+    });
+    expect(getEventProgress(event)).toBe('(25%)');
+  });
+});
+
+describe('getUpgradeableVolumeIds', () => {
+  it('should return the id of volumes that have a corosponding upgrade notification', () => {
```
jdamore-linode, line 61:
```suggestion
  it('should return the id of volumes that have a corresponding upgrade notification', () => {
```

packages/manager/src/features/Volumes/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+    event.percent_complete === null ||
+    event.status !== 'started'
+  ) {
+    return null;
+  }
+
+  return `(${event.percent_complete}%)`;
+};
+
+export const getUpgradeableVolumeIds = (
```
jdamore-linode, line 59:
```suggestion
/**
 * Returns an array of IDs of Volumes that are scheduled to be upgraded.
 * 
 * @param volumes - Array from which to retrieve upgradeable Volumes.
 * @param notifications - Notifications containing Volume migration statuses.
 * 
 * @returns Array of upgradeable Volume IDs.
 */
export const getUpgradeableVolumeIds = (
```
Mind if we add doc comments for this util? (Not sure how accurate the suggestion above is)

jdamore-linode (approved):
Awesome work, @bnussman-akamai! Posted a couple suggestions for the tests but nothing major. Thanks for taking this on!

packages/manager/src/features/Volumes/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+        (notification) =>
+          notification.entity?.id === volume.id &&
+          notification.type === 'volume_migration_scheduled'
+      )
+    ) {
+      upgradeableVolumeIds.push(volume.id);
+    }
+  }
+
+  return upgradeableVolumeIds;
```
jdamore-linode, line 85: Definite nitpick, but I think this could be optimized by filtering out any non-`volume_migration_scheduled` notifications before iterating:

```suggestion
  return notifications
    .filter((notification) => notification.type === 'volume_migration_scheduled')
    .reduce((volumeIds: number[], notification: Notification) => {
      const upgradeableVolume = volumes.find((volume) => volume.id === notification.entity?.id);
      if (upgradeableVolume) {
        volumeIds.push(upgradeableVolume.id);
      }
      return volumeIds;
    }, []);
```

I did a quick experiment by adapting your unit test to call the function 100,000 times against arrays of 100 volumes and 250 notifications, and got ~3.5s with the function above vs ~8s with the original implementation. (I did this purely out of curiosity -- I think there's exactly zero chance this function ever becomes a bottleneck, but on the other hand it felt silly to scrap this, so just thought I'd share 🤷‍♂️)

bnussman-akamai: Great call. 

I didn't stress too much about perf because I believe Linodes can only have 8 volumes max, but this is much better.

PR #10245, opened 2024-03-01
test: Resolve OBJ create and delete E2E test flake

jdamore-linode:
## Description 📝
This resolves the test failures that we've recently been seeing in `object-storage.e2e.spec.ts`. There are a couple issues at play that contribute to these failures:

- After a Bucket has been created, the landing page re-renders (presumably in response to some API request resolving), causing occasional failures when the re-render occurs after Cypress has located some element (e.g. the "Delete" table row button) but before it has interacted with it. (I think this is irrelevant from a user experience point of view)
- Less frequently, the test fails due to an issue with the Buckets landing page where the page fails to display a newly created Bucket if the Bucket was created before the initial requests to fetch Buckets have all completed (see M3-7833)

This PR only seeks to resolve test failures related to these behaviors, and it does so by waiting for all related API requests to resolve before attempting to create a Bucket. Because Cloud makes an indeterminate number of requests to fetch Buckets, the test waits for the loader to disappear instead of intercepting and waiting on the requests directly.

I was able to reproduce these failures easily on `develop`, and after making these changes I ran the test on repeat 50 times and all 50 runs passed.

## Changes  🔄
- Wait for loader to disappear and for network utilization data to be fetched before creating an OBJ bucket

## How to test 🧪
We can lean on CI for this, but you can use this command to run the tests locally:

```bash
yarn cy:run -s "cypress/e2e/core/objectStorage/object-storage.e2e.spec.ts"
```

If you want to run them repeatedly, you can use this command if you're using `zsh` as your shell:

```bash
repeat 5 yarn cy:run -s "cypress/e2e/core/objectStorage/object-storage.e2e.spec.ts"
```

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

### DIFF
--- a/packages/manager/.changeset/pr-10245-tests-1709311899662.md
+++ b/packages/manager/.changeset/pr-10245-tests-1709311899662.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Tests
+---
+
+Resolve OBJ Bucket create/delete E2E test flake ([#10245](https://github.com/linode/manager/pull/10245))
--- a/packages/manager/cypress/e2e/core/objectStorage/object-storage.e2e.spec.ts
+++ b/packages/manager/cypress/e2e/core/objectStorage/object-storage.e2e.spec.ts
@@ -6,6 +6,7 @@ import 'cypress-file-upload';
 import { createBucket } from '@linode/api-v4/lib/object-storage';
 import { objectStorageBucketFactory } from 'src/factories';
 import { authenticate } from 'support/api/authentication';
+import { interceptGetNetworkUtilization } from 'support/intercepts/account';
 import {
   interceptCreateBucket,
   interceptDeleteBucket,
@@ -120,6 +121,7 @@ describe('object storage end-to-end tests', () => {
     interceptGetBuckets().as('getBuckets');
     interceptCreateBucket().as('createBucket');
     interceptDeleteBucket(bucketLabel, bucketCluster).as('deleteBucket');
+    interceptGetNetworkUtilization().as('getNetworkUtilization');
 
     mockAppendFeatureFlags({
       objMultiCluster: makeFeatureFlagData(false),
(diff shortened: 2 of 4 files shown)


jaalah-akamai:
Screenshot 2024-03-01 at 2 00 27 PM

jdamore-linode:
> (Test failure picture)

Thanks @jaalah-akamai, pushed a fix for that! Very glad you caught that as it would've been a guaranteed failure for all of our test accounts with no buckets 🙏

bnussman-akamai (approved):

mjac0bs (approved):
The `repeat` command is useful, thanks for including that!

This test passed consistently for me, and your 50x good runs are good assurance that the flake is resolved. 🚀

mjac0bs:
One note, though: in some subsequent runs I'd kicked off in the background after that initial 5, I've seen this failure occasionally (twice within ~20 runs):

image

I haven't seen it when watching the test run in debug mode, didn't get to the video output before it was overwritten, and not sure why this would be happening if we've got our `wait` for `@uploadObject`, anyway. 🤔

jdamore-linode:
> One note, though: in some subsequent runs I'd kicked off in the background after that initial 5, I've seen this failure occasionally (twice within ~20 runs):
> 
> (Screenshot failure)
> 
> I haven't seen it when watching the test run in debug mode, didn't get to the video output before it was overwritten, and not sure why this would be happening if we've got our `wait` for `@uploadObject`, anyway. 🤔

Thanks @mjac0bs! I'll take a look at this before merging, and if I can reproduce it I'll try to fix it 👍 I was using `cy.only` when I was testing my changes so this totally fell under the radar.

(Also worth pointing out that I'm experimenting with some Jenkins pipeline changes using this PR, so if you see unusual results for the `cloud-manager-tests` run, that's probably why!)

jdamore-linode:
@mjac0bs I've been totally unable to reproduce this locally (tried something like 80 times 😭) but I did find a recording of it happening in Cypress. I wrote up M3-7850 to revisit this soon!

PR #10248, opened 2024-03-01
fix: [M3-7269] - Display parent/proxy username in user menu and toast if no company name is available

mjac0bs:
## Description 📝
We learned on the backend that `company` may be `null` for some users, even if they payByAkamai. (Though awaiting final confirmation that this won't change going forward and can be retroactively applied to parent/child accounts when the relationship made on the backend.) Assuming company can still be `null`,  we won't always be able to display the company under the helper text "You are currently logged in as:" in the user menu dropdown for parent and proxy users.

This PR falls back on username in those cases, so we are no longer rendering an empty string in the UserMenu dropdown and success toast.

**UPDATE (3/4)**: Marked as Do Not Merge and Requires Changes because:

- Product wants us to fall back on email rather than username (pending final confirmation)
   - We'd need to ensure parent email cannot be updated to null
- The child account list in the Switch Account Drawer uses company name; we would also need to replace with email

## Changes  🔄
- For a proxy user, we fall back on proxy username rather than parent username + child company if we can’t display the child company name.
- For a parent user, we fall back on parent username if we can’t display the parent company name.
- Updates test coverage in `UserMenu.test.tsx` to reflect this. 

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-03-01 at 11 35 24 AM | Screenshot 2024-03-01 at 11 30 15 AM |
| Screenshot 2024-03-01 at 7 05 36 AM | Screenshot 2024-03-01 at 11 30 30 AM |

### Prerequisites
- This can be tested with mocks more easily than in the dev environment because you'd need to ask me for provisioned parent/child account access, then log into alpha admin and delete a child account's company name, and make one manual code edit due to an API bug with `child_account_access` (the grant isn't being returned correctly) to eliminate the final `!isChildAccountAccessRestricted` check from `const canSwitchBetweenParentOrProxyAccount` if you want the Switch Account button to show up.

### Reproduction steps
- Look at the diff and notice how this code before was defaulting to `company` being an empty string if the conditional was false, and then displaying that empty string in the user menu and tooltip.

### Verification steps
- Check out this PR and `yarn up`.
- Ensure the parent/child feature flag is on.
With mocks:
- Edit the account endpoint [here](https://github.com/linode/manager/blob/develop/packages/manager/src/mocks/serverHandlers.ts#L1201) in serverHandlers.ts to return a `null` company name.  Confirm your `user_type` is [`parent`](https://github.com/linode/manager/blob/develop/packages/manager/src/mocks/serverHandlers.ts#L558).
- Open the user menu and confirm the username is visible. Click the Switch Account button and switch into a proxy account by selecting a child.
- Edit the profile endpoint in serverHandlers.ts to have a `user_type` of [`proxy`](https://github.com/linode/manager/blob/develop/packages/manager/src/mocks/serverHandlers.ts#L558).
- Confirm the success toast includes the username.
- Open the user menu and confirm the username is visible.
- (Note that it will be the same username ("mock-user") since mocks will use the same profile endpoint.)

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/TopMenu/UserMenu/UserMenu.tsx
+++ b/packages/manager/src/features/TopMenu/UserMenu/UserMenu.tsx
@@ -82,11 +82,11 @@ export const UserMenu = React.memo(() => {
   const open = Boolean(anchorEl);
   const id = open ? 'user-menu-popover' : undefined;
   const companyName =
-    (hasParentChildAccountAccess &&
-      profile?.user_type !== 'default' &&
-      account?.company) ??
-    '';
-  const showCompanyName = hasParentChildAccountAccess && companyName;
+    hasParentChildAccountAccess &&
+    profile?.user_type !== 'default' &&
+    account?.company
+      ? account.company
+      : undefined;
   const { isParentTokenExpired } = useParentTokenManagement({ isProxyUser });
 
   // Used for fetching parent profile and account data by making a request with the parent's token.
@@ -99,9 +99,12 @@ export const UserMenu = React.memo(() => {
 
   const { data: parentProfile } = useProfile({ headers: proxyHeaders });
 
+  // For a proxy user, if there is no company name to identify a child account, fall back on the proxy username.
   const userName =
-    (hasParentChildAccountAccess && isProxyUser ? parentProfile : profile)
-      ?.username ?? '';
+    (hasParentChildAccountAccess && isProxyUser && companyName
+      ? parentProfile
+      : profile
+    )?.username ?? '';
 
   const matchesSmDown = useMediaQuery((theme: Theme) =>
     theme.breakpoints.down('sm')
@@ -118,7 +121,7 @@ export const UserMenu = React.memo(() => {
   React.useEffect(() => {
     // Run after we've switched to a proxy user.
     if (isProxyUser) {
-      enqueueSnackbar(`Account switched to ${companyName}.`, {
+      enqueueSnackbar(`Account switched to ${companyName ?? userName}.`, {
         variant: 'success',
       });
     }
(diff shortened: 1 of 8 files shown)


packages/manager/src/features/TopMenu/UserMenu/UserMenu.tsx
```diff
@@ -290,7 +295,7 @@ export const UserMenu = React.memo(() => {
           <Box>
             <Heading>My Profile</Heading>
             <Divider color="#9ea4ae" />
-            <Grid container>
+            <Grid columnGap={userName.length >= 29 ? 1 : undefined} container>
```
mjac0bs, line 298: Intended to fix the spacing issue with the second column of links being too close to the first when there is a long username. Open to suggestions if you can do this in a better way than hardcoding the length of the username.

Screenshot 2024-03-01 at 11 31 37 AM

jaalah-akamai: Let me think about this, but everything else looks good from testing ✅

jaalah-akamai (approved):

packages/manager/src/features/TopMenu/UserMenu/UserMenu.tsx
jaalah-akamai: Remove `columnGap` and the `rowGap` on each of the grid items. Instead do `columnSpacing={2} rowSpacing={1}` on the container grid and that will fix things.

Screenshot 2024-03-01 at 7 52 02 PM

hkhalil-akamai (approved):
Looks good when testing with mocks. Unable to verify account switching toasts.

mjac0bs:
Closing this PR as it has been determined that we do not need a fallback for companyName. Elsewhere on the backend, we were incorrectly retrieving that data from a value that could be optional. This will be changed on the backend to a value that is required, so the companyName will always exist.

mjac0bs:
For posterity: we reopened this PR after finding an edge case where if a restricted parent account has no Billing Access but has child_account_access, we can’t access the account/ endpoint to grab the company. However, with the enabled grant, they do have account switching access, so we had a blank in the user menu again.

Following guidance from Product, we can fall back on parent email here.

Issue:
Screenshot 2024-03-08 at 9 01 19 AM

Fix:
Screenshot 2024-03-08 at 10 00 40 AM

mjac0bs:
We received approval on this from Product so I'm going to go ahead and merge this ahead of the demo since it is a bug that otherwise will be visible.

PR #10249, opened 2024-03-04
fix: React Query `updateInPaginatedStore` helper function not working as expected

bnussman-akamai:
## Description 📝

With React Query, it is reccomneded that you update the cache in an "immutable" way. 

Essentially, we were updating the cache, but the changes were not getting propagated and rendered because updating an existing value within the same array. The fix in this case to just copy the array before committing to the cache. 

See https://tanstack.com/query/v4/docs/framework/react/guides/updates-from-mutation-responses#immutability

> [!note]
> This PR fixes Cypress test failures in `machine-image-upload.spec.ts`

## How to test 🧪

- Run `yarn cy:debug` and watch the `uploads machine image, mock finish event` test. After `@getImage`, we should see the changes render immediately and the test should pass.

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/queries/base.ts
+++ b/packages/manager/src/queries/base.ts
@@ -244,16 +244,16 @@ export const updateInPaginatedStore = <T extends { id: number | string }>(
         return oldData;
       }
 
-      const updatedDATA = [...oldData.data];
+      const updatedPaginatedData = [...oldData.data];
 
-      updatedDATA[toUpdateIndex] = {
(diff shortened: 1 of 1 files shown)


jdamore-linode (approved):
Thanks @bnussman-akamai! Confirmed that the Image status issue is resolved. Thanks for the explanation + TanStack docs link!

mjac0bs (approved):
Observed the spec pass as described. Thanks for fixing this! 🚢

PR #10257, opened 2024-03-05
upcoming: [M3-7613] - Grants and Permissions for Placement Groups

carrillo-erik:
## Description 📝
This PR sets up the Grants and Permissions for users to be able to create, delete, and update Placement Groups. More specifically, write access to create a Placement Group will be dictated by the `add_linodes` grant. Access to assign/unassign an individual Linode to a Placement Group will be based on having `read_write` access to the specific Linode.

## Changes  🔄
- The majority of the changes revolve around the `useRestrictedGlobalGrantCheck` hook. Once it's determined if the user has the `add_linodes` grant, certain components will be hidden from the user or disabled to the user; and in other cases, we display components to convey a message to the user.
- The other change in this PR comes from the `useIsResourceRestricted` hook. This time, we are interested to know if a user has `read_write` access for specific Linodes in order to assign (add) them to a Placement Group.

## Preview 📷
Screenshot 2024-03-11 at 10 07 00 AM
Screenshot 2024-03-11 at 9 59 39 AM
Screenshot 2024-03-11 at 10 00 48 AM
Screenshot 2024-03-11 at 9 59 14 AM

### Prerequisites
- Enable the following using the Cloud Manager Dev Tools:
    - Placement Groups feature flag
    - MSW
- Read the comments in the `serverHandlers.ts` file on line 559 and line 574.

### Reproduction steps
- Given that we are still working with API mocks, the best way to verify the UI is to follow the instructions from the last bullet point in the `How to test` section above. Kudos to @mjac0bs for this solution!
  - In the alternative, one can negate the `isLinodeReadOnly` value in the files touched by this PR. 
- As a consequence of the previous point, some actions like deleting a Placement Group or unassigning a Linode will not take effect.
- Testing the functionality of the `useIsResourceRestricted` hook in the current state proved to be difficult since it would require having a testing account which we can set the grants and permissions, then log into that account while using the MSW. After the API hits the `alpha` environment, this will be more thoroughly tested.
- The scenario where a user may not have the necessary access to update Placement Groups but have knowledge of a url route was considered, and can be verified by negating the `isLinodeReadOnly` value on the Placement Groups landing page and then visiting `http://localhost:3000/placement-groups/create` or `http://localhost:3000/placement-groups/1/linodes`

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/components/LandingHeader/LandingHeaderProps.tsx
mjac0bs, line 1: What's this file for? We're exporting LandingHeaderProps in `LandingHeader.tsx`.

packages/manager/src/features/PlacementGroups/PlacementGroupsDetail/PlacementGroupsLinodes/PlacementGroupsLinodesTableRow.tsx
```diff
@@ -46,6 +53,7 @@ export const PlacementGroupsLinodesTableRow = React.memo((props: Props) => {
       <TableCell actionCell>
         <InlineMenuAction
           actionText="Unassign"
+          disabled={isAllowedToEditPlacementGroup}
```
mjac0bs, line 56: The naming of this variable and the way that this util works feels confusing here.

`isAllowedToEditPlacementGroup` should return `true` if that linode has `read_write` permissions, so that makes sense. But then we'd be saying that the "Unassign" button should be disabled if the user has read_write abilities, and otherwise enabled? That sounds like it would be the opposite of what we want, so we're missing the `!`.

```suggestion
          disabled={!isAllowedToEditPlacementGroup}
```

packages/manager/src/features/PlacementGroups/PlacementGroupsUnassignModal.tsx
```diff
@@ -48,10 +50,16 @@ export const PlacementGroupsUnassignModal = (props: Props) => {
...
+    grantLevel: 'read_write',
+    grantType: 'linode',
+    id: +linodeId,
+  });
+
   const actions = (
     <ActionsPanel
       primaryButtonProps={{
-        disabled: isLoading,
+        disabled: isLoading || isAllowedToEditPlacementGroup,
```
mjac0bs, line 62: Same comment here as above for disabling.

packages/manager/src/features/PlacementGroups/PlacementGroupsCreateDrawer.tsx
mjac0bs, line 1: Have you considered disabling the form fields in this drawer, as well? We did ended up doing this for parent/child when a child user can't access the Edit Billing Info Drawer, and it looks pretty nice as far as a unified disabled experience:

Screenshot 2024-03-12 at 3 42 28 PM

mjac0bs (commented):
Nice job with pointing out the drawer can be reached by URL in the testing instructions. 

Left some feedback in comments, and I also wanted to mention how I tested at least the read only linodes part with a couple updates to the mocks that allow us to mock `add_linodes` as `false`. 

```
  rest.get('*/profile', (req, res, ctx) => {
    const profile = profileFactory.build({
      restricted: true, // This is important! The grants endpoint is only called for restricted users!
      user_type: 'default',
    });
    return res(ctx.json(profile));
  }),
  rest.get('*/profile/grants', (req, res, ctx) => {
    return res(
      ctx.json(grantsFactory.build({ global: { add_linodes: false }},))
    );
  }),

```

abailly-akamai:
@jaalah-akamai would love to have your eyes on this initial permission PR to confirm the overall direction before I dive into a review. thx!

packages/manager/src/features/PlacementGroups/PlacementGroupsLanding/PlacementGroupsLanding.tsx
```diff
@@ -116,6 +123,7 @@ export const PlacementGroupsLanding = React.memo(() => {
     <>
       <LandingHeader
         breadcrumbProps={{ pathname: '/placement-groups' }}
+        disabledCreateButton={isLinodeReadOnly}
```
jaalah-akamai, line 126: We want to display tooltip text here if disabled:
```
buttonDataAttrs={{
  tooltipText: getRestrictedResourceText({
    action: 'create',
    isSingular: false,
    resourceType: 'PlacementGroups',
  }),
}}
```

carrillo-erik: @jaalah-akamai Are we doing this for all disabled button states now? Also, are you implying just the tooltip icon or icon with explanation?

packages/manager/src/features/PlacementGroups/PlacementGroupsCreateDrawer.tsx
carrillo-erik: @mjac0bs I did considered what you've suggested. We shouldn't give users a false sense of capability by allowing them to fill in the form, only to disable the CTA. I will do this in my next push.

packages/manager/src/features/PlacementGroups/PlacementGroupsCreateDrawer.tsx
```diff
@@ -131,6 +133,17 @@ export const PlacementGroupsCreateDrawer = (
       open={open}
       title="Create Placement Group"
     >
+      {disabledCreateButton && (
+        <Notice
+          text={getRestrictedResourceText({
+            action: 'edit',
+            resourceType: 'PlacementGroups',
```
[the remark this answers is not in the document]
carrillo-erik: @mjac0bs Looking at the `grantTypeMap` object, which is where these grant type values are defined, maybe the following will work: `placementGroups: 'Placement Groups',`

These values are only read by the utility function to generate the copy for the Notice. I don't see how including the space in the string would create issues elsewhere.

packages/manager/src/components/LandingHeader/LandingHeaderProps.tsx
carrillo-erik: This is weird, my editor must have done some auto-generated thingy because I manually did not create this file. Git history shows it was introduced by this PR. Honestly, this is weird. In any event, I will delete this file.

packages/manager/src/features/PlacementGroups/PlacementGroupsDetail/PlacementGroupsLinodes/PlacementGroupsLinodesTableRow.tsx
carrillo-erik: Correct, I will update.

packages/manager/src/features/PlacementGroups/PlacementGroupsUnassignModal.tsx
carrillo-erik: Will update.

packages/manager/src/features/PlacementGroups/PlacementGroupsLanding/PlacementGroupsLanding.tsx
jaalah-akamai: The tooltip won't display if there's no tooltip text. Not for all disabled states, because some flows are disabled and display a notice which is sufficient enough to where we don't need button tooltip text

jaalah-akamai (approved):

packages/manager/src/features/PlacementGroups/PlacementGroupsCreateDrawer.tsx
mjac0bs: Thanks for the updates! Did UX have any guidance about making this an important error notice versus just an error notice. I think (but may be wrong) that we don't often make error notices important in drawers, and we didn't for Parent/Child. Achieving consistency would be best here, as we improve disabled states throughout the app, so could be worth bringing up with them.

Screenshot 2024-03-18 at 11 48 17 AM

packages/manager/src/features/PlacementGroups/PlacementGroupsCreateDrawer.tsx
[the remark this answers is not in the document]
mjac0bs: Yep, good call! 🚢

mjac0bs (commented):
Addressed feedback looks good. Couple more things:

As far as URLs, it looks like a user with read only access could still potentially navigate to the edit drawer and delete modal via URL with the placement group id, so these UIs would need disabled state too.

e.g. http://localhost:3000/placement-groups/edit/1
Screenshot 2024-03-18 at 11 42 09 AM

e.g. http://localhost:3000/placement-groups/delete/1
Screenshot 2024-03-18 at 11 51 23 AM

packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -556,6 +556,10 @@ export const handlers = [
       restricted: false,
       // Parent/Child: switch the `user_type` depending on what account view you need to mock.
       user_type: 'parent',
+      // PLACEMENT GROUPS TESTING - Permissions and Grants:
+      // Uncommnet the two lines below: This is important! The grants endpoint is only called for restricted users.
```
mjac0bs, line 560:
```suggestion
      // Uncomment the two lines below: This is important! The grants endpoint is only called for restricted users.
```

(+ if you want to keep these commented changes for easier testing with mocks in the future, we'll need to resolve the merge conflicts stemming from the [MSW update PR](https://github.com/linode/manager/pull/10285))

packages/manager/src/components/LandingHeader/LandingHeader.tsx
```diff
@@ -19,6 +19,7 @@ export interface LandingHeaderProps {
   buttonDataAttrs?: { [key: string]: boolean | string };
   createButtonText?: string;
   disabledCreateButton?: boolean;
+  disableEditButton?: boolean;
```
abailly-akamai, line 22: this should read `disabledBreadcrumbEditButton` for clarity since Landing header has other CTAs

packages/manager/src/features/PlacementGroups/PlacementGroupsCreateDrawer.tsx
```diff
@@ -146,7 +159,7 @@ export const PlacementGroupsCreateDrawer = (
               autoFocus: true,
             }}
             aria-label="Label for the Placement Group"
-            disabled={false}
+            disabled={disabledCreateButton || false}
```
abailly-akamai, line 162: same here - the naming convention could lead to confusion.`disabledCreateButton` implies only the button and disabling the form using this variable isn't optimal. The button can also be disabled for other reasons limits) so we could run into logic problems down the line.

How about renaming `disabledCreateButton` with `disablePlacementGroupCreation` (or similar) where it applies?

abailly-akamai (commented):
@carrillo-erik please rebase at your earliest convenience so we can rely on serverHandler to test the PR, thx.

packages/manager/src/features/PlacementGroups/PlacementGroupsCreateDrawer.tsx
carrillo-erik: I removed the `important` prop, there was no formal discussion around this notice. Rather we implemented it on the fly. There is a separate notice where we do show the icon in the mock ups but that's inconsistent as you mentioned, for now I also removed it from that component.

packages/manager/src/mocks/serverHandlers.ts
carrillo-erik: Thanks for catching this typo. The merge conflicts have been resolved.

packages/manager/src/components/LandingHeader/LandingHeader.tsx
carrillo-erik: Good call. The recent code push reflects this change.

packages/manager/src/features/PlacementGroups/PlacementGroupsCreateDrawer.tsx
carrillo-erik: Same as the other comment.

packages/manager/src/features/PlacementGroups/PlacementGroupsCreateDrawer.tsx
jaalah: Yea moving forward `important` will only be for top level pages, not drawers 👍

jaalah (approved):

jaalah-akamai (approved):

packages/manager/src/features/PlacementGroups/PlacementGroupsLanding/PlacementGroupsLanding.tsx
carrillo-erik: The tooltip text is now shown to the user as suggested.

mjac0bs (approved):
✅  Confirmed that the edit drawer and delete modal, when accessed via URL, are disabled as expected when the user has read only grant access.

Thanks for the update to the error notice in the create drawer, too.

| Cannot Edit | Cannot Delete |
| --- | --- |
| Screenshot 2024-03-22 at 1 56 31 PM |Screenshot 2024-03-22 at 1 56 18 PM |

PR #10237, opened 2024-02-28
upcoming: [M3-7816] - Hide "Switch Account" buttons if child_account_access is false

jaalah-akamai:
## Description 📝
For parent account users (non-admin), we need to ensure Switch Account button doesn't show if they don't have the child_account_access grant.

## Changes  🔄
- Checking for `child_account_access` as part of condition
- Added unit test for UserMenu

## Target release date 🗓️
03/04/2024

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screenshot 2024-02-28 at 12 06 24 PM | Screenshot 2024-02-28 at 12 06 47 PM |
| Screenshot 2024-02-28 at 12 06 31 PM | Screenshot 2024-02-28 at 12 06 52 PM |

### Prerequisites
- Create a parent account
- Create a non-admin user on the account
- Ensure this new users `child_account_access` grant is set to `false`

### Reproduction steps
- Go to http://localhost:3000/account/billing
- Observe that without changes you can see the Switch Account buttons

### Verification steps
- Checkout local changes or preview link
- Observe that the Switch Account button is not rendered

## As an Author I have considered 🤔

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ❔ Our [contribution guidelines](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md)
- [ ] 🤏 Splitting feature into small PRs
- [x] ➕ Adding a changeset
- [x] 🧪 Providing/Improving test coverage
- [ ] 🔐 Removing all sensitive information from the code and PR description
- [x] 🚩 Using a feature flag to protect the release
- [ ] 👣 Providing comprehensive reproduction steps
- [ ] 📑 Providing or updating our documentation
- [ ] 🕛 Scheduling a pair reviewing session
- [ ] 📱 Providing mobile support
- [ ] ♿  Providing accessibility support

PR #10266, opened 2024-03-07
upcoming: [M3-7816-v2] - Adjust logic for when to show Switch Account button

jaalah-akamai:
## Description 📝
We had to adjust the logic for `isChildAccountAccessRestricted` to parent users only since this is where the grant can be toggled. This was being applied to the proxy user inadvertently preventing the proxy from switching to accounts.

## Changes  🔄
- Added E2E test for this use-case

## Target release date 🗓️
3/18

## Screenshots
Screenshot 2024-03-07 at 2 32 50 PM

## How to test 🧪
See testing instructions: https://github.com/linode/manager/pull/10237

### DIFF
--- a/packages/manager/cypress/e2e/core/parentChild/account-switching.spec.ts
+++ b/packages/manager/cypress/e2e/core/parentChild/account-switching.spec.ts
@@ -306,7 +306,7 @@ describe('Parent/Child account switching', () => {
 
       cy.visitWithLogin('/account/billing');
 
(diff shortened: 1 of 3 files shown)


jaalah-akamai:
I may actually add two more tests just to check the user menu.

bnussman-akamai (approved):

mjac0bs (approved):

PR #10269, opened 2024-03-08
upcoming: [M3-7855] - Updates to PATs for Restricted Parent Users

jaalah-akamai:
## Description 📝
As a Restricted Parent User, I should not be able to create Personal Access Tokens (PATs) with the 'child_account_access' grant if I do not possess the 'child_account_access' permission. This ensures that only users with the appropriate permissions can generate PATs with the 'child_account_access' grant. 
 
When viewing scopes, if the child_account_access grant for the parent user is later revoked, we'll still display this scope for existing PATs, but not for new ones.

## Target release date 🗓️
3/18

## Preview 📷
| Video |
| --- |
| <video src="https://github.com/linode/manager/assets/125309814/4d783e12-341c-4d92-9171-d63090ca8d49"/> |

### Prerequisites
- Log into a unrestricted parent user account
- Log into a restricted parent user account (incognito)

### Reproduction steps
- Currently, when viewing scopes the `child account access` grant always appears
- Currently, when viewing the scope action menu in mobile view, a help icon/tooltip always appears when it should only appear for proxy users

### Verification steps
- Follow steps in video preview
- Make sure to check the `PUT /token` request payload in addition to what the drawer is/isn't displaying. 
- Then turn the feature flag *off* and confirm that when creating a PAT with all scopes set to Read Only, the `child_account` scope is not included in `scopes` in the payload. 
- Change screen width to mobile view and, on any non-proxy account, confirm that the help icon/tooltip is not visible

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset


packages/manager/src/features/Profile/APITokens/CreateAPITokenDrawer.tsx
```diff
@@ -203,6 +212,13 @@ export const CreateAPITokenDrawer = (props: Props) => {
       defaultAccessLevel: 0,
       invalidAccessLevels: [1],
       name: 'vpc',
+      optionallyInclude: false,
+    },
+    {
```
mjac0bs, line 217: I don't think we need to do this *if* we change [L181](https://github.com/linode/manager/pull/10269/files#diff-d179f0d0f9fc3644a265f97bc10ba57f5e177c4c96b9960e7b9f779bb7c724aeR181) to `(showFilteredPermissions ? filteredPermissions : allPermissions).map` because that will use the filtered scopes without Child Account Access when Selecting All, and not add back in the Child Account Access scope.

If the intent here was something else: can we add a comment for clarity? It wasn't initially obvious to me what was happening and I wrote this interface for `vpc` recently. 😅

mjac0bs (commented):
Going to hold off approving for now because I'm questioning whether we need the `optionallyInclude` prop and additions to the excluded scopes array + test changes, as long as we use the correctly filtered perms rather than all perms.

As far as functionality, though, I confirmed that as a restricted parent user with the child_account_access grant *disabled*:
- I could not create a PAT with the Child Account Access scope - it was not displayed and was not submitted in PUT request 
- I could see the Child Account Access scope when viewing *previously created* tokens created prior to when child_account_access was revoked
- When access as restored via the grant, I could once again create a PAT with Child Account Access and I could view whether the grant was enabled on all tokens

packages/manager/src/features/Profile/APITokens/CreateAPITokenDrawer.tsx
```diff
@@ -128,7 +135,7 @@ export const CreateAPITokenDrawer = (props: Props) => {
     },
     {
       name: 'child_account',
-      shouldBeIncluded: hasParentChildAccountAccess,
+      shouldBeIncluded: hasParentChildAccountAccess && isParentUser,
```
mjac0bs, line 138: Our TODO comment isn't accurate if we're checking for `isParentUser` here, is it? We'd still want to continue to manipulate the perm name map. Unless:

If we were to omit the `isParentUser` here and eventually not manipulate `basePermNameMap` when we release the feature, I think we would be able to remove 
[L217-221](https://github.com/linode/manager/pull/10269/files#diff-d179f0d0f9fc3644a265f97bc10ba57f5e177c4c96b9960e7b9f779bb7c724aeR217-R221) entirely, at that point. On L238-242, we'd always filter that child account access scope out for all users except parent accounts with child account access on. Then it's not available in the scopes list, so we can't possibly select it with "select all", and it doesn't need to be in the excludedScopesFromSelectAll array. 

Right now, it seems like we're filtering twice. 'Child Account Access' is being filtered out of `basePermNameMap` for all non-parent users by `shouldBeIncluded: hasParentChildAccountAccess && isParentUser`, so it can't be used to filter allPermissions on L242. We then end up with `child_account` in the `filteredPermissions` and have to include it in `excludedScopesFromSelectAll`.

packages/manager/src/features/Profile/APITokens/CreateAPITokenDrawer.tsx
```diff
@@ -204,6 +214,11 @@ export const CreateAPITokenDrawer = (props: Props) => {
       invalidAccessLevels: [1],
       name: 'vpc',
     },
+    {
+      defaultAccessLevel: 0,
+      invalidAccessLevels: [1, 2],
```
mjac0bs, line 219: `excludedScopesFromSelectAll` is really intended to be for scopes that have some options disabled with defaults set instead, and therefore they're excluded from select all, so this isn't working quite as intended for a parent user with child account access. They should be able "select all" and include the child access scope. My earlier suggestion was misleading because it would require that we not include based on `isParentUser` and then remove this from the array -- see the comment about the TODO.

Current behavior:

https://github.com/linode/manager/assets/114685994/3ba58642-f29d-4fdb-945a-d44f70cab0c4

cpathipa (approved):
Functionality looks good. Left a minor comment for improvement.

jaalah-akamai:
@mjac0bs 👍

mjac0bs (approved):
I spent a while testing this because there are so many accounts that can be in so many states.  With the current changes, I am seeing:

- Unrestricted parents see the Child Account Access scope in Create drawer and it is included in payload, visible in View drawer
- Restricted parents with the grant see the scope in Create drawer and it is included in payload, visible in View drawer
- Restricted parents without the grant do NOT see the scope in Create drawer, it is NOT included in the payload, and it is only visible in the View drawer if they previously created it with RO/RW
- When the parent/child feature flag is OFF, all users do NOT see the scope in either the Create or View drawer and it is NOT included in the payload

PR #10274, opened 2024-03-11
change: [M3-7860] - Source ACLB region info from API data and use Jakarta instead of Sydney

bnussman-akamai:
## Description 📝

- Updates ACLB to pull region information from API data instead of using hardcoded region data wherever possible 🗺️ 
- Updates the hardcoded beta regions to use Jakarta 🇮🇩 instead of Sydney 🇦🇺 

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-03-11 at 4 15 10 PM | Screenshot 2024-03-11 at 4 15 00 PM |
| Screenshot 2024-03-11 at 4 15 52 PM | Screenshot 2024-03-11 at 4 16 04 PM | 
| Screenshot 2024-03-11 at 4 17 55 PM | Screenshot 2024-03-11 at 4 17 45 PM | 
| | ⬆️ Yes this is expected. The dev environment does not have these regions so we intentionally fallback to a blank flag and region ID | 

## How to test 🧪

> [!note]
> Behavior will be different in dev compared to production because dev lacks certain regions in the API response.
> You should test this PR using both dev and production 

### Testing dev
- Login to `dev-test-aglb`
- Verify you see region IDs listed on the landing page, details page, and settings page
- Verify the create page lists the proper regions (uses Jakarta 🇮🇩 instead of Sydney 🇦🇺)

Yes, this is expected:
Screenshot 2024-03-11 at 4 23 14 PM

### Testing staging/production
- Login to `prod-test-065`
- Verify you see region ID and labels on the landing page table
- Verify you see region ID and labels, and country flags on the details page and settings page
- Verify the create page lists the proper regions (uses Jakarta 🇮🇩 instead of Sydney 🇦🇺)

Screenshot 2024-03-11 at 4 24 52 PM

- [ ] 👀 Doing a self review
- [x] ➕ Adding a changeset


jaalah-akamai (approved):
I do find it weird that at times we'll only display the region id, but I'm sure that's temporary?

✅ Confirmed empty flag state in dev and that Jakarta (id-cgk) displays instead of Sidney
✅ Confirmed flags display in prod account

bnussman-akamai:
Yeah, it's not great @jaalah-akamai. 

It's more of a configuration problem. The VMs that run the underlying ACLBs  are deployed in regions that don't exist in alpha. What really should have happened is that alpha ACLBs should be using alpha only regions (basically just Newark), but they chose to deploy alpha ACLBs nodes using production Linodes.

We could continue to hardcode the values to keep alpha looking nice but I don't think we should be doing that to cover up for other team's design choices.

carrillo-erik (approved):

PR #10227, opened 2024-02-23
refactor: [M3-7687] - Linodes Restricted User Experience 2/2

jaalah-akamai:
## Description 📝
Overall improvement to the Linode restricted experience. This PR disables lot's of buttons and fields that a user doesn't have access to.

## Changes  🔄
- Removed the banners from beneath each tab on the Linodes detail page since we will display an overall banner at the top of the page.
- If you don't have `add_linodes` grant, we disable create buttons. If you do have the grant, you will be able to initiate the creation of Linodes, as well as manage those you have personally established. However, restrictions may still apply for modifying of Linodes configured by the administrator.

## Target release date 🗓️
03/04/2024

### BEFORE
| | |
| ------- | ------- |
| Screenshot 2024-02-26 at 5 01 59 PM | Screenshot 2024-02-26 at 5 02 24 PM |
| Screenshot 2024-02-26 at 5 02 40 PM | Screenshot 2024-02-26 at 5 02 07 PM |

### AFTER
| | |
| ------- | ------- |
|Screenshot 2024-02-26 at 5 09 06 PM | Screenshot 2024-02-26 at 5 08 27 PM |
| Screenshot 2024-02-26 at 5 14 58 PM | Screenshot 2024-02-26 at 5 15 33 PM |
|Screenshot 2024-02-26 at 5 20 57 PM| |

### Prerequisites
(How to setup test environment)
- Log into two accounts side by side:
   - An unrestricted admin user account: full access
   - A restricted user account (use Incognito for this)
       - Start with `Read Only` for everything

### Reproduction steps
(How to reproduce the issue, if applicable)
- Create a linode from an unrestricted admin account and give your restricted user view only access.
- Observe that before changes, a user can attempt to edit/modify linodes that they don't have access to edit/modify and they will get denied by the API. This is what we're looking to prevent.

### Verification steps
(How to verify changes)
- Verify that restricted users can access only what they have permissions for.
- Verify that navigating via URL still display correct banners

## As an Author I have considered 🤔

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ❔ Our [contribution guidelines](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md)
- [x] 🤏 Splitting feature into small PRs
- [ ] ➕ Adding a changeset
- [ ] 🧪 Providing/Improving test coverage
- [ ] 🔐 Removing all sensitive information from the code and PR description
- [ ] 🚩 Using a feature flag to protect the release
[...]

PR #10275, opened 2024-03-11
fix: [M3-7862] - Fix display of TagsPanel in read-only entities

hkhalil-akamai:
## Description 📝
#10227 disables the `TagPanel` component for read-only entities. However, this change would not work correctly for non-Linode entities, such as domains and nodebalancers, since it was hard-coded to `grantType: 'linode'`.

## Changes  🔄
- Fixes hard-coded `grantType` value
- Moves `useIsResourceRestricted` out of `TagsPanel` and into each respective component

## Target release date 🗓️
3/18/2024

### Prerequisites
- Create and log in to a restricted user with a read-only Domain and NodeBalancer.

### Reproduction steps
- While logged in to the restricted user, navigate to the read-only Domain/NodeBalancer
- Observe that the Add Tag button is not disabled
- Observe that trying to add a tag results in an Unauthorized Error.

### Verification steps
- Follow reproduction steps above
- Verify that Add Tag button is disabled

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset


packages/manager/src/components/TagsPanel/TagsPanel.tsx
```diff
@@ -66,12 +61,6 @@ export const TagsPanel = (props: TagsPanelProps) => {
     isLoading: userTagsLoading,
   } = useTagSuggestions(!profile?.restricted);
 
-  const isLinodesGrantReadOnly = useIsResourceRestricted({
-    grantLevel: 'read_only',
-    grantType: 'linode',
-    id: entityId,
-  });
-
```
hkhalil-akamai, line 74: It's better to call this hook in each respective component instead of cluttering `TagsPanel` with unrelated logic.

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeSummaryPanel.tsx
hkhalil-akamai, line 1: It seems like restricted users can't view Kubernetes clusters at all, so the logic in this file may have no effect.

dwiley-akamai (commented):
For Domains, should we provide some sort of indication as to why the tags are disabled? For NodeBalancers, we have a notice at the top so the reason can at least be ascertained. Maybe explanatory hover text could be appropriate here in the Domain case? Alternatively, we could give a similar treatment to Domains, with a notice at the top (this could be a separate ticket).

| NodeBalancer  | Domain |
| Screenshot 2024-03-13 at 11 19 26 AM | Screenshot 2024-03-13 at 11 20 46 AM |

hkhalil-akamai:
> For Domains, should we provide some sort of indication as to why the tags are disabled? For NodeBalancers, we have a notice at the top so the reason can at least be ascertained. Maybe explanatory hover text could be appropriate here in the Domain case? Alternatively, we could give a similar treatment to Domains, with a notice at the top (this could be a separate ticket).

Going to defer to @jaalah-akamai, who's leading this project -- maybe this should be a follow-up change?

jaalah-akamai:
> For Domains, should we provide some sort of indication as to why the tags are disabled?

@dwiley-akamai No because we're going to update the domains section next, so it will also have a banner. We're currently working our way through each feature or resource entity.

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeSummaryPanel.tsx
dwiley-akamai, line 1: If the read-only logic isn't doing anything for LKE, we should remove the code IMO. Or if API will support it in the future, add a `@todo` placeholder about it and comment the grant-related code out

dwiley-akamai (approved):
Disabled `TagsPanel` for restricted users on read-only entities for Linodes (card view & on Details page), Domains, and NodeBalancers ✅

jaalah-akamai (approved):

hana-linode (commented):
There seems to be an issue loading tags for a restricted user that has r/w Nodebalancer access

https://github.com/linode/manager/assets/115299789/813702d7-7ba4-4cdc-a20d-96fe894e6ac5

hkhalil-akamai:
> There seems to be an issue loading tags for a restricted user that has r/w Nodebalancer access
> 
>  Screen.Recording.2024-03-22.at.10.55.46.AM.mov

Good catch! Seems to be caused by the way [react query v4 defines `isLoading`](https://github.com/linode/manager/pull/10236#discussion_r1506614349). Addressed in [`c14561`](https://github.com/linode/manager/pull/10275/commits/c14561fe4a679901154d2d749e4fa09490b38298)

PR #10276, opened 2024-03-12
fix: [M3-7826-v2] - Fix email displaying in top menu for all users without a company name

mjac0bs:
## Description 📝
In #10248, I didn't test the regular user experience thoroughly enough, because the conditional logic was defaulting to the profile email of the user if no company name was set, even if the user was not a parent/child/proxy account. We don't want this. 

We **should** see the following:
- For a regular (default) user, there should be no changes to the top menu and user menu (i.e. it should display username only). 
- For parent/child/proxy users, the top menu should display username and company names. 
- For parent/proxy users, the user menu dropdown should display the company name.
- For **restricted parent users who do not have `account_access` set to at least `read_only`**, the top menu should display username and email (in place of company).
For **restricted parent users who do not have `account_access` set to at least `read_only`**, the dropdown menu should display email (in place of company).

## Changes  🔄
- Fixed the conditional logic to check if the user is a parent user and is missing an account company in order to resort to email, otherwise defaults to undefined.
- Added the other piece of unit test coverage that was missing for this edge-case, confirming a regular user does not see their email in the top menu or dropdown user menu.

## Target release date 🗓️
3/18

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-03-12 at 8 53 37 AM | Screenshot 2024-03-12 at 8 52 57 AM |

### Prerequisites
- You can use the MSW to toggle the different user types, or you can use real parent/child/proxy and default user accounts (ask me). 

### Reproduction steps
- Check out develop.
- Observe for your regular user account, the top menu now includes your email if you don't have a company name associated with your account.

### Verification steps
(How to verify changes)
- Check out this branch. 
- With mocks off: confirm that the top menu and user menu dropdown include *only* your username (no email). There should be no change from prod.
- With mocks on: update the `user_type` to `parent`, `proxy`, `child`, and `default` in the [MSW](https://github.com/linode/manager/blob/develop/packages/manager/src/mocks/serverHandlers.ts#L558). For a restricted parent user, you'd want to mock a 403 on the [account](https://github.com/linode/manager/blob/develop/packages/manager/src/mocks/serverHandlers.ts#L1203) endpoint.
```
      rest.get('*/account', (req, res, ctx) => {
        return res(ctx.status(403));
      }),
```
- For each user_type, confirm that the top menu and user menu match the description outlined in the PR description. 

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset - Note: this was a bug fix to a bug fix, and did not ever make it into a release so I left the changeset off. 

### DIFF
--- a/packages/manager/src/features/TopMenu/UserMenu/UserMenu.tsx
+++ b/packages/manager/src/features/TopMenu/UserMenu/UserMenu.tsx
@@ -27,6 +27,8 @@ import { useGrants, useProfile } from 'src/queries/profile';
 import { sendSwitchAccountEvent } from 'src/utilities/analytics';
 import { getStorage } from 'src/utilities/storage';
 
+import { getCompanyNameOrEmail } from './utils';
+
 interface MenuLink {
   display: string;
   hide?: boolean;
@@ -81,15 +83,11 @@ export const UserMenu = React.memo(() => {
   const open = Boolean(anchorEl);
   const id = open ? 'user-menu-popover' : undefined;
 
-  // For parent users lacking `account_access`: without a company name to identify an account, fall back on the email.
-  const companyNameOrEmail =
-    hasParentChildAccountAccess &&
-    profile?.user_type !== 'default' &&
-    account?.company
-      ? account.company
-      : isParentUser && profile?.email
-      ? profile.email
-      : undefined;
+  const companyNameOrEmail = getCompanyNameOrEmail({
+    company: account?.company,
+    isParentChildFeatureEnabled: hasParentChildAccountAccess,
+    profile,
+  });
 
   const { isParentTokenExpired } = useParentTokenManagement({ isProxyUser });
 
--- a/packages/manager/src/features/TopMenu/UserMenu/UserMenu.tsx
+++ b/packages/manager/src/features/TopMenu/UserMenu/UserMenu.tsx
@@ -81,14 +81,15 @@ export const UserMenu = React.memo(() => {
   const open = Boolean(anchorEl);
   const id = open ? 'user-menu-popover' : undefined;
 
-  // If there is no company name to identify an account, fall back on the email.
-  // Covers an edge case in which a restricted parent user without `account_access` cannot access the account company.
+  // For parent users lacking `account_access`: without a company name to identify an account, fall back on the email.
   const companyNameOrEmail =
     hasParentChildAccountAccess &&
     profile?.user_type !== 'default' &&
     account?.company
       ? account.company
-      : profile?.email;
+      : isParentUser && profile?.email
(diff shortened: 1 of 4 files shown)


jaalah-akamai (approved):
Thanks for the tests! These changes are now working - not seeing email above my normal user accounts

packages/manager/src/features/TopMenu/UserMenu/UserMenu.tsx
```diff
@@ -81,14 +81,15 @@ export const UserMenu = React.memo(() => {
...
+  // For parent users lacking `account_access`: without a company name to identify an account, fall back on the email.
   const companyNameOrEmail =
     hasParentChildAccountAccess &&
     profile?.user_type !== 'default' &&
     account?.company
       ? account.company
-      : profile?.email;
+      : isParentUser && profile?.email
+      ? profile.email
+      : undefined;
```
bnussman-akamai, line 92: Should we break this out into a util function and unit test it?

bnussman-akamai (approved):

packages/manager/src/features/TopMenu/UserMenu/UserMenu.tsx
bnussman-akamai: Sounds good! I think both testing approaches are fine, but I do think a helper function using if/else would be more readable than the ternaries here.

Feel free to revisit later if it will keep parent/child unblocked

abailly-akamai (approved):

packages/manager/src/features/TopMenu/UserMenu/utils.ts
```diff
@@ -25,9 +25,11 @@ export const getCompanyNameOrEmail = ({
   }
 
   // For parent users lacking `account_access`: without a company name to identify an account, fall back on the email.
-  // In all other parent/child/proxy cases, company will be available, as it is a required field.
+  // We do not need to do this for child users lacking `account_access` because we do not need to display the email.
```
mjac0bs, line 28: Trying to confirm this with the team. 

Difference is just a matter of top menu consistency; we're not in danger of rendering any blanks.

RestrictedChildWithEmailWenNoCompanyName
RestrictedChildWithOnlyUsernameWhenNoCompanyName

packages/manager/src/features/TopMenu/UserMenu/UserMenu.tsx
mjac0bs: @bnussman-akamai Done in [2a259d0](https://github.com/linode/manager/pull/10276/commits/2a259d054628a87115622109879d0ea0b3d09162).

PR #10281, opened 2024-03-14
upcoming: [M3-7869] - Linode Create Refactor - Part 2 - Images and Distributions

bnussman-akamai:
## Description 📝

Part 2 of the Linode Create refactor effort 🎉 

## Changes  🔄
- Introduces **Tabs** (Distribution, Marketplace , StackScripts, Backups, Clone)
- Introduces some **helpers for managing query params** (Linode Create relies somewhat heavily on query params)
- Completely **implements the Distribution and Image Tabs**
- Implements a new **`ImageSelectv2`**
  - Yes, this does expand this scope of this project, and yes, it does add a 3rd ImageSelect to the codebase but we'd eventually write this code anyway. The current LinodeSelect literally has a `Paper` baked into it 🙄 
- Adds `LandingHeader` to match the existing Linode Create
- Switches use of `watch` to `useWatch` to fix other components re-rendering when changing `region`
- Configure MUI Autocomplete Group Label theme so that it looks more like our old component

## Preview 📷

Putting a before and after so you can how the refactor is progressing 🤩

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-03-13 at 8 07 33 PM | Screenshot 2024-03-13 at 8 07 04 PM |

## How to test 🧪

- Manually test the new functionality
- Check for unnecessary renders
- Look and my code and roast it 😳
  - I want this new code to be super clean and maintainable. If I'm doing something questionable, please call me out! 📣 

- [x] 👀 Doing a self review
- [ ] ➕ Adding a changeset


packages/manager/src/features/Linodes/LinodeCreatev2/Plan.tsx
```diff
@@ -12,16 +12,17 @@ import { extendType } from 'src/utilities/extendType';
...
+  const regionId = useWatch({ control, name: 'region' });
+
   const { field, fieldState } = useController<CreateLinodeRequest>({
     name: 'type',
   });
 
   const { data: regions } = useRegionsQuery();
   const { data: types } = useAllTypes();
 
-  const regionId = watch('region');
```
bnussman-akamai, line 23: Turns out, this is very bad.

It resulted in the entire form re-rendering when the region changed. With this new flow, I'm going to focus on performance and fine-grain rendering. 

Changing out this `watch` call for a `useWatch` makes it so that **only** this component subscribes to `region` changing. 


This is what makes `react-hook-form` great. They build things in a way so that you have to be very intentional with what state updates you subscribe to.

packages/manager/src/features/Linodes/LinodeCreatev2/index.tsx
```diff
@@ -37,11 +44,45 @@ export const LinodeCreatev2 = () => {
...
+              </SafeTabPanel>
+              <SafeTabPanel index={1}>Marketplace</SafeTabPanel>
+              <SafeTabPanel index={2}>StackScripts</SafeTabPanel>
+              <SafeTabPanel index={3}>
+                <Images />
+              </SafeTabPanel>
+              <SafeTabPanel index={4}>Bckups</SafeTabPanel>
+              <SafeTabPanel index={5}>Clone Linode</SafeTabPanel>
+            </TabPanels>
+          </Tabs>
```
bnussman-akamai, line 85: I debated moving this chunk into another component to keep this root file super clean but I decided keeping it here helps your brain match up the code with the UI. 

Curious to hear what others think!

packages/manager/src/components/ImageSelectv2/ImageSelectv2.tsx
```diff
@@ -0,0 +1,65 @@
...
+    getAPIFilterForImageSelect(variant)
+  );
+
+  // We can't filter out Kubernetes images using the API so we filter them here
+  const options = getFilteredImagesForImageSelect(images, variant);
+
+  const value = images?.find((i) => i.id === props.value);
+
+  return (
+    <Autocomplete
```
hkhalil-akamai, line 46: Would it be possible to display the selected distro's icon (to match the current functionality)? This might be possible using `startAdornment`

Screenshot 2024-03-18 at 3 25 15 PM

packages/manager/src/features/Linodes/LinodeCreatev2/index.tsx
```diff
@@ -25,8 +34,6 @@ export const LinodeCreatev2 = () => {
 
       history.push(`/linodes/${linode.id}`);
     } catch (errors) {
-      // @todo this is temporary API error handling. We will develop a more
-      // robust helper that can convert API errors to react-hook-form errors
```
hkhalil-akamai, line 29: What is the reason for removing this todo?

packages/manager/src/features/Linodes/LinodeCreatev2/index.tsx
hkhalil-akamai: I'm fine with it this way, the file remains pretty short and readable.

packages/manager/src/features/Linodes/LinodeCreatev2/index.tsx
```diff
@@ -37,11 +44,45 @@ export const LinodeCreatev2 = () => {
...
+            index={currentTabIndex}
+            onChange={(index) => updateParams({ type: tabs[index] })}
+          >
+            <TabList>
+              <Tab>Distributions</Tab>
+              <Tab>Marketplace</Tab>
+              <Tab>StackScripts</Tab>
+              <Tab>Images</Tab>
+              <Tab>Backups</Tab>
+              <Tab>Clone Linode</Tab>
```
hkhalil-akamai, line 71: Would this be better expressed as a `tabs.map`?

hkhalil-akamai (approved):
Changes look good in testing and code is clean and readable. Left a few minor comments. Great to see this work proceed @bnussman-akamai!

packages/manager/src/features/Linodes/LinodeCreatev2/index.tsx
bnussman-akamai: I thought about it, but because this page is likely not going to change much, I think keeping it static is okay.

packages/manager/src/components/ImageSelectv2/ImageSelectv2.tsx
bnussman-akamai: Yeah, I should do that for the 1:1 parity with the old component.

I left it simple for now because I prefer the clean look 😖

packages/manager/src/components/DistributionIcon.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  const className = distribution
+    ? `fl-${distroIcons[distribution] ?? 'tux'}`
+    : `fl-tux`;
+
+  return (
+    <i
+      className={className}
+      data-testid="distro-icon"
+      style={{ fontSize: size ?? '1.8em' }}
```
abailly-akamai, line 34: Do we want the icon size to be relative to the parent? I am not saying we don't, i am frankly wondering cause sometimes it can cause unwanted side effects. Since we can override the size am wondering about using px instead for the default. No big deal either way

packages/manager/src/components/ImageSelect/ImageOption.tsx
```diff
@@ -1,8 +1,8 @@
 import DescriptionOutlinedIcon from '@mui/icons-material/DescriptionOutlined';
 import { Theme } from '@mui/material/styles';
-import { makeStyles } from 'tss-react/mui';
 import * as React from 'react';
 import { OptionProps } from 'react-select';
+import { makeStyles } from 'tss-react/mui';
```
abailly-akamai, line 5: ?

packages/manager/src/components/ImageSelect/ImageSelect.tsx
```diff
@@ -137,6 +137,9 @@ const isMemo = (prevProps: ImageSelectProps, nextProps: ImageSelectProps) => {
   );
 };
 
+/**
+ * @deprecated Start using ImageSelectv2 when possible
+ */
```
abailly-akamai, line 142: Nice 👍

packages/manager/src/components/ImageSelectv2/ImageSelectv2.tsx
abailly-akamai, line 1: Extending the scope is fine, but let's keep going with our recommanded practices? This one definitely warrants a unit test and to be in Storybook since we do that for our custom selects (LinodeSelect, RegionSelect etc). A bit of of overhead but worth it IMO.

packages/manager/src/components/ImageSelectv2/utilities.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Using API filter, I can't think of a way to filter out
+ * LKE images that we don't want customers to see.
+ *
+ * This function exists to filter out public Kubernetes images
+ * for the Images Select.
+ *
+ * Please use API filtering (getAPIFilterForImageSelect) when possible!
+ */
```
abailly-akamai, line 31: What is the original behavior here (with legacy `ImageSelect`)?

packages/manager/src/features/Linodes/LinodeCreatev2/index.tsx
abailly-akamai: ya, it's fine to keep it here.

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Hook that allows you to read and manage Linode Create flow query params.
+ *
+ * We have this because react-router-dom's query strings are not typesafe.
+ */
+export const useLinodeCreateQueryParams = () => {
+  const history = useHistory();
+
+  const rawParams = getQueryParamsFromQueryString(history.location.search);
```
abailly-akamai, line 22: Any reason you are not using `useParams`?

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const updateParams = (params: Partial<LinodeCreateQueryParams>) => {
+    const newParams = new URLSearchParams({ ...rawParams, ...params });
+    history.replace({ search: newParams.toString() });
```
abailly-akamai, line 26: `history.replace` will prevent the browser from remember the history, therefore not allowing the browser back and forward buttons. This introduces regressions with the current behavior. `history.push` is probably what you want here.

abailly-akamai (commented):
Nice! super clean - some improvement to be made but the direction is great as far as i can see the scaffolding! 

Apart from my comments, the Image Select Option still needs some TLC (alignment w/ icon, padding, hover styles)
Screenshot 2024-03-19 at 09 57 43
Screenshot 2024-03-19 at 09 56 22

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
bnussman-akamai: I'm going to 1:1 parity with the existing create flow and it only uses query params. I _think_ `useParams` is only for router path params

packages/manager/src/components/ImageSelectv2/utilities.ts
bnussman-akamai: `ImageSelect` just had intense client side filtering / sorting

abailly-akamai: right right just trying to understand if anything is changing to the logic

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
bnussman-akamai: Good catch! Thanks! 4a3e081041

packages/manager/src/components/ImageSelect/ImageOption.tsx
bnussman-akamai: `perfectionist/sort-imports` 🤷

packages/manager/src/components/ImageSelectv2/ImageSelectv2.tsx
bnussman-akamai: Added some testing in 1bf9aa5

packages/manager/src/components/DistributionIcon.tsx
bnussman-akamai: I updated this component in 5981137

It will now just inherit the fontSize and have no default.

packages/manager/src/components/ImageSelectv2/utilities.ts
bnussman-akamai: Yes, there are changes to logic. `packages/manager/src/components/ImageSelect/ImageSelect.tsx` does a lot of crazy stuff and this new component likely doesn't work exactly like the old one. 

For example, there is an API bug that is breaking the ordering on this new component. I filed an ARB for it already!

This is something we can iterate on as we need other functionality

packages/manager/src/features/Linodes/LinodeCreatev2/index.tsx
hkhalil-akamai: My only concern is with `currentTabIndex` going out of sync, but as you said with these not changing often its unlikely.

packages/manager/src/components/ImageSelectv2/utilities.ts
dwiley-akamai: @bnussman-akamai Does the bug you're referring to cover the reason why, for example, openSUSE images get listed under separate headers?

packages/manager/src/features/Linodes/LinodeCreatev2/Plan.tsx
```diff
@@ -47,7 +48,7 @@ export const Plan = () => {
       selectedId={field.value}
       selectedRegionID={regionId}
       showTransfer
-      types={types?.map(extendType) ?? []}
+      types={types?.map(extendType) ?? []} // @todo don't extend type
```
dwiley-akamai, line 51: Maybe a notation like `@TODO Linode Create Refactor` to make it easy to track these down during this refactoring effort?

packages/manager/src/components/ImageSelectv2/ImageSelectv2.tsx
dwiley-akamai: Keeping it vs. not keeping it might be worth raising during Cafe this week

dwiley-akamai (commented):
Will circle back to wrap up my review tomorrow morning!

packages/manager/src/components/ImageSelectv2/utilities.ts
bnussman-akamai: Yes exactly that! @dwiley-akamai

packages/manager/src/features/Linodes/LinodeCreatev2/Plan.tsx
bnussman-akamai: I think I might leave it as a basic todo because that change isn't within the scope of this project.

abailly-akamai (approved):
Thanks for addressing the changes!

PR is good to go from my end, for future ones I'd still love to see out of scope commits (such as the new ImageSelect) decoupled in a separate PR - especially since it seems like this is prompting the most questions on this PR and diverting the attention from the create flow refactor. My two cents!

PR #10294, opened 2024-03-18
chore: [M3-7896] - Typescript v5

bnussman-akamai:
## Description 📝

- This is an effort to get us more up to date on our typescript version. We've been holding this off for a long time now 😖
- What makes this PR hard and ugly is the fact that our previous config and version allowed many un-typesafe things so I did the minimum to fix those instances
- Our `eslint` setup is pretty crusty. Some eslint packages are not compatible with the latest typescript versions so I'll either fix or workaround this for now (can we please just delete it all and use Biome😩)

## How to test 🧪

- Check the diff and verify 
  - No applicable logic was changed or broken
  - Type changes make sense 

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset


packages/manager/.eslintrc.cjs
```diff
@@ -107,7 +107,6 @@ module.exports = {
     'cypress',
     'prettier',
     'testing-library',
-    'scanjs-rules',
```
bnussman-akamai, line 110: This eslint plugin is deprecated. I don't think we need to keep it around anyway. It isn't providing a ton of value.

PR #10296, opened 2024-03-19
test: [M3-7508] - Add tests to check Parent and Child Close Account flows

cliu-akamai:
## Description 📝
Add regression tests to check Parent/Child Close Account flows

## Major Changes 🔄
- Add new tests to tests to check Parent/Child Close Account flows

## How to test 🧪
```
yarn cy:run -s "cypress/e2e/core/account/account-cancellation.spec.ts"
```

### DIFF
--- a/packages/manager/cypress/e2e/core/account/account-cancellation.spec.ts
+++ b/packages/manager/cypress/e2e/core/account/account-cancellation.spec.ts
@@ -18,6 +18,11 @@ import {
 } from 'support/util/random';
 import type { CancelAccount } from '@linode/api-v4';
 import { mockWebpageUrl } from 'support/intercepts/general';
+import {
+  mockAppendFeatureFlags,
+  mockGetFeatureFlagClientstream,
+} from 'support/intercepts/feature-flags';
+import { makeFeatureFlagData } from 'support/util/feature-flags';
 
 // Data loss warning which is displayed in the account cancellation dialog.
 const cancellationDataLossWarning =
@@ -29,6 +34,18 @@ may not be able be restored.';
 const cancellationPaymentErrorMessage =
   'We were unable to charge your credit card for services rendered. \
 We cannot cancel this account until the balance has been paid.';
+
+// Tooltip message that appears when a child account tries to close the account.
+const contactParentUserTooltipsMessage =
+  'Contact your parent user to close your account.';
+
+// Tooltip message that appears when a child account tries to close the account.
+const contactCustomerSupportTooltipsMessage =
+  'Contact customer support to close this account.';
+
+// Tooltip message that appears when a parent account with one and more child accounts tries to close the account.
+const removeChildAccountTooltipsMessage =
+  'Remove child accounts before closing the account.';
 
 describe('Account cancellation', () => {
   /*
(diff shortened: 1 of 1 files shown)


packages/manager/cypress/e2e/core/account/account-cancellation.spec.ts
```diff
@@ -198,3 +215,223 @@ describe('Account cancellation', () => {
...
+        ui.tooltip
+          .findByText(contactCustomerSupportTooltipsMessage)
+          .should('be.visible');
+      });
+  });
+
+  /**
+   * Confirms that a parent account with one or more active child accounts cannot close the account
+   */
+  it('disables a parent account with one or more active child accounts to close the account', () => {
```
jdamore-linode, line 291:
```suggestion
  it('disables "Close Account" button for parent users', () => {
```

jdamore-linode (approved):
Looks good @cliu-akamai! I'd like to get either Mariah or Jaalah's approval before merging just to make sure all these scenarios are correct, but it's looking good to me! Thanks!

jaalah-akamai (approved):

packages/manager/cypress/e2e/core/account/account-cancellation.spec.ts
```diff
@@ -30,6 +35,18 @@ const cancellationPaymentErrorMessage =
...
+const contactParentUserTooltipsMessage =
+  'Contact your parent user to close your account.';
+
+// Tooltip message that appears when a child account tries to close the account.
+const contactCustomerSupportTooltipsMessage =
+  'Contact customer support to close this account.';
+
+// Tooltip message that appears when a parent account with one and more child accounts tries to close the account.
+const removeChildAccountTooltipsMessage =
+  'Remove child accounts before closing the account.';
```
jaalah-akamai, line 48: Nit: We started created [constants](https://github.com/linode/manager/blob/develop/packages/manager/src/features/Account/constants.ts#L22) for things like this which would be nice to use especially when verbiage changes.

jaalah-akamai: Same for the other two if @jdamore-linode approves, but I understand if in a testing scenario we want to keep things silo'd

jaalah-akamai (approved):

packages/manager/cypress/e2e/core/account/account-cancellation.spec.ts
jdamore-linode: I agree @jaalah-akamai, I think using the constants from `src` here would make sense.

(The only situations where I might insist that the strings be defined in the tests rather than imported from `src` is when the language itself is really important, like with notices and warnings involving billing or potential data loss, and that's specifically _so_ the tests break if the language changes: it basically forces us to double check the wording and gives us an extra opportunity to spot typos, etc.)

mjac0bs (approved):
This is looking good overall.  Approved pending the suggestion comments are cleaned up, we make the final default user closing accounts test case accurate, and a changeset is added.

packages/manager/cypress/e2e/core/account/account-cancellation.spec.ts
jaalah-akamai: Yup I agree with that

PR #10307, opened 2024-03-22
change: [M3-7898] - Add ticket severity fields to API package

hkhalil-akamai:
## Description 📝
Adds the proposed `severity` fields to support ticket request/response interfaces.

## Target release date 🗓️
4/15


mjac0bs (commented):
Looks like we've got a TS error -- we'll also need to update the [`supportTicketFactory`](https://github.com/linode/manager/blob/46ca9e128f21205918f7985eaef55f2655de2c4a/packages/manager/src/factories/support.ts#L4-L18) to include the new `severity` field. 

Screenshot 2024-03-22 at 9 52 17 AM

PR #10310, opened 2024-03-22
test: [M3-7852] Use test weights for Jenkins/Cypress parallelization

jdamore-linode:
## Description 📝
Proof of concept to use test weights when distributing tests across containers to improve the efficiency of our Jenkins runs. Always annoys me when I see one runner finished 10 minutes before all the rest, and this should help bring our Jenkins test performance closer to GHA/Cypress Cloud.

Works by reading a JSON file containing weights for each spec and using that info to determine which runner gets which specs. The JSON file is generated when the `CY_TEST_GENWEIGHTS` environment variable is defined. Developed specifically to be resilient to these scenarios:

- Weights file is absent or malformed
- Weights file contains paths to spec files that no longer exist
- Specs exist that aren't listed in the weights file

When weights aren't available, it falls back to the old round robin method of distributing the specs.

Will follow up Monday with documentation and rework of `split-run.ts` which is kind of a mess. 

## Changes  🔄
- ...
- ...

## Target release date 🗓️

## Preview 📷

| Before  | After   |
| ------- | ------- |
| 📷 | 📷 |

### Prerequisites
- ...
- ...

### Reproduction steps
- ...
- ...

### Verification steps
(How to verify changes)
- ...
- ...

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

---

**Commit Types:**

---


hkhalil-akamai:
This looks really interesting @jdamore-linode! Do you have any stats on the speedup with weights? Also, it may be useful to include some documentation on how to use the `generate-weights.ts` and `split-run.ts` utilities -- might be good to include them in the testing section of our docs site.

jdamore-linode:
> This looks really interesting @jdamore-linode! Do you have any stats on the speedup with weights? Also, it may be useful to include some documentation on how to use the `generate-weights.ts` and `split-run.ts` utilities -- might be good to include them in the testing section of our docs site.

Thanks for the feedback and questions @hkhalil-akamai!

I don't have a ton of concrete data but I did archive the most recent Jenkins run against this PR since it sort of represents the best case (as in the test weights are up-to-date and accurate, and the run itself didn't have any failures/flaky tests), and it took 20:38 to run the tests. Meanwhile another test run that day which also had no failures or flaky tests took 23:25 -- not a huge difference.

Because the current way we distribute tests is essentially random (sorts specs by filename then distributes them to runners round-robin style), we could theoretically run into situations where a single runner gets assigned all the slowest specs and dramatically increases the length of the run, and the weights help guard against that type of scenario (goes without saying that it'd be very unlikely for one runner to get _all_ the slowest test, but it doesn't require too much bad luck to impact the results).

The real and hypothetical gains made by this should definitely be weighed against the maintenance burden of having to keep the weights up-to-date; it only requires running the tests to generate the weights, and we could even automate that, but it does get complicated when you have to start factoring in test flake and failures, which throw off the weights. (Cypress solves this by using test weights collected/generated over a long span of time, so it accounts for each test's tendency to fail, but I'm not sure we have a straightforward way to do something similar. Definitely open to suggestions!)

> Also, it may be useful to include some documentation on how to use the `generate-weights.ts` and `split-run.ts` utilities -- might be good to include them in the testing section of our docs site.

Thanks for the reminder! Had been thinking about adding docs (particularly for the environment variables) the entire time I worked on this ticket and still managed to forget in the end 🤷‍♂️ appreciate the callout! Will add some docs soon

packages/manager/cypress/support/plugins/split-run.ts
```diff
@@ -24,14 +26,108 @@ export const splitCypressRun: CypressPlugin = (_on, config) => {
...
-    // Sort spec filenames deterministically.
-    // Or at least as deterministically as we can in a pinch...
-    specs.sort((a: string, b: string): number => {
+
+    // Attempt to locate a weights file for the test suite being run.
+    const weightsPath = `cypress/weights/${suiteName}-weights.json`;
+
+    // Information about test weights to be included in the info table, if applicable.
+    // TODO type.
+    let weightsInfo: any = {
```
jaalah-akamai, line 41: nit: Record<string, any>

jaalah-akamai (approved):
I agree with what has been said especially if this involves some maintenance overhead.

jdamore-linode:
Re: maintenance burden: planning to push a change soon that will allow weight file input and output paths to be specified so that we can script the test pipeline to push/pull test weights from git (not this repo) and ease the maintenance burden/automate it entirely

cc @jaalah-akamai @hkhalil-akamai

hkhalil-akamai:
> Re: maintenance burden: planning to push a change soon that will allow weight file input and output paths to be specified so that we can script the test pipeline to push/pull test weights from git (not this repo) and ease the maintenance burden/automate it entirely

Making it hands-free/automated would be the dream!! But I'm curious about how this would work -- would it be another git repo? Could it be possible to pull from previous runs' jenkins artifacts to automatically calculate the optimal weights (to minimize the number of separate places we have to touch/maintain).

jdamore-linode:
> Making it hands-free/automated would be the dream!! But I'm curious about how this would work -- would it be another git repo? Could it be possible to pull from previous runs' jenkins artifacts to automatically calculate the optimal weights (to minimize the number of separate places we have to touch/maintain).

Yeah, this is my thought!: we'd have a separate git repo containing 1) timestamped test weight files 2) a script that reads all the test weight files for some given time period (2 weeks? 30 days?) and outputs to stdout new test weight data that aggregates and averages all of the input data.

So from the pipeline side, we'd pull that repo, run the script to aggregate the most recent data, write _that_ test weight data to the filesystem somewhere where our Cypress tests can read it, run the tests with the weights applied while simultaneously generating new test weight data via `CY_TEST_GENWEIGHTS`, then finally once Cypress finishes running we'd just commit and push the new test weight file so that those results get included in the aggregated test weights next time the pipeline runs.

hkhalil-akamai:
> That would be pretty cool! Sadly we only store artifacts for the most recent few dozen builds (for the sake of saving disk space), so we wouldn't have access to very long lived data to generate the average weights

I'm fine with going any way you deem appropriate, but I had an idea and I wonder if you'd be willing to indulge me: to make the artifact concept work, we wouldn't necessary need the entire history of runs, we could keep a running average of the weights and just update them with a weighted average. (e.g., new_weights = (.25 * weights_from_this_run) + (.75 * weights_running_average). That way, weights would slowly adapt to changing tests without a single errant test run significantly affecting the weights.

jdamore-linode:
@hkhalil-akamai @jaalah-akamai Requesting re-review when you both have the chance (no urgency whatsoever, this is a really low priority).

Quick overview of the changes made:

- Allowed test weight file read path to be specified using `CY_TEST_SPLIT_RUN_WEIGHTS`
- Documented `CY_TEST_GENWEIGHTS` and `CY_TEST_SPLIT_RUN_WEIGHTS`
- Deleted the `core-weights.json` file from the repo since it can/should be maintained elsewhere once we integrate this with CI

Thanks guys!

docs/development-guide/08-testing.md
```diff
@@ -199,24 +199,26 @@ These environment variables are used by Cloud Manager's UI tests to override reg
...
 
-| Environment Variable      | Description                                | Example        | Default                    |
-|---------------------------|--------------------------------------------|----------------|----------------------------|
-| `CY_TEST_SPLIT_RUN`       | Enable run splitting                       | `1`            | Unset; disabled by default |
-| `CY_TEST_SPLIT_RUN_INDEX` | Numeric index for each Cypress test runner | `1`, `2`, etc. | Unset                      |
-| `CY_TEST_SPLIT_RUN_TOTAL` | Total number of runners for the tests      | `2`            | Unset                      |
+| Environment Variable        | Description                                | Example          | Default                    |
+|-----------------------------|--------------------------------------------|------------------|----------------------------|
+| `CY_TEST_SPLIT_RUN`         | Enable run splitting                       | `1`              | Unset; disabled by default |
```
jaalah-akamai, line 206: Should these be added to our `.env.example`? What are the defaults you've found to be a good balance for these?

packages/manager/cypress/support/plugins/split-run.ts
```diff
@@ -1,15 +1,18 @@
...
+import { resolve } from 'path';
+import { readFileSync } from 'fs';
+import { SpecWeight, SpecWeights, specWeightsSchema } from './generate-weights';
 import type { CypressPlugin } from './plugin';
 
 export const splitCypressRun: CypressPlugin = (_on, config) => {
   const splitRunEnabled = config?.env?.['CY_TEST_SPLIT_RUN'];
   const splitRunTotalRunners = config?.env?.['CY_TEST_SPLIT_RUN_TOTAL'];
   const splitRunRunnerIndex = config?.env?.['CY_TEST_SPLIT_RUN_INDEX'];
+  const splitRunWeightsPath = config?.env?.['CY_TEST_SPLIT_RUN_WEIGHTS'];
```
jaalah-akamai, line 15: Could we do something like this?
```suggestion
  const {
    CY_TEST_SPLIT_RUN: splitRunEnabled,
    CY_TEST_SPLIT_RUN_TOTAL: splitRunTotalRunners,
    CY_TEST_SPLIT_RUN_INDEX: splitRunRunnerIndex,
    CY_TEST_SPLIT_RUN_WEIGHTS: splitRunWeightsPath,
  } = config.env;
```

packages/manager/cypress/support/plugins/split-run.ts
```diff
@@ -27,11 +30,111 @@ export const splitCypressRun: CypressPlugin = (_on, config) => {
...
+          .filter((specWeight): specWeight is SpecWeight => !!specWeight);
+
+        // Get an array of relative paths to specs that do not have weight data available.
+        unweighedSpecs = specs.filter((specPath: string) => {
+          return !specWeights.find(
+            (specWeight) => specWeight.filepath === specPath
+          );
+        });
+
+        const totalWeight = weightsData.meta.totalWeight;
```
jaalah-akamai, line 77: Nit: Only using this once, may not need to store in variable

jaalah-akamai (approved):

docs/development-guide/08-testing.md
jdamore-linode: Hm, I don't think these belong there since they're more intended for CI and not super practical to use locally (you _could_, but the cost of having multiple Cypress and Chrome instances running at once plus the hassle of setting up separate test accounts for each runner probably makes it more trouble than it's worth).

However, I've never really considered documenting any of these in `.env.example`, so I'll take a quick look later and add any that might make sense! Thanks Jaalah!

packages/manager/cypress/support/plugins/split-run.ts
```diff
@@ -27,11 +30,111 @@ export const splitCypressRun: CypressPlugin = (_on, config) => {
...
 
-    // Override configuration spec pattern to reflect test subset for this runner.
+    // Override configuration spec pattern to reflect test subset for this runner...
     const specs = globSync(config.specPattern);
-    // Sort spec filenames deterministically.
-    // Or at least as deterministically as we can in a pinch...
-    specs.sort((a: string, b: string): number => {
+
+    // Attempt to locate a weights file for the test suite being run.
+    // const weightsPath = `cypress/weights/${suiteName}-weights.json`;
```
hkhalil-akamai, line 37:
```suggestion
```

packages/manager/cypress/support/plugins/split-run.ts
```diff
@@ -27,11 +30,111 @@ export const splitCypressRun: CypressPlugin = (_on, config) => {
...
+
+    // Information about test weights to be included in the info table, if applicable.
+    // TODO type.
+    let weightsInfo: Record<string, any> = {
```
hkhalil-akamai, line 41: The type here isn't super critical since this is just for informative purposes
```suggestion
    let weightsInfo: Record<string, string | number> = {
```

hkhalil-akamai (approved):
Looking good aside from some minor code style suggestions to consider. Where did we land on how weights will be generated/consumed in our CI pipeline? We might want to document it internally (maybe KB? I'm not sure if/where we have existing documentation describing our CI).

packages/manager/cypress/support/plugins/split-run.ts
jdamore-linode: Nice catch, thanks!

jdamore-linode:
@hkhalil-akamai @jaalah-akamai Applied your suggestions and refactored `split-run.ts` a bit -- mind taking another look next time you have a chance? (Very, very low priority)

Thanks again for all this feedback, it's been extremely helpful!

packages/manager/cypress/support/plugins/split-run.ts
hkhalil-akamai, line 1: One more minor change... what are your thoughts on moving the helper function definitions to *after* `splitCypressRun`? This way, someone reading the file would see the main function first and can jump to the function definitions below if they wanted to see how they were implemented.

hkhalil-akamai (approved):
Thanks for addressing feedback! Would definitely say `split-run.ts` is much more readable now! I left one minor suggestion about re-ordering the function definitions but feel free to heed or ignore that. Really impressive work!

packages/manager/cypress/support/plugins/split-run.ts
jdamore-linode: Done!

Thanks again @hkhalil-akamai for all your feedback and patience! Unless there are any objections in the meantime, I'll plan to merge this in the next day or two 👍

PR #10313, opened 2024-03-25
feat: [M3-7888] - Revoke proxy PAT when switching accounts

jaalah-akamai:
## Description 📝
As a user who frequently switches between accounts, I notice that proxy tokens (PATs) accumulate when switching into the same account repeatedly. To mitigate this, we want to ensure that we automatically revoke proxy tokens whenever a user switches accounts.

## Changes  🔄
- New `usePendingRevocationToken` hook with accompanying unit test.
    - Needed to consolidate code required for both `UserMenu` and `AccountLanding` 
- Added `getPersonalAccessTokenForRevocation` to compare the token we have in local storage to what we have stored in our database from the API endpoint.

## Target release date 🗓️
4/1

## Preview 📷
| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/linode/manager/assets/125309814/d0453e41-86cf-40ab-a73b-dc983ef249e3" /> | <video src="https://github.com/linode/manager/assets/125309814/17d5418b-4cf3-4382-8047-f7a160ff9511" /> |

### Prerequisites
- Log into Parent account using developer credentials

### Reproduction steps
- Switch into proxy account and go to PAT page
- Keep switching into proxy and observe multiple PATs accumulating

### Verification steps
- Follow same reproduction steps and observe that this issue is fixed.

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/Account/SwitchAccountDrawer.tsx
+++ b/packages/manager/src/features/Account/SwitchAccountDrawer.tsx
@@ -1,4 +1,5 @@
 import { createChildAccountPersonalAccessToken } from '@linode/api-v4';
+import { useSnackbar } from 'notistack';
 import React from 'react';
 
 import { StyledLinkButton } from 'src/components/Button/StyledLinkButton';
@@ -12,6 +13,7 @@ import {
   updateCurrentTokenBasedOnUserType,
 } from 'src/features/Account/utils';
 import { useCurrentToken } from 'src/hooks/useAuthentication';
+import { useRevokePersonalAccessTokenMutation } from 'src/queries/tokens';
 import { sendSwitchToParentAccountEvent } from 'src/utilities/analytics';
 import { getStorage, setStorage } from 'src/utilities/storage';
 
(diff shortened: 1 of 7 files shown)


dwiley-akamai (approved):
Test passes ✅
Code review ✅
Confirmed proxy PATs do not continue accumulating when switching accounts ✅

packages/manager/src/features/Account/SwitchAccountDrawer.tsx
```diff
@@ -37,14 +40,57 @@ export const SwitchAccountDrawer = (props: Props) => {
...
     getStorage('authentication/parent_token/token') ?? '';
 
   const handleClose = React.useCallback(() => {
     onClose();
   }, [onClose]);
 
+  const handleProxyTokenRevocation = React.useCallback(async () => {
+    try {
+      await revokeToken();
+      enqueueSnackbar(`Successfully revoked ${proxyTokenId}.`, {
```
mjac0bs, line 59: Is it possible to use the proxy token name instead of id, as we do when revoking a token manually? Alternatively, to preface the id number with "Successfully revoked proxy token {id}".

Proxy revoke on account switch:
Screenshot 2024-03-26 at 1 19 23 PM

Manual revoke via action menu:
Screenshot 2024-03-26 at 1 18 49 PM

mjac0bs (approved):
Observed the expected behavior and confirmed the tests for the hook pass. I think we can slightly improve the success toast copy, so left a comment about that.

packages/manager/src/features/Account/SwitchAccountDrawer.tsx
jaalah-akamai: Yea, that would require returning the whole token from the hook instead of the id. I'll make those changes

PR #10316, opened 2024-03-26
fix: [M3-7508] - Follow up on tests to check Parent and Child 'Close Account' flows

cliu-akamai:
## Description 📝
Fix the PR comments in previous PR [Add tests to check Parent and Child Close Account flows](https://github.com/linode/manager/pull/10296).

### DIFF
--- a/packages/manager/.changeset/pr-10316-tests-1711480175744.md
+++ b/packages/manager/.changeset/pr-10316-tests-1711480175744.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Tests
+---
+
+Add tests to check Parent and Child Close Account flows ([#10316](https://github.com/linode/manager/pull/10316))
--- a/packages/manager/.changeset/pr-10316-tests-1711480175744.md
+++ b/packages/manager/.changeset/pr-10316-tests-1711480175744.md
@@ -2,4 +2,4 @@
 "@linode/manager": Tests
 ---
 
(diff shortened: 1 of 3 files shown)


packages/manager/.changeset/pr-10316-tests-1711480175744.md
```diff
@@ -0,0 +1,5 @@
+---
+"@linode/manager": Tests
+---
+
+Add tests to check Parent and Child Close Account flows ([#10316](https://github.com/linode/manager/pull/10316))
```
mjac0bs, line 5:
```suggestion
Add tests to check Parent and Child Close Account flows ([#10316](https://github.com/linode/manager/pull/10316), [#10296](https://github.com/linode/manager/pull/10296))
```
Let's reference the original PR, too.

mjac0bs (approved):
Thanks for addressing the feedback. Test cases and comment look accurate now. Based on comments in the last PR, the suggestion may have been to use [CM's constants from src to avoid duplicating the tooltip text](https://github.com/linode/manager/blob/develop/packages/manager/src/features/Account/constants.ts#L23), but if there's a reason to avoid that or it can't be done, this seems fine too.

dwiley-akamai (approved):
`account-cancellation.spec.ts` passes ✅

I agree re: the use of `src/features/Account/constants.ts` if possible for the constants (and also renaming the ones in this PR to match the usual casing convention for constants, e.g. `PARENT_USER_CLOSE_ACCOUNT_TOOLTIP_TEXT`)

PR #10317, opened 2024-03-26
upcoming: [M3-7898] - Support ticket severity

hkhalil-akamai:
## Description 📝
Adds Cloud Manager support for the upcoming support ticket severity changes.

## Changes  🔄
- APIv4:
  - Updated `SupportTicket` and `TicketRequest` types with new `severity` field
  - New `Support Ticket Severity` account capability
- Cloud Manager:
  - New "Severity" column in support ticket list
  - Display ticket severity in ticket details status bar
  - Add severity select input when creating a new support ticket

## Target release date 🗓️
4/15

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Severity column |
| Screenshot 2024-03-26 at 3 26 44 PM | Screenshot 2024-03-26 at 3 26 55 PM |
| Ticket details |
|  Screenshot 2024-03-26 at 3 28 25 PM |  Screenshot 2024-03-26 at 3 28 30 PM |
| New Ticket Dialog |
| Screenshot 2024-03-26 at 3 28 58 PM |  Screenshot 2024-03-26 at 3 29 15 PM |

### Prerequisites
- Enable 'Support Ticket Severity' feature flag
- Enable MSW

### Verification steps
- Navigate to Support > Tickets and verify that the 'Severity' column appears, displays each ticket's severity and can be sorted
- Click on a ticket and verify the ticket's severity appears in the status bar at the top of the page
- Click "Open New Ticket" and verify a severity can be selected
- Hover over tooltip and verify copy (from ticket)

> [!NOTE]
> As of 3/26, ticket severity is not supported by the API. Submitting a ticket will ignore the specified severity.

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/features/Support/SupportTickets/SupportTicketDialog.tsx
```diff
@@ -596,24 +640,27 @@ export const SupportTicketDialog = (props: SupportTicketDialogProps) => {
...
                     label="What is this regarding?"
-                    onChange={handleEntityTypeChange}
+                    onChange={(e, type) => handleEntityTypeChange(type.value)}
                     options={topicOptions}
                     value={selectedTopic}
                   />
                   {!['general', 'none'].includes(entityType) && (
                     <>
-                      <Select
+                      <Autocomplete
```
hkhalil-akamai, line 653: Replaced the existing react `Select`s in this file with MUI `Autocomplete`.

packages/manager/src/features/Support/SupportTickets/ticketUtils.ts
```diff
@@ -44,3 +48,20 @@ export const getTicketsPage = (
...
+    Boolean(flags.supportTicketSeverity),
+    account?.capabilities ?? []
+  );
+};
+
+export const severityLabelMap: Map<TicketSeverity, string> = new Map([
+  [1, '1-Major Impact'],
+  [2, '2-Moderate Impact'],
+  [3, '3-Low Impact'],
+]);
```
hkhalil-akamai, line 67: Choosing to use a map here since integer-keyed objects are not well supported. For example, calling `Object.keys` converts the integer keys to strings.

packages/manager/src/features/Support/SupportTickets/SupportTicketDialog.tsx
```diff
@@ -587,6 +613,24 @@ export const SupportTicketDialog = (props: SupportTicketDialogProps) => {
...
+          {hasSeverityCapability && (
+            <Autocomplete
+              onChange={(e, severity) =>
+                setSelectedSeverity(
+                  severity != null ? severity.value : undefined
+                )
+              }
+              textFieldProps={{
+                tooltipPosition: 'right',
+                tooltipText: TICKET_SEVERITY_TOOLTIP_TEXT,
```
mjac0bs, line 625: Can we increase the width of this tooltip to make it easier to read? 
Screenshot 2024-03-26 at 1 43 11 PM

packages/manager/src/features/Support/SupportTicketDetail/TicketStatus.tsx
```diff
@@ -42,37 +44,42 @@ export const TicketStatus = (props: Props) => {
...
+        <Typography>
+          | {statusUpdateText} by {updated_by} at {formattedDate}
+        </Typography>
+        &nbsp;
+        {renderEntityLabel()}
+      </Grid>
+      {severity && (
+        <Grid alignItems="center" container direction="row" spacing={1}>
+          <Grid>
+            <Typography>Severity</Typography>
```
mjac0bs, line 64:
```suggestion
            <Typography>Severity:</Typography>
```

Should we consider putting a colon after "Severity" here, just to make it clear that the helper text and the chip are related? 

Currently:
Screenshot 2024-03-26 at 1 44 12 PM

Potentially:
Screenshot 2024-03-26 at 2 00 09 PM

packages/manager/src/features/Support/SupportTicketDetail/SeverityChip.tsx
mjac0bs, line 1: Did we consider making the text color of the Moderate Severity chip black in dark mode? The contrast here is not great.

Screenshot 2024-03-26 at 2 21 18 PM

packages/api-v4/.changeset/pr-10317-added-1711481937702.md
```diff
@@ -0,0 +1,5 @@
+---
+"@linode/api-v4": Added
+---
+
+Added severity fields to support ticket endpoints and new account capability ([#10317](https://github.com/linode/manager/pull/10317))
```
mjac0bs, line 5:
```suggestion
Severity fields to support ticket endpoints and new account capability ([#10317](https://github.com/linode/manager/pull/10317))
```
Nit: don't need to duplicate "added" since everything in this section is added

packages/manager/src/features/Support/SupportTickets/SupportTicketDialog.tsx
mjac0bs: Likely the cause of a test failure:

Screenshot 2024-03-26 at 1 40 52 PM

packages/manager/src/features/Support/SupportTickets/ticketUtils.ts
```diff
@@ -44,3 +48,20 @@ export const getTicketsPage = (
...
+    'Support Ticket Severity',
+    Boolean(flags.supportTicketSeverity),
+    account?.capabilities ?? []
+  );
+};
+
+export const severityLabelMap: Map<TicketSeverity, string> = new Map([
+  [1, '1-Major Impact'],
+  [2, '2-Moderate Impact'],
+  [3, '3-Low Impact'],
```
mjac0bs, line 66: This may just be a personal annoyance, but I wish we had spacing between the number and the impact (e.g. "1 - Major Impact"). Does anyone else feel the same?

mjac0bs (commented):
Great work  -- here's some UI feedback I have and a note that `open-support-ticket.spec.ts` is a legit failure that seems related to the Select component replacement work.

packages/manager/src/features/Support/SupportTickets/ticketUtils.ts
bnussman-akamai: Yes, I agree, but I feel strongly that there shouldn't even be a number in the first place.

jaalah-akamai (approved):

packages/manager/src/features/Support/SupportTickets/ticketUtils.ts
mjac0bs: Me too, but that seems to be something Support specifically requested. 🙃

jaalah-akamai: Let me circle back with an update

jaalah-akamai: Unfortunately, the numbers will have to stay for the time being so that we have parity with the implementation from other teams, but we can add spacing.

packages/manager/src/features/Support/SupportTickets/SupportTicketDialog.tsx
hkhalil-akamai: I agree but I can't figure out how to do this without changing the setting in [the global theme](https://github.com/linode/manager/blob/0fdc974e71c030437219b9491322077f1a0bd5ea/packages/manager/src/foundations/themes/light.ts#L1342). I tried using the `tooltipClasses` prop in conjunction with `makeStyles` as well as the `sx` prop but neither seem to have any impact.

Any suggestions or help in accomplishing this would be appreciated!! 🙏🏽

jdamore-linode:
@hkhalil-akamai This is great! Planning to follow up with some test additions soon, but I'm noticing some inconsistencies with the severity autocomplete field compared to some of the other autocompletes in the app:

- The severity autocomplete allows the user to enter an invalid value. I can type "asdf", press enter, and proceed with the ticket creation, but the UI never indicates that I've entered an invalid severity. When the ticket gets submitted, the "severity" field is omitted from the outgoing API request and the user isn't made aware. (Other autocompletes, like the Region select, seem to clear the user's input when they enter something invalid and move focus to another element)
- Similarly, if I type "1", "2", or "3" into the field and hit "enter", I would expect it to automatically select the relevant severity, but instead it just leaves the typed value as-is. Contrast to the Region select autocomplete, where I can type "us-e", hit enter, and "Newark, NJ (us-east)" gets selected. The crux of the issue seems to be that the top autocomplete result isn't getting focused automatically.

packages/manager/cypress/e2e/core/helpAndSupport/open-support-ticket.spec.ts
hkhalil-akamai, line 1: Thanks for updating these tests!

hkhalil-akamai:
> @hkhalil-akamai This is great! Planning to follow up with some test additions soon, but I'm noticing some inconsistencies with the severity autocomplete field compared to some of the other autocompletes in the app:

Great observations and I agree that these negatively affect usability. Both are easily solved by adding the `autoHighlight` and `clearOnBlur` props, but I wonder if we should be enabling these by default across all Autocompletes?

packages/manager/src/features/Support/SupportTickets/ticketUtils.ts
hkhalil-akamai: @jaalah-akamai can you double check the verdict on spacing? From Andrew: "I originally had spaces between characters for the severity and removed them for absolute consistency with Akamai convention"

packages/manager/src/features/Support/SupportTicketDetail/TicketStatus.tsx
```diff
@@ -42,37 +44,42 @@ export const TicketStatus = (props: Props) => {
...
-      })}
-    >
-      <Stack
-        sx={{
-          display: 'inline-flex',
-          flexDirection: 'row',
-        }}
-      >
+    <StyledStatusBar container data-qa-ticket-status>
+      <Grid alignItems="center" container direction="row" xs>
```
bnussman-akamai, line 48: Does `Grid` make more sense than `Stack` here? I tend to Stacks if the layout is simple enough

packages/manager/src/features/Support/SupportTicketDetail/TicketStatus.tsx
```diff
@@ -42,37 +44,42 @@ export const TicketStatus = (props: Props) => {
...
-        },
-      })}
-    >
-      <Stack
-        sx={{
-          display: 'inline-flex',
-          flexDirection: 'row',
-        }}
-      >
+    <StyledStatusBar container data-qa-ticket-status>
```
bnussman-akamai, line 47: Can we try to avoid one off styled components and just use a `Paper`?

packages/manager/src/features/Support/SupportTicketDetail/TicketStatus.tsx
bnussman-akamai, line 1: Can we improve wrapping?

Screenshot 2024-03-29 at 12 45 16 PM

packages/manager/src/features/Support/SupportTicketDetail/SeverityChip.tsx
```diff
@@ -0,0 +1,25 @@
...
+  <StyledChip
+    color={severityColorMap[severity]}
+    label={severityLabelMap.get(severity)}
+  />
+);
+
+const StyledChip = styled(Chip, { label: 'StyledChip' })(({ theme }) => ({
+  paddingLeft: theme.spacing(),
+  paddingRight: theme.spacing(),
+}));
```
bnussman-akamai, line 25: Does the Chip really need margin? Could we handle spacing by putting it in a flex box that has `spacing` or `gap`?

bnussman-akamai (commented):
Things look solid, just have a few concerns with styling / components

packages/manager/src/features/Support/SupportTicketDetail/TicketStatus.tsx
hkhalil-akamai: Ah, thanks for pointing this out. Improved this based on the mobile UX mockup.

packages/manager/src/features/Support/SupportTicketDetail/SeverityChip.tsx
hkhalil-akamai: I added the padding to stay true to the UX mockups. Simplified this somewhat, let me know your thoughts.

bnussman-akamai (approved):

packages/manager/src/features/Support/SupportTickets/ticketUtils.ts
jaalah-akamai: Let's keep what we have for now as to not hold this up. We can revisit later

jdamore-linode (approved):
Awesome work, thanks again @hkhalil-akamai!

packages/manager/src/features/Support/SupportTickets/SupportTicketDialog.tsx
mjac0bs: > I agree but I can't figure out how to do this 

I honestly don't know either. We use `tooltipClasses` in a few places and pass in `classes.tooltip`, where we define a custom width, but that doesn't work in this case. It does look like tooltipClasses are getting passed in [here](https://github.com/linode/manager/blob/f08356f003411798f52d8a4b5edd272fb2e18712/packages/manager/src/components/TextField.tsx#L486), but applied to the popper, so I'm not sure if they're being handled correctly.

mjac0bs (approved):
Not sure what can be done about that tooltip text (and to Joe's point, I wonder what that would look like in the form) in the future, but that's a minor improvement we could revisit.  Everything else looks good. 🚢

PR #10330, opened 2024-03-29
fix: [M3-7997] - Charts Y-axis values are trimmed when scale is increased

hana-linode:
## Description 📝
Fix chart y-axis values being trimmed for large values and add support for Billion and Trillion values

## Preview 📷
| Before  | After   |
| ------- | ------- |
| image | image |
| Screenshot 2024-04-16 at 2 41 39 PM | image |

### Verification steps
- Go to a Longview client details page and check the charts (you may need to refresh/play with the dropdown if there's an error loading the data). The y-axis label values should no longer be cut off

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/utilities/unitConversions.ts
```diff
@@ -90,14 +90,18 @@ export const readableBytes = (
     num = -num;
   }
 
-  const power = determinePower(num, storageUnits, options);
+  // If no maxUnit is provided, default to the highest unit
+  const power = determinePower(num, storageUnits, {
+    ...options,
+    maxUnit: options.maxUnit ?? storageUnits[storageUnits.length - 1],
+  });
```
hkhalil-akamai, line 97: This fixes the unit overflow bug in case we ever encounter PB or larger values in the future (unlikely)

packages/manager/src/components/AreaChart/utils.ts
```diff
@@ -17,6 +17,22 @@ export const tooltipLabelFormatter = (timestamp: number, timezone: string) =>
...
+  }
+  if (value >= 1000000) {
+    return +(value / 1000000).toFixed(1) + 'M';
+  }
+  if (value >= 1000) {
+    return +(value / 1000).toFixed(1) + 'K';
+  }
+  return `${value}`;
+};
+
```
hana-linode, line 35: Consolidate `humanizeLargeData` function and add support for billion and trillion

packages/manager/src/components/LineGraph/LineGraph.tsx
```diff
@@ -186,6 +177,11 @@ export const LineGraph = (props: LineGraphProps) => {
   ) => {
     const finalChartOptions: ChartOptions = {
       animation: { duration: 0 },
+      layout: {
+        padding: {
+          left: 8,
+        },
+      },
```
hana-linode, line 184: Fix for y-axis label being cut off

hana-linode:
I will remove the mocked data and changes to display the mocked data before merging this PR

packages/manager/src/components/AreaChart/utils.ts
hkhalil-akamai: I wonder if we still need this function or whether we can use `readableBytes` instead.

mjac0bs (approved):
✅  No longer seeing any access values being trimmed for billions or trillions.

It does look like we have some linter complaints in NetworkGraph.ts that caused the TS check to fail. Just some unused imports.

image

Approving pending removal of mocks before merge and feedback is addressed.

Question unrelated to bug but related to chart axises: Do we know what the y-axis unit is supposed to be for the `Load (Target < 1.00)`? Seconds?

packages/manager/src/components/AreaChart/utils.ts
hkhalil-akamai: Bumping this as I believe we can use `readableBytes(value).formatted` instead to avoid redundancy.

packages/manager/src/features/Longview/LongviewDetail/DetailTabs/OverviewGraphs/DiskGraph.tsx
```diff
@@ -39,10 +40,10 @@ export const DiskGraph = (props: GraphProps) => {
   const { error, read, swap, write } = React.useMemo(
     () =>
       processDiskData(
-        pathOr({}, ['Disk'], data),
-        pathOr('kvm', ['SysInfo', 'type'], data)
+        db1disk[0].DATA.Disk,
+        db1disk[0].DATA.SysInfo.type
       ),
```
hkhalil-akamai, line 45: When we remove the mock data, we can remove the ramda util too:

```suggestion
    () => processDiskData(data.Disk ?? {}, data.SysInfo?.type ?? 'kvm'),
```

hkhalil-akamai: Same with the other usages of `pathOr` in this file so we can get rid of the ramda import all together.

packages/manager/src/features/Longview/LongviewDetail/DetailTabs/OverviewGraphs/MemoryGraph.tsx
```diff
@@ -40,10 +41,10 @@ export const MemoryGraph = (props: GraphProps) => {
 
   const _convertData = React.useCallback(convertData, [data, start, end]);
 
-  const buffers = pathOr<Stat[]>([], ['Memory', 'real', 'buffers'], data);
-  const cache = pathOr<Stat[]>([], ['Memory', 'real', 'cache'], data);
+  const buffers = db1memory[0].DATA.Memory.real.buffers;
+  const cache = db1memory[0].DATA.Memory.real.cache;
```
hkhalil-akamai, line 45: Same comment as above

packages/manager/src/features/Longview/LongviewDetail/DetailTabs/OverviewGraphs/NetworkGraph.tsx
```diff
@@ -33,10 +34,7 @@ export const NetworkGraph = React.memo((props: GraphProps) => {
     end
   );
 
-  const networkData = React.useMemo(
-    () => sumNetwork(pathOr({}, ['Interface'], data.Network)),
-    [data.Network]
-  );
+  const networkData = db1network[0].DATA.Network.Interface.eth0;
```
hkhalil-akamai, line 37: And here

hkhalil-akamai:
> Do we know what the y-axis unit is supposed to be for the `Load (Target < 1.00)`? Seconds?

I think it's a unitless value -- [see here](https://en.wikipedia.org/wiki/Load_(computing)#Interpretation).

hkhalil-akamai (approved):
Nice work on fixing up some crusty old components. Approved but I left some feedback that I think are minor improvements.

packages/manager/src/components/AreaChart/utils.ts
hana-linode: @hkhalil-akamai Looks like this messes up the y axis'

image

hkhalil-akamai: Ohh it might work with the following options set:

```
readableBytes(value, {
  base10: true,
  round: 0,
  unitLabels: {
    GB: 'B',
    KB: 'K',
    MB: 'M',
    TB: 'T',
    bytes: '',
  },
}).formatted;
```

I'll leave it up to you which route you want to take.

hana-linode: Going to keep it as is, the y-axis labels are still a bit weird with `readableBytes` and I don't want to change the function too much since it's used in a lot of places

PR #10331, opened 2024-03-29
refactor: [M3-7944] - Linode Create Refactor v2 - User Data - Part 7

bnussman-akamai:
## Description 📝

Adds User Data section to the new Linode Create page 👤📑 🎉 

> [!note]
> The User Data section will only show up if  you select an Image that is cloud-init capable and a region that supports Metadata. I don't necessarily agree with this UX, but this matches the existing behavior.

## Preview 📷

Screenshot 2024-03-29 at 11 38 08 AM

## How to test 🧪

- Turn on the `Linode Create v2` feature flag using load dev tools
- Verify you see the Metadata section if you select a `cloud-init` capable image and a Region that supports metadata
- Verify you can specify metadata 
 - Observe that the base64 encoded user data is in the payload

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserData.test.tsx
+++ b/packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserData.test.tsx
@@ -1,0 +1,57 @@
+import { fireEvent } from '@testing-library/react';
+import * as React from 'react';
+
+import { renderWithThemeAndHookFormContext } from 'src/utilities/testHelpers';
+
+import { UserData } from './UserData';
+
+describe('Linode Create v2 UserData', () => {
+  it('should display a warning message if the user data is not in an accepted format', () => {
+    const inputValue = '#test-string';
+    const { getByLabelText, getByText } = renderWithThemeAndHookFormContext({
+      component: <UserData />,
+    });
+
+    const input = getByLabelText('User Data');
+    fireEvent.change(input, { target: { value: inputValue } });
+    fireEvent.blur(input); // triggers format check
+
+    expect(
+      getByText('The user data may be formatted incorrectly.')
+    ).toBeInTheDocument();
+  });
+
+  it('should display a warning in the header for cloning', () => {
+    const { getByText } = renderWithThemeAndHookFormContext({
+      component: <UserData />,
+      options: {
+        MemoryRouter: {
+          initialEntries: [{ pathname: '/linodes/create?type=Clone+Linode' }],
+        },
+      },
+    });
+
+    expect(
+      getByText(
+        'Existing user data is not cloned. You may add new user data now.'
+      )
+    ).toBeVisible();
+  });
+
+  it('should display a warning in the header for creating from a Linode backup', () => {
+    const { getByText } = renderWithThemeAndHookFormContext({
+      component: <UserData />,
+      options: {
+        MemoryRouter: {
+          initialEntries: [{ pathname: '/linodes/create?type=Backups' }],
+        },
+      },
+    });
+
+    expect(
+      getByText(
+        'Existing user data is not accessible when creating a Linode from a backup. You may add new user data now.'
+      )
+    ).toBeVisible();
+  });
+});
--- a/packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserData.test.tsx
+++ b/packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserData.test.tsx
@@ -1,57 +1,63 @@
 import { fireEvent } from '@testing-library/react';
 import * as React from 'react';
 
+import { imageFactory, regionFactory } from 'src/factories';
+import { makeResourcePage } from 'src/mocks/serverHandlers';
+import { HttpResponse, http, server } from 'src/mocks/testServer';
 import { renderWithThemeAndHookFormContext } from 'src/utilities/testHelpers';
 
 import { UserData } from './UserData';
 
 describe('Linode Create v2 UserData', () => {
-  it('should display a warning message if the user data is not in an accepted format', () => {
-    const inputValue = '#test-string';
-    const { getByLabelText, getByText } = renderWithThemeAndHookFormContext({
+  it('should render if the selected image supports cloud-init and the region supports metadata', async () => {
+    const image = imageFactory.build({ capabilities: ['cloud-init'] });
+    const region = regionFactory.build({ capabilities: ['Metadata'] });
+
+    server.use(
+      http.get('*/v4/images/*', () => {
+        return HttpResponse.json(image);
+      }),
+      http.get('*/v4/regions', () => {
+        return HttpResponse.json(makeResourcePage([region]));
+      })
+    );
+
+    const { findByText } = renderWithThemeAndHookFormContext({
       component: <UserData />,
+      useFormOptions: { defaultValues: { image: image.id, region: region.id } },
     });
 
-    const input = getByLabelText('User Data');
+    const userDataHeading = await findByText('Add User Data');
+
+    expect(userDataHeading).toBeVisible();
+    expect(userDataHeading.tagName).toBe('H2');
+  });
+
+  it('should display a warning message if the user data is not in an accepted format', async () => {
+    const image = imageFactory.build({ capabilities: ['cloud-init'] });
+    const region = regionFactory.build({ capabilities: ['Metadata'] });
+
+    server.use(
+      http.get('*/v4/images/*', () => {
+        return HttpResponse.json(image);
+      }),
+      http.get('*/v4/regions', () => {
+        return HttpResponse.json(makeResourcePage([region]));
+      })
+    );
+
+    const inputValue = '#test-string';
+    const { findByLabelText, getByText } = renderWithThemeAndHookFormContext({
+      component: <UserData />,
+      useFormOptions: { defaultValues: { image: image.id, region: region.id } },
+    });
+
+    const input = await findByLabelText('User Data');
     fireEvent.change(input, { target: { value: inputValue } });
     fireEvent.blur(input); // triggers format check
 
     expect(
       getByText('The user data may be formatted incorrectly.')
     ).toBeInTheDocument();
-  });
-
-  it('should display a warning in the header for cloning', () => {
-    const { getByText } = renderWithThemeAndHookFormContext({
-      component: <UserData />,
-      options: {
-        MemoryRouter: {
-          initialEntries: [{ pathname: '/linodes/create?type=Clone+Linode' }],
-        },
-      },
-    });
-
-    expect(
-      getByText(
-        'Existing user data is not cloned. You may add new user data now.'
-      )
-    ).toBeVisible();
-  });
-
-  it('should display a warning in the header for creating from a Linode backup', () => {
-    const { getByText } = renderWithThemeAndHookFormContext({
-      component: <UserData />,
-      options: {
-        MemoryRouter: {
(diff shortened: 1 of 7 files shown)


packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserDataHeading.tsx
```diff
@@ -0,0 +1,55 @@
...
+    <Stack>
+      <Stack direction="row" spacing={1}>
+        <Typography variant="h2">Add User Data</Typography>
+        <TooltipIcon
+          text={
+            <>
+              User data allows you to provide additional custom data to
+              cloud-init to further configure your system.{' '}
+              <Link to="https://www.linode.com/docs/products/compute/compute-instances/guides/metadata/">
+                Learn more.
```
mjac0bs, line 39: Nit: generally, we don't include periods in the link.

packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserDataHeading.test.tsx
mjac0bs, line 1: Great test, thank you!

We could add test coverage to check that the help icon displays as expected.

packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserData.test.tsx
mjac0bs, line 1: We could add test coverage to check that the help icon displays as expected.

(One of them would actually be in UserDataHeading, I know.)
Screenshot 2024-03-29 at 10 32 20 AM

mjac0bs (approved):
Looks good from my review and testing! I didn't see test coverage for our help icons/tooltips in any e2es, so that could be something worth adding to these component unit tests. And we need a changeset, please.

✅ I see the Metadata section if I select a cloud-init capable image and a Region that supports metadata -- though I'm with you on the poor visibility of this feature. I had to google our [Metadata availability](https://www.linode.com/docs/products/compute/compute-instances/guides/metadata/?tabs=linode-api#availability) because I never remember.
✅  I can specify metadata and if it's not in the intended formats, I see the warning
✅ Observed that the base64 encoded user data is in the payload
✅  Observed the warnings for backups and cloning

packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserData.tsx
```diff
@@ -0,0 +1,114 @@
...
+                hasInputValueChanged: false,
+                userData: e.target.value,
+              })
+            }
+            onChange={(e) => {
+              checkFormat({
+                hasInputValueChanged: true,
+                userData: e.target.value,
+              });
+              field.onChange(utoa(e.target.value));
```
hkhalil-akamai, line 97: Instead of performing the `utoa` conversion on change, could we store the original value and only do the conversion on submit? This would also save us from having to convert back on line 106.

packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserData.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+import { imageFactory, regionFactory } from 'src/factories';
+import { makeResourcePage } from 'src/mocks/serverHandlers';
+import { HttpResponse, http, server } from 'src/mocks/testServer';
+import { renderWithThemeAndHookFormContext } from 'src/utilities/testHelpers';
+
+import { UserData } from './UserData';
+
+describe('Linode Create v2 UserData', () => {
+  it('should render if the selected image supports cloud-init and the region supports metadata', async () => {
```
hkhalil-akamai, line 12: Should we have a test for the opposite (i.e., doesn't appear when the image or regions doesn't support metadata)?

packages/manager/src/features/Linodes/LinodesCreate/utilities.tsx
```diff
@@ -27,6 +27,13 @@ export const utoa = (data: string) => {
   }
 };
 
+/**
+ * ASCII to Unicode (decode Base64 to original data)
+ */
+export function atou(b64: string) {
+  return decodeURIComponent(escape(atob(b64)));
+}
+
```
hkhalil-akamai, line 36: It might be worth adding a test for this utility (and its counterpart) with some common examples and edge cases (like hex escape codes).

packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserData.tsx
```diff
@@ -0,0 +1,114 @@
...
+    const userDataLower = userData.toLowerCase();
+    const validPrefixes = ['#cloud-config', 'content-type: text/', '#!/bin/'];
+    const isUserDataValid = validPrefixes.some((prefix) =>
+      userDataLower.startsWith(prefix)
+    );
+    if (userData.length > 0 && !isUserDataValid && !hasInputValueChanged) {
+      setFormatWarning(true);
+    } else {
+      setFormatWarning(false);
+    }
```
hkhalil-akamai, line 45: I think I was the one who wrote this code originally but looking at it again:

```suggestion
    setFormatWarning(userData.length > 0 && !isUserDataValid && !hasInputValueChanged);
```

hkhalil-akamai (approved):
Noticed no regressions after manual testing. Left a few suggestions. Nice work, really appreciate the bite-sized PRs! 🙏🏽

packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserData.test.tsx
bnussman-akamai: Yes, but I don't really know how to reliably do this because the unit test waits for API data to load

jdamore-linode: @bnussman-akamai This is a good candidate to mock the `useImages` and `useRegions` hooks so you can ignore the HTTP stuff and test directly against the state you want, but you can also do this kind of hacky thing to wait for the requests before testing the state of the component:

```typescript
    const image = imageFactory.build({ capabilities: [] });
    const region = regionFactory.build({ capabilities: ['Metadata'] });
    let [hasLoadedImages, hasLoadedRegions] = [false, false];

    server.use(
      http.get('*/v4/images/*', () => {
        hasLoadedImages = true;
        return HttpResponse.json(image);
      }),
      http.get('*/v4/regions', () => {
        hasLoadedRegions = true;
        return HttpResponse.json(makeResourcePage([region]));
      })
    );

    const { queryByText } = renderWithThemeAndHookFormContext({
      component: <UserData />,
      useFormOptions: { defaultValues: { image: image.id, region: region.id } },
    });

    // Wait for requests...
    await waitFor(() => {
      expect(hasLoadedImages && hasLoadedRegions).toBe(true);
    });

    // Assert that user data is hidden here.
```

bnussman-akamai: yeah good call. Mocking `useImages` and `useRegions` would probably be best. Thank you for the tip 🙏

packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserData.tsx
bnussman-akamai: I'm considering both options. At first, when building this new create flow, I wanted to try to keep the form data in the format the API expects, but not I'm realizing it might make more sense to perform some transformation on submit. 

I might merge as-is, but might come back and change this.

abailly-akamai: FWIW, i agree with a submit data transformation. This looks a bit expensive especially without debouncing - you also transform the value

packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserDataHeading.tsx
```diff
@@ -0,0 +1,56 @@
...
+          status="help"
+          sxTooltipIcon={{ p: 0 }}
+        />
+      </Stack>
+      {warningMessage && (
+        <Notice spacingBottom={16} spacingTop={16} variant="warning">
+          {warningMessage}
+        </Notice>
+      )}
+    </Stack>
```
abailly-akamai, line 54: nit, let's not add another div if not needed and use a fragment instead

packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserData.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { Accordion } from 'src/components/Accordion';
+import { Link } from 'src/components/Link';
+import { Notice } from 'src/components/Notice/Notice';
+import { TextField } from 'src/components/TextField';
+import { Typography } from 'src/components/Typography';
+import { useRestrictedGlobalGrantCheck } from 'src/hooks/useRestrictedGlobalGrantCheck';
+import { useImageQuery } from 'src/queries/images';
+import { useRegionsQuery } from 'src/queries/regions/regions';
+
+import { atou, utoa } from '../../LinodesCreate/utilities';
```
abailly-akamai, line 13: I assume you're planning to move utils eventually? wondering about duplicating them but then we'd miss potential updates

abailly-akamai (commented):
Looks great and appreciate the small, well tested PR

Prob an existing behavior - we should prevent event propagation on the accordion header tooltip so it does not collapses it onclick. This is prob very annoying on mobile

https://github.com/linode/manager/assets/130582365/cd6f3d80-c5dc-42aa-bcfe-f91b8c0ba906

packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserData.tsx
bnussman-akamai: I'll do all of the moving when we start deleting the old create flow code. It will be a pain, but TS should help me identify the utils that need to be kept and moved when I delete a file and see an error

packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserDataHeading.tsx
bnussman-akamai: Kind of addressed in cf72f9fd45

bnussman-akamai:
> we should prevent event propagation on the accordion header tooltip 

Good catch on this @abailly-akamai . Do you think this should be a global change to the `TooltipIcon` component?

packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserData.tsx
bnussman-akamai: Fair enough. I will change this ✅

abailly-akamai:
@bnussman-akamai i can't really think of a time we'd want the click to propagate so I'd say yes?

bnussman-akamai:
Sounds good, I agree! I also can't think of any instances where we'd want the click to propagate. Thanks @abailly-akamai

packages/manager/src/features/Linodes/LinodeCreatev2/UserData/UserData.tsx
bnussman-akamai: Added a transform step in 1803a36

abailly-akamai (approved):
Thanks for the changes!

PR #10337, opened 2024-04-01
upcoming: [M3-7861]: Include powered-off data in clone analytics event

hkhalil-akamai:
## Description 📝
Updates the Create Linode analytics event when cloning to include whether the source Linode was powered-off.

## Target release date 🗓️
4/15

## How to test 🧪

- Enable the Linode Clone UI Feature flag
- Go to http://localhost:3000 and turn on analytics logs for Adobe: 
   - Ensure `REACT_APP_ADOBE_ANALYTICS_URL` is set in your .env
   - In the browser console, type `_satellite.setDebug(true)`
- Clone a powered-on/off Linode
- Verify the Create Linode event is fired and the `data.isLinodePoweredOff` field reflects the power status of the origin Linode
- Optional (it takes about ~1 hr for Adobe dashboards to update): 
   - Log into Adobe Analytics
   - Go to the Projects folder and open the Linode Cloud - Dev Dashboard. Scroll down to the bottom two graphs, "Component Details vs. Component Clicks" and "Number of Events With Linodes Powered Off" and observe the data before and ~1 hour after your testing, which should now include metrics for your event(s) that fired.

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

---


hkhalil-akamai:
@mjac0bs While we wait on changes on the analytics side, I wanted to prototype what custom analytics key-value pairs might look like based on my understanding of the changes.

packages/manager/src/utilities/analytics.ts
```diff
@@ -8,19 +8,13 @@ declare global {
...
 };
 
-interface PageViewEvent {
-  url: string;
-}
-
 interface AnalyticsEvent {
   action: string;
   category: string;
+  details?: Record<string, unknown>;
```
hkhalil-akamai, line 17: This is the new field. I'm allowing all types for the value and the whole object is stringified before being sent.

packages/manager/src/features/Linodes/LinodesCreate/LinodeCreateContainer.tsx
```diff
@@ -995,11 +998,13 @@ const actionsAndLabels = {
...
-  type: CreateTypes,
-  payload: CreateLinodeRequest,
-  label?: string
-) => {
+const handleAnalytics = (details: {
+  fromLinode?: Linode;
+  label?: string;
+  payload: CreateLinodeRequest;
+  type: CreateTypes;
+}) => {
```
mjac0bs, line 1006: Should we pull this out into an interface?

packages/manager/src/features/Linodes/LinodesCreate/LinodeCreateContainer.tsx
```diff
@@ -995,11 +998,13 @@ const actionsAndLabels = {
   fromStackScript: { action: 'stackscript', labelPayloadKey: 'stackscript_id' },
 };
 
-const handleAnalytics = (
-  type: CreateTypes,
-  payload: CreateLinodeRequest,
-  label?: string
-) => {
+const handleAnalytics = (details: {
+  fromLinode?: Linode;
```
mjac0bs, line 1002:
```suggestion
  linode?: Linode;
```
`fromLinode` feels misleading because I would expect a boolean, and we're receiving an actual linode.

packages/manager/src/features/Linodes/LinodesCreate/LinodeCreateContainer.tsx
```diff
@@ -1018,5 +1023,11 @@ const handleAnalytics = (
     eventLabel = label;
   }
 
-  sendCreateLinodeEvent(eventAction, eventLabel);
+  sendCreateLinodeEvent(
+    eventAction,
+    eventLabel,
+    fromLinode
+      ? { linodeWasPoweredOff: fromLinode.status === 'offline' }
+      : undefined
```
mjac0bs, line 1031: I'd pull this out into something like:
```
  let eventData: CustomAnalyticsData = {};
  [ . . .]
  if (linode) {
    eventData['isLinodePoweredOff'] = linode?.status === 'offline';
  }
  sendCreateLinodeEvent(eventAction, eventLabel, eventData);
```
to make it consistent with the function/cleaner to read. See my other comment in analytics.ts about the interface.

packages/manager/src/utilities/analytics.ts
mjac0bs, line 17:
```suggestion
  data?: Record<string, unknown>;
```
I think Richa went with `data` here, rather than `details`, so that's a change we would need to make to get this to the backend. 

I was also thinking that we create an interface for the `data` (details) object:
```
export interface CustomAnalyticsData {
  isLinodePoweredOff?: boolean;
  value?: number; //We *could* move our existing "value" field here
  //We could add other things in future
}
```
so that if we wanted to add something new to our `data`, we could, and we're making sure we're typing it correctly. Thoughts?

packages/manager/src/utilities/analytics.ts
```diff
@@ -150,11 +145,13 @@ export const sendCreateNodeBalancerEvent = (eventLabel: string): void => {
 // LinodeCreateContainer.tsx
 export const sendCreateLinodeEvent = (
   eventAction: string,
-  eventLabel: string
+  eventLabel: string,
+  details?: Record<string, unknown>
```
mjac0bs, line 149:
```suggestion
  data: CustomAnalyticsData
```
If we did the above

packages/manager/src/utilities/analytics.ts
```diff
@@ -150,11 +145,13 @@ export const sendCreateNodeBalancerEvent = (eventLabel: string): void => {
...
 ): void => {
   sendEvent({
     action: eventAction,
     category: 'Create Linode',
+    details,
```
mjac0bs, line 154:
```suggestion
    data,
```

packages/manager/src/utilities/analytics.ts
```diff
@@ -8,19 +8,13 @@ declare global {
 }
 
 type DTMSatellite = {
-  track: (
-    eventName: string,
-    eventPayload: AnalyticsEvent | PageViewEvent
-  ) => void;
+  track: (eventName: string, eventPayload: Record<string, unknown>) => void;
```
mjac0bs, line 11: Hmm, I don't think we should change this.

mjac0bs (commented):
Thanks for prototyping this!  As I played with this locally, I actually wasn't seeing the event fire with `data` (after I changed it from `details`) like I would expect, so I've let Richa know and will follow up with what she says -- I think there's probably still a misconfiguration on the backend because we can do something like this:

Screenshot 2024-04-02 at 12 16 01 PM

In the meantime, I'm leaving some thoughts here.

packages/manager/src/features/Linodes/LinodesCreate/LinodeCreateContainer.tsx
hkhalil-akamai: How about `clonedLinode`? -- I feel that's even more explicit.

packages/manager/src/features/Linodes/LinodesCreate/LinodeCreateContainer.tsx
hkhalil-akamai: I'm hesitant to do so since it's only used once...

packages/manager/src/features/Linodes/LinodesCreate/LinodeCreateContainer.tsx
mjac0bs: A little misleading, though, since this function is still going to run for other methods of creating a linode, and some of those do have linodes associated with them. (Create from Backup, for example, and the event will fire and include the powered on/off data.)

packages/manager/src/features/Linodes/LinodesCreate/LinodeCreateContainer.tsx
```diff
@@ -1018,5 +1023,11 @@ const handleAnalytics = (
     eventLabel = label;
   }
 
-  sendCreateLinodeEvent(eventAction, eventLabel);
+  sendCreateLinodeEvent(
+    eventAction,
+    eventLabel,
+    clonedLinode
+      ? { isLinodePoweredOff: clonedLinode.status === 'offline' }
+      : undefined
```
mjac0bs, line 1031: if we only want to send on clone create events, not any create event with an existing linode:
```suggestion
    linode && eventAction === 'clone'
      ? { isLinodePoweredOff: linode.status === 'offline' }
      : undefined
```
Since we have the graph that shows component details vs component clicks and we can see clone events vs any other type of event there, we don't *need* to only collect powered on/off data on clone events, but it does make it more relevant to graph the "total number of linodes powered on/off".

mjac0bs (approved):
This is looking good from local testing with the changes in the Adobe dev environment! Approving pending my couple of additional comments re: variable name and tracking isLinodePoweredOff for events other than clone.

| Clone with Linode Powered On | Clone with Linode Powered Off |
| --- | --- |
| Screenshot 2024-04-09 at 8 08 18 AM Screenshot 2024-04-09 at 8 08 31 AM | Screenshot 2024-04-09 at 8 10 54 AM Screenshot 2024-04-09 at 8 11 08 AM|

While it takes about 1 hour for event data to show up in the AA dashboard, I did set up a graph in the Linode Cloud - Dev Dashboard and can see that our new custom component data is being associated with their respective click events:
image

mjac0bs:
@hkhalil-akamai I'm going to use M3-7964 to update our documentation on Adobe in our developer guide to reflect the latest changes.

packages/manager/src/utilities/analytics.ts
```diff
@@ -10,21 +10,28 @@ declare global {
...
+
 interface AnalyticsEvent {
   action: string;
   category: string;
+  data?: CustomAnalyticsData;
   label?: string;
   value?: number;
 }
 
+type AnalyticsPayload = Omit<AnalyticsEvent, 'data'> & { data?: string };
```
jaalah-akamai, line 33: nit: I would prefer we extend via an interface 
```
interface AnalyticsPayload extends Omit<AnalyticsEvent, 'data'> {
  data?: string;
}
```

jaalah-akamai (approved):

mjac0bs:
I'm going to go ahead and merge this. e2es failed on the final runner because of the hanging test in `account-switching.spec.ts`. This test is a known issue, unrelated to these PR changes.

PR #9992, opened 2023-12-12
upcoming, change: [M3-7524] - Add child_account OAuth scope to Create and View PAT drawers

mjac0bs:
## Description 📝
New Scope:
Introduce a new OAuth scope named "child_account" for all users, but visible only to "Parent" accounts. As per API spec: its purpose is to allow Parent Account Users with the `child_account_access` grant to specify a Token with less permission than their User, i.e. a Token that is unable to list and/or create Tokens for the Child Account(s).

Token Configuration:
Default all scopes to `none` for users. Users should enable scopes as needed, contrary to having the initial state set to `read_write`. 

## Changes  🔄
- In the Create PAT drawer, a new "Child Account Access" scope is visible for **parent** users.
- In the View PAT drawer, a new "Child Account Access" scope is visible for **parent** users.

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2023-12-13 at 4 01 40 PM | Screenshot 2023-12-13 at 4 00 48 PM |
| Screenshot 2023-12-13 at 4 02 58 PM | Screenshot 2023-12-13 at 4 03 14 PM |

### Prerequisites
(How to setup test environment)
- Check out this PR and `yarn dev`.
- In the dev tools:
   - Make sure the Parent/Child Account Switching feature flag is on.
   - Turn mocks on.

### Verification steps 
(How to verify changes)
- Go to http://localhost:3000/profile/tokens and click the Create a Personal Access Token button.
- Observe that "Child Account Access" is visible as a new scope and all scopes default to None.
- On the tokens landing page, click the View Scope button for any mock PAT.
- Observe that the "Child Account Access" scope is visible in the table.
- Toggle the feature flag off and observe that the Child Account Access token is not visible in the Create or View drawers and all scopes default to Read/Write.
- Toggle the feature flag back on.
- Go to `serverHandler.ts` and edit the following request to change the user_type to `null`, `child`, or `proxy`:
```
  rest.get('*/account/users/:user', (req, res, ctx) => {
    // Parent/Child: switch the `user_type` depending on what account view you need to mock.
    return res(ctx.json(accountUserFactory.build({ user_type: 'parent' })));
  }),
```
- Observe that the Child Account Access token is not visible in the Create or View drawers.
- Verify tests pass:
```
yarn test utils CreateAPITokenDrawer ViewAPITokenDrawer

```

## As an Author I have considered 🤔

*Check all that apply*

- [x] 👀 Doing a self review
[...]

PR #10338, opened 2024-04-01
change: [M3-7729] - Improve UX of Personal Access Token and Access Key drawers

mjac0bs:
## Description 📝
Based on discussion in cafe, we want to make a couple of improvements to the Personal Access Token and OBJ Access Key drawers to prevent users from unintentionally limiting their access (e.g. creating a PAT with all scopes set to None), while forcing the user to make a conscious decision about the access selections that they make.

Note: `AccessKeyDrawer.tsx` is going to be deprecated as a component once OBJ Multi-Cluster is live, so updates are limited to the OMC bucket access keys components.

## Changes  🔄
In both the Create PAT Drawer and Create OBJ Access Key Drawer:

- No scopes are selected by default
- "None" is changed to "No Access" (_already_ done in #10188 for OBJ access keys; _also_ done in ViewAPITokenDrawer)
- The "Select All" row is separated from the other rows by a bolded bottom border
- The submit button is disabled until every scope has a selection made
- Tests are updated

Note: If anyone has suggestions on how to fix the padding for the Read/Write column, I'd appreciate it! Other than making the PAT drawer wide, I haven't been able to find a way to add spacing to that column that actually works with the table `display` CSS properties. 

## Preview 📷

| Component | Before  | After   |
| ------- | ------- | --- |
| Create a PAT | cloud linode com_profile_tokens (2) | localhost_3000_object-storage_access-keys |
| Create an OBJ Access Key | <video src="https://github.com/linode/manager/assets/114685994/3eb78805-a2b5-4815-aa94-73bae46962e6" /> | <video src="https://github.com/linode/manager/assets/114685994/e06c3bf1-5a7c-4d0f-b885-d13ce686b71a" /> |

### Reproduction steps
- Go to https://cloud.linode.com/profile/tokens and click "Create a Personal Access Token".
- Observe the default scope selection is "None" and the Select All row is not obviously separated from the other scopes.
- Check out develop and turn the OBJ Multi-Cluster feature flag on and the mocks on.
- Go to http://localhost:3000/object-storage/access-keys/ and click "Create Access Key".
- Observe that when there are no regions selected, all three radio buttons are selected in the Bucket Permissions Table.
- Observe that once a region is selected, the radio button selection is "No Access" by default. 
- Observe that the Select All row is not obviously separated from the other scopes.

### Verification steps
(How to verify changes)
- Check out this PR and `yarn up`.
- Go to http://localhost:3000/profile/tokens and click "Create a Personal Access Token".
- Observe the Changes outlined above. 
- Confirm that you can still submit PATs with no scopes, read-only scopes, and `*` (all scopes).
- Go to http://localhost:3000/object-storage/access-keys and click "Create Access Key".
- Observe the Changes outlined above.
- Confirm the request payload when creating access keys with various region and permissions selections.
   - Example 1:
      - Limited Access field is toggled on with No Access selected
      - Payload contains `bucket_access: []` empty array
   - Example 2:  
      - Limited Access field is toggled off (unlimited access)
      - Payload contains `bucket_access: null`
   - Example 3: 
      - Limited Access field toggled on with some read / read write
      - Payload contains `bucket_access : [{....}]`
- Confirm all tests pass:
```
yarn test utils CreateAPITokenDrawer ViewAPITokenDrawer
```
```
yarn cy:run -s "cypress/e2e/core/account/personal-access-tokens.spec.ts, cypress/e2e/core/objectStorage/access-keys.smoke.spec.ts"
```

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


cpathipa (approved):
LGTM! 
- [x] Confirming on visual changes on OBJ Access key drawer. 
- [x] Able to create / update limited access key.
- [x] Create a Personal Access Token

cpathipa:
@mjac0bs Hold on my approval, I think we should find a way to control default behavior of radio buttons with out diverging form API. I think there is an opportunity for little bit cleanup.  I will take a deeper look later today.

packages/manager/.changeset/pr-10338-changed-1712685790473.md
```diff
@@ -0,0 +1,5 @@
+---
+"@linode/manager": Changed
+---
+
+Improve the UX of access key drawers ([#10338](https://github.com/linode/manager/pull/10338))
```
dwiley-akamai, line 5:
```suggestion
Improve the UX of Access Token & Access Key drawers ([#10338](https://github.com/linode/manager/pull/10338))
```

dwiley-akamai (commented):
From the PR description: "The submit button is disabled until every scope has a selection made" -- is this the case for the PAT drawer currently? I'm observing this:

<details>
<summary>PAT drawer screenshots</summary>

Screenshot 2024-04-10 at 2 02 03 PM

Tried to actually submit on this one, but got an error about the `child_account` scope (which shouldn't be at play for this particular account since it isn't a parent/child scenario?)

Screenshot 2024-04-10 at 2 04 36 PM


</details>

carrillo-erik (approved):

carrillo-erik:
Verified functionality and that all tests are passing. Below is a screenshot of the `Access Keys` (redacted) I created and the payload contents which show it aligns with the description of the PR.

Screenshot 2024-04-10 at 12 38 14 PM

Screenshot 2024-04-10 at 12 37 58 PM

packages/api-v4/src/object-storage/types.ts
```diff
@@ -17,7 +17,7 @@ export type AccessType = 'read_only' | 'read_write' | 'none';
 
 export interface Scope {
   bucket_name: string;
-  permissions: AccessType;
+  permissions: AccessType | null;
```
cpathipa, line 20: There could a potential risk that we could send `none` or `null` as the permissions, which are not inline with the API [doc](https://www.linode.com/docs/api/object-storage/#object-storage-key-create).

mjac0bs: After internal discussion: working on reverting this change, keeping this Scope interface true to the API spec, and creating a new Scope interface for the radio buttons, for which the permissions can have `none` and `null` values.

bnussman-akamai:
I really liked how we set all scopes to "None" by default in the old flow. I feel like this new UX creates slightly more work and confusion for the user. 

For example if they fill out the form as shown in the screenshot, what will the expected behavior be? I don't think it's obvious to the user what permission will be set of the unselected scopes.
<details>
  <summary>Screenshot</summary>
  
</details>

I think defaulting to "No Access" makes a lot of sense. This new flow also requires more clicks / interaction from the user.

I assume UX's argument is that they want the choices to be intentional. This is kind of fair, but I think it will create more work for the average user. This is speculation, but there's probably two types of users who create PATs. 

1. A user who knows exactly what they are doing. I don't see a need to make them do an extra click to set all permissions to none
2. A user who is following some guide on an external site and the guide will tell them what scopes to grant. That guide should't have to have an extra step that says "first, set all grants to "No access", then select Read-Write for VPC"

Additionally, If a user wants to create a PAT with zero scopes, I think we should let them 🤷 


Just my thoughts, they can be completely ignored

mjac0bs:
@bnussman-akamai - There's some history to this one. We used to default to "Read Write", rather than "No Access"/"None". The new default was a [fairly recent change](https://github.com/linode/manager/pull/9992) requested for security best practices that got rolled into some Parent/Child updates. Once we implemented that change, we received some reports that (internal, though I imagine there were also  external) users were mistakenly creating useless keys because they didn't realize that the default had changed to "No Access"/"None". In the internal ticket, I linked a thread with some discussion on this, but we've got UX and backend folks on board with this solution: defaulting to no selections and disabling the button until the required selections have been made. 

> For example if they fill out the form as shown in the screenshot, what will the expected behavior be? I don't think it's obvious to the user what permission will be set of the unselected scopes.

You're right that this was confusing, and it was unintended. Thanks to you and Dajahi for pointing it out; I fixed this in [5e8b3e8](https://github.com/linode/manager/pull/10338/commits/5e8b3e85d38667c5b575c54bc648f6a5377a1096). If we feel  reason for the disabled button is not obvious to the user, we could add a help tooltip to the disabled button, but I'm not sure it's necessary.

<details>
<summary>Screenshot</summary>

Screenshot 2024-04-16 at 2 55 12 PM

</details>

> Additionally, If a user wants to create a PAT with zero scopes, I think we should let them

DX did say there's potentially a legit use-case for this, so it will remain possible like it is today, it will just require the user to Select "No Access" for all scopes.

bnussman-akamai (approved):
Thanks for the history and context!


Changes look good ✅

packages/manager/src/features/ObjectStorage/AccessKeyLanding/OMC_AccessKeyDrawer.tsx
```diff
@@ -52,6 +56,11 @@ export interface AccessKeyDrawerProps {
   open: boolean;
 }
 
+// Access key scopes displayed in the drawer can have no permission or "No Access" selected, which are not valid API permissions.
+export interface DisplayedAccessKeyScope extends Omit<Scope, 'permissions'> {
+  permissions: AccessType | null;
```
mjac0bs, line 61: @cpathipa I looked into also removing `'none'` from the AccessType, but it required more clean up in the original AccessKeyDrawer files, which I wanted to avoid touching since they will be soon deprecated. I did create the new interface here for the `null` value displayed scopes in the OMC drawer and can create a follow up ticket for correcting the AccessType to only the API's values once we switch to fully OMC components.

cpathipa (approved):

PR #10341, opened 2024-04-02
test: [M3-7496] - Add test for Proxy user -> Parent account token expiration prompt

jdamore-linode:
## Description 📝
Adds a test to confirm the flow when a proxy user attempts to switch back to their parent account when their parent account auth token has expired.

## Changes  🔄
- Adds Proxy -> Parent expiration flow test
- Adds missing imports to factories index (and alphabetizes the imports)
- Cleans up Cypress auth by removing invalid local storage entries

## How to test 🧪

```bash
yarn cy:run -s "cypress/e2e/core/parentChild/token-expiration.spec.ts
```
(Also recommend using `cy:debug` so you can observe the flow in action, in case there are any mistakes)


packages/manager/cypress/e2e/core/parentChild/token-expiration.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+const mockChildAccount = accountFactory.build({
+  company: 'Partner Company',
+});
+
+const mockChildAccountProxyUser = accountUserFactory.build({
+  username: randomLabel(),
+  user_type: 'proxy',
+});
+
+const mockChildAccountProfile = profileFactory.build({
```
mjac0bs, line 30:
```suggestion
const mockChildAccountProxyProfile = profileFactory.build({
```
This feels like it might be slightly more clear?

packages/manager/cypress/e2e/core/parentChild/token-expiration.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+      parentChildAccountAccess: makeFeatureFlagData(true),
+    });
+    mockGetFeatureFlagClientstream();
+  });
+
+  /*
+   * - Confirms flow when a Proxy user attempts to switch back to a Parent account with expired auth token.
+   * - Uses mock API and local storage data.
+   */
+  it('shows session expiry prompt upon switching back to Parent account', () => {
```
mjac0bs, line 48:
```suggestion
  it('shows session expiry prompt upon switching back to Parent account with expired Parent token', () => {
```
Small nitpick suggestion for clarity, just thinking about the test to follow in the future when the *proxy* token expires.

mjac0bs (approved):
Thanks, Joe!

image

packages/manager/cypress/e2e/core/parentChild/token-expiration.spec.ts
jdamore-linode: Absolutely, these are both much clearer! Thanks @mjac0bs!

mjac0bs: [L51](https://github.com/linode/manager/pull/10341/files#diff-a4411cf9e5c7abfd273b78dd5c4b06281d6032972bd82182ed26d771b246bfd8R51) too 😅

jdamore-linode: @mjac0bs Ah that's my bad, I should've realized that 🤦‍♂️ will fix tomorrow! Thanks again!

jaalah (approved):
Thanks for capturing this use-case in our test, albeit hopefully it's never an issue!

jaalah-akamai (approved):

carrillo-erik (approved):

PR #10342, opened 2024-04-02
upcoming: [M3-7926] - Linode Create Refactor - VLANs - Part 9

bnussman-akamai:
## Description 📝

Adds VLANs to the new Linode Create flow ✨ 

## Changes  🔄
- Use query key factory for VLANs
- Adds new `VLANSelect` component
- Add infinite query for VLANs to power the VLAN Select
- Adds VLANs section to the new Linode Create flow

## Preview 📷

Screenshot 2024-04-02 at 1 37 58 PM

### Prerequisites
- Turn on the `Linode Create v2` feature flag using our local dev tools

### Verification steps
- Test the new `VLAN` section
- Test the new `VLANSelect`
- Check new logic, utils, and UI for issues, regressions, and quality
- Verify you can create a Linode with a VLAN attached
- Verify VLANs throughout the app still work as expected

- [ ] 👀 Doing a self review
- [x] ➕ Adding a changeset


dwiley-akamai (commented):
I observed a bit of odd behavior in the Linode Create flow with a "ghost" VLAN appearing in the dropdown:

https://github.com/linode/manager/assets/114682940/28b018c3-18fb-439f-9888-b2d12f2eceda

Additionally, [based on the API docs](https://www.linode.com/docs/products/networking/vlans/#additional-limits-and-considerations), VLANs can only be attached to Linodes within the same DC. Our functionality in prod currently filters the VLAN dropdown based on region, but we don't have that on this branch

packages/manager/src/components/VLANSelect.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  const {
+    data,
+    error,
+    fetchNextPage,
+    hasNextPage,
+    isLoading,
+  } = useVLANsInfiniteQuery(searchFilter);
+
+  const vlans = data?.pages.flatMap((page) => page.data) ?? [];
```
abailly-akamai, line 60: curious about the use of `flatMap` here - can we encounter nested arrays in pages? if so, do we need to merger anything with a reduce? Very candid questions cause I know little about those

packages/manager/src/components/VLANSelect.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  const newVlanPlacehodler = VLANFactory.build({
+    label: inputValue,
+  });
```
abailly-akamai, line 64: Never seen a factory used in a customer facing feature. I am not crazy about this cause it's not unusual for developers to change a factory (for various reasons) without having to worry at all how it would affect production code. I'd be in favor of passing a fully typed object instead.

packages/manager/src/components/VLANSelect.tsx
```diff
@@ -0,0 +1,105 @@
...
+          const listboxNode = event.currentTarget;
+          if (
+            listboxNode.scrollTop + listboxNode.clientHeight >=
+              listboxNode.scrollHeight &&
+            hasNextPage
+          ) {
+            fetchNextPage();
+          }
+        },
+      }}
```
abailly-akamai, line 88: just curious if this is over-optimization or we know real cases for hundreds of vlans for a customer. Do we need this considering the search?

packages/manager/src/components/VLANSelect.tsx
```diff
@@ -0,0 +1,105 @@
...
+          props.onChange(value?.label ?? null);
+        }
+      }}
+      onInputChange={(_, value, reason) => {
+        if (reason === 'input' || reason === 'clear') {
+          setInputValue(value);
+        }
+      }}
+      disabled={props.disabled}
+      errorText={props.errorText ?? error?.[0].reason}
```
abailly-akamai, line 106: can we destructure all these props?

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
```diff
@@ -77,5 +78,58 @@ export const getLinodeCreatePayload = (
     payload.metadata.user_data = utoa(payload.metadata.user_data);
   }
 
+  payload.interfaces = getInterfacesPayload(payload.interfaces);
+
   return payload;
 };
+
+export const getInterfacesPayload = (
```
abailly-akamai, line 90: To be consistent we should add JSDocs above the util

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
```diff
@@ -77,5 +78,58 @@ export const getLinodeCreatePayload = (
...
+  interfaces: InterfacePayload[] | undefined
+): InterfacePayload[] | undefined => {
+  if (!interfaces) {
+    return undefined;
+  }
+
+  interfaces = interfaces.filter((i) => {
+    if (i.purpose === 'vpc' && !i.vpc_id) {
+      // If no vpc was selected, clear remove it from the interfaces array
+      return false;
```
abailly-akamai, line 100: feels a bit strange to early return `false` - should it be `undefined` for consistency?

abailly-akamai (commented):
Looks good and well tested. Left some comments.

Also, maybe i am doing something wrong, but ran into this error when submitting

```JSON
{
    "reason": "metadata dict must contain key user_data",
    "field": "metadata"
}
```

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
bnussman-akamai: This is in a filter function. I think returning false is what we want here.

packages/manager/src/components/VLANSelect.tsx
bnussman-akamai: I don't see any concern here because we specify `{ label: inputValue }` and because the API doesn't introduce breaking changes, I don't expect this to need to change.

packages/manager/src/components/VLANSelect.tsx
bnussman-akamai: There won't be any nested arrays in the pages. This flatMap just combines all of the pages into one `VLAN[]`. The same could be accomplished with a reduce

packages/manager/src/components/VLANSelect.tsx
cpathipa: Agree with @abailly-akamai we could limit the usage of factory in components.

packages/manager/src/components/VLANSelect.tsx
```diff
@@ -0,0 +1,149 @@
...
+        }
+      }}
+      disabled={disabled}
+      errorText={errorText ?? error?.[0].reason}
+      inputValue={selectedVLAN ? selectedVLAN.label : inputValue}
+      label="VLAN"
+      loading={isLoading}
+      noOptionsText="You have no VLANs in this region. Type to create one."
+      options={vlans}
+      placeholder="Create or select a VLAN"
```
cpathipa, line 118: UX observation: Currently, the user will only know the loading state upon clicking the field. Could a dynamic placeholder help?
image

cpathipa (approved):
Overall, LGTM I agree with the feedback from A and B. Approving, pending the addressing of their feedback.

packages/manager/src/components/VLANSelect.tsx
bnussman-akamai: I'm hoping that the changes from https://github.com/linode/manager/pull/10122 improves the loading state

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
abailly-akamai: yes my bad

packages/manager/src/components/VLANSelect.tsx
abailly-akamai: > I think for anything customer driven, we should assume they could have thousands of them

That's fair. In this case (outside of this PR), should we consider baking this in the autocomplete, controlled via prop?

packages/manager/src/components/VLANSelect.tsx
abailly-akamai: It's a super light maintenance and entirely catchable via types so still in favor of a typed object

packages/manager/src/components/VLANSelect.tsx
abailly-akamai: should've asked a better question - no risk of duplicated data in the array? (reduce VS flatMap will handle that differently). again, very unfamiliar with this

packages/manager/src/components/VLANSelect.tsx
bnussman-akamai: Because of the complexity and case-by-case nature of the API filtering logic, baking it into the existing component would probably be pretty hard. In the future I think we could POC this and see how it works.

packages/manager/src/components/VLANSelect.tsx
bnussman-akamai: There won't be any risk of duplicate data. All that I'm doing here is flattening the React Query infinite query data into a single array.

packages/manager/src/components/VLANSelect.tsx
bnussman-akamai: Changed to plain object in 3d2600f

packages/manager/src/components/VLANSelect.tsx
```diff
@@ -0,0 +1,154 @@
...
+import { useVLANsInfiniteQuery } from 'src/queries/vlans';
+
+import { Autocomplete } from './Autocomplete/Autocomplete';
+
+import type { Filter } from '@linode/api-v4';
+import type { SxProps, Theme } from '@mui/material';
+
+interface Props {
+  /**
+   * Disabled the Select
```
abailly-akamai, line 12: nit: `Disable the Select`

packages/manager/src/components/VLANSelect.tsx
abailly-akamai: thanks for confirming

packages/manager/src/components/VLANSelect.tsx
abailly-akamai: sounds good 👍

abailly-akamai (approved):
Looks great on my end, nice work! Thanks for addressing those changes as well ✅

hkhalil-akamai:
> I observed a bit of odd behavior in the Linode Create flow with a "ghost" VLAN appearing in the dropdown:

Was there a response to this comment by @dwiley-akamai? I am not able to replicate it myself -- I think it might be different VLANs in different regions with the same name.

packages/manager/src/components/VLANSelect.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+import { VLANFactory } from 'src/factories';
+import { makeResourcePage } from 'src/mocks/serverHandlers';
+import { HttpResponse, http, server } from 'src/mocks/testServer';
+import { renderWithTheme } from 'src/utilities/testHelpers';
+
+import { getVLANSelectFilter, VLANSelect } from './VLANSelect';
+
+describe('VLANSelect', () => {
+  it('should render a default "Images" label', () => {
```
hkhalil-akamai, line 12:
```suggestion
  it('should render a default "VLAN" label', () => {
```

packages/manager/src/components/VLANSelect.tsx
```diff
@@ -0,0 +1,154 @@
...
+  sx?: SxProps<Theme>;
+  /**
+   * The label of the selected VLAN
+   */
+  value: null | string;
+}
+
+/**
+ * A VLAN select component that has the following features
+ * - Infinitly loads VLANs
```
hkhalil-akamai, line 40: Such a small nit but since this component might be used elsewhere... 😅

```suggestion
 * - Infinitely loads VLANs
```

packages/manager/src/features/Linodes/LinodeCreatev2/VLAN.tsx
```diff
@@ -0,0 +1,119 @@
...
+    'Vlans'
+  );
+
+  const isCreatingFromBackup = params.type === 'Backups';
+
+  const disabled =
+    !imageId ||
+    isCreatingFromBackup ||
+    isLinodeCreateRestricted ||
+    !regionSupportsVLANs;
```
hkhalil-akamai, line 45: For parity with the current flow, the VLANs field should reset when disabled goes from false to true.

In the (admittedly contrived) case that the user selects a VLAN and then e.g., clears the image field, the form will still be submittable with the VLAN attached. This causes an API error and also breaks the form (this might be an unrelated), see attached video:

https://github.com/linode/manager/assets/122488130/6c7fbccd-fbdf-4fd8-84aa-6bad07005d07

bnussman-akamai:
I wasn't able to reproduce the "ghost" issue, but the component has undergone some significant changes and it now filters by region. I think we should be good now @hkhalil-akamai @dwiley-akamai

packages/manager/src/features/Linodes/LinodeCreatev2/VLAN.tsx
bnussman-akamai: Yes, good catch. This will have to be addressed at some point. I was holding off on this for now because there will be significant work dedicated to "resetting" form fields under certain conditions like switching tabs. 

I need to find a clean way to handle these "side-effect" state changes

packages/manager/src/components/VLANSelect.tsx
bnussman-akamai: Good catch, thank you!

packages/manager/src/components/VLANSelect.test.tsx
bnussman-akamai: Good catch. Thanks!

packages/manager/src/features/Linodes/LinodeCreatev2/VLAN.tsx
hkhalil-akamai: One idea might be with a `useEffect` as so:

```
React.useEffect(() => {
  if (disabled) {
    // Reset input here
  }
}, [disabled]);

**Review by hkhalil-akamai — APPROVED** [review]
Thanks for addressing feedback @bnussman-akamai!

File: packages/manager/src/components/VLANSelect.tsx [hunk 12]
```diff
@@ -0,0 +1,154 @@
...
+    data,
+    error,
+    fetchNextPage,
+    hasNextPage,
+    isFetching,
+  } = useVLANsInfiniteQuery(apiFilter, open);
+
+  const vlans = data?.pages.flatMap((page) => page.data) ?? [];
+
+  const newVlanPlacehodler = {
```
dwiley-akamai, line 65:
```suggestion
  const newVlanPlaceholder = {
```

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
```diff
@@ -77,5 +78,70 @@ export const getLinodeCreatePayload = (
...
+    payload.metadata = undefined;
+  }
+
+  payload.interfaces = getInterfacesPayload(payload.interfaces);
+
   return payload;
 };
+
+/**
+ * Performans transformation and ordering on the Linode Create "interfaces" form data.
```
dwiley-akamai, line 91:
```suggestion
 * Performs a transformation and ordering on the Linode Create "interfaces" form data.
```

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
```diff
@@ -77,5 +78,70 @@ export const getLinodeCreatePayload = (
...
+ *
+ * We need this so we can put interfaces in the correct order and omit unused iterfaces.
```
dwiley-akamai, line 93:
```suggestion
 * We need this so we can put interfaces in the correct order and omit unused interfaces.
```

dwiley-akamai (approved):
Functionality LGTM now, no longer observing the ghost VLAN issue

bnussman-akamai:
Oops. I forgot to fix those spelling issues @dwiley-akamai. My fault 🤦 

I'll make sure to fix them in the next Linode Create reactor PR

PR #10322, opened 2024-03-27
refactor: [M3-7920, M3-7930] – Refactor VPC queries

dwiley-akamai:
## Description 📝
Refactor the VPC queries to use `createQueryKeys` from `query-key-factory`.

The main objective was to achieve parity with existing functionality, although some small enhancements at the margins were done as well.

## Changes  🔄
- Use `createQueryKeys` to generate VPC & subnet query keys
- Update mock region data to include VPC capability
- Add new `useAllVPCsQuery` and use it in `VPCPanel.tsx` & `useVPCConfigInterface.tsx`
- Remove manual uses of `vpcQueryKey` and `subnetQueryKey` throughout codebase, notably for various query invalidations
- Remove unused `useSubnetQuery`

## Target release date 🗓️
4/15/24

### Verification steps
Essentially the entire VPC offering in Cloud needs to be tested to ensure nothing has broken with these changes.

- VPC landing page
  - query for paginated results should refresh upon VPC creation
- Assign/Unassign Linode flow via drawer accessed from Subnets table
  - should invalidate the paginated VPC, individual VPC, and VPC subnets queries
- Assigning Linode to VPC via Linode Config dialog
  - all VPC queries should be invalidated
- Linode Create flow
  - a newly-created linode assigned to a VPC should have the VPC section present in the Linode Detail header
- Linode Detail page
  - the VPC section in the Linode Detail header should be accurate, and the "IP Addresses" table on the "Network" tab should contain the expected VPC-related IPs
- Deleting a Linode
  - the 'all' VPCs, paginated VPCs, individual VPC, and VPC subnets queries should be invalidated upon Linode deletion

## As an Author I have considered 🤔
- [X] 👀 Doing a self review
- [X] ❔ Our [contribution guidelines](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md)
- [X] 🤏 Splitting feature into small PRs
- [X] ➕ Adding a changeset
- [X] 🧪 Providing/Improving test coverage
- [X] 🔐 Removing all sensitive information from the code and PR description
- [X] 🚩 Using a feature flag to protect the release
- [X] 👣 Providing comprehensive reproduction steps
- [X] 📑 Providing or updating our documentation
- [X] 🕛 Scheduling a pair reviewing session
- [X] 📱 Providing mobile support
- [X] ♿  Providing accessibility support

PR #10354, opened 2024-04-05
upcoming: [M3-7924] - Linode Create Refactor - VPC - Part 10

bnussman-akamai:
## Description 📝

- Adds a VPC section to the new Linode Create flow ✨

## Changes  🔄
- Builds **new** VPC panel that uses react hook from tooling 🔧
- Adds a VPC Select
- Adds `Create a Linode` to the document title to match the existing flow
- Improves error handling for VPC IP ranges

## Preview 📷

Screenshot 2024-04-04 at 8 00 07 PM

### Prerequisites
- Turn on the `Linode Create v2` feature flag 🎏 

### Verification steps
- Test the new VPC selction 🧪 
- Compare functionality to the existing Linode Create flow 🔍 
- Test error handling in the VPC section ❌ 

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/components/VPCSelect.tsx
+++ b/packages/manager/src/components/VPCSelect.tsx
@@ -1,0 +1,37 @@
+import React from 'react';
+
+import { useVPCsQuery } from 'src/queries/vpcs';
+
+import { Autocomplete } from './Autocomplete/Autocomplete';
+
+import type { EnhancedAutocompleteProps } from './Autocomplete/Autocomplete';
+import type { Filter, VPC } from '@linode/api-v4';
+
+interface Props extends Partial<Omit<EnhancedAutocompleteProps<VPC>, 'value'>> {
+  /**
+   * An optional API filter to filter the VPC options
+   */
+  filter?: Filter;
+  /**
+   * The ID of the selected VPC
+   */
+  value: null | number | undefined;
+}
+
+export const VPCSelect = (props: Props) => {
+  const { filter, value, ...rest } = props;
+
+  const { data, isFetching } = useVPCsQuery({}, filter ?? {});
+
+  const selectedVPC = data?.data.find((vpc) => vpc.id === value) ?? null;
+
+  return (
+    <Autocomplete
+      label="VPC"
+      loading={isFetching}
+      options={data?.data ?? []}
+      value={selectedVPC}
+      {...rest}
+    />
+  );
+};
--- a/packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
+++ b/packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
@@ -46,9 +46,9 @@ export const VPC = () => {
     control,
     name: [
       'region',
-      'interfaces.2.vpc_id',
-      'interfaces.2.subnet_id',
-      'interfaces.2.ipv4.vpc',
+      'interfaces.0.vpc_id',
+      'interfaces.0.subnet_id',
+      'interfaces.0.ipv4.vpc',
     ],
   });
 
@@ -91,7 +91,7 @@ export const VPC = () => {
               />
             )}
             control={control}
-            name="interfaces.2.vpc_id"
+            name="interfaces.0.vpc_id"
           />
           {regionId && !regionSupportsVPCs && (
             <Typography>
@@ -127,7 +127,7 @@ export const VPC = () => {
                   />
                 )}
                 control={control}
-                name="interfaces.2.subnet_id"
+                name="interfaces.0.subnet_id"
               />
               {selectedSubnetId && (
                 <>
@@ -156,7 +156,7 @@ export const VPC = () => {
                       />
                     )}
                     control={control}
-                    name="interfaces.2.ipv4.vpc"
+                    name="interfaces.0.ipv4.vpc"
                   />
                   {linodeVPCIPAddress !== null &&
                     linodeVPCIPAddress !== undefined && (
@@ -172,7 +172,7 @@ export const VPC = () => {
                           />
                         )}
                         control={control}
-                        name="interfaces.2.ipv4.vpc"
+                        name="interfaces.0.ipv4.vpc"
                       />
                     )}
                   <Controller
@@ -200,7 +200,7 @@ export const VPC = () => {
                       />
                     )}
                     control={control}
-                    name="interfaces.2.ipv4.nat_1_1"
+                    name="interfaces.0.ipv4.nat_1_1"
                   />
                   <Divider />
                   <Typography variant="h3">
@@ -228,7 +228,7 @@ export const VPC = () => {
         </Stack>
       </Stack>
       <VPCCreateDrawer
-        handleSelectVPC={(vpcId) => setValue('interfaces.1.vpc_id', vpcId)}
+        handleSelectVPC={(vpcId) => setValue('interfaces.0.vpc_id', vpcId)}
         onClose={() => setIsCreateDrawerOpen(false)}
         open={isCreateDrawerOpen}
         selectedRegion={regionId}
--- a/packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
+++ b/packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
@@ -139,57 +139,57 @@ export const VPC = () => {
               />
               {selectedSubnetId && (
                 <>
-                  <Box pl={1.5}>
+                  <Stack>
                     <Controller
                       render={({ field }) => (
-                        <FormControlLabel
-                          checked={
-                            field.value === null || field.value === undefined
-                          }
-                          label={
-                            <Stack alignItems="center" direction="row">
-                              <Typography>
-                                Auto-assign a VPC IPv4 address for this Linode
-                                in the VPC
-                              </Typography>
-                              <TooltipIcon
-                                status="help"
-                                text={VPC_AUTO_ASSIGN_IPV4_TOOLTIP}
-                              />
-                            </Stack>
-                          }
-                          onChange={(e, checked) =>
-                            // If "Auto-assign" is checked, set the VPC IP to null
-                            // so that it gets auto-assigned. Otherwise, set it to
-                            // an empty string so that the TextField renders and a
-                            // user can enter one.
-                            field.onChange(checked ? null : '')
-                          }
-                          control={<Checkbox sx={{ ml: -1 }} />}
-                        />
+                        <Box>
+                          <FormControlLabel
+                            checked={
+                              field.value === null || field.value === undefined
+                            }
+                            label={
+                              <Stack alignItems="center" direction="row">
+                                <Typography>
+                                  Auto-assign a VPC IPv4 address for this Linode
+                                  in the VPC
+                                </Typography>
+                                <TooltipIcon
+                                  status="help"
+                                  text={VPC_AUTO_ASSIGN_IPV4_TOOLTIP}
+                                />
+                              </Stack>
+                            }
+                            onChange={(e, checked) =>
+                              // If "Auto-assign" is checked, set the VPC IP to null
+                              // so that it gets auto-assigned. Otherwise, set it to
+                              // an empty string so that the TextField renders and a
+                              // user can enter one.
+                              field.onChange(checked ? null : '')
+                            }
+                            control={<Checkbox sx={{ ml: 0.5 }} />}
+                          />
+                        </Box>
                       )}
                       control={control}
                       name="interfaces.0.ipv4.vpc"
                     />
-                  </Box>
-                  {linodeVPCIPAddress !== null &&
-                    linodeVPCIPAddress !== undefined && (
-                      <Controller
-                        render={({ field, fieldState }) => (
-                          <TextField
-                            errorText={fieldState.error?.message}
-                            label="VPC IPv4"
-                            noMarginTop
-                            onChange={field.onChange}
-                            required
-                            value={field.value}
-                          />
-                        )}
-                        control={control}
-                        name="interfaces.0.ipv4.vpc"
-                      />
-                    )}
-                  <Box pl={1.5}>
+                    {linodeVPCIPAddress !== null &&
+                      linodeVPCIPAddress !== undefined && (
+                        <Controller
+                          render={({ field, fieldState }) => (
+                            <TextField
+                              errorText={fieldState.error?.message}
+                              label="VPC IPv4"
+                              onChange={field.onChange}
+                              required
+                              sx={{ my: 2 }}
+                              value={field.value}
+                            />
+                          )}
+                          control={control}
+                          name="interfaces.0.ipv4.vpc"
+                        />
+                      )}
                     <Controller
                       render={({ field }) => (
                         <FormControlLabel
@@ -210,14 +210,14 @@ export const VPC = () => {
                             field.onChange(checked ? 'any' : null)
                           }
                           checked={field.value === 'any'}
-                          control={<Checkbox sx={{ ml: -1 }} />}
+                          control={<Checkbox sx={{ ml: 0.5 }} />}
                           sx={{ mt: 0 }}
                         />
                       )}
                       control={control}
                       name="interfaces.0.ipv4.nat_1_1"
                     />
-                  </Box>
+                  </Stack>
                   <Divider />
                   <Typography fontFamily={(theme) => theme.font.bold}>
                     Assign additional IPv4 ranges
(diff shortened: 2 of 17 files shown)


packages/manager/src/components/VPCSelect.tsx
```diff
@@ -0,0 +1,37 @@
...
+  /**
+   * The ID of the selected VPC
+   */
+  value: null | number | undefined;
+}
+
+export const VPCSelect = (props: Props) => {
+  const { filter, value, ...rest } = props;
+
+  const { data, isFetching } = useVPCsQuery({}, filter ?? {});
```
bnussman-akamai, line 24: I will come back and update this to use `useAllVPCsQuery` or an infinite query after https://github.com/linode/manager/pull/10322 gets merged

packages/manager/src/hooks/useCreateVPC.ts
```diff
@@ -181,6 +181,7 @@ export const useCreateVPC = (inputs: UseCreateVPCInputs) => {
   };
 
   const formik = useFormik({
+    enableReinitialize: true,
```
bnussman-akamai, line 184: Had to add this so that `region` updates when the region is changed on the Linode Create flow. This was not an issue with the existing create flow because the drawer was re-mounted when it was opened (causing the VPC create drawer to not animate)

abailly-akamai (commented):
I'll spend more time with it in the UI once we get more reviews from folks more familiar than me with the feature. 

All together this looks great, concise, well thought & put together

Left some initial comments to look at a couple areas that could be improved

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
```diff
@@ -0,0 +1,238 @@
...
+          {regionId && !regionSupportsVPCs && (
+            <Typography>
+              VPC is not available in the selected region.
+            </Typography>
+          )}
+          {regionId && regionSupportsVPCs && (
+            <Box>
+              <LinkButton onClick={() => setIsCreateDrawerOpen(true)}>
+                Create VPC
+              </LinkButton>
```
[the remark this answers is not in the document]
bnussman-akamai: I agree with this. I did it this way to match the existing create flow. Should I make this change now or make a todo to do it after this refactor?

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
```diff
@@ -0,0 +1,238 @@
...
+                              the VPC
+                            </Typography>
+                            <TooltipIcon
+                              status="help"
+                              text={VPC_AUTO_ASSIGN_IPV4_TOOLTIP}
+                            />
+                          </Stack>
+                        }
+                        onChange={(e, checked) =>
+                          field.onChange(checked ? null : '')
```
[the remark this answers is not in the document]
bnussman-akamai: `null` / `undefined` in the payload will make the VPC Auto-assign an IPv4. A string will tell the API that we are specifying an explicit IPv4. 

So, the logic is
- If the "Auto-assign" Checkbox is checked, set the value to `null` so that an IP get auto-assigned
- If the "Auto-assign" Checkbox is unchecked, set the value to an empty string so that the text field shows and a user can type in an IPv4

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
```diff
@@ -96,31 +99,41 @@ export const getLinodeCreatePayload = (
...
-  return interfaces;
+  if (hasVLAN) {
+    return [publicInterface, vlanInterface];
+  }
+
+  if (hasVPC) {
+    return [vpcInterface];
+  }
+
+  return undefined;
```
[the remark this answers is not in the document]
bnussman-akamai: Will add some comments here

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPCRanges.tsx
```diff
@@ -0,0 +1,54 @@
...
+              aria-label={`Remove IP Range ${index}`}
+              onClick={() => remove(index)}
+            >
+              <CloseIcon />
+            </IconButton>
+          </Stack>
+        ))}
+      </Stack>
+      <Box>
+        <LinkButton onClick={() => append('')}>Add IPv4 Range</LinkButton>
```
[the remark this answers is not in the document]
bnussman-akamai: This adds an empty string to the `ip_ranges` array on the VPC interface. Appending an empty string item to the array will cause a new empty TextField to render so that the user can add a new IP range.

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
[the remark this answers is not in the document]
bnussman-akamai: I'll look into adding some comments to clarify how this works

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPCRanges.tsx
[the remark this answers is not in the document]
abailly-akamai: Yeah, i am getting to understand that part. i think even doing 

// Adding an empty string to the array will render to new empty field
`const newRangeField = ''`

=> `append(newRangeField)`

or something like that could help with readability

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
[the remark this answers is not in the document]
abailly-akamai: It's really up to you. Not a blocker for this PR for sure since we're at parity, but perhaps a cafe item cause there's been friction on my end with UX not always agreeing on doing this consistently

dwiley-akamai (commented):
I'm still testing and combing through the code, but it's looking good overall so far. A few comments:

<details>
<summary>Points with screenshots</summary>

1. Can we keep the "X" vertically centered in relation to the input field when there's an error? I think this has come up as an issue in other contexts, maybe there's a related/similar ticket in the backlog somewhere
Screenshot 2024-04-09 at 4 31 38 PM

2. Can we keep both dropdowns the same width to match prod?
Screenshot 2024-04-09 at 5 00 44 PM

3. Can we make some width adjustments so if you click far away from the Auto-assign & Assign checkboxes, they don't get toggled?

https://github.com/linode/manager/assets/114682940/89e4a897-c217-427e-8f23-25f5ad66cef9

4. There were some concerns during the initial IPv4 range work a couple of months ago about large header text like this (doesn't make the section hierarchy as clear) instead of the more understated one currently used in prod

Screenshot 2024-04-09 at 5 36 55 PM
</details>

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPCRanges.tsx
```diff
@@ -0,0 +1,54 @@
...
+            <Controller
+              render={({ field, fieldState }) => (
+                <TextField
+                  errorText={fieldState.error?.message}
+                  hideLabel
+                  label={`IP Range ${index}`}
+                  onChange={field.onChange}
+                  placeholder="10.0.0.0/24"
+                  value={field.value}
+                />
```
hkhalil-akamai, line 35: The width of this textfield is shorter (166.5px vs 362.41px) than the original. Personally I prefer the shorter width since it more closely matches the expected length of the subnet value but just wanted to point it out.

hkhalil-akamai (approved):
Tested and verified parity with existing VPC flow, code review looks clean too -- excellent work as always!

Also, thanks for including comprehensive unit tests!

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
```diff
@@ -0,0 +1,242 @@
...
+    selectedVPCId ?? -1,
+    Boolean(selectedVPCId)
+  );
+
+  return (
+    <Paper>
+      <Stack spacing={2}>
+        <Typography variant="h2">VPC</Typography>
+        <Typography>
+          Assign this Linode to an existing VPC.{' '}
```
abailly-akamai, line 71: Production has "Allow Linode to communicate in an isolated environment." is this intentional?

packages/manager/src/components/VPCSelect.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  const selectedVPC = data?.data.find((vpc) => vpc.id === value) ?? null;
+
+  return (
+    <Autocomplete
+      label="VPC"
+      loading={isFetching}
+      options={data?.data ?? []}
+      value={selectedVPC}
+      {...rest}
```
abailly-akamai, line 34: We need to reset the VPC panel on region change. Repro:

- select an available region (ex: chicago)
- select a VPC
- change to a non-available region. ex: Newark
- see screenshot below

Screenshot 2024-04-10 at 09 37 10

You can basically keep interacting with the whole VPC form as the previous valus is still selected

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
```diff
@@ -0,0 +1,242 @@
...
+            </>
+          )}
+        </Stack>
+      </Stack>
+      <VPCCreateDrawer
+        handleSelectVPC={(vpcId) => setValue('interfaces.0.vpc_id', vpcId)}
+        onClose={() => setIsCreateDrawerOpen(false)}
+        open={isCreateDrawerOpen}
+        selectedRegion={regionId}
+      />
```
abailly-akamai, line 243: It looks to me like the VPC drawer gets unmounted after the first close. Subsequent openings break transition and we can notice a weird delay when it opens

https://github.com/linode/manager/assets/130582365/2d589bb9-c7cc-485e-8346-c02c5ae2c7f0

packages/manager/src/components/VPCSelect.tsx
```diff
@@ -0,0 +1,37 @@
...
+    />
+  );
+};
```
abailly-akamai, line 37: Also, this is at parity with current behavior, but it would be nice to have consistency there.

For firewalls, we can either select a firewall and clear the field which will set it back to "None"

in VPC, "None" is an option, which i feel is unecessary.

Either way would be nice to have consistency and maybe ask UX what should be the default.

bnussman-akamai:
I fixed point 1, 3, and 4. @dwiley-akamai 

I personally don't agree with 4. I was thinking an `h3` would be fine if every other header on the page is an `h2`. I've made the change regardless. 

I'm declaring point 2 out of scope because I told myself I wouldn't write one-off styling in this project 😖 I think the issue would have to be fixed at the Autocomplete component level if we want it to work that way globally.

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPCRanges.tsx
bnussman-akamai: Yeah, it does look slightly odd. I wasn't able to easily fix it without writing one-off styles, so I'd prefer to leave it be

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
```diff
@@ -0,0 +1,242 @@
...
+                  )}
+                  <Typography>
+                    Assign additional IPv4 address ranges that the VPC can use
+                    to reach services running on this Linode.{' '}
+                    <Link to="https://www.linode.com/docs/products/networking/vpc/guides/assign-services/">
+                      Learn more
+                    </Link>
+                    .
+                  </Typography>
+                  <VPCRanges />
```
abailly-akamai, line 231: Same thing here. I expect my subnet IPv4 ranges to be reset when clearing the field or selecting a new subnet

https://github.com/linode/manager/assets/130582365/125b5566-713c-4b96-b777-d2862f2d74db

abailly-akamai (changes requested):
Took a second pass at the UI and found regressions/improvements to be made to the UX. See comments for details

packages/manager/src/components/VPCSelect.tsx
bnussman-akamai: This is expected for now. I am saving "side effect" state update for later in this project.

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
bnussman-akamai: Will handle these "side-effect" state updates in a later PR.

packages/manager/src/components/VPCSelect.tsx
abailly-akamai: This is a local side effect, not a global one tho. You gotta point those out in your PR description

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
bnussman-akamai: I'm unable to reproduce this.

packages/manager/src/components/VPCSelect.tsx
bnussman-akamai: Yeah, I thought using an option for "None" was pretty crazy.  

I went ahead an fixed this both here and in the Firewall section to use a placeholder instead of "None" placeholder option. 

This does break parity slightly, but it is a no-brainer change imo.

packages/manager/src/components/VPCSelect.tsx
abailly-akamai: Got it. Would still be helpful for reviewers to know what's not supposed to work. I remembered default values being one, not those (tho they are clearly related).

bnussman-akamai: I have a running list so I assure you these will be fixed! I will do my best to call these out in the future

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
abailly-akamai: hmm it's on FF only. Let's welcome new unfixable behaviors 😓 

Resolving this, not worth going into

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
abailly-akamai: Note for UX: 
Screenshot 2024-04-10 at 10 39 30

It's a tiny bit unclear for the user you need to select a region to be able to select a VPC. Again, this is parity but something to think about. Weird it never came up 🤷

bnussman-akamai: Agreed. I'll make an effort to circle back around to all of the weird UX things after this refactor happens

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
```diff
@@ -0,0 +1,254 @@
...
+    Boolean(selectedVPCId)
+  );
+
+  // This is here only to determine which copy should show...
+  const { data } = useVPCsQuery({}, {});
+
+  const copy =
+    data?.results === 0
+      ? 'Allow Linode to communicate in an isolated environment.'
+      : 'Assign this Linode to an existing VPC.';
```
dwiley-akamai, line 72: minor: We lost some parity here -- a fresh form in prod shows `Allow Linode to communicate in an isolated environment.` but this branch shows `Assign this Linode to an existing VPC.`

dwiley-akamai (approved):
The core functionality LGTM, nice work as the code patterns are pretty clean 🧽

I do think the remaining visual inconsistencies and regressions have to be addressed, but that can be in a follow-up or at least a circle-back before the Linode Create refactor is wrapped up.

<details>
<summary>Other quirks with IPv4 ranges</summary>

When there's an error, the field it belongs to gets horizontally widened, which makes me favor retaining the wider field width that we currently have in prod.

Screenshot 2024-04-10 at 11 22 13 AM

The vertical distance between the fields is now greatly compressed compared to prod. I think that should be adjusted.

This isn't an issue per se, but I noticed that compared to prod, you now have to input something into the most recent field before clicking "Add IPv4 Range" adds a new input field.
</details>

Also, were you going to address the leftover changes from https://github.com/linode/manager/pull/10342#issuecomment-2037686354 in this PR?

packages/manager/src/features/Linodes/LinodeCreatev2/VPC/VPC.tsx
bnussman-akamai: Good catch. I believe the logic should now match the prod flow.

PR #10361, opened 2024-04-08
feat: [M3-7921] - Added Dialog to Refresh Proxy Tokens as Time Expires

jaalah-akamai:
## Description 📝
When time is about to expire, we want to prompt the user with a modal window that indicates they will be logged out soon (show when 5min is remaining) and ask them if they want to continue working (refresh their token).

> [!note]
I also did refactoring in `SwitchAccountDrawer` to implement the `useParentChildAuthentication`, hopefully making the file more readable.

## Additional Issues
- **Parent Account:** The drawer lacks a loading indicator when choosing a child account. This can lead to multiple selections and redundant requests for slow networks.
- **Proxy Account:** Slow networks can cause the "Switch back to your account" action to initiate multiple requests.
- **Token Revocation:** Currently, a toast notification appears when revoking a token. However, it often fails to display before the page is refreshed during the switch. Given the refresh, the toast notification is unnecessary.

### New Files:
- **SessionExpirationDialog**: Displayed for users logged into a proxy account. It performs three main functions:
   - Initiates a countdown timer based on the token's expiration time, displaying the dialog when only 5 minutes remain.
   - Provides options to either refresh the token or dismiss the dialog, allowing the token to expire naturally.
   - Provides option to logout (returns to parent account if parent token is valid or logs out if not)
- **useInterval**: Introduced a familiar hook for reliable timer management.
- **useParentChildAuthentication**: In order to make the creation/revoking/updating/validating of tokens, I created this hook to make things more DRY. This allows us to do a lot of cleanup and in several files (SessionExpirationDialog, SwitchAccountDrawer, AccountLanding).

### Deleted:
- `usePendingRevocationToken`: We are now using `usePersonalAccessTokensQuery` directly in places it's needed.

### Changes:
- Renamed `useParentTokenManagement` to `useIsParentTokenExpired` for clarity

## Target release date 🗓️
4/29

## Preview 📷
https://github.com/linode/manager/assets/125309814/1fc62b69-4ed0-430e-b79d-8c6d69c9a68e

### Prerequisites
- Log into Parent account using developer credentials

#### Loading Indicators:
- Open user menu and "Switch Account"
- Select a child company and observe there's no loading indicator
- In addition, if you slowed network you can repeatedly send requests
- In proxy account (after switching to child company), open user menu and "Switch Account"
- Observe same issue with "Switch back to your account"

#### Loading Indicators:
- Checkout branch and log into Parent account
- Open user menu and "Switch Account"
- Observe that selecting child company shows loading indicator
- Observe the same process for Proxy account when trying to "Switch back to your account"

#### Session Expiry:
- Checkout branch and log into Parent account
- Change [L70](https://github.com/linode/manager/blob/ec2d75ad72a9d8b17d841e3dc4f3744f6aa8aac4/packages/manager/src/features/Account/SwitchAccounts/SessionExpirationDialog.tsx#L70) to `if (timeRemaining.minutes < 15) {` to observe modal
- Test that closing modal should do nothing allowing user to return to work without refreshing token
- Test that logging out sends user back to parent account
- Change [L100](https://github.com/linode/manager/blob/ec2d75ad72a9d8b17d841e3dc4f3744f6aa8aac4/packages/manager/src/features/Account/SwitchAccounts/SessionExpirationDialog.tsx#L100) to `true` to test that logging out sends you to log in
- Test that "Continue Working" refreshes token and closes modal.

*Check all that apply*

- [x] 👀 Doing a self review
- [ ] ➕ Adding a changeset


packages/manager/src/MainContent.tsx
```diff
@@ -295,105 +301,107 @@ export const MainContent = () => {
...
-                    desktopMenuIsOpen ||
-                    (desktopMenuIsOpen && desktopMenuIsOpen === true),
-                })}
-              >
-                <TopMenu
-                  desktopMenuToggle={desktopMenuToggle}
-                  isSideMenuOpen={!desktopMenuIsOpen}
-                  openSideMenu={() => toggleMenu(true)}
-                  username={username}
+      <SessionExpirationProvider value={sessionExpirationContextValue}>
```
jaalah-akamai, line 304: Only change in JSX was to add a new context provider

packages/manager/src/features/Account/SwitchAccountDrawer.tsx
```diff
@@ -1,73 +1,73 @@
...
+import { updateParentTokenInLocalStorage } from './SwitchAccounts/utils';
 
-import type { APIError, Token, UserType } from '@linode/api-v4';
+import type { APIError, UserType } from '@linode/api-v4';
 import type { State as AuthState } from 'src/store/authentication';
 
 interface Props {
   onClose: () => void;
   open: boolean;
-  proxyToken?: Token;
```
jaalah-akamai, line 30: No need to pass token in anymore, we're handling this within component now.

packages/manager/src/features/Account/SwitchAccountDrawer.tsx
```diff
@@ -1,73 +1,73 @@
...
-    proxyToken?.id ?? -1
-  );
-  const { enqueueSnackbar } = useSnackbar();
-  const currentTokenWithBearer = useCurrentToken() ?? '';
 
+  const isProxyUser = userType === 'proxy';
   const currentParentTokenWithBearer =
     getStorage('authentication/parent_token/token') ?? '';
 
-  const handleProxyTokenRevocation = React.useCallback(async () => {
```
jaalah-akamai, line 55: Handled by `useParentChildAuthentication.revokeToken` and removed toasts since they're not necessary when switching accounts.

packages/manager/src/features/Account/SwitchAccountDrawer.tsx
```diff
@@ -1,73 +1,73 @@
...
+  const {
+    createToken,
+    createTokenError,
+    createTokenLoading,
+    revokeToken,
+    updateCurrentToken,
+    validateParentToken,
+  } = useParentChildAuthentication({
+    tokenIdToRevoke: pendingRevocationToken?.id ?? -1,
+  });
```
jaalah-akamai, line 68: This is how the new hook will be used

packages/manager/src/features/Account/SwitchAccounts/utils.ts
```diff
@@ -45,33 +39,79 @@ export const updateParentTokenInLocalStorage = ({
 };
 
 /**
- * Headers are required for proxy users when obtaining a proxy token.
- * For 'proxy' userType, use the stored parent token in the request.
+ * Determine whether the tokens used for switchable accounts are still valid.
+ */
+export const isParentTokenValid = (): boolean => {
```
jaalah-akamai, line 44: `isParentTokenValid` and `setTokenInLocalStorage` were moved in from parent utils file since they pertain to Account Switching.

packages/manager/src/features/Account/utils.ts
```diff
@@ -65,89 +63,3 @@ export const getRestrictedResourceText = ({
...
-export async function getPersonalAccessTokenForRevocation(
-  tokens: Token[],
-  currentTokenWithBearer: string
-): Promise<Token | undefined> {
-  return tokens.find(
-    (token) =>
-      token.token &&
-      currentTokenWithBearer.replace('Bearer ', '').startsWith(token.token)
-  );
-}
```
jaalah-akamai, line 153: Moved all these to `/SwitchAccounts/utils`

packages/manager/src/features/Account/SwitchAccounts/useParentChildAuthentication.tsx
```diff
@@ -0,0 +1,73 @@
...
+          Authorization: getStorage('authentication/parent_token/token'),
+        },
+      });
+    },
+    [createProxyToken]
+  );
+
+  const revokeToken = useCallback(async (): Promise<{}> => {
+    return revokeAccessToken();
+  }, [revokeAccessToken]);
```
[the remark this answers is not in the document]
jaalah-akamai: @bnussman-akamai I like this, but we would lose the error and loading states for revoking... but we're not actually using those, so I think we're good

jaalah-akamai:
- @bnussman-akamai Good suggestion on using the endpoint directly, was able to clean things up in several files
- @mjac0bs Hitting the endpoints on slow network requests should be fixed with the way I switched things around, essentially the loading state from RQ was taking too long, so handling this with state

mjac0bs (approved):
Switching accounts from parent to proxy and back, as well as to other child accounts via proxy is still looking good with the latest. Confirmed the slow network edge case behavior I was seeing before is fixed. The dialog still works as expected; I set it to 14 minutes and confirmed it popped up on time. New dialog copy looks good to me. 

This should be unrelated to this PR; I just happened to notice it while testing: token revoke events, which are plentiful if you switch between accounts or continue working many times, try to make a network request to `/account/users/[proxy-username]` and get 403 forbidden as a response. It looks gravatar related to me. 😓  (just furthering our hopes 🪓 )

Screenshot 2024-05-02 at 2 43 56 PM

bnussman-akamai (approved):

PR #10365, opened 2024-04-10
change: [M3-7964] - Update developer guide documentation for Adobe Analytics

mjac0bs:
## Description 📝
Based on cafe discussion, we want to be able to send a new variable to Adobe Analytics to capture more information about custom events. An example of this is a boolean like "isLinodePoweredOn": [true/false], which is not easily decipherable via our existing variables. See #10337.

After working with the Adobe Analytics backend team, we have added a new object variable, which we stringify before sending to Adobe Analytics, to allow us to capture more descriptive aspects of our custom events, as needed. (**Note**: the changes to the AA backend are testable Cloud + their dev environment right now, and were released to their prod environment on 4/10, ahead of our next release.)

**This PR handles only the documentation updates for the described changes above.**

## Changes  🔄
- Removes the Adobe Analytics section from the Coding Standards page. It never totally fit in there, anyway and adding it as its own page to the Tooling section will make it more discoverable.
- Updates the docs to include information about the new data property we can now pass in with our custom events. Makes a couple of other miscellaneous copy updates for clarity.

## Target release date 🗓️
4/15/24

## Preview 📷
Screenshot 2024-04-10 at 7 26 44 AM

### Verification steps
(How to verify changes)
- Check out this PR and `yarn docs`, then go to http://localhost:5173/manager/tooling/analytics.html.
- Or just read the diff, check for typos or any confusing statements. 

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/.changeset/pr-10365-tech-stories-1712760397845.md
+++ b/packages/manager/.changeset/pr-10365-tech-stories-1712760397845.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Tech Stories
+---
+
+Update documentation on Adobe Analytics to cover data property ([#10365](https://github.com/linode/manager/pull/10365))
--- a/docs/development-guide/13-coding-standards.md
+++ b/docs/development-guide/13-coding-standards.md
@@ -158,39 +158,3 @@ For this reason, extending types with interfaces/extends is suggested over creat
 ```
 
 Source: [TypeScript Wiki](https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections)
-
-## Adobe Analytics
-
-### Writing a Custom Event
-
-Custom events live (mostly) in `src/utilities/analytics.ts`. Try to write and export custom events in this file if possible, and import them in the component(s) where they are used.
-
-```tsx
-// Component.tsx {file(s) where the event is called, for quick reference}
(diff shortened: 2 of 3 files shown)


packages/manager/.changeset/pr-10365-tech-stories-1712760397845.md
hkhalil-akamai, line 1: I wonder if a manager changeset is appropriate since technically docs are outside of the manager package -- not sure where it should go though.

hkhalil-akamai (approved):
Agree with relocating AA docs, and the new section looks good!

packages/manager/.changeset/pr-10365-tech-stories-1712760397845.md
mjac0bs: Yeah. This is consistent with our changesets for [past docs PRs](https://github.com/linode/manager/pulls?q=is%3Apr+label%3ADocumentation+is%3Aclosed) - I don't think we have a better place. The other option is to not include a changeset at all, but it seems better to have a reference to it (even if in an imperfect place) rather than not.

bnussman-akamai (approved):

abailly-akamai (approved):
Thanks for the update, this is helpful and well put together! ✅

PR #10374, opened 2024-04-12
upcoming: [M3-7987] - Linode Create Refactor - Validation

bnussman-akamai:
## Description 📝

- Adds dynamic validation to the new Linode Create flow 🔥 
- Adds `onBlur` and `ref` to many components
  - `onBlur` allows the form to perform validation `onBlur` 
  - `ref` allows react-hook-form to focus errors aromatically onSubmit
- Expected behavior ⌨️ 
  - before submit, fields will be validated onBlur
  - after submission, fields will be validated onChange 

## Preview 📷

https://github.com/linode/manager/assets/115251059/54cdefdb-3962-4f5f-a585-c8a10e5b5d56

## How to test 🧪

- Test validation throughout the new form

- [ ] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/components/SelectRegionPanel/SelectRegionPanel.tsx
+++ b/packages/manager/src/components/SelectRegionPanel/SelectRegionPanel.tsx
@@ -26,8 +26,10 @@ import { DocsLink } from '../DocsLink/DocsLink';
 import { Link } from '../Link';
 
 import type { LinodeCreateType } from 'src/features/Linodes/LinodesCreate/types';
+import { RegionSelectProps } from '../RegionSelect/RegionSelect.types';
 
 interface SelectRegionPanelProps {
+  RegionSelectProps?: Partial<RegionSelectProps>;
   currentCapability: Capabilities;
   disabled?: boolean;
   error?: string;
@@ -49,6 +51,7 @@ export const SelectRegionPanel = (props: SelectRegionPanelProps) => {
     helperText,
     selectedId,
     selectedLinodeTypeId,
+    RegionSelectProps,
   } = props;
 
   const flags = useFlags();
@@ -147,6 +150,7 @@ export const SelectRegionPanel = (props: SelectRegionPanelProps) => {
         regions={regions ?? []}
         selectedId={selectedId || null}
         showEdgeIconHelperText={showEdgeIconHelperText}
+        {...RegionSelectProps}
       />
       {showClonePriceWarning && (
         <Notice
--- a/packages/manager/src/features/Linodes/LinodeCreatev2/Access.tsx
+++ b/packages/manager/src/features/Linodes/LinodeCreatev2/Access.tsx
@@ -32,9 +32,11 @@ export const Access = () => {
               autoComplete="off"
               disabled={isLinodeCreateRestricted}
               errorText={fieldState.error?.message}
+              inputRef={field.ref}
               label="Root Password"
               name="password"
               noMarginTop
+              onBlur={field.onBlur}
               onChange={field.onChange}
               placeholder="Enter a password."
               value={field.value ?? ''}
--- a/packages/manager/src/features/Linodes/LinodeCreatev2/VLAN.tsx
+++ b/packages/manager/src/features/Linodes/LinodeCreatev2/VLAN.tsx
@@ -88,7 +88,7 @@ export const VLAN = () => {
               filter={{ region: regionId }}
               onBlur={field.onBlur}
               onChange={field.onChange}
-              sx={{ minWidth: 300 }}
+              sx={{ width: 300 }}
               value={field.value ?? null}
             />
           )}
@@ -101,6 +101,7 @@ export const VLAN = () => {
               tooltipText={
                 'IPAM address must use IP/netmask format, e.g. 192.0.2.0/24.'
               }
+              containerProps={{ maxWidth: 335 }}
               disabled={disabled}
               errorText={fieldState.error?.message}
               inputRef={field.ref}
@@ -109,7 +110,6 @@ export const VLAN = () => {
               onChange={field.onChange}
               optional
               placeholder="192.0.2.0/24"
-              sx={{ maxWidth: 300 }}
               value={field.value ?? ''}
             />
           )}
--- a/packages/manager/src/features/Linodes/LinodeCreatev2/VLAN.tsx
+++ b/packages/manager/src/features/Linodes/LinodeCreatev2/VLAN.tsx
@@ -86,6 +86,7 @@ export const VLAN = () => {
               disabled={disabled}
               errorText={fieldState.error?.message}
               filter={{ region: regionId }}
+              onBlur={field.onBlur}
               onChange={field.onChange}
               sx={{ minWidth: 300 }}
               value={field.value ?? null}
@@ -102,7 +103,9 @@ export const VLAN = () => {
               }
               disabled={disabled}
               errorText={fieldState.error?.message}
+              inputRef={field.ref}
               label="IPAM Address"
+              onBlur={field.onBlur}
               onChange={field.onChange}
               optional
               placeholder="192.0.2.0/24"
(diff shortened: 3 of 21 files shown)


packages/manager/src/features/Linodes/LinodeCreatev2/Access.tsx
```diff
@@ -32,9 +32,11 @@ export const Access = () => {
...
               disabled={isLinodeCreateRestricted}
               errorText={fieldState.error?.message}
+              inputRef={field.ref}
               label="Root Password"
               name="password"
               noMarginTop
+              onBlur={field.onBlur}
               onChange={field.onChange}
               placeholder="Enter a password."
               value={field.value ?? ''}
```
hkhalil-akamai, line 42: As a small enhancement, I wonder if it would be possible to update the validation on this field when the selected Image changes (i.e., so the validation error clears if the user deselects the image).

packages/manager/src/features/Linodes/LinodeCreatev2/VLAN.tsx
```diff
@@ -102,7 +103,9 @@ export const VLAN = () => {
               }
               disabled={disabled}
               errorText={fieldState.error?.message}
+              inputRef={field.ref}
               label="IPAM Address"
+              onBlur={field.onBlur}
```
hkhalil-akamai, line 108: At certain window widths, this component seems to break onto the next line when a validation error is present:

<video src="https://github.com/linode/manager/assets/122488130/c47d4cf1-7fec-4c69-bcb6-8773f357d48e" width=100></video>

abailly-akamai (commented):
Super clean! I like the pattern, looks super maintainable. ✨ 

A few things I noticed (not sure if you plan to address them later)

- We also need validation for the image select (it can be reset to no value so the same validation as region should happen?)
Screenshot 2024-04-16 at 11 20 24
- We don't scroll to error to the plan selection panel, so we may never notice the selection is missing
- two problems with the video below
  - I was able to submit without a password 
  - we're not scrolling to the top most error (plan, password for instance)
<video src="https://github.com/linode/manager/assets/130582365/0158d282-e4ff-4960-91d3-230faa98bd67" />

I'll take another pass with the other panels once those have been ironed out

hkhalil-akamai (approved):
Validation looks really good and an improvement over create V1. Left some minor feedback.

packages/manager/src/features/Linodes/LinodeCreatev2/Access.tsx
bnussman-akamai: Good catch. I'll look into this edge case

bnussman-akamai:
@abailly-akamai

> I was able to submit without a password

I think this is expected behavior. You can deploy a Linode without an image, which does not require a root password.

> We don't scroll to error to the plan selection panel

Good catch. I'm still trying to figure out how to do this properly with hook form's `ref` 👀  I'm trying to avoid `scrollErrorIntoView`, but maybe we'll need it here

> We also need validation for the image select

I don't think we need anything additional because it is valid to deploy a Linode without an image. Let me know if I overlooked something about this

dwiley-akamai (commented):
Things are looking great overall! I found an error that we weren't surfacing before:

Screenshot 2024-04-16 at 2 05 04 PM

I got that error by removing the image but trying to assign the linode to a VPC

bnussman-akamai:
@dwiley-akamai  Fixed in f8e6221. 

> [!Note]
> Errors that don't have an associated text field will not automatically scroll into view yet. I am exploring ways to do this, it might be a later addition

packages/manager/src/components/SelectRegionPanel/SelectRegionPanel.tsx
```diff
@@ -26,8 +26,10 @@ import { DocsLink } from '../DocsLink/DocsLink';
 import { Link } from '../Link';
 
 import type { LinodeCreateType } from 'src/features/Linodes/LinodesCreate/types';
+import { RegionSelectProps } from '../RegionSelect/RegionSelect.types';
 
 interface SelectRegionPanelProps {
+  RegionSelectProps?: Partial<RegionSelectProps>;
```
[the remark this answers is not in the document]
bnussman-akamai: Hmmm. MUI uses the style `ChipProps`, `LabelProps`, etc... 

Should use do lowercase?

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
```diff
@@ -161,3 +165,29 @@ export const defaultValues: CreateLinodeRequest = {
   region: '',
   type: '',
 };
+
+/**
+ * Provides dynamic validation to the Linode Create form.
+ *
+ * Unfortunately, we have to wrap `yupResolver` so that we can transform the payload
+ * using `getLinodeCreatePayload` before validaton happens.
```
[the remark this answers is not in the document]
dwiley-akamai: (typo still there so I unresolved)

packages/validation/src/linodes.schema.ts
```diff
@@ -173,7 +173,9 @@ export const LinodeInterfaceSchema = object().shape({
   primary: boolean().notRequired(),
   subnet_id: number().when('purpose', {
     is: 'vpc',
-    then: number().required('Subnet is required.'),
+    then: number()
+      .transform((value) => (isNaN(value) ? undefined : value))
```
dwiley-akamai, line 177: What made the `.transform(...)` necessary here?

dwiley-akamai (approved):

packages/validation/src/linodes.schema.ts
bnussman-akamai: If you cleared the subnet select, you'd get some weird error saying "subnet_id is NaN". This fixes the weird casting and gives a better error like "Subnet is required."

abailly-akamai (approved):
> I think this is expected behavior. You can deploy a Linode without an image, which does not require a root password.

@bnussman-akamai yeah of course the password is only for the image 🤦 in the light of that validation looks good, but it makes the UX a bit weird.. (unrelated to this PR). Now it really makes me feel that these the password field should be associated in the image panel in the UI, especially cause it's super easy to miss the password field in the middle of the flow. 
I will bring that to the cafe and see if there's any opinion on the matter.

> Errors that don't have an associated text field will not automatically scroll into view yet. I am exploring ways to do this, it might be a later addition

Sounds good. 

And just to confirm, do we have logic to scroll to the top most error already?

bnussman-akamai:
@abailly-akamai 

> And just to confirm, do we have logic to scroll to the top most error already? 

Not yet. It seems like react-hook-form needs an actual `input` we can attach a ref to for it to auto-scroll. Going to spend some time investigating how that works. It's not super well documented 😣

hkhalil-akamai:
> It seems like react-hook-form needs an actual `input` we can attach a ref to for it to auto-scroll.

Wonder if the ref has to be an `input` or if it could be any HTML component, then you might be able to pass the error component ref.

packages/manager/src/components/SelectRegionPanel/SelectRegionPanel.tsx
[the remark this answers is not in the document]
hkhalil-akamai: Hmm yeah it seems like the convention is a little inconsistent (e.g., `textFieldProps`). I'm in favor of lowerCase for all props (since I associate UpperCase with components) but I don't feel too strongly about it.

bnussman-akamai:
I think I tried that, but I will try again @hkhalil-akamai

hkhalil-akamai:
A hacky way to do it might be to have an invisible input component but yeah that's definitely not ideal.

cpathipa (approved):
Thank you @bnussman-akamai confirming that error validation is working as expected for regions field. Approving since root password validation is out of this PR's scope.

PR #10375, opened 2024-04-12
upcoming: [M3-7981] - Show the selected regions as chips in the AccessKeyDrawer

cpathipa:
## Description 📝
Renders the selected regions as chips in the AccessKeyDrawer component.
The buckets table will be updated once region selection is completed.

## Preview 📷 
| Before | After |
| --- | --- |
| Screenshot 2024-04-18 at 10 20 02 AM Screenshot 2024-04-18 at 10 20 18 AM | Screenshot 2024-04-18 at 9 10 45 AM Screenshot 2024-04-18 at 9 10 50 AM |

### Prerequisites
- Turn on OBJ feature flag and MSW

### Verification steps
(How to verify changes)
- Navigate to http://localhost:3000/object-storage/access-keys
- Click on "Create Access Key"
- Select the multiple regions. 
- Verify the selected regions are rendered as chips.
- Verify the buckets table will be updated once regions selection is complete.  
- Verify there is no regression in OBJ Create / Edit Access flow. 

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset


mjac0bs (commented):
I'm excited about the chips and think this will definitely be an UI improvement. A few things I noticed:

- I think the CI e2e failure is legit. When I run `objectStorage/access-keys.smoke.spec.ts` locally, I saw the three test failures that seem related to that region select. The first OMC test fails at the `@createAccessKey` intercept, the second test fails when looking for a bucket in the list after adding a region, and the third fails because it's looking for a selection list.
- Is UX set on including the flags in the chips? I'm not a huge fan of them and think I would prefer just the `city, country (id)`. But if we do want the added visual, can we display the flags smaller and with the correct aspect ratio? They look slightly squished.
- Once some regions are selected, can we get rid of the placeholder text? We can't read all of it, since it gets truncated, and removing it would keep that form field as visually simple as possible, which I think is important since it can already contain a lot of information.

Screenshot 2024-04-15 at 3 57 02 PM

cpathipa:
@mjac0bs Thank you for the feedback. Addressed feedback PR is ready for re-review. 
- Adjusted cypress tests, 
- Reduced flag size in chips, 
- Lifted placed holder when regions are selected.

mjac0bs (approved):
Thanks for addressing feedback! This looks really nice. 🚢 

- Confirmed the flag aspect ratio in the chips looks the same as in the region select
- Confirmed the placeholder text is not shown once a user has at least one region chip selected
- Confirmed I can still create limited and unlimited access keys with the expected bucket access set
- Confirmed the "Select at least one region" validation error displays if the user attempts to submit the form without any chips, or if they have chips and remove them all
- Confirmed tests are passing now

bnussman-akamai (approved):

PR #10383, opened 2024-04-16
change: [M3-7994] - Clean up Top Menu and Fix Inconsistencies

bnussman-akamai:
## Description 📝

Cleans up and fixes many inconsistencies with the Top Menu 🧹 

### 1. Fixes bug where opening the Notification/Event Menu would scroll to top in Safari browsers 🔧 

| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/linode/manager/assets/115251059/5b1558cb-ce6e-423b-88a1-fec18ed6f73e" /> | <video src="https://github.com/linode/manager/assets/115251059/69f93066-23c5-496f-8cdd-29b4faf6032f" /> | 

### 2. Fixes inconsistent ripple effects

| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/linode/manager/assets/115251059/05adb51e-d1e2-4100-88b0-7e0b993a856b" /> | <video src="https://github.com/linode/manager/assets/115251059/a308640f-2371-4932-90d3-ddafe3aae5e3" /> |

### 3. Makes Notification/Event Menu work and look like the `UserMenu` and the `AddNewMenu` 🎨 

| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/linode/manager/assets/115251059/8e9e336d-6d9e-412f-8fc4-71900edefe76" /> | <video src="https://github.com/linode/manager/assets/115251059/3ca6fa6f-22f5-48f4-a282-fd45e752d3a9" /> | 

## Other Changes 🔁
- Makes "Help and Support" and actual `Link`, so it should be more accessible
- Makes "Community" an actual `Link`, so it should be more accessible

## How to test 🧪

- Test the Top Menu
- Verify accessibility does not get worse 

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/TopMenu/Community.tsx
+++ b/packages/manager/src/features/TopMenu/Community.tsx
@@ -10,7 +10,10 @@ import { TopMenuTooltip, topMenuIconButtonSx } from './TopMenuTooltip';
 export const Community = () => {
   return (
     <TopMenuTooltip title="Linode Cloud Community (opens in new tab)">
-      <Link to={LINODE_COMMUNITY_URL}>
+      <Link
+        accessibleAriaLabel="Linode Cloud Community"
+        to={LINODE_COMMUNITY_URL}
+      >
         <IconButton sx={topMenuIconButtonSx}>
           <CommunitySVGIcon height="20px" width="20px" />
         </IconButton>
--- a/packages/manager/src/features/TopMenu/Community.tsx
+++ b/packages/manager/src/features/TopMenu/Community.tsx
@@ -5,23 +5,13 @@ import CommunitySVGIcon from 'src/assets/icons/community_nav.svg';
 import { Link } from 'src/components/Link';
 import { LINODE_COMMUNITY_URL } from 'src/constants';
 
-import { TopMenuTooltip } from './TopMenuIcon';
+import { TopMenuTooltip, topMenuIconButtonSx } from './TopMenuTooltip';
 
 export const Community = () => {
   return (
     <TopMenuTooltip title="Linode Cloud Community (opens in new tab)">
       <Link to={LINODE_COMMUNITY_URL}>
-        <IconButton
-          sx={(theme) => ({
-            '&:hover, &:focus': {
-              color: '#606469',
-            },
-            color: '#c9c7c7',
-            [theme.breakpoints.down('sm')]: {
-              padding: 1,
-            },
-          })}
-        >
+        <IconButton sx={topMenuIconButtonSx}>
           <CommunitySVGIcon height="20px" width="20px" />
         </IconButton>
       </Link>
--- a/packages/manager/src/features/TopMenu/Community.tsx
+++ b/packages/manager/src/features/TopMenu/Community.tsx
@@ -2,7 +2,6 @@ import { IconButton } from '@mui/material';
 import * as React from 'react';
 
 import CommunitySVGIcon from 'src/assets/icons/community_nav.svg';
-import { Link } from 'src/components/Link';
 import { LINODE_COMMUNITY_URL } from 'src/constants';
 
 import { TopMenuTooltip, topMenuIconButtonSx } from './TopMenuTooltip';
@@ -10,14 +9,14 @@ import { TopMenuTooltip, topMenuIconButtonSx } from './TopMenuTooltip';
 export const Community = () => {
   return (
     <TopMenuTooltip title="Linode Cloud Community (opens in new tab)">
-      <Link
-        accessibleAriaLabel="Linode Cloud Community"
-        to={LINODE_COMMUNITY_URL}
+      <IconButton
+        aria-label="Linode Cloud Community - link opens in a new tab"
+        href={LINODE_COMMUNITY_URL}
+        sx={topMenuIconButtonSx}
+        target="_blank"
       >
-        <IconButton sx={topMenuIconButtonSx}>
-          <CommunitySVGIcon height="20px" width="20px" />
-        </IconButton>
-      </Link>
+        <CommunitySVGIcon height="20px" width="20px" />
+      </IconButton>
     </TopMenuTooltip>
   );
 };
--- a/packages/manager/src/features/TopMenu/Community.tsx
+++ b/packages/manager/src/features/TopMenu/Community.tsx
@@ -1,27 +1,27 @@
+import { IconButton } from '@mui/material';
 import * as React from 'react';
 
 import CommunitySVGIcon from 'src/assets/icons/community_nav.svg';
-import { StyledLinkButton } from 'src/components/Button/StyledLinkButton';
+import { Link } from 'src/components/Link';
 import { LINODE_COMMUNITY_URL } from 'src/constants';
 
-import { StyledTopMenuIconWrapper, TopMenuIcon } from './TopMenuIcon';
+import { TopMenuTooltip } from './TopMenuIcon';
 
 export const Community = () => {
-  const openInNewTab = (url: string) => {
-    window.open(url, '_blank', 'noopener noreferrer');
-  };
-
   return (
-    <TopMenuIcon title="Linode Cloud Community (opens in new tab)">
-      <StyledLinkButton
-        aria-label="Linode Cloud Community (opens in new tab)"
-        onClick={() => openInNewTab(LINODE_COMMUNITY_URL)}
-        role="link"
-      >
-        <StyledTopMenuIconWrapper>
-          <CommunitySVGIcon status="community" />
-        </StyledTopMenuIconWrapper>
-      </StyledLinkButton>
-    </TopMenuIcon>
+    <TopMenuTooltip title="Linode Cloud Community (opens in new tab)">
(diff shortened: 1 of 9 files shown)


packages/manager/src/features/TopMenu/NotificationMenu/NotificationMenu.tsx
```diff
@@ -111,92 +80,60 @@ export const NotificationMenu = () => {
...
-          zIndex: 3000,
         }}
         anchorEl={anchorRef.current}
-        disablePortal
+        onClose={handleClose}
         open={notificationContext.menuOpen}
-        transition
       >
-        <ClickAwayListener onClickAway={handleClose}>
-          <MenuList
```
bnussman-akamai, line 171: As far as I can tell, this Menu stuff isn't even doing anything meaningful. For it to work properly, the direct children probably need to be Menu Items.

packages/manager/src/features/TopMenu/TopMenu.tsx
```diff
@@ -64,7 +64,7 @@ export const TopMenu = React.memo((props: TopMenuProps) => {
           variant="dense"
         >
           <Hidden mdDown>
-            <TopMenuIcon key={navHoverText} title={navHoverText}>
+            <TopMenuTooltip title={navHoverText}>
```
bnussman-akamai, line 67: I don't think `key` is needed here. Please correct me if I'm wrong

packages/manager/src/features/TopMenu/TopMenuTooltip.tsx
bnussman-akamai, line 1: I just renamed `TopMenuIcon.tsx` to `TopMenuTooltiptsx` because TopMenuIcon didn't make sense

packages/manager/src/features/TopMenu/NotificationMenu/NotificationMenu.tsx
```diff
@@ -111,92 +80,60 @@ export const NotificationMenu = () => {
...
+            <StyledChip
+              color="success"
+              component="span"
+              label={numNotifications}
+              size="small"
+            />
+          )}
+        </IconButton>
+      </TopMenuTooltip>
+      <Popover
```
bnussman-akamai, line 111: Now using a `Popover` instead of a `Popper` to match the behavior of the `AddNewMenu` and the `UserMenu`. This fixes a lot of styling and adds a nice animation to match the other menus. 

| Before  |  This PR  |
|---|---|
| <video src="https://github.com/linode/manager/assets/115251059/405fba80-d8cb-4003-85cd-9cff03bbbd11" />  | <video src="https://github.com/linode/manager/assets/115251059/b3931c59-ad21-420c-8a0c-1fa568c623e2" />  |
| Notice how the Popper background does not stay when scroll bounces and how there is no open/close animation   |  Notice how the background of the Popover is constant when scroll bounces and there is a nice open/close animation |

packages/manager/src/features/TopMenu/NotificationMenu/NotificationMenu.tsx
```diff
@@ -111,92 +80,60 @@ export const NotificationMenu = () => {
...
-          sx={{
-            '&:hover': {
-              backgroundColor: 'unset',
-            },
-            margin: 0,
-            minWidth: 'unset',
-            padding: 0,
-          }}
+      <TopMenuTooltip title="Notifications">
+        <IconButton
```
bnussman-akamai, line 89: An `IconButton` is much more fitting than a `Button`

packages/manager/src/features/TopMenu/Help.tsx
```diff
@@ -1,25 +1,19 @@
...
-        onClick={() => history.push('/support')}
-        role="link"
-      >
-        <StyledTopMenuIconWrapper>
-          <HelpSVGIcon status="help" />
-        </StyledTopMenuIconWrapper>
-      </StyledLinkButton>
-    </TopMenuIcon>
+    <TopMenuTooltip title="Help & Support">
+      <Link to="/support">
```
bnussman-akamai, line 12: Using an actual `Link` instead of using JS to go to `/support`. This is probably good for accessibility and probably just better practice?

packages/manager/src/features/TopMenu/Help.tsx
```diff
@@ -1,25 +1,19 @@
...
-
   return (
-    <TopMenuIcon title="Help & Support">
-      <StyledLinkButton
-        aria-label="Help & Support"
-        onClick={() => history.push('/support')}
-        role="link"
-      >
-        <StyledTopMenuIconWrapper>
-          <HelpSVGIcon status="help" />
```
bnussman-akamai, line 20: I don't think `status="help"` does anything meaningful. Let me know if I'm mistaken

carrillo-erik: This `status="help"` prop and the other one you removed from the `Community` component `status="community"` appear to be remaining bits from previous implementations of the tooltip icons. I agree that these don't belong in the codebase any longer. Good catch.

carrillo-erik:
I ran the lighthouse tool against your PR and the develop branch. Although the tool is not perfect it does help point out some useful information. In this case Accessibility dropped from 92 to 87.  It appears to be complaining about the `IconButton` not having any text for screen readers.
### Develop
dev-branch
### This PR
pr-branch

The good news is that the workaround seems straightforward based on the recommendations. I tried `aria-label="Help"`, I suppose the same can be done with `accessibleAriaLabel`. The Accessibility score remained the same, however, the message went away.
Screenshot 2024-04-17 at 1 52 24 PM

carrillo-erik (approved):

packages/manager/src/features/TopMenu/Community.tsx
```diff
@@ -1,27 +1,23 @@
...
+        to={LINODE_COMMUNITY_URL}
       >
-        <StyledTopMenuIconWrapper>
-          <CommunitySVGIcon status="community" />
-        </StyledTopMenuIconWrapper>
-      </StyledLinkButton>
-    </TopMenuIcon>
+        <IconButton sx={topMenuIconButtonSx}>
+          <CommunitySVGIcon height="20px" width="20px" />
+        </IconButton>
```
abailly-akamai, line 19: I don't think we want a button here (button inside a link) 
It appears to be semantically invalid HTML and perhaps a bad accessibility pattern as well. 
The last iteration was semantically correct. Is that the fix for the Safari problem?

see this [link](https://stackoverflow.com/questions/6393827/can-i-nest-a-button-element-inside-an-a-using-html5) or [this](https://stackoverflow.com/questions/64522551/accessible-nested-button-inside-button) for more info about acceptable children of interactive elements

packages/manager/src/features/TopMenu/NotificationMenu/NotificationMenu.tsx
```diff
@@ -111,92 +80,60 @@ export const NotificationMenu = () => {
...
-          right: '15px',
-          [theme.breakpoints.down('sm')]: {
-            right: 0,
-            width: '100%',
+          <Bell height="20px" width="20px" />
+          {numNotifications > 0 && (
+            <StyledChip
+              color="success"
+              component="span"
+              label={numNotifications}
```
abailly-akamai, line 105: while we're at it we should make this label more descriptive, such as 

```TypeScript
${numNotifications} unread notifications
```

packages/manager/src/features/TopMenu/Community.tsx
bnussman-akamai: Should be fixed by b4f82ab

packages/manager/src/features/TopMenu/NotificationMenu/NotificationMenu.tsx
bnussman-akamai: Can you clarify? Should we add some additional attribute? `${numNotifications} unread notifications` won't fit in the chip

abailly-akamai: oh sorry, I mean the "aria-label="Notifications" of the parent

bnussman-akamai: Changed in a3a2cee004

bnussman-akamai: Actually reverted this. We can address this later!

bnussman-akamai: Changing the aria label caused a handful of e2e test failures, just want to keep the scope of this PR narrow

abailly-akamai (approved):
Thanks for addressing the fixes. looks good and consistent. This is a nice cleanup!

- Semantic HTML & accessibility ✅ 
- Behavior & no regressions ✅ 

I noticed that we don't have a `rel="noopener noreferrer"` for the cloud community link but I don't think we even need those anymore as it seems to have been set as the default behavior in most browsers. Not to mentioned, the risk of browser hijacking by our own web properties seems fairly low 😄

PR #10389, opened 2024-04-18
change: [M3-7688] - Gravatar Analytics

bnussman-akamai:
## Description 📝
- Adds analytics to Gravatars 👤 
  - Adds one event that says whether or not the current user has a Gravatar when the page loads
  - Adds one event for when the users clicks `Manage photo` on the Profile Display page (`http://localhost:3000/profile/display`)

### Prerequisites
- Run `window._satellite.setDebug(true)` in your browser console
- Refresh your page

### Verification steps

| Before  | After   |
| ------- | ------- |
| When the page loads, verify you see an event like this if you have a Gravatar | Screenshot 2024-04-18 at 3 57 07 PM |
| When the page loads, verify you see an event like this if you *don't* have a Gravatar | Screenshot 2024-04-18 at 4 01 36 PM |
| Verify you see this event when you click `Manage Photo` on the `http://localhost:3000/profile/display` page | Screenshot 2024-04-18 at 4 04 14 PM |

- [ ] 👀 Doing a self review
- [x] ➕ Adding a changeset


abailly-akamai (approved):
Very nice! confirmed events are being sent. ✅ 

i could see:
```TypeScript
{
  "action": "Load",
  "category": "Gravatar",
  "label": "Has Gravatar"
  "value": true
}
{
  "action": "Load",
  "category": "Gravatar",
  "label": "Has Gravatar"
  "value": false
}
```
for payloads, but i guess it reporting ends up the same that's prob fine

mjac0bs:
> Very nice! confirmed events are being sent. ✅
> 
> i could see:
> 
> ```ts
> {
>   "action": "Load",
>   "category": "Gravatar",
>   "label": "Has Gravatar"
>   "value": true
> }
> {
>   "action": "Load",
>   "category": "Gravatar",
>   "label": "Has Gravatar"
>   "value": false
> }
> ```
> 
> for payloads, but i guess it reporting ends up the same that's prob fine

`value` has to be a `number`, and that's got to stay consistent because the variable that defines it in Adobe Analytics is shared with and used for linode.com.

We could use a boolean here with the new `data` value, but I think `label` is totally fine in this case, since it's the main data we're capturing and reason for this event, and not an additional thing we're also adding on.

bnussman-akamai:
I'm a bit confused. Why did @abailly-akamai see `"value": true` and `"value": false` when I'm not even passing a `value`?

abailly-akamai:
@mjac0bs @bnussman-akamai is correct. I was envionning such payload as a more logical payload - thx for the clarification however!

is value typed that way by us? is it required to be typed as such by the script?

bnussman-akamai:
OHHH. My fault, I understand now 🤦 haha

So are we cool with using `label` of `Has Gravatar` or `Does not have Gravatar`?

abailly-akamai:
am cool with that but in the future i would love to use the value field if we can control its type. not a big deal

mjac0bs:
`value` is typed as a `number` by us, and I know that you technically *could* send a string there to Adobe. 
However, if we start having boolean data in a field that is intended for numbers, our filtering in dashboards are going to get weird. 

`data` is intended to represent booleans, as we did with is linode powered on/off. ([our docs](https://linode.github.io/manager/tooling/analytics.html))

Screenshot 2024-04-19 at 9 08 40 AM

mjac0bs (approved):
Confirmed the events are firing as expected. One thing I guess we'll have to take into account when looking at this data is that page refreshes for a user will send another Has Gravatar/Does not have Gravatar event. I know we can measure the number of unique visitors, and what I think we'd really want to do in Adobe is look at the unique visitors this event is firing for. 

Thank you for doing this - I'm glad we're finally collecting data!

Data for dev environment is displaying in dashboard: 
Screenshot 2024-04-19 at 8 31 20 AM

abailly-akamai:
ah i see - thx @mjac0bs!

mjac0bs:
> sure - the user info is passed

> ah i see - thx @mjac0bs!
> 
> > I think we'd really want to do in Adobe is look at the unique visitors this event is firing for.
> 
> for sure - the user info is passed along, right?

I think so, but tbh I don't know all of the variables in Adobe Analytics or the correct way to graph it in a dash. I'll follow up with Laura and/or Richa on this.

bnussman-akamai:
I am also curious about that. I feel like the data could be skewed. For example, if users **with** Gravatars load Cloud Manager more often, more "Has Gravatar" events will be captured. This data won't be representative of the actual percentages of users that have Gravatars setup because it is just based on page loads as far as I have it setup. @mjac0bs @abailly-akamai

abailly-akamai:
@bnussman-akamai @mjac0bs 

I actually want to retract my previous statement because I think we can't send any user information (event an userId) to 3rd party service without authorization. With that being said, we have a couple options:
- AA has its own system for identifying unique users via algorithm ([vid](https://experienceleague.adobe.com/en/docs/analytics/components/metrics/unique-visitors)) (usually a mix of browser settings, encoded geoloc etc, stored as a cookie on the user device) which we can employ to count unique visits
- If this proves flawed, we could attach a UUID to the payload to do that, since this would not be personal info

mjac0bs:
@abailly-akamai Yeah, I see Unique Visitors in Adobe. I don't understand how to correctly graph that against component type (custom click events) because the numbers don't look right. I did reach out to Laura to let her know we're working on this again and see if she knows the correct way to chart this.

image

bnussman-akamai:
@mjac0bs Want me to go ahead and merge or hold off?

cpathipa (approved):
Confirming `Gravatar` events are firing as expected.

mjac0bs:
@bnussman-akamai We've got two solid days of stats that make sense. I'm removing the `Wait to Merge` and I think we can get this into next release.

Chandra, you, and I tested today:
Screenshot 2024-04-23 at 1 41 07 PM

I tested yesterday on various browsers:
Screenshot 2024-04-23 at 1 41 25 PM

PR #10396, opened 2024-04-22
test: [M3-7982] - Fix hanging account switching test

jdamore-linode:
## Description 📝

This un-skips and fixes the Parent/Child account switching test that was recently disabled by #10353 due to an issue with it hanging.

The root cause of the hanging is unknown but appears to be related to the timing of Cypress's assertions when a page reload or other browser navigation is involved. I detailed my findings in the Jira ticket, but the only reliable workaround I found to resolve the issue involves detecting and waiting for the page load before asserting any HTTP requests, state of DOM elements, etc. (credit to [a commenter on this very old and loosely related thread](https://github.com/cypress-io/cypress/issues/1805#issuecomment-525482440) on Cypress's repo)

To facilitate the workaround, I added a couple new Cypress commands: `cy.trackPageVisit()` and `cy.expectNewPageVisit()`:

```typescript
// After initial call to `cy.visit()` or `cy.visitWithLogin()`:
cy.trackPageVisit().as('pageVisit');

// Perform some actions that are known to trigger a page reload or similar...
// ...

// Immediately after page reload, assert that reload has occurred:
cy.expectNewPageVisit('@pageVisit');

// Continue with the test as normal...
```

## Changes  🔄
- Unskip Proxy -> Parent account switching test
- Add `cy.trackPageVisit()` and `cy.expectNewPageVisit()` commands
- Fix hanging by waiting for page reload before waiting for API responses

## How to test 🧪
We can rely on CI for this, but to run the tests locally you can use this command:

```bash
yarn cy:run -s "cypress/e2e/core/parentChild/account-switching.spec.ts"
```

To better confirm that the hanging is really resolved, I ran the test 20 times in a row locally using zsh's `repeat`:

```bash
repeat 20 yarn cy:run -s "cypress/e2e/core/parentChild/account-switching.spec.ts"
```

### DIFF
--- a/packages/manager/.changeset/pr-10396-tests-1713821841742.md
+++ b/packages/manager/.changeset/pr-10396-tests-1713821841742.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Tests
+---
+
+Fix hanging account switching test ([#10396](https://github.com/linode/manager/pull/10396))
--- a/packages/manager/cypress/e2e/core/parentChild/account-switching.spec.ts
+++ b/packages/manager/cypress/e2e/core/parentChild/account-switching.spec.ts
@@ -179,6 +179,7 @@ describe('Parent/Child account switching', () => {
       interceptGetInvoices().as('getInvoices');
 
       cy.visitWithLogin('/account/billing');
+      cy.trackPageVisit().as('pageVisit');
       cy.wait(['@getPayments', '@getInvoices', '@getPaymentMethods']);
 
       // Confirm that "Switch Account" button is present, then click it.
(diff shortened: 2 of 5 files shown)


cpathipa (approved):
LGTM! confirming tests passed locally. 
image

mjac0bs (approved):
Thanks for the detailed investigation into this, @jdamore-linode! It seems like a strange thing for Cypress to not have addressed, but I'm glad we have a reasonable workaround. Documentation of the new commands looks great; I agree with the choice to add the page visit to every parent/child test with an account switch out of caution and because it does no harm.

The spec passed consistently for me the 20+ times I ran it.

image

Edit: we seem to have an unrelated CI test failure on `objectStorage/object-storage.e2e.spec`.

jdamore-linode:
Thanks @mjac0bs!

(Also opened M3-8027 to address the bug causing the OBJ failure, and am separately working on a branch with a bunch of test fixes, including a workaround for 8027 👍)

PR #10404, opened 2024-04-24
upcoming: [M3-7875] - Linode Create Refactor - Backups

bnussman-akamai:
## Description 📝

- Linode Create Refactor for creating from a backup 💾 

## Preview 📷

Screenshot 2024-04-25 at 1 20 15 PM
 
### Prerequisites
- Turn on the `Linode Create v2` feature flag

### Verification steps
- Check for parity with the existing Linode Create flow backups tab
- Verify you can deep link to a pre-selected backup from the Linode Details page
- Verify existing tabs still work as expected
- Verify the page feel quick and snappy (no lag when making selections)

> [!note]
> The API does not support filtering on `backups.enabled`, so for now, the table will show all Linodes rather than just Linodes with backups enabled. I will work with the API team to get this sorted!

> [!note]
> I moved the position of the warning notice. I think the change is tasteful, acceptable, and in-line with our Notice patterns. If anyone has concerns about moving it I can move it to match production.

- [ ] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Backups/BackupSelect.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const hasSelectedLinode = Boolean(linode);
+
+  const { data, isFetching } = useLinodeBackupsQuery(
+    linode?.id ?? -1,
+    hasSelectedLinode
+  );
+
+  const hasNoBackups =
+    !isFetching && data?.automatic.length === 0 && !data.snapshot.current;
```
abailly-akamai, line 32: I always wonder if there is a risk of API data used in computation being stale right after the re-render in cases like this.

packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Backups/LinodeSelectTable.tsx
```diff
@@ -0,0 +1,170 @@
...
+              {data?.data.map((linode) => (
+                <LinodeSelectTableRow
+                  key={linode.id}
+                  linode={linode}
+                  onSelect={() => handleSelect(linode)}
+                  selected={linode.id === field.value?.id}
+                />
+              ))}
+            </TableBody>
+          </Table>
```
abailly-akamai, line 144: You had previously dismissed a comment about the handling of the radio/label table column in the Stackscripts create flow and it would be nice for those to be consistent (they aren't in Prod either). we have a mix of floating left, extra cell, extra spacing etc

Am in favor of floating everything left without an extra column for the radio but whatever is consistent works for me in the end.

| Stackscript  | Backups   |
| ------- | ------- |
|  Screenshot 2024-04-26 at 09 21 13 | Screenshot 2024-04-26 at 09 21 25 |

packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Backups/BackupsWarning.tsx
```diff
@@ -0,0 +1,27 @@
...
+      >
+        <li>
+          This newly created Linode will be created with the same password and
+          SSH Keys (if any) as the original Linode.
+        </li>
+        <li>
+          This Linode will need to be manually booted after it finishes
+          provisioning.
+        </li>
+      </ul>
```
abailly-akamai, line 24: Let's please use `<List />` & `<ListItem />`

I know they also need care to be bulletted (need to make a ticket to add a prop to do so), but it will make things easier to refactor and it's still better to use SX on those than `style`

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
```diff
@@ -196,39 +221,77 @@ const defaultPublicInterface = {
...
+/**
+ * We extend the API's payload type so that we can hold some extra state
+ * in the react-hook-form form.
+ *
+ * For example, we add `linode` so we can store the currently selected Linode
+ * for the Backups and Clone tab.
+ *
+ * For any extra values added to the form, we should make sure `getLinodeCreatePayload`
+ * removes the from the payload before it is sent to the API.
+ */
```
abailly-akamai, line 233: nice, thanks for documenting

`removes the from the payload before it is sent to the API.` small typo here

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
```diff
@@ -196,39 +221,77 @@ const defaultPublicInterface = {
...
+const getDefaultImageId = (params: ParsedLinodeCreateQueryParams) => {
+  // You can't have an Image selected when deploying from a backup.
+  if (params.type === 'Backups') {
+    return null;
+  }
+
+  // Always default debian for the distributions tab.
+  if (!params.type || params.type === 'Distributions') {
+    return 'linode/debian11';
+  }
```
abailly-akamai, line 284: this is probably fine, but could be made as a constant since you are using it down on line 306

nit: I would also argue that we should check that it exists and if not select the first in array via a util. I know it's unlikely we will remove it but it's a small safety next

abailly-akamai (commented):
Looks great! Patterns are solid and consistent

Added a few comments before taking a second pass

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
bnussman-akamai: Yeah, handling default values has been hard. We can't really "check that it exists" at this point because we don't have images loaded into memory yet. Hoping to find better way to do it and clean up over time.  Will make a constant

packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Backups/LinodeSelectTable.tsx
bnussman-akamai: I created M3-8041 so we can follow up, find the best pattern, and implement it on all tables

packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Backups/BackupSelect.tsx
bnussman-akamai: Do you mean something like this component showing backups for the previously selected Linode? Or the React Query cache being out of date containing stale data?

abailly-akamai: Sorry not being very clear. It's frankly something I am always struggling a bit with (and this example is probably just fine). This type of computation in the react world has always been advocated to be done in useEffects cause we can control rendering cycles a bit more granularly. While very happy to move away from using those, I sometimes wonder if (in this case for instance), `hasNoBackups` gets recalculated properly, when needed.

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
abailly-akamai: Sounds good. I imagine we will want to set defaults via custom hooks so we can hook in the data in memory?

packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Backups/BackupSelect.tsx
bnussman-akamai: In this case, I think we're good.  `hasNoBackups` should always get derived directly from the React Query state on each and every render.

Side note: I removed the `isFetching` from the condition. It wasn't needed based on how I did the conditional rendering.

packages/manager/src/features/Linodes/LinodeCreatev2/utilities.ts
```diff
@@ -280,12 +348,14 @@ export const defaultValuesMap: Record<LinodeCreateType, CreateLinodeRequest> = {
...
-export const resolver: Resolver<CreateLinodeRequest> = async (
+export const resolver: Resolver<LinodeCreateFormValues> = async (
   values,
   context,
   options
 ) => {
-  const transformedValues = getLinodeCreatePayload(values);
+  const transformedValues = getLinodeCreatePayload(
+    values
+  ) as LinodeCreateFormValues;
```
SanPilot, line 358: Can we avoid casting by adding the `linode` value back in?

```suggestion
  const transformedValues: LinodeCreateFormValues = {
    ...getLinodeCreatePayload(values),
    linode: values.linode,
  };
```

packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Backups/LinodeSelectTable.tsx
SanPilot, line 1: Why are we no longer using a common selection table component for backups and clone?

packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Backups/LinodeSelectTable.tsx
```diff
@@ -0,0 +1,173 @@
...
+  };
+
+  return (
+    <Stack pt={1} spacing={2}>
+      <DebouncedSearchTextField
+        customValue={{
+          onChange: (value) => {
+            if (!value && preselectedLinodeId) {
+              setPreselectedLinodeId(undefined);
+            }
```
SanPilot, line 94: Bug: when a Linode is pre-selected, changing the search filter has no effect until the clear button is pressed.

SanPilot: https://github.com/linode/manager/assets/11482841/f033858e-bc1b-41cd-88fb-ab04de9b4a54

packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Backups/LinodeSelect.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import { BackupsWarning } from './BackupsWarning';
+import { LinodeSelectTable } from './LinodeSelectTable';
+
+export const LinodeSelect = () => {
+  return (
+    <Paper>
+      <Stack spacing={1}>
+        <Typography variant="h2">Select Linode</Typography>
+        <BackupsWarning />
```
SanPilot, line 15: I'm not opposed to moving the notice below the header but I think UX should be informed of the change

packages/manager/src/features/Linodes/LinodeCreatev2/index.tsx
```diff
@@ -115,11 +116,13 @@ export const LinodeCreatev2 = () => {
               <SafeTabPanel index={3}>
                 <Images />
               </SafeTabPanel>
-              <SafeTabPanel index={4}>Bckups</SafeTabPanel>
```
SanPilot, line 118: Finally, this typo has been bugging me 😂

SanPilot (approved):
Manual testing and code review both look good. Left some feedback including one possible bug.

hkhalil-akamai (approved):
Whoops, signed into the wrong account 😅

Approved pending feedback from the previous review

packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Backups/BackupsWarning.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export const BackupsWarning = () => {
+  return (
+    <Notice variant="warning">
+      <List sx={{ listStyleType: 'disc', pl: 2.5 }}>
+        <ListItem sx={{ display: 'list-item', pl: 1, py: '4px' }}>
+          This newly created Linode will be created with the same password and
+          SSH Keys (if any) as the original Linode.
+        </ListItem>
+        <ListItem sx={{ display: 'list-item', pl: 1, py: '4px' }}>
```
abailly-akamai, line 15: same here I agree using ints is always better for spacing values when possible

abailly-akamai (approved):
Approving pending addressing @hkhalil-akamai's suggestions and comments. 

Also, agree a shared Linode table for Backups/Clone would be nice if easily doable

packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Backups/BackupsWarning.tsx
```diff
@@ -0,0 +1,22 @@
...
+import React from 'react';
+
+import { List } from 'src/components/List';
+import { Notice } from 'src/components/Notice/Notice';
+
+export const BackupsWarning = () => {
+  return (
+    <Notice variant="warning">
+      <List sx={{ listStyleType: 'disc', pl: 2.5 }}>
+        <ListItem sx={{ display: 'list-item', pl: 1, py: '4px' }}>
```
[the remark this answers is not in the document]
bnussman-akamai: Yeah! I updated the value to `0.5` instead of `4px`. I believe MUI takes ints and multiples them by `theme.spacing`

packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Backups/LinodeSelectTable.tsx
bnussman-akamai: I wen't ahead and created a new table because this new one is API filtered and ordered. Also, creating a new table allows us to use `react-hook-form` hooks closer to where they are consumed (less props required).  I think these reasons warrant new code but I'm happy to reevaluate if you think I should be using more code from the old flow

PR #10405, opened 2024-04-25
test: [various] - Cypress test cleanup and flake fixes

jdamore-linode:
## Description 📝

Cleans up and fixes various tests that have been especially flaky lately.
## Changes  🔄

- Fixes a flaw with VPC subnet Linode assignment test that would cause it to randomly fail. This was because the mock subnet object contained [5 Linodes with random IDs](https://github.com/linode/manager/blob/develop/packages/manager/src/factories/subnets.ts#L27), and occasionally the mock Linode would have a matching ID, preventing it from appearing in the assignment select drop-down. See M3-8036.
- Fixes an OBJ access key test flaw where the test would fail on accounts that have many OBJ buckets*
- Works around an OBJ bug to prevent failures in Object Storage end-to-end tests. See also M3-8027.
- Refactors Linode config tests (and related Linode create utils) to be resilient to future kernel upgrades. See M3-8034.
- Cleans up power state tests, removes some redundant assertions to shed some time off the tests
- Marks various utils as deprecated to discourage their use in new development

## How to test 🧪
We can rely on CI.


packages/manager/cypress/e2e/core/linodes/linode-config.spec.ts
```diff
@@ -16,529 +10,631 @@ import {
...
+      cleanUp('linodes');
+
+      // Fetch Linode kernel data from the API.
+      // We'll use this data in the tests to confirm that config labels are rendered correctly.
+      cy.defer(fetchAllKernels(), 'Fetching Linode kernels...').then(
+        (fetchedKernels) => {
+          kernels = fetchedKernels;
+        }
+      );
+    });
```
bnussman-akamai, line 88: I'm cool with this, but are there other ways to do it?

For example, could we intercept the kernels response that Cloud Manager makes to avoid needing to defer here?

bnussman-akamai (approved):
Very nice 🔥 Thanks for getting all of these things addressed

Left one question, but it doesn't need to hold up merging

packages/manager/cypress/e2e/core/linodes/linode-config.spec.ts
jdamore-linode: Great question! Intercepting the request and looking at the response was my first approach, and it technically worked, but I didn't really like it because of these points:

- We don't get any type safety on the response object, so we have to explicitly check and guard against `undefined` values, etc.. The end result was I had all these conditionals in the middle of the tests checking the response data and `throw`ing if they didn't have the right shape, and it was just kind of ugly. We _do_ do this in other places throughout the tests, though.
  - Long term I'd like to come up with a command that can intercept requests _and_ validate the response so we don't have to do it manually every time, but I'm not 100% certain it's possible (haven't tried yet), and it doesn't address the next point...
- This is a paginated response, so it only works when we're certain that the data we want is included in the response that we're intercepting. The `kernels` endpoint responds with fewer than 500 items so, for now, `linode/latest-64bit` is guaranteed to be in the response, but that might not be the case as more and more items get added to the response (there are 328 items right now). Things start getting really tricky when we start trying to intercept multiple requests.

It is a bummer to have to fire off more API requests in the tests so definitely happy to keep discussing. I'd love a better approach and I bet we could figure one out!

bnussman-akamai: That all makes sense! Thanks for the detailed explanation 💯

mjac0bs (approved):
Thank you for all this clean up and the corresponding explanations! 🧼 

CI looks good; I read through the PR and ran the linode-config test locally to watch it. (That form has a lot of possible situations resulting in warning notices to display, and I don't think we're testing all of them, currently, so I might add to it at the end of M3-7977.)

PR #10415, opened 2024-04-26
upcoming: [M3-7983] - Use 'edge'-class plans for edge regions

hkhalil-akamai:
## Description 📝
Updates the logic in `PlansPanel` to display plans with the new 'edge' class when deploying a Linode in an edge region.

## Target release date 🗓️
ASAP (out of cycle)

## Preview 📷

|Before|After|
|------|------|
|Screenshot 2024-04-30 at 1 46 04 PM|Screenshot 2024-04-30 at 1 43 36 PM|

* *Note: The "Edge Nanode 1" plan is used in internal testing only.*

### Prerequisites
- Add the `edge_compute` and `edge_testing` customer tags.

### Verification steps
- Enter the Linode Create flow
- Select an edge region
- Verify only 'edge'-class plans are displayed

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

---


packages/manager/src/features/components/PlansPanel/PlanContainer.tsx
```diff
@@ -53,7 +53,7 @@ export interface Props {
   selectedDiskSize?: number;
   selectedId?: string;
   selectedRegionId?: Region['id'];
-  showTransfer?: boolean;
```
hkhalil-akamai, line 56: Renaming to avoid confusion since this actually affects both the transfer and network columns.

packages/manager/src/features/components/PlansPanel/PlanSelection.tsx
```diff
@@ -275,16 +275,22 @@ export const PlanSelection = (props: PlanSelectionProps) => {
...
-              {type.network_out / 1000} Gbps
+              {type.network_out ? (
+                <>
+                  {LINODE_NETWORK_IN} Gbps{' '}
+                  <span style={{ color: '#9DA4A6' }}>/</span>{' '}
+                  {type.network_out / 1000} Gbps
+                </>
+              ) : (
+                ''
+              )}
```
hkhalil-akamai, line 293: The decision to display a column should not come at the row level. This resolves an edge-case where some plans have transfer data and others don't, causing cells to fall into the wrong column.

|Before|After|
|------|------|
|Screenshot 2024-04-30 at 1 42 35 PM|Screenshot 2024-04-30 at 1 43 36 PM|

packages/manager/src/features/components/PlansPanel/utils.ts
```diff
@@ -50,7 +51,7 @@ export const getPlanSelectionsByPlanType = <
   T extends { class: LinodeTypeClass }
 >(
   types: T[]
-): PlansByType<T> => {
+): Partial<PlansByType<T>> => {
```
hkhalil-akamai, line 54: This stricter type avoids crashes in cases where a plan class is empty.

packages/manager/src/features/components/PlansPanel/PlansPanel.tsx
```diff
@@ -96,8 +96,13 @@ export const PlansPanel = (props: PlansPanelProps) => {
     getIsEdgeRegion(regionsData ?? [], selectedRegionID ?? '');
 
   const getDedicatedEdgePlanType = () => {
+    if (_plans.edge?.length) {
+      return _plans.edge;
+    }
+
+    // @TODO Remove fallback once edge plans are activated
```
hkhalil-akamai, line 103: As an alternative to using a feature flag, the new plans will be displayed if they are returned in the API response; otherwise, we will fallback to the existing logic. This reduces the chance of mis-timing.

abailly-akamai (approved):
- Code and functionality both look good ✅ 
- Confirmed the transfer column/data in both the table and the card ✅

bnussman-akamai (commented):
I'm seeing a ghost tab 👻 

https://github.com/linode/manager/assets/115251059/e0e37873-fdb4-485d-9b78-15e37a2513a7

I noticed the API is returning the plans out of order. Will we need to client side sort them or can we make them fix that?

packages/manager/src/features/components/PlansPanel/PlansPanel.tsx
```diff
@@ -96,8 +96,9 @@ export const PlansPanel = (props: PlansPanelProps) => {
     getIsEdgeRegion(regionsData ?? [], selectedRegionID ?? '');
 
   const getDedicatedEdgePlanType = () => {
-    if (_plans.edge?.length) {
-      return _plans.edge;
+    const edgePlans = types.filter((type) => type.class == 'edge');
```
bnussman-akamai, line 99: Should we use `===`?

bnussman-akamai (approved):
Good stuff! Thanks for picking up this work 🔥

hkhalil-akamai:
> I noticed the API is returning the plans out of order. Will we need to client side sort them or can we make them fix that?

I mentioned this to the API team, they will resolve it on their side.

packages/manager/src/features/components/PlansPanel/PlanSelection.tsx
hana-linode: @hkhalil-akamai A bit out of the loop. Is there a reason that the Edge Nanode 1GB has transfer but the other Edge plans don't? For the plans that don't have transfer, should we show `N/A` or something instead of blank?

packages/manager/src/features/components/PlansPanel/utils.ts
hana-linode: When would a plan class be empty?

packages/manager/src/features/components/PlansPanel/PlanSelection.tsx
hkhalil-akamai: The 'Edge Nanone 1GB' plan is an internal testing plan that won't be displayed to users. I have reached out to API and they confirmed all user facing plans will not have a network limit.

packages/manager/src/features/components/PlansPanel/utils.ts
hkhalil-akamai: If you look at [lines 74-81](https://github.com/linode/manager/pull/10415/files#diff-134803c151aeef092660488dc5ec07ce5d711cdc6571f5d3edf25812d7778ceaR74-L81), the `getPlanSelectionsByPlanType` util removes empty plan classes -- in an earlier version of this PR this was happening for edge plans before I added the customer tag causing a crash.

hana-linode (approved):

PR #10417, opened 2024-04-29
test: [M3-8050] - Fix OBJ E2E tests following API release

jdamore-linode:
## Description 📝
This fixes a couple OBJ E2E tests that are failing following the latest API release. The failures are caused by the API responding with a `400` when we pass the `region` property in OBJ requests; previously this property was ignored, but now the API handles the property by rejecting it when a certain API-side config flag is not set.

Our OBJ factories include values for both the `region` and `cluster` properties, so this fixes the issue by manually setting `region` to `undefined` in the impacted tests.

## How to test 🧪
We can rely on CI for this; in its current state, these two tests fail 100% of the time, so it should be pretty easy to determine whether or not this PR resolves the problem.

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

### DIFF
--- a/packages/manager/.changeset/pr-10417-tests-1714418018265.md
+++ b/packages/manager/.changeset/pr-10417-tests-1714418018265.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Tests
+---
+
+Fix failing OBJ E2E tests following API release ([#10417](https://github.com/linode/manager/pull/10417))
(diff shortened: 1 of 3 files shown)


bnussman-akamai (approved):

cpathipa (approved):
Thank you @jdamore-linode for fixing it.

PR #10420, opened 2024-04-30
upcoming: [M3-8012] - Fix PG event text formatting

carrillo-erik:
## Description 📝
This PR updates the types and constants for action events for Placement Groups. Specifically, the action events are updated with a present-tense and imperative-style for consistency with other entity events. These changes resolve issues with Placement Group events missing the link and username.

## Changes  🔄
- Remove the `d` and `ed` from events to make them consistent with other entity action events.
(Ex. `placement_group_created` becomes `placement_group_create` and `placement_group_unassigned` becomes `placement_group_unassign`.

## Preview 📷
| Before  | After   |
| ------- | ------- |
| before-aaa | after-bb |

### Prerequisites
- Have the `placement-group` customer tag on the user account used for testing.
- Using the Cloud Manager Dev Tools:
  - Enable the `Placement Groups` feature flag
  - Switch to `Dev` environment
  - Disable the MSW

### Verification steps
(How to verify changes)
- Verify the following events emit an event that is formatted properly (with link and username).
  - `placement_group_create`
  - `placement_group_assign`
  - `placement_group_unassign`
  - `placement_group_update`
  - `placement_group_delete`

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/features/Events/eventMessageGenerator.ts
```diff
@@ -717,18 +717,18 @@ export const eventMessageCreators: { [index: string]: CreatorsForStatus } = {
     notification: (e) =>
```
carrillo-erik, line 717: We need additional info with the distinction between the `placement_group_assign` and `placement_group_assigned` events. There have been updates since the initial work was completed for Placement Group events.

mjac0bs: I don't know the full context here, but I am seeing `placement_group_assign` events when assigning a linode to a PG. Based on the comment in the file, this should be a `placement_group_assigned` event instead. 

https://github.com/linode/manager/assets/114685994/78d85d05-8c37-476d-b8a3-3297946914e6

In the create flow, where code comments indicate `placement_group_assign` events should display, I see only the` placement_group_create` event, currently.

It sounds like there is still some clarification to get for this ticket?

mjac0bs (commented):
✅  Confirmed that the PG event verb tenses are consistent with our existing events.
✅  Confirmed that user name, the correct action for create/update/delete/unassign, and correct linkable entities show up in the events lists. 

A couple of questions about consistency with toasts, which are slightly tangential to this event PR, so I'll approve once there's clarity with the assign/assigned events:
- Can we punctuate consistently in toasts? This has been a cafe discussion before, so I'm mentioning here. I think they can be considered a sentence (subject and verb, even if it is a fragment) with a period. 

Has period (delete only):
Screenshot 2024-04-30 at 8 14 26 AM

Does not have period (everything else):
Screenshot 2024-04-30 at 8 14 58 AM
Screenshot 2024-04-30 at 8 14 36 AM
Screenshot 2024-04-30 at 8 16 25 AM
Screenshot 2024-04-30 at 8 15 16 AM


- Is it an intentional choice (for brevity, maybe) to not mention what the linode has been assigned/unassigned from in the toasts? (e.g. 'Linode successfully unassigned from  Placement Group.') This differs from the event message, which does mention a  PG. 
Screenshot 2024-04-30 at 8 16 57 AM

carrillo-erik:
After removing the `placement_group_assigned` event, the events were manually tested. The image below depicts a workflow starting from the Create Linode form.
1. A Placement Group is created
2. The Linode is created and is assigned to the newly create Placement Group.
3. A new stand alone Linode instance is created in the same region as the Placement Group from step 2.
4. From the Placement Group details page, the Linode instance from step 3 is assigned via the`PlacementGroupAssignLinodeDrawer`.
5. The event message formatting (entity label, username, and entity links) is verified in the `Events` page.

Screenshot 2024-05-01 at 7 02 18 AM

carrillo-erik:
> * Can we punctuate consistently in toasts? This has been a cafe discussion before, so I'm mentioning here. I think they can be considered a sentence (subject and verb, even if it is a fragment) with a period.
> * Is it an intentional choice (for brevity, maybe) to not mention what the linode has been assigned/unassigned from in the toasts? (e.g. 'Linode successfully unassigned from  Placement Group.') This differs from the event message, which does mention a  PG.

@mjac0bs I have updated the toast notifications to include the entity label and punctuation as suggested.

Screenshot 2024-05-01 at 8 48 18 AM
Screenshot 2024-05-01 at 8 46 01 AM
Screenshot 2024-05-01 at 8 45 52 AM

packages/manager/src/features/Events/eventMessageGenerator.ts
carrillo-erik: @mjac0bs This should be addressed with the new commits.

mjac0bs (approved):
Thanks for the updates -- confirmed that the event messages and their corresponding toasts are looking consistent. How does a placement group become compliant or non-compliant?

The CI e2e failures are unrelated, and I believe they could be fixed by merging in develop.

packages/manager/src/features/PlacementGroups/PlacementGroupsUnassignModal.tsx
```diff
@@ -56,7 +56,7 @@ export const PlacementGroupsUnassignModal = (props: Props) => {
 
   const onUnassign = async () => {
     await unassignLinodes(payload);
-    const toastMessage = 'Linode successfully unassigned';
+    const toastMessage = `Linode ${selectedLinode?.label} successfully unassigned.`;
```
abailly-akamai, line 59: For reading this code, that toast message could be "Linode undefined successfully unassigned."

I'd suggest using a condition to dispaly a safe message instead if `selectedLinode` is undefined

carrillo-erik:
> How does a placement group become compliant or non-compliant?

From the API docs:
Screenshot 2024-05-01 at 2 07 18 PM

My understanding is that we don't trigger these events and just listen to them. We have logic that prevents the assignment of Linodes that would break compliance and the user is shown a toast notification of the failure. There's actions that take place in the backend that are async, which is where these might get triggered. Maybe @abailly-akamai can chime in and correct me where I'm wrong.

abailly-akamai:
@carrillo-erik @mjac0bs sounds right to me. We just really want to populate those events. A placement group to fall out of compliance for reasons out of our control (usually due to availability, hardware failure etc). Completely async. On top of that event, the user will see a non-compliant status for the PG in Cloud Manager.

packages/manager/src/features/PlacementGroups/PlacementGroupsUnassignModal.tsx
carrillo-erik: Please check the latest commit.

abailly-akamai (approved):

PR #10422, opened 2024-04-30
test: [M3-7844] - Add script to generate JSON payload for TOD test results

jdamore-linode:
## Description 📝
This adds another script in `scripts/` to generate a payload for our internal test result store, TOD. See M3-7843 for the Cloud/TOD epic which provides more context and outlines our plan to integrate Cloud with TOD.

The purpose of this script is to consume a JUnit XML file (or a bunch of JUnit XML files) and output a TOD-compatible JSON payload. This payload contains the JUnit data itself, along with various metadata describing the project and test run.

I opted to add the script here because it will need to be invoked from Jenkins and from GitHub Actions, and wanted to avoid having to implement the same script twice.

## Changes  🔄
- Add `yarn generate-tod` script

### Pre-Requisites
You'll need at least one JUnit XML file to test with. The quickest way to generate one is to pass `--reporter=junit` when invoking `yarn test`, like this:

```bash
yarn --silent test features/Profile/APITokens/CreateAPITokenDrawer.test --reporter=junit >> junit.xml
```

Alternatively, you can generate JUnit files using Cypress:

```bash
CY_TEST_JUNIT_REPORT=1 yarn cy:run -s "cypress/e2e/core/account/account-cancellation.spec.ts"
```

The JUnit files will be output to `cypress/results`.

### Testing

Confirm that `yarn generate-tod` script works as expected. Run `yarn generate-tod --help` for a detailed breakdown of command syntax and optional parameters. Some scenarios to test:

- Generating a payload containing a single JUnit file by passing a path to the JUnit file
- Generating a payload containing multiple JUnit files by passing a path to a directory containing JUnit files
- Generating a payload containing extra metadata using `--appName`, `--appBuild`, `--appBuildUrl`, `--fail`, `--tag`, and related parameters

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "coverage": "yarn workspace linode-manager coverage",
     "coverage:summary": "yarn workspace linode-manager coverage:summary",
     "junit:summary": "ts-node scripts/junit-summary/index.ts",
+    "generate-tod": "ts-node scripts/tod-payload/index.ts",
     "docs": "bunx vitepress@1.0.0-rc.44 dev docs"
   },
   "resolutions": {
--- a/packages/manager/.changeset/pr-10422-tech-stories-1714510098465.md
+++ b/packages/manager/.changeset/pr-10422-tech-stories-1714510098465.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Tech Stories
+---
+
+Add script to generate internal test results payload ([#10422](https://github.com/linode/manager/pull/10422))
--- a/scripts/tod-payload/index.ts
+++ b/scripts/tod-payload/index.ts
@@ -17,6 +17,7 @@ program
   .option('-v, --appVersion <str>', 'Application version')
   .option('-t, --appTeam <str>', 'Application team name')
   .option('-f, --fail', 'Treat payload as failure')
+  .option('-t, --tag <str>', 'Optional tag for run')
   .option('-o, --output <str>', 'Optional path to output TOD payload file')
 
   .action((junitPath: string) => {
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -45,6 +45,7 @@
     "coverage": "yarn workspace linode-manager coverage",
     "coverage:summary": "yarn workspace linode-manager coverage:summary",
     "junit:summary": "ts-node scripts/junit-summary/index.ts",
+    "generate-tod": "ts-node scripts/tod-payload/index.ts",
```
bnussman-akamai, line 48: In the future it would be nice to run these helper scripts with bun so we don't have to depend on `ts-node`

I guess we'd have to update our Jenkins pipelines and Github Actions to install bun first

bnussman-akamai (approved):

package.json
jdamore-linode: I doubt that'd be a big lift! I'll follow up with a ticket soon

carrillo-erik (approved):

jdamore-linode:
Thanks @carrillo-erik, and sorry for the slow response!

> Tested the commands and was able to successfully execute them. There was a warning message, I'll paste here for your attention:
> 
> ```
> WARNING: Secure coding is not enabled for restorable state! Enable secure coding by implementing NSApplicationDelegate.applicationSupportsSecureRestorableState: and returning YES.
> ```

This one's new to me! I haven't been able to reproduce it, but `NSApplicationDelegate` is a Cocoa API and judging from some Google results, it might be related to executing code from within another program (VSCode?). All the results I've found seem to be related to Python on MacOS 14, however, so kind of a shot in the dark. Either way, if everything is working I'd wager this isn't an issue for us.

> Also, would this be an issue for maintainability in the future?
> 
> ```
> You are running Node v20.12.2. Only the following versions of Node are supported:
>   - v16.x
>   - v18.x
> ```

The usefulness of these warnings is debatable, but I think the intent is to let outside contributors know if they're running Cloud/the tests on a version of Node that isn't compatible with our codebase. The "supported" version is Node 18.x just because that's the version of Node we use in our test and deploy pipelines, but I'd be extremely surprised if there are compatibility issues with modern versions of Node.

I was on the fence about getting rid of this warning when we upgraded from Node 14 -- curious if you have any opinion on that? Seems like it might cause more confusion than anything else

PR #10448, opened 2024-05-08
test: [M3-7955] - Cypress integration test to add SSH key via Linode Create

cliu-akamai:
## Description 📝
Add integration test to add SSH key via Linode create.

## Major Changes 🔄
- Check users can add SSH key during Linode creation

## How to test 🧪
```
```


packages/manager/cypress/e2e/core/linodes/create-linode.spec.ts
```diff
@@ -76,6 +76,7 @@ authenticate();
 describe('create linode', () => {
   before(() => {
     cleanUp('linodes');
+    cleanUp('ssh-keys');
```
jdamore-linode, line 79:
```suggestion
```
Very minor nit: This call to `cleanUp('ssh-keys')` isn't really necessary since your test mocks the SSH key (no real SSH key is being created, so nothing will need to be cleaned up). However, I really appreciate you setting up the `deleteAllTestSSHKeys` util and hooking it up to the `cleanUp` function since this will be needed for M3-7956!

packages/manager/cypress/e2e/core/linodes/create-linode.spec.ts
```diff
@@ -511,4 +512,177 @@ describe('create linode', () => {
...
+      uppercase: true,
+      lowercase: true,
+      numbers: true,
+      spaces: false,
+      symbols: false,
+    });
+    const sshPublicKey = `ssh-rsa e2etestkey${randomKey} e2etest@linode`;
+    const linodeLabel = randomLabel();
+    const region: Region = getRegionById('us-southeast');
+    const diskLabel: string = 'Debian 10 Disk';
```
jdamore-linode, line 529: I think a lot of this mock data is held over from other tests, and we can get rid of them for this test just to make things a little cleaner! Locally, I was able to get rid of all the mock data and setup for disks, VLANs, VPCs & subnets, configs, and the mocks for the Linode DC-specific pricing types and your test still ran really nicely and passed!

Not a super big deal since your test is still testing all of the important things we need to cover for SSH keys 👍

jdamore-linode (approved):
Nice work @cliu-akamai! Left some comments for potential clean up, but nothing that impacts the quality or effectiveness of your test (which is great, by the way!)

You have my approval pending `yarn changeset`. Thanks Cassie!

dwiley-akamai (approved):
`create-linode.spec.ts` passes ✅ (it didn't for me locally due to a timeout, but it passes remotely)

Code review ✅

Approved pending the addition of a changeset & resolution of the merge conflict

AzureLatte (approved):
Test passed

hkhalil-akamai (approved):
Thanks Cassie!

PR #10462, opened 2024-05-13
upcoming: [M3-8074] – Add "Disk Encryption" section to Linode Create flow

dwiley-akamai:
## Description 📝
Add "Disk Encryption" section to the Linode Create flow to allow user to choose if their VM will be encrypted or not.

## Changes  🔄
- `Disk Encryption` added as `AccountCapability` and region `Capabilities`
- New props added to `AccessPanel` component
- Conditional notice for Backups added in `AddonsPanel` component
- `useIsDiskEncryptionFeatureEnabled` hook added

## Target release date 🗓️
5/28/24

## Preview 📷
| Before (LDE not supported)  | After (LDE supported) |
| ------- | ------- |
| Screenshot 2024-05-16 at 5 42 04 PM | Screenshot 2024-05-16 at 5 42 14 PM |
| Screenshot 2024-05-16 at 5 44 11 PM | Screenshot 2024-05-16 at 5 45 04 PM |
| Screenshot 2024-05-16 at 5 46 16 PM | Screenshot 2024-05-16 at 5 46 42 PM |

### Prerequisites
1. Have the `linode_disk_encryption` tag on your account
2. Point at the alpha/dev environment
3. Have the `Linode Disk Encryption (LDE)` flag in our dev tool toggled on

### Verification steps
Without the customer tag and/or with the LDE flag in our dev tool toggled off, you should not see any LDE-related things in the UI (or in the Linode Create payload). Otherwise,

1. "Disk Encryption" should be below "SSH Keys" in the same paper, and that paper should be labeled "Security"
 a. No region selected or region w/o Disk Encryption support selected --> "Encrypt Disk" checkbox should be unchecked and disabled with an explanatory tooltip; region w/ Disk Encryption support selected (Newark has support in alpha) --> checkbox should be enabled, and default to checked
 b.   Toggling the "Encrypt Disk" checkbox should work as expected
 c. The "Summary" section at the bottom should reflect your choice to encrypt or not, i.e., if "Encrypt Disk" is checked, you should see "Encrypted" in the summary
 d. "Encrypt Disk" checked & Backups add-on checked --> observe: notice under Backups stating, "Virtual Machine Backups are not encrypted."
3. `disk_encryption` shows with the expected value for cURL and Linode CLI when you click "Create Using Command Line" (and should be absent when the selected region doesn't support LDE)
4. No adverse or unexpected impacts where `<AccessPanel />` is used elsewhere (e.g., UDF fields, Rebuild from Image & Stackscript, Create Disk drawer --> Create from Image, etc.)

- [X] 👀 Doing a self review
- [X] ➕ Adding a changeset


bnussman-akamai:
I'm not sure how I feel about putting disk encryption in the Access Panel on the Linode create flow. Disk encryption feels unrelated to configuring credentials to me.

I think the Addons section would be a lot more fitting. (and maybe rename addons to "Options" or "Additional Configuration" if UX is concerned about the "Addons" naming)

dwiley-akamai:
> I'm not sure how I feel about putting disk encryption in the Access Panel on the Linode create flow. Disk encryption feels unrelated to configuring credentials to me.
> 
> I think the Addons section would be a lot more fitting. (and maybe rename addons to "Options" or "Additional Configuration" if UX is concerned about the "Addons" naming)

I surfaced this feedback to UX and Product, and the consensus was to roll forward with the current UX given encryption's centrality to the VM creation workflow

bnussman-akamai:
Got it! Thanks for surfacing 👍🙏🏼

packages/manager/cypress/e2e/core/linodes/create-linode.spec.ts
```diff
@@ -511,4 +517,77 @@ describe('create linode', () => {
...
+      id: 'us-east',
+      label: 'Newark, NJ',
+    });
+
+    ui.regionSelect.find().click();
+    ui.regionSelect
+      .findItemByRegionLabel(regionSupportingDiskEncryption.label)
+      .click();
+
+    cy.get(`[data-testid="${checkboxTestId}"]`).should('be.enabled');
```
dwiley-akamai, line 591: @jdamore-linode @cliu-akamai I believe it is this line specifically that's causing the failure 🤔

jdamore-linode: @dwiley-akamai Just pushed a fix! It was basically just tripping up because the regions request wasn't getting mocked to include the `Disk Encryption` capability for the part of the test that asserts that the checkbox is enabled

dwiley-akamai: Ah I missed mocking the regions, that makes sense. Thanks for taking a look and for the fix @jdamore-linode!

jaalah-akamai (approved):
✅ Encryption works based on regions
✅ Encryption shows in summary
✅ Encryption enabled shows in payload and request

Screenshot 2024-05-21 at 8 23 01 AM

carrillo-erik:
I wonder if it would be worth it to rename the `AccessPanel` component to `SecurityPanel` or maybe change the header of the `AccessPanel` to reflect the name of the component. It appears this is the only panel whose header does not match the component name in the code.

Aside from comment above, I was able to verify all functionality described in the "Verification Steps" to work as intended. LGTM ✅

carrillo-erik (approved):

dwiley-akamai:
> I wonder if it would be worth it to rename the `AccessPanel` component to `SecurityPanel` or maybe change the header of the `AccessPanel` to reflect the name of the component. It appears this is the only panel whose header does not match the component name in the code.

That file probably won't be around more than a couple of months given the Linode Create v2 effort so I'd say it's not worth it, but I will keep a possible renaming in mind for the equivalent v2 file (`src/features/Linodes/LinodeCreatev2/Access.tsx`)

PR #10466, opened 2024-05-14
feat: [M3-8016] - Add TagSelect to edit images drawer

hkhalil-akamai:
## Description 📝
Adds a TagSelect to the edit image drawer based on the new field in the API.

## Changes  🔄
- Add TagSelect to edit image drawer
- Update APIv4 Image type and factory with new `tags` field
- Clean up `ImageDrawer` component, removing unused code

## Target release date 🗓️
5/28

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-05-14 at 1 52 31 PM | Screenshot 2024-05-14 at 1 52 19 PM |

### Prerequisites
- Configure Cloud Manager to use the alpha API (e.g., via DevTools)

### Verification steps
- Navigate to the Images page
- Click on the Action menu for an image and select 'Edit'
- Verify tag select appears
- Add/remove tags and click "Save Changes"
- Re-open edit drawer and verify changes are reflected
- Verify that there are no regressions in the Restore from Image drawer, which reuses this component

## TODO
- [x] Create unit tests for ImageDrawer

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/components/TagsInput/TagsInput.tsx
```diff
@@ -42,15 +42,15 @@ export interface TagsInputProps {
   /**
    * Callback fired when the value changes.
    */
-  onChange: (selected: Item[]) => void;
```
hkhalil-akamai, line 45: By default, this type is `string | number`

packages/manager/src/features/Images/ImagesDrawer.tsx
hkhalil-akamai, line 1: The 'create' and 'imagize' modes of this drawer are no longer used, allowing for significant clean-up.

bnussman-akamai (commented):
Tags are working well ✅ 

- Found a few things 🐛 
  - The `Cancel` button does not work for me currently
  - The loading state persists when the drawer is reopened
  
 

https://github.com/linode/manager/assets/115251059/3509a6eb-47d4-4231-a26a-f5f14a5ed41f

<hr />

We decided as a team that going forward that we'd like to keep drawers purpose very concise. So in this case, we could split up `ImageDrawer` into `RestoreImageDrawer` and `EditImageDrawer`. We **don't** need to do this now, but you can if you'd like to!  It could also give us a chance to rewrite these forms to use `react-hook-form`, but again, not required

hkhalil-akamai:
>   * The `Cancel` button does not work for me currently
>   * The loading state persists when the drawer is reopened

Good catch, fixed! ✅

hkhalil-akamai:
> We decided as a team that going forward that we'd like to keep drawers purpose very concise. So in this case, we could split up `ImageDrawer` into `RestoreImageDrawer` and `EditImageDrawer`. We **don't** need to do this now, but you can if you'd like to! It could also give us a chance to rewrite these forms to use `react-hook-form`, but again, not required

There is definitely room for cleanup in this component! Making this PR as WIP while I address this.

bnussman-akamai (approved):
Tagging looks good! 🏷️ 🎉 

I did notice this behavior shown in the video. Clearing a field and saving results in no change to the Image. I'd expect a validation error if I clear the label field. We can address this later

https://github.com/linode/manager/assets/115251059/25580481-c277-456a-946d-f83cce006f06

abailly-akamai (approved):
Great PR and cleanup!

left a couple code comments.

Also, to add to either this PR or the follow up ticket you created, when encountering an API error (ex: block PUT | https://api.dev.linode.com/v4/images/private requests), the error will persist when reopening the drawer.

Screen Shot 2024-05-21 at 15 58 54

Lastly, it's pretty strange to be able to add to add/edit tags in create/edit flow yet they are not visible anywhere in the UI when looking at the image list (prolly a follow up with UX would be good there)

Screen Shot 2024-05-21 at 16 01 22

PR #10471, opened 2024-05-15
feat: [M3-8103] - Refactor Image Create and Add Tags

bnussman-akamai:
## Description 📝

Refactors the Image Create page and adds the ability to add tags during creation. This PR does not touch the upload page. **These changes will not be feature flagged**.

## Changes  🔄
- Completely refactored the Image Create page to use `react-hook-form` ✏️ 
- Added tags input to the Image Create flow 🏷️ 
- Implements new UI and copy according the Figma mockup 🎨 
- Improves cypress tests for Image creation 🧪 
  - I added one spec that is 100% mocked data and another spect that is 100% real API requests (cc @jdamore-linode)

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-05-14 at 7 52 30 PM | Screenshot 2024-05-14 at 7 52 42 PM |

## How to test 🧪

- Test creating an Image 💾 
- Using the "Network" tab of dev tools, verify the payload looks correct (includes tags) 🌏 
- Test validation on the Image Create form ☑️ 
- Test the Image Create page as a restricted user 👤 
- Verify there are no breaking changes ⛓️‍💥

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
bnussman-akamai, line 1: The primary changes are in here 🚨

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
```diff
@@ -1,355 +1,330 @@
...
+              getOptionDisabled={
+                grants
+                  ? (linode) =>
+                      grants.linode.some(
+                        (grant) =>
+                          grant.id === linode.id &&
+                          grant.permissions === 'read_only'
+                      )
+                  : undefined
+              }
```
hkhalil-akamai, line 140: Instead of disabling these options would it make sense to filter them out via `optionsFilter`?

hkhalil-akamai (approved):
The new create flow looks good and the code is very readable and clean 🧼.

✅ Image create page matches mockups
✅ Confirmed I can create an Image from a Linode and Disk and select Label, Tags and Description
✅ CI passing

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
bnussman-akamai: That was the existing behavior. We can do that, but in my opinion, the better UX would be to show the items and explain why they are disabled. That's why I added the helper text of `You can only create Images from Linodes you have read/write access to.` for this case. Happy to revert if other think this is worse UX

I have a cafe item today that will touch on this

PR #10512, opened 2024-05-23
feat: [M3-7919] Add Fonoa tax id event notification on invalidation

jaalah-akamai:
## Description 📝
To enhance our tax ID validation process, we've updated Cloud Manager to offer clearer guidance on these UX changes.

## Changes  🔄
- Non-US countries will display a helper text below the tax id input field
   - When this value changes, we display a toast alerting the user that their tax id is being validated
- A new event `tax_id_invalid` is returned and emitted if the tax id is invalid.

## Target release date 🗓️
10/10

## Preview 📷
| Desc  | Screenshot   |
| ------- | ------- |
| Non-US Helper Text | Screenshot 2024-05-23 at 12 35 29 PM |
| Event Notification / Event Page | Screenshot 2024-05-23 at 12 12 36 PM Screenshot 2024-05-23 at 12 12 14 PM |
| Non-US Helper Text | Screenshot 2024-05-24 at 11 06 19 AM |

### Prerequisites
- If you want to test the events part of this, you'll need to setup DevEnv (see me about details)

### Reproduction steps
- Pull down branch
- Turn on feature flag for `TaxId`
- Edit tax id field: http://localhost:3000/account/billing/edit 

### Verification steps
- Observe no helper text is visible and toast doesn't appear for US locations
- For non-US countries, observe both helper text and toast when field if changed

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/features/Events/constants.ts
```diff
@@ -96,6 +96,7 @@ export const EVENT_ACTIONS: Event['action'][] = [
   'subnet_create',
   'subnet_delete',
   'subnet_update',
+  'tax_id_invalid',
```
jaalah-akamai, line 99: This is the only new action, all the others needed alphabetization

packages/manager/src/features/Events/eventMessageGenerator.ts
```diff
@@ -861,6 +861,9 @@ export const eventMessageCreators: { [index: string]: CreatorsForStatus } = {
   tag_delete: {
     notification: (e) => `Tag ${e.entity!.label} has been deleted.`,
   },
+  tax_id_invalid: {
+    notification: (e) => `Tax Identification Number format is invalid.`,
+  },
```
jaalah-akamai, line 866: This is the only new notification, all the others were just formatting done by eslint

packages/manager/src/features/Billing/BillingPanels/ContactInfoPanel/UpdateContactInformationForm/UpdateContactInformationForm.tsx
```diff
@@ -63,6 +70,25 @@ const UpdateContactInformationForm = ({ focusEmail, onClose }: Props) => {
...
+            hideIconVariant: false,
+            style: {
+              display: 'flex',
+              flexWrap: 'nowrap',
+              width: '372px',
+            },
+            variant: 'info',
+          }
+        );
+      }
```
bnussman-akamai, line 90: Would it make sense to put this in the React Query `onSuccess` function of the `useMutateAccount` so it would apply globally if the `tax_id` was ever changed elsewhere?

packages/manager/src/features/Events/EventsLanding.tsx
```diff
@@ -46,6 +48,10 @@ export const EventsLanding = (props: Props) => {
     isLoading,
   } = useEventsInfiniteQuery(filter);
 
+  const filteredEvents = isTaxIdEnabled
+    ? events?.filter((event) => event.action !== 'tax_id_invalid')
+    : events;
```
bnussman-akamai, line 53: This could probably be x-filtered like `profile_update`, but this is probably fine.

jaalah-akamai:
@bnussman-akamai thanks for review, see: [0d23300](https://github.com/linode/manager/pull/10512/commits/0d23300a4acb54bea2251c06037be36191b945b3)

mjac0bs (commented):
I tested everything but the events/notifications. Generally, things looked good but I noticed a couple of edge cases:

- Do we expect the US Virgin Islands and US Minor outlying islands have tax IDs?
Screenshot 2024-06-03 at 8 12 14 AM

Screenshot 2024-06-03 at 8 13 09 AM


- If a user lives in a country with a tax ID, and then opens the drawer to update their country to the `United States of America`, they will still see the Tax ID text field and, when submitting the form, also see the Tax ID show up in the saved Billing Contact section of the Account landing page. They will not see a toast upon form submission, though, as we'd expect.

Screenshot 2024-06-03 at 8 16 10 AM


Screenshot 2024-06-03 at 8 19 02 AM

And one UX question: did UX opt to place the helper text outside of a tooltip for visibility, even though it is long?

bnussman-akamai:
General question: I see we have an event for an invalid tax ID but nothing else. I'm curious as to why we didn't go for an account notification or a new endpoint/field.  There is no way for us to know on the frontend if the event is still relevant/true

jaalah-akamai:
@mjac0bs 
> Do we expect the US Virgin Islands and US Minor outlying islands have tax IDs?
This is a good question proposed it to the team for an answer.
> If a user lives in a country with a Tax ID and sets + saves that value, and then reopens the drawer to update their country to the United States of America, they will still see the Tax ID text field and, when submitting the form, also see the Tax ID show up in the saved Billing Contact section of the Account landing page. They will not see a toast upon form submission, though, as we'd expect.
I believe this is expected if I'm understanding what you're saying correctly. If the country is the US, they will not see a toast indicating their tax id is being validated even if it's included in the POST.
> Did UX opt to place the helper text outside of a tooltip for visibility, even though it is long?
Yea due to the fact that invalid or no tax ids (when required) will be caught and rejected at the time of initiating any B2C sales

jaalah-akamai:
@bnussman-akamai 
> I'm curious as to why we didn't go for an account notification or a new endpoint/field. 
Spoke with API and we're going to improve this (transition to using `/notifications`) in a separate PR. We will also be able to add the additional UX invalid tax warning to the "Billing Contact" card. 👍 Great feedback

packages/manager/src/features/Account/utils.ts
```diff
@@ -63,3 +65,23 @@ export const getRestrictedResourceText = ({
 
   return message;
 };
+
+/**
+ * Hook to determine if the Tax Id feature should be visible to the user.
+ * Based on the user's account capability and the feature flag.
+ *
+ * @returns {boolean} - Whether the Placement Group feature is enabled for the current user.
```
mjac0bs, line 73:
```suggestion
 * @returns {boolean} - Whether the Tax Id feature is enabled for the current user.
```

mjac0bs (commented):
Thanks for clarifying and checking on the US islands.

> > If a user lives in a country with a Tax ID and sets + saves that value, and then reopens the drawer to update their country to the United States of America, they will still see the Tax ID text field and, when submitting the form, also see the
> > Tax ID show up in the saved Billing Contact section of the Account landing page. They will not see a toast upon form submission, though, as we'd expect.
> 
> I believe this is expected if I'm understanding what you're saying correctly. If the country is the US, they will not see a toast indicating their tax id is being validated even if it's included in the POST.

If this is the case, then the UI can show another country's Tax ID beside a US address, which seems strange. I think a user would expect that, if they were to select a new country (or at least a country without Tax ID - i.e. US and maybe US islands) then the Tax ID text field would get cleared.

packages/manager/src/features/Events/EventsLanding.tsx
```diff
@@ -29,8 +30,16 @@ interface Props {
...
+  const { isTaxIdEnabled } = useIsTaxIdEnabled();
 
-  const filter: Filter = { action: { '+neq': 'profile_update' } };
+  const filter: Filter = {
+    action: {
+      '+and': [
+        { '+neq': 'profile_update' },
+        isTaxIdEnabled ? { '+neq': 'tax_id_invalid' } : {},
+      ],
+    },
```
bnussman-akamai, line 41: This filter isn't valid

Screenshot 2024-06-05 at 11 05 52 AM

This should do what you are intending:
```suggestion
'+and': [
      { action: { '+neq': 'profile_update' } },
      ...(isTaxIdEnabled ? [{ action: { '+neq': 'tax_id_invalid' } }] : []),
    ],
```

I'm fine with this change, but do feel like it might be an over-flag-ization. I feel like the API should be responsible for hiding/showing **and/or** I wonder if we even need to hide it because it will only start showing up for a customer once it has been enabled on the API-side

Also, this goes back to what @abailly-akamai and I were talking about [here](https://linode.slack.com/archives/GUN7PKZR6/p1717519268385269?thread_ts=1717517920.519189&cid=GUN7PKZR6). Ideally, this API filter should eventually be shared with the Event Menu, but no action is needed now. @abailly-akamai or I will address that later

packages/manager/src/queries/account/account.ts
```diff
@@ -36,10 +38,40 @@ export const useAccount = () => {
 
 export const useMutateAccount = () => {
   const queryClient = useQueryClient();
+  const { enqueueSnackbar } = useSnackbar();
+  const { isTaxIdEnabled } = useIsTaxIdEnabled();
 
   return useMutation<Account, APIError[], Partial<Account>>(updateAccountInfo, {
     onSuccess(account) {
-      queryClient.setQueryData(accountQueries.account.queryKey, account);
+      queryClient.setQueryData<Account | undefined>(
```
bnussman-akamai, line 46: Nice 🔥

packages/manager/src/features/Events/EventsLanding.tsx
abailly-akamai: Correct - this is about to be handled and we will be able to configure the filter so that it's sharable for EventsLanding and the NotificationCenter. I would suggest removing this change and add a TODO

packages/manager/src/features/Events/eventMessageGenerator.ts
abailly-akamai: You will need to rebase your PR cause there's been a whole lot of changes to this and this new event will need to be added to event messages V2. pls reach out if needs clarification

packages/manager/src/features/Events/EventsLanding.tsx
jaalah-akamai: Sounds good - I'll revert if that's the case

packages/api-v4/src/account/types.ts
```diff
@@ -432,6 +432,7 @@ export const EventActionKeys = [
   'subnet_update',
   'tag_create',
   'tag_delete',
+  'tax_id_invalid',
```
jaalah-akamai, line 435: This is the only real change here

jaalah-akamai:
@mjac0bs I should have fixed the issue you pointed out

mjac0bs (approved):
`factory.test.tsx` is failing because we no longer have a defined event message for the new event key `tax_id_invalid`. This looks like it will need a new factory for tax, as described [here](https://linode.github.io/manager/development-guide/15-api-events.html#adding-a-new-action-and-composing-messages) with "Tax Identification Number format is invalid." added as a message for this action.

image


Approving pending that is fixed and tests pass, as well as we get confirmation of whether the US Virgin and Minor Outlying Islands are not being validated for tax ids.
The clearing of the Tax ID field upon selection of another tax id is working as expected now.

jaalah-akamai:
I haven't received any information about US territories, so we'll proceed without that for now

jaalah-akamai:
> This looks like it will need a new factory for tax, as described [here](https://linode.github.io/manager/development-guide/15-api-events.html#adding-a-new-action-and-composing-messages) with "Tax Identification Number format is invalid." added as a message for this action.

Thanks, I like the new way events are typed now 🎉

packages/manager/src/dev-tools/FeatureFlagTool.tsx
```diff
@@ -28,6 +28,7 @@ const options: { flag: keyof Flags; label: string }[] = [
   { flag: 'linodeDiskEncryption', label: 'Linode Disk Encryption (LDE)' },
   { flag: 'objMultiCluster', label: 'OBJ Multi-Cluster' },
   { flag: 'placementGroups', label: 'Placement Groups' },
+  { flag: 'taxId', label: 'Tax Id' },
```
abailly-akamai, line 31: nit: not sure we need an entry in dev tools

packages/manager/src/features/Events/factories/tax.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+
+import type { PartialEventMap } from '../types';
+
+export const tax: PartialEventMap<'tax'> = {
+  tax_id_invalid: {
+    notification: () => <>Tax Identification Number format is invalid.</>,
```
abailly-akamai, line 7:
```suggestion
    notification: () => <>Tax Identification Number format is <strong>invalid</strong>.</>,
```

abailly-akamai (approved):
Approving pending a couple nits

abailly-akamai:
Private User Image

fortunately new event messages will read better than this with the username separated from the message, cause this reads weird

bnussman-akamai (approved):

PR #10521, opened 2024-05-28
upcoming: [M3-8139, M3-8140, M3-8141] – Add warning notices re: non-encryption when creating Images & enabling Backups

dwiley-akamai:
## Description 📝
Add notices to the Imagize dialog, "Enable Backups" dialog, and the Create Image --> "Capture Image" tab advising users that Images and Backups are not encrypted.

## Target release date 🗓️
6/10

## Preview 📷
<details>
<summary>Screenshots</summary>

Screenshot 2024-05-30 at 11 22 36 AM

Screenshot 2024-05-30 at 11 30 22 AM

Screenshot 2024-05-30 at 11 31 18 AM
</details>

### Prerequisites
1. Use an account that doesn't have Managed enabled (since you'll need to check the "Enable Backups" flow)
2. Have the `linode_disk_encryption` tag on your account
3. Point at the alpha/dev environment
4. Have the `Linode Disk Encryption (LDE)` flag in our dev tool toggled on

### Verification steps
Without the tag and/or with the LDE flag in our dev tool toggled off, you should not see any LDE-related things in the UI. Otherwise,

1. Go to the Backups tab on a Linode Details page and click "Enable Backups". Observe: a notice that says "Virtual Machine Backups are not encrypted."
2. Go to the Storage tab on a Linode Details page and select a disk to imagize. Observe in the dialog a notice that says: "Virtual Machine Images are not encrypted."
3. Go to create an Image and on the "Capture Image" tab select a linode in a core region. Observe a notice that says: "Virtual Machine Images are not encrypted." (if you select a linode in an Edge region, you should not see the notice)

The last item is a bit difficult to test at the moment outside of DevEnv, but you can turn the MSW on and first select `metadata-test-image` to see the notice and then switch to the next two (`metadata-test-region` and `Gecko Edge Test`) to see that it doesn't appear. Nonetheless, the `linodeIsInEdgeRegion` logic in `CreateImageTab.tsx` should ensure that the notice is not displayed if the selected linode is in an Edge region.

- [X] 👀 Doing a self review
- [X] ➕ Adding a changeset


packages/manager/cypress/e2e/core/images/create-image.spec.ts
```diff
@@ -84,4 +120,136 @@ describe('create image (e2e)', () => {
...
+    // Select the Linode
+    ui.autocompletePopper
+      .findByTitle(mockLinodes[1].label)
+      .should('be.visible')
+      .should('be.enabled')
+      .click();
+
+    // Check if notice is visible
+    cy.get(`[data-testid=${encryptionCaveatNoticeTestId}]`).should('not.exist');
+  });
```
dwiley-akamai, line 254: @jdamore-linode This assertion is failing, any suggested tweaks to get it passing? The intent is to mock regions, one of which is an edge site, and the mocked linode that is selected has that edge site as its region. Would that be enough for `getIsEdgeRegion` in `CreateImageTab.tsx` to return the expected `true`?

packages/manager/cypress/e2e/core/images/create-image.spec.ts
```diff
@@ -84,4 +120,136 @@ describe('create image (e2e)', () => {
...
+    mockGetFeatureFlagClientstream().as('getClientStream');
+
+    // Mock responses
+    const mockAccount = accountFactory.build({
+      capabilities: ['Linodes', 'Disk Encryption'],
+    });
+
+    mockGetAccount(mockAccount).as('getAccount');
+    mockGetRegions(mockRegions).as('getRegions');
+    mockGetLinodes(mockLinodes).as('getLinodes');
```
jdamore-linode, line 226: @dwiley-akamai It looks like Cloud fires off a request to get the details for the individual Linode that was selected, and uses the region ID from that response when it calls `getIsEdgeRegion`.

Importing the `mockGetLinodeDetails` util and calling that with the data for `mockLinodes[1]` gets the test passing!

```suggestion
    mockGetLinodes(mockLinodes).as('getLinodes');
    mockGetLinodeDetails(mockLinodes[1].id, mockLinodes[1]);
```

packages/manager/cypress/e2e/core/images/create-image.spec.ts
```diff
@@ -84,4 +120,136 @@ describe('create image (e2e)', () => {
...
+
+    // Select the Linode
+    ui.autocompletePopper
+      .findByTitle(mockLinodes[0].label)
+      .should('be.visible')
+      .should('be.enabled')
+      .click();
+
+    // Check if notice is visible
+    cy.get(`[data-testid=${encryptionCaveatNoticeTestId}]`).should('exist');
```
jdamore-linode, line 165:
```suggestion
    cy.get(`[data-testid=${encryptionCaveatNoticeTestId}]`).should('be.visible');
```
Just a small suggestion, since this will also catch cases where the element is hidden (e.g. by a CSS rule) or obscured by another element

jdamore-linode, line 165: In a hypothetical scenario where this warning notice fails to appear or the language inadvertently gets changed, is there any risk a customer could create an Image and expect it to be encrypted and suffer some kind of incident as a result?

If we're not certain the answer is no, I might suggest searching for the actual contents of the notice (i.e. via `cy.findByText('...')`), and I might additionally suggest that we define a constant in this spec containing that language (not importing the const from `src`) so that this test fails if the language is ever inadvertently changed or fails to render inside the `data-testid="encryption-caveat-notice"` element.

dwiley-akamai: I don't think there would be any risk of them suffering an incident since there's currently no ability for a user to choose to encrypt an image, but down the line it is possible some users may assume that all resources get encrypted after being conditioned by LDE.

That is a reasonable adjustment though, included it in the most recent commit 🚀

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
```diff
@@ -123,9 +149,9 @@ export const CreateImageTab = () => {
                 text="open a support ticket"
                 title="Request to increase Image size limit when capturing from Linode disk"
               />{' '}
-              to request a higher limit. Additionally, images can't be created
-              from a raw disk or a disk that's formatted using a custom file
-              system.
```
dwiley-akamai, line 128: prompted by linter -- small change to use apostrophe as we do in other places throughout the app instead of the single quotation mark

packages/manager/src/features/Images/ImagesCreate/CreateImageTab.tsx
```diff
@@ -92,6 +101,23 @@ export const CreateImageTab = () => {
...
+  /*
+    We only want to display the notice about disk encryption if:
+    1. the Disk Encryption feature is enabled
+    2. the selected linode is not in an Edge region
+  */
+  const { data: regionsData } = useRegionsQuery();
+
+  const { data: linode } = useLinodeQuery(
+    selectedLinodeId ?? -1,
+    Boolean(selectedLinodeId) && isDiskEncryptionFeatureEnabled
```
dwiley-akamai, line 113: the objective here is prevent this query from firing off constantly if a linode hasn't be selected yet or if LDE is not enabled

AzureLatte (approved):
e2e test create-image.spec.tc passed.

hkhalil-akamai (approved):
Verified the notice appears:

✅ In the Backups tab
✅ In the Storage tab
✅ In the Capture Image tab

✅ Code review looks good. Thanks for including tests.

Approved pending CI passes (`machine-image-upload.spec.ts` appears to be failing).

PR #10524, opened 2024-05-29
refactor: [M3-8183] - Clean up loading components

hana-linode:
## Description 📝
We have a lot of loading components that are all basically just variations of CircleProgress. This PR cleans up those components.

I also moved the overage pricing tooltip text position in the Object Storage Create Bucket drawer to the top so that the tooltip doesn't block the CTA buttons

## Preview 📷
| Before  | After   |
| ------- | ------- |
| image | image |
| image | image |
| image | image |

### Verification steps
(How to verify changes)
- Check the affected loading spinners throughout the app

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/components/MenuItem/MenuItem.tsx
```diff
@@ -88,9 +87,6 @@ export const WrapperMenuItem = (props: WrapperMenuItemCombinedProps) => {
       {...rest}
       className={`${classes.root} ${className} ${tooltip && 'hasTooltip'}`}
     >
-      {isLoading && (
```
hana-linode, line 91: This loading state was never used

packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Backups/BackupSelect.tsx
```diff
@@ -38,11 +38,7 @@ export const BackupSelect = () => {
     }
 
     if (isFetching) {
-      return (
-        <Stack alignItems="center">
-          <CircularProgress />
-        </Stack>
-      );
+      return <LinearProgress />;
```
hana-linode, line 41: It seems like the intention for using CircularProgress was to have a small loading state, I think LinearProgress is a good replacement

packages/manager/src/features/ObjectStorage/BucketLanding/OveragePricing.tsx
```diff
@@ -75,6 +78,7 @@ export const OveragePricing = (props: Props) => {
             the{' '}
             <TextTooltip
               displayText="network transfer pool for this region"
+              placement="top"
```
hana-linode, line 81: Unblock CTA buttons

| Before  | After   |
| ------- | ------- |
| image | image |

packages/manager/src/components/LinkButton.tsx
```diff
@@ -58,7 +56,9 @@ export const LinkButton = (props: Props) => {
     return (
       <Box alignItems="center" display="flex">
         {Button}
-        <CircularProgress className={classes.spinner} size={12} />
+        <Box marginLeft={1}>
+          <CircleProgress mini noPadding size={12} />
```
abailly-akamai, line 60: Would it make sense to not allow custom `size` number and override it with tShirt sizes (this way we can also remove "mini"

ex:
```Typescript
size: 'xs' | 'sm' | 'md' | 'lg'
```

and have JSDocs in the components describe the relative size in number?

```
/**
sizes:
* xs = 12
* md = 18 (or whatever)
* ...
*/

or similar solution? i don't think we need a lot of custom sizes and this may lead to inconsistencies. Just a thought!

**Review by abailly-akamai — COMMENTED** [review]
Nice cleanup! PR looks good. left some comments & suggestions

File: packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Backups/BackupSelect.tsx (hunk 2 above)
Comment (hana-linode) [reply packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Backups/BackupSelect.tsx:41]: Confirmed with Andrew that he is fine with this change 🎉 . We use LinearProgress in the payment drawer

File: packages/manager/src/components/LinkButton.tsx (hunk 4 above)
Comment (hana-linode) [reply packages/manager/src/components/LinkButton.tsx]: Good point, addressed!

**Review by abailly-akamai — APPROVED** [review]
Great Cleanup! did not notice regressions through the flows. Thanks for looking at improving the size API ✨ 

One improvement remains IMO:

when we don't enforce a size, `<CircleProgress />` has different treatment:
Screenshot 2024-06-04 at 09 16 46

It's larger and features an inner circle. it's a common loader since it is used for the splashscreen and and loading landing page
For the sake of addressing this globally and make the component's API clear (right now it's unclear NOT adding a prop renders the largest version), i think we should either:
- require size, add a `lg` option and add it to all circle progress without a size prop (will increase PR size but it's a no regression change)
- still add an `lg` option, and set it by default in the component. 
- for either option, i would suggest removing the inner circle cause i am not sure what it adds?

These are just suggestions, it's not a deal breaker to go along with this cleanup. Approving pending whatever solution you go with and others chiming in

File: packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Marketplace/AppsList.tsx [hunk 5]
```diff
@@ -53,7 +53,7 @@ export const AppsList = ({ onOpenDetailsDrawer }: Props) => {
         justifyContent="center"
         width="100%"
       >
-        <CircularProgress />
+        <CircleProgress />
```
bnussman-akamai, line 56: Can we use size `md` here?

packages/manager/src/components/CircleProgress/CircleProgress.tsx
bnussman-akamai, line 1: I think my feedback is similar to @abailly-akamai's. 

- I think `lg` should be a valid option (especially if it is the default)
- I don't think the `size` should change the appearance of the sinner. (Basically, I think we should remove that weird gray circle or make is show on all sizes)
- Should we rename this component to `CircularProgress` so that it matches MUI's naming. Maybe this will deter people from importing `@mui/material/CircularProgress` directly

hana-linode: Addressed first two bullets. I think `CircleProgress` is easier to type than `CircularProgress`

packages/manager/src/components/CircleProgress/CircleProgress.tsx
```diff
@@ -1,58 +1,62 @@
...
-   */
-  noInner?: boolean;
-  /**
-   * Removes the padding for `mini` circle progresses only.
+   * Removes the padding
    */
   noPadding?: boolean;
   /**
-   * To be primarily used with mini and noPadding. Set spinner to a custom size.
+   * Set spinner to a smaller custom size
```
bnussman-akamai, line 20:
```suggestion
   * Sets the size of the spinner
   * @default "lg"
```

packages/manager/src/components/CircleProgress/CircleProgress.tsx
```diff
@@ -1,58 +1,62 @@
...
 const CircleProgress = (props: CircleProgressProps) => {
-  const { children, mini, noInner, noPadding, size, sx, ...rest } = props;
+  const { children, noPadding, size, sx, ...rest } = props;
 
   const variant =
     typeof props.value === 'number' ? 'determinate' : 'indeterminate';
   const value = typeof props.value === 'number' ? props.value : 0;
 
-  if (mini) {
+  if (size) {
```
bnussman-akamai, line 52: It would be great if we could remove this `if` so that `lg` isn't treated any differently, but this might require some work to not mess up the padding/centering

bnussman-akamai (commented):
I wonder if we could consider `disableShrink `. I think it looks slightly better, but I'm not sure

packages/manager/src/components/CircleProgress/CircleProgress.tsx
hana-linode: Yeah, this is a bit out of scope as the padding and centering is messed up

bnussman-akamai:
Can we move the `disableShrink`  option into the theme?

Screenshot 2024-06-05 at 1 41 25 PM

bnussman-akamai (approved):
Thanks! Nice clean up! 🧹

PR #10519, opened 2024-05-28
fix: [M3-8164] - regions length check

cpathipa:
## Description 📝
This PR checks the length of regions before reading its data while rending Table cell in Access Key table.

## Target release date 🗓️
6/7

### Prerequisites
(How to setup test environment)
- Use dev env and dev account with appropriate account tags.(reach me if you need help on account tags).

### Reproduction steps
(How to reproduce the issue, if applicable)
- Navigate to http://localhost:3000/object-storage/access-keys
- Verify Access Key landing page renders.

### Verification steps
(How to verify changes)
- Navigate to http://localhost:3000/object-storage/access-keys
- Verify Access Key landing page renders.

## As an Author I have considered 🤔

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ❔ Our [contribution guidelines](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md)
- [ ] 🤏 Splitting feature into small PRs
- [ ] ➕ Adding a changeset
- [ ] 🧪 Providing/Improving test coverage
- [ ] 🔐 Removing all sensitive information from the code and PR description
- [ ] 🚩 Using a feature flag to protect the release
- [ ] 👣 Providing comprehensive reproduction steps
- [ ] 📑 Providing or updating our documentation
- [ ] 🕛 Scheduling a pair reviewing session
- [ ] 📱 Providing mobile support
- [ ] ♿  Providing accessibility support

PR #10530, opened 2024-05-30
Release v1.120.2 - release → staging

bnussman-akamai:
### Upcoming Features:

- Fix regions length check in HostNameTableCell ([#10519](https://github.com/linode/manager/pull/10519))

### DIFF
--- a/packages/manager/CHANGELOG.md
+++ b/packages/manager/CHANGELOG.md
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ### Upcoming Features:
 
-- fix regions length check in HostNameTableCell ([#10519](https://github.com/linode/manager/pull/10519))
+- Fix regions length check in HostNameTableCell ([#10519](https://github.com/linode/manager/pull/10519))
 
 ## [2024-05-29] - v1.120.1
 
--- a/packages/manager/CHANGELOG.md
+++ b/packages/manager/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [2024-05-30] - v1.120.2
+
+
+### Upcoming Features:
+
+- fix regions length check in HostNameTableCell ([#10519](https://github.com/linode/manager/pull/10519))
+
 ## [2024-05-29] - v1.120.1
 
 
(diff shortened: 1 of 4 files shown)


jdamore-linode (dismissed):
Thanks @bnussman-akamai! Approved pending tests

cpathipa (approved):

packages/manager/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/) and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [2024-05-30] - v1.120.2
+
+
+### Upcoming Features:
+
+- fix regions length check in HostNameTableCell ([#10519](https://github.com/linode/manager/pull/10519))
```
mjac0bs, line 12:
```suggestion
- Fix regions length check in HostNameTableCell ([#10519](https://github.com/linode/manager/pull/10519))
```

We're doing a hotfix for this, so is it still accurate to have it in Upcoming Features? That feels strange.

bnussman-akamai: Yeah, I'm not sure what makes sense here. I just generated this from the changeset. 

I think the fix is specific to the **upcoming** multi-cluster work, but I guess multi-cluster is launching this release because we are hotfixing this change?

Can you let us know what works best @cpathipa?

cpathipa: We have 6/7 launch for multi-cluster and our next release is 6/10,  Technically this not an hot fix, off cycle release make sense to me to support for upcoming multi-cluster feature .

bnussman-akamai: In that case, "upcoming" makes sense to me. I will push a fix for the capitalization

mjac0bs: Yeah, thanks for the context @cpathipa. It would make sense to update the labels on this PR as well - I believe we have one for off-cycle releases.

bnussman-akamai: Is this really an off-cycle release?

When I hear off-cycle-release I think we're deploying whatever is in develop.  I think hotfix is appropriate because we are releasing a singular fix.

mjac0bs: @bnussman-akamai Honestly, I don't know. You make a good point. 🤷🏼‍♀️ 

My main concern with the changelog was that for people not in-the-know about where MultiCluster is at, it is confusing (for us as a team, but also for anyone who occasionally references our changelogs/PRs when there's issues, like Customer Support) to have something labeled as a hotfix, but be Upcoming. Pretty sure this comment thread will be sufficient explanation for posterity though and we should just fix the capitalization and carry on.

(Idk if labels factor into John's metrics capturing, as a side note.)

PR #10539, opened 2024-06-03
upcoming: [DI-18829] - Resources MultiSelect component in cloudpulse global filters view

venkymano:
## Description 📝

1. Added Resources Selection component inside the Global Filters of CloudPulse view

## Changes  🔄
1. Created new component CloudPulseResourcesSelect
2. Added the component into GlobalFilters of Monitor landing page
3. Added a new useResourcesQuery method to fetch the instances based on the resource type
4. Added unit tests for the resources selection component

## Target release date 🗓️

## Preview 📷

| Before  | After   |
| ------- | ------- |
| 
image
 | 
image
 |
|
image
 |
image
 |

### Prerequisites
- ...
- ...

### Reproduction steps
- ...
- ...

### Verification steps
(How to verify changes)
- ...
- ...

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

---

**Commit Types:**

---

### DIFF
--- a/packages/manager/src/features/CloudPulse/Overview/GlobalFilters.tsx
+++ b/packages/manager/src/features/CloudPulse/Overview/GlobalFilters.tsx
@@ -5,6 +5,10 @@ import * as React from 'react';
 import { WithStartAndEnd } from 'src/features/Longview/request.types';
 
 import { CloudPulseRegionSelect } from '../shared/RegionSelect';
+import {
+  CloudPulseResources,
+  CloudPulseResourcesSelect,
+} from '../shared/ResourceMultiSelect';
 import { CloudPulseTimeRangeSelect } from '../shared/TimeRangeSelect';
 
 export interface GlobalFilterProperties {
@@ -26,6 +30,9 @@ export const GlobalFilters = React.memo((props: GlobalFilterProperties) => {
   });
 
   const [selectedRegion, setRegion] = React.useState<string>();
+  const [selectedResources, setResources] = React.useState<
+    CloudPulseResources[]
+  >();
   React.useEffect(() => {
     const triggerGlobalFilterChange = () => {
       const globalFilters: FiltersObject = {
@@ -54,6 +61,13 @@ export const GlobalFilters = React.memo((props: GlobalFilterProperties) => {
     setRegion(region);
   }, []);
 
+  const handleResourcesSelection = React.useCallback(
+    (resources: CloudPulseResources[]) => {
+      setResources(resources);
+    },
+    []
+  );
+
   return (
     <Grid container sx={{ ...itemSpacing, padding: '8px' }}>
       <StyledGrid xs={12}>
@@ -62,7 +76,15 @@ export const GlobalFilters = React.memo((props: GlobalFilterProperties) => {
             handleRegionChange={handleRegionChange}
           />
         </Grid>
-        <Grid sx={{ marginLeft: 12, width: 250 }}>
+
+        <Grid sx={{ marginLeft: 2, width: 350 }}>
+          <StyledCloudPulseResourcesSelect
+            handleResourcesSelection={handleResourcesSelection}
+            region={selectedRegion}
+            resourceType={'linode'} // for now passing this static value, will be made dynamic once resource selection component is ready
+          />
+        </Grid>
+        <Grid sx={{ marginLeft: 2, width: 250 }}>
           <StyledCloudPulseTimeRangeSelect
             defaultValue={'Past 30 Minutes'}
             handleStatsChange={handleTimeRangeChange}
@@ -85,6 +107,12 @@ const StyledCloudPulseTimeRangeSelect = styled(CloudPulseTimeRangeSelect, {
   label: 'StyledCloudPulseTimeRangeSelect',
 })({
   width: 150,
+});
+
+const StyledCloudPulseResourcesSelect = styled(CloudPulseResourcesSelect, {
+  label: 'StyledCloudPulseResourcesSelect',
+})({
+  width: 250,
 });
 
 const StyledGrid = styled(Grid, { label: 'StyledGrid' })(({ theme }) => ({
--- a/packages/manager/src/features/CloudPulse/shared/RegionSelect.tsx
+++ b/packages/manager/src/features/CloudPulse/shared/RegionSelect.tsx
@@ -14,7 +14,9 @@ export const CloudPulseRegionSelect = React.memo(
     const [selectedRegion, setRegion] = React.useState<string>();
 
     React.useEffect(() => {
-      props.handleRegionChange(selectedRegion);
+      if (selectedRegion) {
+        props.handleRegionChange(selectedRegion);
+      }
       // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [selectedRegion]);
 
--- a/packages/manager/src/features/CloudPulse/shared/ResourceMultiSelect.test.tsx
+++ b/packages/manager/src/features/CloudPulse/shared/ResourceMultiSelect.test.tsx
@@ -1,0 +1,160 @@
+import { fireEvent, screen } from '@testing-library/react';
+import * as React from 'react';
+
+import { linodeFactory } from 'src/factories';
+import { renderWithTheme } from 'src/utilities/testHelpers';
+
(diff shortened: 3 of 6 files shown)


packages/manager/src/features/CloudPulse/Overview/GlobalFilters.tsx
```diff
@@ -26,6 +30,9 @@ export const GlobalFilters = React.memo((props: GlobalFilterProperties) => {
   });
 
   const [selectedRegion, setRegion] = React.useState<string>();
+  const [selectedResources, setResources] = React.useState<
```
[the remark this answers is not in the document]
venkymano: @cpathipa I had kept this state variable for future uses where we will implement a dashboard component, so at that time, we will use this

packages/manager/src/features/CloudPulse/shared/RegionSelect.tsx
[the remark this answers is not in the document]
venkymano: done

packages/manager/src/features/CloudPulse/shared/ResourceMultiSelect.test.tsx
```diff
@@ -0,0 +1,160 @@
+import { fireEvent, screen } from '@testing-library/react';
+import * as React from 'react';
+
+import { linodeFactory } from 'src/factories';
+import { renderWithTheme } from 'src/utilities/testHelpers';
+
+import { CloudPulseResourcesSelect } from './ResourceMultiSelect';
```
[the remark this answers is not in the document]
venkymano: done

packages/manager/src/features/CloudPulse/shared/ResourceMultiSelect.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+export interface CloudPulseResourcesSelectProps {
+  defaultSelection?: number[];
+  handleResourcesSelection: (resources: CloudPulseResources[]) => void;
+  placeholder?: string;
+  region: string | undefined;
+  resourceType: string | undefined;
+}
+
+export const CloudPulseResourcesSelect = React.memo(
```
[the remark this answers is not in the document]
venkymano: done

packages/manager/src/queries/cloudpulse/resources.ts
bnussman-akamai, line 1: We should reuse existing queries rather than building `useResourcesQuery` so that we have the benefit of using existing data that may be cached.  (See `useAllLinodesQuery`)

If it doesn't make sense to use `useAllLinodesQuery`, we should at least use the same query key so reads the same data from the cache.

packages/manager/src/features/CloudPulse/shared/CloudPulseResourcesSelect.tsx
```diff
@@ -0,0 +1,96 @@
...
+    };
+
+    const applyRegionFilter = (instance: CloudPulseResources): boolean => {
+      return (
+        (props.region === undefined ||
+          instance.region === props.region ||
+          (instance.regions && instance.regions.includes(props.region))) ??
+        false
+      );
+    };
```
bnussman-akamai, line 49: Instead of using client side filter, should we X-Filter by region using the Linode API?

For customers with thousands of entities, this _might_ grant us better performance

packages/manager/src/queries/cloudpulse/resources.ts
```diff
@@ -0,0 +1,47 @@
...
+import { getAllLinodesRequest } from '../linodes/requests';
+import { getAllVolumes } from '../volumes/requests';
+
+export const useResourcesQuery = (
+  key: any,
+  enabled: boolean,
+  resourceType: string | undefined
+) => {
+  return useQuery<any[], unknown, CloudPulseResources[]>(
+    [key, resourceType],
```
bnussman-akamai, line 15: To manage query keys, we use https://github.com/lukemorales/query-key-factory. Please use a query key factor for any new CloudPulse queries. You can use `packages/manager/src/queries/volumes/volumes.ts` as an example

packages/manager/src/queries/cloudpulse/resources.ts
```diff
@@ -0,0 +1,47 @@
...
+import { getAllLoadbalancers } from '../aclb/requests';
+import { getAllLinodesRequest } from '../linodes/requests';
+import { getAllVolumes } from '../volumes/requests';
+
+export const useResourcesQuery = (
+  key: any,
+  enabled: boolean,
+  resourceType: string | undefined
+) => {
+  return useQuery<any[], unknown, CloudPulseResources[]>(
```
bnussman-akamai, line 14: When building new queries, please use single object notation. 
We need to do this because this format is [not supported in TanStack Query v5](https://tanstack.com/query/v5/docs/framework/react/guides/migrating-to-v5#supports-a-single-signature-one-object) 

Use this ✅ 
```ts
useQuery({ queryKey: '', queryFn: () => {} });
```

Not this ⛔  
```ts
useQuery("", () => {});
```

packages/manager/src/features/CloudPulse/shared/CloudPulseResourcesSelect.tsx
```diff
@@ -0,0 +1,96 @@
...
+        onChange={(_: any, resourceSelections: any) => {
+          setResources(resourceSelections);
+          props.handleResourcesSelection(resourceSelections);
+        }}
+        autoHighlight
+        clearOnBlur
+        data-testid={'Resource-select'}
+        disabled={!props.region || !props.resourceType || isLoading}
+        isOptionEqualToValue={(option, value) => option.label === value.label}
+        key={'multi-select-resource'}
```
bnussman-akamai, line 75: Is a `key` necessary?

cpathipa:
@venkymano Thank you for addressing the feedback. Changes looks good to me. I agree with @bnussman-akamai feedback.

packages/manager/src/queries/cloudpulse/resources.ts
venkymano: Hi @bnussman-akamai ,

Thanks for this.

Our use case is bit different, our service types can be anything like, linodes, volumes, loadbalancers etc. Using the useAllLinodesQuery hook directly is not possible for this because, inside the component itself we don't want maintain lot of boilerplate code. 

I have done and updated the resources.ts with the following, using your review comments as well

1.  I have created a new query hook, that acts like a factory between different services like linode, volumes, loadbalancers etc., and converts the data to a specific format that component can understand it easily.

2. According to your comments, now the query hook tries to re use the query factories defined in the service query files, like volumes.ts and linodes.ts.

3. If we have the query factory defined in the respective files, it will re use so that cache and everything work properly. If the files like linodes.ts does not have any query factory implementation, in this case we use the same query keys in our factory, so that cache works fine.

4. I also addressed your comment of passing the region directly as filter, instead of doing it on our own.

packages/manager/src/features/CloudPulse/shared/CloudPulseResourcesSelect.tsx
venkymano: I also addressed your comment of passing the region directly as filter, instead of doing it on our own. Done

packages/manager/src/queries/cloudpulse/resources.ts
venkymano: Since this component borrows the queryFn, queryKey from query factories of other services like defined in volumes.ts..we don't need a separate setup here.

According to your comments, now the query hook tries to re use the query factories defined in the service query files, like volumes.ts and linodes.ts.

If we have the query factory defined in the respective files, it will re use so that cache and everything work properly. If the files like linodes.ts does not have any query factory implementation, in this case we use the same query keys in our factory, so that cache works fine.

packages/manager/src/queries/cloudpulse/resources.ts
venkymano: This is done

nikhagra (approved):

bnussman-akamai (approved):

bnussman-akamai:
Looks like there is a typescript error that needs to be resolved

cpathipa (approved):

kmuddapo (approved):

PR #10545, opened 2024-06-04
upcoming: [M3-8015] - Update Images Landing table

hkhalil-akamai:
## Description 📝
Updates the Images landing table with new columns in preparation for the launch of Image Service v2.

## Changes  🔄
- New columns added to `ImagesLanding` and `ImageRow`:
  - Regions
  - Compatibility
  - Total size
  - Image Id
- Added unit tests for `ImagesLanding` and `ImageRow`
- Slight refactor to `EditImageDrawer` and `RestoreImageDrawer` to simplify props and state management

## Target release date 🗓️
6/24

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-06-14 at 4 49 37 PM| Screenshot 2024-06-14 at 4 48 44 PM |

### Prerequisites
Although the alpha API supports the new images endpoints, the underlying image service is not currently operational. You can use the mock service worker to test these changes instead.

For now, these changes are not being feature flagged. Instead, CM will detect if Image Service v2 is enabled by checking if images returned from the API have the new `regions` field.

### Verification steps
- Verify new columns appear in the expected order when Image Service V2 is enabled

### UX
* Default values for "Compatibility" column? (e.g., blank or "None")
* Which columns to hide for `md` breakpoint? Currently, column breakpoints are:
  * `sm` and up: image label and size
  * `md` and up: all columns (regions, compatibility, total size, created, image id)
    * This causes wrapping and overflow on medium-sized viewports

### API
* Is `total_size` field available for all images -- (e.g., is it available for failed or pending images)?
* Is there plans to support sorting on `id` and `total_size` fields?

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -176,19 +177,22 @@ export const ImagesLanding = () => {
     imageEvents
   );
 
+  // TODO Image Service V2: delete after GA
+  const multiRegionsEnabled =
+    manualImages?.data.some((image) => image.regions.length) ?? false;
+
```
hkhalil-akamai, line 183: There is no feature flag for this release. We can detect if Image Service v2 is enabled if the `regions` field is non-empty.

packages/manager/src/features/Images/EditImageDrawer.tsx
hkhalil-akamai, line 1: Slightly simplified the props in this file

packages/manager/src/features/Images/ImagesActionMenu.tsx
```diff
@@ -60,6 +64,15 @@ export const ImagesActionMenu = (props: Props) => {
               ? 'Image is not yet available for use.'
               : undefined,
           },
+          ...(onManageRegions
```
hkhalil-akamai, line 67: This option shouldn't appear until image service v2 is enabled.

packages/manager/src/features/Images/RebuildImageDrawer.tsx
hkhalil-akamai, line 1: See comment on `EditImageDrawer`

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -403,14 +424,19 @@ export const ImagesLanding = () => {
...
+              {multiRegionsEnabled && (
+                <Hidden smDown>
+                  <TableSortCell
+                    active={manualImagesOrderBy === 'id'}
+                    direction={manualImagesOrder}
+                    // eslint-disable-next-line @typescript-eslint/no-empty-function
+                    handleClick={/* handleManualImagesOrderChange*/ () => {}} // TODO: sorting on id currently broken
+                    label="id"
+                  >
+                    Image Id
```
bnussman-akamai, line 436:
```suggestion
                    Image ID
```

packages/manager/src/features/Images/ImagesLanding.tsx
```diff
@@ -403,14 +424,19 @@ export const ImagesLanding = () => {
...
-                active={manualImagesOrderBy === 'size'}
-                direction={manualImagesOrder}
-                handleClick={handleManualImagesOrderChange}
-                label="size"
-              >
-                Size
-              </TableSortCell>
+              {multiRegionsEnabled && (
+                <Hidden smDown>
+                  <TableSortCell
```
bnussman-akamai, line 429: We should probably do a `TableCell` for now because we likely won't have filterability on the ID

packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -640,6 +651,7 @@ export const handlers = [
     const images = [
       cloudinitCompatableDistro,
       cloudinitCompatableImage,
+      multiRegionsImage,
```
bnussman-akamai, line 654: Could we update the MSW handler to check the X-Filter so that the Custom Images / Recovery Images show properly on the landing page? This would help me visualize/test the changes because I haven't been able to create an Image with `regions` yet using the alpha api.

packages/manager/src/features/Images/ImagesActionMenu.tsx
bnussman-akamai: I wonder if we should also feature flag this in case they turn on the API feature flag in production but the manage drawer isn't ready. We'll probably be fine, but we'll just need to make sure our work is fully implemented before they turn on this stuff in prod if we choose not to feature flag.

packages/manager/src/features/Images/ImageRow.tsx
```diff
@@ -1,31 +1,76 @@
...
+            ,{' '}
+            <StyledLinkButton onClick={() => handlers.onManageRegions?.(image)}>
+              +{regions.length - 1}
+            </StyledLinkButton>
+          </>
+        )}
+      </>
+    ) : (
+      ''
+    ));
```
bnussman-akamai, line 68: Should move this into it's own component?

packages/manager/src/features/Images/ImagesActionMenu.tsx
hkhalil-akamai: I'm not opposed to using a feature flag instead of the `.regions` field for detecting image service v2 support.

packages/manager/src/features/Images/ImagesLanding.tsx
hkhalil-akamai: Asked Susana for UX confirmation

packages/manager/src/features/Images/ImageRow.tsx
```diff
@@ -63,15 +86,41 @@ const ImageRow = (props: Props) => {
...
+      {multiRegionsEnabled && (
+        <>
+          <Hidden smDown>
+            <TableCell>
+              {regions && regions.length > 0 && (
+                <RegionsList
+                  onManageRegions={() => handlers.onManageRegions?.(image)}
+                  regions={regions}
+                />
+              )}
```
bnussman-akamai, line 99: I know UX said to show nothing if there are no regions, but I think `None` might be cleaner 🤷‍♂️

packages/manager/src/features/Images/ImageRow.tsx
```diff
@@ -63,15 +86,41 @@ const ImageRow = (props: Props) => {
...
+            </TableCell>
+          </Hidden>
+          <Hidden smDown>
+            <TableCell>{compatibilitiesList}</TableCell>
```
bnussman-akamai, line 103: Same thing here: I know UX said to show nothing if there are no capabilities, but I think `None` or `N/A` might be cleaner 🤷‍♂️  

When I see an empty table cell, I think it looks broken / not intentional, but maybe that's just me.

bnussman-akamai (approved):
Looks great!

Test with MSW on and also verified there were no regressions with MSW off

AzureLatte (approved):

packages/manager/src/features/Images/ImageRow.tsx
hkhalil-akamai: My understanding is that, after the launch of Image Service v2, all images will have at least one region?

packages/manager/src/features/Images/ImageRow.tsx
hkhalil-akamai: I'm on the fence here because showing nothing would leave a lot of blank space if most images don't have compatibility, but showing `None` could be confusing too since it suggests the image is somehow incompatible. I think we should follow UX's guidelines and adjust based on customer feedback during LA/GA?

packages/manager/src/features/Images/ImagesLanding.tsx
hkhalil-akamai: Confirmed capitalization should be "Image ID"

packages/manager/src/features/Images/ImagesActionMenu.tsx
hkhalil-akamai: @bnussman-akamai added `imageServiceGen2` feature flag which must be turned on to see the new columns.

PR #10468, opened 2024-05-14
refactor: [M3-7382] - Use `object-storage/types` endpoint for pricing data 

mjac0bs:
## Description 📝
API is now returning pricing data through the `/object-storage/types` endpoint. 

In this PR, prices can be dynamically displayed for base region pricing and DC-specific region pricing, using the prices returned in the new endpoint, for Object Storage.

> [!NOTE]  
> Outbound transfer overages will be done in a follow-up PR, since pricing for network transfer overages is the same across Cloud Manager and not currently being returned by the API. They are creating a new endpoint for this.

## Changes  🔄
- Created the necessary api-v4 endpoint, query, and interface to use the object-storage/types endpoint.
- Modified `getDCSpecificPriceByType` to optionally accept interval (hourly or monthly) and decimal precision params and updated util unit test accordingly.
- Created the necessary factories and endpoint in the MSW to support a mocked response to the API endpoint.
- Added UX for error handling in the unlikely event that pricing cannot be calculated: disable primary action button and display tooltip error. 
- Added loading state for prices.
- Updated unit tests.

## Target release date 🗓️
5/28/24

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-05-17 at 12 50 05 AM Screenshot 2024-05-17 at 12 50 23 AM Screenshot 2024-05-17 at 12 50 33 AM | Screenshot 2024-05-20 at 8 08 52 AM Screenshot 2024-05-17 at 12 47 38 AM Screenshot 2024-05-17 at 12 47 49 AM |
| N/A | <video src="https://github.com/linode/manager/assets/114685994/c2551927-a383-4fda-a8b7-afa4cc026501"/>|
| Screenshot 2024-05-17 at 12 50 46 AM | Screenshot 2024-05-17 at 12 46 48 AM |
| Screenshot 2024-05-17 at 12 51 03 AM | Screenshot 2024-05-17 at 12 47 10 AM |

### Prerequisites
(How to setup test environment)
- Check out this PR and `yarn up`.
- Go to http://localhost:3000/account/settings and Cancel Object Storage if it is enabled on your account.

### Verification steps
(How to verify changes)
- Go to http://localhost:3000/object-storage/buckets/create.
- Select a region without DC specific pricing (e.g. `us-east`).
- Open the network tab and filter on `/types`. Confirm the "additional storage overage" pricing is consistent with the overage pricing returned by the API endpoint AND that price is consistent with prod.
[...]

PR #10560, opened 2024-06-10
fix: [M3-8162] - Inability to create access keys due to incorrectly running `/types` query

cpathipa:
## Description 📝
This PR fixes two bugs in the Access Key drawer caused by the transition to `object-storage/types` (#10468).

The two issues were that:
1. Since we were disabling the query conditionally, we needed to be using `isInitialLoading` rather than `isLoading` to avoid getting stuck in perpetual loading state
2. Even with the above change, we were still using a value from the query to determine the fallback pricing. If the query was disabled, then that value would be undefined, and we'd see the $--.-- (unknown pricing) for Access Keys. 

## Changes  🔄
- Remove the condition that disabled the query when `regionId` was not provided. This means that the query will always run, which we want to be the case so we can get pricing from the API even when there is no `regionId` (e.g. for Access Keys). If there *is* a regionId provided, we can calculate pricing through the DC pricing util (which will also be $5, but is dynamic in case base pricing by region changes in the future). 

## Target release date 🗓️
6/10

## Preview 📷

| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/linode/manager/assets/114685994/f91b3981-faeb-459e-87e7-c589ab5ae90a" /> | image |

### Prerequisites
- **Cancel Object Storage on your account.**

### Reproduction steps
- On dev, go to object-storage/access-keys (the Create Access key page), and attempt to create a key via the drawer.
- Observe that the Enable Object Storage Modal pops up and is stuck in loading state.

### Verification steps
(How to verify changes)
- On this branch, go to http://localhost:3000/object-storage/access-keys and attempt to create an access key via the drawer.
- Observe that the Enable Object Storage Modal pops up, has pricing filled in, and allows the user to Enable Object Storage and create an access key successfully.
- Confirm that with the obj_use_reagions_mode and regions_allowed tags on your account, you can still see the pricing and create an access key.

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/ObjectStorage/EnableObjectStorageModal.tsx
+++ b/packages/manager/src/features/ObjectStorage/EnableObjectStorageModal.tsx
@@ -1,4 +1,3 @@
-import { Region } from '@linode/api-v4';
 import { styled } from '@mui/material/styles';
 import * as React from 'react';
 
@@ -14,6 +13,8 @@ import {
 } from 'src/utilities/pricing/constants';
 import { getDCSpecificPriceByType } from 'src/utilities/pricing/dynamicPricing';
 
+import type { Region } from '@linode/api-v4';
+
 export const OBJ_STORAGE_STORAGE_AMT = '250 GB';
 export const OBJ_STORAGE_NETWORK_TRANSFER_AMT = '1 TB';
 export interface EnableObjectStorageProps {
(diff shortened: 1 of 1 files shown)


bnussman-akamai (approved):

mjac0bs (approved):
Tested this with and without the OMC tags and confirmed:

1. No loading state, no unknown pricing for Access Keys - can create an access key
2. No regressions to Buckets 

(Note that I pushed [f6983ca](https://github.com/linode/manager/pull/10560/commits/f6983caf4951e4df42c2c5c00c4073eb33645720) with an update and explained in the PR description - as long as this query is always enabled, we don't need to worry about a separate const or handling the loading state differently).

PR #10579, opened 2024-06-13
test: [M3-8107] - Refactor Cypress Longview test to use mock API data/events

AzureLatte:
## Description 📝
Refactor this test so that it uses mock API data/events to test the Longview client creation and installation (and Cloud UI update) flow. This will involve eliminating the script that connects to a Linode and installs Longview and using mock API events to simulate the flow instead.

## Changes  🔄
- Added more factories for mock data
- Set up mocks for each api_action request: lastUpdated, getValues, and getLatestValue
- Refactor the tests to use the new factories

## How to test 🧪
```
yarn cy:run -s "cypress/e2e/core/longview/longview.spec.ts
```

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/cypress/e2e/core/longview/longview.spec.ts
```diff
@@ -100,85 +73,133 @@ const waitForLongviewData = (
...
+      .within(() => {
+        cy.findByText(client.label).should('be.visible');
+        cy.findByText(client.api_key).should('be.visible');
+        cy.contains(installCommand).should('be.visible');
+        cy.findByText('Waiting for data...');
+      });
+
+    // Update mocks after initial Longview fetch to simulate client installation and data retrieval.
+    // The next time Cloud makes a request to the fetch endpoint, data will start being returned.
+    cy.wait(['@fetchLongview', '@fetchLongview', '@fetchLongview']).then(() => {
```
hkhalil-akamai, line 176: Can we reduce this to 2 or even 1 fetch so the test doesn't take as long to complete? Or is this necessary? Same for line 237.

AzureLatte: 3 fetches here was intended to avoid some intermittent issue that was observed during local testing. Let me try to remove some and test again in the jenkins to see if it is stable enough.

jdamore-linode: I'll follow up later this afternoon with an explanation why this is necessary, but I suspect that removing any of them may result in inconsistencies with the mock data that the Longview endpoint responds with (which may not necessarily manifest as a failure)

packages/manager/cypress/e2e/core/longview/longview.spec.ts
```diff
@@ -100,85 +73,133 @@ const waitForLongviewData = (
...
+      .within(() => {
+        cy.findByText(client.label).should('be.visible');
+        cy.findByText(client.api_key).should('be.visible');
+        cy.contains(installCommand).should('be.visible');
+        cy.findByText('Waiting for data...');
+      });
+
+    // Update mocks after initial Longview fetch to simulate client installation and data retrieval.
+    // The next time Cloud makes a request to the fetch endpoint, data will start being returned.
+    cy.wait(['@fetchLongview']).then(() => {
```
jdamore-linode, line 176:
```suggestion
    cy.wait(['@fetchLongview', '@fetchLongview', '@fetchLongview']).then(() => {
```
We do want to revert both of these back to the way they were -- we need to wait for all 3 requests to resolve before proceeding to ensure that each response is mocked with the correct data.

packages/manager/cypress/e2e/core/longview/longview.spec.ts
jdamore-linode: Sorry for the slow follow up!

This is necessary because the Longview landing page, for each client, fires 3 requests to the Longview fetch endpoint when the page loads (and then repeats this process at some interval, kind of like how we poll for events). It's worth pointing out that all 3 of these requests share the same endpoint, but they have a different outgoing payload to specify e.g. what action is being performed or what data is being requested.

Because these all hit the same endpoint, we can't use Cypress to assign an alias to each request -- if we did, they would all satisfy one another. So instead we have to assign one alias to one of our intercepts, and wait for the alias to resolve 3 times so that we can ensure that our _next_ intercepts get set up correctly (e.g. the right data being mocked for the right request). If we don't wait for all 3, depending on how quickly Cypress is executing, it's possible for the next mock data to get set up before the 2nd or 3rd request resolves, causing the landing page to behave as if it's receiving data when the test expects it to be in its "waiting for data..." state.

This weird and careful sequencing of events is only really necessary because all these requests hit the same endpoint, but it's worth noting that there isn't really a serious performance penalty to doing this -- Cloud fires these 3 requests all at once, and they resolve quickly, so waiting for all 3 isn't adding any perceivable time to the test duration, it just ensures that Cypress doesn't proceed a fraction of a second too quickly.

hkhalil-akamai: Ahh this makes a lot of sense. It might be worth including a comment including a brief explanation (or even a link to this comment, lol). Thanks for following up with the detailed explanation!

hkhalil-akamai (approved):
Thanks for addressing feedback! These tests are looking good to me, approved pending the CI passes (they are currently failing across all PRs)

bnussman-akamai (approved):

PR #10611, opened 2024-06-25
upcoming: [M3-8282] - Prevent Linode Create v2 from toggling mid-creation

bnussman-akamai:
## Description 📝

- Prevent Linode Create v2 from toggling mid-creation
  - We are doing this so that if we need to toggle the flag, the user's form progress to be lost. That could be frustrating to a user 😠 

## Preview 📷

| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/linode/manager/assets/115251059/546cb772-b4a8-41f9-a1c9-6c2780de08f2" /> | <video src="https://github.com/linode/manager/assets/115251059/2463eab5-9559-490f-8982-f04f0cbc09ab" /> |
| Notice how the form resets when the flag is toggled because the create flows changed | Notice how the create flow does not change when the flag is toggled | 

## How to test 🧪

- By toggling the flag with local dev tools, verify the form does not reset when the flag is toggled

- [ ] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/.changeset/pr-10611-upcoming-features-1719332278758.md
+++ b/packages/manager/.changeset/pr-10611-upcoming-features-1719332278758.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Upcoming Features
+---
+
+Prevent Linode Create v2 from toggling mid-creation ([#10611](https://github.com/linode/manager/pull/10611))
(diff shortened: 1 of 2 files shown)


abailly-akamai (approved):
Good idea - now, this also means that when we want to turn off `LinodeCreateV2` if we have a bug for instance, more users could encounter it (and report it) while carrying on with the erroneous flow till they restart it. Probably not a huge concern considering we don't anticipate such scenario, this current approach is still improving a much more likely case. Just wanted to point it out in case there's a 100% proof way to handle this.

jaalah-akamai (approved):

PR #10616, opened 2024-06-26
test: [M3-8135] - Add Cypress integration test for Help & Support landing page

cliu-akamai:
## Description 📝
Add integration test for Help & Support landing page.

## Major Changes 🔄
- Check users can view support tickets listed in the Help & Support landing page as expected.

## How to test 🧪
```
yarn cy:run -s "cypress/e2e/core/helpAndSupport/support-tickets-landing-page.spec.ts"
```


AzureLatte (approved):
Test passed

packages/manager/cypress/e2e/core/helpAndSupport/support-tickets-landing-page.spec.ts
```diff
@@ -0,0 +1,315 @@
+/* eslint-disable prettier/prettier */
```
mjac0bs, line 1: Why are we disabling prettier?

packages/manager/cypress/e2e/core/helpAndSupport/support-tickets-landing-page.spec.ts
```diff
@@ -0,0 +1,315 @@
...
+      cy.findByLabelText('Sort by severity').should('be.visible');
+
+      // Confirm that other table columns are shown.
+      cy.findAllByText('Subject').should('be.visible');
+      cy.findAllByText('Ticket ID').should('be.visible');
+      cy.findAllByText('Regarding').should('be.visible');
+      cy.findAllByText('Date Created').should('be.visible');
+      cy.findAllByText('Last Updated').should('be.visible');
+      cy.findAllByText('Updated By').should('be.visible');
+    });
```
mjac0bs, line 293: Do we need to test this again here? If the focus of the test is `"can navigate to the entity's page when clicking the entity's label"`, it would be nice to reduce some of the repetition of coverage in this test and the previous test (`"can navigate to the ticket's page when clicking on the ticket subject"`) as far as page layout, since that's covered in the test case `"lists support tickets in the table as expected"`.

packages/manager/.changeset/pr-10616-tests-1719431797288.md
```diff
@@ -0,0 +1,5 @@
+---
+"@linode/manager": Tests
+---
+
+Add Cypress integration test for Help & Support landing page ([#10616](https://github.com/linode/manager/pull/10616))
```
mjac0bs, line 5:
```suggestion
Add Cypress integration test for Support Ticket landing page ([#10616](https://github.com/linode/manager/pull/10616))
```

mjac0bs (commented):
Thanks for adding coverage on the support ticket landing page and for the changeset. It's great to have more coverage, between this spec and the open-support-ticket spec!

I'm getting consistent local failures when running `yarn cy:run -s "cypress/e2e/core/helpAndSupport/support-tickets-landing-page.spec.ts"`, even though CI passed and I *don't* see the failures when running the spec with `yarn cy:debug`. I'm confused why that's the case.

It seems like the test is unhappy with having multiple support tickets listed in the table and doesn't know which to select. Are you seeing the same failures?

Local output:
Screenshot 2024-06-27 at 8 04 29 AM

Video:

https://github.com/linode/manager/assets/114685994/91c79ee0-3b32-44d3-b63c-2db404ab2c5b

mjac0bs (approved):
Thanks for addressing feedback. I'm approving this because it passes in CI and passed locally in the debugger. My question still stands about whether you'd seen any errors on local runs via `yarn cy:run` and why that might be. (@jdamore-linode, maybe you have some insight? I'm just confused why behavior is different, and not sure if it's only me.)

packages/manager/cypress/e2e/core/helpAndSupport/support-tickets-landing-page.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+      // Confirm that "Severity" table column is not shown.
+      cy.findByLabelText('Sort by severity').should('not.exist');
+
+      // Confirm that other table columns are shown.
+      cy.findAllByText('Subject').should('be.visible');
+      cy.findAllByText('Ticket ID').should('be.visible');
+      cy.findAllByText('Regarding').should('be.visible');
+      cy.findAllByText('Date Created').should('be.visible');
+      cy.findAllByText('Last Updated').should('be.visible');
+      cy.findAllByText('Updated By').should('be.visible');
```
jdamore-linode, line 64: @cliu-akamai Any reason to prefer `cy.findAllByText` in this test? I think we should only use this when we want to select multiple elements, and otherwise I think we should be using `cy.findByText`

cliu-akamai: Right, will change them.

PR #10628, opened 2024-07-01
upcoming: [M3-8066] - Debounce the Linode Create v2 `VLANSelect`

bnussman-akamai:
## Description 📝

- Adds debouncing the `VLANSelect` created for and used on Linode Create v2 ⏲️ 
- Adds a new `useDebouncedValue` hook to make this easy
  - Let me know if you know of a better way 💡  

## Preview 📷

| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/linode/manager/assets/115251059/9bc3cab4-9c88-4980-8a66-d3dcc36b75a5" /> | <video src="https://github.com/linode/manager/assets/115251059/9b7c1a6e-fb92-41b7-b4fe-34e637c0e6b5" /> |

### Prerequisites
- Turn on `Linode Create v2`

### Verification steps
- Verify that the API search performed by the `VLANSelect` is debounced and works as expected

- [ ] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/hooks/useDebouncedValue.ts
+++ b/packages/manager/src/hooks/useDebouncedValue.ts
@@ -1,0 +1,17 @@
+import { useEffect, useState } from 'react';
+
+export const useDebouncedValue = <T>(value: T, delay: number) => {
+  const [debouncedValue, setDebouncedValue] = useState(value);
+
+  useEffect(() => {
+    const handler = setTimeout(() => {
+      setDebouncedValue(value);
+    }, delay);
+
+    return () => {
+      clearTimeout(handler);
+    };
+  }, [value, delay]);
+
+  return debouncedValue;
+};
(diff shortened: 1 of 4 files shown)


packages/manager/src/hooks/useDebouncedValue.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useState } from 'react';
+
+export const useDebouncedValue = <T>(value: T, delay: number) => {
```
jaalah-akamai, line 3: Nit: This is pretty much perfect, my opinion is that we should agree on what a good average is and set it as the default to avoid varying delay implementations. Anywhere between 500-800ms is fine with me - Let me know your thoughts.

jaalah-akamai (approved):

packages/manager/src/hooks/useDebouncedValue.ts
jaalah-akamai: Also I can't believe we didn't have a debounce on this... 🤯

bnussman-akamai: The component is new and exclusive to Linode Create v2 so it hasn't been customer facing just yet.

I just didn't think to add debouncing at the time. @jdamore-linode caught this so thank you Joe!

AzureLatte (approved):
Looks good

PR #10629, opened 2024-07-01
upcoming: [M3-8310] - Add Validation to Linode Create v2 Marketplace Tab

bnussman-akamai:
## Description 📝

Surfaces a `You must select a Marketplace App.` error on the Linode Create v2 Marketplace tab if no app is selected 🚫 

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-07-01 at 11 38 26 AM | Screenshot 2024-07-01 at 11 37 48 AM |

### Prerequisites
- Turn on Linode Create v2

### Verification steps
- Verify you see an error saying `You must select a Marketplace App.` if you try to submit the form without selecting an app 👀 
- Verify this is parity with Linode Create v1 

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/Linodes/LinodeCreatev2/resolvers.ts
+++ b/packages/manager/src/features/Linodes/LinodeCreatev2/resolvers.ts
@@ -4,12 +4,13 @@ import { CreateLinodeSchema } from '@linode/validation';
 import {
   CreateLinodeByCloningSchema,
   CreateLinodeFromBackupSchema,
+  CreateLinodeFromMarketplaceAppSchema,
   CreateLinodeFromStackScriptSchema,
 } from './schemas';
 import { getLinodeCreatePayload } from './utilities';
-import { LinodeCreateFormValues } from './utilities';
 
 import type { LinodeCreateType } from '../LinodesCreate/types';
+import type { LinodeCreateFormValues } from './utilities';
 import type { Resolver } from 'react-hook-form';
 
 export const resolver: Resolver<LinodeCreateFormValues> = async (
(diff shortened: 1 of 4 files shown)


dwiley-akamai (approved):
Error appears and gets scrolled to on Marketplace tab when creation attempted but no app is selected ✅

Improved UX compared to prod ✅

packages/manager/src/features/Linodes/LinodeCreatev2/resolvers.ts
```diff
@@ -107,6 +128,6 @@ export const linodeCreateResolvers: Record<
   'Clone Linode': cloneResolver,
   Distributions: resolver,
   Images: resolver,
```
mjac0bs, line 130: Not relevant to this PR, just curious for my own understanding: is using the `CreateLinodeSchema` `resolver`   (not the additional resolvers like the one in this PR) useful for us in a way that calling `yupResolver(CreateLinodeSchema)` isn't?

mjac0bs (approved):
Verified the validation error is surfaced and scrolled into view without an app selection, consistent with v1. 🚢

packages/manager/src/features/Linodes/LinodeCreatev2/resolvers.ts
bnussman-akamai: I'm not sure I fully understand the question, but the reason we have all of these weird wrappers over `yupResolver` is because we perform some hacky transformations (using `getLinodeCreatePayload`) before submitting. 

Essentially, these wrappers exist so we can run `getLinodeCreatePayload` on the form values before they get validated. When I started Linode Create v2, I really tried to avoid this, but we kind of had to go this route based on PR feedback

mjac0bs: > we perform some hacky transformations (using getLinodeCreatePayload) before submitting

Okay, I see -- this answered my question, thanks!

PR #10653, opened 2024-07-08
feat: [M3-8207] - Auto-close Stale Issues/PRs with Github Actions

bnussman-akamai:
## Description 📝

- Adds a Github Action to auto-close stale PRs
- Based on the [DX team's workflow](https://github.com/linode/linodego/blob/main/.github/workflows/stale.yml)

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,19 @@
+name: 'Stale issue handler'
+
+on:
+  workflow_dispatch: null
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        id: stale
+        with:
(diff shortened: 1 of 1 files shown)


hkhalil-akamai (commented):
Curious what the team thinks about closing PRs after 20 days -- did we already decide in cafe?

dwiley-akamai (commented):
Should we exempt any labels like `Do Not Merge` or `Wait to Merge`?

bnussman-akamai:
I'd be in favor of having no exceptions, but happy to add those exemptions if the team thinks we should

mjac0bs (approved):
My preference is to try it out like it's configured now, and then we can add some exceptions via tags if we find that good reason for it (e.g. there's actually PRs that need to be open for a long time for testing and it becomes annoying to bump it with a new comment saying that). It would be a good retro topic to check in on once this is part of our flow.

dwiley-akamai (approved):
I would be fine with giving this a go and tweaking as needed 👍🏾

PR #10662, opened 2024-07-09
refactor  [M3-8339] - Clean up and fix Linode Details after CDS 2.5

bnussman-akamai:
## Description 📝
- Removes many unnecessary `styled` components 🗑️ 
- Removes unnecessary usages of `Grid` 🗑️
- Fixes dark mode paper theme color 🎨 
- Replaces usages of `AddNewLink` with `Button` 🔘 

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-07-09 at 3 03 28 PM | Screenshot 2024-07-09 at 3 03 44 PM |
| Screenshot 2024-07-09 at 3 04 22 PM | Screenshot 2024-07-09 at 3 04 40 PM |

## How to test 🧪

- Check Linode Details for general styling and functionality 🎨 🔘 
- Verify the theme change looks correct 👁️ 

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset


packages/manager/src/foundations/themes/dark.ts
```diff
@@ -858,7 +858,7 @@ export const darkTheme: ThemeOptions = {
   palette: {
     background: {
       default: customDarkModeOptions.bg.app,
-      paper: Color.Neutrals[100],
+      paper: Color.Neutrals[90],
```
bnussman-akamai, line 861: `bg.bgPaper` is set to `Color.Neutrals[90]` so I think this should probably be set to the same thing. 

`bg.bgPaper` should probably be phased out in the long run in favor of MUI's built in palette value `theme.palette.background.paper`

packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/LinodeIPAddresses.tsx
```diff
@@ -133,101 +133,91 @@ export const LinodeIPAddresses = (props: LinodeIPAddressesProps) => {
...
-          alignItems="flex-end"
-          container
-          justifyContent="space-between"
-          spacing={1}
-        >
-          <Grid className="p0">
-            <StyledTypography variant="h3">IP Addresses</StyledTypography>
-          </Grid>
-          <StyledWrapperGrid>
-            <Hidden smDown>
```
bnussman-akamai, line 151: I removed this `Hidden` because we should probably be allowing mobile users to access `IP Transfer` and `IP Sharing`

hkhalil-akamai: This does lead to some wrapping at narrow widths but it doesn't affect usability:

Screenshot 2024-07-09 at 5 46 22 PM

packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/LinodeFirewalls/LinodeFirewalls.tsx
```diff
@@ -72,18 +72,14 @@ export const LinodeFirewalls = (props: LinodeFirewallsProps) => {
...
-            marginBottom: theme.spacing(),
-            marginLeft: '15px',
-            marginTop: theme.spacing(),
-          })}
-          data-testid="linode-firewalls-table-header"
-          variant="h3"
-        >
+    <Stack>
+      <Box
+        bgcolor={(theme) => theme.palette.background.paper}
```
hkhalil-akamai, line 77: Why are we adding a background color to the headers in this section, is it to match what we had before?

hkhalil-akamai (approved):
Great job simplifying styles while maintaining pixel-parity with what we had before. Tested and looks good in all breakpoints. A few minor questions.

packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/LinodeFirewalls/LinodeFirewalls.tsx
bnussman-akamai: Yeah, exactly. The previous implementation was using `theme.color.white`, but since CDS was merged, the color changed causing these headers to have no visible background in dark mode.  I updated this to use `theme.palette.background.paper`, which seems more fitting anyways.

We could probably use a `Paper` instead of a `Box` here, but I went with Box to keep things more "raw"

abailly-akamai: I don't understand what `keep things more "raw"`means but using the right component is always the better option VS a one time style, especially as the theme will go through more changes in the near future

packages/manager/src/features/Linodes/LinodesDetail/LinodeStorage/LinodeVolumes.tsx
```diff
@@ -164,26 +164,25 @@ export const LinodeVolumes = () => {
...
 
   return (
-    <div style={{ marginTop: '20px' }}>
-      <StyledRootGrid
-        alignItems="flex-end"
-        container
+    <Box>
+      <Box
+        alignItems="center"
+        bgcolor={(theme) => theme.palette.background.paper}
```
abailly-akamai, line 170: same here - if we mean to use a paper treatment, let's use a Paper component

packages/manager/src/foundations/themes/dark.ts
abailly-akamai: We will clean some of that up in future theme layer updates. Ideally we don't want to use either directly anyway. In our perfect world we won't have any exportable theme constants.

packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/LinodeIPAddresses.tsx
```diff
@@ -133,101 +133,91 @@ export const LinodeIPAddresses = (props: LinodeIPAddressesProps) => {
...
+                    {...handlers}
+                    isVPCOnlyLinode={
+                      isVPCOnlyLinode && ipDisplay.type === 'IPv4 – Public'
+                    }
+                    key={`${ipDisplay.address}-${ipDisplay.type}`}
+                    linodeId={linodeID}
+                    readOnly={isLinodesGrantReadOnly}
+                  />
+                ))}
+              </TableBody>
```
abailly-akamai, line 216: Can we clean up inline styles and use `sx` for consistency while we're at it?

packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/LinodeFirewalls/LinodeFirewalls.tsx
bnussman-akamai: Good point. I initially hesitated to use `Paper` because I was worried that its built-in padding would cause problems, but it doesn't matter because I'm specifying the padding manually anyway

Addressed in 8730aaef3e

packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/LinodeIPAddresses.tsx
bnussman-akamai: Yes! Done in 0af5c2e8b5

packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/LinodeIPAddresses.tsx
bnussman-akamai: Updated this to switch over to an `ActionMenu` on small screens in a22ef075bd

abailly-akamai (approved):
thx for addressing the changes!

PR #10675, opened 2024-07-15
change: [M3-7683] - Keep Create & Upload Image Pages' Error Notification Position Consistent

harsh-akamai:
## Description 📝
Keeping the Create & Upload Image Pages' error notification position consistent with other tabs like NodeBalancers.

## Changes  🔄
 - For restricted users:
    - The position of error notification has been moved from inside the box to outside, which is now consistent with other tabs like NodeBalancers.
    - Used getRestrictedResourceText utility to show a Notice / error notification message.

## Preview 📷

Create Image Page:
| Before  | After   |
| ------- | ------- |
|||

Upload Image Page:
| Before  | After   |
| ------- | ------- |
|||

### Prerequisites
- For create page:
  - Log in with a restricted user account (read-only access) and go to the Create Images Page via images/create/disk route directly from you browser.
- For upload page: 
  - Log in with a restricted user account (read-only access) and go to the Upload Images Page via images/create/upload route directly from you browser.

### Reproduction steps
- Observe as restricted user, error notification shows out of the box (not inside as before) and you cannot create/upload Images.

### Verification steps
- As a restricted user, compare Create and Upload Images Page (routes: images/create/disk & images/create/upload respectively) with Create NodeBalancers Page (route: nodebalancers/create) to ensure consistency.

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


hkhalil-akamai (approved):
Thanks for making this fix -- looks good!

mjac0bs (approved):
Thanks Harsh. 

This is not related to the changes in this PR, but noticed while testing and is related to restricted users - do we have a ticket for disabling the "Upload Using Command Line" button? This is a recent addition, so it probably wasn't in the original ticket's acceptance criteria. Since the primary action button is disabled, we should disable the secondary action button too. A user would run into the same permissions issue via CLI. This can be a follow up PR if you'd prefer to keep the scope of this one as-is.

Screenshot 2024-07-15 at 3 08 29 PM

hkhalil-akamai:
> Since the primary action button is disabled, we should disable the secondary action button too. A user would run into the same permissions issue via CLI. This can be a follow up PR if you'd prefer to keep the scope of this one as-is.

I'm not sure if I would agree with this, since the "Upload Using Command Line" button isn't really a regular CTA, it's intended to inform users about the Linode CLI -- perhaps a better option would be a notice in the top of the dialog?

mjac0bs:
> perhaps a better option would be a notice in the top of the dialog?

@hkhalil-akamai Good point about the informational nature. An error notice in the dialog that has the same copy as the restricted user notice at the top of the page would work. It might seem somewhat redundant to have two notices... but it would make it consistent and totally clear that the restricted user couldn't expect a successful action via the CLI either - that was my main concern, but this is a minor issue.

harsh-akamai:
> Thanks Harsh.
> 
> This is not related to the changes in this PR, but noticed while testing and is related to restricted users - do we have a ticket for disabling the "Upload Using Command Line" button? This is a recent addition, so it probably wasn't in the original ticket's acceptance criteria. Since the primary action button is disabled, we should disable the secondary action button too. A user would run into the same permissions issue via CLI. This can be a follow up PR if you'd prefer to keep the scope of this one as-is.
> 
> Screenshot 2024-07-15 at 3 08 29 PM

I have added the disabling feature in the latest commit.

PR #10620, opened 2024-06-26
feat: [M3-7673] - Add Account Limit support ticket to Linode Create flow

mjac0bs:
## Description 📝
This PR adds a new type of support ticket form: the account limit ticket. Currently, when creating a Linode and a user has:
- reached a limit on their account that prevents them from deploying the selected Linode plan
or
- reached the limit for the number of active services on their account

a conditional error notice is displayed. Clicking the contact support link the user is redirected to the Support page and now, in this PR, a new support ticket dialog with custom account-limit-related fields is displayed. Once submitted by the user, the support team can determine if the user is eligible to have access to more Linodes and/or larger plans. According to Customer Support, this is one of the top requests they receive from users. Similar to the [SMTP ticket](https://github.com/linode/manager/pull/8636) form, this change aims to improve the user experience and reduce the amount of back-and-forth between customers and Support when handling requests.
 
> [!NOTE]
> This PR *only* links this account limit support ticket in the Linode Create flow, to start. It will be added to the rest of the flows that users can encounter this notice in follow up PRs.

Follow up tasks:
- Be able to pass data from the submitted form to pre-populate a field like the Linode Plan selection (M3-8336)
- Add the Account Limit support ticket to other product create pages (M3-8335)

## Changes  🔄
- Adds a new support ticket type: `accountLimit`
- Adds new account limit support ticket fields
- Adds a validation schema
- Adds an integration test case
- Handles the already-globally-intercepted support ticket API error with JSX on the Linode Create flows to be able to link to this new ticket type when the expected error text is present in the error message

## Target release date 🗓️
7/22

## Preview 📷

| Description | Before  | After   | 
| ------- | ------- | ------- |
| Error notice | v1: Screenshot 2024-07-08 at 1 30 29 PM Screenshot 2024-07-08 at 1 50 22 PM v2: Screenshot 2024-07-08 at 1 31 28 PM Screenshot 2024-07-08 at 1 49 44 PM| No change |
| Linked support ticket | a general support ticket | Screenshot 2024-07-08 at 1 30 43 PM |
| Submitted ticket | N/A | Screenshot 2024-07-08 at 1 39 33 PM |

### Prerequisites
(How to setup test environment)
- **Please test this on your dev account so we don't bother Customer Support with real tickets.**
- **Please test these changes on Linode Create flow v1 and v2.**
- Check out this PR and go to http://localhost:3000/linodes/create.
- Log into admin and be ready to change some things on your dev account.

### Verification steps
(How to verify changes)
[...]

PR #10684, opened 2024-07-16
feat: [M3-8335] - Link Account Limit ticket in more flows with error notice

mjac0bs:
## Description 📝
#10620 focused on the creation of the new account limit support ticket, linking to it from the Linode Create flow, and providing test coverage.

This ticket will link to the the account limit support ticket from the "contact Support" text on the Volumes, NodeBalancers, Firewall, Kubernetes (LKE create and Add a Node Pool), and Database create flows. Currently, the link in these flows takes users to a general support ticket.

## Changes  🔄
- Removes the global intercept on the text "contact Support" and "open a support ticket"
- Links to an account limit ticket for Create flows listed above when the expected error text is present in the error message
- Fixes a bug in a `handleGeneralError` Formik util that didn't account for the APIError reason being a string *or* JSX

## Target release date 🗓️
8/5/24

## Preview 📷

<details>
<summary>
Videos 🎥 
</summary>

| Flow | Limit Error |
| ------- | ------- | 
| Volumes | <video src="https://github.com/user-attachments/assets/5fdb4771-b4b6-4928-82db-bc7b47d3595d" />  | 
| NodeBalancers | <video src="https://github.com/user-attachments/assets/2fb32a68-1ac5-47ad-b302-db53e0f8a760" />  | 
| LKE Create - Thing Limit | <video src="https://github.com/user-attachments/assets/61288ed6-6eef-4959-874e-89042492d456" />  |
| LKE Create - Rep Limit | <video src="https://github.com/user-attachments/assets/10dc6db1-be18-4d47-b3bb-46bd1ef51dd0" /> | 
| LKE Add Node Pool - Thing Limit | <video src="https://github.com/user-attachments/assets/c4aebdf4-fc4d-48cf-bfae-de4ceb59b3a7" />  |
| LKE Add Node Pool - Rep Limit | <video src="https://github.com/user-attachments/assets/c4aebdf4-fc4d-48cf-bfae-de4ceb59b3a7" />  | <video src="https://github.com/user-attachments/assets/46e04253-41b7-4267-9d4d-117e38a15be4" /> | 
| Firewalls | <video src="https://github.com/user-attachments/assets/1befcbfe-0086-4d80-a27d-d7580cd5e286" />  | 
| Databases | <video src="https://github.com/user-attachments/assets/6f8d5543-3603-407f-bb7a-d203804cc866" /> | 

</details>

### Prerequisites
- **Please test this on your dev account so we don't bother Customer Support with real tickets.**
- Check out this PR and switch your environment to dev.
- Log into admin and be ready to change some things on your dev account.

### Verification steps
(How to verify changes)
- In admin, set your thing limit to 0.
- For each of the above create flows (Volumes, NBs, LKE Create + Add a Node Pool, Firewalls, Databases):
- Fill out the create form and try to create the entity. 
- Confirm that the UI displays an error notice when user has reached a limit on their account: `You’ve reached a limit for the number of active services on your account. Please contact Support to request an increase and provide the total number of services you may need.`
- Click on the "contact Support" link and confirm that you are redirected to the Account Limit ticket and the expected fields show.
   - The Linode plan field should only show for Linode and LKE account limit tickets.
   - The correct entity should be shown in the "total entities" field, with the correct number of existing entities from your account as helper text.
- For LKE only, setting your reputation to 0 should result in the error: `A limit on your account is preventing the deployment of the selected service. To request access, please contact Support and provide the service name.` You should also link to an account limit support ticket.

- Verify that this test isn't flaking in CI:
```
yarn cy:run -s "cypress/e2e/core/helpAndSupport/open-support-ticket.spec.ts"
```

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/utilities/formikErrorUtils.ts
```diff
@@ -49,12 +49,16 @@ export const handleGeneralErrors = (
...
+  const generalError =
+    typeof _apiErrors[0].reason !== 'string'
+      ? _apiErrors[0].reason
+      : _apiErrors
+          .reduce(
+            (result, { field, reason }) =>
+              field ? result : [...result, reason],
+            []
+          )
+          .join(',');
```
mjac0bs, line 61: The misleading string type of _apiErrors was resulting in `[Object object]` being displayed to the user in any case where the error was already globally intercepted converted to JSX, as support errors were. 

I'm proposing that we remove the global intercept for support tickets in this PR, but we still global intercept for verification and migrate errors, so I left the change in this util.

packages/manager/src/request.tsx
```diff
@@ -103,16 +102,6 @@ export const handleError = (
...
-    {
-      condition: (e) => {
-        return (
-          (!!e.reason.match(/.*open a support ticket/i) ||
-            !!e.reason.match(/contact Support/i)) &&
-          !e.field
-        );
-      },
-      replacementText: <SupportError errors={errors} />,
-    },
```
mjac0bs, line 115: Ultimately, I think we should remove this and handle support errors at the component level. As we begin to add more ticket types, it will depend on the context of the component which type of ticket we link to, so it is no longer useful to override at a global level to open a general support ticket. Additionally, our lack of type safety for a `reason` that is either string or JSX adds complexity and errors (see the `[Object object]` in Volume Create; another place this happens is the Configuration dialog when hitting a VLAN limit). We could mitigate these if we leave the construction of JSX to `ErrorMessage.tsx` and `SupportTicketGeneralError.tsx`.

If we do this here, in this PR, some API errors not covered in this PR will no longer link directly to a support ticket. If we're okay with that for now, we can proceed and create a follow up ticket to use `ErrorMessage` with general errors throughout the app - or at least where we think "contact support" API errors are likely.

AzureLatte (approved):
Test passed

bnussman-akamai (approved):
Looks great! I like how we're handling errors at the component level now.

packages/manager/src/request.tsx
bnussman-akamai: I'm good with this!  The interceptor solution was handy, but I'm much more sold on handling at the component level. 

I remember @hkhalil-akamai saying we might even be able to remove the migration and verification interceptors as well because they might not be returned by the API nowadays, but I might be wrong. We can investigate that at a later point.

hana-linode (approved):
UI displays contact support error notice and expected fields show in the support modal ✅ 

Non-blocking: If it's not too complicated, what are your thoughts on replacing the word `entity` with the actual entity?
image

mjac0bs:
> > Non-blocking: If it's not too complicated, what are your thoughts on replacing the word `entity` with the actual entity?
> 
> @hana-linode That would be ideal, agreed. It felt like it would be more complicated than it's worth when I was first implementing, but I'll revisit now that this is a much smaller PR. 👀

The way that we're converting special ticket type form fields into one description field makes this messier than I'd like. I'm going to hold off on it unless we get feedback from customers/Support that 'entities' is confusing.

PR #10690, opened 2024-07-17
fix: [M3-7877] - Incorrect Linode network interface configuration being displayed

hkhalil-akamai:
## Description 📝
Fixes a bug in the Linode configuration dialog causing network interfaces to be displayed incorrectly.

## Changes 🔄
* Fix race condition causing network interfaces to be displayed incorrectly
* Add unit test to verify fix
    * Add `LinodeKernelFactory` to mock kernels
    * I haven't been able to replicate the bug with this unit test -- see [this comment](https://github.com/linode/manager/pull/10690/files#r1681720782) for more details

## Bug pathology 🪳

The following represents my best explanation of the bug but may not be correct:

### The bug is caused by a race condition between two `useEffect`s:
1. Updates the formik form state when config values become available
https://github.com/linode/manager/blob/e60faef92e3670909cf793b758f98e32c92d10a1/packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.tsx#L467-L531

2. Modifies VPC interfaces to add IPv4 information
https://github.com/linode/manager/blob/e60faef92e3670909cf793b758f98e32c92d10a1/packages/manager/src/features/Linodes/LinodesDetail/LinodeSettings/InterfaceSelect.tsx#L230-L282

### The race condition occurs in the following sequence:

1. The `LinodeConfigDialog` is opened
2. `config` is updated from `undefined` to the selected Linode's config -> this sets off `useEffect` 1 which queues the new interface values in formik
    - Due to [Automatic Batching](https://react.dev/blog/2022/03/29/react-v18#new-feature-automatic-batching) in React 18, the formik state will not actually be updated until the next render
4. `InterfaceSelect` is mounted -> this sets off `useEffect` 2, overriding the updates from step 2

### The fix:
Update `useEffect` 2 to return early if the interface is non-VPC

## Target release date 🗓️
7/22

## Preview 📷
| Before  | After   |
| ------- | ------- |
|  <video src="https://github.com/user-attachments/assets/6876a997-b83a-45e9-afd1-494ddd6440ed" />|<video src="https://github.com/user-attachments/assets/cb8a65fa-d32d-40ea-9cc0-fef2eb4f63ab" />|

### Prerequisites
- Create a Linode with a non-default network configuration
   - For example:
        **`eth0`:** VPC
        **`eth1`:** Public internet

### Reproduction steps
This bug is caused by a race condition and is tricky to reproduce consistently. The following steps seem to trigger the bug reliably:

- Navigate to the newly created Linode (from Prerequisites)
- Navigate to the "Configurations" tab and click "Edit" on the non-default configuration to open the configuration dialog
- Exit out of the configuration dialog and navigate to the different section of the app (e.g., Volumes, NodeBalancers, etc.)
- Return to the Linode configuration dialog and observe the network interface fields are not correct (i.e., they show the default network interfaces)

### Verification steps
- Perform the same steps as above but observe the correct network interfaces are always dispalyed

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/features/Linodes/LinodesDetail/LinodeSettings/InterfaceSelect.tsx
```diff
@@ -229,11 +229,7 @@ export const InterfaceSelect = (props: InterfaceSelectProps) => {
 
   React.useEffect(() => {
     if (purpose !== 'vpc') {
-      return handleChange({
-        ipam_address: ipamAddress,
-        label,
-        purpose,
-      });
```
hkhalil-akamai, line 236: This is the actual fix.

packages/manager/src/utilities/testHelpers.tsx
```diff
@@ -146,8 +142,15 @@ export const wrapWithTableBody = (ui: any, options: Options = {}) =>
...
-export const renderWithTheme = (ui: any, options: Options = {}) => {
-  return render(wrapWithTheme(ui, options));
+export const renderWithTheme = (
+  ui: React.ReactNode,
+  options: Options = {}
+): RenderResult => {
+  const renderResult = render(wrapWithTheme(ui, options));
+  return {
+    ...renderResult,
+    rerender: (ui) => renderResult.rerender(wrapWithTheme(ui, options)),
```
hkhalil-akamai, line 152: Adding support for the `rerender` method

packages/manager/src/features/Linodes/LinodesDetail/LinodeConfigs/LinodeConfigDialog.test.tsx
```diff
@@ -146,4 +146,29 @@ describe('LinodeConfigDialog', () => {
...
+          interfaces: [vpcInterface, publicInterface],
+        })}
+        open
+        {...props}
+      />
+    );
+
+    await findByText('VPC');
+    await findByText('Public Internet');
+  });
```
hkhalil-akamai, line 173: In theory, this test should replicate the exact sequence causing the bug. However, I haven't been able to fail this test without the fix, which I believe is due to inconsistencies between how [Automatic Batching](https://react.dev/blog/2022/03/29/react-v18#new-feature-automatic-batching) works in tests vs in the browser.

I'm still keeping this test because may catch regressions in the future.

packages/manager/src/features/Linodes/LinodesDetail/LinodeSettings/InterfaceSelect.tsx
hkhalil-akamai: @hana-linode can you verify this change has no unintended consequences? I wasn't able to figure out the original purpose of this.

AzureLatte (approved):

bnussman-akamai (commented):
After making no changes, clicking "Save Changes", and refreshing, the config's VPC looked different than how it was originally configured. The after refreshing again, it went back to the expected values.

Not sure if this is related. I haven't been able to easily reproduce.

https://github.com/user-attachments/assets/e191f99f-6164-42c4-b866-0276c8814ab6

hkhalil-akamai:
> After making no changes, clicking "Save Changes", and refreshing, the config's VPC looked different than how it was originally configured. The after refreshing again, it went back to the expected values.

I have been unable to reproduce this. There is a lot of complicated logic in the `LinodeConfigDialog` and `InterfaceSelect` components that needs to be cleaned up but I don't know if it's being caused by this change. Are you able to reproduce it in prod?

packages/manager/src/features/Linodes/LinodesDetail/LinodeSettings/InterfaceSelect.tsx
hana-linode: @hkhalil-akamai It's been a while so I don't really remember, but I _think_ the original purpose was to return early if the purpose wasn't vpc so that vpc properties would not be passed to the `handleChange` function

bnussman-akamai (approved):
I wasn't able to reproduce the issue I showed above, so I think we're good

hana-linode:
I have been able to reproduce the bug @bnussman-akamai saw on this branch on a fresh Linode and new VPC created during the flow. I was not able to reproduce in prod 🤔

dwiley-akamai (approved):
Code review ✅
Observing correct network interfaces ✅

I was not able to reproduce the issue Banks observed, and I didn't observe any adverse impacts from the changes in `InterfaceSelect.tsx`

hkhalil-akamai:
@bnussman-akamai @hana-linode @dwiley-akamai

I was able to trace back the issue to the same `useEffect` in *InterfaceSelect.tsx*. I believe I have solved the issue by removing the `useEffect` and rewriting the onChange handlers to modify the interface directly ([545f6db](https://github.com/linode/manager/pull/10690/commits/545f6dba88546e1a7a73def1bb3d6f1da9b18b76)).

The following steps seem to reliably reproduce the issue:

1. Create a new Linode with the `eth0` interface set to VPC.
    - **Deselect** "Auto-assign a VPC IPv4 address for this Linode in the VPC" and manually set an IPv4 address such as "10.0.0.2"
    - **Select** "Assign a public IPv4 address for this Linode"
    - For example, your configuration could look like this:
    
Screenshot 2024-07-23 at 2 06 48 PM

2. Navigate to the Linode details page -> Configuration tab and edit the configuration
3. Verify the VPC interface appears as expected (i.e., with the first checkbox deselected and the second checkbox selected) -- you may need to refresh the page to ensure the interface appears as expected.
4. Navigate to a different page in CM (e.g., NodeBalancers, Volumes, etc.)
5. Return to the Linode details page -> Configuration tab -> Edit configuration and observe the VPC interface has been reset (the first checkbox is selected and the second checkbox is deselected)

hana-linode:
So it seems like editing the config doesn't update the UI now 💀 . See the checkbox still being checked on this branch vs unchecked on prod

https://github.com/user-attachments/assets/36d8d119-eba3-42df-ac10-b598d0b9acbf

bnussman-akamai:
I haven't been able to replicate the first bug I brought up, which is good ✅ 

I am seeing the form reset to a weird state when I try to remove a VPC IP range, which does not happen in production.

https://github.com/user-attachments/assets/fef93041-7ac3-4cde-a745-a36d58fd3e64

bnussman-akamai (approved):
No more bugs found from my testing 🎉

PR #10692, opened 2024-07-17
upcoming: [M3-8315] - Add EU Agreement to Linode Create v2

bnussman-akamai:
## Description 📝

- Adds the "EU Agreement" to Linode Create v2 🇪🇺 
  - Should show if you have not previously agreed to the EU agreement and you are trying to deploy to an EU region

## Preview 📷

Screenshot 2024-07-17 at 6 16 26 PM

### Prerequisites
- Use an account that **does not** have the `eu_model` agreement set to `true` (look for the `/v4/account/agreements` API request on the Linode Create page )
  - I am using devenv to do this because you can't unsign agreements as far as I know🙄 

### Verification steps
- Verify the checkbox shows if
  - You have not agreed to the `eu_model`
  - You have selected an EU region
- Verify you get a validation error if you try to create a Linode in an EU region when the checkbox is showing and you have not checked it
- Verify that creating a Linode in an EU region causes the EU agreement to be signed

- [ ] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/Linodes/LinodeCreatev2/SMTP.test.tsx
+++ b/packages/manager/src/features/Linodes/LinodeCreatev2/SMTP.test.tsx
@@ -1,0 +1,27 @@
+import React from 'react';
+
+import { accountFactory } from 'src/factories';
+import { HttpResponse, http, server } from 'src/mocks/testServer';
+import { renderWithTheme } from 'src/utilities/testHelpers';
+
+import { SMTP } from './SMTP';
+
+describe('SMTP', () => {
+  it('should render if the account was actived before MAGIC_DATE_THAT_EMAIL_RESTRICTIONS_WERE_IMPLEMENTED', async () => {
+    const account = accountFactory.build({
+      active_since: '2022-11-29T00:00:00.000',
+    });
+
+    server.use(
+      http.get('*/account', () => {
+        return HttpResponse.json(account);
+      })
+    );
+
+    const { findByText } = renderWithTheme(<SMTP />);
+
+    await findByText('SMTP ports may be restricted on this Linode', {
+      exact: false,
+    });
+  });
+});
--- a/packages/manager/src/features/Linodes/LinodeCreatev2/SMTP.test.tsx
+++ b/packages/manager/src/features/Linodes/LinodeCreatev2/SMTP.test.tsx
@@ -7,7 +7,7 @@ import { renderWithTheme } from 'src/utilities/testHelpers';
 import { SMTP } from './SMTP';
 
 describe('SMTP', () => {
-  it('should render if the account was actived before MAGIC_DATE_THAT_EMAIL_RESTRICTIONS_WERE_IMPLEMENTED', async () => {
+  it('should render if the account was activated before MAGIC_DATE_THAT_EMAIL_RESTRICTIONS_WERE_IMPLEMENTED', async () => {
     const account = accountFactory.build({
       active_since: '2022-11-29T00:00:00.000',
     });
--- a/packages/manager/src/features/Linodes/LinodeCreatev2/resolvers.ts
+++ b/packages/manager/src/features/Linodes/LinodeCreatev2/resolvers.ts
@@ -1,6 +1,10 @@
 import { yupResolver } from '@hookform/resolvers/yup';
 import { CreateLinodeSchema } from '@linode/validation';
 
+import { accountQueries } from 'src/queries/account/queries';
+import { regionQueries } from 'src/queries/regions/regions';
+import { getRegionCountryGroup, isEURegion } from 'src/utilities/formatRegion';
+
 import {
   CreateLinodeByCloningSchema,
   CreateLinodeFromBackupSchema,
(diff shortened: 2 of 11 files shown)


packages/manager/src/features/Linodes/LinodeCreatev2/resolvers.ts
```diff
@@ -11,123 +15,68 @@ import { getLinodeCreatePayload } from './utilities';
...
-) => {
-  const transformedValues = getLinodeCreatePayload(structuredClone(values));
+      if (!hasSignedEUAgreement && !values.hasSignedEUAgreement) {
+        errors['hasSignedEUAgreement'] = {
+          message:
+            'You must agree to the EU agreement to deploy to this region.',
+          type: 'validate',
+        };
+      }
+    }
```
mjac0bs, line 63: I never saw this validation error because when the checkbox isnt't checked (rest of the form is filled out), the create button is disabled. Did I miss something?

Screenshot 2024-07-18 at 3 11 11 PM

mjac0bs (approved):
✅ Verified the checkbox shows if
- You have not agreed to the eu_model
- You have selected an EU region

❓  Verify you get a validation error if you try to create a Linode in an EU region when the checkbox is showing and you have not checked it
✅  Verify that creating a Linode in an EU region causes the EU agreement to be signed

EU agreement copy/checkbox displays and then is no longer displayed as expected depending on the user's agreement/lack thereof. Wasn't sure about the validation error, but the user can't submit without acknowledging the agreement, so the same result is achieved.

packages/manager/src/features/Linodes/LinodeCreatev2/resolvers.ts
bnussman-akamai: That is Linode Create v1 you are looking at. You'll need to refresh after you toggle the flag to see v2

abailly-akamai (approved):
Confirm the toggling and validation of the EU agreement - tested on a real account without any resource created ✅

packages/manager/src/features/Linodes/LinodeCreatev2/resolvers.ts
mjac0bs: 🤦🏼‍♀️ Thanks, should have realized that based on the paper. Agreements section with EU agreement looks good.

Screenshot 2024-07-19 at 7 03 11 AM

Did we intentionally not include SMTP helper text in v2?

mjac0bs:
Non-blocking: Just noticed... can we modify some margin in the `EUAgreementCheckbox` to line up the text and checkbox with those in the Add-ons section above it?

I think we can get rid of the [negative margin](https://github.com/linode/manager/blob/develop/packages/manager/src/features/Account/Agreements/EUAgreementCheckbox.tsx#L24-L27) on tablet size on down. And then make the [left margin on the text](https://github.com/linode/manager/blob/develop/packages/manager/src/features/Account/Agreements/EUAgreementCheckbox.tsx#L59) `16` instead of `4`.


| Currently | Adjusted |
| --- | --- |
| Screenshot 2024-07-19 at 7 22 12 AM Screenshot 2024-07-19 at 7 22 01 AM  | Screenshot 2024-07-19 at 7 21 18 AM Screenshot 2024-07-19 at 7 21 42 AM |

packages/manager/src/features/Linodes/LinodeCreatev2/resolvers.ts
bnussman-akamai: Oh boy. We'll probably need SMTP helper text. Completely forgot that was a thing

mjac0bs: Yeah. I've never loved SMTP's current placement, but think we should keep it where it currently lives in v1, since it's not an agreement.

bnussman-akamai: Added in this PR 😌  Ready to get this flow turned on in dev 😮‍💨

packages/manager/src/features/Linodes/LinodeCreatev2/SMTP.test.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+import { accountFactory } from 'src/factories';
+import { HttpResponse, http, server } from 'src/mocks/testServer';
+import { renderWithTheme } from 'src/utilities/testHelpers';
+
+import { SMTP } from './SMTP';
+
+describe('SMTP', () => {
+  it('should render if the account was actived before MAGIC_DATE_THAT_EMAIL_RESTRICTIONS_WERE_IMPLEMENTED', async () => {
```
mjac0bs, line 10:
```suggestion
  it('should render if the account was activated before MAGIC_DATE_THAT_EMAIL_RESTRICTIONS_WERE_IMPLEMENTED', async () => {
```
or `active`?

mjac0bs (approved):
Thanks! 🚀

PR #10694, opened 2024-07-18
test: [M3-8226] - Cypress integration tests for account "Maintenance" tab

AzureLatte:
## Description 📝
Add Cypress integration tests using mock API data to ensure the Cloud Manager "Maintenance" tab on the Account page (/account/maintenance) works as expected.

## Changes  🔄
- Add new test in e2e packages/manager/cypress/e2e/core/account/account-maintenance.spec.ts
- Add new intercepts functions in account.ts

## How to test 🧪
```
yarn cy:run -s "cypress/e2e/core/account/account-maintenance.spec.ts"
```

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


AzureLatte:
observing a timeout issue during integration test. So remove "ready for review" and checking now

AzureLatte:
Ready for review

packages/manager/cypress/e2e/core/account/account-maintenance.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+            .each(($row, idx, $rows) => {
+              cy.wrap($row).within(() => {
+                // Check the content of each <td> element
+                cy.get('td').each(($cell, idx, $cells) => {
+                  cy.wrap($cell).should('not.be.empty');
+                });
+              });
+            });
+        });
+      });
```
jdamore-linode, line 82: We might want to take a different approach here since this is a little more tightly coupled to the page's HTML structure than we'd probably want.

One issue is that this table does not have an `aria-label` attribute -- we normally assign our tables `aria-label`s for accessibility reasons ([example](https://github.com/linode/manager/blob/b78edb7/packages/manager/src/features/Profile/APITokens/APITokenTable.tsx#L214)), and a nice benefit of that is it allows us to use `cy.findByLabelText()` to select whichever table we want from Cypress. It looks like that might've been overlooked for these account maintenance tables, however.

I'd suggest adding an ARIA label to these Account Maintenance tables in `src/features/Account/Maintenance/MaintenanceTable.tsx` on line 162:

```typescript
<Table aria-label={`List of ${type} maintenance`}>
```

That gives you a little bit more flexibility in how you can interact with these tables and assert their contents. For example, you could do something like this:

```typescript
    cy.findByLabelText('List of pending maintenance')
      .should('be.visible')
      .find('tbody')
      .within(() => {
        accountpendingMaintenance.forEach((pendingMaintenance) => {
          cy.get('tr')
            .should('have.length', accountpendingMaintenance.length)
            .each((row, index) => {
              const pendingMaintenance = accountpendingMaintenance[index];
              cy.wrap(row).within(() => {
                cy.contains(pendingMaintenance.entity.label).should('be.visible');
                cy.findByText(pendingMaintenance.reason).should('be.visible');
                // Etc... continue asserting that expected values are displayed in maintenance table row.
              });
            });
        });
      });
```

You could then repeat that for the completed maintenance table, or even wrap the entire thing in a loop that runs once for the pending table and again for the completed table.

(Note that if you do add an ARIA label to these tables, we should probably run `yarn changeset` again so that there's a separate changelog entry for that addition since it'd be a legitimate accessibility improvement)

AzureLatte: Thanks for the advice! I just updated it as suggested. But I also noticed a corner case that one of the completed Maintenance reason is too long to fit in the table (e.g. `volumn7` in the screenshot), so this reason was truncated directly and not displayed in full. I understand the size limit for each cell but just would like to double check it is expected for us.
Screenshot 2024-07-25 at 10 55 17 AM

packages/manager/cypress/e2e/core/account/account-maintenance.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    // Confirm download buttons work
```
AzureLatte, line 84: Ideally we would like to check the contents of the CSV files but looks like the file is empty after downloading. I think it is because we are mocking responses but not actually have those maintenance messages.

Just a quick note that we might unable to confirm the contents of the CSV match an expected string or similar that was mentioned in the ticket.

packages/manager/cypress/e2e/core/account/account-maintenance.spec.ts
jdamore-linode: Oh, great catch!

It looks like Cloud is truncating this string in the `MaintenanceTableRow` component on [line 34](https://github.com/linode/manager/blob/b78edb7/packages/manager/src/features/Account/Maintenance/MaintenanceTableRow.tsx#L34):

```typescript
// MaintenanceTableRow.tsx, ln 34:
const truncatedReason = truncate(reason, 93);
```

We're truncating the string at the 93rd character, but looking at the [implementation of this `truncate` util](https://github.com/linode/manager/blob/a196794/packages/manager/src/utilities/truncate.ts#L5) reveals some information that can help us:

- The util avoids truncating a string mid-word
- To accomplish this, the util performs the truncation _beyond_ the specified max character limit, not before.

In other words, we know that the string that'll be rendered on-screen will only be truncated if it's at least 93 characters long, and we can use this info to perform an accurate-enough assertion that the expected contents are being displayed on-screen. Here's one way you could accomplish that:

```typescript
// Around line 75 of `account-maintenance.spec.ts`, and again around line 102...
// ...
const truncatedReason = pendingMaintenance.reason.substring(0, 90);
cy.findByText(truncatedReason, { exact: false }).should('be.visible');
// ...
```

This will confirm that the first 90 characters of each reason string are rendered on screen, and will work for strings that do not get truncated as well due to the behavior of [JavaScript's substring method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#:~:text=Any%20argument%20value%20that%20is%20less%20than%200%20or%20greater%20than%20str.length%20is%20treated%20as%20if%20it%20were%200%20and%20str.length%2C%20respectively.).

There are certainly other ways you could handle this, so feel free to experiment in case another solution comes to mind that might be a better fit! (You could import the `truncate` util and use it directly within the test for your assertion, you could override the mocks' reason strings to sidestep the issue altogether, etc. -- it's your call!)

packages/manager/cypress/e2e/core/account/account-maintenance.spec.ts
jdamore-linode: @AzureLatte I'm puzzled by this one! I’m observing the empty CSVs being downloaded when the test runs, but if I click either of the "Download CSV" buttons manually inside the Cypress debugger, the file that gets downloaded has the expected mock data:

```csv
"Entity Label","Entity Type","Entity ID","Date","Type","Status","Reason"
"linode-1","linode","1","2023-04-17T03:16:20.148Z","live_migration","pending","We must replace faulty RAM in your Linode's host."
"linode-2","linode","2","2023-10-24T20:57:33.195Z","cold_migration","started","There is an issue with the power supply of your Linode's host."
```

Very strange!

Tomorrow I'll take a closer look at how Cloud handles this to see if it gives us any hints what's happening here. In the meantime, this is all looking really great! If we can't figure out what's going on with the CSV files, we can talk about it spinning that off into a separate task so that all of this other work doesn't get held up!

AzureLatte: > but if I click either of the "Download CSV" buttons myself inside the Cypress debugger, the file that gets downloaded has the expected mock data:

Did you config something special before? I am asking because on my side, the downloaded file is also empty even if I clicked the button manually.

jdamore-linode: Even weirder! Nothing specific that I can think of, but I may have been using `it.only(...)` on the test while I was looking at it -- not that that should make any difference, though 🤔

packages/manager/cypress/e2e/core/account/account-maintenance.spec.ts
AzureLatte: Thank you!

packages/manager/cypress/e2e/core/account/account-maintenance.spec.ts
jdamore-linode: @AzureLatte You're right, seeing the same behavior on my end! Taking a look right now!

jdamore-linode: @AzureLatte I'm about 90% sure that this is actually a Cloud Manager bug and not a test/mock issue.

Got a bit distracted by another issue and about to step away for lunch, but I'll follow up here with more info soon! If there's an easy fix it might be nice to just take care of it in this PR, but there's a good chance we'll spin this off into a bug ticket!

jdamore-linode: @AzureLatte Ok, definitely a bug, and pretty easy to reproduce outside of Cypress:

1. Build and serve Cloud via `yarn dev`
2. Using the dev tools in the bottom left, enable Mock Service Worker
3. Navigate to the Maintenance page at `/account/maintenance`
4. Click "Download CSV" for either table, observe empty CSV data (besides column headers)
5. Click "Download CSV" again, observe CSV is populated with data

Not the strongest with React, but my understanding of what's going on in [`MaintenanceTable.tsx`](https://github.com/linode/manager/blob/7952cfc/packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx) is that the CSV data is not automatically fetched when the MaintenanceTable component mounts (that's what the `false` value on line 94 does). That makes sense because this could potentially fetch a lot of data, and doing it automatically wouldn't make sense since it's only used when customers click the "Download CSV" button.

Instead, [it fetches the data when the user clicks the download button](https://github.com/linode/manager/blob/7952cfc/packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx#L138), but because the [CSV data gets passed to the `DownloadCSV` component as a prop](https://github.com/linode/manager/blob/7952cfc/packages/manager/src/features/Account/Maintenance/MaintenanceTable.tsx#L155), it doesn't actually take effect, so to speak, until after the next re-render which presumably happens after the CSV file is downloaded.

I don't know what the proper fix for this would be, so I wrote a pair of tickets:

- M3-8392 for the bug itself
- M3-8393 to test CSV download functionality (it may be helpful to have a test _before_ M3-8392 is worked on so that whoever does the development for that has a test case they can use to confirm their work, but not crucial since this is pretty reproducible)

jdamore-linode (approved):
Nice work @AzureLatte! Approved pending some kind of handling of the Download CSV stuff:

- Remove or comment out what's here and re-add it for M3-8193
- Leave it the way it is and just add the content assertions for M3-8193 (probably the approach that makes the most sense)

Thanks again for all this work! The tests are great and they helped catch a bug! 🐛

packages/manager/cypress/e2e/core/account/account-maintenance.spec.ts
AzureLatte: Thank you, I will add a TODO comment in this PR.

bnussman-akamai (approved):

PR #10695, opened 2024-07-19
test: [M3-8359] - Make Existing Linode Create Cypress Tests work for Linode Create v2

bnussman-akamai:
## Description 📝

- Doing this so Cypress tests still pass when the Linode Create v2 flow is toggled on ✅ 
- Makes Existing Linode Create Cypress Tests work for both Linode Create v1 and v2 🔧 
  - I'm doing this because currently, enabling Linode Create v2 in dev, causes many cypress failures.
  - This shouldn't be the case because Linode Create v2 should in parity with Linode Create v1, but the problem is that some cypress test use selectors that are specific to old components like EnhancedSelect.
  - This PR just updates the tests to use more generic selectors so that **they work for both flows** 
  - Alternatively, we could duplicated these tests, adapt them work for v2, and manually specify the flag to use, but I decided against this because I think it would be a lot more messy and confusing

## How to test 🧪

- There isn't a great way to test this locally unless you manually go into the code and mock the Linode Create v2 feature flag as **enabled**
- I used [this](https://linode.slack.com/archives/CL1L0ML7Q/p1720031512448079) to know what tests to fix. This link will show you what tests failed when we last tried turning on Linode Create v2 in dev

## Before (Linode Create v2 turned **ON** in dev)
Screenshot 2024-07-19 at 1 41 23 AM

## After (Linode Create v2 turned **ON** in dev)

Screenshot 2024-07-19 at 1 38 37 AM

> [!note]
> The remaining GDPR tests should start passing when https://github.com/linode/manager/pull/10692 is merged in

- [ ] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/cypress/e2e/core/images/create-linode-from-image.spec.ts
+++ b/packages/manager/cypress/e2e/core/images/create-linode-from-image.spec.ts
@@ -1,4 +1,4 @@
-import { containsClick, fbtClick, fbtVisible, getClick } from 'support/helpers';
+import { fbtClick, fbtVisible, getClick } from 'support/helpers';
 import { apiMatcher } from 'support/util/intercepts';
 import { randomLabel, randomNumber, randomString } from 'support/util/random';
 import { mockGetAllImages } from 'support/intercepts/images';
@@ -33,14 +33,17 @@ const createLinodeWithImageMock = (url: string, preselectedImage: boolean) => {
   cy.visitWithLogin(url);
 
   cy.wait('@mockImage');
+
   if (!preselectedImage) {
-    cy.get('[data-qa-enhanced-select="Choose an image"]').within(() => {
-      containsClick('Choose an image');
-    });
-    cy.get(`[data-qa-image-select-item="${mockImage.id}"]`).within(() => {
-      cy.get('span').should('have.class', 'fl-tux');
-      fbtClick(mockImage.label);
-    });
+    cy.findByPlaceholderText('Choose an image')
+      .should('be.visible')
+      .should('be.enabled')
+      .click();
+
+    cy.findByText(mockImage.label)
+      .should('be.visible')
+      .should('be.enabled')
+      .click();
   }
 
   ui.regionSelect.find().click();
@@ -49,7 +52,13 @@ const createLinodeWithImageMock = (url: string, preselectedImage: boolean) => {
   fbtClick('Shared CPU');
   getClick('[id="g6-nanode-1"][type="radio"]');
   cy.get('[id="root-password"]').type(randomString(32));
-  getClick('[data-qa-deploy-linode="true"]');
+
+  ui.button
+    .findByTitle('Create Linode')
+    .scrollIntoView()
+    .should('be.visible')
+    .should('be.enabled')
+    .click();
 
   cy.wait('@mockLinodeRequest');
 
--- a/packages/manager/cypress/e2e/core/oneClickApps/one-click-apps.spec.ts
+++ b/packages/manager/cypress/e2e/core/oneClickApps/one-click-apps.spec.ts
@@ -49,14 +49,14 @@ describe('OneClick Apps (OCA)', () => {
       // Check the content of the OCA listing
       cy.findByTestId('one-click-apps-container').within(() => {
         // Check that all sections are present (note: New apps can be empty so not asserting its presence)
-        cy.findByTestId('Popular apps').should('exist');
-        cy.findByTestId('All apps').should('exist');
+        cy.findByText('Popular apps').should('be.visible');
+        cy.findByText('All apps').should('be.visible');
 
         trimmedApps.forEach((stackScript) => {
           const { decodedLabel, label } = handleAppLabel(stackScript);
 
           // Check that every OCA is listed with the correct label
-          cy.get(`[data-qa-select-card-heading="${label}"]`).should('exist');
+          cy.get(`[data-qa-select-card-heading="${label.trim()}"]`).should('exist');
 
           // Check that every OCA has a drawer match
           // This validates the regex in `mapStackScriptLabelToOCA`
@@ -76,7 +76,7 @@ describe('OneClick Apps (OCA)', () => {
       const candidateLabel = handleAppLabel(trimmedApps[0]).label;
 
       const stackScriptCandidate = cy
-        .get(`[data-qa-selection-card-info="${candidateLabel}"]`)
+        .get(`[data-qa-selection-card-info="${candidateLabel.trim()}"]`)
         .first();
       stackScriptCandidate.should('exist').click();
 
@@ -92,7 +92,7 @@ describe('OneClick Apps (OCA)', () => {
       }
 
       ui.drawer
-        .findByTitle(trimmedApps[0].label)
+        .findByTitle(trimmedApps[0].label.trim())
         .should('be.visible')
         .within(() => {
           containsVisible(app.description);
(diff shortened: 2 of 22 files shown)


packages/manager/src/components/EnhancedSelect/components/SelectControl.tsx
```diff
@@ -8,6 +8,7 @@ type Props = ControlProps<any, any>;
 const SelectControl: React.FC<Props> = (props) => {
   return (
     <TextField
+      placeholder={props.selectProps.placeholder}
```
bnussman-akamai, line 11: Fixes `findByPlaceholderText` not working for legacy EnhancedSelect

packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Images.tsx
bnussman-akamai, line 1: Adds missing empty state for parity to fix a test

packages/manager/src/features/Linodes/LinodesCreate/SelectionCardWrapper.tsx
```diff
@@ -77,7 +77,7 @@ export const SelectionCardWrapper = (props: Props) => {
     <InfoGrid xs={2}>
       <Info
         aria-label={`Info for "${label}"`}
-        data-qa-selection-card-info={label}
+        data-qa-selection-card-info={label.trim()}
```
bnussman-akamai, line 80: All of this trimming is unfortunate. Lots of labels have trailing spaces causing the Cypress test matches to fail. Trimming is the easiest fix for now. Once Linode Create v1 is deleted, it will be much easier to clean up the trimming.

packages/manager/src/features/Linodes/index.tsx
bnussman-akamai, line 1: Makes it so that in local dev, the flag will toggle immediately. In production builds, it will not toggle mid-creation so users don't get interrupted.

packages/manager/cypress/e2e/core/images/create-linode-from-image.spec.ts
```diff
@@ -33,14 +33,17 @@ const createLinodeWithImageMock = (url: string, preselectedImage: boolean) => {
   cy.visitWithLogin(url);
 
   cy.wait('@mockImage');
+
   if (!preselectedImage) {
-    cy.get('[data-qa-enhanced-select="Choose an image"]').within(() => {
-      containsClick('Choose an image');
-    });
```
bnussman-akamai, line 39: This is a good summary of this PR's intention. 

`data-qa-enhanced-select` is very rigid and specific to the old EnhancedSelect. Using `cy.findByPlaceholderText` allows this to work for both flows.

packages/manager/cypress/e2e/core/oneClickApps/one-click-apps.spec.ts
```diff
@@ -168,6 +168,7 @@ describe('OneClick Apps (OCA)', () => {
 
     mockGetStackScripts([stackScripts]).as('getStackScripts');
     mockAppendFeatureFlags({
+      linodeCreateRefactor: makeFeatureFlagData(false),
```
bnussman-akamai, line 171: Linode Create v2 handles the "new" section of Marketplace apps differently, which causes this test to fail when Linode Create v2 is enabled. For now, we'll just make this test run on Linode Create v1 exclusively and come back later to adapt it to Linode Create v2

packages/manager/src/features/Linodes/LinodeCreatev2/Security.tsx
```diff
@@ -67,6 +67,7 @@ export const Security = () => {
               label="Root Password"
               name="password"
               noMarginTop
+              id="linode-password"
```
bnussman-akamai, line 70: Out of scope change, but should be harmless. This should fix the bug where Safari scrolls to the top of the page when focusing the Root Password Input.

AzureLatte (approved):
Looks good!

packages/manager/cypress/e2e/core/stackscripts/smoke-community-stackscrips.spec.ts
```diff
@@ -370,15 +369,21 @@ describe('Community Stackscripts integration tests', () => {
       .should('be.visible')
       .should('be.enabled')
       .click();
-    cy.contains('Password does not meet complexity requirements.');
+
+    cy.findByText('Password does not meet', { exact: false }).should(
+      'be.visible'
+    );
```
bnussman-akamai, line 375: Don't love this change, but once we delete Linode Create v1, this could be updated to check for the exact string.

Linode Create v1 says `Password does not meet complexity requirements.`
Linode Create v2 says `Password does not meet strength requirement.`

packages/manager/cypress/e2e/core/images/create-linode-from-image.spec.ts
abailly-akamai: eventually we should add a method to https://github.com/linode/manager/blob/develop/packages/manager/cypress/support/ui/autocomplete.ts for this one

abailly-akamai (approved):
The approach makes sense, there will be some cleanup to do but this is more efficient than duplicating suites/ mocking flags

Left a couple non-blocking quetions/comments

packages/manager/src/features/Linodes/LinodesCreate/SelectionCardWrapper.tsx
bnussman-akamai: The StackScripts that power marketplace have labels like "WordPress One-Click".  There is logic in Cloud Manager that replaces "One-Click" with "", causing the trailing whitespace. 

Ideally, I'd just fix the source of the issue and replace " One-Click" with "", but the Marketplace logic is so messy I think it's best for me to hold off until we can just focus on addressing it when we just need to focus on v2

packages/manager/src/features/Linodes/index.tsx
```diff
@@ -32,7 +32,11 @@ export const LinodesRoutes = () => {
 
   // Hold this feature flag in state so that the user's Linode creation
   // isn't interupted when the flag is toggled.
-  const [isLinodeCreateV2Enabled] = useState(flags.linodeCreateRefactor);
+  const [isLinodeCreateV2EnabledStale] = useState(flags.linodeCreateRefactor);
+
+  const isLinodeCreateV2Enabled = import.meta.env.DEV
+    ? flags.linodeCreateRefactor
+    : isLinodeCreateV2EnabledStale;
```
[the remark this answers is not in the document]
bnussman-akamai: My concern is if we release to prod and find a bug, I'd want to be able to turn off v2, but not interrupt any user that is in progress of creating a Linode.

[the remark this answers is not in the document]
bnussman-akamai: I added the `import.meta.env.DEV` check so that we can keep a nice local dev experience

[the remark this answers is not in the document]
abailly-akamai: Right but imagine there's a bug on v2, we switch the flag off and the user does not refresh the browser - they will keep encountering the same bug. Maybe it's better to force an interruption the one time rather than keeping users in a bad state?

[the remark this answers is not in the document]
abailly-akamai: That's fair, I'll leave the decision to you. I just don't like making assumption on the kind of bugs we can get but all together the chances of switching the flow back are pretty thin.

PR #10683, opened 2024-07-16
test: [M3-8147] - Confirm refactored Linode Create flow with Firewalls attached

AzureLatte:
## Description 📝
Add Cypress integration tests using mock API data to confirm refactored Linode Create flow with Firewalls attached

## Changes  🔄
- Add new test in e2e packages/manager/cypress/e2e/core/linodes/create-linode-with-firewall.spec.ts
- Add new intercepts functions in packages/manager/cypress/support/intercepts/firewalls.ts
- Update TODO comment to confirm toast notification should appear on Linode create.

## How to test 🧪
```
yarn cy:run -s "cypress/e2e/core/linodes/*"
```

## As an Author I have considered 🤔

*Check all that apply*

- [x] 👀 Doing a self review
- [ ] ❔ Our [contribution guidelines](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md)
- [ ] 🤏 Splitting feature into small PRs
- [x] ➕ Adding a changeset
- [ ] 🧪 Providing/Improving test coverage
- [ ] 🔐 Removing all sensitive information from the code and PR description
- [ ] 🚩 Using a feature flag to protect the release
- [ ] 👣 Providing comprehensive reproduction steps
- [ ] 📑 Providing or updating our documentation
- [ ] 🕛 Scheduling a pair reviewing session
- [ ] 📱 Providing mobile support
- [ ] ♿  Providing accessibility support

PR #10698, opened 2024-07-19
fix: Fix Cypress Test Flakes and Failures

bnussman-akamai:
## Description 📝

- `vpc-navigation.spec.ts` & ` smoke-community-stackscrips.spec.ts` & `betas.spec.ts` 
   - Mocks preferences so that the Primary Nav (our sidebar) is open 📂 
  - The test should run faster because preferences are mocked 💨 
- `clone-linode.spec.ts`
  - Fix bug in Linode Create v2 Clone flow causing this test to fail
  - Validation logic was flawed

## Preview 📷

| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/user-attachments/assets/4acc0c61-a993-4575-b9ea-a8f2daeda8fd" /> | <video src="https://github.com/user-attachments/assets/de9e2f8b-4bdb-4b3c-b800-666549f6e949" />  |

## How to test 🧪

- Verify e2es pass ✅ 

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

### DIFF
--- a/packages/manager/cypress/e2e/core/account/betas.spec.ts
+++ b/packages/manager/cypress/e2e/core/account/betas.spec.ts
@@ -8,6 +8,7 @@ import {
 } from 'support/intercepts/feature-flags';
 import { makeFeatureFlagData } from 'support/util/feature-flags';
 import { ui } from 'support/ui';
+import { mockGetUserPreferences } from 'support/intercepts/profile';
 
 // TODO Delete feature flag mocks when feature flag is removed.
 describe('Betas landing page', () => {
@@ -22,8 +23,11 @@ describe('Betas landing page', () => {
     }).as('getFeatureFlags');
     mockGetFeatureFlagClientstream().as('getClientStream');
 
+    // Ensure that the Primary Nav is open
+    mockGetUserPreferences({ desktop_sidebar_open: false }).as('getPreferences');
+
     cy.visitWithLogin('/linodes');
-    cy.wait(['@getFeatureFlags', '@getClientStream']);
+    cy.wait(['@getFeatureFlags', '@getClientStream', '@getPreferences']);
 
     ui.nav.findItemByTitle('Betas').should('be.visible').click();
 
--- a/packages/manager/cypress/e2e/core/stackscripts/smoke-community-stackscrips.spec.ts
+++ b/packages/manager/cypress/e2e/core/stackscripts/smoke-community-stackscrips.spec.ts
@@ -15,6 +15,7 @@ import { Profile } from '@linode/api-v4';
 import { formatDate } from '@src/utilities/formatDate';
 
 import type { StackScript } from '@linode/api-v4';
+import { mockGetUserPreferences } from 'support/intercepts/profile';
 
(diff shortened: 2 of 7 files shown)


packages/manager/cypress/e2e/core/account/betas.spec.ts
```diff
@@ -22,8 +23,11 @@ describe('Betas landing page', () => {
     }).as('getFeatureFlags');
     mockGetFeatureFlagClientstream().as('getClientStream');
 
+    // Ensure that the Primary Nav is open
+    mockGetUserPreferences({ desktop_sidebar_open: false }).as('getPreferences');
```
bnussman-akamai, line 27: It looks crazy, but `false` is actually correct here.`false` makes the sidebar open.  

I'll do follow up PR to fix this globally.

hkhalil-akamai (approved):
Verified this spec consistently passes when tested locally. Thanks for improving our CI!

packages/manager/cypress/e2e/core/stackscripts/smoke-community-stackscrips.spec.ts
```diff
@@ -262,9 +263,11 @@ describe('Community Stackscripts integration tests', () => {
     const region = chooseRegion({ capabilities: ['Vlans'] });
     const linodeLabel = randomLabel();
 
+    // Ensure that the Primary Nav is open
+    mockGetUserPreferences({ desktop_sidebar_open: false }).as('getPreferences');
```
bnussman-akamai, line 267: Also mocking this here because this test checks for the "StackScripts" sidebar item

packages/manager/src/features/Linodes/LinodeCreatev2/schemas.ts
```diff
@@ -1,15 +1,6 @@
...
 import { number, object } from 'yup';
 
-/**
- * Extends the Linode Create schema to make `linode` required for the Clone Linode tab
- */
-export const CreateLinodeByCloningSchema = CreateLinodeSchema.concat(
-  object({
-    linode: object().required('You must select a Linode to clone from.'),
-  })
-);
```
bnussman-akamai, line 11: Oopsies on my part. 

We strip `linode` from the payload before it is validated by the yup schema. Unfortunately, we need to handle this as a special case in the hook-from resolver directly.

packages/manager/cypress/e2e/core/stackscripts/smoke-community-stackscrips.spec.ts
hkhalil-akamai: I can't get these to pass locally (probably the same issue from [this comment](https://github.com/linode/manager/pull/10683#pullrequestreview-2181186323)) but I'm good with the changes if they pass in CI.

mjac0bs (approved):
Thanks for the investigation and fixes - they make sense (well, except where they don't and need the follow up 😆 ). 

I was able to run all three tests locally on `repeat` and noticed no flake or failures.

PR #10599, opened 2024-06-20
fix: [M3-8142] - Alignment Regression Of The IP Address Inside Node Pools Table

carrillo-erik:
## Description 📝
Fixes the the vertical alignment of the IP Address inside of the Node Pools table. 

## Changes  🔄
List any change relevant to the reviewer.
- For testing purposes, I disabled the `background-color: transparent` css property on the `copy` icon in order to show the cause of the visual bug.
- ~~Updated the row width percentage values, allowing the `IP Address` column more room since some addresses will cause the text to wrap. The `Status` column does not need as much room and the values in these change maintain a similar proportion as before.~~
- Pass the `noWrap` prop set to `true` on the `<TableCell>` for the IP Address column to prevent longer addresses from wrapping the `copy` icon and resulting in vertical alignment discrepancies. 

## Preview 📷
| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/linode/manager/assets/119514965/b55cb5bf-018a-4a17-a537-599f8ca20d65" /> | <video src="https://github.com/linode/manager/assets/119514965/10ac1227-7686-43dc-8977-9c6d904eb2de" /> |

## How to test 🧪
- Open the web browser developer tools and modify the viewport width as shown in the `Preview` video to ensure consistent results. 

### Verification steps
(How to verify changes)
- Verify that as you adjust the viewport width, the IP Address which are wider than normal do not cause the copy icon to wrap to the next line (causing the IP Address to shift vertically).

## As an Author I have considered 🤔
*Check all that apply*

- [x] 👀 Doing a self review
- [ ] ❔ Our [contribution guidelines](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md)
- [ ] 🤏 Splitting feature into small PRs
- [x] ➕ Adding a changeset
- [ ] 🧪 Providing/Improving test coverage
- [ ] 🔐 Removing all sensitive information from the code and PR description
- [ ] 🚩 Using a feature flag to protect the release
- [ ] 👣 Providing comprehensive reproduction steps
- [ ] 📑 Providing or updating our documentation
- [ ] 🕛 Scheduling a pair reviewing session
- [ ] 📱 Providing mobile support
- [ ] ♿  Providing accessibility support

PR #10699, opened 2024-07-22
upcoming: [M3-8364] - Use React Hook Form instead of Formik for Create Bucket

jaalah-akamai:
## Description 📝
Switching from Formik to React Hook Form before updating the drawer with the new UI / API requirements.

## Changes  🔄
- Using `isFeatureEnabledV2` instead of deprecated `isFeatureEnabled`
- Unique label validation is happening in the schema now
- Reset of values is now happening `onExited` 

## Target release date 🗓️
N/A

## Preview 📷
N/A

## How to test 🧪

>[!note]
Ensure you test both legacy and multi-cluster (gen1) by toggling flag

### Verification steps
- Create a new bucket and ensure the creation process still works
- Try creating a bucket with a label that's already used and ensure the validation works
   - You should see an error with helper text under label input field 
- The bucket creation form should be more responsive, with real-time validation feedback
- Form fields should validate input as the user types, not just on submission
- The form should prevent submission if there are any validation errors

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset
- [  ] 🧪 Providing/Improving test coverage


packages/manager/src/features/ObjectStorage/BucketLanding/ClusterSelect.tsx
```diff
@@ -11,7 +11,7 @@ interface Props {
   onBlur: (e: any) => void;
   onChange: (value: string) => void;
   required?: boolean;
-  selectedCluster: string;
+  selectedCluster: string | undefined;
```
jaalah-akamai, line 14: The RegionSelect for which this is a wrapper has the `value` as `string | undefined` which I believe to be correct.

cpathipa (approved):
Create Bucket functionality is working as expected for Gen1 and legacy. 

- [x] Able to create a new bucket for Gen1 and legacy.
- [x] Validation is working as expected for label that's already used.
- [x] Confirming on  error with helper text under label input field
- [x] - Confirming form prevents submission if there are any validation errors.

**Note:** Since we are using V2 flag, we should enable the flag in LD in respective environment when this is shipped to production.

packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
```diff
@@ -109,121 +108,134 @@ export const CreateBucketDrawer = (props: Props) => {
...
-    initialValues: {
+  const {
+    control,
+    formState: { errors },
+    handleSubmit,
+    reset,
+    setError,
+    watch,
+  } = useForm<CreateObjectStorageBucketPayload>({
+    context: { buckets: bucketsData?.buckets ?? [] },
```
bnussman-akamai, line 119: Neat. Did not know about `context`

bnussman-akamai (commented):
On an account without Object Storage enabled, I wasn't able to enable Object Storage

https://github.com/user-attachments/assets/cfbc11dc-520c-4d51-b11e-7c080182bcc5

jaalah-akamai:
I need to fix the OBJ test since I'm now disabling `Create Bucket` unless they provide a label and a region, and the tests aren't accounting for that. 

@hana-linode The clone-linode.spec.ts is looking for the old region format, surprised this wasn't caught in the PR where we changed this? 🤔 

https://github.com/user-attachments/assets/2a9263b4-c90f-4497-9cc2-ef5ddf255eae

carrillo-erik:
I still need to test this with the multi-cluster feature flag turned on. The legacy experience worked as expected and did not see any visual regressions. Nothing in the code stood out on first inspection.

AzureLatte (approved):
Looks good!

hana-linode:
> The clone-linode.spec.ts is looking for the old region format, surprised this wasn't caught in the PR where we changed this? 🤔

@jaalah tests were not updated since we're still going to display the old format for now since Gecko LA/GA got pushed 2 months (new format is feature flagged under Gecko GA). The tests passed on my PR 🤔

carrillo-erik:
> > The clone-linode.spec.ts is looking for the old region format, surprised this wasn't caught in the PR where we changed this? 🤔
> 
> @jaalah tests were not updated since we're still going to display the old format for now since Gecko LA/GA got pushed 2 months (new format is feature flagged under Gecko GA). The tests passed on my PR 🤔

@jaalah-akamai @hana-linode 
I experienced the region label issue in the `clone-linode.spec.ts` when I tried to merge [M3-8142](https://github.com/linode/manager/pull/10599). In order to have the tests pass without updating the tests, I had to comment out the `gecko2` feature flag from the `useFlags.ts` file. 
```
  return {
    ...flags,
    ...mockFlags,
    // gecko2: {
    //   enabled: true,
    //   ga: true,
    // },
  };
```

jaalah-akamai:
@carrillo-erik Yea this issue is stemming from https://github.com/linode/manager/pull/10702. We are looking into a patch to fix this without having to revert the work. Until then if you need to get your tests passing, you can disable the `gecko2` flag in your mock test. Something like this should work
```
mockAppendFeatureFlags({
   gecko2: makeFeatureFlagData(false),
}).as('getFeatureFlags');
```

packages/manager/cypress/e2e/core/objectStorage/enable-object-storage.spec.ts
```diff
@@ -149,6 +149,8 @@ describe('Object Storage enrollment', () => {
       .findByTitle('Create Bucket')
       .should('be.visible')
       .within(() => {
+        cy.get('[data-testid="label"] input').click().type(randomLabel());
```
bnussman-akamai, line 152: Would `findByLabelText` work?

packages/validation/src/buckets.schema.ts
```diff
@@ -9,7 +9,26 @@ export const CreateBucketSchema = object()
...
+              return true;
+            }
+
+            return !buckets.some(
+              (bucket) =>
+                bucket.label === value &&
+                (bucket.cluster === cluster || bucket.region === region)
+            );
+          }
+        ),
```
bnussman-akamai, line 31: In my opinion, I think we should just remove this logic and rely on API errors for uniqueness, but this seems fine!

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
```diff
@@ -195,12 +216,12 @@ export const OMC_CreateBucketDrawer = (props: Props) => {
           primaryButtonProps={{
             'data-testid': 'create-bucket-button',
             disabled:
-              !formik.values.region ||
+              !watchLabel ||
+              !watchRegion ||
```
bnussman-akamai, line 220: From previous cafe discussion, UX decided that we should to keep form submit buttons enabled globally so that the user can submit at any time and receive validation errors.

Not a huge deal here because this is a small form, but this will diverge from our pattern.

bnussman-akamai (approved):

carrillo-erik (approved):

carrillo-erik:
Tested and verified all workflows related to these changes. In addition, all tests are now ✅.

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
jaalah-akamai: Ok no problem, I might have forgotten we talked about this 👍

packages/validation/src/buckets.schema.ts
jaalah-akamai: Hmm, I haven't looked too deeply into this yet, but it seems that no API error is caught in the case of a duplicate. Instead, it silently reverts the optimistic UI rendering of the bucket or something similar. I'll investigate this further and try to address it in my next PR. For now, I think the validation is useful. 👍

bnussman-akamai: Ohh, got it! Thanks for the investigation 🙏

packages/manager/cypress/e2e/core/objectStorage/enable-object-storage.spec.ts
jaalah-akamai: You would think `cy.findByLabelText('Label').should('be.visible').type(randomLabel());` would work, but it doesn't. Even when it's right there in the output 🤔 
```
Running:  enable-object-storage.spec.ts                                                   (1 of 1)
Browser information:
┌─────────────────┬──────────────────────────┐
│     (index)     │          Values          │
├─────────────────┼──────────────────────────┤
│     Browser     │ 'Chrome v126.0.6478.183' │
│     Family      │        'chromium'        │
│    Headless     │          'Yes'           │
│ Chrome Headless │          'Old'           │
└─────────────────┴──────────────────────────┘


  Object Storage enrollment
    1) can enroll in Object Storage
    ✓ can cancel Object Storage (3200ms)


  1 passing (1m)
  1 failing

  1) Object Storage enrollment
       can enroll in Object Storage:
     AssertionError: Timed out retrying after 80000ms: Unable to find a label with the text of: Label

Ignored nodes: comments, script, style
      
      { ... More Output Above ... }

      <div
        class="MuiBox-root css-0"
      >
        <div
          class="MuiBox-root css-15ybjgl"
          data-testid="inputLabelWrapper"
        >
          <label
            class="MuiFormLabel-root MuiInputLabel-root MuiInputLabel-animated MuiFormLabel-colorPrimary MuiInputLabel-root MuiInputLabel-animated css-17y63dv-MuiFormLabel-root-MuiInputLabel-root-noTransform"
            data-qa-textfield-label="Label"
            for="label"
          >
            Label
            <span
              class="css-1pz1r01-label"
            >
               (required)
            </span>
          </label>
        </div>
        <div
          class=""
        >
          <div
            class="MuiFormControl-root MuiFormControl-fullWidth MuiTextField-root css-1ccy8ut-MuiFormControl-root-MuiTextField-root-root"
            data-qa-cluster-label="true"
            data-testid="label"
          >
            <div
              class="MuiInputBase-root MuiInput-root MuiInputBase-colorPrimary MuiInputBase-fullWidth MuiInputBase-formControl input css-1jeecg1-MuiInputBase-root-MuiInput-root"
            >
              <input
                aria-invalid="false"
                class="MuiInputBase-input MuiInput-input css-1hp0d5q-MuiInputBase-input-MuiInput-input"
                data-testid="textfield-input"
                id="label"
                name="label"
                type="text"
                value=""
              />
            </div>
          </div>
        </div>
      </div>
            
      { ... More Output Below ... }
```

bnussman-akamai: `findByLabelText("Label (required)")` might work, but if not, no worries!

jaalah-akamai: Yup that was it 🔥

jaalah-akamai:
This appears to be a flaky e2e test looking at `detail/PR-10699/9/pipeline`.

Screenshot 2024-07-26 at 9 13 00 PM

PR #10702, opened 2024-07-23
upcoming: [M3-8209] - Region label updates

hana-linode:
## Description 📝
When the Gecko GA flag is true:
- Display Country, City instead of City, State anywhere there is a region label
- Ensure Linode Create v2 displays Country, City

I decided to keep the region label updates feature flagged for now since it is a bit of an update, LA/GA has been delayed further, and to avoid confusion with the patching csv list, etc.

## Preview 📷
| Before  | After   |
| ------- | ------- |
| image | image |

### Prerequisites

### Verification steps
(How to verify changes)
- Check region labels throughout the app
- Regions should display as Country, City with the gecko flag on and as normal with the flag off

*Check all that apply*

- [ ] 👀 Doing a self review
- [x] ➕ Adding a changeset


mjac0bs (approved):
I spot-checked around the app, including most places we're using the `useRegionsQuery` hook without the transformRegionLabel option to confirm we didn't need it. Everything was looking consistent with the new updates.

I just restarted the tests in CI because they didn't kick off right, so approving pending those pass/

hkhalil-akamai (approved):
Nice work! Changes look good across the app. I wonder if we should just make `transformRegionLabel` true by default?

PR #10704, opened 2024-07-23
upcoming: Minor Improvements to Linode Create v2

bnussman-akamai:
## Description 📝

Makes some minor changes to Linode Create v2 to improve the code and make the from work better 🔧  

## Changes  🔄
- Uses `setValue` instead of `reset` to handle side effect state updates 🔄
  - React hook form does not have a built in way to set many form values at once 😩
  - Originally, I used `reset` to set many values at once when the region changes, but `reset` has causes many unintended side effects so I am migrating to `setValue`
- Uses React Query in a few places where the raw api-v4 function was called 🌎 

## Preview 📷

| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/user-attachments/assets/0f74aec0-098b-4389-b8ab-3f158340c292" /> | <video src="https://github.com/user-attachments/assets/8541753e-c570-4dad-8858-7894ba7c53ac" /> |
| Notice how the validation error does not go away when you select a Linode Plan 👎 This is happening because `reset` causes the validation behavior to change back to `onBlur` when the region is changed. We want the from to remain in "revalidationMode"  | Notice how the validation error does go away when you choose a plan ✅  |

### Prerequisites
- Ensure the Linode Create v2 feature flag is toggled on 

### Verification steps
- Test general functionality on Linode Create v2
- Verify that region specific things like VLANs, VPC, and Placement Groups properly reset if the region is changed
- Looks at logic in `LinodeCreatev2/Region.tsx` and ensure it makes sense
- Verify the Linode payload and summary are correct with various options selected
- Verify the dynamic validation works as expected
  - Expected behavior is:
    - Validate onBlur before first submit
    - Validate onChange after first submit

- [ ] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/features/Linodes/LinodeCreatev2/Region.tsx
```diff
@@ -80,53 +79,62 @@ export const Region = () => {
...
+      values.metadata?.user_data &&
+      !region.capabilities.includes('Metadata')
+    ) {
+      // Clear metadata only if the new region does not support it
+      setValue('metadata.user_data', null);
+    }
+
+    if (values.placement_group?.id) {
+      // If a placement group is selected, clear it because they are region sepecific
+      setValue('placement_group.id', 0);
```
dwiley-akamai, line 108: Do we have to use `0` instead of `null` here?

dwiley-akamai (approved):
Code review ✅
Tested a variety of option combinations to check validation & ability to create ✅

packages/manager/src/features/Linodes/LinodeCreatev2/Region.tsx
bnussman-akamai: Good question!

Using `0` here because the type `CreateLinodeRequest` does not allow `null`. It does get a bit deeper though.

Ideally we'd do something like `setValue('placement_group', undefined);` rather than `setValue('placement_group.id', 0);`, but react-hook-form has some [weird behavior](https://github.com/react-hook-form/react-hook-form/issues/7758#issuecomment-1030812564) around update nested objects. That's why I'm having to `setValue` of `placement_group.id` rather than the entire `placement_group` object. 

The reason why this is okay here is because our "post submit transformation"  function `getLinodeCreatePayload` removes the entire `placement_group` object if `placement_group.id` is falsy, hence why I used `0` here.

I don't love that this is the case, but I think it's the best solution for now

packages/manager/src/features/Linodes/LinodeCreatev2/Region.tsx
```diff
@@ -80,53 +79,62 @@ export const Region = () => {
...
-        keepSubmitCount: true,
-        keepTouched: true,
-      }
-    );
+    const values = getValues();
+
+    field.onChange(region.id);
+
+    if (values.interfaces?.[0].vpc_id) {
+      // If a VPC is selected, clear it because VPCs are region sepecific
```
mjac0bs, line 87:
```suggestion
      // If a VPC is selected, clear it because VPCs are region specific
```

packages/manager/src/features/Linodes/LinodeCreatev2/Region.tsx
```diff
@@ -80,53 +79,62 @@ export const Region = () => {
...
+      setValue('interfaces.0.vpc_id', null);
+      setValue('interfaces.0.subnet_id', null);
+    }
+
+    if (values.interfaces?.[1].label) {
+      // If a VLAN is selected, clear it because VLANs are region sepecific
```
mjac0bs, line 93:
```suggestion
      // If a VLAN is selected, clear it because VLANs are region specific
```

packages/manager/src/features/Linodes/LinodeCreatev2/Region.tsx
```diff
@@ -80,53 +79,62 @@ export const Region = () => {
...
+    if (
+      values.metadata?.user_data &&
+      !region.capabilities.includes('Metadata')
+    ) {
+      // Clear metadata only if the new region does not support it
+      setValue('metadata.user_data', null);
+    }
+
+    if (values.placement_group?.id) {
+      // If a placement group is selected, clear it because they are region sepecific
```
mjac0bs, line 107:
```suggestion
      // If a placement group is selected, clear it because they are region specific
```

packages/manager/src/features/Linodes/LinodeCreatev2/Region.tsx
```diff
@@ -80,53 +79,62 @@ export const Region = () => {
...
-      'Disk Encryption'
-    )
-      ? 'enabled'
-      : undefined;
-
-    reset(
-      (prev) => ({
-        ...prev,
-        // reset EU agreement
-        hasSignedEUAgreement: undefined,
```
mjac0bs, line 96: Is my understanding correct that we're good to exclude this because if the user has signed the agreement with one EU region selection, they've really agreed for all regions in the EU?

mjac0bs (approved):
This form and its error handling seemed to be behaving as expected from my testing and triggering of various validation errors. The region-dependent fields reset, and the logic makes sense in Region.tsx.

I think we should be good here with form analytics as well, since fields that have their value set will be considered "touched" and the thing that we wanted to avoid with `reset` was touched being cleared.

packages/manager/src/features/Linodes/LinodeCreatev2/Region.tsx
bnussman-akamai: Yeah, that's what I was thinking. 

By not resetting it, if the user checks the EU checkbox, then changes region, they won't have to re-check it. 

We could reset `hasSignedEUAgreement` so that the user **would** have to re-check if they change regions. 

Any thoughts on what behavior is best?

mjac0bs: Not having to recheck = user is not slowed down with the extra selection.
Forced to recheck = user is able to read the helper text and click the link in the text (not sure if users really do that anyway?)

I honestly don't have a strong opinion and not having to recheck is probably optimal for most users. (But personally, as a user who doesn't like things hidden from me, I'd prefer to recheck it.)

bnussman-akamai: I'm learning towards re-check. I'll update the PR to reflect that.

PR #10717, opened 2024-07-25
hotfix: Distributed Images Incorrect Capability

bnussman-akamai:
## Description 📝

- Turns out we had the wrong Image capability for distributed images 💾 🌎 
  - The capability should be `distributed-sites`, **not** `distributed-images`.
- Adds change-log entries and bumps versions of `manager` and `api-v4`📝 ⬆️  

## How to test 🧪

- As far as I know, there is no environment where this can be easily tested 😞 
- Test Image Service Gen2 features with the MSW 👁️  and verify automated testing passes ✅

### DIFF
--- a/packages/api-v4/CHANGELOG.md
+++ b/packages/api-v4/CHANGELOG.md
@@ -1,3 +1,9 @@
+## [2024-07-25] - v0.122.1
+
+### Fixed:
+
+- Incorrect Image capability of `distributed-images` by changing it to `distributed-sites` ([#10717](https://github.com/linode/manager/pull/10717))
+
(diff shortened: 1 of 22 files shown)


jdamore-linode (dismissed):
Thanks @bnussman-akamai!

Code changes all look good, Image tests are all passing locally, and couldn't find any other stray instances of `'distributed-images'` in the codebase. Additionally checked the behavior of the test in `manage-image-regions.spec.ts` with the mock set back to `'distributed-images'` to confirm that Cloud does _not_ recognize that capability string (test failed as expected). Approved pending CI 👍

jaalah-akamai (dismissed):

PR #10720, opened 2024-07-27
upcoming: [M3-8350] - Obj Gen2 - MSW, Factories, Cypress Updates

jaalah-akamai:
## Description 📝
Laying some more groundwork

## Changes  🔄
- The `setUpBucket` and `setUpBucketMulticluster` functions should use the `createObjectStorageBucketFactory` factory instead of `objectStorageBucketFactory`, which is the mocked response for the create request.
- Added new intercept `mockGetObjectStorageEndpoints` which is the new endpoint we're adding for this project.
- Added new factory `getObjectStorageEndpointsFactory`
- Added new MSW endpoint for `/v4/object-storage/endpoints`
- Added `s3_endpoint` and `endpoint_type` which are all optional properties to some factories

## Target release date 🗓️
N/A

## Preview 📷
N/A

### Verification steps
- Ensure existing e2e tests all pass.

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/cypress/support/intercepts/object-storage.ts
+++ b/packages/manager/cypress/support/intercepts/object-storage.ts
@@ -9,6 +9,7 @@ import { paginateResponse } from 'support/util/paginate';
 import { makeResponse } from 'support/util/response';
 
 import type {
+  CreateObjectStorageBucketPayload,
   ObjectStorageBucket,
   ObjectStorageCluster,
   ObjectStorageKey,
@@ -86,7 +87,7 @@ export const interceptCreateBucket = (): Cypress.Chainable<null> => {
  * @returns Cypress chainable.
  */
 export const mockCreateBucket = (
-  bucket: ObjectStorageBucket
+  bucket: CreateObjectStorageBucketPayload
 ): Cypress.Chainable<null> => {
   return cy.intercept(
     'POST',
@@ -477,4 +478,13 @@ export const interceptUpdateBucketAccess = (
     'PUT',
     apiMatcher(`object-storage/buckets/${cluster}/${label}/access`)
   );
+};
+
+/**
+ * Intercepts GET request to get object storage endpoints.
+ *
+ * @returns Cypress chainable.
+ */
+export const mockGetObjectStorageEndpoints = (): Cypress.Chainable<null> => {
+  return cy.intercept('GET', apiMatcher(`object-storage/endpoints`));
 };
--- a/packages/manager/cypress/support/intercepts/object-storage.ts
+++ b/packages/manager/cypress/support/intercepts/object-storage.ts
@@ -485,6 +485,6 @@ export const interceptUpdateBucketAccess = (
  *
  * @returns Cypress chainable.
  */
-export const mockGetObjectStorageEndpoints = (): Cypress.Chainable<null> => {
+export const interceptGetObjectStorageEndpoints = (): Cypress.Chainable<null> => {
   return cy.intercept('GET', apiMatcher(`object-storage/endpoints`));
 };
--- a/packages/manager/src/factories/objectStorage.ts
+++ b/packages/manager/src/factories/objectStorage.ts
@@ -1,9 +1,10 @@
 import Factory from 'src/factories/factoryProxy';
 
 import type {
-  ObjectStorageBucket,
   CreateObjectStorageBucketPayload,
+  ObjectStorageBucket,
   ObjectStorageCluster,
+  ObjectStorageEndpoint,
   ObjectStorageKey,
   ObjectStorageObject,
 } from '@linode/api-v4/lib/object-storage/types';
@@ -12,12 +13,14 @@ export const objectStorageBucketFactory = Factory.Sync.makeFactory<ObjectStorage
   {
     cluster: 'us-east-1',
     created: '2019-12-12T00:00:00',
+    endpoint_type: 'E1',
     hostname: Factory.each(
       (i) => `obj-bucket-${i}.us-east-1.linodeobjects.com`
     ),
     label: Factory.each((i) => `obj-bucket-${i}`),
     objects: 103,
     region: 'us-east',
+    s3_endpoint: 'us-east-1.linodeobjects.com',
     size: 999999,
   }
 );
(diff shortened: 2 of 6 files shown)


packages/manager/cypress/support/intercepts/object-storage.ts
```diff
@@ -478,3 +479,12 @@ export const interceptUpdateBucketAccess = (
     apiMatcher(`object-storage/buckets/${cluster}/${label}/access`)
   );
 };
+
+/**
+ * Intercepts GET request to get object storage endpoints.
+ *
+ * @returns Cypress chainable.
+ */
+export const mockGetObjectStorageEndpoints = (): Cypress.Chainable<null> => {
```
jdamore-linode, line 488:
```suggestion
export const interceptGetObjectStorageEndpoints = (): Cypress.Chainable<null> => {
```
(Since this util only intercepts the outgoing request and doesn't mock the response)

packages/manager/src/factories/objectStorage.ts
```diff
@@ -12,12 +13,14 @@ export const objectStorageBucketFactory = Factory.Sync.makeFactory<ObjectStorage
...
     created: '2019-12-12T00:00:00',
+    endpoint_type: 'E1',
     hostname: Factory.each(
       (i) => `obj-bucket-${i}.us-east-1.linodeobjects.com`
     ),
     label: Factory.each((i) => `obj-bucket-${i}`),
     objects: 103,
     region: 'us-east',
+    s3_endpoint: 'us-east-1.linodeobjects.com',
     size: 999999,
```
jdamore-linode, line 24: Any thoughts on having separate factories for Gen1/Gen2 Object Storage Buckets (and for any other types where we're adding/changing properties specific to Gen2 but which don't apply to existing operations)?

These new properties will be applied everywhere we're already using this factory, so we'll probably have to go through and manually set these properties back to `undefined` for all the existing uses of the factories. Since we have to maintain both OBJ implementations anyway it might be nice to have factories for each and save us the trouble.

jdamore-linode (commented):
Looks good so far!

Posted a quick suggestion for the intercept util, but more curious what your thoughts are about the types and factories. I think having more specialized factories (i.e. a factory for each OBJ implementation) could help us avoid some of the issues we had with OBJ multicluster

packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -836,6 +837,10 @@ export const handlers = [
     ];
     return HttpResponse.json(makeResourcePage(objectStorageTypes));
   }),
+  http.get('*/v4/object-storage/endpoints', ({}) => {
```
mjac0bs, line 840:
```suggestion
  http.get('*/v4beta/object-storage/endpoints', ({}) => {
```
My understanding is that this is a new endpoint, and new endpoints should be `v4beta` rather than `v4` - is that correct here?

packages/manager/src/factories/objectStorage.ts
```diff
@@ -77,3 +82,11 @@ export const makeObjectsPage = (
 });
 
 export const staticObjects = objectStorageObjectFactory.buildList(250);
+
+export const getObjectStorageEndpointsFactory = Factory.Sync.makeFactory<ObjectStorageEndpoint>(
```
mjac0bs, line 86: Were you thinking it adds clarity to include `get` here? Our factories for get requests don't generally include the verb.

```suggestion
export const objectStorageEndpointsFactory = Factory.Sync.makeFactory<ObjectStorageEndpoint>(
```

jaalah-akamai: Typically I like to use the verbs for POST/PUT, not necessarily for GET - Good catch

packages/manager/src/factories/objectStorage.ts
jaalah-akamai: Yes, let me add those in 👍

packages/manager/src/mocks/serverHandlers.ts
jaalah-akamai: As of right now, API has implemented this using `v4`, but I'm waiting for final reply from API/Architect - this could be a small amendment in a follow-up if I don't get an answer quickly

packages/manager/src/factories/objectStorage.ts
jaalah-akamai: I've split out the factories. For instance `createObjectStorageBucketFactory` is now broken up as:
- `createObjectStorageBucketFactoryLegacy`
- `createObjectStorageBucketFactoryGen1`
- `createObjectStorageBucketFactoryGen2`

packages/manager/src/mocks/serverHandlers.ts
jaalah-akamai: No need to prefix this with `v4beta` since the Beta is currently in progress and not integrated with the linode API at all. This will be the endpoint for GA.

packages/manager/src/factories/objectStorage.ts
```diff
@@ -22,11 +23,46 @@ export const objectStorageBucketFactory = Factory.Sync.makeFactory<ObjectStorage
...
+    label: Factory.each((i) => `obj-bucket-${i}`),
+    region: 'us-east-1',
+  }
+);
+
+// TODO: OBJ Gen2 - Once we eliminate legacy and Gen1 support, we can rename this to `createObjectStorageBucketFactory` and set it as the default.
+export const createObjectStorageBucketFactoryGen2 = Factory.Sync.makeFactory<CreateObjectStorageBucketPayload>(
+  {
+    acl: 'private',
+    cors_enabled: true,
```
mjac0bs, line 65: `"@param cors_enabled - Enable CORS on the bucket: defaults to true for Gen1 and false for Gen2."`

Should `cors_enabled` be `false` here, then?

mjac0bs (approved):
Thanks for checking on the endpoint. The latest changes look good, assuming CI finishes and passes.

AzureLatte (approved):
Looks good

jdamore-linode (approved):
Looks good, thanks Jaalah!

hana-linode (approved):

PR #10723, opened 2024-07-29
upcoming: [M3-8401] - Add support for Two-step region select in Linode Create v2

hana-linode:
## Description 📝
Add support for the two-step Gecko GA region select in the Linode Create v2 flow

## Changes  🔄
- Add `TwoStepRegion.tsx` for Linode Create v2

## Preview 📷
| Before  | After   |
| ------- | ------- |
| image | image |

### Prerequisites

### Verification steps
(How to verify changes)
- There should be no visual or functional differences between v1 and v2 Create flows

```
yarn test TwoStepRegion
```

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/Linodes/LinodeCreatev2/index.tsx
+++ b/packages/manager/src/features/Linodes/LinodeCreatev2/index.tsx
@@ -7,10 +7,6 @@ import { useHistory } from 'react-router-dom';
 
 import { DocumentTitleSegment } from 'src/components/DocumentTitle';
 import { LandingHeader } from 'src/components/LandingHeader';
-import {
-  isDistributedRegionSupported,
-  useIsGeckoEnabled,
-} from 'src/components/RegionSelect/RegionSelect.utils';
 import { Stack } from 'src/components/Stack';
 import { SafeTabPanel } from 'src/components/Tabs/SafeTabPanel';
 import { Tab } from 'src/components/Tabs/Tab';
@@ -42,7 +38,6 @@ import { Images } from './Tabs/Images';
 import { Marketplace } from './Tabs/Marketplace/Marketplace';
 import { OperatingSystems } from './Tabs/OperatingSystems';
 import { StackScripts } from './Tabs/StackScripts/StackScripts';
-import { TwoStepRegion } from './TwoStepRegion';
 import { UserData } from './UserData/UserData';
 import {
   captureLinodeCreateAnalyticsEvent,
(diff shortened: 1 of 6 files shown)


packages/manager/src/features/Linodes/LinodeCreatev2/index.tsx
```diff
@@ -180,7 +189,8 @@ export const LinodeCreatev2 = () => {
               </SafeTabPanel>
             </TabPanels>
           </Tabs>
-          {params.type !== 'Backups' && <Region />}
+          {params.type !== 'Backups' && !showTwoStepRegion && <Region />}
+          {params.type !== 'Backups' && showTwoStepRegion && <TwoStepRegion />}
```
bnussman-akamai, line 193: Rather than having two separate Region components and duplicating logic, could we just conditionally render the `TwoStepRegionSelect` inside the existing `Region.tsx` file?

AzureLatte (approved):
Test passed

packages/manager/src/features/Linodes/LinodeCreatev2/index.tsx
bnussman-akamai: I was thinking more along the lines of how `SelectRegionPanel.tsx` does it 

```typescript
{isGeckoGAEnabled && isDistributedRegionSupported(params.type) ? <TwoStepRegionSelect /> : <RegionSelect />}
```

Would this be possible?

hana-linode: @bnussman-akamai Ahh gotcha, addressed in [540e2d3](https://github.com/linode/manager/pull/10723/commits/540e2d3deece41e1767cd4ef39a97bdce5a80b00). I went for the early return instead since the clone warning notices doesn't position nicely by just in-line replacing

bnussman-akamai: Hm. Could we just reposition the clone warning notices? I'd really prefer if we could accomplish this without the need for `TwoStepRegion.tsx`

hana-linode: @bnussman-akamai It's a bit tricky since the two-step is inside a tab panel so unless we add conditional logic there, the notices will still be "outside". Cloning is out of scope for this PR as it's not currently supported in v1 either.

I think keeping the logic in `TwoStepRegion` is cleaner than adding ~100 lines to `RegionSelect`. Anyone else feel strongly about this?

bnussman-akamai (approved):

jaalah-akamai (approved):

PR #10724, opened 2024-07-29
fix: [M3-8219] - Firewall rule sources not displaying correctly

hkhalil-akamai:
## Description 📝
Fixes a bug in `FirewallRuleDrawer` causing some addresses not to appear.

## Target release date 🗓️
8/5

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-07-29 at 5 29 03 PM  |Screenshot 2024-07-29 at 5 29 49 PM  |

### Reproduction steps
- Create a new rule in a new/existing firewall
- Enter any value for "Label", "Protocol" and "Ports"
- For sources, select "IP / Netmask"
- For the IP addresses, either
    - Add all IPv4 addresses (0.0.0.0/0) and some IPv6 addresses or an IPv6 subnet
    - Add all IPv6 addresses (::/0) and some IPv4 addresses or an IPv4 subnet
- Click "Add Rule" then "Save Changes"
- Edit the new rule that was just created
- Observe "Sources" is set to "All IPv4" or "All IPv6" instead of "IP / Netmask" as expected

## Verification steps
- Follow the same steps as above and verify "Sources" is set to "IP / Netmask" and displays all of the configured IP addresses

### DIFF
--- a/packages/manager/.changeset/pr-10724-fixed-1722289139869.md
+++ b/packages/manager/.changeset/pr-10724-fixed-1722289139869.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Fixed
+---
+
+Sources not displaying correctly in Firewall Rule drawer ([#10724](https://github.com/linode/manager/pull/10724))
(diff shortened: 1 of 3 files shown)


jdamore-linode (approved):
Thanks @hkhalil-akamai! Was able to reproduce the issue, and confirmed that it's fixed with your changes

hana-linode (approved):

mjac0bs (approved):
Thanks for the fix. Reproduced in prod and then confirmed the sources are displaying accurately when editing a firewall rule with some dummy IPs:

✅  All IPv4 + some IPv6:
Screenshot 2024-07-31 at 8 33 33 AM

✅  All IPv6 + some IPv4:
Screenshot 2024-07-31 at 8 34 45 AM

PR #10726, opened 2024-07-30
refactor: [M3-7438] - Query Key Factory for Object Storage

bnussman-akamai:
## Description 📝

- Uses a Query Key Factory for Object Storage related queries 🏭 
- Fixed deprecations in preparation for TanStack Query v5 ([docs](https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5)) 🔧 

## How to test 🧪

> [!note]
> Object Storage should still work with the `OBJ Multi-Cluster` flag both **on and off** 🏴 

- Test all Object Storage functionality in Cloud Manager 👀 
- Verify all automated testing for Object Storage pass ✅ 

- [ ] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/components/PrimaryNav/PrimaryNav.tsx
```diff
@@ -93,68 +87,20 @@ export const PrimaryNav = (props: PrimaryNavProps) => {
...
-      : undefined,
-  });
-
-  const allowObjPrefetch =
-    !buckets &&
-    !clusters &&
-    !clustersLoading &&
-    !bucketsLoading &&
-    !clustersError &&
-    !bucketsError;
```
jaalah-akamai, line 156: Just leaving a comment here for others: Most of the logic that was in this file has been moved into the query now

jaalah-akamai:
Thanks for tackling this before I get more involved with Gen2 changes 🔥. Initial code changes look good and the tests pass locally for me despite what the CI is saying. I will manually test things tomorrow. 

cc: @cpathipa

packages/manager/src/features/ObjectStorage/AccessKeyLanding/AccessKeyDrawer.tsx
```diff
@@ -94,43 +87,12 @@ export const AccessKeyDrawer = (props: AccessKeyDrawerProps) => {
...
     error: bucketsError,
     isLoading: areBucketsLoading,
-  } = useObjectStorageBuckets({
-    clusters: isObjMultiClusterEnabled ? undefined : objectStorageClusters,
-    enabled: true,
-    isObjMultiClusterEnabled,
-    regions: isObjMultiClusterEnabled
-      ? regionsSupportingObjectStorage
-      : undefined,
-  });
```
jaalah-akamai, line 133: Just leaving a comment here for others: Most of the logic that was in this file (and others where we use `useObjectStorageBuckets`) has been moved into the query now.

packages/manager/src/components/PrimaryNav/PrimaryNav.tsx
bnussman-akamai: Yeah. I removed the prefetching in this PR because I don't believe the optimization is worth it for the very slight user experience improvement. 

Happy to add it back if we think it should be kept.

jaalah-akamai (approved):
For Legacy / Gen1
✅ Bucket Creation 
✅ Access Keys
✅ Bucket Access
✅ Object Uploads / Folder Creation

Nice work - didn't notice anything broken from the flow and this is needed for Gen2 progress

hana-linode (commented):
It looks like we're unnecessarily getting all buckets for every action now

https://github.com/user-attachments/assets/8bb81e1d-3a07-453e-ab8b-2b7345850b4c

packages/manager/src/queries/object-storage/utilities.ts
```diff
@@ -11,3 +18,49 @@
...
+
+      if (
+        indexOfBucket === undefined ||
+        indexOfBucket === -1 ||
+        previousData?.buckets === undefined
+      ) {
+        // If the bucket does not exist in the cache or the list of buckets has not been fecthed yet,
+        // don't update the cache.
+        return undefined;
+      }
```
jaalah-akamai, line 52: Nit: For `previousData`, I would probably short circuit this before we try to find an index on L38 - then in the case of `-1` could we just return `previousData`?

jaalah-akamai (approved):
Screenshot 2024-08-02 at 9 56 49 AM

Looks good - only seeing the bucket changed being updated

packages/manager/src/queries/object-storage/utilities.ts
bnussman-akamai: Good catch. I early returned if there is no `previousData`.

I kept the `return undefined` because according to the React Query [docs](https://tanstack.com/query/latest/docs/reference/QueryClient/#queryclientsetquerydata), that will bail out of doing any cache update, which I think might be slightly better than returning previousData, but I think they'd accomplish the same thing.

hana-linode (approved):

cpathipa (commented):
Overall LGTM, confirming on create/update/delete buckets and keys.

PR #10728, opened 2024-07-30
test: [M3-8403] - Avoid cleaning up Volumes whose status is not active in Cypress

jdamore-linode:
## Description 📝
This updates the Cypress clean up logic for Volumes to avoid attempting to delete any Volumes with statuses other than 'active'. This should work around some failures we're seeing in CI related to stuck volumes that are unable to be deleted via the API, and allow the tests to run and pass while we work to clean up the stuck resources.

## Changes  🔄
- Update Volume clean up logic to avoid attempting to delete non-active Volumes

## How to test 🧪
This is difficult to test manually since it depends on having a stuck Volume (a non-stuck Volume would probably be created too quickly to effectively test this). However, we should be able to compare the CI results with other (failing) runs to see whether or not this change is effective.

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ➕ Adding a changeset

### DIFF
--- a/packages/manager/cypress/support/api/volumes.ts
+++ b/packages/manager/cypress/support/api/volumes.ts
@@ -18,7 +18,10 @@ export const deleteAllTestVolumes = async (): Promise<void> => {
   );
 
   const detachDeletePromises = volumes
-    .filter((volume: Volume) => isTestLabel(volume.label))
+    .filter(
+      (volume: Volume) =>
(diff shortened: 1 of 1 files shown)


bnussman-akamai (approved):

hkhalil-akamai (approved):

jdamore-linode:
Merging despite test failures because the failures were confirmed to be unrelated to these changes.

PR #10738, opened 2024-07-31
upcoming: [M3-8301] - New TextFieldHelperText Component

jaalah-akamai:
## Description 📝
There was a need for OBJ Gen2 to implement custom helper text with links. To avoid duplicating `RegionHelperText`, I decided to create a generic component that can replace `RegionHelperText` and accommodate any future needs.

## Changes  🔄
- The API is straightforward. It takes an array of strings or links, allowing you to build the helper text as needed.
- Added component unit tests
- Added storybook stories

## Target release date 🗓️
N/A

## Preview 📷
See storybook for examples, but with the API you could build text that looks like this:
Screenshot 2024-07-31 at 3 50 00 PM

### Verification steps
- `yarn storybook`
- `yarn test src/components/TextField/TextFieldHelperText.test.tsx`

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/.changeset/pr-10738-added-1722456106442.md
+++ b/packages/manager/.changeset/pr-10738-added-1722456106442.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Added
+---
+
+New TextFieldHelperText component ([#10738](https://github.com/linode/manager/pull/10738))
--- a/packages/manager/src/components/TextField/TextFieldHelperText.stories.tsx
+++ b/packages/manager/src/components/TextField/TextFieldHelperText.stories.tsx
@@ -1,0 +1,73 @@
+import { action } from '@storybook/addon-actions';
+import React from 'react';
+
+import { TextFieldHelperText } from 'src/components/TextField/TextFieldHelperText';
+
+import type { Meta, StoryFn, StoryObj } from '@storybook/react';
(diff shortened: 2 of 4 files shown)


AzureLatte (approved):
Looks good

dwiley-akamai (approved):
Stories in Storybook ✅
`TextFieldHelperText.test.tsx` passes locally and in CI ✅
Code review ✅

Has a ticket already been created for replacing `RegionHelperText` with this, or will that be done in upcoming OBJ Gen2 work?

bnussman-akamai (commented):
What's the benefit of using this over using `Box` + `Typography` + `Links` directly?

The same way I think `ActionsPanel` is an unnecessary abstraction, I feel like this component might also be an unnecessary abstraction. 

Code looks good, but my preference would to be use `Typography` + `Links` directly and abstract only if the helper text is reused (like how we did `RegionHelperText`)

jaalah-akamai:
> What's the benefit of using this over using `Box` + `Typography` + `Links` directly?
> 
> The same way I think `ActionsPanel` is an unnecessary abstraction, I feel like this component might also be an unnecessary abstraction.
> 
> Code looks good, but my preference would to be use `Typography` + `Links` directly and abstract only if the helper text is reused (like how we did `RegionHelperText`)

That's a valid point, so I'm going to refrain from creating this component. If there's a need for something like this in the future we can revisit.

PR #10744, opened 2024-08-02
upcoming: [M3-8301] - Object Storage Gen2 Create Bucket Additions

jaalah-akamai:
## Related / Blocked By:
- [x] https://github.com/linode/manager/pull/10736

## Description 📝
This PR handles all the new features to our v2 `CreateBucketDrawer`.

If OBJ Gen2 flag is enabled and capability is added
- Moved `getObjectStorageEndpoints` from `/objects.ts` to `/buckets.ts` - not sure why I had it there to begin with.
- Added new factories and mocks for testing
- Added new "Object Storage Endpoint Type` field
- Added new "Bucket Rate Limits" table
   - The table component is currently using static data until and API endpoint is available to return rate limits for each endpoint type.

## Target release date 🗓️
N/A

## Preview 📷
| Legacy  | Gen1   | Gen2     | Gen2 (E3) |
| ------- | ------- |  ------- | ------- |
| Screenshot 2024-08-02 at 12 42 04 PM | Screenshot 2024-08-02 at 12 42 33 PM |  Screenshot 2024-08-02 at 12 43 14 PM | Screenshot 2024-08-02 at 12 43 55 PM|

<details>
  <summary>Videos for E0, E1, E2 🎥 </summary>
   
| Legacy  | Gen1   | Gen2      |
| ------- | ------- |  ------- |
| <video src="https://github.com/user-attachments/assets/09456d41-c2be-4f7a-8b00-7417dd5846e8" /> | <video src="https://github.com/user-attachments/assets/b3c8c948-c75c-478d-81ea-a45db1312e51" /> | <video src="https://github.com/user-attachments/assets/2babb561-9e81-4327-9c16-979542fa1684" /> |
</details>

### Prerequisites
- Enable MSW and ensure `OBJ Gen2` flag is enabled

#### Endpoint Type Selector Control
- [x] Add a new control immediately below the region selector to display available endpoint types.
- [x] Automatically select the endpoint type if only one is available in the chosen region.
    - In `serverHandlers.ts` comment out all endpoints except one
- [x] Display endpoint types (E0, E1, E2, E3) in the dropdown regardless if there's multiple of the same type. The label for the option should always include endpoint type.

#### Displaying Endpoints
- [x] Display the S3 endpoint hostname for endpoints to differentiate between identical options in the dropdown.
    - In `serverHandlers.ts` feel free to change `s3_endpoint` from `null` to any hostname for an `E3` type and see that the UI displays it. There will need to be two `E3` endpoints for this to work.
- [x] Omit the S3 endpoint hostname for proposed auto-assignments.
- [x] Pass the `s3_endpoint` in the POST CreateBucket payload if it's part of the endpoints response.

#### Bucket Rate Limits
- [x] For Legacy/E1: Do not display the Bucket Rate Limits table; instead, display: "This endpoint type supports up to 750 Requests Per Second (RPS)."
- [x] For E2/E3: Show the table with basic and high numbers. All other columns should display 000.
- [x] E3 high should be 20,000.
- [x] For E2/E3, display: "Specifies the maximum Requests Per Second (RPS) for a bucket. To increase it to High, open a support ticket. Understand bucket rate limits."

#### CORS Policy
- [x] For endpoints of type E0 and E1, the behavior remains unchanged, and an "allow all" CORS policy is silently set when creating a bucket.
- [x] For endpoints of type E2 and E3, no CORS policy is set when creating a bucket due to gen2 not supporting CORS, so this is set to false.
- [x] Do not make any changes to the GUI to display CORS policies during bucket creation, but ensure the underlying API call reflects the necessary adjustments.

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.tsx
@@ -16,14 +16,11 @@ import type { ObjectStorageEndpointTypes } from '@linode/api-v4';
  */
 
 interface BucketRateLimitTableProps {
-  selectedEndpointType: {
-    label: string;
-    value: ObjectStorageEndpointTypes;
-  } | null;
+  endpointType: ObjectStorageEndpointTypes | undefined;
 }
 
 const tableHeaders = ['Limits', 'GET', 'PUT', 'LIST', 'DELETE', 'OTHER'];
-const tableData = ({ selectedEndpointType }: BucketRateLimitTableProps) => [
+const tableData = ({ endpointType }: BucketRateLimitTableProps) => [
   {
     checked: true,
     values: ['1000', '000', '000', '000', '000'],
@@ -31,7 +28,7 @@ const tableData = ({ selectedEndpointType }: BucketRateLimitTableProps) => [
   {
     checked: false,
     values: [
-      selectedEndpointType?.value === 'E3' ? '20000' : '5000',
+      endpointType === 'E3' ? '20000' : '5000',
       '000',
       '000',
       '000',
@@ -41,7 +38,7 @@ const tableData = ({ selectedEndpointType }: BucketRateLimitTableProps) => [
 ];
 
 export const BucketRateLimitTable = ({
-  selectedEndpointType,
+  endpointType,
 }: BucketRateLimitTableProps) => (
   <Table
     sx={{
@@ -67,7 +64,7 @@ export const BucketRateLimitTable = ({
       </TableRow>
     </TableHead>
     <TableBody>
-      {tableData({ selectedEndpointType }).map((row, rowIndex) => (
+      {tableData({ endpointType }).map((row, rowIndex) => (
         <TableRow key={rowIndex}>
           <TableCell>
             <Radio
--- a/packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.tsx
@@ -1,0 +1,101 @@
+import React from 'react';
+
+import { Radio } from 'src/components/Radio/Radio';
+import { Table } from 'src/components/Table';
+import { TableBody } from 'src/components/TableBody';
+import { TableCell } from 'src/components/TableCell';
+import { TableHead } from 'src/components/TableHead';
+import { TableRow } from 'src/components/TableRow';
+
+import type { ObjectStorageEndpointTypes } from '@linode/api-v4';
+
+/**
+ * TODO: This component is currently using static data until
+ * and API endpoint is available to return rate limits for
+ * each endpoint type.
+ */
+
+interface BucketRateLimitTableProps {
+  selectedEndpointType: {
+    label: string;
+    value: ObjectStorageEndpointTypes;
+  } | null;
+}
+
+const tableHeaders = ['Limits', 'GET', 'PUT', 'LIST', 'DELETE', 'OTHER'];
+const tableData = ({ selectedEndpointType }: BucketRateLimitTableProps) => [
+  {
+    checked: true,
+    values: ['1000', '000', '000', '000', '000'],
+  },
+  {
+    checked: false,
+    values: [
+      selectedEndpointType?.value === 'E3' ? '20000' : '5000',
+      '000',
+      '000',
+      '000',
+      '000',
+    ],
+  },
+];
+
+export const BucketRateLimitTable = ({
+  selectedEndpointType,
+}: BucketRateLimitTableProps) => (
+  <Table
+    sx={{
+      marginBottom: 3,
+    }}
+  >
+    <TableHead>
+      <TableRow>
+        {tableHeaders.map((header, index) => {
+          return (
+            <TableCell
+              sx={{
+                '&&:last-child': {
+                  paddingRight: 2,
+                },
+              }}
+              key={`${index}-${header}`}
+            >
+              {header}
+            </TableCell>
+          );
+        })}
+      </TableRow>
+    </TableHead>
+    <TableBody>
+      {tableData({ selectedEndpointType }).map((row, rowIndex) => (
+        <TableRow key={rowIndex}>
+          <TableCell>
+            <Radio
+              checked={row.checked}
+              disabled
+              name="limit-selection"
+              onChange={() => {}}
+              value="2"
+            />
+          </TableCell>
+          {row.values.map((value, index) => {
+            return (
+              <TableCell
+                sx={{
+                  '&&:last-child': {
+                    paddingRight: 2,
+                  },
+                }}
+                key={`${index}-${value}`}
+              >
+                {value}
+              </TableCell>
+            );
+          })}
+        </TableRow>
+      ))}
+    </TableBody>
+  </Table>
+);
+
+export default BucketRateLimitTable;
--- a/packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.tsx
@@ -94,5 +94,3 @@ export const BucketRateLimitTable = ({
     </TableBody>
   </Table>
 );
-
-export default BucketRateLimitTable;
--- a/packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
@@ -170,13 +170,11 @@ export const OMC_CreateBucketDrawer = (props: Props) => {
       const typeLabel = isLegacy ? 'Legacy' : 'Standard';
       const shouldShowHostname =
         endpointCounts && endpointCounts[endpoint_type] > 1;
-      const label =
-        shouldShowHostname && s3_endpoint !== null
-          ? `${typeLabel} (${endpoint_type}) ${s3_endpoint}`
-          : `${typeLabel} (${endpoint_type})`;
+      const label = shouldShowHostname
+        ? `${typeLabel} (${endpoint_type}) ${s3_endpoint}`
+        : `${typeLabel} (${endpoint_type})`;
 
       return {
-        endpoint_type,
         label,
         s3_endpoint: s3_endpoint ?? undefined,
         value: endpoint_type,
@@ -185,16 +183,8 @@ export const OMC_CreateBucketDrawer = (props: Props) => {
   );
 
   // Automatically select the endpoint type if only one is available in the chosen region.
-  const autoSelectEndpointType = React.useMemo(() => {
-    if (filteredEndpointOptions?.length === 1) {
-      const option = filteredEndpointOptions[0];
-      return {
-        ...option,
-        s3_endpoint: option.s3_endpoint ?? undefined,
-      };
-    }
-    return null;
-  }, [filteredEndpointOptions]);
+  const autoSelectEndpointType =
+    filteredEndpointOptions?.length === 1 && filteredEndpointOptions?.[0];
 
   const selectedEndpointType =
     filteredEndpointOptions?.find(
@@ -214,21 +204,20 @@ export const OMC_CreateBucketDrawer = (props: Props) => {
     selectedEndpointType.value !== 'E1';
 
   React.useEffect(() => {
-    const { endpoint_type, s3_endpoint, value } =
-      autoSelectEndpointType || selectedEndpointType || {};
+    if (selectedEndpointType) {
+      const { s3_endpoint, value } = selectedEndpointType;
 
-    if (s3_endpoint) {
-      setValue('s3_endpoint', s3_endpoint);
-    }
+      // For endpoints with multiple assignments, add s3_endpoint to the payload.
+      if (s3_endpoint) {
+        setValue('s3_endpoint', s3_endpoint);
+      }
 
-    if (endpoint_type) {
-      setValue('endpoint_type', endpoint_type);
+      // CORS is not supported for E2 and E3 endpoint types
+      if (['E2', 'E3'].includes(value)) {
+        setValue('cors_enabled', false);
+      }
     }
-
-    if (value && ['E2', 'E3'].includes(value)) {
-      setValue('cors_enabled', false);
-    }
-  }, [autoSelectEndpointType, selectedEndpointType, setValue]);
+  }, [selectedEndpointType, setValue]);
 
   return (
     <Drawer
--- a/packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
@@ -43,6 +43,10 @@ interface Props {
   onClose: () => void;
 }
 
+interface EndpointCount {
+  [key: string]: number;
+}
+
 export const OMC_CreateBucketDrawer = (props: Props) => {
   const { data: profile } = useProfile();
   const { isOpen, onClose } = props;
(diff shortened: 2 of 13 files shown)


packages/manager/src/features/ObjectStorage/BucketLanding/OveragePricing.tsx
```diff
@@ -66,51 +66,46 @@ export const OveragePricing = (props: Props) => {
...
+          your{' '}
+          <TextTooltip
+            displayText="global network transfer pool"
+            placement="top"
+            tooltipText={GLOBAL_TRANSFER_POOL_TOOLTIP_TEXT}
+          />
+        </>
+      )}
+      .
+    </StyledTypography>
```
jaalah-akamai, line 108: Only changes in this file is that I made them a part of the same paragraph, eliminating the space between the two text blocks (UX change).

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
```diff
@@ -186,7 +268,60 @@ export const OMC_CreateBucketDrawer = (props: Props) => {
...
+                    helperTextPosition: 'top',
+                  }}
+                  errorText={fieldState.error?.message}
+                  label="Object Storage Endpoint Type"
+                  loading={isEndpointLoading}
+                  onBlur={field.onBlur}
+                  options={filteredEndpointOptions ?? []}
+                  placeholder="Object Storage Endpoint Type"
+                  value={autoSelectEndpointType || selectedEndpointType}
+                />
```
jdamore-linode, line 297: Because we're using MSW, selecting a region other than Seattle or Newark causes the endpoint type drop-down to have no options:

Screenshot 2024-08-06 at 10 06 02 AM

I just want to confirm that this scenario won't be possible once we have real API data to work with -- if there are any regions that don't support any endpoint types, they wouldn't be selectable to begin with, right?

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
```diff
@@ -186,7 +268,60 @@ export const OMC_CreateBucketDrawer = (props: Props) => {
...
+                    ),
+                    helperTextPosition: 'top',
+                  }}
+                  errorText={fieldState.error?.message}
+                  label="Object Storage Endpoint Type"
+                  loading={isEndpointLoading}
+                  onBlur={field.onBlur}
+                  options={filteredEndpointOptions ?? []}
+                  placeholder="Object Storage Endpoint Type"
+                  value={autoSelectEndpointType || selectedEndpointType}
```
jdamore-linode, line 296: Seeing some strange behavior involving endpoint auto selection and validation error handling:

1. Select a region with only 1 endpoint type (Newark if MSW is on)
2. Observe "Standard (E3)" is automatically selected as expected
3. Click the "X" button in the autocomplete
4. Change focus to another UI element
5. Observe that "endpoint_type must be one of the following values: E0, E1, E2, E3" validation error appears even though "Standard (E3)" is still selected

I suppose making this autocomplete not clearable would work around the issue -- I can't think of any reason that wouldn't work but might not be considering every use case

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
```diff
@@ -186,7 +268,60 @@ export const OMC_CreateBucketDrawer = (props: Props) => {
...
+              <>
+                <FormLabel>
+                  <Typography marginBottom={1} marginTop={2} variant="inherit">
+                    Bucket Rate Limits
+                  </Typography>
+                </FormLabel>
+                <Typography marginBottom={isGen2EndpointType ? 2 : 3}>
+                  {isGen2EndpointType
+                    ? 'Specifies the maximum Requests Per Second (RPS) for a bucket. To increase it to High, open a support ticket. '
+                    : 'This endpoint type supports up to 750 Requests Per Second (RPS).'}
```
jdamore-linode, line 313:
```suggestion
                    : 'This endpoint type supports up to 750 Requests Per Second (RPS). '}
```

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
```diff
@@ -186,7 +268,60 @@ export const OMC_CreateBucketDrawer = (props: Props) => {
...
+                  Understand <Link to="#">bucket rate limits</Link>.
+                </Typography>
+              </>
+            )}
+            {isGen2EndpointType && (
+              <BucketRateLimitTable
```
jdamore-linode, line 319: I'm seeing just the tiniest bit of overflow when E3 is selected (presumably caused by the `20000`). I'm guessing this will be more severe once we're showing real data here.

Screenshot 2024-08-06 at 10 20 31 AM

jdamore-linode (approved):
This looks great @jaalah-akamai! Noticed a couple possible issues but nothing too serious. Thanks!

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
jaalah-akamai: Yea that won't be a possible scenario 👍

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
jaalah-akamai: Yea I need to work with UX on this, but it's following our table component. In fact, the Table component needs some major refactoring for several reasons, but this has been noted. 👍

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
jaalah-akamai: Nice catch - fixed

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
```diff
@@ -126,17 +142,83 @@ export const OMC_CreateBucketDrawer = (props: Props) => {
...
+  );
+
+  const filteredEndpointOptions = filteredEndpoints?.map(
+    ({ endpoint_type, s3_endpoint }) => {
+      const isLegacy = endpoint_type === 'E0';
+      const typeLabel = isLegacy ? 'Legacy' : 'Standard';
+      const shouldShowHostname =
+        endpointCounts && endpointCounts[endpoint_type] > 1;
+      const label = shouldShowHostname
+        ? `${typeLabel} (${endpoint_type}) ${s3_endpoint}`
```
jdamore-linode, line 174: This might be a non-issue but I'm noticing that if there are 2 endpoints of a given type, but one of them has `null` for `s3_endpoint`, `'null'` actually gets displayed on-screen. (Another case where I'm not sure this is even a valid scenario but wanted to call it out just in case)

Screenshot 2024-08-06 at 3 16 03 PM

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
```diff
@@ -126,17 +142,83 @@ export const OMC_CreateBucketDrawer = (props: Props) => {
...
+      if (s3_endpoint) {
+        setValue('s3_endpoint', s3_endpoint);
+      }
+
+      // CORS is not supported for E2 and E3 endpoint types
+      if (['E2', 'E3'].includes(value)) {
+        setValue('cors_enabled', false);
+      }
+    }
+  }, [selectedEndpointType, setValue]);
```
jdamore-linode, line 220: Think I'm seeing more strangeness related to the endpoint type auto-selection. 

It seems like if the endpoint type is automatically selected (e.g. if you select Newark with MSW enabled), `selectedEndpointType` evaluates to `null` here. The outgoing API request payload looks like this as a result:

```json
{
  "cors_enabled": true,
  "label": "test-bucket-fdsfdsf",
  "region":"us-east"
}
```
(no `endpoint_type` or `s3_endpoint` field)

Similarly, you can get the create form into a weird state if you follow these steps:

1. Select a region that has more than one endpoint type available (e.g. Seattle if MSW is on)
2. Select any endpoint type
3. Clear the endpoint type selection with the "X" button
4. Change your region selection to a region which only has one endpoint type (e.g. Newark if MSW is on)
5. Observe that while "Standard (E3)" is automatically selected, the endpoint type validation error appears and you are not able to dismiss the error (e.g. by re-selecting E3) and consequently cannot create the bucket without closing the form and starting over

jaalah-akamai: Ok I believe I fixed this issue with latest changes 👍

jdamore-linode: I think it's better! I can create (via MSW) an E3 cluster in Newark and the API request contains the expected payload data, but I am seeing that E3 is no longer selected automatically in that case (that might've been an intentional change so feel free to disregard if so).

However, I am still able to get the form into a weird state and as a result have gotten Cloud to make outgoing requests with invalid payload data (e.g. an "E2" endpoint type for Newark)

1. Open Create bucket drawer, select a region that has at least 1 endpoint type but not every endpoint type (e.g. Newark if MSW is enabled)
2. Give it a label and click "Create Bucket", observe that "Endpoint Type is required" validation error appears as expected
3. Without closing the drawer, select another region which has endpoint type(s) that the previous region did not have (e.g. Seattle if MSW is enabled)
4. Select one of the endpoint types that the previous region did not have (E0, E1, or E2 if MSW is enabled)
5. Without closing the drawer, re-select the region which had previously been selected in step 1
6. Observe that "Object Storage Endpoint Type" field is cleared
7. Click "Create Bucket", observe that no validation error appears and that an outgoing bucket create request is sent which contains the endpoint type previously selected in step 4 (but which is invalid for the selected region)


https://github.com/user-attachments/assets/49228832-f158-42c3-8470-0e0108f59648

jaalah-akamai: Ahh I see, some values weren't being reset in the form and there were some edge-case side effects even though the UI appears to be fine. I refactored things quite a bit, but I believe all the pain points you mentioned in manual testing should be fixed including the auto-selection.

I appreciate the reviews ❤️ 

https://github.com/user-attachments/assets/66bbf314-9b0f-4471-b714-1cd6d2d9e261

jdamore-linode: Nice! Messed around with switching between regions/endpoint types and every outgoing request contained the expected payload. Thanks @jaalah-akamai!

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
```diff
@@ -117,26 +158,136 @@ export const OMC_CreateBucketDrawer = (props: Props) => {
...
+  React.useEffect(() => {
+    // Auto-select an endpoint option if there's only one
+    if (filteredEndpointOptions && filteredEndpointOptions.length === 1) {
+      updateEndpointType(filteredEndpointOptions[0]);
+    } else {
+      // When region changes, reset values
+      resetSpecificFormFields();
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [watchRegion]);
```
bnussman-akamai, line 289: Can this be done in the `onChange` of region?

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
```diff
@@ -117,26 +158,136 @@ export const OMC_CreateBucketDrawer = (props: Props) => {
...
+    const label =
+      shouldShowHostname && s3_endpoint !== null
+        ? `${typeLabel} (${endpoint_type}) ${s3_endpoint}`
+        : `${typeLabel} (${endpoint_type})`;
+
+    return {
+      endpoint_type,
+      label,
+      s3_endpoint: s3_endpoint ?? undefined,
+      value: `${typeLabel}-${endpoint_type}-${index}`,
```
bnussman-akamai, line 223: Could we just remove `value` if it isn't crucial to the functionality?

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
```diff
@@ -158,47 +309,114 @@ export const OMC_CreateBucketDrawer = (props: Props) => {
...
+          <>
+            <Controller
+              render={({ field }) => (
+                <Autocomplete
+                  isOptionEqualToValue={(option, value) => {
+                    if (!option || !value) {
+                      return option === value;
+                    }
+                    return option.value === value.value;
+                  }}
```
bnussman-akamai, line 353: Do we need a `isOptionEqualToValue` for this? I think it might work without one, but double check me on that.

bnussman-akamai (commented):
Looks solid. Thanks for the nice jsdoc and comments.

Optional: Can we add some unit testing and/or cypress testing

coliu-akamai (commented):
There was something strange with the objectStorageGen2 flag on my end, more on that below 😖 

Besides that, things looked good!
✅ Endpoint type selector control
- A bit confused by the "Display endpoint types (E0, E1, E2, E3) in the dropdown regardless of the number of options" check - was this saying to display everything regardless of what's available for the regions?

✅ Displaying endpoints
✅ Bucket rate limits 
✅ CORS policy

Regarding the objectStorageGen2 flag: Inside the `useObjectStorageEndpoints` query in [packages/manager/src/queries/object-storage/queries.ts](https://github.com/linode/manager/blob/develop/packages/manager/src/queries/object-storage/queries.ts), I had to change line 98 from `Boolean(flags.objectStorageGen2?.enabled)` to `Boolean(flags.objectStorageGen2)` bc the flags were coming in as

`{objectStorageGen2: true, objMultiCluster: true}`

Not too sure if this is only happening on my end (could be some things I'm still missing access to that's causing this?), but without that change I couldn't see the endpoints 😖

packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.tsx
```diff
@@ -0,0 +1,98 @@
...
+              </TableCell>
+            );
+          })}
+        </TableRow>
+      ))}
+    </TableBody>
+  </Table>
+);
+
+export default BucketRateLimitTable;
```
carrillo-erik, line 98: Is there a reason why this includes both default and named export?

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
jaalah-akamai: I attempted to do that several times, but let me try once more since I did a bit of refactoring here

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
jaalah-akamai: I'll check

packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.tsx
jaalah-akamai: Nope good catch

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
```diff
@@ -158,47 +309,114 @@ export const OMC_CreateBucketDrawer = (props: Props) => {
...
+              <>
+                <FormLabel>
+                  <Typography marginBottom={1} marginTop={2} variant="inherit">
+                    Bucket Rate Limits
+                  </Typography>
+                </FormLabel>
+                <Typography marginBottom={isGen2EndpointType ? 2 : 3}>
+                  {isGen2EndpointType
+                    ? 'Specifies the maximum Requests Per Second (RPS) for a bucket. To increase it to High, open a support ticket. '
+                    : 'This endpoint type supports up to 750 Requests Per Second (RPS). '}
```
carrillo-erik, line 390: I think there's a good opportunity to use constants for the text copy here and maybe even for text in lines 360-361.

carrillo-erik (approved):

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
jaalah-akamai: So I can do this, but I'm open to hearing if it's worth it. So this is a true side effect in my opinion where we'll have to get the new filtered endpoint options based on the region change:

jaalah-akamai: I'm open to ideas

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
jaalah-akamai: Thanks, most likely remnants of an old refactor 👍

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
jaalah-akamai: Agree - I'll keep this in mind if we plan to re-use this text elsewhere in the feature, which at the moment we aren't

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
jaalah-akamai: 🤔 this is true! I must've been under some assumption that it was needed for the form / autocomplete 🙏

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
bnussman-akamai: Ah I see what you mean about the endpoint options. Maybe the endpoint option filtering could be extracted into a util function that could be shared if we want to try to eliminate the useEffect?

I'm okay with whichever you prefer. The useEffect looked pretty well documented and safe.  I think both are acceptable because it is a true side effect like you said.

jaalah-akamai: I'll create an optimization ticket (M3-8187) to address this 👍

bnussman-akamai (approved):

PR #10750, opened 2024-08-05
upcoming: [M3-8367] – Add "Volume Encryption" section to Volume Create page

dwiley-akamai:
## Description 📝
Add "Volume Encryption" section to Volume Create page.

> [!NOTE]
> The wireframes have a warning notice below the Linode dropdown when the selected linode uses an old qemu version. However, the API team has not finalized yet where that data point will be placed, so I created M3-8420 to circle back to that once finalized.

## Changes  🔄
- Unit test coverage added with `VolumeCreate.test.tsx`
- Added support in `Encryption.tsx` for warning notices above checkbox via optional `notices` prop
- Added `useIsBlockStorageEncryptionFeatureEnabled` util, updated constants
- Added "Volume Encryption" section to Volume Create page
 
## Target release date 🗓️
8/19/24

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screenshot 2024-08-05 at 4 40 05 PM | Screenshot 2024-08-06 at 2 16 45 PM |

<details>
<summary>Notices when checkbox is checked</summary>

Screenshot 2024-08-05 at 5 28 40 PM
</details>

### Prerequisites
Point at the dev environment with the `blockstorage-encryption` tag on your account

### Verification steps
- Confirm that you can create a volume with the indicated encryption status (i.e., unchecked box should result in `encryption: "disabled"` in the payload and a checked box should result in `encryption: "enabled"` in the payload)

- [X] 👀 Doing a self review
- [X] ➕ Adding a changeset

### DIFF
--- a/packages/manager/cypress/e2e/core/volumes/create-volume.spec.ts
+++ b/packages/manager/cypress/e2e/core/volumes/create-volume.spec.ts
@@ -8,9 +8,6 @@ import { interceptCreateVolume } from 'support/intercepts/volumes';
 import { randomNumber, randomString, randomLabel } from 'support/util/random';
 import { chooseRegion } from 'support/util/regions';
 import { ui } from 'support/ui';
-import { mockAppendFeatureFlags } from 'support/intercepts/feature-flags';
-import { accountFactory } from 'src/factories';
-import { mockGetAccount } from 'support/intercepts/account';
 
 // Local storage override to force volume table to list up to 100 items.
 // This is a workaround while we wait to get stuck volumes removed.
@@ -203,47 +200,5 @@ describe('volume create flow', () => {
           });
       }
     );
-  });
-
-  it('should not have a "Volume Encryption" section visible if the feature flag is off and user does not have capability', () => {
-    // Mock feature flag -- @TODO BSE: Remove feature flag once BSE is fully rolled out
-    mockAppendFeatureFlags({
-      blockStorageEncryption: false,
-    }).as('getFeatureFlags');
-
-    // Mock account response
-    const mockAccount = accountFactory.build({
-      capabilities: ['Linodes'],
-    });
-
-    mockGetAccount(mockAccount).as('getAccount');
-
-    // intercept request
-    cy.visitWithLogin('/volumes/create');
-    cy.wait(['@getFeatureFlags', '@getAccount']);
-
-    // Check if section is visible
-    cy.findByText('Encrypt Volume').should('not.exist');
-  });
-
-  it('should have a "Volume Encryption" section visible if feature flag is on and user has the capability', () => {
-    // Mock feature flag -- @TODO BSE: Remove feature flag once BSE is fully rolled out
-    mockAppendFeatureFlags({
-      blockStorageEncryption: true,
-    }).as('getFeatureFlags');
-
-    // Mock account response
-    const mockAccount = accountFactory.build({
-      capabilities: ['Linodes', 'Block Storage Encryption'],
-    });
-
-    mockGetAccount(mockAccount).as('getAccount');
-
-    // intercept request
-    cy.visitWithLogin('/volumes/create');
-    cy.wait(['@getFeatureFlags', '@getAccount']);
-
-    // Check if section is visible
-    cy.findByText('Encrypt Volume').should('exist');
   });
 });
--- a/packages/manager/cypress/e2e/core/volumes/create-volume.spec.ts
+++ b/packages/manager/cypress/e2e/core/volumes/create-volume.spec.ts
@@ -219,9 +219,7 @@ describe('volume create flow', () => {
     mockGetAccount(mockAccount).as('getAccount');
 
     // intercept request
-    cy.visitWithLogin('/volumes/create', {
-      localStorageOverrides: pageSizeOverride,
-    });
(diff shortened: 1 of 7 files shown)


packages/manager/src/features/Volumes/VolumeCreate.tsx
```diff
@@ -127,6 +138,12 @@ export const VolumeCreate = () => {
   const [hasSignedAgreement, setHasSignedAgreement] = React.useState(false);
   const { mutateAsync: updateAccountAgreements } = useMutateAccountAgreements();
 
+  const [encryptVolume, setEncryptVolume] = React.useState(false);
```
bnussman-akamai, line 141: Is this state necessary? 

The Formik form has `encryption` in its type, but we are using this separate state instead.  Can we find a way to use Formik state exclusively?

carrillo-erik:
I don't see a changeset, maybe you haven't gotten around to it? Nice job improving the e2e coverage. 👍

dwiley-akamai:
> I don't see a changeset, maybe you haven't gotten around to it? Nice job improving the e2e coverage. 👍

Whoops, yes, meant to add it this morning. Just pushed it

packages/manager/src/components/Encryption/utils.ts
bnussman-akamai, line 1: Would `isFeatureEnabledV2` be suitable here?

packages/manager/cypress/e2e/core/volumes/create-volume.spec.ts
```diff
@@ -201,4 +205,50 @@ describe('volume create flow', () => {
...
+    // Mock account response
+    const mockAccount = accountFactory.build({
+      capabilities: ['Linodes'],
+    });
+
+    mockGetAccount(mockAccount).as('getAccount');
+
+    // intercept request
+    cy.visitWithLogin('/volumes/create', {
+      localStorageOverrides: pageSizeOverride,
```
bnussman-akamai, line 224: I don't think these overrides are needed for the create page.

packages/manager/cypress/e2e/core/volumes/create-volume.spec.ts
```diff
@@ -201,4 +205,50 @@ describe('volume create flow', () => {
...
+    });
+    cy.wait(['@getFeatureFlags', '@getAccount']);
+
+    // Check if section is visible
+    cy.get(`[data-testid=${headerTestId}]`).should('not.exist');
```
bnussman-akamai, line 229: Could we check for the actual header text instead of the`testid`? 

There's nothing wrong with checking for the testid, but I'm thinking it might be slightly more beneficial to check for text that the user should see.

packages/manager/cypress/e2e/core/volumes/create-volume.spec.ts
bnussman-akamai, line 1: Optional: These tests might be okay to have as unit tests because they are relatively simple. 

Just looking for ways to save on the e2e run time ⌛ 😄

bnussman-akamai (approved):
Looks good! 🚀  Left a few non-blocking comments

It might be nice to very slightly decrease the spacing between the notices and the checkbox, but it's probably fine as-is. 

Screenshot 2024-08-09 at 11 05 14 AM

packages/manager/src/components/Encryption/utils.ts
dwiley-akamai: It could be used, but so far we have used it only within UI components. And we could use it in components like `VolumeCreate`, but it's nice to not have to pull in `useAccount()` and `useFlags()` in so many files

dwiley-akamai:
Hey @carrillo-erik, can you review this today?

packages/manager/src/components/Encryption/utils.ts
bnussman-akamai: Good point.

packages/manager/src/features/Volumes/VolumeCreate.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+        return {
+          isBlockStorageEncryptionFeatureEnabled: true,
+        };
+      }
+    );
+
+    const { queryByText } = renderWithTheme(<VolumeCreate />);
+
+    expect(queryByText('Encrypt Volume')).toBeInTheDocument();
+  });
```
bnussman-akamai, line 50: There might be some value in using a more "e2e" approach so that `useIsBlockStorageEncryptionFeatureEnabled` is also tested instead of mocked. This approach does make it a bit harder to test the falsy case though

```suggestion
it('should have a "Volume Encryption" section visible if feature flag is on and user has the capability', async () => {
    server.use(
      http.get('*/v4/account', () => {
        return HttpResponse.json(
          accountFactory.build({ capabilities: ['Block Storage Encryption'] })
        );
      })
    );

    const { findByText } = renderWithTheme(<VolumeCreate />, {
      flags: { blockStorageEncryption: true },
    });

    await findByText('Encrypt Volume');
  })
```

dwiley-akamai: Opting to keep things as they are in the interest of moving things along for BSE, but we can have confidence in this pattern (was used for LDE also) and the general logic underlying `isBlockStorageEncryptionFeatureEnabled`

PR #10752, opened 2024-08-06
upcoming: [M3-7942] - Use object-storage/endpoints to get buckets

jaalah-akamai:
## Related / Blocked By:
- [x] https://github.com/linode/manager/pull/10744

## Description 📝
Use the new OBJ endpoint GET `/v4/object-storage/endpoints` that returns an array of regions to fetch buckets by region. This endpoint returns only regions that support Object Storage and is a much faster `O(n)` compared to filtering regions by capabilities `O(n*k)`.

## Changes  🔄
- Upon error, instead of returning a `Region`, an `Endpoint` will be returned which has the `region`, `s3_endpoint`, and `endpoint_type` which I believe to be more important information

Screenshot 2024-08-05 at 10 03 17 PM
-  Add `getAllBucketsFromEndpoints` which will be the way to get buckets moving forward. Other queries will eventually become deprecated and removed once OBJ Gen2 is in GA.

## Target release date 🗓️
N/A

### Prerequisites
- Test all UI paths: Legacy, Multicluster (Gen1), Gen2 by toggling feature flags in dev environment

### Reproduction steps
- Turn on MSW
- With Gen2 flag on, go to Object Storage landing page
- Observe that you should see a bunch of seattle buckets fetched and a few other regions (you won't see every region because we're using the /endpoints mock where I've provisioned just enough to test the feature).

### Verification steps
- Ensure buckets are fetched without failures
- You can mock a failure by adding this snippet to `/object-storage/requests.ts`
```
// Mock a failure
const getBucketsInRegion = async (regionId: string, params: any) => {
  throw new Error('Forced error for testing');
};
```

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset


packages/manager/src/mocks/serverHandlers.ts
```diff
@@ -94,6 +94,9 @@ import { LinodeKernelFactory } from 'src/factories/linodeKernel';
 import { pickRandom } from 'src/utilities/random';
 import { getStorage } from 'src/utilities/storage';
 
+const getRandomWholeNumber = (min: number, max: number) =>
+  Math.floor(Math.random() * (max - min + 1) + min);
+
```
jaalah-akamai, line 99: Purely an MSW  - happy to omit, but I'm sure it'll be improved with MSW refactor we're doing

packages/manager/src/queries/object-storage/requests.ts
```diff
@@ -120,3 +150,43 @@ export const getAllBucketsFromRegions = async (
...
+      errors.push({ endpoint: result.endpoint, error: result.error });
+    }
+  });
+
+  if (errors.length === endpoints.length) {
+    throw new Error('Unable to get Object Storage buckets.');
+  }
+
+  return { buckets, errors };
+};
```
jaalah-akamai, line 192: This is based off the related functionality above, but based around `ObjectStorageEndpoints`

bnussman-akamai (approved):

packages/manager/src/queries/object-storage/requests.ts
```diff
@@ -34,23 +34,50 @@ export const getAllObjectStorageEndpoints = () =>
...
+// TODO: OBJGen2 - Remove the `Gen2` suffix when Gen2 is in GA.
+export interface BucketErrorGen2 {
+  endpoint: ObjectStorageEndpoint;
+  error: APIError[];
+}
+
+// TODO: OBJGen2 - Only needed during interim period when Gen2 is in beta.
+export type BucketsResponseType<T> = T extends true
+  ? BucketsResponseGen2
+  : BucketsResponse;
```
dwiley-akamai, line 69: This pattern seems like a better way of handling for situations where we need to maintain legacy & upcoming code for a feature

dwiley-akamai (approved):
Code review ✅
Tested Legacy, Gen1, and Gen2 paths ✅

packages/api-v4/src/object-storage/types.ts
```diff
@@ -99,11 +99,6 @@ export interface DeleteObjectStorageBucketPayload {
 }
 
 export interface ObjectStorageBucket {
-  /*
-   @TODO OBJ Multicluster: 'region' will become required, and the 'cluster' field will be deprecated
-   once the feature is fully rolled out in production as part of the process of cleaning up the 'objMultiCluster'
-   feature flag.
-  */
   region?: string;
```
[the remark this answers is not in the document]
jaalah-akamai: No, not yet! I just removed that comment since it's outdated. I'll clean it up as part of OBJ Gen2

packages/manager/src/queries/object-storage/requests.ts
jaalah-akamai: Yea it's much better than doing `BucketsResponse | BucketsResponseGen2` and then having to add type narrowing wherever it's used.

PR #10753, opened 2024-08-06
upcoming: [APL-55] - Application platform for Linode Kubernetes 

dennisvankekem:
## Description 📝
UI implementation for the upcoming APL ( Application platform for Linode) feature 

## Changes  🔄
- APL UI elements added to the createKubernetes and KubernetesOverview pages 
- HA mode will automatically be set whenever APL is enabled
- Kubernetes plans that are too small will be grayed out whenever APL is enabled
- Breadcrumb is extended so it support a array of numbers

## Target release date 🗓️
N/A

## Preview 📷

https://github.com/user-attachments/assets/9a5a1fcd-12bc-4cbd-8414-8552f1a61b84

image

## How to test 🧪

- coming soon


dennisvankekem:
please add the 'do not merge' and 'work in progress' tags

abailly-akamai:
@dennisvankekem is this PR still relevant or a work in progress? I am going to help move things with APL and wondering about the state of this contribution. If moving forward with it, can we get it rebased and updated? I'll start reviewing the code then - thx!

dennisvankekem:
@abailly-akamai I will update this PR with some final minor changes and then the code is ready to test. To test the functionality of the feature however we have to wait until the APL feature in API is merged first. I will rebase and mark this PR for review once I've pushed the last changes which is expected somewhere tomorrow

packages/api-v4/src/kubernetes/kubernetes.ts
```diff
@@ -38,20 +38,23 @@ export const getKubernetesClusters = (params?: Params, filters?: Filter) =>
 export const getKubernetesCluster = (clusterID: number) =>
   Request<KubernetesCluster>(
     setMethod('GET'),
-    setURL(`${API_ROOT}/lke/clusters/${encodeURIComponent(clusterID)}`)
+    setURL(`${BETA_API_ROOT}/lke/clusters/${encodeURIComponent(clusterID)}`)
```
abailly-akamai, line 41: This will update **all** users to a beta endpoint, which is essentially a breaking change.

My understanding is that only the people enrolling to the beta program will see their upstream change. 

In order to respect the contract we want to create a new function (ex: `getKubernetesClusterBeta`) and have instances of it called behind a feature flag.

packages/api-v4/src/kubernetes/kubernetes.ts
```diff
@@ -38,20 +38,23 @@ export const getKubernetesClusters = (params?: Params, filters?: Filter) =>
...
 
 /**
  * createKubernetesClusters
  *
  * Create a new cluster.
  */
-export const createKubernetesCluster = (data: CreateKubeClusterPayload) =>
-  Request<KubernetesCluster>(
+export const createKubernetesCluster = (data: CreateKubeClusterPayload) => {
+  const apiRoot = data.apl_enabled ? BETA_API_ROOT : API_ROOT;
```
abailly-akamai, line 50: Same sentiment here, a new call is preferable. Conditioning your API destination based on the payload is rather error prone. We want to have a flag controlling this feature as a whole.

packages/api-v4/src/kubernetes/types.ts
```diff
@@ -10,6 +10,7 @@ export interface KubernetesCluster {
   id: number;
   tags: string[];
   control_plane: ControlPlaneOptions;
+  apl_enabled?: boolean;
```
abailly-akamai, line 13: At the risk of sounding redundant, the same applies here. `apl_enabled` is not an optional field. It is a required boolean for the beta program, so it needs to be typed as such. Marking the field optional ends up weakening type safety. We can just make a `KubernetesClusterBeta` type that extends `KubernetesCluster`. (i'll let you decide on your naming conventions). It will increase the complexity of your logic a little, but facilitate any beta upstream change that may occur in the future, till we can consolidate it all.

packages/api-v4/src/kubernetes/types.ts
```diff
@@ -66,6 +67,7 @@ export interface ControlPlaneOptions {
 export interface CreateKubeClusterPayload {
   label?: string; // Label will be assigned by the API if not provided
   region?: string; // Will be caught by Yup if undefined
+  apl_enabled?: boolean;
```
abailly-akamai, line 70: ditto

packages/manager/src/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -64,10 +64,19 @@ export const Breadcrumb = (props: BreadcrumbProps) => {
...
+    // Sort the indices in descending order
+    const indicesToRemove = new Set(removeCrumbX.map((index) => index - 1));
+
+    // Filter out the indices to remove
+    pathMap = allPaths.filter((_, index) => !indicesToRemove.has(index));
+  } else if (removeCrumbX != null) {
+    pathMap = removeByIndex(allPaths, removeCrumbX - 1);
+  } else {
+    pathMap = allPaths;
+  }
```
abailly-akamai, line 79: It does make sense to have this, however please provide coverage for it.

packages/manager/src/featureFlags.ts
```diff
@@ -85,6 +85,7 @@ export interface Flags {
   aclpReadEndpoint: string;
   aclpResourceTypeMap: CloudPulseResourceTypeMapFlag[];
   apiMaintenance: APIMaintenance;
+  apl: boolean;
```
abailly-akamai, line 88: Just a note here, you may want to set up your flag as a JSON payload to have more control over time, shall the feature complexity increase a little. ex:

```
apl: {
  enabled: boolean;
}
```

This will allow you to have more granular control as the feature grows

packages/manager/src/features/Kubernetes/CreateCluster/CreateCluster.tsx
```diff
@@ -18,9 +18,11 @@ import { RegionHelperText } from 'src/components/SelectRegionPanel/RegionHelperT
 import { Stack } from 'src/components/Stack';
 import { TextField } from 'src/components/TextField';
 import {
+  getAPLAvailability,
```
abailly-akamai, line 21: we usually make a hook for this, so we don't have to check for flag + account capability everywhere. ex:
https://github.com/linode/manager/blob/develop/packages/manager/src/components/Encryption/utils.ts#L10-L31

packages/manager/src/features/Kubernetes/CreateCluster/CreateCluster.tsx
```diff
@@ -264,6 +274,25 @@ export const CreateCluster = () => {
             placeholder={' '}
             value={versions.find((v) => v.value === version) ?? null}
           />
+          {showAPL && flags.apl ? (
```
abailly-akamai, line 277: see comment above

packages/manager/src/features/Kubernetes/CreateCluster/HAControlPlane.tsx
```diff
@@ -12,6 +12,7 @@ import { Typography } from 'src/components/Typography';
 
 export interface HAControlPlaneProps {
   highAvailabilityPrice: string;
+  isAPLEnabled: boolean | undefined;
```
abailly-akamai, line 15: and using the hook avoids prop drilling

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/APLSummaryPanel.tsx
```diff
@@ -0,0 +1,104 @@
...
+      }
+    } catch (error) {
+      if (error.response && error.response.status === 404) {
+        setStatus({
+          message:
+            'APL is still being deployed, please check back in a minute.',
+          resolved: false,
+        });
+      } else {
+        console.error('An error occurred:', error.message);
```
abailly-akamai, line 57: you'll have to remove this to pass linting checks. we sentry log by default so it's prolly fine to swallow it.

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/APLSummaryPanel.tsx
```diff
@@ -0,0 +1,104 @@
...
+  message?: string;
+  resolved?: boolean;
+  url?: string;
+}
+
+const checkConsoleURL = async (
+  cluster: KubernetesCluster,
+  setStatus: React.Dispatch<React.SetStateAction<StatusState>>
+) => {
+  const consoleURL = `https://console.lke${cluster.id}.akamai-apl.net`;
```
abailly-akamai, line 39: nit: put this into your `constants.ts` file

abailly-akamai (commented):
PR is looking good & clean, thanks for your contribution!

I did not review the UI yet, just looked at the code and pointed changes you'll want to consider to move the feature forward. Please reach out with any question/concern 👍

dennisvankekem:
@abailly-akamai Thanks for your review, sadly I didn't get notified by Github in my email for some reason,  so your review wasn't noticed by me. I will look at it tomorrow!

packages/manager/src/components/Breadcrumb/Breadcrumb.tsx
dennisvankekem: Will do

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/APLSummaryPanel.tsx
dennisvankekem: I would like to challenge this idea. Although consoleURL is a const, the URL is dependent on one external variable `cluster.id` whereas the other constants in `constants.ts` are all completely static without external variables.

abailly-akamai:
@dennisvankekem thx for addressing this initial feedback - I will give it another go and see what we can come up with to avoid too much duplication

abailly-akamai:
@dennisvankekem in the interest of time we may want to loosen the typing grip based on the refactor and duplication needed. I did not expect things to cascade that far and I agree the debt may trump the original intent. 

Am thinking we could do:
```
export interface CreateKubeClusterPayload {
  label?: string; // Label will be assigned by the API if not provided
  region?: string; // Will be caught by Yup if undefined
  node_pools: CreateNodePoolData[];
  k8s_version?: string; // Will be caught by Yup if undefined
  control_plane?: ControlPlaneOptions;
  // an important comment about this type, such as this needs a value for BETA
  apl_enabled: boolean | undefined
}
```
which is not idea but would get us to a compromise where we are always reminder to declare the property. Thoughts?

Meanwhile
Can you:
- fix the current type issues
- fix any broken test

also, you probably to conditionaly render copy referring to APL. ex (i don't think we want that copy for non-adopters since APL isn't a thing to them): 
Screenshot 2024-10-03 at 16 47 46

packages/manager/src/features/Kubernetes/KubernetesPlansPanel/KubernetesPlansPanel.tsx
```diff
@@ -27,6 +27,7 @@ interface Props {
   getTypeCount: (planId: string) => number;
   hasSelectedRegion: boolean;
   header?: string;
+  isAPLEnabled?: boolean | undefined;
```
abailly-akamai, line 30: it's either 

```TS
isAPLEnabled: boolean | undefined;
```

or 

```TS
isAPLEnabled?: boolean;
```

please replace all instances

packages/manager/src/features/Kubernetes/kubeUtils.ts
```diff
@@ -112,6 +113,17 @@ export const getKubeHighAvailability = (
   };
 };
 
+export const useGetAPLAvailability = (): boolean => {
+  // const showAPL = account?.capabilities.includes('LKE APL');
```
abailly-akamai, line 117: what about capabilities here?

dennisvankekem: can be removed if we CustomerTags are not handled through it at least.

abailly-akamai: Not sure I follow

dennisvankekem:
'work in progress' and 'do not merge' labels may be removed

packages/manager/src/features/Kubernetes/KubernetesPlansPanel/KubernetesPlansPanel.tsx
dennisvankekem: `isAPLEnabled` is now by default defined as `false` as I do not see why it should be parsed as undefined to child components anyway. It only needs to be an optional parameter in the createKubernetes and KubernetesQuery as the beta and non-Beta Types are shared. Please let me know if you agree on that.

dennisvankekem: never mind, I was looking at it the wrong way, forgot that some components or types are being used outside of kubernetes feature. Will revert

packages/manager/src/features/Kubernetes/KubernetesPlansPanel/KubernetesPlansPanel.tsx
```diff
@@ -80,13 +82,14 @@ export const KubernetesPlansPanel = (props: Props) => {
 
   const tabs = Object.keys(plans).map(
     (plan: Exclude<LinodeTypeClass, 'nanode' | 'standard'>) => {
-      const plansMap: PlanSelectionType[] = plans[plan]!;
+      const plansMap: PlanSelectionType[] = plans[plan];
```
dennisvankekem, line 85: this line throws a type error: ` Type 'undefined' is not assignable to type 'PlanSelectionType[]'`, but I have no clue where this is coming from as I can't seem to figure out where this undefined is coming from. 

It's currently the only type check thats fails atm

packages/api-v4/src/kubernetes/types.ts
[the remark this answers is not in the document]
dennisvankekem: - Aligned the Chip better, it was indeed missing `AlignItems`, nice catch!
- Replaced the APL copy with a question mark `TooltipIcon` as this seems to be the convention as well as clean up the interface a bit 
- Regarding Label switching to darker: This seems to be intented behaviour whenever you have the field highlighted, notice that the color will return to normal whenever another field is highlighted (which will in return be darker)
- added comment to type with `apl_enabled?: boolean` and duplicated functions as to why they're there

packages/manager/src/featureFlags.ts
j-zimnowoda: @abailly-akamai we expect that this feature flag by default is set to falsy (false|undefined). Is that correct?

abailly-akamai: The feature flag can be anything you want it to be (to a certain limit)

What I was suggesting here is setting up a modular flag that can be extended to add more properties shall the feature get more complex (right now it's just a pure boolean). ex:

```TS
apl: {
  enabled: boolean;
  showBetaChip: boolean;
  useBetaAPI: boolean;
}
```

this is completely optional but worth pointing at this is probably the best time to introduce something like that if that's something you'd want.

abailly-akamai (approved):
Great work @dennisvankekem - thanks for addressing all changes, this is good to go on my end. 

- confirmed payloads and posting to beta API when flag `ON` ✅ 
- confirmed payloads and posting to prod API when flag `OFF` ✅ 
- UI consistent with Cloud Manager and feature requirements ✅ 

Non-blocking notes:
K8 detail page dos not show anythign about APL when opted in. Something to consider in the UI
Screenshot 2024-10-08 at 09 43 55

Noticing an error when POST to clusters
Screenshot 2024-10-08 at 09 50 30

ElderMatt (approved):

jaalah-akamai:
✅ Payload is looking good

Non-blocking:
UX needs to consider the tablet/mobile breakpoints for this. As of right now, the UI does not look great at these sizes.

Screenshot 2024-10-08 at 10 54 41 AM

jaalah-akamai (approved):

PR #10754, opened 2024-08-06
feat: [M3-7666] - Remove Animation from Search Results and Animate Icon Instead

pmakode-akamai:
## Description 📝
As a user, the animation on the first character of no search results can result in potentially offensive words. Let's remove the animation from the character and animate the logo instead.

## Changes  🔄
- Remove the animation from the character/search query text - The first character is no longer animated.
- Add animation on svg icon.
- Update the unit test case.

## Target release date 🗓️
8/19

## Preview 📷
| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/user-attachments/assets/d6f005f2-b348-4f5e-b96b-17878e99a65e" /> |  <video src="https://github.com/user-attachments/assets/b3e6f328-2d77-4910-ab11-c26e8f4b712d" />|

### Reproduction steps
- Type something irrelevant in the Search input and Enter.

### Verification steps
- Ensure the first character is no longer animated.
- Ensure the icon is animated in some fun way.

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ➕ Adding a changeset

### DIFF
--- a/packages/manager/src/features/Search/SearchLanding.styles.ts
+++ b/packages/manager/src/features/Search/SearchLanding.styles.ts
@@ -30,9 +30,6 @@ const blink = keyframes`
   0%, 50%, 100% {
     transform: scaleY(0.1);
   }
-  25%, 75% {
-    transform: scaleY(0.1);
-  }
 `;
 
 const rotate = keyframes`
--- a/packages/manager/src/features/Search/SearchLanding.styles.ts
+++ b/packages/manager/src/features/Search/SearchLanding.styles.ts
@@ -1,6 +1,7 @@
+import { keyframes } from '@emotion/react';
 import { Stack } from '@mui/material';
-import Grid from '@mui/material/Unstable_Grid2';
 import { styled } from '@mui/material/styles';
+import Grid from '@mui/material/Unstable_Grid2';
 
 import Error from 'src/assets/icons/error.svg';
 import { H1Header } from 'src/components/H1Header/H1Header';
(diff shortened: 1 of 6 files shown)


packages/manager/src/features/Search/SearchLanding.tsx
```diff
@@ -55,14 +55,6 @@ export interface SearchLandingProps
...
     RouteComponentProps<{}> {}
 
-const splitWord = (word: any) => {
-  word = word.split('');
-  for (let i = 0; i < word.length; i += 2) {
-    word[i] = <span key={i}>{word[i]}</span>;
-  }
-  return word;
-};
-
```
pmakode-akamai, line 65: No need to split the word as we are removing any animation on the search query text.

packages/manager/src/features/Search/SearchLanding.tsx
```diff
@@ -55,14 +55,6 @@ export interface SearchLandingProps
   extends SearchProps,
     RouteComponentProps<{}> {}
 
-const splitWord = (word: any) => {
```
pmakode-akamai, line 58: No need to split the word as we are removing any animation on the search query text.

packages/manager/src/features/Search/SearchLanding.styles.ts
```diff
@@ -25,9 +26,50 @@ export const StyledGrid = styled(Grid, {
   padding: `${theme.spacing(10)} ${theme.spacing(4)}`,
 }));
 
+const blink = keyframes`
+  0%, 50%, 100% {
+    transform: scaleY(0.1);
+  }
+  25%, 75% {
+    transform: scaleY(0.1);
+  }
```
hkhalil-akamai, line 35: Both `transform` values are identical -- is this intentional?

hkhalil-akamai (approved):
Thanks for handling this and nice work on the SVG animation, cute!

packages/manager/src/features/Search/SearchLanding.styles.ts
pmakode-akamai: Let me refactor this a bit.

bnussman-akamai (approved):
Nice!

packages/manager/src/features/Search/SearchLanding.tsx
```diff
@@ -285,11 +277,9 @@ export const SearchLanding = (props: SearchLandingProps) => {
             <Typography style={{ marginBottom: 16 }}>
               You searched for ...
             </Typography>
-            <Typography className="resultq">
-              {query && splitWord(query)}
-            </Typography>
+            <Typography className="resultq">{query}</Typography>
             <Typography className="nothing" style={{ marginTop: 56 }}>
-              Sorry, no results for this one
+              Sorry, no results for this one.
```
jdamore-linode, line 282: Looks like this improvement is causing test failures in one of our Cypress UI specs 😅

You should be able to fix it by making the same change here:
https://github.com/linode/manager/blob/3a6584c/packages/manager/cypress/e2e/core/linodes/smoke-delete-linode.spec.ts#L23

jdamore-linode (changes requested):
Thanks for the improvement @pmakode-akamai! Looks like we're getting a couple test failures in CI that we'll want to fix before merging this (and otherwise have no reason to be alarmed by), but otherwise this all looks great!

Don't hesitate to reach out on Slack if you have any questions!

packages/manager/src/features/Search/SearchLanding.tsx
pmakode-akamai: @jdamore-linode Fixed in 268046f48bb34bf214a4901d4e94bcd4614e6d1f

jdamore-linode (approved):
Thanks @pmakode-akamai!

PR #10633, opened 2024-07-01
test: [M3-8122] - Disable access to internet for Linodes created during Cypress tests

jdamore-linode:
## Description 📝
This modifies our Cypress tests so that any Linodes created during the course of a test are created with no access to the internet. It does this by modifying the `interceptCreateLinode` util to mutate the outgoing request.

Note that this applies only to Linodes that are created via Cloud Manager during the tests -- Linodes created directly by Cypress via our JS SDK already have extra measures taken to secure them (see #10538).

## Changes  🔄
- Mutate outgoing Linode create requests to place Linodes behind VLAN and, more importantly, without access to the internet
- Increase clone test timeout

## How to test 🧪
1. Confirm that tests continue to pass (we can do this via CI)
2. Confirm that created Linodes are not connected to the internet
  - There are a couple ways to do this, but this is the easiest:
    1. Run a Linode create test, then sign into Cloud and find the Linode that was created
    2. Observe the IP address of the created Linode
    3. Run `ping <IP address>` and confirm that you cannot connect
    4. Run `ssh root@<IP address>` and confirm that you cannot connect
  - To be extra thorough, you can confirm that the Linode itself can't connect to the internet
    1. Run a Linode create test via `yarn cy:debug` with the Chrome dev tools open
    2. In the network tab, observe the outgoing Linode create request and make note of the root password
    3. Sign into Cloud, find the Linode that was created, and launch Glish/Weblish
    4. Sign into the root account using the root password you observed in step 2
    5. From the Linode, run `ping 1.1.1.1` and confirm that there is no internet connectivity

## As an Author I have considered 🤔

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ❔ Our [contribution guidelines](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md)
- [ ] 🤏 Splitting feature into small PRs
- [ ] ➕ Adding a changeset
- [ ] 🧪 Providing/Improving test coverage
- [ ] 🔐 Removing all sensitive information from the code and PR description
- [ ] 🚩 Using a feature flag to protect the release
- [ ] 👣 Providing comprehensive reproduction steps
- [ ] 📑 Providing or updating our documentation
- [ ] 🕛 Scheduling a pair reviewing session
- [ ] 📱 Providing mobile support
- [ ] ♿  Providing accessibility support

PR #10765, opened 2024-08-08
test: [M3-8146] - Cypress tests for refactored Linode Create via CLI flow

cliu-akamai:
## Description 📝
Refactor Cypress test to the flow of creating Linode through CLI.

## Major Changes 🔄
- Refactor Cypress test that creates Linode through CLI.

## How to test 🧪
```
yarn cy:run -s "cypress/e2e/core/linodes/legacy-create-linode.spec.ts"
```

### DIFF
--- a/packages/manager/cypress/e2e/core/linodes/legacy-create-linode.spec.ts
+++ b/packages/manager/cypress/e2e/core/linodes/legacy-create-linode.spec.ts
@@ -17,6 +17,7 @@ import { chooseRegion } from 'support/util/regions';
 import { getRegionById } from 'support/util/regions';
 import {
   accountFactory,
+  createLinodeRequestFactory,
   subnetFactory,
   vpcFactory,
   linodeFactory,
@@ -29,6 +30,8 @@ import {
 import { authenticate } from 'support/api/authentication';
 import { cleanUp } from 'support/util/cleanup';
 import { mockGetRegions } from 'support/intercepts/regions';
+import { API_ROOT } from 'src/constants';
+import { oauthToken } from 'support/constants/api';
 import {
   dcPricingPlanPlaceholder,
   dcPricingMockLinodeTypes,
@@ -179,6 +182,20 @@ describe('create linode', () => {
     const linodeLabel = randomLabel();
     const linodePass = randomString(32);
     const linodeRegion = chooseRegion();
+    const linodeType = 'g6-dedicated-2';
+    const linodeImage = 'linode/debian11';
+    const payload = createLinodeRequestFactory.build({
+      authorized_users: [],
+      backups_enabled: false,
+      booted: true,
+      image: linodeImage,
+      label: linodeLabel,
+      private_ip: false,
+      region: linodeRegion.id,
+      root_pass: linodePass,
+      tags: [],
+      type: linodeType,
+    });
 
     cy.visitWithLogin('/linodes/create');
 
@@ -188,7 +205,7 @@ describe('create linode', () => {
       .should('exist')
       .click();
 
-    cy.get('[id="g6-dedicated-2"]').click();
+    cy.get(`[id="${linodeType}"]`).click();
 
     cy.findByLabelText('Linode Label')
       .should('be.visible')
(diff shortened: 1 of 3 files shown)


yarn.lock
mjac0bs, line 1: How did this get in here?

packages/manager/cypress/e2e/core/linodes/legacy-create-linode.spec.ts
mjac0bs, line 1: I may be misunderstanding the point of this ticket, but based on the comment on L4-L5 of this file:

```
// TODO Delete this test file when `linodeCreateRefactor` feature flag is retired.
// Move out any tests (e.g. region select test) for flows that aren't covered by new tests in the meantime.
```

It sounds like we wouldn't want to make updates in this file, but rather write a new test file that tests CLI in the **refactored** (v2) flow. Since this test suite mocks the linode create feature flag [false](https://github.com/linode/manager/pull/10765/files#diff-e9a1b9cf41bab377cfc730b07faf7f55fa53eaa24f08d821fa2625fd6fd3bdfdL95), we don't seem to be doing that here.

packages/manager/cypress/e2e/core/linodes/legacy-create-linode.spec.ts
```diff
@@ -207,42 +224,93 @@ describe('create linode', () => {
...
+        expect(resBody.label).to.equal(payload.label);
+        expect(resBody.region).to.equal(payload.region);
+        expect(resBody.type).to.equal(payload.type);
+      });
+
+      cy.visitWithLogin('/linodes/');
+
+      // Confirm that the linode is created.
+      cy.findByText(linodeLabel).should('be.visible');
+      cy.findByText('Provisioning', { exact: false }).should('be.visible');
```
mjac0bs, line 312: Not quite what the acceptance criteria requested here: `Confirm Cloud Manager behavior to automatically redirect to created Linode after CLI command is executed`. Isn't the redirect to the Linode's details page?

mjac0bs (changes requested):
This test failed in CI, but it does run and pass locally (5+ times), and I can't quite follow what happened in the CI run.

It looks like we're testing the wrong flow here, though! See my comment on the `legacy-create-linode.spec.ts`.

packages/manager/cypress/e2e/core/linodes/legacy-create-linode.spec.ts
```diff
@@ -207,42 +224,93 @@ describe('create linode', () => {
...
+    "booted": ${payload.booted},
+    "image": "${payload.image}",
+    "label": "${payload.label}",
+    "private_ip": ${payload.private_ip},
+    "region": "${payload.region}",
+    "root_pass": "${payload.root_pass}",
+    "tags": [],
+    "type": "${payload.type}"
+}' ${API_ROOT}/linode/instances`;
+      cy.exec(curlRequest).then((res) => {
```
hkhalil-akamai, line 296: The CI error is being caused by executing the curl command. It seems like the jenkins instance doesn't have curl installed on it.

```
Stderr:
/usr/bin/bash: line 1: curl: command not found
```

@jdamore-linode can we update the jenkins docker config to install cURL if we want to go ahead with this approach?

jdamore-linode: @cliu-akamai (cc @hkhalil-akamai) we probably _shouldn't_ be doing this E2E, and instead should mock the relevant events to trigger/test the redirect behavior, even though this is really, really cool as it is.

(I'm only reluctant to do this because we wouldn't have the normal security guardrails available, like ensuring that the created Linode can't connect to the internet, etc. -- basically none of the protections that were added by #10538 and #10633 would apply to this cURL call)

hkhalil-akamai:
Here's the code that detects the event and triggers the redirect:

https://github.com/linode/manager/blob/14b02430623d770ddb6c60ac91957aec4863ed88/packages/manager/src/features/Linodes/LinodesCreate/ApiAwarenessModal/ApiAwarenessModal.tsx#L64-L69

We can try mocking this instead of manually executing the curl command.

@cliu-akamai

packages/manager/cypress/e2e/core/linodes/legacy-create-linode.spec.ts
cliu-akamai: No, it doesn't redirect after sending the CLI.

PR #10766, opened 2024-08-09
refactor: Clean up Account Settings Object Storage

bnussman-akamai:
## Description 📝

- Refactors the Object Storage section of the https://cloud.linode.com/account/settings page 🔁 
- Adds a React Query mutation called `useCancelObjectStorageMutation` so we don't have to manage our own loading and error states 🧼 😃 
- Renames the component from `EnableObjectStorage` to  `ObjectStorageSettings` ✏️ 
  -  `EnableObjectStorage` was very confusing. This component's main purpose is to allow users to _**disable**_ object storage
- Adds a new toast for when you successfully disable object storage 🍞 
- Updates the unit tests to account for changes 🧪 

## Preview 📷

> [!note]
> UI should not look any different. All changes were to just clean up the code and use React Query for the mutation

https://github.com/user-attachments/assets/d1c11ff9-cbe8-40fe-8665-d0a236011554

## How to test 🧪

- Test the disable Object Storage flow on http://localhost:3000/account/settings
  - The `enable-object-storage.spec.ts` covers this flow so ensure that test passes ✅ 

- [ ] 👀 Doing a self review


mjac0bs:
> Renames the component from EnableObjectStorage to ObjectStorageSettings ✏️
> EnableObjectStorage was very confusing. This component's main purpose is to allow users to disable object storage

Thank you! 🙏🏼 This has tripped me up more than once.

packages/manager/src/features/Account/ObjectStorageSettings.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+    // The confirm button should be disabled because the user need to type to confirm
+    expect(confirmButton).toBeDisabled();
+
+    const typeToConfirmTextField = getByLabelText('Username');
+
+    // Type the user's username to confirm
+    await userEvent.type(typeToConfirmTextField, 'my-username-1');
+
+    // The confirm button should become enabled because we typed username
```
coliu-akamai, line 103: these comments walking through the test are really nice!

packages/manager/src/features/Account/ObjectStorageSettings.tsx
```diff
@@ -0,0 +1,112 @@
...
+          </Stack>
+        ) : (
+          <Typography variant="body1">
+            Content storage and delivery for unstructured data. Great for
+            multimedia, static sites, software delivery, archives, and data
+            backups. <br />
+            To get started with Object Storage, create a{' '}
+            <Link to="/object-storage/buckets">Bucket</Link> or an{' '}
+            <Link to="/object-storage/access-keys">Access Key.</Link>{' '}
+            <Link to="https://www.linode.com/docs/platform/object-storage/">
```
coliu-akamai, line 79: If there's already an existing constants file, maybe this link could be moved there - it appears a few times throughout the codebase by itself

packages/manager/src/features/Account/ObjectStorageSettings.tsx
```diff
@@ -0,0 +1,112 @@
...
+    false
+  );
+
+  const handleCloseCancelDialog = () => {
+    setIsCancelDialogOpen(false);
+    reset();
+  };
+
+  const handleCancelObjectStorage = async () => {
+    cancelObjectStorage().then(() => {
```
coliu-akamai, line 40: could remove the `async` here since you use `then` (unless the async is necessary? if so would love to learn)

coliu-akamai (approved):
✅ UI still looks the same on light/dark mode
✅ integration + unit tests passed

The added toast is a really nice touch 🤩 

Unrelated to this PR, but I did see some weird bolding when testing on Safari:
image
I don't have access yet to create a ticket - would someone be able to create one if this is something that should be addressed?

packages/manager/src/features/Account/ObjectStorageSettings.tsx
bnussman-akamai: Good call. `async` can be removed!

jaalah-akamai (approved):

PR #10769, opened 2024-08-12
upcoming: [DI-20204] - Added Interface type setup, query updates and enhanced preferences type for upcoming Cloudpulsecustomselect component

venkymano:
## Description 📝
Added new Use Query setup for the upcoming CloudPulseCustomSelect component, that can either gets the static options from the configuration and list it down or call an API Url and list down the filters in the dropdown.

## Changes  🔄

1.  Added code for new use query setup, that takes a api URL, filters as input and returns the formatted filters as output
2. Enhanced existing dbaas list all instances query with XFILTER support to make it compatible for our dbaas use cases.
3. Renamed the filterKey from dbEngine to engine in our FilterConfig.ts
4. Enhanced the type for our AclpConfig to more generic type because we can't have fixed set of keys for CloudPulseCustomSelect Component

## Target release date 🗓️
20th August 2024

## Preview 📷

| Before  | After   |
| ------- | ------- |

## How to test 🧪

1. This is just interface enhancements and query call setup and doesn't introduce any new UI and functionality changes to existing code.

*Check all that apply*

- [x] 👀 Doing a self review


packages/manager/src/queries/databases/requests.ts
```diff
@@ -21,3 +28,11 @@ export const getAllDatabaseTypes = () =>
   getAll<DatabaseType>((params) => getDatabaseTypes(params))().then(
     (data) => data.data
   );
+
+export const getAllDatabasesWithFilters = (
```
venkymano, line 32: FYI - The existing getAllDatabases method doesn't accept  xFilter parameter, where as the xFilter is supported in the actual list API, so we created this function following existing cloud manager standards

bnussman-akamai: Instead of doing that, can you please just add a `filter` param to the `getAllDatabases` function and query. There is no need to have two "get all databases" functions

packages/manager/src/queries/databases/databases.ts
```diff
@@ -57,6 +58,10 @@ export const databaseQueries = createQueryKeys('databases', {
         queryFn: getAllDatabases,
         queryKey: null,
       },
+      allWithFilters: (params: Params, filter: Filter) => ({
+        queryFn: () => getAllDatabasesWithFilters(params, filter),
+        queryKey: [params, filter],
+      }),
```
bnussman-akamai, line 64:
```suggestion
```

Please just add `filter` to the `all` query instead of duplicating it.

packages/manager/src/queries/cloudpulse/customfilters.ts
```diff
@@ -0,0 +1,52 @@
...
+  idField: string,
+  labelFields: string,
+  params: Params = {},
+  filter: Filter = {}
+) => {
+  return useQuery<
+    { [key: string]: unknown }[], // the use case here is api url and api response here is not consistent across multiple service types, it can list of db engines, list of node types etc.,
+    unknown,
+    CloudPulseServiceTypeFiltersOptions[]
+  >([queryKey], () => getAllFilters(params, filter, url), {
```
bnussman-akamai, line 22: Please use "single object notation" here (see [here](https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-v5#supports-a-single-signature-one-object)). We need all queries to use this syntax to make our upgrade to TanStack Query v5 easier. 

Just move the `queryKey` and `queryFn` into a single object.

bnussman-akamai (changes requested):

packages/manager/src/queries/cloudpulse/customfilters.ts
venkymano: Moved it, thanks

packages/manager/src/queries/databases/requests.ts
venkymano: @bnussman-akamai , we are assuming getAllDatabases this might be already used by the dbaas and adding a params and filters to the all query might break, so keeping that , can I go ahead make these params optional in the getAllDatabases query and remove the newly created filters

bnussman-akamai: Adding the `filter` option shouldn't  break anything. We should be okay to add it but keep it optional

venkymano: Hi @bnussman-akamai , updated to support optional filter object and removed newly created functions

packages/manager/src/queries/databases/databases.ts
venkymano: Hi @bnussman-akamai , updated to support optional filter object and removed newly created functions

jaalah-akamai:
@venkymano Can we break this up into two PRs, one that focuses on adding the queries, endpoints and types. And the other that contains the actual has the implementation?

venkymano:
@jaalah-akamai , if we just alone create the queries and types in a single PR, we might have the ability to test it. Is it still okay, please confirm?

kmuddapo:
> @venkymano Can we break this up into two PRs, one that focuses on adding the queries, endpoints and types. And the other that contains the actual has the implementation?
> 
> This PR can be the queries, endpoints and types since Banks has commented already on changes for that

@jaalah-akamai , We have actually thought of breaking this into smaller, but the testing capability as a feature for the reviewer is getting effected, so considering that in mind, we have created this PR,  Please suggest if its ok not to have the testing capability in this PR and have the same testing capability in the next PR, then will try to split it accordingly as you suggested. Thanks!! inadvance

packages/manager/src/features/CloudPulse/shared/CloudPulseCustomSelect.tsx
```diff
@@ -0,0 +1,323 @@
...
+   */
+  type: CloudPulseSelectTypes;
+}
+
+export enum CloudPulseSelectTypes {
+  dynamic,
+  static,
+}
+
+export const CloudPulseCustomSelect = React.memo(
```
hkhalil-akamai, line 103: This component is very long and I'm concerned that it will be difficult to maintain in the future. Can we make it more readable by breaking it up and extracting functionality into utility functions? For example, `getDefaultSelectionsFromPreferences` and `handleChange` can be extracted.

hkhalil-akamai (commented):
Thanks for this contribution and for accepting feedback from the team. Looking over this PR, I'm concerned about the complexity of the custom select component and the maintainability of some of this code.

Please review my suggestions to improve the readability and maintainability of the changes.

packages/api-v4/src/cloudpulse/customfilter.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * This function makes an API-V4 call based on given URL and returns the response
+ * @param params - The params of the api call, contains information related to pagination
+ * @param filters - The xFilters that needs to applied in the API
+ * @param url - The URL that is going to be called
+ * @returns - The filters loaded from the API
+ */
+export const getFilters = (params: Params, filters: Filter, url: string) =>
+  Request<ResourcePage<{ [key: string]: unknown }>>( // since our API's can be anything like list of db engines, node types, region etc., come up with this structure
```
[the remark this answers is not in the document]
venkymano: Hi @hkhalil-akamai , Thanks for asking, Our use case is, we have a config to maintain the filters we show on top of the dashboard. So in some of the filters (custom select), the filter values we get from calling an API in the linode. The API can be anything like list of db engines, fetch list of db clusters, list of node types, list of regions etc., Mostly in this case we don't know predefined what will be response from the API, it can be a JSONObject which contains list of filters we need to show, so even we tried using like something below

```
export type JsonValue = string | number | boolean | JsonObject | JsonArray;
export interface JsonObject {
  [key: string]: JsonValue;
}
export type JsonArray = JsonValue[];
```

But it is more or less the same as **{ [key: string]: unknown }**,  using unknown is better than using any as it offers type safety.

Please see if this helps and let me know whether it is fine or give some suggestions, Thanks.

packages/api-v4/src/cloudpulse/types.ts
```diff
@@ -45,12 +45,15 @@ export interface Filters {
...
-  timeDuration: string;
-  widgets: { [label: string]: AclpWidget };
+  // we maintain only the filters selected in the preferences for latest selected dashboard
+  [key: string]:
+    | number
+    | string
+    | string[]
+    | number[]
+    | { [key: string]: AclpWidget }
+    | undefined;
```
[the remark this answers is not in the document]
venkymano: @hkhalil-akamai , Thanks, here the use case is to store all the latest selected filters in preferences, along with widget level preferences, Since we don't have fixed set of filters and filter keys change based on dashbaord service type, 

Example 

1. Dashboard A - service type - linode, filters - resources, region
2. Dashboard B - service type - dbaas, filters - clusters, region, engine, node type

Thinking of the future, for another service types we might have some more unique filter keys, this makes it harder to come up with a interface type, so we have used a more generic type like this. 

Since these filters are configurable to avoid releases, in future, a service owner can introduce a new filter key as well.

Is this fine, or if you have any suggestions please let us know

nikhagra (approved):
approved

venkymano:
@jaalah-akamai / @bnussman-akamai / @hkhalil-akamai , Thanks for reviewing, We have descoped this work in to two smaller PR's as suggested by Jaalah. In this PR we will have only the changes associated with the following

1. Any type changes
2. New Query setup and Existing query enhancements

The CloudPulseCustomSelect implementation will be done in the next PR

@jaalah-akamai , please mark this PR as open if the changes are as you suggested.

Please review and provide your suggestions

Thanks

packages/manager/src/queries/databases/databases.ts
```diff
@@ -53,10 +53,10 @@ export const databaseQueries = createQueryKeys('databases', {
   }),
   databases: {
     contextQueries: {
-      all: {
-        queryFn: getAllDatabases,
-        queryKey: null,
-      },
+      all: (filter?: Filter) => ({
```
bnussman-akamai, line 56: Can we default to an empty object
```suggestion
      all: (filter: Filter = {}) => ({
```

packages/manager/src/queries/cloudpulse/customfilters.ts
```diff
@@ -0,0 +1,85 @@
...
+        }): CloudPulseServiceTypeFiltersOptions => {
+          return {
+            id: String(filter[idField]),
+            label: String(filter[labelField]),
+          };
+        }
+      );
+    },
+  });
+};
```
bnussman-akamai, line 72: Frankly, I don't think this should exist. 

It looks like you are trying to make a query that can GET _anything_ from _any_ API endpoint.  I don't think this is a good pattern for many reasons

- It will be difficult to keep everything typesafe
- It will not share our existing cache. For example, if we use `useInfiniteLinodesQuery` somewhere to fetch a list of Linodes, this query might duplicate the same data and CloudPulse won't pull from that existing cache because it is a complete separate query with a different query key
- Adds unnecessary complexity
- Does not follow our standard query key patterns

I think our existing queries should be sufficient to build what you are trying to accomplish

packages/api-v4/src/cloudpulse/customfilter.ts
```diff
@@ -0,0 +1,10 @@
+import { Params, Filter, ResourcePage } from 'src/types';
+import Request, { setMethod, setParams, setURL, setXFilter } from '../request';
+
+export const getFilters = (params: Params, filters: Filter, url: string) =>
+  Request<ResourcePage<{ [key: string]: unknown }>>(
```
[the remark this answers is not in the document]
bnussman-akamai: Why does this exist? We already have typesafe APIv4 functions (`getDatabaseEngines`, `getRegions`) for getting the things you mentioned.

bnussman-akamai (changes requested):

packages/manager/src/queries/cloudpulse/customfilters.ts
venkymano: @bnussman-akamai , As discussed with @kmuddapo , our approach is here is to get any filters provided by any service owners like, linode, dbaas etc.,

We have designed this **more generic** like, accept a filter endpoint and list down. So that for the **service owners, it will be just a addition of a filter config with API-URL** and **avoid code changes** in the actual component

Another use case, A service owner like linode, dbaas etc., might only have a **completed API available whereas in the Cloud Manager UI application we don't have the appropriate query factory and use query implementation**. In this case, the suggested approach will not work, as we don't have the implementation available in our Cloud Manager UI. We need to **enforce the service owner to implement those in the Cloud Manager UI**, even though they have the API available.

Also, if we go by the existing queries approach, our **useGetCustomFiltersQuery** will have the capability of getting filters from any service provider, it can get a list of Linode[], Engine[], NodeType[] ,....Test[], in this case we need to maintain a type that unions most of the existing type.

Please suggest if we still need to re-use the existing things, or if we need to discuss more on this, as a team we can connect over a call and discuss.

Thanks

bnussman-akamai: >  to get any filters provided by any service owners 

Can you elaborate on this? I don't know of any API endpoints that return "filters"? I'm not sure what filter means in this case. What would be an example of a `url` here?

nikhagra: https://github.com/linode/manager/blob/develop/packages/manager/src/features/CloudPulse/Utils/FilterConfig.ts

@bnussman-akamai if you see this sample filter config file, so similar configuration will be there for each service type that service owners will configure and based on the dashboard they select from the dashboard select drop down, different filters from configuration will be shown to user which may vary from service type to service how it is confugured. Now to show those filters from configuration, we have to make it fully dynamic and give flexibility to service owners to configure and either provide us static drop down option lists or an endpoint from which we fetch the options and show in filter drop downs. To achieve this we can't completely  use existing queries because we don't know what service owners have configured.

packages/api-v4/src/cloudpulse/customfilter.ts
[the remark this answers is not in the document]
jaalah-akamai: I have to agree, this APIv4 has been built to mirror the Linode API and adding this generic approach bypasses, and I would even say undermines, that principle in allowing any arbitrary endpoint to be called with any set of filters. If you need some brainstorming as to how to accomplish what you're trying to do with our already existing functions, I'm happy to setup a discussion around that!

packages/manager/src/queries/cloudpulse/customfilters.ts
bnussman-akamai: Can you give an example use of `useGetCustomFiltersQuery` or an example of what `url ` could be here?

jaalah-akamai:
> Why does this exist? We already have typesafe APIv4 functions (getDatabaseEngines, getRegions) for getting the things you mentioned.

I agree with this, we're essentially allowing here any arbitrary endpoint to be called with any set of filters. This really cuts across the grain of the core intent of the `APIv4` which is to provide specific, type-safe functions that mirror the Linode API.

packages/manager/src/queries/cloudpulse/customfilters.ts
jaalah-akamai: If there’s a need to centralize filter logic, have you considered creating a higher-level utility that manages filter state and logic but still uses specific queries under the hood? I think what we all are concerned about here is this really deviates from the clear one-to-one mapping of between the SDK and the API.

venkymano: @bnussman-akamai , sample URL may be

1. ${API_ROOT}/databases/types
2. ${API_ROOT}/databases/engines
3. ${API_ROOT}/regions
4. ${API_ROOT}/nodebalancers/types

These are just assumptions, it will grow based on config, like this we can receive any endpoints. Suppose these query factory and usequery implementation is not available, we need to enforce the service providers to implement it

jaalah-akamai: To add to what I mean, what do you all think about something like this? 

We create a new hook called something like `FilterManager` that can be used to manage the filter state and filter data based on the configuration provided in FILTER_CONFIG:
1. Get the filter options based on the service type
3. Update the filter state
4. Filter the data based on the current filter state


Here's kind of what I mean in terms of usage:
```
const { data: linodes } = useLinodesQuery();
const { filterState, updateFilter, filteredData } = useFilterManager('linode'); // 'linode' or 'dbaas' from the FILTER_CONFIG
const filteredLinodes = filteredData(linodes);
```

On any sort of change event, we could update the filter accordingly:

```
const handleRegionChange = (e) => {
  updateFilter('region', e.target.value);
};
const handleResourceIdChange = (e) => {
  updateFilter('resource_id', e.target.value);
};

Just a thought, but this approach could enable us to apply it across different service types by simply passing the serviceType as an argument. This would map to the corresponding config, retrieve the relevant filter options, and then apply those filters to the service-specific data.

**Comment by venkymano** [comment]
@jaalah-akamai / @bnussman-akamai  - Thanks for reviewing, We removed the API-V4 setup for customfilter and started using existing queryfactories from service providers like linode, dbaas etc., we have modified our filter config type to accept the queryKey and queryFn from the existing queryfactories and make use of it to list down the filters.

For now we have changed our approach like the above and we can discuss and enhance it later when we start extending the use cases.

**Comment by jaalah-akamai** [comment]
I'll review this again tomorrow

File: packages/api-v4/src/cloudpulse/types.ts [hunk 10]
```diff
@@ -44,13 +44,21 @@ export interface Filters {
...
+
 export interface AclpConfig {
-  dashboardId: number;
-  interval: string;
-  region: string;
-  resources: string[];
-  timeDuration: string;
-  widgets: { [label: string]: AclpWidget };
+  // we maintain only the filters selected in the preferences for latest selected dashboard
+  [key: string]: FilterValue;
```
hkhalil-akamai, line 60: My earlier comment seems to have been removed when this PR was descoped.

Is it possible to narrow this type based on the API spec? It's better to have a more specific type even if it may require code changed if the spec is updated. Having more specific types increases the usefulness of autocomplete, catches type errors early and avoids anti-patterns like casting.

I would like to see the API spec and assist with developing this type; if it's available can you send it to me privately?

venkymano: Hi @hkhalil-akamai , This type is for saving the Global and Widget level filter values from the dashboard in the user preferences, the type of our filter values can be string , number , string[], number[], TimeDuration like those. There is no API spec for this , these values are stored in our user preferences API. Hope this helps, happy to explain further over a quick chat as well

packages/manager/src/features/CloudPulse/Utils/models.ts
```diff
@@ -30,6 +39,28 @@ export interface CloudPulseServiceTypeFilters {
   name: string;
 }
 
+/**
+ * As of now, the list of possible custom filters are engine, database type, this union type will be expanded if we start enhancing our custom select config
+ */
+export type QueryFunctionType = DatabaseEngine[] | DatabaseType[];
+
+export type QueryFunctionNonArrayType = DatabaseEngine | DatabaseType;
```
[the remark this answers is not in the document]
venkymano: @jaalah-akamai , thanks for the suggestion, we will consider this in upcoming PR's. But for now instead of maintaining two type one Array based and one non array based, we have inferred the type from array types and used it, making it easy to maintain in only one place.

packages/manager/src/features/CloudPulse/Utils/models.ts
```diff
@@ -46,8 +79,9 @@ export interface CloudPulseServiceTypeFiltersConfiguration {
 
   /**
    * This is an optional field, it is required if the type is dynamic for call the respective API to get the filters
+   * example, databaseQueries.types, databaseQueries.engines etc., makes use of existing query key and optimises cache
    */
-  apiUrl?: string;
+  apiV4FactoryFunction?: QueryFunctionAndKey;
```
jaalah-akamai, line 84: I would just call this `apiV4QueryKey` as to not create confusion with the way we use the name `Factory` for our mocking functions.

packages/manager/src/queries/cloudpulse/customfilters.ts
```diff
@@ -0,0 +1,91 @@
...
+} from 'src/features/CloudPulse/Utils/models';
+
+/**
+ * Interface that has the parameters required for making custom filter query call
+ */
+interface CustomFilterQueryProps {
+  /**
+   * The Built in API-V4 query factory functions like databaseQueries.types, databaseQueries.engines etc., makes use of existing query key and optimises cache
+   */
+  apiFactoryFunction: QueryFunctionAndKey;
```
jaalah-akamai, line 18: Same here for this name

packages/manager/src/queries/cloudpulse/customfilters.ts
```diff
@@ -0,0 +1,91 @@
...
+  const { apiFactoryFunction, enabled, idField, labelField } = queryProps;
+  return useQuery<
+    QueryFunctionType,
+    unknown,
+    CloudPulseServiceTypeFiltersOptions[]
+  >({
+    // receive filters and  return only id and label
+    enabled,
+    ...apiFactoryFunction,
+    select: (filters: QueryFunctionType) => {
```
jaalah-akamai, line 58:
```suggestion
    select: (filters: QueryFunctionType): CloudPulseServiceTypeFiltersOptions[] => {
```

packages/manager/src/queries/cloudpulse/customfilters.ts
```diff
@@ -0,0 +1,91 @@
...
+      return filters
+        .map(
+          (filter): CloudPulseServiceTypeFiltersOptions => {
+            return {
+              id: getStringValue(filter, idField) ?? '',
+              label: getStringValue(filter, labelField) ?? '',
+            };
+          }
+        )
+        .filter((value) => value.id.length && value.label.length);
```
jaalah-akamai, line 69:
```suggestion
        .filter(({ id, label }) => id.length && label.length);
```

packages/manager/src/queries/databases/databases.ts
```diff
@@ -53,10 +53,10 @@ export const databaseQueries = createQueryKeys('databases', {
   }),
   databases: {
     contextQueries: {
-      all: {
-        queryFn: getAllDatabases,
-        queryKey: null,
-      },
+      all: (filter: Filter = {}) => ({
```
jaalah-akamai, line 56: In addition to filter, do we need to allow params here for any reason?

packages/manager/src/queries/databases/databases.ts
```diff
@@ -94,7 +94,7 @@ export const useDatabasesQuery = (params: Params, filter: Filter) =>
 
 export const useAllDatabasesQuery = (enabled: boolean = true) =>
```
jaalah-akamai, line 95: We should probably pass `filter: Filter = {},` to this query now

packages/manager/src/queries/databases/databases.ts
```diff
@@ -94,7 +94,7 @@ export const useDatabasesQuery = (params: Params, filter: Filter) =>
...
   useQuery<DatabaseInstance[], APIError[]>({
-    ...databaseQueries.databases._ctx.all,
+    ...databaseQueries.databases._ctx.all(),
```
jaalah-akamai, line 97: Same here, `...databaseQueries.databases._ctx.all(filter),`

jaalah-akamai (approved):
Approving pending a few changes. cc: @bnussman-akamai or @hkhalil-akamai can you give this one more pass / approval?

packages/manager/src/queries/databases/databases.ts
venkymano: As of now we don't have the use case, but thinking of the future, enhanced to support optional params, as per standards

packages/manager/src/queries/databases/databases.ts
venkymano: enhanced this existing method to have optional filter and params

kmuddapo:
> Approving pending a few changes. cc: @bnussman-akamai or @hkhalil-akamai can you give this one more pass / approval?

Thanks!! @jaalah-akamai  for reviewing and approving this taking your off hours for this.  Really appreciate the efforts here.

venkymano:
@jaalah-akamai , Thanks for review, addressed the comments

packages/manager/src/queries/cloudpulse/customfilters.ts
bnussman-akamai, line 1: This is looking better!

I'm not exactly sure how `useGetCustomFiltersQuery` is going to be used, but does it need to have a dedicated query?  The name `useGetCustomFiltersQuery` doesn't make any sense to because we're not getting any filters from the API. It seems like the goal is to get entities, not filters. 

If you need to implement a select component for any arbitrary entity, it could look something like
```ts

const queries = {
  databases: useAllDatabasesQuery,
  linode: useAllLinodesQuery,
};

interface Props {
  entity: keyof typeof queries;
}

const MyComponent = (props: Props) => {
  const useResourceQuery = queries[props.entity];
  const { data, isLoading } = useResourceQuery(
    {},
    { label: { '+contains': 'test' } }
  );

  const options = data?.map((entity) => ({
    id: entity.id,
    label: entity.label,
  }));

  return <Autocomplete label="Select Entity" options={options ?? []} />;
};
```

venkymano: @bnussman-akamai , we are already using the same factory functions used underneath the suggested hooks, but it will be better if we have the controls like when to enable, what fields to pick etc., which might differ according to hooks,  so we have designed our own hook, which will help us in having more flexibility.

Also, as we are driving everything from config like below

```
{
      configuration: {
        apiV4FactoryFunction: databaseQueries.types, // this is the existing queryfactory function
        filterKey: 'dbType',
        filterType: 'string',
        isFilterable: true,
        isMetricsFilter: true,
        isMultiSelect: false,
        name: 'DB Type',
        neededInServicePage: true,
        placeholder: 'Select a DB Type',
        priority: 3,
        type: CloudPulseSelectTypes.dynamic,
      },
      name: 'DB Type',
    },
```
As we are abstracting the implementation aspects from the select component, it will be easy for service provider to construct a config like above with factory functions rather relying on both the config and component for getting the functionality out.

venkymano: Also, in this context, since these are filters listed as a dropdown in our global filters view, come up with this name of the hook

jaalah-akamai: I'm looking forward to seeing the implementation! I really like how clean the suggested API Banks showed above - hopefully we can achieve something similar 👍

venkymano: @jaalah-akamai , yes, in our current implementation we have followed the similar approach what banks suggested, along with this, we have made it more of it as a config driven approach along with our own useQuery hook for more flexibility and control. It will come in the next PR as we have split this PR in to two.

nikhagra:
@jaalah-akamai & @bnussman-akamai thanks for approvals. Lets merge it so that we will raise another

PR #10770, opened 2024-08-12
upcoming: [M3-8398] - Generate firewall dialog

hkhalil-akamai:
## Description 📝
Implements the generate firewall dialog, which guides users through the process of generating a compliant firewall from a template.

## Changes  🔄
- New firewall template endpoints to `api-v4` package
    - Add RQ hook for firewall templates
- New `GenerateFirewallDialog` component that steps users through generating a firewall from a template
    - Add hook `useCreateFirewallFromTemplate` which exposes a script to fetch firewall templates, select a label and create the new firewall
    - Implement dialog in `FirewallLanding` and Linode create flows
    - Unit tests
- New utility `replaceNewlinesWithLineBreaks` to handle rendering multi-line content from LD
    - Unit tests

## Target release date 🗓️
8/19

## Preview 📷
See the internal ticket.

## How to test 🧪
See the internal ticket.

*Check all that apply*

- [x] 👀 Doing a self review


packages/manager/src/components/GenerateFirewallDialog/GenerateFirewallDialog.tsx
```diff
@@ -0,0 +1,219 @@
...
+  onClose: () => void;
+  onFirewallGenerated?: (firewall: Firewall) => void;
+  open: boolean;
+}
+
+export type DialogState =
+  | ErrorDialogState
+  | ProgressDialogState
+  | PromptDialogState
+  | SuccessDialogState;
```
hkhalil-akamai, line 27: `DialogState` is used to track the users current progress in the generate dialog and stores relevant data to each step (e.g., the progress or the generated firewall)

packages/manager/src/components/GenerateFirewallDialog/useCreateFirewallFromTemplate.ts
```diff
@@ -0,0 +1,93 @@
...
+  // Determine new firewall name
+  const label = await getUniqueFirewallLabel(template, firewalls);
+  updateProgress(80);
+
+  // Create new firewall
+  return await createFirewall({ label, rules: template.rules });
+};
+
+const getFirewallTemplate = async (queryClient: QueryClient) => {
+  const templates = await queryClient.fetchQuery(firewallQueries.templates);
```
bnussman-akamai, line 68: I think we might want to use `ensureQueryData` because it will read from the cache if the data is already fetched. 

```suggestion
  const templates = await queryClient.ensureQueryData(firewallQueries.templates);
```

packages/manager/src/components/GenerateFirewallDialog/useCreateFirewallFromTemplate.ts
```diff
@@ -0,0 +1,93 @@
...
+  if (templates.length < 1) {
+    throw Error('No firewall templates are available.');
+  }
+  return templates[0];
```
bnussman-akamai, line 72: Is it safe to assume that the first template is the one we want?

Rather than doing a fetch-all for firewall templates, could we just fetch the specific template for this project `getTemplate("akamai-non-prod")`?

packages/manager/src/components/GenerateFirewallDialog/GenerateFirewallDialog.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+        onFirewallGenerated={onFirewallGenerated}
+        open
+      />
+    );
+
+    getByText('Generate an Akamai Compliant Firewall');
+
+    act(() => {
+      userEvent.click(getByText('Generate Firewall Now'));
+    });
```
bnussman-akamai, line 104: I believe `userEvent` should be awaited and does not need to be in an `act`
```suggestion
    await userEvent.click(getByText('Generate Firewall Now'));
```

packages/manager/src/components/GenerateFirewallDialog/useCreateFirewallFromTemplate.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  // Get firewalls and firewall template in parallel
+  const [template, firewalls] = await Promise.all([
+    getFirewallTemplate(queryClient),
+    queryClient.fetchQuery(firewallQueries.firewalls._ctx.all),
+  ]);
+  updateProgress(50); // this gives the appearance of linear progress
+
+  // Determine new firewall name
+  const label = await getUniqueFirewallLabel(template, firewalls);
```
bnussman-akamai, line 60: Looks like `getUniqueFirewallLabel` does not need to be async. I'm wondering if the linear progress implementation here is a bit over-engineered. From some of my testing, it appears to be at 80% most of the time (moves first time, but is stuck at 80% on the second creation).

https://github.com/user-attachments/assets/3dfaaa20-fc49-4304-a1ef-dbb2d04b5f24

packages/manager/src/components/GenerateFirewallDialog/GenerateFirewallDialog.test.tsx
```diff
@@ -0,0 +1,110 @@
...
```
bnussman-akamai, line 50: I think `userEvent` should be awaited and I don't think we need the `act`

```suggestion
    await userEvent.click(getByText('Generate Firewall Now'));
```

bnussman-akamai (commented):
From an UI/UX standpoint, I'm not sure how I feel about putting the buttons on the left side. I feel like they should be on the right to be consistent with our patterns.

Screenshot 2024-08-13 at 1 22 08 PM

hkhalil-akamai:
> From an UI/UX standpoint, I'm not sure how I feel about putting the buttons on the left side. I feel like they should be on the right to be consistent with our patterns.

Good point. Let me follow up with UX about this

packages/manager/src/components/GenerateFirewallDialog/useCreateFirewallFromTemplate.ts
```diff
@@ -0,0 +1,93 @@
...
+  queryClient: QueryClient;
+  updateProgress: (progress: number | undefined) => void;
+}): Promise<Firewall> => {
+  const { createFirewall, queryClient, updateProgress } = handlers;
+  updateProgress(0);
+
+  // Get firewalls and firewall template in parallel
+  const [template, firewalls] = await Promise.all([
+    getFirewallTemplate(queryClient),
+    queryClient.fetchQuery(firewallQueries.firewalls._ctx.all),
```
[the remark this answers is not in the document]
hkhalil-akamai: I think the naming was decided after a lot of back and forth from project folks and technical writers. But also from a technical standpoint, how will we avoid duplication without doing a fetch all? Another approach could be to retry the create firewall API call until it passes but there's a lot of concerns with that too:
- We don't have error enums of any kind so we would have to do a string search which could potentially change and break in the future
- Could potentially result in a lot of network requests

packages/manager/src/components/GenerateFirewallDialog/useCreateFirewallFromTemplate.ts
hkhalil-akamai: Good catch, changed! The other `fetchQuery` (to fetch the firewalls) is necessary however, since if you generate successive firewalls the cache won't update.

packages/manager/src/components/GenerateFirewallDialog/useCreateFirewallFromTemplate.ts
hkhalil-akamai: For the time being, I believe this is the simplest approach. It saves us from having to release code changes when the slug name changes (which has already happened once). In the future, if there are multiple templates, we can revisit this approach.

hkhalil-akamai: To make sure, I will verify with the API team that they aren't planning to add multiple templates.

packages/manager/src/components/GenerateFirewallDialog/useCreateFirewallFromTemplate.ts
hkhalil-akamai: Removed the redundant `async`. The progress bar not animating on subsequent creations is because the firewall and template data is accessed from the cache. We can force the animation to start from 0 by using a `setTimeout` of 0 but also I believe most users will only generate a single firewall so won't see this behavior.

packages/manager/src/components/GenerateFirewallDialog/GenerateFirewallDialog.tsx
```diff
@@ -0,0 +1,217 @@
...
+  return (
+    <Stack gap={3}>
+      {dialogCopy && (
+        <Typography>
+          {replaceNewlinesWithLineBreaks(dialogCopy.text)}
+          {dialogCopy.link && (
+            <>
+              {' '}
+              <Link external to={dialogCopy.link?.url}>
+                {dialogCopy.link?.text}
```
carrillo-erik, line 115: I don't believe you need the `?.` on the two lines above based on the check on Line 111, both `url` and `text` are required in the `link` object.

packages/manager/src/features/Firewalls/FirewallDetail/index.tsx
carrillo-erik, line 1: Small nitpick and I know it's not part of your work, but this file contains both a named export (Line 37) and a default export (Line 193). Can we remove the default export please?

carrillo-erik (approved):
Looks good, nice work getting this completed.

packages/manager/src/components/GenerateFirewallDialog/useCreateFirewallFromTemplate.ts
hkhalil-akamai: Update: switched to fetching the specific template from the slug

packages/manager/src/components/GenerateFirewallDialog/useCreateFirewallFromTemplate.ts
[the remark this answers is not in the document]
bnussman-akamai: Fair point 👍

jaalah-akamai (approved):
I think the loader can be improved on the second generation of a FW, but otherwise this looks good. nice job!

packages/api-v4/src/firewalls/firewalls.ts
```diff
@@ -243,3 +244,32 @@ export const deleteFirewallDevice = (firewallID: number, deviceID: number) =>
       )}/devices/${encodeURIComponent(deviceID)}`
     )
   );
+
+// TEMPLATES
```
bnussman-akamai, line 248: Here's a neat trick I found. Maybe we could start using it

Screenshot 2024-08-15 at 10 51 46 AM

bnussman-akamai (approved):
Looks good! Nice job on this one

packages/api-v4/src/firewalls/firewalls.ts
hkhalil-akamai: This is cool! Doing this in this file

PR #10774, opened 2024-08-12
test: [M3-8432] - Cypress integration test for Object Storage Gen2: E0 endpoint, some additional unit tests

coliu-akamai:
## Description 📝
Add new tests for Object Storage gen2 - bucket create

## Changes  🔄
- Created new objStorageGen2 directory and new file, bucket-create-gen2.spec.ts for new integration tests
- Added unit tests for new create bucket drawer 

## Target release date 🗓️
n/a

## How to test 🧪
- yarn cy:run -s "cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts"
- yarn test OMC_CreateBucketDrawer.test.tsx
- yarn test BucketRateLimitTable.test.tsx

*Check all that apply*

- [x] 👀 Doing a self review


packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.tsx
```diff
@@ -44,6 +44,7 @@ export const BucketRateLimitTable = ({
     sx={{
       marginBottom: 3,
     }}
+    data-testid="bucket-rate-limit-table"
```
coliu-akamai, line 47: adding this so that we can confirm this table doesn't exist when selecting an E0 endpoint (@jdamore-linode - is it supposed to be data-qa...etc instead? I've seen both in the cypress tests so wasn't too sure)

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+    cy.wait('@createBucket').then((xhr) => {
+      const requestPayload = xhr.request.body;
+      expect(requestPayload['endpoint_type']).to.equal('E0');
+      expect(requestPayload['cors_enabled']).to.equal(true);
+    });
+
+    ui.drawer.find().should('not.exist');
+
+    // Confirm that bucket is created, initiate deletion for cleanup
+    // TODO: when M3-8304 is complete, confirm endpoint type and endpoint url are visible
```
coliu-akamai, line 148: M3-8304 is up for review rn, will be keeping an eye on it!

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+      objectStorageEndpointsFactory.build({
+        endpoint_type: 'E3',
+        region: 'us-sea',
+        s3_endpoint: null,
+      }),
+    ];
+
+    mockAppendFeatureFlags({
+      objMultiCluster: makeFeatureFlagData(true),
+      objectStorageGen2: makeFeatureFlagData(true),
```
coliu-akamai, line 63: I'm still having the relevant feature flags coming in as `{objectStorageGen2: true, objMultiCluster: true}` rather than 
`{objectStorageGen2: {beta: false, enabled: true}, objMultiCluster: true}`

 I had to change line 98 in [useObjectStorageEndpoints](https://github.com/linode/manager/blob/develop/packages/manager/src/queries/object-storage/queries.ts) locally when testing this (see my comment on #10744) 😖. Left it as is when pushing up my changes, but if this causes issues will investigate more!

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+        region: 'us-sea',
+        s3_endpoint: null,
+      }),
+    ];
+
+    mockAppendFeatureFlags({
+      objMultiCluster: makeFeatureFlagData(true),
+      objectStorageGen2: makeFeatureFlagData<Flags['objectStorageGen2']>({
+        enabled: true,
+      }),
```
coliu-akamai, line 66: this should take care of [my previous comment](https://github.com/linode/manager/pull/10774#discussion_r1715570162) of the test potentially not working, actually! 🤞 

One thing to note - objectStorageGen2 is typed as a [BaseFeatureFlag](https://github.com/linode/manager/blob/9d082055da7f5f82a9cea494d142ccaca4c9add1/packages/manager/src/featureFlags.ts#L49) but seems to have fields of a [BetaFeatureFlag](https://github.com/linode/manager/blob/9d082055da7f5f82a9cea494d142ccaca4c9add1/packages/manager/src/featureFlags.ts#L53) (additional beta field) - is this intentional? Otherwise am happy to change its [type](https://github.com/linode/manager/blob/9d082055da7f5f82a9cea494d142ccaca4c9add1/packages/manager/src/featureFlags.ts#L107) to BetaFeatureFlag

packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const getHeader = getByText('GET');
+    expect(getHeader).toBeVisible();
+    const putHeader = getByText('PUT');
+    expect(putHeader).toBeVisible();
+    const listHeader = getByText('LIST');
+    expect(listHeader).toBeVisible();
+    const deleteHeader = getByText('DELETE');
+    expect(deleteHeader).toBeVisible();
+    const otherHeader = getByText('OTHER');
+    expect(otherHeader).toBeVisible();
```
jaalah-akamai, line 25: Nit: I think if this wasn't a test, you might have done something similar to:
```
const headers = ['Limits', 'GET', 'PUT', 'LIST', 'DELETE', 'OTHER'];

headers.forEach((header) => {
  expect(screen.getByText(header)).toBeInTheDocument();
});
```

packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+    // ensure table cells render as expected
+    const limitValue000 = getAllByText('000');
+    expect(limitValue000).toHaveLength(8);
+    const limitValue1000 = getByText('1000');
+    expect(limitValue1000).toBeVisible();
+    const limitValue5000 = getByText('5000');
+    expect(limitValue5000).toBeVisible();
```
jaalah-akamai, line 33: These are somewhat malleable data points at least for now, I don't want these to be the reason the test breaks in the future. What we do know is that there will only be 2 rows, basic / high. So maybe we can check if 3 rows exist?
```
const rows = screen.getAllByRole('row');
// 1 header row + 2 data rows
expect(rows).toHaveLength(3);
```

packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+    const limitValue000 = getAllByText('000');
+    expect(limitValue000).toHaveLength(8);
+    const limitValue1000 = getByText('1000');
+    expect(limitValue1000).toBeVisible();
+    const limitValue20000 = getByText('20000');
+    expect(limitValue20000).toBeVisible();
+    const limitValue5000 = queryByText('5000');
+    expect(limitValue5000).not.toBeInTheDocument();
+  });
```
jaalah-akamai, line 49: For now I would check that 5,000 changes to 20,000 when selecting an E3

packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.test.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
+
+import { renderWithTheme } from 'src/utilities/testHelpers';
+
+import { BucketRateLimitTable } from './BucketRateLimitTable';
+
+describe('BucketRateLimitTable', () => {
+  it('should render a BucketRateLimitTable', () => {
+    const { getAllByText, getByText } = renderWithTheme(
+      <BucketRateLimitTable endpointType="E1" />
```
jaalah-akamai, line 10: Table will never show for anything other than E2 or E3
```suggestion
      <BucketRateLimitTable endpointType="E2" />
```

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
jaalah-akamai: There will technically be no beta period for CM. I've been ignoring that field since I didn't want to go back through and update the flag everywhere.

jaalah-akamai (approved):
✅ Overall this looks good, all tests passed and e2e looks good through my non-SDET eyes, but I'll let one of the @linode/frontend-sdet comment on the rest

packages/manager/cypress/support/intercepts/object-storage.ts
```diff
@@ -485,6 +486,12 @@ export const interceptUpdateBucketAccess = (
...
-export const interceptGetObjectStorageEndpoints = (): Cypress.Chainable<null> => {
-  return cy.intercept('GET', apiMatcher(`object-storage/endpoints`));
+export const interceptGetObjectStorageEndpoints = (
+  endpoints: ObjectStorageEndpoint[]
+): Cypress.Chainable<null> => {
+  return cy.intercept(
+    'GET',
+    apiMatcher(`object-storage/endpoints*`),
+    paginateResponse(endpoints)
+  );
```
jdamore-linode, line 496: I think we'll want to rename this to `mockGetObjectStorageEndpoints` -- we use the `intercept` prefix for utils that intercept requests but don't mock the response e.g. so we can wait for requests to resolve, or inspect the outgoing request/incoming response, etc..

(Don't forget to document the `endpoints` param too!)

AzureLatte (approved):
Test passed

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
```diff
@@ -0,0 +1,176 @@
...
```
jdamore-linode, line 66:
```suggestion
      objMultiCluster: true,
      objectStorageGen2: { enabled: true },
```
You _should_ be able to do it this way now -- we made the feature flag mocking a little bit more ergonomic in #10635 (you can also get rid of the call to `mockGetFeatureFlagClientstream`!)

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+   * Confirms S3 endpoint hostname displayed to differentiate between identical options in the dropdown
+   */
+  it('can create a bucket with endpoint type 0', () => {
+    const bucketLabel = randomLabel();
+    const bucketRegion = 'Seattle, WA';
+    const bucketCluster = 'us-sea';
+
+    interceptGetBuckets().as('getBuckets');
+    interceptDeleteBucket(bucketLabel, bucketCluster).as('deleteBucket');
+    interceptCreateBucket().as('createBucket');
```
jdamore-linode, line 32: We want to mock all the bucket operations for this test, so we'll probably want to use `mockGetBuckets`/`mockDeleteBucket`/`mockCreateBuckets` here instead

jdamore-linode (commented):
Noticed one little thing! Hoping to review this properly in the next day or so but no need to hold up merging once that's taken care of

packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.tsx
jdamore-linode: @coliu-akamai there's no hard rule here! What you have here is totally fine.

`data-testid` is a [Testing Library](https://testing-library.com/) convention, so it works nicely with functions like `getByTestId()`, `queryByTestId()`, etc. in Vitest land and `cy.findByTestId()` in Cypress land. If you only need to identify some element so it can be selected from a test, this is a good go-to.

I believe `data-qa-*` is a custom convention of ours. It's mostly useful for selecting elements in Cypress tests via `cy.get('[data-qa-my-component]')`, but its main advantage is that you can use the attribute name to identify the element and its value to contain some other data, and make selections or assertions based on that data. That makes it really helpful for cases where there may be multiple of some element on the screen and you need to select a single one. You can do some other tricks with it, but that's the main way we use it.

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
coliu-akamai: Updated to use mock requests instead of intercepts! I had to make a change to the `makeResponse` portion of `mockCreateBucket`, otherwise would get a bunch of type errors when trying to render a bucket row. There's definitely a lot I'm not familiar with for integration tests, so would love to schedule a call to review at some point (kinda in line with the SME meetings we discussed at retro actually!)

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+  /**
+   * Confirms UI flow for creating a gen2 Object Storage bucket with endpoint E0
+   * Confirms all endpoints are displayed regardless if there's multiple of the same type
+   * Confirms S3 endpoint hostname displayed to differentiate between identical options in the dropdown
+   */
+  it('can create a bucket with endpoint type 0', () => {
+    const endpointTypeE0 = 'Legacy (E0)';
+    const bucketLabel = randomLabel();
+    const bucketRegionSelect = 'US, Seattle, WA';
+    const bucketRegion = 'us-sea';
```
jdamore-linode, line 28: Very minor nit! I'd recommend mocking the regions when possible. In this case, it'd look something like this:

```typescript
// Requires importing `regionFactory` (src/factories), `chooseRegion` (support/util/regions), and `mockGetRegions` (support/intercepts/regions)
const mockRegions = regionFactory.buildList(10, { capabilities: ['Object Storage'] });
const mockRegion = chooseRegion({ regions: mockBucketRegions });

// ... later, sometime before `cy.visitWithLogin()`:
mockGetRegions(mockRegions);
```

Then elsewhere in the test, you can refer to the region's ID via `mockRegion.id` and its label (as shown in Cloud) via `mockRegion.label`

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+
+    mockAppendFeatureFlags({
+      objMultiCluster: true,
+      objectStorageGen2: { enabled: true },
+    }).as('getFeatureFlags');
+    mockGetAccount(
+      accountFactory.build({
+        capabilities: [
+          'Object Storage Endpoint Types',
+          'Object Storage Access Key Regions',
```
jdamore-linode, line 75: Another nit!

```suggestion
          'Object Storage',
          'Object Storage Endpoint Types',
          'Object Storage Access Key Regions',
```
Let's add the `Object Storage` capability to this mock, too! Even though it's not strictly required to get the test passing, it might be helpful from a tests-as-docs point of view since it signals that all 3 capabilities are expected to be present together when interacting with OBJ Gen 2 features.

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+        endpoint_type: 'E2',
+        region: bucketRegion,
+        s3_endpoint: null,
+      }),
+      objectStorageEndpointsFactory.build({
+        endpoint_type: 'E3',
+        region: bucketRegion,
+        s3_endpoint: null,
+      }),
+    ];
```
jdamore-linode, line 65: Super minor nit, but I like to define all the test-specific mock data (and any other vars) at the top of the tests when possible, followed by mock/intercept setup via the `mockGetWhatever()` functions, and then finally `cy.visitWithLogin()` and all the test interactions. It makes it easier to see how all the mock data is set up at a glance, which is super useful when troubleshooting since a lot of test flaws come down to data setup.

In this case, though, we can probably even move `mockEndpoints` to the very top scope since we'll eventually have 3 or 4 more tests in this spec and most will probably be able to use this same mock data.

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+          .click();
+      });
+
+    //wait for the newly 'created' mocked bucket to appear
+    const mockBucket = objectStorageBucketFactoryGen2.build({
+      label: bucketLabel,
+      region: bucketRegion,
+      endpoint_type: 'E0',
+      s3_endpoint: undefined,
+    });
```
jdamore-linode, line 149: Pretty much a repeat of my other suggestion, let's move this up to the top of the test!

jdamore-linode (approved):
Nice work @coliu-akamai! Tests look great! Shared some really minor suggestions, but nothing at all that impacts the test's quality or effectiveness

coliu-akamai:
Unskipped the unit test that was failing here (oh man I need to relearn rebase) - we will get to the bottom of this failing test!! :D

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
pmakode-akamai: Hey Connie (@coliu-akamai), I think you might be facing the same issue I encountered locally with the `Object Gen 2` flag. After some investigation, I was able to identify a possible cause. You can see my comment and the related thread from the other PR [here](https://github.com/linode/manager/pull/10797#discussion_r1729460553). Hope this helps!

coliu-akamai:
Some updates:
- adding additional timeout to the failing OMC_CreateBucketDrawer test - this is honestly just a bandaid solution, but will continue investigating the underlying issue
- skipping BucketRateLimitTableTests for now - tests are now failing + table crashes the page on the Create Bucket Drawer now, will investigate the issue separately

Will be merging after github status checks (especially for test-manager - will run this a few times) pass

PR #10725, opened 2024-07-30
refactor: [M3-6913] Replace Select with Autocomplete in: linodes

harsh-akamai:
## Description 📝
We want to get rid of our dependency on react-select for accessibility reasons and to consolidate our usage of third-party libraries.

## Changes  🔄
- Replaced all instances of Select with Autocomplete in the Linode feature.

### Verification steps
(How to verify changes)
- Navigate to [http://localhost:3000/linodes](http://localhost:3000/linodes)
- Verify the select input fields remains unchanged when using Autocomplete

## As an Author I have considered 🤔

*Check all that apply*

- [x] 👀 Doing a self review
- [x] ❔ Our [contribution guidelines](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md)
- [ ] 🤏 Splitting feature into small PRs
- [x] ➕ Adding a changeset
- [x] 🧪 Providing/Improving test coverage
- [ ] 🔐 Removing all sensitive information from the code and PR description
- [ ] 🚩 Using a feature flag to protect the release
- [ ] 👣 Providing comprehensive reproduction steps
- [x] 📑 Providing or updating our documentation
- [ ] 🕛 Scheduling a pair reviewing session
- [ ] 📱 Providing mobile support
- [ ] ♿  Providing accessibility support

PR #10783, opened 2024-08-13
fix: 'Oh snap' error when creating VPC from Linode Create

coliu-akamai:
## Description 📝
When the Linode Create v2 feature flag is not checked, trying to create a VPC from Linode Create results in the page crashing
image

## Changes  🔄
- Update value field in VPC panel's autocomplete to ensure 'undefined' is never the value

## Target release date 🗓️
8/14 - trying to get it into this upcoming release, as the bug it fixes is still unreleased!

### Reproduction steps
- Make sure Linode Create v2 flag is _not_ checked
- Try creating a VPC from the Linode Create page 

### Verification steps
- Verify that creating a VPC from the Linode Create page does not crash the page when Linode Create v2 flag isn't checked
- Double check that when Linode Create v2 flag _is_ checked, everything still works as expected

*Check all that apply*

- [x] 👀 Doing a self review

---

**Commit Types:**

---

### DIFF
--- a/packages/manager/src/features/Linodes/LinodesCreate/VPCPanel.tsx
+++ b/packages/manager/src/features/Linodes/LinodesCreate/VPCPanel.tsx
@@ -133,6 +133,8 @@ export const VPCPanel = (props: VPCPanelProps) => {
     );
   }, [vpcs, region, fromLinodeCreate]);
 
+  const defaultVPCValue = fromLinodeConfig ? null : vpcDropdownOptions[0];
+
   const subnetDropdownOptions: DropdownOption[] =
     vpcs
       .find((vpc) => vpc.id === selectedVPCId)
@@ -198,9 +200,6 @@ export const VPCPanel = (props: VPCPanelProps) => {
         <Stack>
           <Typography>{getMainCopyVPC()}</Typography>
           <Autocomplete
-            isOptionEqualToValue={(option, value) => {
-              return option.label === value.label;
-            }}
             onChange={(_, selectedVPC) => {
               handleSelectVPC(selectedVPC?.value || -1);
               // Track clearing and changing the value once per page view, configured by inputValue in AA backend.
@@ -228,11 +227,9 @@ export const VPCPanel = (props: VPCPanelProps) => {
                 ? vpcDropdownOptions.find(
                     (option) => option.value === selectedVPCId
                   ) || null
-                : null
+                : defaultVPCValue
             }
             autoHighlight
-            clearIcon={null}
-            defaultValue={fromLinodeConfig ? null : vpcDropdownOptions[0]} // If we're in the Config dialog, there is no "None" option at index 0
             disabled={!regionSupportsVPCs}
             errorText={vpcIdError ?? vpcError}
             label={from === 'linodeCreate' ? 'Assign VPC' : 'VPC'}
--- a/packages/manager/src/features/Linodes/LinodesCreate/VPCPanel.tsx
+++ b/packages/manager/src/features/Linodes/LinodesCreate/VPCPanel.tsx
@@ -226,7 +226,7 @@ export const VPCPanel = (props: VPCPanelProps) => {
               selectedVPCId && selectedVPCId !== -1
                 ? vpcDropdownOptions.find(
                     (option) => option.value === selectedVPCId
-                  ) || null
+                  ) ?? null
                 : defaultVPCValue
             }
             autoHighlight
--- a/packages/manager/src/features/Linodes/LinodesCreate/VPCPanel.tsx
+++ b/packages/manager/src/features/Linodes/LinodesCreate/VPCPanel.tsx
@@ -199,7 +199,7 @@ export const VPCPanel = (props: VPCPanelProps) => {
           <Typography>{getMainCopyVPC()}</Typography>
           <Autocomplete
             isOptionEqualToValue={(option, value) => {
-              return option.label === value?.label;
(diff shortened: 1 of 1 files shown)


packages/manager/src/features/Linodes/LinodesCreate/VPCPanel.tsx
```diff
@@ -227,7 +227,7 @@ export const VPCPanel = (props: VPCPanelProps) => {
               selectedVPCId && selectedVPCId !== -1
                 ? vpcDropdownOptions.find(
                     (option) => option.value === selectedVPCId
-                  )
+                  ) || null
```
coliu-akamai, line 230: Originally, I'd changed line 202 to be `return option.label === value?.label;`, since the bug is caused by the value being undefined. I thought that would be a super quick fix 😅 but this led to two MUI errors:
- controlled changing to uncontrolled input error
- "The value provided to Autocomplete is invalid. None of the options match with `undefined`."

ngl it took me a hot second to remember that array.find can return an undefined value oops 😅  so this change ensures that the item being passed to the value prop is always 'null' or some vpc dropdown option, not undefined. This fixes both the original problem and gets rid of the two MUI warnings above too! 

This originally slipped past likely because the value prop is an optional parameter, so it can technically still be undefined (but we won't be passing it an undefined value).

image
image

coliu-akamai: (I've also been trying to play around with making it look cleaner so if anyone has suggestions lmk 😆)

bnussman-akamai:
I see a few regressions still that we'll want to address
- We should probably add `disableClearable` 
- We should probably remove the `clearIcon` prop
- We can probably remove `isOptionEqualToValue`

I'm thinking we can delete the `defaultValue` prop  in favor of this for the `value`:

```typescript
            value={
              selectedVPCId && selectedVPCId !== -1
                ? vpcDropdownOptions.find(
                    (option) => option.value === selectedVPCId
                  )
                : vpcDropdownOptions[0]
            
```

Linode Create v2 handled the "empty" state better by using a placeholder instead of a "None" option, but I think these steps will get Linode Create v1 back to what is use to be

coliu-akamai:
👀 oh sweet thanks Banks! There does seem to be a slight change in initial behavior with the panel inside the Edit Config dialogue with those changes - the first VPC option gets selected instead of a default no option selected. Maybe this is fine though, since a person editing their Linode's configs to include a VPC wouldn't leave the VPC as 'none' anyway? 

local (also going to look into the styling change)
image

prod
image

coliu-akamai:
> Maybe this is fine though, since a person editing their Linode's configs to include a VPC wouldn't not select a VPC anyway?

Looks like we do need to keep the default value for the Linode config, otherwise we run into the strange case where it seems we've selected a VPC, but can't select subnets for that VPC
image


I probably missing a few things but here were my thoughts on what I just pushed up:
- adding `disableClearable` prevented me from using `null`, which it seems we need for the Edit Config portion - using undefined led to controlled vs not controlled errors (?)
- essentially moved the value of `defaultValue` to be inside the `value` prop - so that initial values inside Edit Config and Linode Create should both now match prod, hopefully
- removed `clearIcon` and `isOptionEqualToValue` props
- there's a weird flicker when creating a VPC in the autocomplete compared to prod - I'm not too sure how to get rid of it tbh
- style regression still not fixed 😢 

I think my takeaway from this is that Linode Create v2 sounds really great 😅

mjac0bs:
> * adding `disableClearable` prevented me from using `null`, which it seems we need for the Edit Config portion - using undefined led to controlled vs not controlled errors (?)

This field was [a point of **trouble**](https://github.com/linode/manager/pull/10725#discussion_r1701791209) in the original PR too. Linking for context in case you hadn't read.

coliu-akamai:
oh man, thanks for pointing that out @mjac0bs! I'm gonna add back in the `clearIcon` prop then + recheck everything

packages/manager/src/features/Linodes/LinodesCreate/VPCPanel.tsx
```diff
@@ -227,12 +226,11 @@ export const VPCPanel = (props: VPCPanelProps) => {
               selectedVPCId && selectedVPCId !== -1
                 ? vpcDropdownOptions.find(
                     (option) => option.value === selectedVPCId
-                  )
-                : null
+                  ) || null
```
hkhalil-akamai, line 229:
```suggestion
                  ) ?? null
```

It might be better to use the [null-coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) (`??`) here but either way is fine since `DropdownOption` can never be falsey anyways. Just a good habit I guess (since using `||` can lead to unexpected behavior when working with strings and numbers for example).

hkhalil-akamai (approved):
Great catch and thanks for addressing this promptly! Verified `VPCPanel` works as expected across the app.

packages/manager/src/features/Linodes/LinodesCreate/VPCPanel.tsx
coliu-akamai: done + ty! 🚀

bnussman-akamai (approved):
Looks good!

I still think `clearIcon` is a "hack" for not using `disableClearable` from a type safety perspective, but this seems to be working alright.

jaalah-akamai (approved):

PR #10784, opened 2024-08-14
feat: [M3-8170] - Add CheckoutBar Story

pmakode-akamai:
## Description 📝
Add CheckoutBar Story to Storybook

## Changes  🔄
- Add CheckoutBar Story

## Target release date 🗓️
N/A

### Verification steps
- Run yarn storybook to test locally

*Check all that apply*

- [x] 👀 Doing a self review

### DIFF
--- a/packages/manager/src/components/CheckoutBar/CheckoutBar.stories.tsx
+++ b/packages/manager/src/components/CheckoutBar/CheckoutBar.stories.tsx
@@ -1,4 +1,7 @@
+import { styled } from '@mui/material/styles';
 import React from 'react';
+
+import { Typography } from 'src/components/Typography';
 
 import { CheckoutBar } from './CheckoutBar';
 
@@ -7,8 +10,22 @@ import type { Meta, StoryFn, StoryObj } from '@storybook/react';
 type Story = StoryObj<typeof CheckoutBar>;
 
 const Item = ({ children }: { children?: React.ReactNode }) => (
-  <div style={{ color: 'white', fontStyle: 'italic' }}>{children}</div>
+  <StyledItem>{children}</StyledItem>
 );
+
+const StyledItem = styled(Typography, {
+  label: 'StyledItem',
+})(() => ({
+  fontStyle: 'italic',
+}));
+
+const defaultArgs = {
+  calculatedPrice: 30.0,
+  children: <Item>Child items can go here!</Item>,
+  heading: 'Checkout',
+  onDeploy: () => alert('Deploy clicked'),
+  submitText: 'Submit',
+};
 
 const meta: Meta<typeof CheckoutBar> = {
   argTypes: {
@@ -28,39 +45,33 @@ const meta: Meta<typeof CheckoutBar> = {
 export default meta;
 
 export const Default: Story = {
-  args: {
-    calculatedPrice: 30.0,
-    children: <Item>Child items can go here!</Item>,
-    heading: 'Checkout',
-    onDeploy: () => alert('Deploy clicked'),
-    submitText: 'Submit',
-  },
+  args: defaultArgs,
 };
 
 export const WithAgreement: Story = {
   args: {
-    ...Default.args,
+    ...defaultArgs,
     agreement: <Item>Agreement item can go here!</Item>,
   },
 };
 
 export const Disabled: Story = {
   args: {
-    ...Default.args,
+    ...defaultArgs,
     disabled: true,
   },
 };
 
 export const Loading: Story = {
   args: {
-    ...Default.args,
+    ...defaultArgs,
     isMakingRequest: true,
   },
 };
 
 export const WithFooter: Story = {
   args: {
-    ...Default.args,
+    ...defaultArgs,
     footer: <Item>Footer element can go here!</Item>,
   },
 };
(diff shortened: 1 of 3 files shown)


packages/manager/src/components/CheckoutBar/CheckoutBar.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+    ...Default.args,
+    isMakingRequest: true,
+  },
+};
+
+export const WithFooter: Story = {
+  args: {
+    ...Default.args,
+    footer: <Item>Footer element can go here!</Item>,
+  },
```
coliu-akamai, line 65: nitpick: could include a custom render function (ie `render: (args) => <CheckoutBar {...args} />`). It might not be necessary in this case if storybook's default render function is doing the same thing tho - does anyone know? I've just noticed that we tend to include `render` even if it's as simple as above.

packages/manager/src/components/CheckoutBar/CheckoutBar.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+import { CheckoutBar } from './CheckoutBar';
+
+import type { Meta, StoryFn, StoryObj } from '@storybook/react';
+
+type Story = StoryObj<typeof CheckoutBar>;
+
+const Item = ({ children }: { children?: React.ReactNode }) => (
+  <div style={{ color: 'white', fontStyle: 'italic' }}>{children}</div>
+);
```
coliu-akamai, line 11: optional: could create a styled component instead like the [ones in here](https://github.com/linode/manager/blob/develop/packages/manager/src/components/CheckoutBar/styles.ts)

coliu-akamai (commented):
thanks for the PR! Had a few thoughts/comments

- Could we change the text color of the children and agreement/footers? I can't see them unless I highlight the text. Maybe a color that works in both light/dark mode, or use the theme
image

- Could we add some documentation for the props of CheckoutBar so that we can see their descriptions in the story? For example ([see corresponding documentation](https://github.com/linode/manager/blob/7341d1673d6384ab85128748a30292a30569403e/packages/manager/src/components/ActionMenu/ActionMenu.tsx#L19))
image

- Could we also add some tests for the CheckoutBar component?

We did a whole bunch of storybook migrations late last year - some of these PRs might be helpful to look at: [here](https://github.com/linode/manager/pulls?q=is%3Apr+author%3Aabailly-akamai+label%3AStorybook) and [here](https://github.com/linode/manager/pulls?q=is%3Apr+author%3Acoliu-akamai+label%3AStorybook)

packages/manager/src/components/CheckoutBar/CheckoutBar.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+    children: <Item>Child items can go here!</Item>,
+    heading: 'Checkout',
+    onDeploy: () => alert('Deploy clicked'),
+    submitText: 'Submit',
+  },
+};
+
+export const WithAgreement: Story = {
+  args: {
+    ...Default.args,
```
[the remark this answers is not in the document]
pmakode-akamai: Got it! I’ll refactor to use a separate constant for Default.args

packages/manager/src/components/CheckoutBar/CheckoutBar.stories.tsx
pmakode-akamai: I believe a render function is optional and we don’t need to add the render function unless we want to customize the rendering.

packages/manager/src/components/CheckoutBar/CheckoutBar.stories.tsx
pmakode-akamai: I’ve tried using styled components in the children argument (e.g., `args: {children: <StyledComponent />}`), but Storybook wasn’t rendering them. It worked for the `const Item` I added earlier.

If we want to use styled components, we have to include them directly in the render function like this:
```
 render: (args) => (
    <CheckoutBar {...args}>
      <StyledItem>Child items can go here!</StyledItem>
    </CheckoutBar>
  ),
````

**Comment by pmakode-akamai** [comment]
> * Could we also add some tests for the CheckoutBar component?

Sure @coliu-akamai, Could we have a separate ticket for adding tests for the `CheckoutBar` component?

File: packages/manager/src/components/CheckoutBar/CheckoutBar.stories.tsx [hunk 4]
```diff
@@ -0,0 +1,77 @@
...
+
+const Item = ({ children }: { children?: React.ReactNode }) => (
+  <StyledItem>{children}</StyledItem>
+);
+
+const StyledItem = styled(Typography, {
+  label: 'StyledItem',
+})(() => ({
+  fontStyle: 'italic',
+}));
```
coliu-akamai, line 20: ahh I'd been hoping that we could use the styled component directly instead of needing `Item` but that doesn't seem to be the case 🥲 

in that case, it might be better to just remove the styled component to avoid the extra import for `styled`? sorry about that! 😅 

```
const Item = ({ children }: { children?: React.ReactNode }) => (
  <Typography sx={{ fontStyle: 'italic' }}>{children}</Typography>
);
```

coliu-akamai (approved):
sure, we can go with a separate ticket for tests - would you be able to create one? Thanks for the changes + good work!

left one small comment below 😅

pmakode-akamai:
> sure, we can go with a separate ticket for tests - would you be able to create one? Thanks for the changes + good work!

@coliu-akamai I have created a new ticket (M3-8463) for tests. Keeping this one as a todo..

bnussman-akamai (approved):

PR #10788, opened 2024-08-15
test: [M3-8098] - Refactor StackScript create test to be resilient to Image deprecations

AzureLatte:
## Description 📝
Update StackScript create test to be more resilient to the changes that Images may be deprecated on the backend by using real Image data retrieved directly from the API.

## Changes  🔄
- ...
- ...

*Check all that apply*

- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/cypress/e2e/core/stackscripts/create-stackscripts.spec.ts
+++ b/packages/manager/cypress/e2e/core/stackscripts/create-stackscripts.spec.ts
@@ -17,6 +17,9 @@ import { chooseRegion } from 'support/util/regions';
 import { SimpleBackoffMethod } from 'support/util/backoff';
 import { cleanUp } from 'support/util/cleanup';
 import { createTestLinode } from 'support/util/linodes';
+import { interceptGetAllImages } from 'support/intercepts/images';
+import type { Image } from '@linode/api-v4';
+import { getFilteredImagesForImageSelect } from 'src/components/ImageSelectv2/utilities';
 
 // StackScript fixture paths.
 const stackscriptBasicPath = 'stackscripts/stackscript-basic.sh';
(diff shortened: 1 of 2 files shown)


jaalah-akamai (approved):

mjac0bs (commented):
Not totally certain what's happening here, because this test seems to be failing due to a regions-related issue. I get consistent failures on the first test case locally. If I skip that first test case and only run the second (the only one this PR touches), that *did* pass. Another case of M3-8471? cc @jdamore-linode 


Screenshot 2024-08-22 at 10 39 38 AM

image

When I checked CI, I see that the first and third runners had failing tests. The test failure on the third runner is expected and fixed recently in #10811. I can't see the failures on the first runner due to some CI UI glitch, but by searching for this PR number (`10788`) in our notifications Slack channel, it looks like this spec is failing there too.

jdamore-linode:
> Another case of M3-8471? cc @jdamore-linode

It sounds like maybe yes _and_ no... when I run the test locally I don't have any issue with Cypress finding the region by its ID, but I _do_ still get failures pretty often but not consistently (seems kinda 50/50). But the fact that it's failing consistently for you makes it sound like you could be experiencing the ID issue -- were you running Cloud with `yarn dev`/`yarn up`, or via `yarn start:manager:ci`? Wondering if you get the same outcome?

In my case, the issue seems to be caused by these two factors:

- Sometimes the (random) region Cypress tries to select is not visible in the drop-down because of overflow, so we just have to scroll it into view before trying to click on it
- There seems to be some kind of Cloud issue where the selected region doesn't get honored if it's made really quickly after page load (e.g. as if it's waiting for some API request to resolve). If Cypress selects a region too quickly, the autocomplete drop-down gets dismissed but the region select remains with no value selected. It _seems_ like waiting for the account availability request to resolve before interacting fixes the issue, but that could be incidental

Either way, I'll have a ticket and PR open later this afternoon, and will circle back and review @AzureLatte's improvements to this other test too! We can disregard this failure for the sake of PR review since we are seeing this happen in CI where Azure's changes aren't present.

mjac0bs (approved):
Approving since test failures (which are addressed in follow-up PRs) are unrelated to these changes, which look good and pass consistently.

AzureLatte:
Thank you all for the review!

PR #10791, opened 2024-08-16
change: [M3-6843] - Move manual snapshot error message to snapshot confirmation dialog

coliu-akamai:
## Description 📝
Moves the manual snapshot error message from the Linode Backups page to the snapshot's confirmation dialog

## Target release date 🗓️
n/a

## Preview 📷
Note - three potential options for displaying the error message - see "Other notes/questions" for more
| Before  | After   |
| ------- | ------- |
| image| imageimageimage|

## How to test 🧪
- Have a busy linode (a linode that is already creating a snapshot, shutting down, turning on, etc) or a linode with no disk and no config and try to create a snapshot
- Confirm error message is now in the dialog
- Confirm that error message clears when closing and then reopening dialog

## Other notes/questions
- several potential ways to display the snapshot error in the confirmation dialog
  - the first (current) one matches ~the rest~ most of the confirmation dialogs 
  - other two more closely resemble how the error message used to appear. 
  - I'm thinking that the first option is probably the way to go for consistency? but can change it if the other options are preferred
- should we add a toast for disk deletion initiation? [_See internal ticket comments for context_]

*Check all that apply*

- [x] 👀 Doing a self review

### DIFF
--- a/packages/manager/src/features/Linodes/LinodesDetail/LinodeBackup/CaptureSnapshot.tsx
+++ b/packages/manager/src/features/Linodes/LinodesDetail/LinodeBackup/CaptureSnapshot.tsx
@@ -53,12 +53,6 @@ export const CaptureSnapshot = (props: Props) => {
 
   const hasErrorFor = getErrorMap(['label'], snapshotError);
 
-  React.useEffect(() => {
-    if (isSnapshotConfirmationDialogOpen) {
-      reset();
-    }
-  }, [isSnapshotConfirmationDialogOpen, reset]);
-
   return (
     <Paper>
       <Typography data-qa-manual-heading variant="h2">
@@ -95,6 +89,7 @@ export const CaptureSnapshot = (props: Props) => {
         error={hasErrorFor.none}
         loading={isSnapshotLoading}
         onClose={() => setIsSnapshotConfirmationDialogOpen(false)}
+        onExited={() => reset()}
         onSnapshot={() => snapshotForm.handleSubmit()}
         open={isSnapshotConfirmationDialogOpen}
       />
--- a/packages/manager/src/features/Linodes/LinodesDetail/LinodeBackup/CaptureSnapshot.tsx
+++ b/packages/manager/src/features/Linodes/LinodesDetail/LinodeBackup/CaptureSnapshot.tsx
@@ -6,7 +6,6 @@ import * as React from 'react';
 import { Box } from 'src/components/Box';
 import { Button } from 'src/components/Button/Button';
 import { FormControl } from 'src/components/FormControl';
-import { Notice } from 'src/components/Notice/Notice';
 import { Paper } from 'src/components/Paper';
 import { TextField } from 'src/components/TextField';
 import { Typography } from 'src/components/Typography';
@@ -65,19 +64,14 @@ export const CaptureSnapshot = (props: Props) => {
         manual snapshot will not be overwritten by automatic backups.
       </Typography>
       <FormControl>
-        {hasErrorFor.none && (
-          <Notice spacingBottom={8} variant="error">
-            {hasErrorFor.none}
-          </Notice>
-        )}
         <StyledBox>
           <TextField
-            sx={{ minWidth: 275 }}
             data-qa-manual-name
             errorText={hasErrorFor.label}
             label="Name Snapshot"
             name="label"
             onChange={snapshotForm.handleChange}
+            sx={{ minWidth: 275 }}
             value={snapshotForm.values.label}
           />
           <Button
@@ -91,6 +85,7 @@ export const CaptureSnapshot = (props: Props) => {
         </StyledBox>
       </FormControl>
       <CaptureSnapshotConfirmationDialog
+        error={hasErrorFor.none}
         loading={isSnapshotLoading}
         onClose={() => setIsSnapshotConfirmationDialogOpen(false)}
         onSnapshot={() => snapshotForm.handleSubmit()}
--- a/packages/manager/src/features/Linodes/LinodesDetail/LinodeBackup/CaptureSnapshot.tsx
+++ b/packages/manager/src/features/Linodes/LinodesDetail/LinodeBackup/CaptureSnapshot.tsx
@@ -30,6 +30,7 @@ export const CaptureSnapshot = (props: Props) => {
     error: snapshotError,
     isLoading: isSnapshotLoading,
     mutateAsync: takeSnapshot,
+    reset,
   } = useLinodeBackupSnapshotMutation(linodeId);
 
   const [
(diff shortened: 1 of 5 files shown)


packages/manager/src/features/Linodes/LinodesDetail/LinodeBackup/LinodeBackups.tsx
```diff
@@ -23,13 +22,15 @@ import { useRegionsQuery } from 'src/queries/regions/regions';
 import { useTypeQuery } from 'src/queries/types';
 import { getMonthlyBackupsPrice } from 'src/utilities/pricing/backups';
 
-import { BackupTableRow } from './BackupTableRow';
 import { BackupsPlaceholder } from './BackupsPlaceholder';
+import { BackupTableRow } from './BackupTableRow';
```
coliu-akamai, line 26: these are just eslint changes

packages/manager/src/features/Linodes/LinodesDetail/LinodeBackup/CaptureSnapshotConfirmationDialog.tsx
```diff
@@ -5,7 +5,7 @@ import { ConfirmationDialog } from 'src/components/ConfirmationDialog/Confirmati
 import { Typography } from 'src/components/Typography';
 
 interface Props {
-  error?: string;
+  error: string | undefined;
```
coliu-akamai, line 8: there's no change in type/functionality of this prop, just made it non optional bc the one place we use this component now passes in an error

packages/manager/src/features/Linodes/LinodesDetail/LinodeBackup/CaptureSnapshotConfirmationDialog.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+    const title = getByText('Take a snapshot?');
+    const body = getByText(
+      'Taking a snapshot will back up your Linode in its current state, overriding your previous snapshot. Are you sure?'
+    );
+
+    expect(takeSnapshotButton).toBeVisible();
+    expect(cancelButton).toBeVisible();
+    expect(title).toBeVisible();
+    expect(body).toBeVisible();
+  });
```
mjac0bs, line 37:
```suggestion
  });
  
```
Nit: could we add spaces between each test case in this file? Small change, but helps with readability.

packages/manager/src/features/Linodes/LinodesDetail/LinodeBackup/CaptureSnapshotConfirmationDialog.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+    const { getByTestId } = renderWithTheme(
+      <CaptureSnapshotConfirmationDialog {...props} loading={true} />
+    );
+
+    const takeSnapshotButton = getByTestId('confirm');
+    expect(takeSnapshotButton).toBeVisible();
+    fireEvent.click(takeSnapshotButton);
+    expect(props.onSnapshot).not.toHaveBeenCalled();
+  });
+  it('should close the dialog', () => {
```
mjac0bs, line 58:
```suggestion
  it('should close the dialog when 'Cancel' button is clicked', () => {
```

packages/manager/src/features/Linodes/LinodesDetail/LinodeBackup/CaptureSnapshot.tsx
```diff
@@ -53,6 +53,12 @@ export const CaptureSnapshot = (props: Props) => {
 
   const hasErrorFor = getErrorMap(['label'], snapshotError);
 
+  React.useEffect(() => {
+    if (isSnapshotConfirmationDialogOpen) {
+      reset();
+    }
+  }, [isSnapshotConfirmationDialogOpen, reset]);
+
```
hkhalil-akamai, line 61: Can we change this to use `Dialog`'s `TransitionProps`? [See here for the reason why.](https://github.com/linode/manager/pull/10287#discussion_r1525455789)

Looks like `ConfirmationDialog` has an [`onExited`](https://github.com/linode/manager/blob/02d51cb86535a16a9caa4d2d655f483ad3e0d613/packages/manager/src/components/ConfirmationDialog/ConfirmationDialog.tsx#L37) prop you can use that is passed to `TransitionProps`.

packages/manager/src/features/Linodes/LinodesDetail/LinodeBackup/CaptureSnapshotConfirmationDialog.test.tsx
hkhalil-akamai, line 1: Thanks for adding this test!!!

hkhalil-akamai (approved):
Tested and looks good! Left one suggestion about using `TransitionProps` instead of `useEffect`.

> several potential ways to display the snapshot error in the confirmation dialog

I agree with the using the first way for consistency, but we could check with UX if you feel like the other pattern is more appropriate. It may be out of scope of this ticket though.

mjac0bs (commented):
Good question about the positioning of the error notice. We're inconsistent with this. 

Searching `<ConfirmationDialog` in the codebase, there are several that places we let the ConfirmationDialog format it below (as shown here) and several other places where we put the error first. (`ConfirmTransferCancelDialog`,  `ImagesLanding`, `DeletionDialog` are a few). 

It does seem like we go with Option 1 *the most* and for that reason, I would suggest sticking to that here... but it might be worth confirming with UX if we want to continue to go that route or make this consistent with one approach throughout our confirmation dialogs (in a separate ticket). Personally, I find it difficult to read, especially in dark mode, because the error message is so small and the color contrast is poor.

Another note: `This Linode has nothing to backup, please add a disk and/or and a config` is grammatically incorrect and really should be fixed on the backend. Could we let the API team know about this? (They may be fine with just a ticket since it looks like it's an error message coming from them and not some other service through APIv4.)

packages/manager/src/features/Linodes/LinodesDetail/LinodeBackup/CaptureSnapshot.tsx
coliu-akamai: adding, thanks for linking this!

coliu-akamai:
Created internal ticket [M3-8453] + will add to frontend cafe discussion too!
Will reach out to API about error message

dwiley-akamai (approved):
Code review ✅
Unit test passes ✅
Error message appears in confirmation dialog now ✅

My personal preference would be for the second of the three possible display options (error notice dialog body text), but connecting with UX to set a convention would be great

mjac0bs (approved):

PR #10811, opened 2024-08-22
test: [M3-8458] - Fix StackScripts pagination test failure

jdamore-linode:
## Description 📝
This fixes the StackScripts pagination test failure. The test fails because Cloud Manager filters certain StackScripts from the landing page, and our test didn't account for that when it checked for content in the StackScripts list.

Shout out to @AzureLatte for identifying the cause of the failure!

## Changes  🔄
- Do not assert that StackScripts exist in the list when they only support deprecated or Kube Images
- Rename `smoke-community-stackscrips.spec.ts` to `smoke-community-stackscripts.spec.ts`

## How to test 🧪
We can rely on CI since the test currently fails 100% of the time. However, you can run the test locally to manually verify the changes:

```bash
yarn cy:run -s "cypress/e2e/core/stackscripts/smoke-community-stackscripts.spec.ts"
```

*Check all that apply*

- [ ] 👀 Doing a self review


bnussman-akamai (approved):
Thanks Joe!  🔥

mjac0bs (approved):
Thanks for the fix(es)! Verified that this is consistently passing locally and the change makes sense.

image


> Rename smoke-community-stackscrips.spec.ts to smoke-community-stackscripts.spec.ts

This has bothered me for a while. 😅 🚀  🛠️ 

> We can rely on CI since the test currently fails 100% of the time. 

Interesting that we've never encountered this failure caused by a stackscript that should be filtered out before now.

jdamore-linode:
> Interesting that we've never encountered this failure caused by a stackscript that should be filtered out before now.

Yeah! It only leads to a failure when the first item on a page returned by the StackScripts endpoint is hidden by Cloud, since those are the only StackScripts the test actually checks.

If we checked every StackScript, or even just a larger random sample, we probably would have noticed it a long time ago (although using a random sample of StackScripts would've been a lot harder to troubleshoot, too!)

abailly-akamai:
getting a failure here consistently - not sure if due to my setup

Screenshot 2024-08-22 at 10 56 41

jdamore-linode:
> getting a failure here consistently - not sure if due to my setup
> 
> (Screenshot)

Hm. This isn't caused by the changes in this PR (different test), but it _has_ been flaky so I don't think it's caused by your setup either (at least not entirely).

My instinct seeing that error message is that Cloud is either selecting a region that doesn't support the Dedicated 8gb plan (don't think such a region even exists) or Cloud is failing to select a region altogether: this test selects a region by typing its name and hitting enter, which doesn't actually manifest as a failure if the desired region is absent from the drop-down, it just silently fails to select a region.

My guess is that [Cypress is randomly selecting a region to use](https://github.com/linode/manager/blob/develop/packages/manager/cypress/e2e/core/stackscripts/smoke-community-stackscrips.spec.ts#L263) which is missing some capability that is required by StackScripts, so when it attempts to select that region in the UI, it's absent from the drop-down and as a result the Dedicated 8gb plan fails to show up. What _doesn't_ make sense to me is why this would be happening consistently for you -- it should eventually pick a region that works. Is your copy of Cloud built against another environment where there might be a lot of regions lacking some capabilities?

(If you have any time today, Cypress should have saved a recording of the failure in `packages/manager/cypress/videos` -- if you send that over on Slack I'd love to take a look!)

abailly-akamai:
@jdamore-linode the error I am getting is indeed due to the region selection failing

`ui.regionSelect.find().click().type(`${region.id}{enter}`);`

Screenshot 2024-08-22 at 11 32 32

typing the ID won't return the region in the select, but the label would. 

confused why I am the only one getting this. It this a flag configuration discrepancy with distributed regions UI?

jdamore-linode:
@abailly-akamai I'm a bit confused by this too -- we've always been able to search by regions by ID as far as I'm aware so not sure why that wouldn't be working now (and seemingly only for you).

One way to rule out a Feature Flag discrepancy would be to build and serve Cloud with `yarn build && yarn start:manager:ci` so any locally configured feature flags via the dev tool won't be a factor. Either way, I've created M3-8471 to look into this since it might become a bigger problem.

abailly-akamai (approved):
Thx for the fix! failure encountered on my end unrelated to this PR

PR #10813, opened 2024-08-22
refactor: [M3-8186] - Clean up DebouncedSearchTextField and fix instances where debouncing is not happening

cpathipa:
## Description 📝
This PR fixes the issue of debouncing is not happening for the instances of DebouncedSearchTextField component. 

## Changes  🔄
- Fix debounce issue in PlacementGroupsLanding.
- Fix debounce issue in LinodeSelectTable.
- Removed customeValue prop usage in LinodeSelectTable and SelectLinodePanel.
- Memoized the debounced onChange handler to prevent unnecessary re-creations. 
- Improved the `useEffect` hook to handle empty values and reduce unnecessary state updates. 
- removed redundant logic of the clear icon from the instanced of DebouncedSearchTextField. 

## Target release date 🗓️
09/02

### Verification steps
(How to verify changes)
- Checkout the branch and verify there is no regression for all the references of `PlacementGroupsLanding`, `SelectLinodePanel` and `LinodeSelectTable`
- Verify debounce behavior for `PlacementGroupsLanding` and `LinodeSelectTable`
- Verify there is no regression in all the references of DebouncedSearchTextField.

*Check all that apply*

- [ ] 👀 Doing a self review


hkhalil-akamai (commented):
Thanks for making this fix. I originally created the `customValue` prop but it may not have been necessary.

Left some suggestions.

Failing unit tests can likely be resolved using `await waitFor(() => { ... });`

mjac0bs (commented):
Thanks for taking this one on, Chandra. Besides the thoughts Hussain raised, I'm wondering how we feel about continuing to use `onSearch` as the event handler prop  here in general. `onChange` seems more standard to me, and I'm concerned that if we proceed with `onSearch`, we'll end up with another mistake like we saw in [PlacementGroupsLanding.tsx](https://github.com/linode/manager/pull/10813/files#diff-acfcaf8a2a711a81ed82bf83de147f9ff4df6dc7e91a98ca7faad52662168f21L199) somewhere down the line. 

Besides needing to update a few more files, is there any other impact that replacing  the`onSearch` prop with `onChange` would have?

hkhalil-akamai (approved):
Thanks for addressing feedback! This looks good, approved pending a decision regarding @mjac0bs' comment about `onSearch` vs `onChange`.

cpathipa:
@hkhalil-akamai @mjac0bs Great feedback.
@mjac0bs Good call, Actually, the root cause of the issue is having `onSearch` as a dependency in the `useEffect`, which is causing unwanted re-triggers. Removing it from the dependencies would prevent this from happening in the future. Also, I don't see the use case for having it in the dependency list. The only concern I see with renaming `onSearch` to `onChange` is that we might override the `onChange` prop of `TextFieldPropsOverrides`. In my opinion, onSearch is a wrapper for onChange in `DebouncedSearchTextField`. Let me know your thoughts—I'm open to changing the prop name to `onChange` as well.

packages/manager/src/components/DebouncedSearchTextField/DebouncedSearchTextField.tsx
```diff
@@ -5,25 +5,18 @@ import * as React from 'react';
...
 import { CircleProgress } from 'src/components/CircleProgress';
 import { InputAdornment } from 'src/components/InputAdornment';
-import { TextField, TextFieldProps } from 'src/components/TextField';
+import { TextField } from 'src/components/TextField';
 
 import { IconButton } from '../IconButton';
 
+import type { TextFieldProps } from 'src/components/TextField';
+
 export interface DebouncedSearchProps extends TextFieldProps {
```
mjac0bs, line 14: Okay, I see, `onSearch` became an optional prop when `customValue` was added, and we don't want that - `onSearch` should be required. This will make it clear what the correct handler to use is. Can we make that change in this interface?

packages/manager/src/components/DebouncedSearchTextField/DebouncedSearchTextField.tsx
```diff
@@ -75,7 +65,7 @@ export const DebouncedSearchTextField = React.memo(
         return () => clearTimeout(timeout);
       }
       return undefined;
-    }, [debounceTime, onSearch, textFieldValue]);
+    }, [debounceTime, textFieldValue]);
```
mjac0bs, line 68: This is old code, but I don't understand why we went with a useEffect and setTimeout when we generally opt to use the `debounce` util throughout the codebase. And with `onSearch` required, we won't need to check for it. Could we clean this up a little and achieve the same by doing? 

```
<Textfield
    [...other props]
    onChange={debounce(debounceTime ?? 400, (e) => {
      onSearch(e.target.value);
      setTextFieldValue(e.target.value);
    })}
/>
```

cpathipa:
@mjac0bs Addressed the above feedback in the commit 9bb33ae
Changes include 

- Memoized the debounced onChange handler to prevent unnecessary re-creations.
- Improved the useEffect hook to handle empty values and reduce unnecessary state updates.
- removed redundant logic of the clear icon from the instanced of DebouncedSearchTextField.

The `onSearch` and `value` props are now required to ensure that the component can function correctly. The `onSearch` prop is essential for handling search. The `value` prop is necessary for managing the controlled input state. Making these props required enforces proper usage and helps prevent unintended behavior if they are accidentally omitted.

mjac0bs (approved):
The only thing I found on this last pass when checking through the places that the component is used throughout CM was a console error in Longview, which seems like it's happening because `inputText` can be undefined [here](https://github.com/linode/manager/blob/develop/packages/manager/src/features/Longview/LongviewDetail/DetailTabs/Processes/ProcessesLanding.tsx#L52). 

Screenshot 2024-09-03 at 9 15 40 AM

I just restarted the e2e tests because they didn't kick off correctly - but approving pending the above is addressed and those look good.

cpathipa:
@mjac0bs fixed the console error in Longview a1b2e8f

jdamore-linode (approved):

PR #10820, opened 2024-08-23
change: [M3-6546] - Volume drawers and action menu refactored

dchyrva-akamai:
## Description 📝
Created new drawer for managing volume tags, action menu for the volume row updated.

## Changes  🔄
- Removed "Show Config" button from the volume table row and added to the actions menu.
- Remove "Edit" button from the volume table row and added to the actions menu .
- Added "Manage Tags" option to the volume table actions menu.

- Added new "Manage Tags" drawer.
- Existing volume tags functionality moved from "Edit Volume" drawer to "Manage Tags" drawer.


packages/manager/cypress/e2e/core/volumes/update-volume.spec.ts
```diff
@@ -64,13 +61,58 @@ describe('volume update flow', () => {
...
+        });
+
+        // Confirm that volume is listed on landing page, click "Edit" to open drawer.
+        cy.findByText(volume.label)
+          .should('be.visible')
+          .closest('tr')
+          .within(() => {
+            cy.get('button').click();
+          });
+        cy.get('[data-testid="Manage Tags"]').click();
```
jdamore-linode, line 99:
```suggestion
        // Confirm that volume is listed on landing page, click "Edit" to open drawer.
        cy.findByText(volume.label)
          .should('be.visible')
          .closest('tr')
          .within(() => {
            ui.actionMenu
              .findByTitle(`Action menu for Volume ${volume.label}`)
              .should('be.visible')
              .click();
          });

        ui.actionMenuItem
          .findByTitle('Manage Tags')
          .click();
```

We've got a bunch of helpers that make selecting common UI elements like drawers, modals, buttons, action menus, etc., a lot easier. You'll have to import the `ui` object via `import { ui } from 'support/ui'`, but using those should make it a lot easier to write tests without having to go searching through the codebase for test tags (and helps keep hardcoded IDs out of the tests)

jdamore-linode (commented):
Thanks @dchyrva! Noticed a couple small issues with the E2E test changes but nothing too serious -- happy to help out if you run into any trouble addressing feedback!

packages/validation/src/volumes.schema.ts
```diff
@@ -46,6 +46,7 @@ export const ResizeVolumeSchema = (minSize: number = 10) =>
 
 export const UpdateVolumeSchema = object({
   label: string().required(),
+  tags: array().of(string()).required(),
```
bnussman-akamai, line 49: The validation package is meant to be representative of what api-v4 allows. I'm wondering if we should remove `required` on the `label` and `tags` here.

dchyrva: Hello @bnussman-akamai,

Thank you for your comment.

I removed "required" modifiers.

packages/manager/cypress/e2e/core/volumes/update-volume.spec.ts
dchyrva: Hello @jdamore-linode,

Thank you for your comments, everything should be resolved now.

packages/manager/src/features/Volumes/Drawers/ManageTagsDrawer.tsx
hkhalil-akamai, line 1: Formik is being deprecated in favor of using `react-hook-form`, which provides improved type-safety. See [RebuildImageDrawer](https://github.com/linode/manager/blob/ff631efec43258ca5eee5f9341695285b17ca996/packages/manager/src/features/Images/ImagesLanding/RebuildImageDrawer.tsx) for an example.

hkhalil-akamai (commented):
Thanks for the contribution! I had a couple of concerns:

1. We have an existing drawer component called [`TagDrawer`](https://github.com/linode/manager/blob/develop/packages/manager/src/components/TagCell/TagDrawer.tsx). Although the UI is slightly different, would it be possible to use that instead of re-implementing a similar component?
2. The Figma prototype in the attached ticket includes the ability to group volumes by tags. Should this be implemented in this or another ticket (as it may require significant changes)?

jaalah-akamai:
> We have an existing drawer component called [TagDrawer](https://github.com/linode/manager/blob/develop/packages/manager/src/components/TagCell/TagDrawer.tsx). Although the UI is slightly different, would it be possible to use that instead of re-implementing a similar component?
@hkhalil-akamai The figma mock is using the version that's currently in place, which is essentially an autocomplete. Let's keep this in place for now until we can ask UX Thursday. Good question 👍 

> The Figma prototype in the attached ticket includes the ability to group volumes by tags. Should this be implemented in this or another ticket (as it may require significant changes)?
@dchyrva lets add this as part of a new PR - part 2.

jaalah-akamai:
Let's move forward with the the autocomplete tag implementation. We're eventually going to sunset the TagDrawer where we display each tag visually on the page.

jaalah-akamai (approved):
Nice job @dchyrva-akamai

bnussman-akamai (commented):
Getting an unexpected error when clearing tags

https://github.com/user-attachments/assets/ad695713-b548-4bc6-89b1-5ae26953bbc0

dchyrva-akamai:
Hello, @bnussman-akamai

Thank you for catching the error.

Should be fixed now.

packages/manager/src/features/Volumes/Drawers/ManageTagsDrawer.tsx
```diff
@@ -0,0 +1,116 @@
...
+    } catch (errors) {
+      errors.forEach((error: APIError) => {
+        if (error.field == 'tags') {
+          setError('tags', {
+            message: error.reason,
+          });
+        } else {
+          setGeneralError(
+            'Unable to edit this Volume at this time. Please try again later.'
+          );
```
bnussman-akamai, line 61: Rather than having a "generalError" state, we could use `react-hook-form`'s `root` error state. This is how we handle general errors on the Linode Create flow. 

We can also just use `error.reason` instead of that hardcoded error

```suggestion
          form.setError('root', { message: error.reason });
```

packages/manager/cypress/e2e/core/volumes/update-volume.spec.ts
```diff
@@ -82,4 +155,8 @@ describe('volume update flow', () => {
       }
     );
   });
+
+  after(() => {
+    cleanUp(['tags', 'volumes']);
+  });
```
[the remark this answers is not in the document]
jdamore-linode: It doesn't hurt anything, but I don't think it should be necessary since we're already doing clean up in the `before` hook ([explanation why](https://docs.cypress.io/guides/references/best-practices#Using-after-Or-afterEach-Hooks)). @dchyrva-akamai any reason in particular this was added?

packages/manager/cypress/e2e/core/volumes/update-volume.spec.ts
```diff
@@ -40,17 +44,86 @@ describe('volume update flow', () => {
...
+      label: randomLabel(),
+      region: chooseRegion().id,
+    });
+
+    const newTags = [randomLabel(5), randomLabel(5), randomLabel(5)];
+
+    cy.defer(() => createVolume(volumeRequest), 'creating volume').then(
+      (volume: Volume) => {
+        // Volume needs some time to become active after creation.
+        cy.wait(1000);
```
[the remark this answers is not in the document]
jdamore-linode: @bnussman-akamai Yep, we should be avoiding `cy.wait(<number of milliseconds>)` if we can ([explanation](https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting)).

I think your suggestion to wait for the UI is what we do in other tests, and it'd look something like this:

```typescript
cy.defer(() => createVolume(volumeRequest), 'creating volume').then(
      (volume: Volume) => {
        // Volume needs some time to become active after creation.
        // Find the Volume's label in the table, then narrow our selection to that table row.
        cy.findByText(volume.label)
          .should('be.visible')
          .closest('tr')
          .within(() => {
            // Wait for "Creating" status to become "Active".
            // I _think_ the actual DOM text is 'active' and not 'Active' and we use CSS for capitalization here, but
            // you might want to fact check me here.
            cy.findByText('active').should('be.visible');
          });
```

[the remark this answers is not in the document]
bnussman-akamai: Yeah, I think there is some API weirdness causing the status to not always update in the UI. I'll create a separate ticket to investigate that.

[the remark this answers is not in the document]
bnussman-akamai: Created M3-8557 for that bug you mentioned @jdamore-linode 

@dchyrva-akamai That example Joe shared using `createActiveVolume` might be a good option here!

packages/manager/src/features/Linodes/LinodeCreatev2/Security.test.tsx
```diff
@@ -17,20 +17,16 @@ import { Security } from './Security';
...
-      });
-
-      const rootPasswordInput = await findByLabelText('Root Password');
-
-      expect(rootPasswordInput).toBeVisible();
-      expect(rootPasswordInput).toBeEnabled();
-    },
-    { timeout: 5_000 }
-  );
+  it('should render a root password input', { timeout: 5_000 }, async () => {
```
mjac0bs, line 20: We [skipped](https://github.com/linode/manager/pull/10859#discussion_r1752386881) this test in M3-8158 (#10859) because it was causing issues in that PR, as well. We can revert this PR's change to the file when addressing the merge conflict, stick with the skip, and address the issues with the test outlined in M3-8559, because there is more happening here than the deprecated order of the arguments. The timeout needs to be adjusted, as it is liable to cause another CI job failure because CI runs slow.

packages/manager/cypress/e2e/core/volumes/update-volume.spec.ts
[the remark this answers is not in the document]
dchyrva-akamai: @jdamore-linode I added clean up after in order to remove volumes, created for testing purposes, from the account.
Maybe you can recommend a better way of doing that?

packages/manager/src/features/Linodes/LinodeCreatev2/Security.test.tsx
dchyrva-akamai: @mjac0bs  Should be resolved now, thank you!

packages/manager/src/features/Volumes/Drawers/ManageTagsDrawer.tsx
dchyrva-akamai: @bnussman-akamai Hello, thank you for the hint, "root" has been used instead of the generalError.

packages/manager/cypress/e2e/core/volumes/update-volume.spec.ts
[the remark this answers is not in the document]
dchyrva-akamai: @bnussman-akamai @jdamore-linode Thank you for the comment, I used `createActiveVolume` instead.

packages/manager/cypress/e2e/core/volumes/update-volume.spec.ts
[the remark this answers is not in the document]
hkhalil-akamai: @jdamore-linode can we double check if this is necessary?

hkhalil-akamai (approved):
Is there anything else blocking this PR, other than [the one open question](https://github.com/linode/manager/pull/10820#discussion_r1756681242)?

packages/manager/cypress/e2e/core/volumes/update-volume.spec.ts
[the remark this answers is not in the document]
jdamore-linode: @hkhalil-akamai No, it shouldn't be. @dchyrva-akamai might be able to speak more to his motivation for adding this (e.g. were you getting test failures that this solved?) but broadly speaking we should not be doing any clean up in `after` hooks (and the link I posted above explains why).

(The obvious elephant in the room is #11028 where I'm doing exactly that; that's an exceptional case that we agreed to for security compliance reasons, unlike this situation where the clean up is being done for the sake of setting up a test's state)

dchyrva-akamai:
@jdamore-linode @hkhalil-akamai Hello and thank you for you patience, I had to switch to the Netstorage for a while.
I have reopened this PR and synchronized the branch with the latest develop branch.

As far as I understand, the last blocking factor here was the `cleanUp` call inside of the `after` method right?

As I had mentioned before, I added it to remove volumes from my account created by the e2e test.

Steps to reproduce the issue are quite simple, you can try them on your own machine:
- Do `git switch` selecting this branch.
- Run `yarn cy:debug` from the terminal.
- Choose "Chrome".
- Find `update-volume.spec.ts` using Cypress UI.
- Run the test suit.
- After the test run, kill the cypress process from the terminal.
- Observe your "Volumes page": 
image

I personally would prefer not to delete them manually all the time, this is why added `cleanUp` at the end of the test.

PR #10824, opened 2024-08-23
fix: [M3-8477] - Event handlers making a proportional number of GET requests to the number of incoming events

bnussman-akamai:
## Description 📝

- Fixes event handler bug discovered by @jdamore-linode  🔧 

## The Issue 🐛 

- A [React Query change](https://tanstack.com/query/latest/docs/framework/react/guides/migrating-to-react-query-4#consistent-behavior-for-cancelrefetch) in v4 makes `invalidateQueries` fire off many GET request if `invalidateQueries` is called many times sequentially

## The Fix 🔧 

- We want to intentionally disable this new behavior (using `cancelRefetch: false`) in our event handlers because our event handlers were built based on the assumption React Query would only fire off one request when `invalidateQueries` was called many times sequentially

## Preview 📷

| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/user-attachments/assets/429b6310-0f74-434b-a3d0-d273ea706106" /> | <video src="https://github.com/user-attachments/assets/102a65b8-508e-47c7-94c6-9b8ee67cb101" /> |
| Notice how the event handers cause 3 repetitive GETs when event handlers are invoked  | Notice how only 1 GET happens when event handlers are invoked  |

### Reproduction steps
- Open the network dev tools
- Within the 16 second polling interval, do something that triggers a new event n times. For example, enable and disable a domain n times.
- When the polling interval fires off verify you see n requests to GET /domains

### Verification steps
- Open the network dev tools
- Within the 16 second polling interval, do something that triggers a new event n times. For example, enable and disable a domain n times.
- When the polling interval fires off verify you don't see any duplicate requests to GET /domains

- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/.changeset/pr-10824-fixed-1724428840947.md
+++ b/packages/manager/.changeset/pr-10824-fixed-1724428840947.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Fixed
+---
+
+Event handlers making a proportional number of GET requests to the number of incoming events ([#10824](https://github.com/linode/manager/pull/10824))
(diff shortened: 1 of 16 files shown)


jaalah-akamai (approved):
Nice find @jdamore-linode 🥇

PR #10828, opened 2024-08-23
fix: [M3-8453] - Update error text color in confirmation dialogs to use theme, refactor styling

coliu-akamai:
## Description 📝
Update error text color in confirmation dialogs to not be hardcoded, as per discussion. Also added some tests/refactored styling to use styled components while I was at it

## Changes  🔄
- updated error text color to use the theme
- added tests for Confirmation Dialogs
- refactor styling

## Target release date 🗓️
n/a

## Preview
| Before  | After   |
| ------- | ------- |
| imageimage| imageimage|

### Verification steps
- yarn test components/ConfirmationDialog.test.tsx
- Verify that Confirmation Dialog styling remains the same throughout the app / verify it remains the same on storybook
- Except the error text color, verify the error text color has been changed + is easier to see in dark mode

*Check all that apply*

- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/src/components/ActionsPanel/ActionsPanel.tsx
+++ b/packages/manager/src/components/ActionsPanel/ActionsPanel.tsx
@@ -2,9 +2,12 @@ import { styled } from '@mui/material/styles';
 import * as React from 'react';
 import { useStyles } from 'tss-react/mui';
 
-import { Button, ButtonProps } from 'src/components/Button/Button';
+import { Button } from 'src/components/Button/Button';
 
-import { Box, BoxProps } from '../Box';
+import { Box } from '../Box';
+
+import type { BoxProps } from '../Box';
+import type { ButtonProps } from 'src/components/Button/Button';
 
 interface ActionButtonsProps extends ButtonProps {
   'data-node-idx'?: number;
--- a/packages/manager/src/components/ConfirmationDialog/ConfirmationDialog.tsx
+++ b/packages/manager/src/components/ConfirmationDialog/ConfirmationDialog.tsx
@@ -1,34 +1,13 @@
-import Dialog, { DialogProps } from '@mui/material/Dialog';
+import Dialog from '@mui/material/Dialog';
 import DialogActions from '@mui/material/DialogActions';
 import DialogContent from '@mui/material/DialogContent';
 import DialogContentText from '@mui/material/DialogContentText';
-import { Theme } from '@mui/material/styles';
+import { styled } from '@mui/material/styles';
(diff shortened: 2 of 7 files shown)


packages/manager/src/components/ActionsPanel/ActionsPanel.tsx
```diff
@@ -2,9 +2,12 @@ import { styled } from '@mui/material/styles';
 import * as React from 'react';
 import { useStyles } from 'tss-react/mui';
 
-import { Button, ButtonProps } from 'src/components/Button/Button';
```
coliu-akamai, line 5: this and Dialog.tsx are just eslint changes I saw when I landed on those files while working on this ticket

coliu-akamai:
Is 'Changed' the right category for this changeset, or would it be 'Fixed'? 
Should I add another changeset for the unit tests and/or style refactoring too?

packages/manager/src/components/ConfirmationDialog/ConfirmationDialog.tsx
```diff
@@ -73,26 +50,52 @@ export const ConfirmationDialog = (props: ConfirmationDialogProps) => {
...
+  label: 'StyledDialogContent',
+})({
+  display: 'flex',
+  flexDirection: 'column',
+});
+
+const StyledErrorText = styled(DialogContentText, {
+  label: 'StyledErrorText',
+})(({ theme }) => ({
+  color: theme.palette.error.dark,
```
coliu-akamai, line 99: ...the actual assigned part of this ticket lol

thanks @jaalah-akamai for letting me know which theme color to use!

mjac0bs (approved):
🚀  Thanks, Connie! Looking much better in dark mode now and you did some good clean up, too.

> Is 'Changed' the right category for this changeset, or would it be 'Fixed'?

Since it was an incorrect color and (more importantly) that low contrast was an accessibility issue, I would lean towards `Fixed` and rewording the changeset to be something like: `Inaccessible, non-theme error text color in confirmation dialogs`. What you have does work, though.

> Should I add another changeset for the unit tests and/or style refactoring too?

I wouldn't add a changeset for style refactoring. Adding a Test changeset for the new unit test is up to you - and we could probably be more consistent about when to do this as a team. I think it's a good idea because it's a new file to test a component we didn't have coverage for previously, as opposed to a new test case for an existing file, which I usually don't bother adding a separate changeset for. Thanks for the test coverage!

AzureLatte (approved):

PR #10838, opened 2024-08-26
fix: [M3-8492] - Create Linode Firewall warning not appearing

hkhalil-akamai:
## Description 📝
Fixes a bug causing the Firewall warning to not appear after the user assigns and then clears the firewall select during the create flow.

## Changes  🔄
- Update the `firewall_id` interface in `CreateLinodePayload` to not accept null values (i.e., `number | undefined`)
- Refactor `SelectFirewallPanel` and associated types in Create v1 to use `undefined` instead of `-1` when the firewall field is cleared
- Remove a redundant `useEffect`

## Target release date 🗓️
9/3

## Preview 📷
| Before  | After   |
| ------- | ------- |
| **Create v2** |
|  <video src="https://github.com/user-attachments/assets/747aaece-561c-447c-a43d-1b691e964c9c" /> | <video src="https://github.com/user-attachments/assets/f52ccd5a-dd9a-4b25-abf9-64f121b0b160" /> |
| **Create v1** |
| <video src="https://github.com/user-attachments/assets/5e3a402a-3edb-4fd8-8a74-4c871cf720f1" />| <video src="https://github.com/user-attachments/assets/34de2782-4111-4363-b1a5-7648ff9a0265" /> |

### Prerequisites
Enable Secure VM notices by adding `secure_vm_notices: 'always'` in the [preference editor](http://localhost:3000/profile/settings?preferenceEditor=true).

### Reproduction steps
- Select and then clear the firewall field in the Linode Create flows (v1 and v2)
- Click the "Create Linode" button
- Observe that the Firewall authorization warning banner does not appear
    - In Create v1, the form is submitted and the Linode is created
    - In Create v2, the Create Linode button becomes disabled

### Verification steps
- Follow the reproduction steps
- Verify that the Firewall authorization banner always appears after clicking the Create Linode button when no firewall is assigned

### DIFF
--- a/packages/api-v4/src/linodes/types.ts
+++ b/packages/api-v4/src/linodes/types.ts
@@ -452,7 +452,7 @@ export interface CreateLinodeRequest {
   /**
    * The `id` of the Firewall to attach this Linode to upon creation.
    */
-  firewall_id?: number;
+  firewall_id?: number | null;
   /**
    * An object that assigns this the Linode to a placement group upon creation.
    */
--- a/packages/api-v4/src/linodes/types.ts
+++ b/packages/api-v4/src/linodes/types.ts
@@ -452,7 +452,7 @@ export interface CreateLinodeRequest {
   /**
    * The `id` of the Firewall to attach this Linode to upon creation.
    */
-  firewall_id?: number | null;
+  firewall_id?: number;
   /**
    * An object that assigns this the Linode to a placement group upon creation.
    */
--- a/packages/manager/src/components/SelectFirewallPanel/SelectFirewallPanel.tsx
+++ b/packages/manager/src/components/SelectFirewallPanel/SelectFirewallPanel.tsx
@@ -1,6 +1,5 @@
 import { styled } from '@mui/material/styles';
 import * as React from 'react';
-import { useLocation } from 'react-router-dom';
 
 import { Box } from 'src/components/Box';
 import { Paper } from 'src/components/Paper';
@@ -10,8 +9,6 @@ import { CreateFirewallDrawer } from 'src/features/Firewalls/FirewallLanding/Cre
 import { useFlags } from 'src/hooks/useFlags';
 import { useSecureVMNoticesEnabled } from 'src/hooks/useSecureVMNoticesEnabled';
 import { useFirewallsQuery } from 'src/queries/firewalls';
-import { sendLinodeCreateFormInputEvent } from 'src/utilities/analytics/formEventAnalytics';
-import { getQueryParamsFromQueryString } from 'src/utilities/queryParams';
 
 import { AkamaiBanner } from '../AkamaiBanner/AkamaiBanner';
 import { Autocomplete } from '../Autocomplete/Autocomplete';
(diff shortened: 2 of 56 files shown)


packages/manager/src/components/SelectFirewallPanel/SelectFirewallPanel.tsx
hkhalil-akamai, line 1: Use `undefined` instead of `-1` when clearing the firewall field.

packages/manager/src/features/Linodes/LinodeCreatev2/Actions.tsx
```diff
@@ -45,7 +45,11 @@ export const Actions = () => {
 
   return (
     <Box sx={{ display: 'flex', gap: 1, justifyContent: 'flex-end' }}>
-      <Button buttonType="outlined" onClick={onOpenAPIAwareness}>
+      <Button
+        buttonType="outlined"
+        disabled={disableSubmitButton}
+        onClick={onOpenAPIAwareness}
+      >
```
hkhalil-akamai, line 52: This matches the behavior from create v1.

packages/manager/src/features/Linodes/LinodeCreatev2/FirewallAuthorization.tsx
```diff
@@ -1,33 +1,23 @@
...
     control,
     name: 'firewallOverride',
   });
 
   const watchFirewall = watch('firewall_id');
-  useEffect(() => {
-    if (isNotNullOrUndefined(watchFirewall)) {
-      clearErrors('firewallOverride');
-    }
-  }, [clearErrors, watchFirewall]);
```
hkhalil-akamai, line 30: I meant to delete this `useEffect` in [an earlier PR](https://github.com/linode/manager/pull/10751#discussion_r1705514532).

packages/manager/src/features/Linodes/LinodesCreate/LinodeCreate.tsx
```diff
@@ -119,10 +119,10 @@ export interface LinodeCreateProps {
   assignPublicIPv4Address: boolean;
   autoassignIPv4WithinVPC: boolean;
   checkValidation: LinodeCreateValidation;
-  checkedFirewallAuthorizaton: boolean;
+  checkedFirewallAuthorization: boolean;
```
hkhalil-akamai, line 122: Typo:
```
checkedFirewallAuthorization
                         ^
```

packages/manager/src/features/Linodes/LinodeCreatev2/Actions.tsx
bnussman-akamai: Should we be excluding this from the API/CLI payloads?

Screenshot 2024-08-26 at 5 00 34 PM

packages/api-v4/src/linodes/types.ts
```diff
@@ -452,7 +452,7 @@ export interface CreateLinodeRequest {
   /**
    * The `id` of the Firewall to attach this Linode to upon creation.
    */
-  firewall_id?: number | null;
+  firewall_id?: number;
```
bnussman-akamai, line 455: I think the API actually accepts `null` so it might be work keeping?

bnussman-akamai (commented):
This is probably unrelated but I'm seeing a `validateDOMNesting` warning

Screenshot 2024-08-26 at 4 58 20 PM

packages/api-v4/src/linodes/types.ts
hkhalil-akamai: Fair enough. Reverted and updated the validation code to account for null `firewall_id`

bnussman-akamai (approved):
Not seeing any regressions ✅

abailly-akamai (approved):
Fix confirmed! well done

unrelated to this PR and non-blocking, we should handle the banner on mobile a bit better

Screenshot 2024-08-28 at 16 45 14

Screenshot 2024-08-28 at 16 44 54

PR #10846, opened 2024-08-28
fix: [M3-8500] - Restricted users without account access unable to create Linodes on Linode Create v2

bnussman-akamai:
## Description 📝

- Updates Linode Create v2 to be in parity with Linode Create v1 when it comes to a restricted user creating a Linode in an EU region

## The Bug 🐛 

- On Linode Create v2, if a user with `add_linodes` access, but with no account/billing permissions tried to create a LInode, they would see an infinite loading spinner. This is because Cloud Manager was trying to GET `/v4/account/agreemeents`, but failed to because the endpoint will 403 for this restricted user.

## The Fix 🔧 

- Allow restricted users with `add_linodes` access but with no account/billing permissions to create Linodes regardless of the account's agreements. 
- This seems questionable, but this is what Linode Create v1 intentionally did. See https://github.com/linode/manager/pull/7901 

## Target release date 🗓️
9/3/24

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-08-28 at 11 14 54 AM | <video src="https://github.com/user-attachments/assets/58b8c653-3cac-491f-9ae1-783f55c29df3" /> |
| User would see an infinite loading spinner | User can create a Linode | 

### Prerequisites
- Create a restricted user on your account with the following permissions
  - `add_linodes: true` (`Can add Linodes to this account` in the UI)
  - `account_access: null` (No billing access in the UI)

### Reproduction steps
- Login as the restricted user
- Try to create a LInode in an EU region
- Observe an infinite loading spinner when you click "Create Linode" button

### Verification steps
- Repeat the "Reproduction steps" with this PR checked out
- Rather than seeing an infinite loading spinner, you should be able to successfully create a Linode in an EU region

- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/src/features/Linodes/LinodeCreatev2/resolvers.ts
+++ b/packages/manager/src/features/Linodes/LinodeCreatev2/resolvers.ts
@@ -13,14 +13,17 @@ import {
 import { getLinodeCreatePayload } from './utilities';
 
 import type { LinodeCreateType } from '../LinodesCreate/types';
-import type { LinodeCreateFormValues } from './utilities';
+import type {
+  LinodeCreateFormContext,
+  LinodeCreateFormValues,
+} from './utilities';
 import type { QueryClient } from '@tanstack/react-query';
 import type { FieldErrors, Resolver } from 'react-hook-form';
 
 export const getLinodeCreateResolver = (
   tab: LinodeCreateType | undefined,
   queryClient: QueryClient
-): Resolver<LinodeCreateFormValues, { secureVMNoticesEnabled: boolean }> => {
+): Resolver<LinodeCreateFormValues, LinodeCreateFormContext> => {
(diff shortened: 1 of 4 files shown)


packages/manager/src/features/Linodes/LinodeCreatev2/resolvers.ts
```diff
@@ -46,7 +49,7 @@ export const getLinodeCreateResolver = (
       getRegionCountryGroup(selectedRegion)
     );
 
-    if (hasSelectedAnEURegion) {
+    if (hasSelectedAnEURegion && !context?.profile?.restricted) {
```
bnussman-akamai, line 52: This is the fix 🔨 

We only check account agreements if the user is **not restricted**

abailly-akamai (approved):
Thx for the fix!

PR #10850, opened 2024-08-28
refactor: [M3-6852] - Remove global error interceptors

hkhalil-akamai:
## Description 📝
Today the Cloud Manager uses a global error interceptor to add links to `APIError`s that call for users to open a support ticket. This interceptor introduces type unsafety (since it necessitates [a cast](https://github.com/linode/manager/blob/2c891f971f51e18f007ec220db003f33dae84639/packages/manager/src/utilities/interceptAPIError.tsx#L18)) and reduces flexibility in displaying error messages.

#10684 made progress towards reducing our reliance on this interceptor by using the `<ErrorMessage />` component, which formats errors at the component level. This PR completes that refactor by removing the two remaining cases where errors were being intercepted:

* Errors about the user reaching their thing limit
    * These are now handled at the component level for all entities (Linodes, firewalls, volumes, etc.)
* Errors caused when trying to migrate in an account with the `dc_migrations_disabled` customer tag
    * This is now handled in the `MigrateLinode` component

## Changes  🔄
- Deletes the global error interceptor in `request.tsx` and associated utilities (`interceptAPIError.ts`)
- Detect migration errors in the `ErrorMessage` component
- Add `ErrorMessage` in a couple of missing places (`GenerateFirewallDialog`, `MigrateLinode`)

## Target release date 🗓️
9/17

## How to test 🧪

*In all of the following instances, displayed errors should include a link to the `SupportTicketDialog`. For example:*

image

* Add the `dc_migrations_disabled` tag to your account and attempt to migrate a Linode
* Reduce your thing limit/reputation in admin and attempt to create new services (e.g., Linodes, volumes, firewalls, etc.)
* Verify `APIError`s otherwise appear as expected across the app


packages/manager/src/features/Linodes/MigrateLinode/MigrateLinode.tsx
```diff
@@ -234,7 +235,11 @@ export const MigrateLinode = React.memo((props: Props) => {
       open={open}
       title={`Migrate Linode ${linode.label ?? ''} to another region`}
     >
-      {error && <Notice text={error?.[0].reason} variant="error" />}
+      {error && (
+        <Notice variant="error">
+          <ErrorMessage message={error[0].reason} />
+        </Notice>
+      )}
```
bnussman-akamai, line 242: Might be out of scope for this PR: it might be nice from a support and UX perspective to pass down the Linode ID so that it auto-fills the Support Ticket form with a bit more context

bnussman-akamai (approved):
Nice clean up! 🧼

mjac0bs (approved):
Thanks for doing this, @hkhalil-akamai! The migrate and support errors are looking good and opening the support ticket dialog at the component level.

There's one minor bit of clean up from [a change I made in a previous PR](https://github.com/linode/manager/pull/10684/files#diff-29f89836c2b7b31e3bfa3e89577d580cd46bbb943e32ea6c4213420cac491ff5R52-R55) due to the casting -  we _could_ revert it now with better type safety.

PR #10858, opened 2024-08-29
change: [M3-8377, M3-8579] - Move Region section above Images in Linode Create and update default OS to Ubuntu 24.04 LTS

hana-linode:
## Description 📝
1. Move Region section above the Images section in the Linode Create flows. This change has been approved by UX/Marketing/Docs
2. Set Default OS to Ubuntu 24.04 LTS

No changes have been made in the Backups/Clone tab since they don't have an Images section

## Preview 📷
| Before  | After   |
| ------- | ------- | 
| image | image |

### Verification steps
(How to verify changes)
- Click through the Linode Create flow tabs and ensure that the Region section is above the Images section and that the default OS is Ubuntu 24.04 LTS
- There should be no regressions in creating a Linode

*Check all that apply*

- [ ] 👀 Doing a self review


packages/manager/src/features/Linodes/LinodeCreatev2/Tabs/Images.tsx
```diff
@@ -95,27 +96,30 @@ export const Images = () => {
...
-            <DistributedRegionIcon height="21px" width="24px" />
-            <Typography>
-              Indicates compatibility with distributed compute regions.
-            </Typography>
-          </Stack>
-        )}
-      </Box>
-    </Paper>
+    <Stack spacing={3}>
+      <Region />
```
bnussman-akamai, line 100: Now that `Region` is rendered in these components, it is more susceptible to re-renders if any state in this component changes.

Could we somehow place `Region` outside of these subcomponents. I think it would help ensure we keep Linode Create v2 performant. Also, rendering a single instance of `Region` at a higher level will keep Linode Create v2 easy to understand from a code perspective. 

This might be tricky with how the Tabs work but maybe this is still possible?

hana-linode: @bnussman-akamai Hmm what about memoizing the `Region` component?

bnussman-akamai: That would help, but I still think it would be great if we could keep one instance of `<Region />` in the JSX so the JSX conceptually matches up with the UI

hana-linode: @bnussman-akamai Pushed the memo change. I tried keeping one instance of `Region` / having it at a higher level but since it's not always in the same place, that makes it difficult. This was the cleanest solution I could come up

bnussman-akamai: Ohhhh. I didn't realize this only applied to the Image tabs. My fault.

bnussman-akamai:
Looks like the `create-linode-mobile.spec.ts` failure might be legit

coliu-akamai:
test failures for 
- `create-linode-view-code-snippet.spec.ts` (creates a linode via CLI - searched for Debian but was Ubuntu. sometimes for this test it also gets hung up on clicking the region? not too sure)
- `linode-storage.spec.ts` ('try to delete in use disk' - searched for Debian 11 but was Ubuntu)

might also be related - ran them locally and the above failed

- `resize-linode.spec.ts` also failed locally but not sure if that's related to these changes

coliu-akamai (approved):
ty Hana! 🎉 
✅ confirmed region moved to top for Linode Create flow
✅ confirmed default is now Ubuntu 24.04
✅ double checked cypress tests, failing test passed locally so doesn't seem related to this pr

am curious about the ongoing thread, if there was another potential solution? 👀

hana-linode:
Yeah, not sure if there's another solution that doesn't require a good amount of refactoring

bnussman-akamai (approved):

PR #10859, opened 2024-08-29
feat: [M3-8158] - Begin to sunset Gravatar

mjac0bs:
## Description 📝
This is the first step in the two step process of sunsetting Gravatar. It adds new components and replaces the default icons for users *currently without Gravatar* with new avatars. For users *with Gravatars*, nothing changes yet except for an info dismissible banner informing Gravatar support will be sunset two weeks from the time this is released.

## Changes  🔄
- Creates a few new components:
   - `ColorPicker` component based on the native HTML ColorPicker
   - `Avatar` component to render a custom background color (stored in local storage) and first initial of username
      - Proxy avatars remain the same as in Parent/Child, but system events by Akamai/Linode have a default color and wave logo
   -  `AvatarColorPickerDialog` to change the background color from its default
- Updates the Profile Display page to allow users without Gravatars to use the `AvatarColorPickerDialog`
- Conditionally renders the new `Avatar` component for users without Gravatars in the following places:
   - Top Menu
   - Notification Center Event
   - Events Landing > Event Table Row
   - User Landing > User Table Row
   - User SSH Key Panel
   - Profile Display
   - Support Ticket Details
- Displays a global gravatar sunsetting banner for users with Gravatars only
- Adds unit tests for new `Avatar` and `AvatarColorPickerDialog` components; updates `EventRow` unit test for changes
- Adds Storybook stories for `Avatar` and `ColorPicker` components

## Target release date 🗓️
09/16/24

## Preview 📷
| Gravatar  | Non-Gravatar  |
| ------- | ------- |
|Screenshot 2024-09-05 at 9 04 19 AM|ProfileDisplayv2|

<details>
<summary>
More Photos 📸 
</summary>

| Before  | After   |
| ------- | ------- |
|ProfileOld |Profilev2|
|NotificationsOld| NotificationsNew|
|EventsOld| EventsNew| 
|SSHKeysOld|SSHKeysNew|
|supportOld | supportNew|
| When logged in as another user on an account, the *other* users will display as the default color and the active user will see their own avatar color from local storage | Screenshot 2024-09-11 at 7 28 02 AM Screenshot 2024-09-11 at 7 29 07 AM Screenshot 2024-09-11 at 7 29 50 AM |

</details>

### Prerequisites
- Have an account with a Gravatar and one without (use a test account if you don't have one without) 

### Verification steps
(How to verify changes)
- For an account with a Gravatar:
   - Check the following places and confirm the user's Gravatar is still displayed:
      - Top Menu
      - Notification Center Event
      - Events Landing > Event Table Row
      - User Landing > User Table Row
      - User SSH Key Panel
      - Profile Display
      - Support Ticket Details (**test this in dev with a dummy support ticket you comment on**)
   - Confirm the sunsetting banner is visible on every page and is dismissible
 - For an account **without** a Gravatar:
   - Check the following places and confirm the new Avatar component is displayed:
      - Top Menu
      - Notification Center Event
      - Events Landing > Event Table Row
      - User Landing > User Table Row
      - User SSH Key Panel
      - Profile Display
      - Support Ticket Details  (**test this in dev with a dummy support ticket you comment on**)
   - Confirm the sunsetting banner is *not* visible
   - Confirm that you can `Change Avatar Color` on http://localhost:3000/profile/display and the changes are reflected everywhere the Avatar is displayed
- `yarn storybook` and check the Avatar and ColorPicker stories in Storybook
  
*Check all that apply*

- [x] 👀 Doing a self review


packages/manager/src/utilities/gravatar.ts
mjac0bs, line 1: Brought back this util from the dead (previous Gravatar work had a similar one) because we need to know the result of the request to fetch a Gravatar. If a user doesn't have one, the default avatar is returned, so that's not a valid way to check if there is one or not.

We use this util to check if we can render the new avatar to users who do *not* have Gravatars. It will be cleaned up once the sunsetting is complete.

packages/manager/src/types/ManagerPreferences.ts
mjac0bs, line 1: At least for now, with no API work required, we can store the user's avatar color in user preferences.

packages/manager/src/features/Profile/DisplaySettings/AvatarColorPickerDialog.tsx
```diff
@@ -0,0 +1,75 @@
...
+      >
+        <Avatar color={avatarColor} height={88} width={88} />
+      </Stack>
+
+      <ActionsPanel
+        primaryButtonProps={{
+          label: 'Save',
+          onClick: () => {
+            if (avatarColor) {
+              updatePreferences({
```
mjac0bs, line 57: We only update preferences once the dialog is confirmed, but by passing a color prop into `Avatar`, we can update the preview so the user can see the changes before they save.

packages/manager/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const { data: preferences } = usePreferences();
+  const { data: profile } = useProfile();
+
+  const _username = username ?? profile?.username ?? '';
+  const isAkamai = /^Linode$|^lke-service-account*/.test(_username);
+
+  const savedAvatarColor =
+    isAkamai || !preferences?.avatarColor
+      ? theme.palette.primary.dark
+      : preferences.avatarColor;
```
mjac0bs, line 62: Avatars from system-generated events have a default color and use the Akamai wave logo. If the user has not set a background color, we default to the same color and their initial. Otherwise, we'll use their saved preference.

mjac0bs (commented):
Leaving a self review since this change touches several files.

packages/manager/src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -0,0 +1,54 @@
...
+      </label>
+      <input
+        onChange={(e) => {
+          setColor(e.target.value);
+          onChange(e.target.value);
+        }}
+        color={color}
+        id="color-picker"
+        style={inputStyles}
+        type="color"
```
bnussman-akamai, line 49: Native color picker 🔥

packages/manager/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,95 @@
...
+      }}
+      alt={`Avatar for user ${username ?? profile?.email ?? ''}`}
+      data-testid="avatar"
+    >
+      {isAkamai ? (
+        <AkamaiWave />
+      ) : (
+        <Typography
+          sx={{
+            color: theme.palette.getContrastText(color ?? savedAvatarColor),
```
bnussman-akamai, line 85: Nice 🔥 🎨

mjac0bs: Ah yeah, I forgot to update you - I was very excited by this line! 🙌🏼

carrillo-erik (approved):
This looks really good, really makes the events pop out. There had been much discussions around Gravatar and it's exciting to see this work getting done.

packages/manager/src/features/Profile/DisplaySettings/DisplaySettings.tsx
```diff
@@ -89,31 +104,47 @@ export const DisplaySettings = () => {
...
+                    text={tooltipIconText}
+                  />
+                )}
               </Typography>
               <StyledProfileCopy variant="body1">
-                Create, upload, and manage your globally recognized avatar from
-                a single place with Gravatar.
+                {hasGravatar
+                  ? 'Create, upload, and manage your globally recognized avatar from a single place with Gravatar.'
+                  : 'Your profile photo is automatically generated using the first character of your username.'}
```
mjac0bs, line 134: Just realized that profile *photo* doesn't make a whole lot of sense if we are just using a colored background and letter, and we refer to it as an avatar in the button/dialog.

Reviewers: what do we think about changing it to "Avatar" for our Gravatarless users?

Screenshot 2024-09-05 at 5 01 43 PM

packages/manager/src/components/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,95 @@
...
+    width = DEFAULT_AVATAR_SIZE,
+  } = props;
+
+  const theme = useTheme();
+
+  const { data: preferences } = usePreferences();
+  const { data: profile } = useProfile();
+
+  const _username = username ?? profile?.username ?? '';
+  const isAkamai = /^Linode$|^lke-service-account*/.test(_username);
```
abailly-akamai, line 57: Could this work to avoid the regex?

```suggestion
  const isAkamai = _username === 'Linode' || _username.startsWith('lke-service-account');
```

abailly-akamai (commented):
Yayyyyy! Great PR, super clean, well tested and documented. I will spend more time with it in the next couple working days, but I left a comment for refactoring how we check for a gravatar.

packages/manager/src/features/Profile/DisplaySettings/DisplaySettings.tsx
abailly-akamai: Is more a "User Icon" than an "Avatar"? Not sure what's best

mjac0bs: image

🤷🏼‍♀️ We could say "Profile icon" everywhere... but it's more wordy.

I checked in with Kendall and Matthew - we've got consensus that "Profile photo" is inaccurate/misleading and they're good with Avatar, so I made that change in [bef6088](https://github.com/linode/manager/pull/10859/commits/bef60881e42faab15c8c098430a2456a7e1544bb).

abailly-akamai: works for me!

packages/manager/src/components/AccessPanel/UserSSHKeyPanel.tsx
```diff
@@ -67,6 +70,11 @@ const UserSSHKeyPanel = (props: Props) => {
 
   const { data: profile } = useProfile();
 
+  const [hasGravatar, setHasGravatar] = React.useState(false);
+  checkForGravatar(getGravatarUrl(profile?.email ?? '')).then((res) =>
+    setHasGravatar(res)
+  );
```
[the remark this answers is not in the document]
mjac0bs: Thanks, agreed that this is cleaner! Currently looking into this, as I was seeing a flash of content for the default state with this hook. 

If we set `hasGravatar` state to `false` by default or don't initialize with a value, then we've got the username/icon showing briefly.

https://github.com/user-attachments/assets/77978f09-d62a-431e-b74d-4e855109f945


If we set it to `true`, the opposite - the old profile icon for Gravatarless uses shows briefly.

packages/manager/src/components/Avatar/Avatar.tsx
mjac0bs: Sure, I didn't think this regex was too bad, but I'm fine with avoiding it altogether. Changed in [bb98f35](https://github.com/linode/manager/pull/10859/commits/bb98f35612e882f943283749496aeb72782a4cee).

While making the change, I also caught a bug fixed in [10f0742](https://github.com/linode/manager/pull/10859/commits/10f07423d801dc2bc0a4e2cb2f7b1834a15ae3a0) related to the event `username` being `null`:
| Bug | Fix |
| --- | --- |
|Screenshot 2024-09-06 at 10 14 02 AM | Screenshot 2024-09-06 at 10 13 44 AM|

packages/manager/src/components/AccessPanel/UserSSHKeyPanel.tsx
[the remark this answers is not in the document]
abailly-akamai: ah yeah, there a few ways to solve this (handling undefined & false or a loading state to wait for a value to load the Notification Center are things i can think of), happy to help if you want to push the code at some point.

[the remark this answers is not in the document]
mjac0bs: I just pushed up [95e2a76](https://github.com/linode/manager/pull/10859/commits/95e2a76b4f32e21bd04235fddf076d8f2a1fcadc)!

packages/manager/src/hooks/useGravatar.ts
```diff
@@ -0,0 +1,39 @@
+import { useEffect, useState } from 'react';
+
+import { checkForGravatar, getGravatarUrl } from 'src/utilities/gravatar';
+
+// set a cache to prevent duplicate requests
+const gravatarCache: Record<string, boolean> = {};
```
bnussman-akamai, line 6: Could we use a `Map`? It might allow us to make it even cleaner!

```suggestion
const gravatarCache = new Map<string, boolean>();
```

packages/manager/src/hooks/useGravatar.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @description
+ * This hook checks if a user has a Gravatar associated with their email.
+ * It uses a cache to prevent duplicate requests.
+ *
+ * @param email - The email address to check for a Gravatar.
+ * @returns - A boolean indicating whether the user has a Gravatar.
+ */
+export const useGravatar = (email: string | undefined) => {
+  const [hasGravatar, setHasGravatar] = useState(false);
```
bnussman-akamai, line 19: I think I see some flickering on http://localhost:3000/profile/display

Could get the default value from the cache to minimize flickering?

For example:
```suggestion
  const [hasGravatar, setHasGravatar] = useState(
    email && gravatarCache.has(email) ? gravatarCache.get(email) : false
  );
```

abailly-akamai: This helps with the profile page a little, but for the notification center and the events page we need to handle first load. We discussed this with @mjac0bs last week and agreed and refactoring things a little to render a placeholder  when `hasGravatar` is `undefined`

```
hasGravatar = undefined => placeholder
hasGravatar = false => icon
hasGravatar = true => gravatar
```

we also agreed we should have **one** component handling the logic above so we don't have to repeat it through the codebase. Heck, the hook could even return the JSX

mjac0bs: Thanks for the suggestion, @bnussman-akamai , and for summarizing, @abailly-akamai . 

[665915c](https://github.com/linode/manager/pull/10859/commits/665915c281054acef293de110b6c902aaf68fc4c) and [130aec7](https://github.com/linode/manager/pull/10859/commits/130aec79bf314adfb14209ad78d34452b811dd6c) address  the above with a `GravatarOrAvatar` component for handing the logic of what to render. In addition to width and height, the component also needed to take in either the Avatar/Gravatar components or their props to be able to display the right color/initial/size of the Avatar or Gravatar. (We also have the two Gravatar components - GravatarByUsername and GravatarByEmail, used depending on the situation.)

I have added a placeholder and some fade-ins. What I was seeing happen was that the request in `useGravatar` would return a response (so `isLoadingGravatar` was false, `hasGravatar` was true), but we were still waiting on the request to get the avatar in `GravatarByUsername` or `GravatarByEmail`, at which point the old Avatar with the UserIcon would still flicker briefly. Adding a fade in to the Gravatar components only seemed to hide that... but in some cases (Event Landing), it's a pretty long/noticable fade in. @abailly-akamai, what are your thoughts? 

Note: On the event landing page, I am seeing a bit of flickering with the default Avatar (i.e. for an Akamai system event) when logged in a user who had Gravatar enabled. Looking at that and the unit test failure now.

packages/manager/src/hooks/useGravatar.ts
abailly-akamai: I still like Banks suggestion here, map and get/set method is nicer

packages/manager/src/hooks/useGravatar.ts
abailly-akamai: This suggestion would still also help I think

abailly-akamai (commented):
This is looking great! some improvement can still be made but I really like the placeholder improvement 💅 and code cleanup 🧹

packages/manager/src/components/GravatarOrAvatar.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const { hasGravatar, isLoadingGravatar } = useGravatar(profile?.email);
+
+  return isLoadingGravatar ? (
+    <Box height={height} width={width} />
+  ) : hasGravatar ? (
+    gravatar
+  ) : (
+    avatar
+  );
+};
```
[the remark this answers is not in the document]
mjac0bs: We can... and an earlier version of this PR did pass props for the components instead of the components, but it honestly felt like more of a pain to do the clean up at the end of the 2 week sunset period. I want to be able to go in and just remove everything but `Avatar` in all the components that use `GravatarOrAvatar`.

packages/manager/src/features/Events/EventRow.styles.ts
```diff
@@ -1,10 +1,14 @@
...
 import { styled } from '@mui/material/styles';
 
+import { fadeIn } from 'src/styles/keyframes';
+
 import { GravatarByUsername } from '../../components/GravatarByUsername';
 
 export const StyledGravatar = styled(GravatarByUsername, {
   label: 'StyledGravatar',
 })(({ theme }) => ({
+  animation: `${fadeIn} 1.75s ease-out forwards`,
```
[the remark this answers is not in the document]
mjac0bs: It does feel slow, but anything faster than this resulted in visible flickering for me. I can do some more experimentation with the fade effect, but the event rows on the event landing page had the most egregious flickering.

packages/manager/src/components/GravatarOrAvatar.tsx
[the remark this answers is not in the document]
abailly-akamai: Ahh ok i wan't thinking of the two week period. You're right, it's totally fine

packages/manager/src/features/Events/EventRow.styles.ts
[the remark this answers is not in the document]
abailly-akamai: I see - this also may mean that the placeholder does not even render at all?

packages/manager/src/features/Linodes/LinodeCreatev2/Security.test.tsx
mjac0bs, line 1: I am confused. This change *did* fix the [CI failure](https://github.com/linode/manager/actions/runs/10796460560/job/29945717058) on the `test-manager` job, which is now passing [here](https://github.com/linode/manager/actions/runs/10797118782/job/29947835413?pr=10859), but it was a failure on a component not touched in this PR. There still seems to be something more going on here because Code Coverage / current branch is still [failing](https://github.com/linode/manager/actions/runs/10797118776/job/29948292801?pr=10859). Local runs of unit tests don't reproduce errors. 

Screenshot 2024-09-10 at 10 05 27 AM

mjac0bs: Reverted this change so it can be addressed in its own PR.

abailly-akamai (approved):
Great work! Confirmed Avatar/Gravatar behaviors are working as expected and async rendering is handled as best as possible.

Thanks for working with me on improving the flickering issues even if this feels like micro-improvements and who knows, the 2 weeks period may be expended and we have a good UI to support it!

Can't wait to remove a lot of logic related to having to fetch the gravatar and kiss goodbye to this unnecessary burden 👋 

Note: fine skipping the the failing test which has become a flake ❄️

packages/manager/src/components/GravatarByUsername.tsx
```diff
@@ -5,18 +5,24 @@ import UserIcon from 'src/assets/icons/account.svg';
...
   const { className, username } = props;
-  const { data: user } = useAccountUser(username ?? '');
+  const { data: user, isLoading } = useAccountUser(username ?? '');
   const url = user?.email ? getGravatarUrl(user.email) : undefined;
 
+  // Render placeholder instead of flashing default user icon briefly
+  if (isLoading) {
+    return <Box height={DEFAULT_AVATAR_SIZE} width={DEFAULT_AVATAR_SIZE} />;
+  }
+
```
mjac0bs, line 25: Latest change: 

Adds `isLoading` to render a placeholder for GravatarByUsername while we're waiting for the `user`.  This impacts the events landing and notification center gravatars. Without a placeholder here, while the query loaded, we were setting `url` to `undefined`, which rendered the default icon briefly and previously had me adding an extra long fade in (>1s) to avoid. With this actual fix, we still slightly (.2s) fade in the gravatar so the transition from loading to the icon is less abrupt.

Between this change and the cache in `useGravatar`, we shouldn't be seeing anymore flickering, even at slow speeds.

bnussman-akamai (approved):

PR #10865, opened 2024-08-30
change: increased block storage max volume size to 16TB

jmarmolm:
## Description 📝
The API is being updated to increase the max size for block storage volumes to 16TB, these are the changes necessary in the front end

## Changes  🔄
-  Increase Block storage max volume size to 16 TB


jaalah-akamai:
cc: @dajahi

jaalah-akamai (approved):
Looks good to me - I didn't notice any other places we may need to update this. I tagged others who are working on this feature to double check

coliu-akamai (commented):
Would we need a changeset for this @jaalah-akamai?

jaalah-akamai:
cc: @abailly-akamai - I updated the release PR and our internal changelog doc

abailly-akamai (approved):
Good to go. Side note, not sure we need to have three separate const for each package. APIv4 could be the source of truth for this one and we remove the others.

dwiley-akamai (approved):
> Good to go. Side note, not sure we need to have three separate const for each package. APIv4 could be the source of truth for this one and we remove the others.

I think it would make sense to get rid of the `/manager` one and import directly from `api-v4` for Cloud's purposes, but `validation` should be able to standalone with no `api-v4` imports (I did a quick search to see if we do any currently and I didn't find any)

PR #10882, opened 2024-09-04
test: [M3-6615, M3-8507] - Cypress test for empty Linode landing page and restricted user Linode landing page 

subsingh-akamai:
## Description 📝
Added cypress test for empty Linode landing page and restricted user Linode landing page  

## Changes  🔄
Added below cypress tests under in spec file - cypress/e2e/core/linodes/smoke-linode-landing-table.spec.ts
- checks empty state on linode landing page
- checks restricted user has no access to create linode on linode landing page

## PRs
- M3-6615
- M3-8507

## How to test 🧪
`yarn cy:run -s cypress/e2e/core/linodes/smoke-linode-landing-table.spec.ts`

### Verification steps
When test executed using this command  `yarn cy:run -s cypress/e2e/core/linodes/smoke-linode-landing-table.spec.ts`all tests should pass successfully and you should see new test added as per below screenshot 

image

*Check all that apply*

- [x] 👀 Doing a self review

### DIFF
--- a/packages/manager/.changeset/pr-10882-tests-1725437459486.md
+++ b/packages/manager/.changeset/pr-10882-tests-1725437459486.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Tests
+---
+
+Cypress test for empty Linode landing page and restricted user Linode landing page ([#10882](https://github.com/linode/manager/pull/10882))
--- a/packages/manager/cypress/e2e/core/linodes/smoke-linode-landing-table.spec.ts
+++ b/packages/manager/cypress/e2e/core/linodes/smoke-linode-landing-table.spec.ts
@@ -15,11 +15,17 @@ import { apiMatcher } from 'support/util/intercepts';
 import { chooseRegion, getRegionById } from 'support/util/regions';
 import { authenticate } from 'support/api/authentication';
 import { mockGetLinodes } from 'support/intercepts/linodes';
-import { userPreferencesFactory } from '@src/factories';
+import { userPreferencesFactory, profileFactory } from '@src/factories';
+import { accountUserFactory } from '@src/factories/accountUsers';
+import { grantsFactory } from '@src/factories/grants';
+import { mockGetUser } from 'support/intercepts/account';
 import {
   mockGetUserPreferences,
   mockUpdateUserPreferences,
+  mockGetProfile,
+  mockGetProfileGrants,
 } from 'support/intercepts/profile';
+import { randomLabel } from 'support/util/random';
 
 const mockLinodes = new Array(5).fill(null).map(
   (_item: null, index: number): Linode => {
@@ -32,6 +38,10 @@ const mockLinodes = new Array(5).fill(null).map(
 );
 
 const mockLinodesData = makeResourcePage(mockLinodes);
+
+const emptyLinode: Linode[] = [];
+
+const emptyLinodeData = makeResourcePage(emptyLinode);
 
 const sortByRegion = (a: Linode, b: Linode) => {
   return a.region.localeCompare(b.region);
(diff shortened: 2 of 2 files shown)


packages/manager/cypress/e2e/core/linodes/smoke-linode-landing-table.spec.ts
```diff
@@ -386,3 +396,115 @@ describe('linode landing checks', () => {
...
+      managed: false,
+    });
+
+    cy.intercept('GET', apiMatcher('account/settings'), (req) => {
+      req.reply(mockAccountSettings);
+    }).as('getAccountSettings');
+    cy.intercept('GET', apiMatcher('profile')).as('getProfile');
+    cy.intercept('GET', apiMatcher('linode/instances/*'), (req) => {
+      req.reply(emptyLinodeData);
+    });
```
jdamore-linode, line 412: Unless the test needs to do something unusual, we'll usually favor the mocking and intercept utils in `support/intercepts` to set up these mocks.

In this case, the equivalent would be:

```suggestion
    mockGetAccountSettings(mockAccountSettings).as('getAccountSettings');
    interceptGetProfile().as('getProfile');
    mockGetLinodes([]);
```

Although I'm wondering if we really need the account setting mocks or profile intercept! It might be worth trying something like this instead:

```typescript
mockGetLinodes([]).as('getLinodes');
cy.visitWithLogin('/linodes');
cy.wait('@getLinodes');
```

packages/manager/cypress/e2e/core/linodes/smoke-linode-landing-table.spec.ts
```diff
@@ -386,3 +396,115 @@ describe('linode landing checks', () => {
...
+        add_linodes: false,
+      },
+    });
+
+    mockGetProfile(mockProfile);
+    mockGetProfileGrants(mockGrants);
+    mockGetUser(mockUser);
+    cy.intercept('GET', apiMatcher('linode/instances/*'), (req) => {
+      req.reply(emptyLinodeData);
+    });
```
jdamore-linode, line 490:
```suggestion
    mockGetLinodes([]);
```
Same suggestion as my other comment!

packages/manager/cypress/e2e/core/linodes/smoke-linode-landing-table.spec.ts
```diff
@@ -386,3 +396,115 @@ describe('linode landing checks', () => {
...
+    cy.visitWithLogin('/', { preferenceOverrides });
+    cy.url().should('endWith', routes.linodeLanding);
```
jdamore-linode, line 492: This is fine in this case (and the other), but want to slightly caution against calling `cy.visit`/`cy.visitWithLogin` in `beforeEach` hooks just because it makes it more difficult for any tests in that scope to configure or add additional mocks: if `cy.visit` has already been called before a test has begun executing, there's no guarantee that any API requests it wants to intercept/mock haven't already been made.

To use this spec as an example, if we wanted to add any hypothetical new tests in this `describe` block, we'd be limited to the profile/grants setup you configured with the mocks in your `beforeEach`. That might be totally fine if all the tests in the `describe` are intended only to cover this specific permissions setup, but if that's the case then [having multiple tests might not be ideal anyway](https://docs.cypress.io/guides/references/best-practices#Creating-Tiny-Tests-With-A-Single-Assertion).

packages/manager/.changeset/pr-10882-tests-1725437459486.md
jdamore-linode, line 1: Thanks for the changeset!

jdamore-linode (commented):
Thanks @subsingh-akamai, nice work! Made a couple small suggestions and offered some advice but overall this is already really solid 👍

packages/manager/cypress/e2e/core/linodes/smoke-linode-landing-table.spec.ts
subsingh-akamai: Made changes in this commit https://github.com/linode/manager/pull/10882/commits/dbc04a5811cc5c60a8e16a7c8c058708ec04025f

packages/manager/cypress/e2e/core/linodes/smoke-linode-landing-table.spec.ts
subsingh-akamai: Made changes in this commit https://github.com/linode/manager/pull/10882/commits/dbc04a5811cc5c60a8e16a7c8c058708ec04025f

packages/manager/cypress/e2e/core/linodes/smoke-linode-landing-table.spec.ts
subsingh-akamai: Made changes in this commit https://github.com/linode/manager/pull/10882/commits/dbc04a5811cc5c60a8e16a7c8c058708ec04025f

jdamore-linode (approved):
Awesome work, thanks @subsingh-akamai!

coliu-akamai (approved):
✅ confirmed tests pass

Nice work and comments throughout the tests were super helpful! 🎉 
(also - just edited your PR comment so that the internal tickets aren't direct links - it shouldn't really matter since they're protected but just in case I guess?)

PR #10894, opened 2024-09-05
change: [M3-8530] - Allow quoted strings in Search v2 and improve StackScript searching on Linode Create v2 

bnussman-akamai:
## Description 📝

- Fixes https://github.com/linode/manager/issues/10869

### Allow spaces in search
- Allows users to use quoted strings (therefore spaces) when searching using Search v2 🔍 
  - For example, if you want to search for a StackScript with the label `My StackScript`, with this PR, you can search using a quoted string: `label: "My StackScript"`
    - ⚠️  Ideally, the user would be able to do `label: My StackScript` instead of `label: "My StackScript"`, but I couldn't quite get that working. I think it would require a significant rework of our search parser. I think a quoted string is probably a good step forward for now, but let me know if you disagree!
    
### Fix searching with `and` on StackScript tab of Linode Create
- Makes the user's search to have priority over the default StackScript filter 💪 
  - This fixes an issue where and queries would not work because they were overwritten by the default filter

## Preview 📷

|   | Before  | After   |
| -------| ------- | ------- |
| Search with string | Screenshot 2024-09-05 at 12 02 09 PM | Screenshot 2024-09-05 at 12 02 17 PM |
| Fix `and` queries | Screenshot 2024-09-05 at 12 03 25 PM | Screenshot 2024-09-05 at 12 03 15 PM |

## How to test 🧪

- Test the Search 🔍  functionality on http://localhost:3000/linodes/create?type=StackScripts&subtype=Community
  - Verify you can use spaces by using a quoted string
  - Verify you can use `and` in your search

- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/.changeset/pr-10894-added-1725553004847.md
+++ b/packages/manager/.changeset/pr-10894-added-1725553004847.md
@@ -2,4 +2,4 @@
 "@linode/manager": Added
 ---
 
-Allow quoted strings in Search v2 ([#10894](https://github.com/linode/manager/pull/10894))
+Support for quoted strings in Search v2 ([#10894](https://github.com/linode/manager/pull/10894))
--- a/packages/manager/.changeset/pr-10894-added-1725553004847.md
+++ b/packages/manager/.changeset/pr-10894-added-1725553004847.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Added
+---
+
+Allow quoted strings in Search v2 ([#10894](https://github.com/linode/manager/pull/10894))
--- a/packages/manager/.changeset/pr-10894-fixed-1725552964331.md
+++ b/packages/manager/.changeset/pr-10894-fixed-1725552964331.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Fixed
+---
+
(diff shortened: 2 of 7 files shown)


dwiley-akamai (approved):
Can use spaces by using a quoted string ✅
Can use `and` in the search ✅

How are users informed of the rules governing the search functionality?

abailly-akamai (approved):
Confirmed both cases now work as expected. 

FWIW, it would be nice for the tooltip to show an actual example because it's confusing with the quotes in there. 

we could change:

```
Hint: try searching for a specific item by prepending your search term with "username:", "label:", or "description:"
```

to 
```
Hint: try searching for a specific item by prepending your search term with `username:`, `label:`, or `description:`. example: `label: "my label"`
```

bnussman-akamai:
Good call @abailly-akamai 

I'll update the tooltip to include that example

bnussman-akamai:
Thoughts? @abailly-akamai @dwiley-akamai 

This might be a bit much

Screenshot 2024-09-06 at 12 28 16 PM

abailly-akamai:
@bnussman-akamai i think that's great! it's a bit intense but it's very clear 👍

dwiley-akamai:
That looks good to me as well 🚀

PR #10895, opened 2024-09-05
fix: [M3-7916] - Typo with Toast Success Notification For Updating Reverse DNS

coliu-akamai:
## Description 📝
- fix typo with toast success notification when updating reverse DNS

## Changes  🔄
- fix typo
- add e2e test to confirm typo is fixed
- added some unit tests
- other small cleanup

## Target release date 🗓️
n/a

## Preview 📷

| Before  | After   |
| ------- | ------- |
| image | image |

### Verification steps
- yarn cy:run -s cypress/e2e/core/linodes/linode-network.spec.ts
- Confirm toast says RDNS when updating RDNS in a Linode's network tab

*Check all that apply*

- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/cypress/e2e/core/linodes/linode-network.spec.ts
+++ b/packages/manager/cypress/e2e/core/linodes/linode-network.spec.ts
@@ -1,0 +1,89 @@
+import type { Linode } from '@linode/api-v4';
+
+import { createLinodeRequestFactory } from '@src/factories';
+
+import { authenticate } from 'support/api/authentication';
+import { interceptGetLinode } from 'support/intercepts/linodes';
+import { interceptUpdateIPAddress } from 'support/intercepts/networking';
+import { cleanUp } from 'support/util/cleanup';
+import { createTestLinode } from 'support/util/linodes';
+import { randomLabel } from 'support/util/random';
+import { chooseRegion } from 'support/util/regions';
+import { ui } from 'support/ui';
+
+authenticate();
+describe('linode networking', () => {
+  before(() => {
+    cleanUp('linodes');
+  });
+
+  /**
+   * - Confirms flow for editing the RDNS of an IP address
+   * - Confirms toast message upon editing RDNS
+   */
+  it('can edit the RDNS of an IP address', () => {
+    const createLinodeRequest = createLinodeRequestFactory.build({
+      label: randomLabel(),
+      region: chooseRegion().id,
+      backups_enabled: false,
+      booted: false,
+    });
+
+    cy.defer(
+      () => createTestLinode(createLinodeRequest),
+      'creating Linode'
+    ).then((linode: Linode) => {
+      interceptGetLinode(linode.id).as('getLinode');
+
+      cy.visitWithLogin(`linodes/${linode.id}/networking`);
+      cy.wait(['@getLinode']);
+
+      cy.findByLabelText('IPv4 Addresses')
+        .should('be.visible')
+        .within(() => {
+          // confirm table headers
+          cy.get('thead').findByText('Address').should('be.visible');
+          cy.get('thead').findByText('Type').should('be.visible');
+          cy.get('thead').findByText('Default Gateway').should('be.visible');
+          cy.get('thead').findByText('Subnet Mask').should('be.visible');
+          cy.get('thead').findByText('Reverse DNS').should('be.visible');
+        });
+
+      const linodeIPv4 = linode.ipv4.length > 0 ? linode.ipv4[0] : '';
+      interceptUpdateIPAddress(linodeIPv4).as('updateIPAddress');
+
+      // confirm row for Linode's (first) IPv4 exists and open up the RDNS drawer
+      cy.get(`[data-qa-ip="${linodeIPv4}"]`)
+        .should('be.visible')
+        .closest('tr')
+        .within(() => {
+          cy.findByText('IPv4 – Public').should('be.visible');
+
+          // open up the edit RDNS drawer
+          cy.get('[data-testid="action-menu-item-edit-rdns"]')
+            .should('be.visible')
+            .click();
+        });
+
+      // confirm RDNS drawer is visible
+      ui.drawer
+        .findByTitle('Edit Reverse DNS')
+        .should('be.visible')
+        .within(() => {
+          cy.findByText('Leave this field blank to reset RDNS').should(
+            'be.visible'
+          );
+
+          // click Save button
+          cy.findByText('Save')
+            .should('be.visible')
+            .should('be.enabled')
+            .click();
+        });
+
+      // confirm RDNS toast message
+      cy.wait('@updateIPAddress');
+      ui.toast.assertMessage(`Successfully updated RDNS for ${linodeIPv4}`);
+    });
+  });
+});
--- a/packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/EditIPRDNSDrawer.tsx
+++ b/packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/EditIPRDNSDrawer.tsx
@@ -45,17 +45,20 @@ export const EditIPRDNSDrawer = (props: Props) => {
     },
   });
 
-  React.useEffect(() => {
-    if (open) {
-      reset();
-      formik.resetForm();
-    }
-  }, [open]);
+  const onExited = () => {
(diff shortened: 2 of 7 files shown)


packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/EditIPRDNSDrawer.tsx
```diff
@@ -38,24 +39,27 @@ export const EditIPRDNSDrawer = (props: Props) => {
...
-  React.useEffect(() => {
-    if (open) {
-      reset();
-      formik.resetForm();
-    }
-  }, [open]);
+  const onExited = () => {
+    formik.resetForm();
+    reset();
+  };
```
coliu-akamai, line 52: see [here](https://github.com/linode/manager/pull/10791#discussion_r1720251672) for reasoning for this change

packages/manager/cypress/e2e/core/linodes/linode-network.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+      label: randomLabel(),
+      region: chooseRegion().id,
+      backups_enabled: false,
+      booted: false,
+    });
+
+    cy.defer(
+      () => createTestLinode(createLinodeRequest),
+      'creating Linode'
+    ).then((linode: Linode) => {
```
bnussman-akamai, line 35: Thoughts on using mocked data for this test? 

I'm all for full e2e tests, but our e2e runs take so long I want to be cautious of test time and resource creation.

bnussman-akamai (approved):
Looks good! Lots of nice clean up and testing 🧼 🧪

packages/manager/cypress/e2e/core/linodes/linode-network.spec.ts
coliu-akamai: actually yes! let me switch to mocks

packages/manager/cypress/e2e/core/linodes/linode-network.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+        .within(() => {
+          // confirm table headers
+          cy.get('thead').findByText('Address').should('be.visible');
+          cy.get('thead').findByText('Type').should('be.visible');
+          cy.get('thead').findByText('Default Gateway').should('be.visible');
+          cy.get('thead').findByText('Subnet Mask').should('be.visible');
+          cy.get('thead').findByText('Reverse DNS').should('be.visible');
+        });
+
+      const linodeIPv4 = linode.ipv4.length > 0 ? linode.ipv4[0] : '';
```
jdamore-linode, line 52: Correct me if I'm wrong: there's no scenario here where we'd expect `linode.ipv4` to be empty, and if it were, all bets are pretty much off for this test?

If so, I'd suggest just failing the test with a nice explanation, e.g.:

```typescript
// You could even put this right at the top of the block so the test can fail as early as possible.
const linodeIPv4 = linode.ipv4[0];
if (!linodeIPv4) {
  throw new Error('The created Linode does not have an IPv4 address. Cya!');
}
```

(Usually I try to steer clear of adding conditions in the test that change its flow, but since we're basically just bailing out of a guaranteed failure early I don't mind leaning on this pattern too much)

In the unlikely event this were ever to happen, the error message will explain what's going on and we won't have to scratch our head trying to figure out why Cypress is looking for an empty string 😅

packages/manager/cypress/e2e/core/linodes/linode-network.spec.ts
jdamore-linode, line 35:
```suggestion
    cy.defer(
      () => createTestLinode(null, { securityMethod: 'powered_off' }),
      'creating Linode'
    ).then((linode: Linode) => {
```

You can save a few lines here by using the [default payload](https://github.com/linode/manager/blob/a678228/packages/manager/cypress/support/util/linodes.ts#L108-L113)! Since we want a powered off Linode anyway, you can avoid having a Firewall created/attached by passing the `securityMethod: 'powered_off'` option (which will enforce the Linode to be created with `booted: false` regardless of whatever is set in the payload).

packages/manager/cypress/e2e/core/linodes/linode-network.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+        .within(() => {
+          cy.findByText('Leave this field blank to reset RDNS').should(
+            'be.visible'
+          );
+
+          // click Save button
+          cy.findByText('Save')
+            .should('be.visible')
+            .should('be.enabled')
+            .click();
```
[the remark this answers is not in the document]
jdamore-linode: That's actually useful context if we ever need to look back at this test after some time has passed. Maybe you can expand on that `// click Save button` comment to explain that we're only trying to trigger the toast and are intentionally not making any change to the form?

packages/manager/cypress/e2e/core/linodes/linode-network.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+      cy.get(`[data-qa-ip="${linodeIPv4}"]`)
+        .should('be.visible')
+        .closest('tr')
+        .within(() => {
+          cy.findByText('IPv4 – Public').should('be.visible');
+
+          // open up the edit RDNS drawer
+          cy.get('[data-testid="action-menu-item-edit-rdns"]')
+            .should('be.visible')
+            .click();
```
jdamore-linode, line 65:
```suggestion
          // open up the edit RDNS drawer
          ui.button
            .findByTitle('Edit RDNS')
            .should('be.visible')
            .should('be.enabled')
            .click();
```
Just a nit pick!

jdamore-linode (approved):
Nice work @coliu-akamai, thanks! The new test looks great!

coliu-akamai:
ty for the feedback!! Just pushed up changes to use mocks + add context and cleanup

packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/EditIPRDNSDrawer.tsx
```diff
@@ -80,7 +84,7 @@ export const EditIPRDNSDrawer = (props: Props) => {
             loading: isPending,
             type: 'submit',
           }}
-          secondaryButtonProps={{ label: 'cancel', onClick: onClose }}
```
hkhalil-akamai, line 83: Nice catch!

packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/EditIPRDNSDrawer.tsx
hkhalil-akamai, line 1: Not really in scope, but can we fix the spacing issue between the field and the helper text? Otherwise I can create a ticket for it:

Screenshot 2024-09-09 at 6 09 03 PM

hkhalil-akamai (approved):
Great catch 👀 and thanks for adding unit and cypress tests!

packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/EditIPRDNSDrawer.tsx
hkhalil-akamai: Looks good, thanks for handling this!

cpathipa (approved):
Nice work! confirming on the RDNS drawers functionality and tests.

packages/manager/cypress/e2e/core/linodes/linode-network.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+import { authenticate } from 'support/api/authentication';
+import {
+  mockGetLinodeDetails,
+  mockGetLinodeIPAddresses,
+  mockGetLinodeFirewalls,
+} from 'support/intercepts/linodes';
+import { mockUpdateIPAddress } from 'support/intercepts/networking';
+import { ui } from 'support/ui';
+
+authenticate();
```
jdamore-linode:
```suggestion
```

Totally harmless, but we don't need this anymore since we're using mocks and not making any real requests to the API from inside the test!

jdamore-linode (approved):
Nice work @coliu-akamai! The new test is really solid -- I ran it on repeat 15 times and it passed each time. Thanks again!

PR #10907, opened 2024-09-09
test: [M3-8433] - Cypress integration test for Object Storage Gen2: E1 Endpoint

jaalah-akamai:
## Changes  🔄
- New Integration Test for Object Storage Gen2: E1 Endpoint - Bucket Creation

## Target release date 🗓️
N/A

## How to test 🧪
- `yarn cy:run -s "cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts"`

Requirements:
- Visiting /object-storage/buckets/create causes "Create Bucket" drawer to open
- Users can select "E1" endpoint type for regions that support E1 endpoint types
    -  This will require mocking the response to the API endpoint types request to ensure that "E1" is available for the selected region
- Upon selecting "E1" endpoint type, request-per-second notice appears in the drawer:
    -  _This endpoint type supports up to 750 Requests Per Second (RPS). Understand bucket rate limits._
- Upon selecting "E1" endpoint type, rate limiting table is not present
- Upon clicking "Create Bucket" with "E1" endpoint type selected, Cloud fires an API bucket create request with the following data in its payload:
    -  `cors_enabled` is `true`
    -  `endpoint_type` is 'E1'
    -  `s3_endpoint` is present if corresponding endpoint mock contains non-null `s3_endpoint`
        -  At least one of these subtasks should test the case when `s3_endpoint` is specified, and at least one of these subtasks should test the case when `s3_endpoint` is omitted
-  Upon creating bucket, the "Create Bucket" drawer closes and new bucket appears in the landing page
    -  The correct information is displayed alongside the new bucket:
        -  Label
        -  Region
        -  Endpoint type
        -  Endpoint URL

*Check all that apply*

- [x] 👀 Doing a self review

### DIFF
--- a/packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
+++ b/packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
@@ -107,7 +107,7 @@ describe('Object Storage Gen2 create bucket tests', () => {
    * Confirms all endpoints are displayed regardless if there's multiple of the same type
    * Confirms S3 endpoint hostname displayed to differentiate between identical options in the dropdown
    */
-  it('can create a bucket with endpoint type 0', () => {
+  it('can create a bucket with E0 endpoint type', () => {
     const endpointTypeE0 = 'Legacy (E0)';
     const bucketLabel = randomLabel();
 
@@ -127,11 +127,6 @@ describe('Object Storage Gen2 create bucket tests', () => {
       cors_enabled: true,
       region: mockRegion.id,
     }).as('createBucket');
-
-    mockAppendFeatureFlags({
-      objMultiCluster: true,
-      objectStorageGen2: { enabled: true },
-    }).as('getFeatureFlags');
 
     mockGetObjectStorageEndpoints(mockEndpoints).as(
       'getObjectStorageEndpoints'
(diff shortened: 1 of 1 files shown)


packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
```diff
@@ -128,11 +128,6 @@ describe('Object Storage Gen2 create bucket tests', () => {
       region: mockRegion.id,
     }).as('createBucket');
 
-    mockAppendFeatureFlags({
```
mjac0bs, line 131: We can also remove [L139](https://github.com/linode/manager/pull/10907/files#diff-c275501373860ff8a10459d0d7f1957e0b05834f08785f8db7bbb1cd0677b27aR139), `@getFeatureFlags`.

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
```diff
@@ -238,10 +233,121 @@ describe('Object Storage Gen2 create bucket tests', () => {
...
+
+    mockGetBuckets([mockBucket]).as('getBuckets');
+    cy.wait(['@getBuckets']);
+
+    // Confirm request body has expected data
+    cy.wait('@createBucket').then((xhr) => {
+      const requestPayload = xhr.request.body;
+      expect(requestPayload['endpoint_type']).to.equal('E1');
+      expect(requestPayload['cors_enabled']).to.equal(true);
+    });
```
mjac0bs, line 315: > s3_endpoint is present if corresponding endpoint mock contains non-null s3_endpoint

Did we still want to verify the `s3_endpoint`?

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
```diff
@@ -238,10 +233,121 @@ describe('Object Storage Gen2 create bucket tests', () => {
...
   });
 
+  /**
+   * Confirms UI flow for creating a gen2 Object Storage bucket with endpoint E1
+   */
+  it('can create a bucket with E1 endpoint type', () => {
+    const endpointTypeE1 = 'Standard (E1)';
+    const bucketLabel = randomLabel();
+
+    //wait for the newly 'created' mocked bucket to appear
```
mjac0bs, line 243: Nit: Does this comment make more sense down before Line 307-308?

mjac0bs (approved):
Test passes and does what was required of it. 🚢 

I just left a couple minor comments, mostly clean up.

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
```diff
@@ -238,10 +233,121 @@ describe('Object Storage Gen2 create bucket tests', () => {
...
+    cy.wait('@createBucket').then((xhr) => {
+      const requestPayload = xhr.request.body;
+      expect(requestPayload['endpoint_type']).to.equal('E1');
+      expect(requestPayload['cors_enabled']).to.equal(true);
+    });
+
+    ui.drawer.find().should('not.exist');
+
+    // Confirm that bucket is created, initiate deletion for cleanup
+    cy.findByText(endpointTypeE1).should('be.visible');
```
mjac0bs, line 320: > The correct information is displayed alongside the new bucket:
Label
Region
Endpoint type
Endpoint URL

Just noticed - are we missing a check for the endpoint URL? I see endpoint type, label, and region.

abailly-akamai (commented):
I have no idea why, but when running 

`yarn cy:debug` all tests are passing, but with

`yarn cy:run -s "cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts"`
it's failing.

anyone else?

abailly-akamai (approved):
Approving since test passes as expected locally and in CI, but curious to know if @jdamore-linode has any insight into the discrepancy shown discussed above

jdamore-linode:
@abailly-akamai @mjac0bs @jaalah-akamai I'm able to reproduce the failures via `yarn cy:run`, too -- in my case, I'm seeing all 3 fail intermittently, and I think I see two reasons why:

- Sometimes the "Object Storage Endpoint Type" drop-down doesn't get populated. Judging by the recording (first vid), Cypress is clicking the "Create Bucket" button before the OBJ landing page has finished loading. In the past, we've had issues where drawers behave unexpectedly when they're opened before certain API requests have resolved, and this resembles that type of issue.
  - Normally I don't consider this type of bug that serious since the user usually has to race as quickly as they can to trigger it, but because the OBJ landing page takes so long to load we might want to take a closer look at this and rule it out as an issue
  - If this really is causing failures, we can work around it in Cypress by waiting for the page to load (using `cy.findByText()` to find one of the blurbs that appears once the page has loaded is usually reliable)
- In each test, we're clicking "Create Bucket" [and _then_ setting up the mocks](https://github.com/linode/manager/blob/516ae68/packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts#L194-L202) that would reflect the new bucket being created. If the browser fires off an HTTP request in the split second before Cypress has set up the request intercept, the request won't be mocked.
  - I think this is what's causing most of the flakiness, and probably explains why this works in `yarn cy:debug` but not in `yarn cy:run`: the debugger is probably _just_ slow enough compared to `cy:run` not to trigger the issue
  - We're doing this twice in each test: when the bucket is created, and when the bucket is deleted, so it'll need to be fixed for all 6 instances
  - To fix, we just have to move `mockGetBuckets()` to right before the preceding command that ends with `.click()`.

https://github.com/user-attachments/assets/ae01302a-1d6f-4bc3-aeab-93596bed1ea6

https://github.com/user-attachments/assets/1ecb59e9-7d9a-496e-9bfa-1a68ee5844df

cc @jaalah-akamai -- Sorry about this! I should've caught this when I reviewed the original E0 test.

jaalah-akamai:
Thanks @jdamore-linode I'll make those changes and confirm back here when it's done 👍

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
jaalah-akamai: Endpoint URL is the s3_endpoint - added 👍

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
jaalah-akamai: Moved it as far down as I could given placement now

packages/manager/cypress/e2e/core/objectStorageGen2/bucket-create-gen2.spec.ts
jaalah-akamai: Howcome?

PR #10913, opened 2024-09-10
fix: [M3-8553] - DisplayPrice story crash when Currency component's `minimumFractionDigits` is negative

pmakode-akamai:
## Description 📝
The Storybook for DisplayPrice is crashing when the `minimumFractionDigits` in the Currency component is set with a negative value.

This PR addresses the issue by updating the Currency component to handle negative decimalPlaces props correctly. Specifically, it ensures that the `minimumFractionDigits` defaults to 2 when the `decimalPlaces `prop is `negative` or `undefined`.

## Changes  🔄
- Handle negative decimalPlaces prop in Currency component to fix DisplayPrice story
   - The `minimumFractionDigits` in the Currency component will now use the default value (2) when `decimalPlaces` prop is `negative` or `undefined`.
- Add a unit test case to cover the above scenario.
- Update all other unit test cases for more clarity.

## Target release date 🗓️
N/A

### Reproduction steps
- Go to the DisplayPrice storybook
- Enter a negative value for decimalPlaces prop

### Verification steps
- Run storybook locally and verify that the Story for DisplayPrice should not crash when the decimalPlaces prop is negative.
- Confirm that the The `minimumFractionDigits` in the Currency component uses the default value (2) when `decimalPlaces` prop is `negative` or `undefined`.
- Verify that all the unit test cases pass.

*Check all that apply*

- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/src/components/Currency/Currency.test.tsx
+++ b/packages/manager/src/components/Currency/Currency.test.tsx
@@ -8,69 +8,79 @@ import { Currency } from './Currency';
 describe('Currency Component', () => {
   it('displays a given quantity in USD', () => {
     const { getByText, rerender } = renderWithTheme(<Currency quantity={5} />);
-    getByText('$5.00');
+    expect(getByText('$5.00')).toBeVisible();
     rerender(<Currency quantity={99.99} />);
-    getByText('$99.99');
+    expect(getByText('$99.99')).toBeVisible();
     rerender(<Currency quantity={0} />);
-    getByText('$0.00');
+    expect(getByText('$0.00')).toBeVisible();
   });
 
   it('handles negative quantities', () => {
     const { getByText, rerender } = renderWithTheme(<Currency quantity={-5} />);
-    getByText('-$5.00');
+    expect(getByText('-$5.00')).toBeVisible();
     rerender(<Currency quantity={-99.99} />);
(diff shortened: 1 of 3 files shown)


packages/manager/src/components/Currency/Currency.test.tsx
```diff
@@ -8,69 +8,79 @@ import { Currency } from './Currency';
 describe('Currency Component', () => {
   it('displays a given quantity in USD', () => {
     const { getByText, rerender } = renderWithTheme(<Currency quantity={5} />);
-    getByText('$5.00');
+    expect(getByText('$5.00')).toBeVisible();
```
hana-linode, line 11: Personally, I don't think this change is necessary; the line is implicitly clear

hana-linode (approved):

packages/manager/src/components/Currency/Currency.test.tsx
pmakode-akamai: I added `expect(....).toBeVisible()` to ensure the element is not only present but also visible, making the test clearer and more robust. Since you mentioned this might be unnecessary, should I revert this change?

hana-linode: No, you're good to merge

PR #10914, opened 2024-09-10
fix: [M3-5753] - Linode Create v2 not handling deprecated and EOL Images 

bnussman-akamai:
## Description 📝

This PR addresses functionality that was missed on Linode Create v2

## Changes  🔄
- Adds `(deprecated)` to Linux distributions that are deprecated on Linode Create v2
- Hides Linux distributions that are > 6 months past their EOL (end of life) on Linode Create 🙈

## Target release date 🗓️

9/16/2024

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-09-10 at 1 07 04 PM | Screenshot 2024-09-10 at 1 06 52 PM |

## How to test 🧪

- Go to http://localhost:3000/linodes/create
- Test the Image/distribution/OS select component and verify the following is true (see M3-5753 for context on why it works like this)
  - Images that have `deprecated: true` show `(deprecated)`
  - Images that are past their `eol` date show `(deprecated)`
  - Images that are > 6 months past their `eol` are not shown in the select 
- Using our local dev tools, toggle between Linode Create v1 and v2 and verify they behave the same
  - You may see `(deprecated)` on a few more options on v2 because we now also respect the  `deprecated ` field returned by the API in addition to the `eol`
-  Verify that private images (on the `Images` tab of Linode Create) are not negatively affected by this change
 
- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/src/components/ImageSelectv2/utilities.ts
+++ b/packages/manager/src/components/ImageSelectv2/utilities.ts
@@ -1,5 +1,8 @@
-import { ImageSelectVariant } from './ImageSelectv2';
+import { DateTime } from 'luxon';
 
+import { MAX_MONTHS_EOL_FILTER } from 'src/constants';
+
+import type { ImageSelectVariant } from './ImageSelectv2';
 import type { Image } from '@linode/api-v4';
 
 /**
@@ -21,11 +24,11 @@ export const getAPIFilterForImageSelect = (
 };
 
 /**
- * Using API filter, I can't think of a way to filter out
- * LKE images that we don't want customers to see.
+ * Unfortunately, we can't use API filtering for all of our filtering needs.
  *
- * This function exists to filter out public Kubernetes images
- * for the Images Select.
+ * This function exists to...
+ * - filter out public Kubernetes images for the Images Select
+ * - filter out images that are too far past their end-of-life date
  *
  * Please use API filtering (getAPIFilterForImageSelect) when possible!
  */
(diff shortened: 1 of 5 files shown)


packages/manager/cypress/e2e/core/stackscripts/create-stackscripts.spec.ts
```diff
@@ -374,7 +374,7 @@ describe('Create stackscripts', () => {
          */
         filteredImageData?.forEach((imageSample: Image) => {
           const imageLabel = imageSample.label;
-          cy.findAllByText(imageLabel)
+          cy.findAllByText(imageLabel, { exact: false })
```
bnussman-akamai, line 377: Had to add `exact: false` so the test can assert images that may have `(deprecated)` in the label

packages/manager/src/components/ImageSelectv2/utilities.ts
```diff
@@ -21,19 +24,67 @@ export const getAPIFilterForImageSelect = (
...
+ */
+export const isImageDeprecated = (image: Image) => {
+  if (image.eol === null) {
+    // If the image does not have an EOL, just use the `deprecated` field returned by the API
+    return image.deprecated;
+  }
+
+  const isImageEOL =
+    DateTime.fromISO(image.eol).toUnixInteger() <
+    DateTime.now().toUnixInteger();
```
hkhalil-akamai, line 87:
```suggestion
  const isImageEOL = DateTime.fromISO(image.eol) < DateTime.now();
```

hkhalil-akamai (approved):
Great catch -- looks good!

packages/manager/src/components/ImageSelectv2/utilities.ts
bnussman-akamai: Good call, much cleaner 🧼

packages/manager/src/components/ImageSelectv2/utilities.ts
```diff
@@ -21,19 +24,63 @@ export const getAPIFilterForImageSelect = (
...
  */
 export const getFilteredImagesForImageSelect = (
   images: Image[] | undefined,
   variant: ImageSelectVariant | undefined
 ) => {
-  return variant === 'private'
-    ? images
-    : images?.filter((image) => !image.id.includes('kube'));
+  if (variant === 'public') {
+    // For public images, we filter out LKE images and images that are > 6 months part their `eol`
```
dwiley-akamai, line 40:
```suggestion
    // For public images, we filter out LKE images and images that are > 6 months past their `eol`
```

packages/manager/src/components/ImageSelectv2/utilities.ts
```diff
@@ -21,19 +24,63 @@ export const getAPIFilterForImageSelect = (
...
+
+  const differenceInMonths = now.diff(imageEOL, 'months').months;
+
+  return differenceInMonths > MAX_MONTHS_EOL_FILTER;
+};
+
+/**
+ * Returns whether or not an image is deprecated
+ *
+ * Unfortunately, as per `M3-5753`, we can't just simpily check `deprecated` on the `Image`,
```
dwiley-akamai, line 74:
```suggestion
 * Unfortunately, as per `M3-5753`, we can't just simply check `deprecated` on the `Image`,
```

dwiley-akamai (approved):
Code review ✅
Images show `(deprecated)` under the described conditions ✅
No adverse impacts on private Images observed ✅

PR #10915, opened 2024-09-10
test: [M3-8316,M3-8317] - Tag cypress tests by adding the "method:e2e" and "purpose:dcTesting"

AzureLatte:
## Description 📝
Tag tests by adding the "method:e2e" tag and "purpose:dcTesting".

## Changes  🔄
- Use cy.tag() to add the "method:e2e" tag to all of our end-to-end Cypress tests.
- Use cy.tag() to add the "purpose:dcTesting" tag to all of our suitable Cypress tests.
- Add new intercept utils in linodes.ts and replace them in end-to-end tests.

## How to test 🧪
CI tests all pass. Or running tests locally using:
```
yarn cy:run
```

*Check all that apply*

- [x] 👀 Doing a self review


packages/manager/cypress/e2e/core/linodes/linode-storage.spec.ts
```diff
@@ -105,10 +112,7 @@ describe('linode storage tab', () => {
   it('try to delete in use disk', () => {
     const diskName = 'Debian 11 Disk';
     cy.defer(() => createTestLinode({ booted: true })).then((linode) => {
-      cy.intercept(
-        'DELETE',
-        apiMatcher(`linode/instances/${linode.id}/disks/*`)
-      ).as('deleteDisk');
+      interceptDeleteDisks(linode.id).as('deleteDisk');
```
jdamore-linode, line 115: Thanks for catching and cleaning all of these intercepts up! 🧼

packages/manager/cypress/e2e/core/nodebalancers/smoke-create-nodebal.spec.ts
```diff
@@ -78,6 +78,10 @@ const createNodeBalancerWithUI = (
 };
 
 authenticate();
+beforeEach(() => {
+  cy.tag('method:e2e');
+  cy.tag('purpose:dcTesting');
```
jdamore-linode, line 83: `cy.tag()` will override previous calls to `cy.tag()`, so these tests only end up with the `purpose:dcTesting` tag.

You can set both tags at once with `cy.tag('method:e2e', 'purpose:dcTesting')` which is probably the right call in this case, but we also have `cy.addTag()` to add tags without overriding existing tags.

I think there are a couple places we'll want to make this fix!

jdamore-linode (approved):
This is so cool, thanks @AzureLatte!

I'm almost looking forward to having to run our next DC tests* -- it's going to be so nice to call `CY_TEST_TAGS='purpose:dcTesting' yarn cy:run` and not have to wait for every test to run.

I did notice a couple spots where we call `cy.tag()` multiple times and made a suggestion around that, but otherwise this is good to go! Thanks again for the extra clean up!!

abailly-akamai (approved):
Thanks! Approving pending fix suggested

PR #10916, opened 2024-09-10
test: [M3-8556] - Add unit tests for NodeBalancer details package

coliu-akamai:
## Description 📝
- Add unit tests for NodeBalancer details package
- Some eslint changes/other small cleanup

## Target release date 🗓️
n/a

## How to test 🧪
- yarn test NodeBalancerDetail

*Check all that apply*

- [x] 👀 Doing a self review


packages/manager/src/features/NodeBalancers/NodeBalancerDetail/NodeBalancerSettings.tsx
```diff
@@ -155,5 +156,3 @@ export const NodeBalancerSettings = () => {
     </div>
   );
 };
-
-export default NodeBalancerSettings;
```
coliu-akamai, line 159: This file had both a named and a default export. Removed the default export in favor of the named one

packages/manager/src/features/NodeBalancers/NodeBalancerDetail/NodeBalancerFirewallsRow.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+    expect(getByText('Unassign')).toBeVisible();
+  });
+
+  it('unassigns the firewall', () => {
+    const { getByText } = renderWithTheme(
+      <NodeBalancerFirewallsRow {...props} />
+    );
+
+    const unassignButton = getByText('Unassign');
+    fireEvent.click(unassignButton);
```
bnussman-akamai, line 62: Can we use `await userEvent` over `fireEvent` wherever possible?

I believe userEvent is slightly better for testing actual user interactions: https://testing-library.com/docs/user-event/intro/#differences-from-fireevent

coliu-akamai: done! + switched to userEvent where possible for earlier NodeBalancer tests I'd already merged in

packages/manager/src/features/NodeBalancers/NodeBalancerDetail/NodeBalancerConfigurations.test.tsx
```diff
@@ -0,0 +1,169 @@
...
+});
+
+describe('NodeBalancerConfigurations', () => {
+  beforeEach(() => {
+    server.resetHandlers();
+  });
+
+  it('renders the NodeBalancerConfigurations component with one configuration', async () => {
+    server.use(
+      http.get(`*/nodebalancers/${String(NaN)}/configs`, () => {
```
hana-linode, line 34: What's with the `String(NaN)`?

coliu-akamai: I had trouble rendering the expected component given the props I was passing in - when I checked the ID of whatever NodeBalancer I was getting, it was `NaN` 😖, so I decided to mock that. 

I'm not really a fan of `NaN` either, so let me look into trying to mock an actual ID and see if that works 🙏

coliu-akamai: update: pushed up a change that gets rid of NaN - thanks for calling this out Hana! Helped me learn more about testing and fix my weird janky solution 😄

packages/manager/src/utilities/testHelpers.tsx
```diff
@@ -98,7 +101,11 @@ export const wrapWithTheme = (ui: any, options: Options = {}) => {
           >
             <SnackbarProvider>
               <MemoryRouter {...options.MemoryRouter}>
-                {ui.children ?? ui}
+                {routePath ? (
+                  <Route path={routePath}>{uiToRender}</Route>
+                ) : (
+                  uiToRender
+                )}
```
coliu-akamai, line 108: I'd tried just using MemoryRouter, but it wasn't until I added a Route that the tests passed - seems a Router is [needed based on this medium article](https://medium.com/@adamjarling/mocking-a-react-router-match-object-in-your-component-tests-fa95904dcc55) for match props. 

I'd tried doing something similar to MemoryRouter, spreading the props of a Router but ran into some typing issues - will look into this some more! Alternatively, if we don't want to change the `wrapWithTheme` helper function (especially if this use case is too narrow) I can do something like LinodeSummary.test.tsx:

```
<MemoryRouter initialEntries={['nodebalancers/1/configurations']}>
  <Route path="nodebalancers/:nodeBalancerId">
    <NodeBalancerConfigurations {...props} />
  </Route>
</MemoryRouter>
```

hana-linode: This looks a lot better than `${String(NaN)}`! I don't have a strong preference of how we're using MemoryRouter/Route so this is fine to me

hana-linode (approved):

bnussman-akamai (approved):

PR #10920, opened 2024-09-11
fix: [M3-8563 & M3-8566] – Fix BSE capability for linodes & prevent unnecessary API request on Volume Create page

dwiley-akamai:
## Description 📝
- Change BSE capability for Linodes from `blockstorage_encryption` --> `Block Storage Encryption`
- Adjust `enabled` condition passed to `useLinodeQuery()` on Volume Create page so it is true only if `linode_id !== null`

## Target release date 🗓️
9/16/24

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screenshot 2024-09-11 at 10 55 05 AM | Screenshot 2024-09-11 at 11 01 21 AM |

### Prerequisites
Point at the dev environment with the `blockstorage-encryption` tag on your account

### Verification steps
On the Volume Create page, you should no longer see requests to `linode/instances/-1` before selecting a linode.

- [X] 👀 Doing a self review
- [X] ➕ Adding a changeset

### DIFF
--- a/packages/api-v4/.changeset/pr-10920-added-1726070878408.md
+++ b/packages/api-v4/.changeset/pr-10920-added-1726070878408.md
@@ -1,0 +1,5 @@
+---
+"@linode/api-v4": Added
+---
+
+LinodeCapabilities type used for `capabilities` property of Linode interface ([#10920](https://github.com/linode/manager/pull/10920))
--- a/packages/api-v4/src/linodes/types.ts
+++ b/packages/api-v4/src/linodes/types.ts
@@ -19,7 +19,7 @@ export interface Linode {
   id: number;
   alerts: LinodeAlerts;
   backups: LinodeBackups;
-  capabilities?: string[]; // @TODO BSE: Remove optionality once BSE is fully rolled out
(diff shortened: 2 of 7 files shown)


bnussman-akamai (approved):
Looks good!

Should we update the `capabilities` type to be more strict than just `string`? 

https://github.com/linode/manager/blob/4d8a24113568ceee0a0a0e788915cacdb0e48461/packages/api-v4/src/linodes/types.ts#L22

mjac0bs (approved):
👍🏼  Extra requests are gone.

Unrelated to this PR, while testing, I did notice a console state error that's happening once a Linode is selected and the config loads - but only initially. This probably existed before BSE changes and can be its own ticket, but mentioning here in case you had any recent context.

https://github.com/user-attachments/assets/7a5c2e0b-764f-43b3-b28a-20737e716670

coliu-akamai (approved):
Confirmed extra requests to -1 are gone!

dwiley-akamai:
> Unrelated to this PR, while testing, I did notice a console state error that's happening once a Linode is selected and the config loads - but only initially. This probably existed before BSE changes and can be its own ticket, but mentioning here in case you had any recent context.

Maybe it has something to do with the replacement of `Select` with `Autocomplete` a few months ago?

PR #10924, opened 2024-09-11
fix: [M3-8161] - Fix misaligned DNS banner text

hkhalil-akamai:
## Description 📝
Fixes an alignment issue in the DNS zones banner. This banner appears in the domains landing when the user has a DNS zone but no Linodes on their account.

## Preview 📷
|Breakpoint|Before|After|
|--|--|--|
|`lg`|Screenshot 2024-09-11 at 3 59 48 PM|Screenshot 2024-09-11 at 3 59 56 PM|
|`md`|Screenshot 2024-09-11 at 4 00 28 PM|Screenshot 2024-09-11 at 4 00 23 PM
|`sm`|Screenshot 2024-09-11 at 4 00 45 PM|Screenshot 2024-09-11 at 4 00 53 PM|

### DIFF
--- a/packages/manager/src/features/Domains/DomainBanner.tsx
+++ b/packages/manager/src/features/Domains/DomainBanner.tsx
@@ -4,6 +4,7 @@ import * as React from 'react';
 
 import { DismissibleBanner } from 'src/components/DismissibleBanner/DismissibleBanner';
 import { Link } from 'src/components/Link';
+import { Stack } from 'src/components/Stack';
 import { Typography } from 'src/components/Typography';
 
 interface DomainBannerProps {
@@ -29,7 +30,7 @@ export const DomainBanner = React.memo((props: DomainBannerProps) => {
       preferenceKey={KEY}
       variant="warning"
     >
-      <>
+      <Stack>
         <Typography>
           <strong>Your DNS zones are not being served.</strong>
         </Typography>
@@ -38,7 +39,7 @@ export const DomainBanner = React.memo((props: DomainBannerProps) => {
           you have at least one active Linode on your account.{` `}
           <Link to="/linodes/create">You can create one here.</Link>
         </Typography>
-      </>
+      </Stack>
     </StyledDismissibleBanner>
   );
 });
@@ -47,6 +48,6 @@ const StyledDismissibleBanner = styled(DismissibleBanner, {
   label: 'StyledDismissableBanner',
 })(({ theme }) => ({
   '& h3:first-of-type': {
-    marginBottom: theme.spacing(1),
+    margin: theme.spacing(1),
   },
 }));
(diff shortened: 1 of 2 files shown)


packages/manager/src/features/Domains/DomainBanner.tsx
```diff
@@ -43,12 +43,6 @@ export const DomainBanner = React.memo((props: DomainBannerProps) => {
   );
 });
 
-const StyledTypography = styled(Typography, { label: 'StyledTypography' })(
-  ({ theme }) => ({
-    marginBottom: theme.spacing(),
```
hkhalil-akamai, line 48: Unsure of the original purpose of this margin.

packages/manager/src/features/Domains/DomainBanner.tsx
```diff
@@ -30,9 +30,9 @@ export const DomainBanner = React.memo((props: DomainBannerProps) => {
       variant="warning"
     >
       <>
-        <StyledTypography>
+        <Typography>
           <strong>Your DNS zones are not being served.</strong>
-        </StyledTypography>
+        </Typography>
```
dwiley-akamai, line 35: As you shrink the window, there are widths where this text and the other piece of text get very close to each other (example below). Maybe some padding or margin can be added on the right here to ensure there's always a gap?

<details>
<summary>Example (1063px width)</summary>

Screenshot 2024-09-11 at 4 45 16 PM
</details>

dwiley-akamai (approved):
Alignment improvement verified ✅

packages/manager/src/features/Domains/DomainBanner.tsx
pmakode-akamai: Perhaps we could keep `StyledTypography` and replace `marginBottom` with `marginRight`?

hkhalil-akamai:
@dwiley-akamai @harsh-akamai I'm experimenting with using `Stack` to layout the elements vertically. This resolves the spacing issue and looks better in my opinion. Are there any objections to the following:

|Breakpoint|Screenshot|
|-|-|
|`lg` and `md`| Screenshot 2024-09-12 at 1 51 07 PM|
|`sm`|Screenshot 2024-09-12 at 1 51 52 PM|

pmakode-akamai:
> @dwiley-akamai @pmakode-akamai I'm experimenting with using `Stack` to layout the elements vertically. This resolves the spacing issue and looks better in my opinion.

Nice work 🚀 Looks good to me!

cpathipa (approved):
DNS Banner alignment verified.

dwiley-akamai:
> @dwiley-akamai @pmakode-akamai I'm experimenting with using `Stack` to layout the elements vertically. This resolves the spacing issue and looks better in my opinion. Are there any objections to the following:

I like this much better as well 👍🏾

PR #10927, opened 2024-09-12
fix: [M3-8519] - Objects Table Refreshing Logic

dchyrva-akamai:
## Description 📝
After uploading several files into the object storage, file table was not updated correctly. In the provided fix, fresh data from the store is used during the store overriding.

## Changes  🔄
- Object buckets query key moved into a separate getter function and exported.
- Fresh store data is used during the overriding process.

## How to test 🧪
- Open the "Object Storage" page.
- Upload several files using drop area or browse button.
- Look at how the table is being updated after each file is uploaded.

*Check all that apply*

- [x] 👀 Doing a self review

### DIFF
--- a/packages/manager/src/features/ObjectStorage/BucketDetail/BucketDetail.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketDetail/BucketDetail.tsx
@@ -1,5 +1,5 @@
 import { getObjectList, getObjectURL } from '@linode/api-v4/lib/object-storage';
-import { useQueryClient } from '@tanstack/react-query';
+import { InfiniteData, useQueryClient } from '@tanstack/react-query';
 import produce from 'immer';
 import { useSnackbar } from 'notistack';
 import * as React from 'react';
@@ -21,14 +21,12 @@ import { OBJECT_STORAGE_DELIMITER } from 'src/constants';
 import { useFlags } from 'src/hooks/useFlags';
 import { useAccount } from 'src/queries/account/account';
 import {
+  getObjectBucketObjectsQueryKey,
   objectStorageQueries,
   useObjectBucketObjectsInfiniteQuery,
   useObjectStorageBuckets,
 } from 'src/queries/object-storage/queries';
-import {
-  fetchBucketAndUpdateCache,
-  prefixToQueryKey,
-} from 'src/queries/object-storage/utilities';
+import { fetchBucketAndUpdateCache } from 'src/queries/object-storage/utilities';
 import { isFeatureEnabledV2 } from 'src/utilities/accountCapabilities';
 import { sendDownloadObjectEvent } from 'src/utilities/analytics/customEventAnalytics';
 import { getQueryParamFromQueryString } from 'src/utilities/queryParams';
@@ -237,11 +235,7 @@ export const BucketDetail = (props: Props) => {
       pageParams: string[];
       pages: ObjectStorageObjectList[];
     }>(
-      [
-        ...objectStorageQueries.bucket(clusterId, bucketName)._ctx.objects
-          .queryKey,
-        ...prefixToQueryKey(prefix),
-      ],
+      getObjectBucketObjectsQueryKey(clusterId, bucketName, prefix),
       (data) => ({
         pageParams: data?.pageParams || [],
         pages,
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/ObjectStorage/BucketDetail/BucketDetail.tsx
```diff
@@ -279,7 +280,13 @@ export const BucketDetail = (props: Props) => {
   };
 
   const addOneFile = (objectName: string, sizeInBytes: number) => {
-    if (!data) {
+    const currentData:
+      | InfiniteData<ObjectStorageObjectList, unknown>
+      | undefined = queryClient.getQueryData(
+      getObjectBucketObjectsQueryKey(clusterId, bucketName, prefix)
+    );
```
bnussman-akamai, line 287: I think we can accomplish the same thing by setting the generic type of `getQueryData`

```suggestion
    const currentData = queryClient.getQueryData<
      InfiniteData<ObjectStorageObjectList>
    >(getObjectBucketObjectsQueryKey(clusterId, bucketName, prefix));
```

bnussman-akamai (commented):
Changes look good. Why we need to get the data using `getQueryData` rather than just using the `data` returned by the hook?

Also, we might be able to update the `updateStore` function to use the `getObjectBucketObjectsQueryKey` util you created

packages/manager/src/features/ObjectStorage/BucketDetail/BucketDetail.tsx
dchyrva-akamai: Good idea, thank you!

abailly-akamai (approved):
Nice concise fix @dchyrva-akamai 👍 

Code looks good and confirmed fix with objects now showing up in the table underneath as they are uploaded. ✅ 

I also tested Obj feature as a whole (delete operations etc) and did not notice any regression ✅

dchyrva-akamai:
> Changes look good. Why we need to get the data using `getQueryData` rather than just using the `data` returned by the hook?

When several objects are added simultaneously, the component is re-rendered only after all of them have been added. This means that the `data` from the hook is not updated with new values after each individual object is added, which causes a bug.

jaalah-akamai:
> > Changes look good. Why we need to get the data using `getQueryData` rather than just using the `data` returned by the hook?
> 
> Hello @bnussman-akamai
> 
> When several objects are added simultaneously, the component is re-rendered only after all of them have been added. This means that the `data` from the hook is not updated with new values after each individual object is added, which causes a bug.

First of all, nice work - this works. I just want to toss out an alternative... I'm interested in what you all think about about reverting back to using the data from the hook (I like the term `currentData` so maybe just do `data: currentData`). Then invalidate the cache at the end of the `maybeAddObjectToTable` function:
```
if (bucket) {
      queryClient.invalidateQueries({
        queryKey: getObjectBucketObjectsQueryKey(
          clusterId,
          bucket.label,
          prefix
        ),
      });
    }
```

dchyrva-akamai:
> > > Changes look good. Why we need to get the data using `getQueryData` rather than just using the `data` returned by the hook?
> > 
> > 
> > Hello @bnussman-akamai
> > When several objects are added simultaneously, the component is re-rendered only after all of them have been added. This means that the `data` from the hook is not updated with new values after each individual object is added, which causes a bug.
> 
> First of all, nice work - this works. I just want to toss out an alternative... I'm interested in what you all think about about reverting back to using the data from the hook (I like the term `currentData` so maybe just do `data: currentData`). Then invalidate the cache at the end of the `maybeAddObjectToTable` function:
> 
> ```
> if (bucket) {
>       queryClient.invalidateQueries({
>         queryKey: getObjectBucketObjectsQueryKey(
>           clusterId,
>           bucket.label,
>           prefix
>         ),
>       });
>     }
> ```

Hello @bnussman-akamai 
I tired it out, unfortunately it did't fix the issue. Additionally this approach will re-fetch the data each time new file is added which I believe is something we are trying to avoid with `updateStore` method.

bnussman-akamai:
I agree. We should keep updating the store manually. 

> When several objects are added simultaneously, the component is re-rendered only after all of them have been added. 

I'm a bit surprised this is happening, but the fix makes sense!

bnussman-akamai (approved):
Thanks for making that `getObjectBucketObjectsQueryKey` helper function!

jaalah-akamai (approved):

PR #10938, opened 2024-09-13
fix: [M3-7149] - Fix textfield tooltip icon focus area

hkhalil-akamai:
## Description 📝
Fixes the focus area for tooltips to be a circle rather than an oval as before.

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-09-13 at 5 21 43 PM | Screenshot 2024-09-13 at 5 21 32 PM |

## How to test 🧪
Verify textfield tooltips' focus areas render as a circle and otherwise work as before.

### DIFF
--- a/packages/manager/src/components/TextField.tsx
+++ b/packages/manager/src/components/TextField.tsx
@@ -1,9 +1,6 @@
 import KeyboardArrowDown from '@mui/icons-material/KeyboardArrowDown';
-import { Theme, useTheme } from '@mui/material/styles';
-import {
-  default as _TextField,
-  StandardTextFieldProps,
-} from '@mui/material/TextField';
+import { useTheme } from '@mui/material/styles';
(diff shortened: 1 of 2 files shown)


packages/manager/src/components/TextField.tsx
```diff
@@ -480,8 +479,10 @@ export const TextField = (props: TextFieldProps) => {
         {tooltipText && (
           <TooltipIcon
             sxTooltipIcon={{
+              height: '34px',
               margin: '0px 0px 0px 4px',
-              padding: '6px',
+              padding: '17px',
+              width: '34px',
```
hkhalil-akamai, line 485: Specifying these as numbers instead of '__px' changes the appearance of the tooltip icon

hana-linode (approved):

PR #10963, opened 2024-09-18
change: [M3-8599] - Disable 'Save' button in Change Avatar Color dialog until color is changed

mjac0bs:
## Description 📝
Internal feedback: a user can forget to hit 'Save' once they've made their color selection.

Enabling the primary action button only if the color is changed makes the enabled Save button a more visible call to action for the user, reminding them to click it once they've selected a color.

## Changes  🔄
- Disables the Save button on the `AvatarColorPickerDialog` 

## Target release date 🗓️
9/30/24

## Preview 📷

| Color not changed | Color changed |
|--------|--------|
| Screenshot 2024-09-18 at 7 55 50 AM | Screenshot 2024-09-18 at 7 56 05 AM |

### Prerequisites
- Use an account without a Gravatar or repro the issue in the dev environment, where Gravatars are no more.

### Reproduction steps
- Go to https://cloud.linode.com/profile/display and click the `Change Avatar Color` button.
- Observe that the Save button is enabled.

### Verification steps
(How to verify changes)
- .Go to http://localhost:3000/profile/display and click the `Change Avatar Color` button.
- Observe that the Save button is disabled as long as the preview avatar color in the dialog matches the existing avatar color.
- Use the color picker to select a different color.
- Observe that the Save button is enabled.

*Check all that apply*

- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/src/features/Profile/DisplaySettings/AvatarColorPickerDialog.test.tsx
+++ b/packages/manager/src/features/Profile/DisplaySettings/AvatarColorPickerDialog.test.tsx
@@ -1,4 +1,5 @@
 import { fireEvent } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import * as React from 'react';
 
 import { renderWithTheme } from 'src/utilities/testHelpers';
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/Profile/DisplaySettings/AvatarColorPickerDialog.test.tsx
```diff
@@ -31,13 +35,4 @@ describe('AvatarColorPicker', () => {
     await fireEvent.click(getByText('Close'));
     expect(mockProps.handleClose).toHaveBeenCalled();
   });
-
-  it('closes when Save button is clicked', async () => {
```
mjac0bs, line 35: I wondered if a component test would be better for this component, since it has interactions and handles state.

bnussman-akamai: A unit test could still bring some value here perhaps


```typescript
  it('closes when Save button is clicked', async () => {
    const { getByLabelText, getByText } = renderWithTheme(
      <AvatarColorPickerDialog {...mockProps} />
    );

    const saveButton = getByText('Save').closest('button');

    expect(saveButton).toBeDisabled();

    fireEvent.input(getByLabelText('Avatar color picker'), {
      target: { value: '#333333' },
    });

    // Verify save button becomes enabled after changing the color
    expect(saveButton).toBeEnabled();

    await userEvent.click(saveButton!);

    // Verify dialog closes after saving
    expect(mockProps.handleClose).toHaveBeenCalled();
  });
```

Side note:  I believe `fireEvent` does not need to be awaited whereas `userEvent` should be awaited. I think `userEvent` should be preferred when possible, but fireEvent needs to be used on the color input because of React Testing Library limitations

mjac0bs: Thank you, this makes sense!

I think I've always defaulted to `fireEvent` with little understanding of `userEvent`, but you're right about `userEvent` being preferred based on the [docs](https://testing-library.com/docs/user-event/intro/#differences-from-fireevent) and [this explanation](https://ph-fritsche.github.io/blog/post/why-userevent). I updated the coverage to adjust the Close button test and add back the Save button test in [17a524e](https://github.com/linode/manager/pull/10963/commits/17a524efa03ce47f396aa61557d0144f191a9c15) based on your snippet.

bnussman-akamai (approved):

abailly-akamai (approved):
Thank you! nice improvement!

Confirmed the button state before/after changing the color ✅

coliu-akamai (approved):
✅ confirmed tests pass - thanks for added test coverage! 🚀 
✅ confirmed save button disabled when no new color is picked, enabled when new color is picked

It was really neat when I accidentally clicked out of the color picker, opened it back up, and the new color I'd picked was still chosen (+ the button was enabled as a result) so I could just save without having to find the color again 🚀

PR #10968, opened 2024-09-19
feat: [LKEAPIFW-428] LKE clusters should have IP ACL integration on CM (part 1)

talmai:
## Description 📝
Enables creation of LKE clusters with IP ACLS; enables updates of already migrated clusters

## Changes  🔄
- Adds new UI components
- adds respective library calls
- some style cleanup

*** tests will be in separate PRs: #11131 and #11132 

## Target release date 🗓️
Oct 28th release

## Preview 📷
Cluster creation:

Cluster summary:
| IPACL enabled  | IPACL not enabled   |
| ------- | ------- |
| Screenshot 2024-09-04 at 14 47 05 | Screenshot 2024-10-08 at 18 28 48 |

New IPACL Drawer:
| Cluster migrated (IPACL enabled/not enabled) | Cluster not yet migrated   |
| ------- | ------- |
| Screenshot 2024-09-04 at 14 47 05 | Screenshot 2024-10-08 at 18 28 48 |

## How to test 🧪

Spin up this branch, update to reach alpha's apinext backend. Once all set, attempt to create new clusters with ipacls and then update them.

### Prerequisites
See how to test

### Verification steps
- Create a Cluster with IPACL enabled (toggled on). Confirm button on summary page lists number of IPs added
- Create a Cluster with IPACL disabled (toggled off). Confirm button on summary page says 'Enable'. If you added any IPs, confirm they show up in the drawer
- Create a Cluster that does not have IPACL. Confirm button on summary page says 'Enable' (it will take several seconds to load) and that drawer matches the screenshot for 'Cluster not yet migrated'. Update/toggle IPACL, confirm flow now follows a cluster with IPACL enabled
  - See Talmai's comment for how to create these clusters. I'v also just been creating clusters on alpha (but not on this branch/using the develop branch), and then going back to this branch
  - general spotcheck of Kubernetes with both the APL flag on and off

*Check all that apply*

- ✅ 👀 Doing a self review
- ✅ ❔ Our [contribution guidelines](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md)
- ✅ 🤏 Splitting feature into small PRs
- [NO see comment] ➕ Adding a [changeset](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md#writing-a-changeset)
- [NO ] 🧪 Providing/Improving test coverage
- ✅ 🔐 Removing all sensitive information from the code and PR description
- [NO] 🚩 Using a feature flag to protect the release
- ✅ 👣 Providing comprehensive reproduction steps
- [NO] 📑 Providing or updating our documentation
- [NO] 🕛 Scheduling a pair reviewing session
- [N/A] 📱 Providing mobile support
- [N/A] ♿  Providing accessibility support


jcallahan-akamai:
Thanks for the contribution @talmai. Would you please move this to Draft until it's ready to be reviewed?

talmai:
Can someone help me get the changeset together? I keep hitting the following error:

```
toliveir@bos-mpdj0 manager % yarn changeset
yarn run v1.22.22
$ node scripts/changelog/changeset.mjs

======================================================

Failed to get pull request number.
Please open a pull request for the current branch and let's try this again!

➔ Error: Pull request number not found.

======================================================

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
toliveir@bos-mpdj0 manager % yarn changeset
yarn run v1.22.22
$ node scripts/changelog/changeset.mjs

======================================================

Failed to get pull request number.
Please open a pull request for the current branch and let's try this again!

➔ Error: Pull request number not found.

======================================================

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

coliu-akamai:
Just pushed up a changeset! 

Giving a quick glance over the files - seems like there are a few new components and flows. Will there be test coverage added later/is there a ticket for it already?

packages/manager/src/features/Kubernetes/CreateCluster/ControlPlaneACLPane.tsx
```diff
@@ -0,0 +1,117 @@
...
+export interface ControlPlaneACLProps {
+  enableControlPlaneACL: boolean;
+  setControlPlaneACL: (enabled: boolean) => void;
+  ipV4Addr: ExtendedIP[];
+  handleIPv4Change: (ips: ExtendedIP[]) => void;
+  ipV6Addr: ExtendedIP[];
+  handleIPv6Change: (ips: ExtendedIP[]) => void;
+}
+
+export const IPACLCopy = () => (
```
talmai, line 21: just noticed this is not used (I prob hit some CSS edge case and ended up incorporating directly).

jdamore-linode:
Hi @talmai! I'm seeing a test has failed repeatedly in `lke-create.spec.ts` against this branch. Would you mind taking a quick look? We can disregard if it's unrelated to your PRs changes.

`yarn && yarn build && yarn start:manager:ci`, then in another console:

```bash
yarn cy:run -s "cypress/e2e/core/kubernetes/lke-create.spec.ts"
```

(You might have to follow [our Cypress first time setup docs](https://github.com/linode/manager/blob/develop/docs/development-guide/08-testing.md#first-time-setup), but feel free to reach out on Slack if you have any questions or run into any trouble)

coliu-akamai:
Thanks Talmai! Overall this is looking really solid! 🎉 
Will also reach out to Daniel to set up a UX review

✅ Tested the following flows so far:
- Creating LKE clusters with only IPv4 IPs, only IPv6s, both, and having IP ACL disabled (and then later enabled via drawer)
- Validation/API error handling for nonsense IPs for both the create flow and the drawer
- Adding/updating/removing IPs in the drawer
- Enabling/disabling IP ACL in the drawer

Still need to go through the code, had a few questions/comments with what I tested:
- Should there be error handling for when I enter the same IP twice? This will say two IP addresses in the cluster summary's IPACL section, but the IPs are the same:


- There's some extra spacing at the bottom of the HA Control Plane on the Create flow now (in comparison to the top of the section):


- On prod, because there's no IP ACL section yet, there's an extra divider:


- in the drawer, the API error appears in the IP section but in the create flow, the error appears at the top of the page - wondering if there should be consistency? (and wondering how other pages/drawers handle this - will need to check)

- In the Create flow: if I type a bad IP, then flip off the Enable IPACL toggle, and try to create a cluster, the bad IP will still get sent to the backend and trigger an error - wondering if toggling off 'Enable ACL' should clear the IPs that might still exist there

talmai:
@coliu-akamai :

> * Should there be error handling for when I enter the same IP twice? This will say two IP addresses in the cluster summary's IPACL section, but the IPs are the same:

We can! Would we then block the user until they have distinct IPs? Would that become an error? Or is the expectation to transparently handle this?

> * On prod, because there's no IP ACL section yet, there's an extra divider:

The expectation is that this only goes live if it's live on prod. And that all clusters in the future will be created with it. So this is... expected (?) at the moment, but would not be a realistic thing in the future/GA.


> * in the drawer, the API error appears in the IP section but in the create flow, the error appears at the top of the page - wondering if there should be consistency? (and wondering how other pages/drawers handle this - will need to check)

Please check with Daniel.

> * In the Create flow: if I type a bad IP, then flip off the Enable IPACL toggle, and try to create a cluster, the bad IP will still get sent to the backend and trigger an error - wondering if toggling off 'Enable ACL' should clear the IPs that might still exist there

Yeah, the original concept was to allow the customer to populate IPs regardless of enabling or not. Daniel requested that the UI be a bit more explicit. Can you check with him if this is something he agrees with it?

talmai:
There is one scenario that is now almost impossible to test without manually creating a ipacl-less cluster via apinext. The scenario is when a NON-IPACL lke cluster already exists, but needs to install/migrate to the IPACL code. 

The reason is that, now that we have flipped the latest apinext code, all clusters get created with IPACL. 

If you want to test this, you will need to call the following code:

```
curl  -X POST -H "Content-Type: application/json" -H "Authorization: Bearer YOUR_TOKEN" https://api.dev.linode.com/v4/lke/clusters -d '{"label": "<YOUR_LABEL>", "region": "<PROBABLY_US-EAST>",  "tags": ["<WHATEVER>"], "node_pools": [{"type": "g6-standard-2", "count": 3}], "control_plane": {"high_availability":false}, "k8s_version": "bleeding"}'  | jq
```

coliu-akamai:
@talmai thanks for clarifying! I'll discuss with Daniel regarding where the API errors should go + toggling of the IPACL.

I just pushed up some small design changes after meeting with Daniel earlier today - they change spacing and other small stuff like that.

Regarding error handling for two IPs that are the same - I think that as long as we're blocking a person from creating an LKE due to same IPs, we'd want to display some error/warning message. Not sure if this is something that should be handled frontend only, or if the backend would return an error that the frontend would then display. @bnussman-akamai do you have any thoughts? 

This is what we do for subnets, and I think the API returns an error that we then display:

bnussman-akamai:
@coliu-akamai I agree. I think it's okay to let the API handle IP range overlap error handling and surface that error on the frontend. We can always try to make Cloud Manager have the same validation as the API, but I like keep the frontend validation simple and letting the API handle anything else that may be more complex.

talmai:
> @coliu-akamai I agree. I think it's okay to let the API handle IP range overlap error handling and surface that error on the frontend. We can always try to make Cloud Manager have the same validation as the API, but I like keep the frontend validation simple and letting the API handle anything else that may be more complex.

Just to clarify this: but the api backend does NOT block ACLs with overlapping/same IPs. And neither does CM (I tested). So we can def get back to this with a change in the behavior (would a jira ticket for future work be enough?), but as of now, the frontend wouldn't ever get an error because of this.

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeControlPaneACLDrawer.tsx
```diff
@@ -0,0 +1,372 @@
...
+          ? 'Update IPACL'
+          : 'Install IPACL'
+      );
+      refetchKubernetesACL();
+    }
+  }, [open]);
+
+  const {
+    mutateAsync: updateKubernetesClusterControlPlaneACL,
+  } = useKubernetesControlPlaneACLMutation(clusterId);
```
bnussman-akamai: This would likely require significant changes, but we are trying to encourage the use of `react-hook-form` for managing state in forms throughout Cloud Manager. 

We also are trying to avoid `useEffect` as much as possible because it is often the source of bugs and the desired behavior can often be achieved without useEffect. 

I think `useForm` with the `values` option generally works as a good way to avoid the need for a useEffect

A very simple example of `react-hook-form` can be found in `packages/manager/src/features/Users/UserProfile/UserEmailPanel.tsx` but there are probably numerous other examples in the codebase.

coliu-akamai:
Thanks @talmai! I'm not familiar with the timelines for this project besides that this PR needs to make it to the next release, but I think a future ticket for API work to be done would make sense! And then we'd just be able to surface that new error for the frontend. 

I just pushed up some changes after meeting with Daniel:
- Made button in the IPACL drawer says 'Install IPACL" when Summary pannel says "Install IPACL"
- the IPACL toggle now appears and the button wording is "Update IPACL" in the drawer as long as IPACL is already installed (toggle shows up even if IPACL is installed but not enabled) 
- Moved the error we were talking about in the Create flow to the IPACL panel's section
- in the create flow, if we toggle off IPACL and then click 'Create', any IPs we'd previously added will be ignored and not sent to the backend

There's still one margin/padding issue that is being a bit more finicky than I expected, so looking into that a bit more 😅

coliu-akamai:
Checked with Daniel - the design should be all set now, but we will need some new copies. Hopefully they'll be in by Wednesday/in time for the release branch being cut, otherwise this might need to be pointed to the release branch.

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeClusterControlPlaneACL.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+  const failedMigrationStatus = () => {
+    // when a cluster has not migrated, the query will always fail
+    setControlPlaneACLMigrated(!isErrorKubernetesACL);
+    return isErrorKubernetesACL;
+  };
+
+  const EnabledCopy = () => {
+    return (
+      <Box className={classes.aclElement} onClick={handleOpenDrawer}>
```
coliu-akamai, line 97: Should we be using buttons instead if these have onClick actions?

We could also have a function that determines the copy itself instead of needing different functions for each case

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeSummaryPanel.styles.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+import { Box } from 'src/components/Box';
+import { CopyTooltip } from 'src/components/CopyTooltip/CopyTooltip';
+import { TableCell } from 'src/components/TableCell';
+import { TableRow } from 'src/components/TableRow';
+import { Typography } from 'src/components/Typography';
+
+import type { Theme } from '@mui/material/styles';
+
+// ---------------------------------------------------------------------
```
coliu-akamai, line 15: a lot of these styled components aren't being used - we could remove the unused ones

jaalah-akamai:
@coliu-akamai will handle the rest of the work needed here. @talmai will leave code comments surrounding the parts where we need to ensure we're accounting for certain edge cases when creating clusters.

packages/manager/cypress/e2e/core/cloudpulse/linode-widget-verification.spec.ts
```diff
@@ -46,8 +46,14 @@ import { Interception } from 'cypress/types/net-stubbing';
 const expectedGranularityArray = ['Auto', '1 day', '1 hr', '5 min'];
 const timeDurationToSelect = 'Last 24 Hours';
 
-const { metrics, id, serviceType, dashboardName, region, resource } =
-  widgetDetails.linode;
+const {
```
coliu-akamai, line 49: looks like this file and `factories/dashboards.ts` snuck in as eslint fixes while I was resolving conflicts - gonna leave them in this PR

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeControlPaneACLDrawer.tsx
```diff
@@ -0,0 +1,298 @@
...
+  const enabledExists = enabled !== undefined;
+  const shouldDefaultToEnabled = !clusterMigrated || !enabled;
+
+  // refetchOnMount isnt good enough for this query because
+  // it is already mounted in the rendered Drawer
+  React.useEffect(() => {
+    if (open && !isLoadingKubernetesACL && !isFetchingKubernetesACL) {
+      refetchKubernetesACL(); // makes it fetch again
+    }
+  }, [open]);
```
coliu-akamai, line 63: when I spoke with Daniel earlier this week, iirc he'd mentioned that we want both the IPACL button on the Summary page and this drawer (once we open it) to get/retry getting for the IPACL data four times before stopping if IPACL isn't installed - keeping this useEffect for that for now... Wondering if we need to refetch the data in this drawer though, especially if the query already failed a bunch while on the summary page - will touch base with him again!

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeControlPaneACLDrawer.tsx
```diff
@@ -0,0 +1,298 @@
...
+    setValue,
+    watch,
+  } = useForm({
+    values: {
+      enabled: !!enabled,
+      ipv4: ipv4 ?? [stringToExtendedIP('')],
+      ipv6: ipv6 ?? [stringToExtendedIP('')],
+      'revision-id': revisionID,
+    },
+  });
```
coliu-akamai, line 87: I'd like to get these values to match `KubernetesControlPlaneACLPayload` and do something like `useForm<KubernetesControlPlaneACLPayload>`. However, we're using the `ExtendedIP[]` type for the `MultipleIPInput` components whereas `KubernetesControlPlaneACLPayload`'s IP fields are just `string[]` - it's making it a bit difficult, but will test some more stuff out

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeControlPaneACLDrawer.tsx
```diff
@@ -0,0 +1,298 @@
...
+export const KubeControlPlaneACLDrawer = (props: Props) => {
+  const { closeDrawer, clusterId, clusterLabel, clusterMigrated, open } = props;
+
+  const {
+    data: data,
+    error: isErrorKubernetesACL,
+    isFetching: isFetchingKubernetesACL,
+    isLoading: isLoadingKubernetesACL,
+    refetch: refetchKubernetesACL,
+  } = useKubernetesControlPlaneACLQuery(clusterId);
```
coliu-akamai, line 43: will be making another version to lift this query and the data up into the parent and pass it down to the drawer - brb

coliu-akamai (commented):
@talmai ty for the additional comments! They've been super helpful in providing context.

@jdamore-linode took a look at `lke-create.spec.ts` - I've seen it failing pretty consistently in the github checks, but then it's been passing relatively consistently for me, locally... 😕🤔  it failed for me once when some region was selected (I think Amsterdam?) that led to some checkbox not being checked above the Create button, disabling the Create button - could that have been an issue? 

I'm planning to write e2e tests for the new flows (see my comment on M3-8725), will  create another PR for them since this PR is already about ~900 added lines. Will be putting this up for review shortly - hopefully by tomorrow, just have a couple more things I want to look into

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeControlPaneACLDrawer.tsx
coliu-akamai: spoke with Daniel - removing this portion

packages/manager/src/components/MultipleIPInput/MultipleIPInput.tsx
```diff
@@ -128,20 +131,21 @@ export const MultipleIPInput = React.memo((props: Props) => {
...
-      onClick={addNewInput}
-    >
-      {buttonText ?? 'Add an IP'}
-    </Button>
-  );
+  const addIPButton =
+    forVPCIPv4Ranges || isLinkStyled ? (
+      <StyledLinkButtonBox sx={{ marginTop: isLinkStyled ? '8px' : '12px' }}>
+        <LinkButton onClick={addNewInput}>{buttonText}</LinkButton>
+      </StyledLinkButtonBox>
```
coliu-akamai, line 138: introduced a new prop, `isLinkStyled`, because of the margin differences (and `forVPCIPv4Ranges` does additional things) - this does feel a bit jank/not clean, but not too sure what else to do 🥲...will keep thinking but putting this up for review

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeControlPaneACLDrawer.tsx
bnussman-akamai: I really think we should eliminate `ExtendedIP` and align with `KubernetesControlPlaneACLPayload` if possible. I don't find `ExtendedIP` to be very maintainable because it diverges from the API's types and has `error` on the type, which is bad because `react-hook-form` has its own more correct way to handle errors.

How hard would the change be?

bnussman-akamai: Happy to help look into this if needed. I really don't like "extended" types like this 😖

coliu-akamai: yep the wrapper idea should work!! Just pushed something specific to lke clusters up, will work on a generalized version now/a wrapper around `MultipleIPInput` itself 

Looking at `CreateCluster.tsx`, `ExtendedIP` might be there to stay since that file doesn't use `react-hook-form` 🥲 but will look a bit more

coliu-akamai: Aligned the drawer with `KubernetesControlPlaneACLPayload`! 

Looking into removing `ExtendedIP` from the LKE Create flow - think I'm onto something with directly using `kubernetesControlPlaneACLPayloadSchema` (ie `schema.validate(somePayload)`) and using the path returned:
 
(unless we shouldn't be doing `schema.validate(payload)` @bnussman-akamai? I only saw this in one other place in the codebase, `AddCreditCardForm`, which then used `yupToFormErrors` / returns  `FormikErrors`, so wanted to double check since we're trying to move away from Formik) 

Also need to double check if we're gonna allow empty strings to be passed in and then just ignore them - both the create and drawer flow filtered out empty strings when creating the payload, but previous validation didn't allow empty strings. (If empty strings are accepted, this would make `schema.validate(somePayload)` difficult due to index issues)

packages/manager/src/components/MultipleIPInput/MultipleNonExtendedIPInput.tsx
bnussman-akamai, line 1: I do wonder if it would make sense to just not use `MultipleIPInput` at all for cases like this (in the future. we can leave it for now unless you want to do more work 😅 ). I do think abstractions are good, but I feel like building the same thing without the abstraction is also fine. `MultipeIPInput` just seems a bit rigid and doesn't enable us to be as flexible. For example, I wonder if `react-hook-form`'s `useFieldArray` could be used for the IPs if we weren't locked into using `MultipleIPInput `.

Most of that is me thing thinking out-loud, but what you did here seems fair to me.

bnussman-akamai (commented):
Looking good! Let me know when you need a final review!

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeSummaryPanel.tsx
```diff
@@ -200,19 +168,56 @@ export const KubeSummaryPanel = React.memo((props: Props) => {
...
                   tags={cluster.tags}
                   updateTags={handleUpdateTags}
                   view="inline"
                 />
-              </Grid>
+              </StyledTagGrid>
             </Grid>
           </Grid>
         }
+        footer={
```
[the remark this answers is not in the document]
coliu-akamai: yup! created a new footer component in [0188f60](https://github.com/linode/manager/pull/10968/commits/0188f60aaf59ed679997361085b9341e64b30661) after I saw the updated footer mocks

packages/manager/src/components/MultipleIPInput/MultipleNonExtendedIPInput.tsx
coliu-akamai: 😅 I'm gonna leave as is for now since I still need to finish the e2e tests, but will hopefully revisit this down the line! Maybe this could be part of the work for transitioning `ClusterCreate` to `react-hook-form`, since that also uses `MultipleIPInput`

packages/manager/src/queries/kubernetes.ts
```diff
@@ -349,6 +359,36 @@ export const useAllKubernetesClustersQuery = (enabled = false) => {
...
 
+export const useKubernetesControlPlaneACLQuery = (
+  clusterId: number,
+  enabled: boolean = true
+) => {
+  return useQuery<KubernetesControlPlaneACLPayload, APIError[]>({
+    enabled,
+    ...kubernetesQueries.cluster(clusterId)._ctx.acl,
+  });
+};
```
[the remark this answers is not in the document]
coliu-akamai: from what I know for non-migrated clusters, the x4 requerying is on purpose - I heard that the API needs to fail four times for the non-migrated scenario... we talked about shortening it to 2-3 checks, but it's an acceptable pain for now?

packages/manager/src/queries/kubernetes.ts
```diff
@@ -139,6 +146,12 @@ export const useKubernetesClusterMutation = (id: number) => {
         queryClient.invalidateQueries({
           queryKey: kubernetesQueries.lists.queryKey,
         });
+        queryClient.invalidateQueries({
+          queryKey: kubernetesQueries.cluster(id).queryKey,
+        });
```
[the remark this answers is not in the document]
bnussman-akamai: We need to keep the `kubernetesQueries.lists.queryKey` invalidation to keep the paginated list up to date.

We shouldn't need to made this invalidation though. 

The issue is that we setQueryData on `kubernetesQueries.cluster(id).queryKey`, but we don't        on `kubernetesQueries.cluster(id)._ctx.beta.queryKey`

I don't agree with the descision to use a different query key for the v4beta fetch. Ideally we should remove this query query and make `kubernetesQueries.cluster(id).queryKey` work with v4 and v4beta.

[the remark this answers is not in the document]
bnussman-akamai: I think this might be what we want:

```typescript
export const useKubernetesClusterQuery = (id: number) => {
  const showAPL = useGetAPLAvailability();
  return useQuery<KubernetesCluster, APIError[]>({
    ...kubernetesQueries.cluster(id),
    queryFn: showAPL
      ? () => getKubernetesClusterBeta(id)
      : () => getKubernetesCluster(id),
  });
};
```

and delete the `useKubernetesClusterBetaQuery` hook and delete the `beta` query key

Hopefully this works because the way it is currently implemented makes cache management much harder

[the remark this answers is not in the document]
coliu-akamai: thanks for the explanation! Let me try that + remove the invalidation I'd previously added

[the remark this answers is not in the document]
bnussman-akamai: Nice. Looks good to me!

packages/manager/src/queries/kubernetes.ts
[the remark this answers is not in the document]
bnussman-akamai: > heard that the API needs to fail four times for the non-migrated scenario

I'm a little surprised by this because I feel like it would be very confusing / complicated for customers that use the API directly and/or DX tools.

Either way, I don't think this is a blocker, just something we should keep in mind as something that could be fixed/improved

bnussman-akamai (approved):

packages/manager/src/queries/kubernetes.ts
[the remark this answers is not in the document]
abailly-akamai: > the x4 requerying is on purpose - I heard that the API needs to fail four times for the non-migrated scenario

this.is.weird.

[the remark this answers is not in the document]
talmai: Let me clarify this a bit..

The IPACL functionality is not enabled (by default) on any of the existing LKE clusters. That means that every single customer - as is - needs to enable/install it. The `../<clusterid>/control_plane_acl` endpoint returns a 400 if the cluster was not created or updated to enable the ACL feature, this request returns a 400 (Bad Request) error with an appropriate message, such as Cluster does not support Control Plane ACL. 

See docs here: https://techdocs.akamai.com/linode-api/reference/get-lke-cluster-acl

The react component - as is - fails multiple times before we actually receive the error (at the UI level). See relevant code here: `features/Kubernetes/KubernetesClusterDetail/KubeClusterControlPlaneACL.tsx`.

packages/validation/src/firewalls.schema.ts
```diff
@@ -7,7 +7,8 @@ export const IP_ERROR_MESSAGE =
   'Must be a valid IPv4 or IPv6 address or range.';
 
 export const validateIP = (ipAddress?: string | null): boolean => {
-  if (!ipAddress) {
+  // ''is falsy, so we must specify that it is OK
+  if (ipAddress !== '' && !ipAddress) {
```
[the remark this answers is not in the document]
coliu-akamai: not that I'd noticed, just wanted a sanity check 😅

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubeControlPaneACLDrawer.tsx
```diff
@@ -0,0 +1,294 @@
...
+        title={clusterLabel}
+      >
+        <form onSubmit={handleSubmit(updateCluster)}>
+          {errors.root?.message && (
+            <Notice spacingTop={8} variant="error">
+              {errors.root.message}
+            </Notice>
+          )}
+          <Stack sx={{ marginTop: 3 }}>
+            <Typography sx={{ width: '90%' }} variant="body1">
```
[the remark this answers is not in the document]
coliu-akamai: Without the 90% the typography extends until the divider's width - tried adding it to a wrapper, but that made the dividers 90% width too 😅 

ended up making a styled component to avoid using the sx everywhere

packages/validation/src/firewalls.schema.ts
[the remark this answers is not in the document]
coliu-akamai: unfortunately doesn't seem like it - Boolean('') evaluates to false like Boolean(undefined/null) 🥲

packages/manager/src/components/MultipleIPInput/MultipleIPInput.tsx
coliu-akamai: Hm - having an `sx` prop but not `isLinkStyled` would get us the secondary Button (lines 140-148) which UX wasn't the biggest fan of (agreed), unless we passed in link styles as sx I think? but then at that point might as well go to the first case 

🤔ing more

coliu-akamai:
@abailly-akamai 
- double checked regarding the drawer - we're wanting to allow people to update IPs and tag it with a revision ID independent of enabling ACL
- created internal ticket for LKE summary mobile/smaller views! [M3-8764]

abailly-akamai (approved):
This is looking great on my end - I did not find functionality issues after the second pass, tho I may still miss some context on the feature as a whole. Thanks for taking this to the finish line @coliu-akamai

packages/manager/src/features/Kubernetes/KubernetesClusterDetail/KubernetesClusterDetail.tsx
```diff
@@ -30,12 +29,8 @@ export const KubernetesClusterDetail = () => {
   const id = Number(clusterID);
   const location = useLocation();
   const showAPL = useGetAPLAvailability();
-  const kubernetesClusterBetaQuery = useKubernetesClusterBetaQuery(id);
```
j-zimnowoda, line 33: why this was removed ?

coliu-akamai: we consolidated the logic for this query with the already existing `kubernetesClusterQuery` so that we'd only need one query key. The functionality should still exist, it's just been moved. Apologies if this caused any confusion!

see [here](https://github.com/linode/manager/pull/10968#discussion_r1805027163) for reasoning/discussion for the change, and [here](https://github.com/linode/manager/pull/10968/files#diff-9fe16b4e86dc963b779071391448b45668e789b1bdf3a38d3d1e97448d8e9940) for the file changes (lines 112-118)

PR #10976, opened 2024-09-19
fix: [M3-8602] - Fix `plan-selection.spec.ts` test failures on retries

jdamore-linode:
## Description 📝

This fixes a minor test flaw in the `plan-selection.spec.ts` spec where mocks are getting set up in `before` hooks instead of `beforeEach` hooks. This prevents mocks from being applied on test retries (because mocks get cleared between tests, but `before` hooks only get called once), causing retries in this spec to be guaranteed to fail.

## Changes  🔄
- Replace `before` hooks with `beforeEach` in `plan-selection.spec.ts` to ensure mocks get applied on test retries

## Target release date 🗓️
N/A

### Reproduction steps
In `develop` (edit: the issue causing the initial failure has been fixed, so reproducing this is now pretty difficult), build Cloud via `yarn && yarn build && yarn start:manager:ci` and then:

```bash
repeat 5 yarn cy:run -s "cypress/e2e/core/linodes/plan-selection.spec.ts
```

Observe flakiness (in my case, the spec failed 4 out of 5 times).

### Verification steps
Check out this branch, build Cloud via `yarn && yarn build && yarn start:manager:ci` and then:

```bash
repeat 5 yarn cy:run -s "cypress/e2e/core/linodes/plan-selection.spec.ts
```

Observe the test passes each time. I was able to run the test 50 times on repeat without failures.

We can also rely on CI to some extent -- this issue theoretically impacts all of our Autocompletes, so any test failures involving an Autocomplete (including region selects, etc.) should receive extra scrutiny.

### Manual verification
Some extra attention should be given to the "Phone Verification" field at `/profile/auth` since it's the only Autocomplete I'm aware of that takes advantage of the recent changes to the `PopperComponent` prop, so it may be at more risk of being broken by these changes than other Autocompletes.

*Check all that apply*

- [x] 👀 Doing a self review
  - Pushed a changeset for the test fix since that's been a flaw all along, but don't think a changeset is necessary for the `Autocomplete` change since 10780 hasn't been released yet. Thoughts?


packages/manager/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -108,11 +109,17 @@ export const Autocomplete = <
...
   const optionsWithSelectAll = [selectAllOption, ...options] as T[];
 
+  /* Memoize popper callback to prevent unnecessary popper re-rendering. */
+  const customPopper = useCallback(
+    (props: PopperProps) => {
+      return <CustomPopper {...props} sx={sxPopperComponent} />;
+    },
+    [sxPopperComponent]
+  );
+
```
jdamore-linode, line 119: This was a suggestion I saw on a StackOverflow thread, but hoping that somebody with more React expertise than me could take a look at this and double check that this is a sensible solution.

It seems to make the tests very happy, and I didn't notice any regressions during my manual testing, but I'd be lying if I said I fully understand the implications of wrapping this in `useCallback`.

packages/manager/cypress/e2e/core/linodes/plan-selection.spec.ts
```diff
@@ -124,7 +124,7 @@ const notices = {
 
 authenticate();
 describe('displays linode plans panel based on availability', () => {
-  before(() => {
+  beforeEach(() => {
```
jdamore-linode, line 127: Using `before` here instead of `beforeEach` causes the mocks not to get applied on re-attempts which leads to failures.

The initial failure is caused by the `cy.click()` issue, but `plan-selection.spec.ts` failures are overrepresented in CI due to the re-attempts being guaranteed to fail.

AzureLatte (approved):

packages/manager/src/components/Autocomplete/Autocomplete.tsx
bnussman-akamai: Hey Joe! Take a look at https://github.com/linode/manager/pull/10978. Do you think this could be a better way to approach it and would it fix the unexpected re-render / arrow function issue?

jdamore-linode: Thank you @bnussman-akamai!

Commented on your PR, but I think your solution is hands down better: it fixes the test failures and I'm guessing it's a lot more idiomatic / in line with the way we normally do things. Happy to either integrate your implementation into this PR or let you take the reins with #10978!

carrillo-erik: @jdamore-linode Do you happen to have the link to the StackOverflow thread? I'm curious why `useCallback` was used instead of `useMemo` and would be interested in reading more into it. Anyway, @bnussman-akamai provided a very good alternative and I see it's been merged.

jdamore-linode: @carrillo-erik not on hand, no -- I could try to find it, but I was already a bit doubtful about `useCallback` as a solution for this, and I think Bank's solution demonstrates that it wasn't really necessary in the first place, so I don't think I'd consider the answer in that thread as authoritative in this case

carrillo-erik (approved):
Approved based on our conversation with the understanding that some minor changes will be implemented.

bnussman-akamai (approved):

coliu-akamai (approved):
thanks Joe! 
✅ confirmed tests passed repeatedly

PR #10977, opened 2024-09-19
poc: MUI v6

bnussman-akamai:
## Description 📝

- Using this POC I did a few weeks ago to see if MUI v6 fixes the Safari autofill issues. My theory is that it will be fixed in MUI v6 because of the removal of internal `useIsFocusVisible`


bnussman-akamai:
Okay.... so preview links don't build when there are conflicts 😑

PR #10780, opened 2024-08-13
refactor: [M3-6908] - Replace react-select in Profile

carrillo-erik:
## Description 📝
This PR helps remove part of the react-select dependency for accessibility reasons and to consolidate our usage of third-party libraries. In particular, it replaces the `EnhancedSelect` with our more versatile MUI-based `Autocomplete` component within the `Profile` feature.

## Changes  🔄
- Introduce a new `sxPopperComponent` prop (optional) for the `Autocomplete` which allows custom CSS styling for the `PopperComponent`.
- The `PhoneVerification` depended heavily on the `react-select` instance, and by removing it, the previous styling required to be redone. This was the reason for introducing the new `sxPopperComponent` prop.
- Update `TimezoneForm` to `TimeZoneForm`.
- Update `default` exports with `named` exports and corresponding `lazy()` imports.

## Target release date 🗓️
09/03/2024

### Verification steps
(How to verify changes)
- Verify that the `Profile` feature works as expected.
- Inspect the `Profile` pages for any visual regressions.

## As an Author I have considered 🤔
*Check all that apply*
- [x] 👀 Doing a self review
- [ ] ❔ Our [contribution guidelines](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md)
- [ ] 🤏 Splitting feature into small PRs
- [x] ➕ Adding a [changeset](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md#writing-a-changeset)
- [x] 🧪 Providing/Improving test coverage
- [ ] 🔐 Removing all sensitive information from the code and PR description
- [ ] 🚩 Using a feature flag to protect the release
- [ ] 👣 Providing comprehensive reproduction steps
- [ ] 📑 Providing or updating our documentation
- [ ] 🕛 Scheduling a pair reviewing session
- [ ] 📱 Providing mobile support
- [ ] ♿  Providing accessibility support

PR #10978, opened 2024-09-20
fix: [M3-8617] - Autocomplete Popper re-rendering causing test failures

bnussman-akamai:
## Description 📝

- Should hopefully fix the test failures we've recently begun seeing throughout the test suite when Cypress tries to interact with Autocompletes:

  > cy.click() failed because the page updated while this command was executing. Cypress tried to locate elements based on this query:
  > <div.MuiBox-root.css-b5zaes>
We initially found matching element(s), but while waiting for them to become actionable, they disappeared from the page.

- Provides an alternative to @jdamore-linode 's fix https://github.com/linode/manager/pull/10976/commits/a90aba8f714b739b4e085261aef44c652047c294 that was originally in  #10976 
- This PR reverts some changes made in https://github.com/linode/manager/pull/10780  in favor of using MUI's built in `slotProps` to customize components that make up our AutoComplete

## How to test 🧪

- Verify the Phone Number select on http://localhost:3000/profile/auth does not break and that the styles look correct
- Verify Autocompletes throughout the app work as expected

### DIFF
--- a/packages/manager/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/manager/src/components/Autocomplete/Autocomplete.tsx
@@ -15,7 +15,6 @@ import {
 } from './Autocomplete.styles';
 
 import type { AutocompleteProps } from '@mui/material/Autocomplete';
-import type { SxProps } from '@mui/system';
 import type { TextFieldProps } from 'src/components/TextField';
 
 export interface EnhancedAutocompleteProps<
@@ -42,10 +41,6 @@ export interface EnhancedAutocompleteProps<
   placeholder?: string;
   /** Label for the "select all" option. */
   selectAllLabel?: string;
-  /**
-   * The prop that allows defining CSS style overrides for the PopperComponent.
-   */
-  sxPopperComponent?: SxProps;
   textFieldProps?: Partial<TextFieldProps>;
 }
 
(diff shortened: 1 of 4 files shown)


jdamore-linode (approved):
Awesome, thanks @bnussman-akamai! 🙏

I'm really confident in this implementation: `plan-selection.spec.ts` passes 50/50 times, tests in `core/linodes` are passing, and I'm not seeing any regressions in any of the Autocompletes I've looked at in the app.

If you're open to it, we can take this out of draft and make this the PR for M3-8617, and I can update #10976 to contain only the M3-8602 fixes (the `before` -> `beforeEach` change in `plan-selection.spec.ts`). Or we can include the spec changes here and just close #10976 altogether

I really appreciate you taking a closer look and proposing this solution! Thanks again Banks!

bnussman-akamai:
> we can take this out of draft and make this the PR for M3-8617, and I can update https://github.com/linode/manager/pull/10976 to contain only the M3-8602 fixes

This sounds good to me! I'll update this PR to reflect M3-8617 @jdamore-linode

dwiley-akamai (approved):
Phone Number select & Autocomplete fields in Cloud look good ✅
Code review ✅

carrillo-erik:
@bnussman Very nice! I seriously didn't understand the power of the `slotProps` and this helps see their usage. This refactor threw me for a spin. Since we are customizing the `Autocomplete` so much. I had started to look into other libraries for a phone number input component with a country flag. However, it felt redundant to introduce a new dependency when the point of the tix was to remove the `react-select` dependency.

PR #10982, opened 2024-09-20
change: [M3-8390] - Initialize Pendo on Cloud Manager

mjac0bs:
## Description 📝
This PR adds Pendo to Cloud Manager by adding the Pendo agent script from the CDN and initializing Pendo with some custom configurations.

## Changes  🔄
- Installs `js-sha256`
- Creates the `usePendo` hook
   - Incorporates Pendo's install script code
   - Loads the Pendo Agent from the CDN with `loadScript`
   - Initializes Pendo with some custom configuration
- Invokes the hook in `App.tsx` with the rest of the GlobalListeners

## Target release date 🗓️
10/14/24

### Prerequisites
- To your `.env`, add the Pendo API Key. This is not a secret key; it will be visible in the browser dev tools, but we also want to be able to stop Pendo from running using an environment variable, which will live elsewhere with our env var configurations. See the internal repo (PR 167) for that change and the API key.
```
# Pendo:
REACT_APP_PENDO_API_KEY='KEY-GOES-HERE'
```

### Verification steps
(How to verify changes)
- We can validate that Pendo is installed by opening the browser console and typing `pendo.validateEnvironment()`. This should output some data, including your hashed visitor and account id.
- We also want to confirm that no sensitive data is sent to Pendo. To do this, filter by `pendo` in the Network tab of the browser dev tools.
   - Then try navigating around Cloud Manager and confirm the following by looking at the `normalizedUrl` of the requests:
      - No ids are being sent (e.g. when viewing a linode details page, `linodes/123` should be truncated to `linodes/`)
      - No search queries are being sent
      - No username or other account data is being sent (e.g. when visiting the User Profile and User Permissions pages, the url doesn't include the username)
      - No token data is being sent

Example:
image

*Check all that apply*

- [x] 👀 Doing a self review


packages/manager/src/hooks/usePendo.ts
```diff
@@ -0,0 +1,120 @@
...
+        pendo[method] =
+          pendo[method] ||
+          function () {
+            pendo._q[method === methodNames[0] ? 'unshift' : 'push'](
+              // eslint-disable-next-line prefer-rest-params
+              [method].concat([].slice.call(arguments, 0))
+            );
+          };
+      })(methodNames[index]);
+    });
```
mjac0bs, line 60: All this stuff is a slightly more readable version of the Pendo install script: https://support.pendo.io/hc/en-us/articles/21362607464987-Components-of-the-install-script.

packages/manager/src/hooks/usePendo.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    // Load Pendo script into the head HTML tag, then initialize Pendo with metadata
+    loadScript(PENDO_URL, {
+      location: 'head',
+    }).then(() => {
+      window.pendo.initialize({
```
mjac0bs, line 66: I left the commented-out sections for reference.

packages/manager/package.json
```diff
@@ -45,6 +45,7 @@
     "highlight.js": "~10.4.1",
     "immer": "^9.0.6",
     "ipaddr.js": "^1.9.1",
+    "js-sha256": "^0.11.0",
```
mjac0bs, line 48: Same hashing method as other Akamai apps use with Pendo.

packages/manager/src/hooks/usePendo.ts
```diff
@@ -0,0 +1,120 @@
...
+
+import { PENDO_API_KEY } from 'src/constants';
+import { useAccount } from 'src/queries/account/account.js';
+import { useProfile } from 'src/queries/profile/profile';
+
+import { loadScript } from './useScript';
+
+declare global {
+  interface Window {
+    pendo: any;
```
mjac0bs, line 12: Unfortunately, [Pendo docs](https://support.pendo.io/hc/en-us/articles/360031862272-Install-Pendo-on-a-single-page-web-application) recommend typing this as `any`.

packages/manager/package.json
bnussman-akamai: I wonder if we can use the native crypto APIs for doing a SHA256 hash 🤔 

https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest

mjac0bs: Interesting! Good find; I didn't know this existed. 

Looks like we can, it's just not nice to look at because we're converting between different data types. If the following looks readable enough, we could skip that `js-sha256` package - though will note it's widely used, no dependencies. We need to use the same hashing algorithm as other Akamai projects (the eventual goal is a unique identifier across apps), but it shouldn't matter how we go about it.

```
 // in hashUniquePendoId():
  [...]

  // Convert to data to uint8 format to hash.
  const dataArray = new TextEncoder().encode(id + pendoEnv);
  const hashBuffer = await crypto.subtle.digest('SHA-256', dataArray);

  // Convert from array buffer back to uint8 array, then to string.
  return Array.from(new Uint8Array(hashBuffer))
    .map((byte) => byte.toString(16).padStart(2, '0'))
    .join('');
};

/**
 * Initializes our Pendo analytics script on mount.
 */
export const usePendo = () => {
  const { data: account } = useAccount();

  let accountId: string | undefined;
  let visitorId: string | undefined;

  hashUniquePendoId(account?.euuid)?.then((hash) => {
    accountId = hash;
  });
  hashUniquePendoId(profile?.uid.toString())?.then((hash) => {
    visitorId = hash;
  });

 [...]
```

packages/manager/src/hooks/usePendo.ts
```diff
@@ -0,0 +1,132 @@
...
+                const userPathMatchingRegex = /(users\/).*/;
+                if (idMatchingRegex.test(url)) {
+                  // Removes everything after the last /
+                  return url.replace(/\/[^\/]*$/, '/');
+                } else if (userPathMatchingRegex.test(url)) {
+                  // Removes everything after /users
+                  return url.replace(userPathMatchingRegex, '$1');
+                }
+                return url;
+              },
```
bnussman-akamai, line 116: Is a URL like this handled?

```
https://cloud.linode.com/oauth/callback#access_token=xxxxxxxxxxxxxx&token_type=bearer&expires_in=7200&scope=*&state=dad80c38-cce6-4426-a027-6a43c83255e7&return=https://cloud.linode.com/oauth/callback?returnTo=/
```

bnussman-akamai (approved):
Script initialization looks good. Didn't 👁️  any issues

packages/manager/src/hooks/usePendo.ts
mjac0bs: Updated in [d861b62](https://github.com/linode/manager/pull/10982/commits/d861b6241bee8ac264531cff27d00455027b4bf7) to remove everything after `oauth/` if the URL matches on `/oauth/callback` to be certain we handle it.

packages/manager/src/hooks/usePendo.ts
coliu-akamai: would we need to cite this in the code?

coliu-akamai (approved):
✅ confirmed pendo.validateEnvironment()
✅ confirmed no sensitive data is sent to Pendo

thanks Mariah! 🎉

packages/manager/src/hooks/usePendo.ts
mjac0bs: I don't think we *need* to as far as Pendo is concerned, but it's good context to have, since we set this up in a way that is consistent with our repo, deconstructing a bit to make use of our loadScript function. I added a link to the Pendo doc in a comment in the code so it's not just ephemerally in this PR - good call out.

mjac0bs:
Going to go ahead and merge this because CI was passing last week and today's updates shouldn't have impacted that; CI is just having known issues today.

PR #10991, opened 2024-09-23
change: [M3-8415] - Hide SMTP warning for Linodes and accounts that have SMTP enabled

hkhalil-akamai:
## Description 📝
Leverages the new `"SMTP Enabled"` account & Linode capabilities to hide the SMTP warning when it doesn't apply.

The SMTP warning appears in two places:

- At the bottom of the Linode create flow. This warning is now hidden for accounts with the new `"SMTP Enabled"` account capability.
- In the Linode details page as a dismissible banner. This warning is now hidden when the `"SMTP Enabled"` Linode capability is present on the selected Linode.

### Prerequisites
Use the alpha API environment.

You will need an account with at least one Linode.

In Admin: remove the `SMTP_ENABLED` customer tag from the account or ensure that it's not present. For the test Linode, check that the *Net Filter Exceptions* field appears as below or change it to match:

Screenshot 2024-09-23 at 4 56 53 PM

### Testing steps
1. Verify the SMTP banner appears in both locations. If it's not appearing for the Linode, you may have dismissed it. Reset your preferences using the Preference Editor: [http://localhost:3000/profile/settings?preferenceEditor=true](http://localhost:3000/profile/settings?preferenceEditor=true).
2. Enable SMTP for a particular Linode by removing the SMTP Net Filter Exceptions.
3. Verify the SMTP warning banner no longer appears for that particular Linode. However, it should still appear for other Linodes and at the bottom of the Create Linode flow
4. Add the `SMTP_ENABLED` customer tag to your account
5. Verify the SMTP warning no longer appears in the Linode Create flow

*Check all that apply*

- [x] 👀 Doing a self review


packages/manager/src/constants.ts
```diff
@@ -233,11 +233,6 @@ export const PAYMENT_HARD_MAX = 50_000;
 
 export const DB_ROOT_USERNAME = 'linroot';
 
-// "In an effort to fight spam, Linode restricts outbound connections on ports 25, 465, and 587 on all Linodes for new accounts created after November 5th, 2019."
-// https://www.linode.com/docs/email/best-practices/running-a-mail-server/
-export const MAGIC_DATE_THAT_EMAIL_RESTRICTIONS_WERE_IMPLEMENTED =
-  '2022-11-30T00:00:00.000Z'; // Date of release for Manager v1.81.0.
-
```
hkhalil-akamai, line 240: Since we are directly using the SMTP status of accounts/Linodes, we no longer need this magic date.

packages/manager/src/features/Linodes/SMTPRestrictionText.tsx
```diff
@@ -3,27 +3,32 @@ import * as React from 'react';
...
     label: string;
   };
 }
 
 export const SMTPRestrictionText = (props: SMTPRestrictionTextProps) => {
-  const { supportLink } = props;
+  const { linode, supportLink } = props;
   const { data: account } = useAccount();
 
+  const displayRestrictionText =
```
mjac0bs, line 24: More of a backend question: Do we know if it's possible for an **account** to have SMTP_ENABLED, but an **individual Linode** is still disabled (with net exceptions on its ports)? If that state were possible, we're currently showing the notice for the Linode, which seems like what we'd want for transparency... unless, somehow, the individual net exceptions don't always get removed when `SMTP_ENABLED` is added account-wide, which could be confusing for customers.

mjac0bs (commented):
Thanks for taking this on, Hussain! Exciting to move away from the magic constant and have backend support for this.

This looks good in dev. From what I can tell based on the API PRs merge dates, this has *not* yet been release to prod yet. Can we update the testing instructions to specify testing in the dev environment?

packages/manager/src/constants.ts
bnussman-akamai: This is great! 🚀

packages/manager/src/features/Linodes/SMTPRestrictionText.tsx
hkhalil-akamai: > Do we know if it's possible for an account to have SMTP_ENABLED, but an individual Linode is still disabled

Yes, it's possible.

I was surprised to learn that there is no logic in the API that links the customer tag with the Net Filter Exceptions of a particular Linode. Adding the tag is a manual process used by support representatives when a customer requests SMTP abilities on their Linode and it is only used to keep track of which customers have had this request granted.

The availability of SMTP for a Linode is determined solely by the presence (or lack thereof) of the SMTP Net Filter Exceptions. There is also a field on the customer called 'Default Net Filter Exceptions' that is applied by default to new Linodes.

So the choice to only use the capability from the Linode object was intentional, since that is the source of truth of whether a particular Linode is SMTP-capable.

For a detailed summary of this topic, I suggest the KB article *SMTP Port Restrictions*.

mjac0bs: > Adding the tag is a manual process used by support representatives when a customer requests SMTP abilities on their Linode and it is only used to keep track of which customers have had this request granted.

Super helpful context and surprises me also. 

> So the choice to only use the capability from the Linode object was intentional, since that is the source of truth of whether a particular Linode is SMTP-capable.

Got it; agree with this.

mjac0bs (approved):
Once you merge, can you update Diana and Avi in the internal ticket, just to let them know when we can expect this to be released? I have a feeling they'll be interested. 

Thank you! 🙌🏼

bnussman-akamai (approved):

PR #11005, opened 2024-09-25
chore: [M3-8524] - Attempt to get dependencies in a more healthy state

bnussman-akamai:
## Description 📝

- This PR does `rm yarn.lock && yarn` to re-generate our lockfile 📄 
  - This will result in dependencies with `~version` “Approximately equivalent to version” and `^version` “Compatible with version” updating 📦 
  - This should help us out on the dependabot front. It will also allow us to remove many of our resolutions 🗑️ 
- Resolves the following Dependabot alerts 🤖 
  - https://github.com/linode/manager/security/dependabot/126
  - https://github.com/linode/manager/security/dependabot/128
  - https://github.com/linode/manager/security/dependabot/116
  - https://github.com/linode/manager/security/dependabot/119
  - https://github.com/linode/manager/security/dependabot/118
  - https://github.com/linode/manager/security/dependabot/117
  - https://github.com/linode/manager/security/dependabot/108
  - Might resolve some of the https://github.com/linode/manager/security/dependabot/113 related alerts
- Removes resolutions for the following (they are no longer needed)
  - `braces`
  - `@babel/traverse`
  - `minimist`
  - `yargs-parser`
  - `kind-of`
  - `ua-parser-js`
  - `immer`
  - `glob-parent`
  - `hosted-git-info` (upgrading `concurrently` allowed this one)
  - `word-wrap`
  - `tough-cookie`
  - `jackspeak` (we added this in https://github.com/linode/manager/pull/10169 as a hacky fix. Now it is no longer needed 🎉 )
- Removes unused dependencies
  - `react-test-renderer` 
  - `reselect-tools`
  - `eslint-plugin-ramda` (removed from api-v4 and validation's `package.json` because they don't use ramda)
  - `@types/jsdom` (We **do** depend on `jsdom` but we **don't** need the types for it because our TypeScript code never imports it)
  - `eslint-plugin-node` (this plugin is not referenced in our eslint config)
  - `eslint-plugin-storybook` (this plugin is not referenced in our eslint config)

### Side Effects 
- I had to fix some issues that TypeScript caught
  - This happened because our `yarn.lock` was re-generated, causing our our TypeScript version to upgrade to the latest minor version. The latest version of TypeScript got smarter and caught a few things!
- Had to do something hacky in our Vite configs. I think it is due to Yarn v1's hoisting behavior because the same issue does not happen with pnpm. Shouldn't be a big deal / blocker at all, we'll just want to clean it up later
- `ts-node` (used by our cypress pipeline) stopped working so I just replaced `ts-node` with a better alternative called [`tsx`](https://www.npmjs.com/package/tsx)
- A UI bug with the Autocomplete was introduced and fixed

### Prerequisites 
- Checkout this PR
- Run `yarn` for good measure
   - Verify yarn.lock does not change when you do this

### Things to check
- Verify everything works like normal locally (Vite dev server, Storybook, etc...) 💻 
- Verify Github Actions pass ✅ 
- Verify Baker jobs pass ✅ 
- Look at the resolutions that are removed 👀 
  - Run `yarn why [package name here]` and verify the version reported by yarn is not vulnerable (verify the version is equal to or newer than what the resolution was) 📦 
- Look at the dependabot alerts and verify this PR installs versions that will resolve the alerts
  - Use `yarn why [package name here]` and verify our repo uses  version that is patched and not vulnerable (the depenedabot page will tell you what versions are acceptable)

- [ ] 👀 Doing a self review


packages/manager/src/features/Linodes/LinodesDetail/LinodeResize/LinodeResizeUnifiedMigrationPanel.tsx
```diff
@@ -112,7 +112,7 @@ export const UnifiedMigrationPanel = (props: Props) => {
               }
               status="help"
               tooltipPosition="right"
-              width={[theme.breakpoints.up('sm')] ? 450 : 300}
```
bnussman-akamai, line 115: We had a handful of things like this.  We would get a type error saying `[theme.breakpoints.up('sm')]` is always truthy so I just updated the code to just use `450`

packages/manager/src/foundations/themes/light.ts
```diff
@@ -336,7 +336,6 @@ export const lightTheme: ThemeOptions = {
           svg: {
             color: Color.Neutrals[40],
           },
-          top: 'unset',
```
bnussman-akamai, line 339: This fixes the UI bug shown below caused by updating our deps.

Screenshot 2024-09-25 at 6 18 11 PM

abailly-akamai (approved):
This is a thing of beauty - thx for tackling this important update

things do appear to be at parity and it's great to be able to trust our tests.

one thing i did notice is new console warnings (PrimaryNav simingly)
Screenshot 2024-09-26 at 12 44 58

I did not notice any others so far but i'll keep looking

packages/manager/package.json
```diff
@@ -208,7 +208,7 @@
     "simple-git": "^3.19.0",
     "storybook": "^8.3.0",
     "storybook-dark-mode": "^4.0.1",
-    "ts-node": "^10.9.2",
+    "tsx": "^4.19.1",
```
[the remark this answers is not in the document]
jdamore-linode: I know we've also talked about using Bun for these scripts too -- iirc it's only a couple of the scripts in the `scripts` dir that we use for the Jenkins pipeline that even rely on `ts-node`, but lately I've been opting to keep that type of stuff in internal repos. I'd be up to move these out of Cloud altogether if it means we can drop the `ts-node`/`tsx` dependency altogether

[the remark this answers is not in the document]
bnussman-akamai: I like the idea of moving scripts like this in internal repos. In the mean time (and because we don't have Bun readily available in CI) I'm totally cool with keeping `tsx` around. It seems very minimal and useful for us even for things like `scripts/package-versions/index.js` (if we were to make it .ts)

Eventually, I'd love Bun to be our go-to tool for running this stuff

dwiley-akamai (commented):
CM, dev server, and all checks look good ✅

I'm seeing this when I attempt to run Storybook:

<details>
<summary>Screenshot</summary>

Screenshot 2024-09-26 at 6 05 16 PM
</details>

packages/manager/package.json
```diff
@@ -201,14 +198,12 @@
...
     "prettier": "~2.2.1",
-    "react-test-renderer": "16.14.0",
     "redux-mock-store": "^1.5.3",
-    "reselect-tools": "^0.0.7",
     "serve": "^14.0.1",
     "simple-git": "^3.19.0",
     "storybook": "^8.3.0",
-    "storybook-dark-mode": "^4.0.1",
-    "ts-node": "^10.9.2",
+    "storybook-dark-mode": "4.0.1",
```
bnussman-akamai, line 205: I pinned this to `4.0.1` because of https://github.com/hipstersmoothie/storybook-dark-mode/issues/282  

This will fix the issue @dwiley-akamai caught with Storybook. We'll just need to keep an eye out for `storybook-dark-mode` to provide a fix before we upgrade storybook-dark-mode next

dwiley-akamai (approved):
Storybook looks good now as well 🚀

PR #11007, opened 2024-09-25
fix: [M3-8588] - Fix `value` selection issue in `ImageSelect`

carrillo-erik:
## Description 📝
This PR fixes the bug in the `ImageSelect` component which resulted in the selected `value` getting cleared after the input lost focus.

## Changes  🔄
- Add logic to handle the case when `value` is `undefined` initially.
- Remove leftover `default export` declarations.
- Replace single letter variables with more meaningful names.

## Target release date 🗓️
10/16/2024

## Preview 📷
| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/user-attachments/assets/e4dd976d-0b7c-470c-a178-d08dcd14c3b3" /> | <video src="https://github.com/user-attachments/assets/74380f5d-715a-425f-8063-14f682093daf" /> |

### Prerequisites
- Choose a Linode from your list of Linodes to work with and power it down.
- Visit the `Storage` tab for your chosen Linode.
- Resize any of the `Disks` in order to enable the `Add A Disk` option.
- Click on the `Add A Disk` button to open the `Create Disk` drawer.
- Click on the `Create from Image` radio option.

### Verification steps
(How to verify changes)
- Select an Image from the `Image` dropdown menu.
- Interact with any other field in the drawer (ex. `Label`, `Root Password`, `Size`).
- Verify that the value selected in the `Image` dropdown component remains after you have focused out of the component and focused into another field in the drawer.
- Verify that the `Add A Disk` feature works as expected.

*Check all that apply*

- [x] 👀 Doing a self review


packages/validation/src/linodes.schema.ts
```diff
@@ -518,6 +518,6 @@ export const UpdateLinodeDiskSchema = object({
 
 export const CreateLinodeDiskFromImageSchema = CreateLinodeDiskSchema.clone().shape(
   {
-    image: string().required('An image is required.'),
+    image: string().required('An image is required.').nullable(),
```
carrillo-erik, line 521: This change is done to remove the validation error message after a user has selected a value from the `ImageSelect` and then clears the value (input).

Screenshot 2024-09-26 at 9 13 25 AM

packages/manager/src/features/Images/ImageSelect.tsx
bnussman-akamai, line 1: I wonder if we could use `ImageSelectv2` here instead. This ImageSelect has some serious type safety issues and hackyness

carrillo-erik: @bnussman-akamai I agree with your concerns about the type safety issues and the hackyness feeling. I gave `ImageSelectv2` a try and it appears that with some minor changes it should work within the `CreateDiskDrawer`.

The one issue I noticed is the way the options are grouped. In the current `ImageSelect`, one the three categories is the recommended 64-bit distributions. In the `ImageSelectv2`, they are grouped by vendor. Would require some conditional logic to determine with grouping function to use.

**ImageSelect**
ImageSelect
**ImageSelectv2**
ImageSelectv2

bnussman-akamai: Ah good point. That would require some UX / product consideration

packages/validation/src/linodes.schema.ts
bnussman-akamai: Yup's .typeError might be more fitting here than `nullable`

bnussman-akamai (approved):

PR #11022, opened 2024-10-01
add: [M3-8630] - Add default root hostname for TXT records

jaalah-akamai:
## Description 📝
For A/AAA records we default to the root hostname if left empty or by using `@`, but we do not do the same for TXT record hostnames. Due to this discrepancy, there has been confusion to whether a TXT record for their root domain was applying correctly, or whether the A/AAAA records they were creating would apply to something like `example.com.example.com`.

## Changes  🔄
- TXT has been added as a case in the `shouldResolve`  function
- TXT hostname now defaults to root domain hostname

## Target release date 🗓️
N/A

## Preview 📷
https://github.com/user-attachments/assets/8383f622-ed04-474f-b61b-c2e303c2d17b

### Reproduction steps
- Create domain in cloud manager
- Add an A record for that domain leaving the hostname field blank or using @
- Create a TXT record for that domain leaving the hostname field blank or using @
- Notice the difference in "hostname" column after those records have been saved.
   - The "hostname" column for A/AAAA/MX records reflects the root domain, while the hostname column for TXT records remains blank.

### Verification steps
- Checkout branch or view preview link
- Observe TXT hostname populated accordingly.

*Check all that apply*

- [x] 👀 Doing a self review

### DIFF
--- a/packages/manager/.changeset/pr-11022-added-1727879935382.md
+++ b/packages/manager/.changeset/pr-11022-added-1727879935382.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Added
+---
+
+Add default root hostname for TXT records ([#11022](https://github.com/linode/manager/pull/11022))
--- a/packages/manager/src/features/Domains/DomainRecordDrawer.test.tsx
+++ b/packages/manager/src/features/Domains/DomainRecordDrawer.test.tsx
@@ -25,6 +25,14 @@ describe('Domain record helper methods', () => {
       expect(shouldResolve('AAAA', 'target')).toBe(false);
     });
 
+    it('should return true for the name of an TXT record', () => {
+      expect(shouldResolve('TXT', 'name')).toBe(true);
+    });
+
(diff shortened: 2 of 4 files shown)


hana-linode (approved):
Approved pending a changeset is added

hkhalil-akamai (approved):
Verified that '@' or leaving Hostname field empty results in a TXT entry displaying the root domain as Hostname.

hkhalil-akamai:
While we're here, I noted a couple other inconsistencies with the other record types (NS, MX and SRV):

| Record Type | Inconsistency |
|--------------|---------------|
|NS|Entering "@" in *Subdomain* field causes validation error|
|MX|Entering "@" in *Subdomain* field causes validation error|
|A/AAAA|None|
|CNAME|None|
|TXT|None|
|SRV|Leaving *Target* field blank causes validation error|

I might be misunderstanding how these record types are supposed to work and this might be the expected behavior. Otherwise, should we create a ticket to address these other discrepancies?

PR #11028, opened 2024-10-01
test: [M3-8114] - Clean up Linodes, LKE clusters, and Firewalls after test run

jdamore-linode:
## Description 📝
This adds a quick-n-dirty clean up pass that will run after all the Cypress tests run in order to clean up Linodes, LKE Clusters, and Firewalls created by the tests.

This should resolve most cases where resources get left behind after a job finishes, but it's not a totally foolproof solution. Resources may not be cleaned up completely if Cypress crashes mid-run (or if the CI job itself is cancelled), or if any API errors occur during the course of clean up.

Once we've had this solution in place for a while, we can evaluate whether any additional measures are necessary.

## Changes  🔄
- Adds an `after` hook that does Linode, LKE Cluster, and Firewall clean up after a run

## Target release date 🗓️
N/A

## How to test / Verification steps 🧪
We can rely on CI by letting the job run, signing into the test accounts, and observing whether there are any leftover Linodes, LKE clusters, or Firewalls left behind with the `cy-test-` prefix.

Alternatively, we can test the changes locally by manually creating some resources prefixed with `cy-test-`, running a couple tests via `yarn cy:run -s "path/to/spec/or/specs/to/run.spec.ts"` or `yarn cy:debug`, and observing whether or not the resources get cleaned up.

To be extra thorough, you can add some `throw new Error('Fail')`s into the tests you run to confirm that this clean up happens even when there are test failures.

*Check all that apply*

- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/.changeset/pr-11028-tests-1727799211177.md
+++ b/packages/manager/.changeset/pr-11028-tests-1727799211177.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Tests
+---
+
+Clean up Linodes, LKE clusters, and Firewalls created by Cypress tests when runs finish ([#11028](https://github.com/linode/manager/pull/11028))
--- a/packages/manager/cypress/support/e2e.ts
+++ b/packages/manager/cypress/support/e2e.ts
@@ -63,8 +63,10 @@ import { deleteInternalHeader } from './setup/delete-internal-header';
 import { mockFeatureFlagClientstream } from './setup/feature-flag-clientstream';
 import { mockAccountRequest } from './setup/mock-account-request';
 import { trackApiRequests } from './setup/request-tracking';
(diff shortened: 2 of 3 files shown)


jdamore-linode:
There appear to be issues with this:

- It runs after every spec instead of once after the entire run
- Mocks may be interfering with the clean up requests

Will find another solution, this won't work.

carrillo-erik:
> Will find another solution, this won't work.
 Hey Joe, will you be updating this PR once you identify a different solution or should this PR be closed?

jdamore-linode:
> Hey Joe, will you be updating this PR once you identify a different solution or should this PR be closed?

Probably a good call @carrillo-erik, it may be a couple days before I get this working. I'll reopen this once I do. Thanks!

PR #10970, opened 2024-09-19
fix: [M3-8578] - Fix console error on Volume Create - Linode Config selection

cpathipa:
## Description 📝
The console Warning: Cannot update a component (VolumeCreate) while rendering a different component (Unknown), usually occurs when a state update is triggered during the rendering phase of another component. In the `ConfigSelect` component, this issue likely arises from the logic that automatically sets a new value. So, moving that logic to `useEffect` hook avoids triggering state updates during the render phase.

## Target release date 🗓️
9/30

### Reproduction steps
(How to reproduce the issue, if applicable)
- Navigate to Volume Create Flow, Select linode and notice warning in the console.

### Verification steps
(How to verify changes)
- Checkout the branch and run in local
-  Navigate to Volume Create Flow, select linode and observe no console warning.

## As an Author I have considered 🤔

*Check all that apply*

- [ ] 👀 Doing a self review
- [ ] ❔ Our [contribution guidelines](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md)
- [ ] 🤏 Splitting feature into small PRs
- [ ] ➕ Adding a [changeset](https://github.com/linode/manager/blob/develop/docs/CONTRIBUTING.md#writing-a-changeset)
- [ ] 🧪 Providing/Improving test coverage
- [ ] 🔐 Removing all sensitive information from the code and PR description
- [ ] 🚩 Using a feature flag to protect the release
- [ ] 👣 Providing comprehensive reproduction steps
- [ ] 📑 Providing or updating our documentation
- [ ] 🕛 Scheduling a pair reviewing session
- [ ] 📱 Providing mobile support
- [ ] ♿  Providing accessibility support

PR #11032, opened 2024-10-01
fix: Add Volume button on Cloud Manager Linode page causing browser to freeze

coliu-akamai:
## Description 📝
Reverts the changes from [this PR](https://github.com/linode/manager/pull/10970) to remove the useEffect so that the page doesn't crash. Note that this means there will be a console error again - we'll have to look into how to handle that issue differently.

## Changes  🔄
- Reverts change to remove useEffect

## Target release date 🗓️
10/2

## reproduction steps
- On the Linode Storage tab, when clicking 'Add Volume', notice how the page freezes afterwards

## Verification steps
- confirm the page freeze no longer happens

*Check all that apply*

- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/CHANGELOG.md
+++ b/packages/manager/CHANGELOG.md
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 ### Fixed:
 
 - DBaaS Landing Page V2 platform error for New Beta Users ([#11024](https://github.com/linode/manager/pull/11024))
+- Add Volume button on Linodes Storage tab causing page crash ([#11032](https://github.com/linode/manager/pull/11032))
 
 ## [2024-09-30] - v1.129.0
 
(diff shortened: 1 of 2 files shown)


abailly-akamai (approved):
Thank you

cpathipa (approved):

bnussman-akamai (approved):
Looks good! Thank you.

I know calling `setState` on render is causing us console errors, but the React docs seem to recommend it... I wonder if those docs only apply to certain circumstances or the console warning can just safely be ignored

Screenshot 2024-10-02 at 10 07 34 AM

bnussman-akamai:
I may have answered my own question after reading this:

> React only lets you update the same component’s state during a render. If you update another component’s state during a render, you’ll see an error. A condition like items !== prevItems is necessary to avoid loops. You may adjust state like this, but any other side effects (like changing the DOM or setting timeouts) should stay in event handlers or Effects to [keep components pure.](https://react.dev/learn/keeping-components-pure)

coliu-akamai:
Merging this to staging as all checks have passed! @bnussman-akamai you might need to resolve a small changelog conflict afterwards

PR #11034, opened 2024-10-01
fix: [M3-8685] - Users unable to update timezone

bnussman-akamai:
## Description 📝

- Fixes bug preventing users from updating / setting their timezone 🌎 🕐 
- I think this might be a regression from https://github.com/linode/manager/pull/10780

## Preview 📷

| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/user-attachments/assets/f5c3392c-138b-4f02-aa4b-b255a9e9bd58" /> | <video src="https://github.com/user-attachments/assets/24a84f42-9c5e-4899-b8eb-a722de6f00af" /> |

## How to test 🧪

- Checkout this PR or use the preview link
- Verify you can update your timezone on http://localhost:3000/profile/display

- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/src/features/Profile/DisplaySettings/TimezoneForm.tsx
+++ b/packages/manager/src/features/Profile/DisplaySettings/TimezoneForm.tsx
@@ -3,30 +3,20 @@ import { DateTime } from 'luxon';
 import { useSnackbar } from 'notistack';
 import * as React from 'react';
 
-import timezones from 'src/assets/timezones/timezones';
-import { ActionsPanel } from 'src/components/ActionsPanel/ActionsPanel';
+import { timezones } from 'src/assets/timezones/timezones';
 import { Autocomplete } from 'src/components/Autocomplete/Autocomplete';
+import { Button } from 'src/components/Button/Button';
 import { CircleProgress } from 'src/components/CircleProgress';
-import { Stack } from 'src/components/Stack';
-import { Typography } from 'src/components/Typography';
+import { Notice } from 'src/components/Notice/Notice';
 import { useMutateProfile, useProfile } from 'src/queries/profile/profile';
 
 interface Props {
   loggedInAsCustomer: boolean;
 }
 
-interface Timezone {
-  label: string;
-  name: string;
-  offset: number;
-}
+type Timezone = typeof timezones[number];
 
-export interface TimezoneOption<T = string, L = string> {
-  label: L;
-  value: T;
-}
-
-export const formatOffset = ({ label, offset }: Timezone) => {
+export const getOptionLabel = ({ label, offset }: Timezone) => {
   const minutes = (Math.abs(offset) % 60).toLocaleString(undefined, {
     minimumIntegerDigits: 2,
     useGrouping: false,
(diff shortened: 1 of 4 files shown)


packages/manager/src/features/Profile/DisplaySettings/TimezoneForm.tsx
```diff
@@ -3,160 +3,111 @@ import { DateTime } from 'luxon';
...
+  const [timezoneValue, setTimezoneValue] = React.useState(profile?.timezone);
 
   const onSubmit = () => {
-    if (timezoneValue === '') {
-      return;
+    if (!timezoneValue) {
+      enqueueSnackbar('Please select a valid timezone.', { variant: 'error' });
     }
 
-    updateProfile({ timezone: String(timezoneValue) }).then(() => {
```
bnussman-akamai, line 71: The is the cause of the bug. `timezoneValue` is an object and we "cast" it to a string, causing us to send `{"timezone": "[Object object]"}`  to the API

packages/manager/src/features/Profile/DisplaySettings/TimezoneForm.tsx
```diff
@@ -3,160 +3,111 @@ import { DateTime } from 'luxon';
...
-const renderTimezonesList = (): TimezoneOption<string>[] => {
+const getTimezoneOptions = () => {
   return timezones
-    .map((tz) => ({ ...tz, offset: DateTime.now().setZone(tz.name).offset }))
-    .sort((a, b) => a.offset - b.offset)
-    .map((tz: Timezone) => {
-      const label = formatOffset(tz);
-      return { label, value: tz.name };
-    });
+    .map((tz) => {
```
bnussman-akamai, line 32: I was able to consolidate the two `map`s into one

packages/manager/src/features/Profile/DisplaySettings/TimezoneForm.tsx
```diff
@@ -3,160 +3,111 @@ import { DateTime } from 'luxon';
...
+export const getOptionLabel = ({ label, offset }: Timezone) => {
   const minutes = (Math.abs(offset) % 60).toLocaleString(undefined, {
     minimumIntegerDigits: 2,
     useGrouping: false,
   });
   const hours = Math.floor(Math.abs(offset) / 60);
   const isPositive = Math.abs(offset) === offset ? '+' : '-';
 
-  return `\(GMT ${isPositive}${hours}:${minutes}\) ${label}`;
+  return `(GMT ${isPositive}${hours}:${minutes}) ${label}`;
```
bnussman-akamai, line 27: I don't think the `\`s are necessary. I removed them

packages/manager/src/features/Profile/DisplaySettings/TimezoneForm.tsx
```diff
@@ -3,160 +3,111 @@ import { DateTime } from 'luxon';
...
-interface Timezone {
-  label: string;
-  name: string;
-  offset: number;
-}
-
-export interface TimezoneOption<T = string, L = string> {
-  label: L;
-  value: T;
-}
```
bnussman-akamai, line 27: We don't need these extra types. We can lean on TypeScript's inference to get a solution that is even more typesafe!

packages/manager/src/features/Profile/DisplaySettings/TimezoneForm.tsx
```diff
@@ -3,160 +3,111 @@ import { DateTime } from 'luxon';
...
+  const disabled = !timezoneValue || profile?.timezone === timezoneValue;
 
   if (!profile) {
     return <CircleProgress />;
   }
 
   return (
     <>
-      {loggedInAsCustomer ? (
-        <StyledLoggedInAsCustomerNotice data-testid="admin-notice">
```
bnussman-akamai, line 90: Removed this one-off component in favor of a simple `Notice` just to clean things up a bit

packages/manager/src/features/Profile/DisplaySettings/TimezoneForm.tsx
```diff
@@ -3,160 +3,111 @@ import { DateTime } from 'luxon';
...
-          <Typography variant="h2">
-            While you are logged in as a customer, all times, dates, and graphs
-            will be displayed in your browser&rsquo;s timezone ({timezone}).
```
bnussman-akamai, line 93: This copy was just factually incorrect. 

 "all times, dates, and graphs will be displayed in your browser's timezone" has been corrected to say "all times, dates, and graphs will be displayed in the user's timezone "

packages/manager/src/features/Profile/DisplaySettings/TimezoneForm.tsx
```diff
@@ -3,160 +3,111 @@ import { DateTime } from 'luxon';
...
-        <Stack>
-          <Autocomplete
-            slotProps={{
-              popper: {
-                sx: {
-                  maxHeight: '285px',
-                  overflow: 'hidden',
-                },
-              },
-            }}
```
bnussman-akamai, line 107: I was able to clean up a lot of these one-off styles by using the `fullWidth` prop instead

packages/manager/src/features/Profile/DisplaySettings/TimezoneForm.tsx
```diff
@@ -3,160 +3,111 @@ import { DateTime } from 'luxon';
...
-            disableClearable
-            errorText={error?.[0].reason}
-            label="Timezone"
-            onChange={(_, option) => handleTimezoneChange(option)}
-            options={timezoneList}
-            placeholder="Choose a Timezone"
-            sx={{ width: '416px' }}
-          />
-        </Stack>
-        <ActionsPanel
```
bnussman-akamai, line 123: Replaced the `ActionsPanel` with a `Button`. I think a plain Button is more fitting here.

packages/manager/src/features/Profile/DisplaySettings/TimezoneForm.tsx
```diff
@@ -3,160 +3,111 @@ import { DateTime } from 'luxon';
...
+          Update Timezone
+        </Button>
+      </TimezoneFormContainer>
     </>
   );
 };
 
-const StyledRootContainer = styled('div', {
-  label: 'StyledRootContainer',
+const TimezoneFormContainer = styled('div', {
```
abailly-akamai, line 102: It's a nit, but I think we should continue to name Styled Components with "Styled", it makes it so much easier for developers to quickly identify components that are purely meant to apply styles and and hove other function.

Doesn't need to be changed for this PR but something to consider.

abailly-akamai (dismissed):
Thanks!

Confirmed the bug ad the resolution ✅ 
Other changes make sense to me and do not add any regression on my end

abailly-akamai:
@bnussman-akamai the e2e failure appears to be relevant here

Screenshot 2024-10-02 at 09 25 10

packages/manager/src/features/Profile/DisplaySettings/TimezoneForm.tsx
bnussman-akamai: Ahh got it. I always though that prefixing the component with "Styled" was kind of an anti-pattern but I see what you mean

cpathipa (dismissed):
Confirming on the bug timezone is successfully updating.

bnussman-akamai:
Good catch @abailly-akamai 

Should be fixed by d7ce808

I removed the `autoHighlight` prop on the Autocomplete in attempt to clean things up, which caused it to behave differently (caused it to not automatically select the first option). I re-added that, which should fix the test.

coliu-akamai (dismissed):
✅ confirmed can now update my timezone
✅ confirmed previously failing e2e passing locally now

thanks for the fix + cleanup! 🧹

coliu-akamai (approved):

jdamore-linode (approved):
Thanks for the catch and fix, @bnussman-akamai!

abailly-akamai (approved):
Thanks! Approving pending e2e passing

hana-linode (approved):

PR #11038, opened 2024-10-02
upcoming: [M3-8613] - Restrict Image Upload to regions with Object Storage

bnussman-akamai:
## Description 📝

- Adds a feature flag that allows us to only show regions with Object Storage capability on the Image Upload page 🌎 
  - Image Service Gen2 uses Object Storage as a storage backend, so that's why we use that capability
  - We using a flag because we will coordinate with API and let customers know before we make this change 🎏 
- This is part of the Image Service Gen2 effort 💾 

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-10-02 at 11 35 44 AM | Screenshot 2024-10-02 at 11 36 06 AM |

## How to test 🧪

- Use Launch Darkly to toggle the flag
- Verify regions get filtered appropriately 
  - **Flag Off:**  Should show all core regions
  - **Flag On:** Should only  show regions with Object Storage

- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/src/featureFlags.ts
+++ b/packages/manager/src/featureFlags.ts
@@ -95,6 +95,7 @@ export interface Flags {
   databases: boolean;
   dbaasV2: BetaFeatureFlag;
   disableLargestGbPlans: boolean;
+  disallowImageUploadToNonObjRegions: boolean;
   gecko2: GeckoFeatureFlag;
   gpuv2: gpuV2;
   imageServiceGen2: boolean;
--- a/packages/manager/.changeset/pr-11038-upcoming-features-1727969322447.md
+++ b/packages/manager/.changeset/pr-11038-upcoming-features-1727969322447.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Upcoming Features
+---
+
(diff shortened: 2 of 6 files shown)


packages/manager/src/featureFlags.ts
```diff
@@ -95,6 +95,7 @@ export interface Flags {
   databases: boolean;
   dbaasV2: BetaFeatureFlag;
   disableLargestGbPlans: boolean;
+  disallowImageUploadToNonObjRegions: boolean;
```
bnussman-akamai, line 98: I'm using a dedicated flag for this because I've been having a hard time getting an exact timeline for when this needs to be live. 

It aligns with the API's feature flag

hana-linode:
Should we display, but disable regions without Object Storage capability instead similar to Linode Create?

image

bnussman-akamai:
I'm thinking no because products like VPC for example just filter the list rather than disable options but I will ask UX! @hana-linode

bnussman-akamai:
UX confirmed they want to filter out incompatible regions for the reason constancy with other parts of the app and to reduce cognitive load so the user can make a choice faster.  @hana-linode

hana-linode (approved):

cpathipa (approved):

PR #11042, opened 2024-10-03
feat: [M3-6989] - Improve helper text accessibility

carrillo-erik:
## Description 📝
Highlight the Pull Request's context and intentions.

## Changes  🔄
- Changes the HTML element that is rendered by the `Typography` component from `<p>` to `<span>`. This is done using the `component` prop.
- Sets an `id` attribute in the `Typography` component to associate the helper text to the `<input />` element.

## Target release date 🗓️
10/14/2024

## Preview 📷
| Before  | After   |
| ------- | ------- |
| Screenshot 2024-10-03 at 9 44 46 AM | Screenshot 2024-10-03 at 9 41 41 AM |
| Screenshot 2024-10-03 at 9 46 03 AM | Screenshot 2024-10-03 at 9 41 56 AM |

### Prerequisites
- Visit the `linodes/create` page.
- Turn on Apple's VoiceOver utility.
- Make sure your volume is up.

### Reproduction steps
- Traverse the form fields using tab navigation on your keyboard.
- Pay close attention to the output read by the VoiceOver utility when the following form fields have been focused.
  - `Region`
  - `Linode Plan`
  - `Disk Encryption`
  - `VPC`
  - `Firewall` 
  - `Backups`
  - `Private IP`

### Verification steps
(How to verify changes)
- Verify that upon focusing on the interactive fields, the helper text is read by the VoiceOver utility. 
- Verify that the form field labels and values are also read by the VoiceOver utility.
- Verify that there are no visual regressions.

*Check all that apply*

- [ ] 👀 Doing a self review


packages/manager/src/components/Encryption/Encryption.tsx
```diff
@@ -79,6 +82,7 @@ export const Encryption = (props: EncryptionProps) => {
         flexDirection="row"
       >
         <Checkbox
+          aria-labelledby="diskEncryptionDesc"
           checked={disabled ? false : isEncryptEntityChecked} // in Create flows, this will be defaulted to be checked. Otherwise, we will rely on the current encryption status for the initial value
           data-testid={checkboxTestId}
           disabled={disabled}
```
hkhalil-akamai, line 88: Noticing that when this checkbox is disabled, it can't be focused using tab (and thus screenreaders won't read the associated helper text). Should we consider adding the `focusableWhenDisabled` prop to all Checkboxes with helper text?

packages/manager/src/components/TabbedPanel/TabbedPanel.tsx
```diff
@@ -118,8 +118,9 @@ const TabbedPanel = React.memo((props: TabbedPanelProps) => {
           <TabPanels>
             {tabs.map((tab, idx) => (
               <TabPanel
-                key={`tabs-panel-${tab.title}-${idx}`}
+                aria-labelledby="planInformationDesc"
```
hkhalil-akamai, line 121: This is a shared component so we can't have anything specific to `PlansPanel` here. Perhaps we can add `labelledby` to the `Tab` interface so we can specify it in `PlansPanel`?

packages/manager/src/features/Linodes/LinodeCreatev2/Addons/PrivateIP.tsx
hkhalil-akamai, line 1: This change prevents the screen reader from reading the headings ("Backups" and "Private IP") and instead it goes straight to the helper text. Before, it would read the heading and then the helper text. Is this the intended behavior?

packages/manager/src/features/Linodes/LinodeCreatev2/TwoStepRegion.tsx
hkhalil-akamai, line 1: Can we add a `labelledby` to one or both of the selects in this component? Currently the helper text is not picked up by VoiceOver.

Screenshot 2024-10-04 at 4 30 53 PM

hkhalil-akamai (commented):
Since the ticket refers to all helper texts across the app, should we create a subtask to capture that this PR is scoped to the Linode Create page only?

hkhalil-akamai:
Merge conflicts caused by renaming in #11043

packages/manager/src/components/Encryption/Encryption.tsx
carrillo-erik: The `focusableWhenDisabled` appears to be a `@mui/base` prop. I don't recall if we've built something similar to accomplish this already. There are a few workarounds I can think of but each would require changes to other shared components. With the current implementation, when the checkbox is disabled, the tooltip gains focuses and VoiceOver reads the text.

packages/manager/src/components/TabbedPanel/TabbedPanel.tsx
carrillo-erik: We could also pass in a `panelId` variable like this:` aria-labelledby=${panelId}-InformationDesc`.

packages/manager/src/features/Linodes/LinodeCreate/Firewall.tsx
```diff
@@ -69,7 +69,7 @@ export const Firewall = () => {
     <Paper>
       <Stack spacing={2}>
         <Typography variant="h2">Firewall</Typography>
-        <Typography>
+        <Typography component="span" id="firewallDesc" variant="body1">
```
abailly-akamai, line 72: I am not sure I understand. This is not helper text. This is perfectly accessible context offered on top of the field.

packages/manager/src/features/Linodes/LinodeCreate/VLAN/VLAN.tsx
```diff
@@ -68,7 +68,7 @@ export const VLAN = () => {
       sx={{ margin: '0 !important', padding: 1 }}
     >
       <VLANAvailabilityNotice />
-      <Typography variant="body1">
+      <Typography component="span" variant="body1">
```
abailly-akamai, line 71: same here

PR #11047, opened 2024-10-04
upcoming: [M3-8452, M3-8603] - Add Resource Links & Fix Test Flake

jaalah-akamai:
## Description 📝
Update OBJ Gen2 resource links and fix issue causing test flake.

## Changes  🔄
- `.includes` was causing test flake because the substring could match several options. IE: If looking for `us-10`, it may return `us-1` or even `us-100`. See `M3-8603`
- Added documentation resource links

## Target release date 🗓️
N/A

## Screenshots
Screenshot 2024-10-04 at 11 53 27 AM

### Verification steps
Ensure tests pass and links work!

*Check all that apply*

- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.tsx
@@ -84,7 +84,11 @@ export const BucketRateLimitTable = ({
         ) : (
           'This endpoint type supports up to 750 Requests Per Second (RPS). '
         )}
-        Understand <Link to="#">bucket rate limits</Link>.
+        Understand{' '}
+        <Link to="https://techdocs.akamai.com/cloud-computing/docs/create-and-manage-buckets">
+          bucket rate limits
+        </Link>
+        .
       </Typography>
 
       {isGen2EndpointType && (
--- a/packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
@@ -128,7 +128,7 @@ export const CreateBucketDrawer = (props: Props) => {
   };
 
   const clusterRegion = watchCluster
-    ? regions?.find((region) => watchCluster.includes(region.id))
+    ? regions?.find((region) => watchCluster === region.id)
     : undefined;
 
   const { showGDPRCheckbox } = getGDPRDetails({
--- a/packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
+++ b/packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
@@ -128,7 +128,7 @@ export const CreateBucketDrawer = (props: Props) => {
   };
 
   const clusterRegion = watchCluster
-    ? regions?.find((region) => watchCluster === region.id)
+    ? regions?.find((region) => watchCluster.includes(region.id))
     : undefined;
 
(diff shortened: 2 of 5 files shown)


packages/manager/src/features/ObjectStorage/BucketLanding/BucketRateLimitTable.tsx
```diff
@@ -84,7 +84,11 @@ export const BucketRateLimitTable = ({
         ) : (
           'This endpoint type supports up to 750 Requests Per Second (RPS). '
         )}
-        Understand <Link to="#">bucket rate limits</Link>.
+        Understand{' '}
+        <Link to="https://techdocs.akamai.com/cloud-computing/docs/create-and-manage-buckets">
```
mjac0bs, line 88: ❓ Do we know if this docs page is going to be updated to mention rate limits? I didn't see anything there about bucket rate limits. One place rate limits are mentioned is: https://techdocs.akamai.com/cloud-computing/docs/object-storage#optimize-applications-to-avoid-rate-limiting.

mjac0bs (commented):
Can we add two changesets? One to note the fixed tests, and another in upcoming features for the docs links.

Confirmed the intended tests pass consistently locally and the fix makes sense. But, seeing a consistent failure with `enable-object-storage.spec.ts`, which is looking for the pricing helper text in the create drawer but it does not display. This also seems to be the case in the UI locally if the OMC feature flag is toggled off. (The test mocks the feature flag as false.) Not sure what's happening there, as the `OveragePricing` component is present in both OMC and non-OMC create drawers. Has this been failing recently and is captured in another ticket? Seems unrelated to this PR.

Also confirmed the docs links redirect to real pages.

packages/manager/src/features/ObjectStorage/BucketLanding/CreateBucketDrawer.tsx
```diff
@@ -128,7 +128,7 @@ export const CreateBucketDrawer = (props: Props) => {
   };
 
   const clusterRegion = watchCluster
-    ? regions?.find((region) => watchCluster.includes(region.id))
+    ? regions?.find((region) => watchCluster === region.id)
```
jdamore-linode, line 131: Ahhh -- changing this line seems to have broken the region select in the non-Multicluster drawer (and is causing the failure in `enable-object-storage.spec.ts`). This `includes` makes more sense for the non-Multicluster drawer since it's comparing a region ID (`us-cgk`) against a cluster ID (`us-cgk-1`), so the substring match kind of makes sense in that context.

mjac0bs: Ohhhh, good find!!

packages/manager/src/features/ObjectStorage/BucketLanding/OMC_CreateBucketDrawer.tsx
```diff
@@ -179,7 +179,7 @@ export const OMC_CreateBucketDrawer = (props: Props) => {
   };
 
   const selectedRegion = watchRegion
-    ? regions?.find((region) => watchRegion.includes(region.id))
+    ? regions?.find((region) => watchRegion === region.id)
```
jdamore-linode, line 182: Nice! Confirming that this fixes the test flake with `bucket-create-gen2.spec.ts` when selecting the mock `us-10` region.

bnussman-akamai (approved):

mjac0bs (approved):
👍🏼  Local e2e run looks good for the OBJ tests. Other CI failures seem to be due to the known auth problem.

Screenshot 2024-10-07 at 7 38 37 AM

PR #11056, opened 2024-10-07
fix: [M3-8726] - Users unable to upgrade Kubernetes version from landing page

bnussman-akamai:
## Description 📝

- Fixes a bug 🐛 preventing users from upgrading their Kubernetes cluster from the Kubernetes landing page ⬆️ 
- I believe the bug was caused by https://github.com/linode/manager/pull/10756 
  - Specifically switching `isLoading` ➡️ `isFetching` - see comment for explanation
- Adds Cypress test that would have caught this bug 🧪 
  - (the test uses mocked API data)
## Preview 📷

| Before  | After   |
| ------- | ------- |
| <video src="https://github.com/user-attachments/assets/87004463-9c6d-4176-847d-d001e454f79c" /> | <video src="https://github.com/user-attachments/assets/eea9b226-7002-4a12-b4d0-074d858beebf" /> |

### Reproduction steps
- Spin up a Kubernetes cluster that is **not** on the latest version
- Attempt the upgrade flow **on the Kubernetes landing page**
- Observe that when clicking "Upgrade Version", the dialog will not proceed to step 2

### Verification steps
- Spin up a Kubernetes cluster that is **not** on the latest version
- Test the upgrade flow **on the Kubernetes landing page**
- Verify you see step 1 and step 2

- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/cypress/e2e/core/kubernetes/lke-update.spec.ts
+++ b/packages/manager/cypress/e2e/core/kubernetes/lke-update.spec.ts
@@ -20,6 +20,7 @@ import {
   mockRecycleAllNodes,
   mockGetDashboardUrl,
   mockGetApiEndpoints,
+  mockGetClusters,
 } from 'support/intercepts/lke';
 import {
   mockGetLinodeType,
@@ -113,7 +114,7 @@ describe('LKE cluster updates', () => {
    * - Confirms that Kubernetes upgrade prompt is shown when not up-to-date.
    * - Confirms that Kubernetes upgrade prompt is hidden when up-to-date.
    */
-  it('can upgrade Kubernetes engine version', () => {
+  it('can upgrade kubernetes version from the details page', () => {
     const oldVersion = '1.25';
     const newVersion = '1.26';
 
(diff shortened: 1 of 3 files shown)


packages/manager/src/features/Kubernetes/KubernetesLanding/KubernetesLanding.tsx
```diff
@@ -98,7 +98,7 @@ export const KubernetesLanding = () => {
 
   const isRestricted = profile?.restricted ?? false;
 
-  const { data, error, isFetching } = useKubernetesClustersQuery(
```
bnussman-akamai, line 101: ### For context
`isLoading` is `true` whenever the first fetch for the query is in-flight
`isFetching` is `true` whenever any fetch for the query is in-flight

### Explanation
In the upgrade dialog, `useKubernetesClusterMutation` is called, which invalidates the paginated list of Kubernets clusters, which causes `useKubernetesClustersQuery`'s `isFetching` to become `true`, which unmounted the upgrade dialog because of the if statement below. Switching to `isLoading` allows the Upgrade dialog to stay mounted during the upgrade, which fixes the bug

There are probably other ways to fix this bug. For example, the `UpgradeVersionModal` component could probably be made more robust. I decided to go with this fix because I think `isLoading` is the desired loading state here rather than `isFetching` anyway

coliu-akamai:
The recent test addition was failing on my end when I ran it locally, but I saw it passed in [this run here](https://baker.linode.com/blue/organizations/jenkins/manager%2Fmanager-test/detail/PR-11056/1/pipeline) (and then I tried rerunning the tests but received a couple errors - not sure if I'm missing permissions or something?). Will look into some more!

image

jaalah-akamai (approved):
✅ Thanks for catching and will keep an eye on this detail for future
Screenshot 2024-10-08 at 11 08 56 AM

packages/manager/cypress/e2e/core/kubernetes/lke-update.spec.ts
```diff
@@ -215,6 +216,61 @@ describe('LKE cluster updates', () => {
...
+      .should('be.visible');
+
+    ui.button
+      .findByTitle('Upgrade Version')
+      .should('be.visible')
+      .should('be.enabled')
+      .click();
+
+    mockGetClusters([updatedCluster]).as('getClusters');
+    cy.wait(['@updateCluster', '@getClusters']);
```
mjac0bs, line 255:
```suggestion
    mockGetClusters([updatedCluster]).as('getClusters');

    ui.button
      .findByTitle('Upgrade Version')
      .should('be.visible')
      .should('be.enabled')
      .click();

    cy.wait(['@updateCluster', '@getClusters']);
```

This appears to be what we actually need here to get the version to update to 1.26. I am not 100% clear on why the mock needs to go before the user action, but know that often when the right requests are being made, but the right data isn't showing, it's this kind of (timing?) issue.

mjac0bs (approved):
Thanks for fixing this. Can we convert the internal PDI ticket to an M3 ticket, so we have it in our project, and update this PR title?

I don't know how Jaalah managed to get the tests to pass. I was seeing the same thing as Connie. Left a comment on the spec about what seems to fix it.

Other than that, this looks good for approving pending the test update!

packages/manager/src/features/Kubernetes/KubernetesLanding/KubernetesLanding.tsx
hkhalil-akamai: Thanks for the context. Agreed that the dialog is not in the best shape but that is out of scope.

packages/manager/cypress/e2e/core/kubernetes/lke-update.spec.ts
coliu-akamai: glad I wasn't going crazy! I'm also curious to how this test managed to pass in the automatic test checks when it's been failing for us.. 😅

hkhalil-akamai: Also seeing this test fail locally.

hkhalil-akamai (approved):
Thanks for looking into this and adding relevant tests 🧪. Also seeing flake in the new test along with @mjac0bs. Approved pending test is fixed and fix changeset is added.

packages/manager/cypress/e2e/core/kubernetes/lke-update.spec.ts
bnussman-akamai: Good catch @mjac0bs 

I knew something felt weird when I wrote that 

Fixed in 1d66657

PR #11057, opened 2024-10-07
upcoming: [M3-8638] - Introduce the new `(at)linode/ui` package

hkhalil-akamai:
## Description 📝
Introduces the new `@linode/ui` package, which will host our shared UI component library. The package also supports unit tests via `vitest` and will eventually become the home of our Storybook declarations.

This is the first step in our modularization effort. These components will eventually be published and used outside of Cloud Manager, such as in downstream microfrontends. Therefore, they need to function independently of any tools, utilities or components specific to Cloud Manager.

The package must support unit testing through `vitest`. Existing unit tests will be migrated alongside their associated components.

Storybook and all storybook declarations will also eventually be hosted from this package.

## Changes  🔄
- Introduces the new `@linode/ui` package
- Update `generate-changelog` and `package-version` scripts to include this package
- Update `ci.yml` to typecheck and test this package
- Copy the `Chip` and `BetaChip` components to demonstrate the usage of this package

## How to test 🧪

- [ ] Importing components from this package into Cloud Manager works as expected
- [ ] Type checking passes
- [ ] All unit tests pass
- [ ] Release scripts continue to work as expected with this package (changelog, package versions)
- [ ] CI passes -- the package passes typechecking and unit testing


bnussman-akamai (commented):
Just realized this was a draft 🤦  Going to leave this early feedback so it doesn't get lost but will wait to continue reviewing, sorry about that!

packages/ui/.prettierrc
hkhalil-akamai, line 1: Copied from `packages/api-v4/.prettierrc`

packages/ui/src/components/Chip.tsx
```diff
@@ -0,0 +1,13 @@
+import { default as _Chip, ChipProps as _ChipProps } from '@mui/material/Chip';
+import * as React from 'react';
+
+export interface ChipProps extends _ChipProps {
+  /**
+   * Optional component to render instead of a span.
+   */
+  component?: React.ElementType;
+}
```
jaalah-akamai, line 9: Something we should address in a follow-up. I don't remember why we ever needed this abstraction, but the `_ChipProps` already has `component` as part of it's type.

jaalah-akamai (approved):
This is looking great. Approving pending addition of CHANGELOG.md.

✅ Changesets are working in package
✅ `yarn package-versions` works as intended
Screenshot 2024-10-08 at 4 38 09 PM

✅ Configs / Dependencies look good
❌ Need to create a `CHANGELOG.md` file for `yarn generate-changelogs`
Screenshot 2024-10-08 at 4 39 38 PM

README.md
mjac0bs, line 1: This can totally be a later thing, but do we want to add chips for the new `linode/ui` package? I find them useful to reference during releases.

Screenshot 2024-10-08 at 2 04 21 PM

packages/ui/vitest.config.ts
mjac0bs, line 1: Minor and non-blocking: Is there anything we can change in our root `package.json` to be able to run tests from both manager and ui packages? It looks like we have `"test": "yarn workspace linode-manager test"` configured in the root `package.json` scripts, but when I ran `yarn test` from the root, I was only seeing `manager` tests run. 

image


Navigating into the `ui` package and running `yarn test` from there worked as expected.


Screenshot 2024-10-08 at 2 31 27 PM

mjac0bs (commented):
- [x] Importing components from this package into Cloud Manager works as expected - confirmed chips look good
- [x]  Type checking passes
- [x]  All unit tests pass - left a ❓ about running tests locally
- [ ]  Release scripts continue to work as expected with this package (changelog, package versions) - no issues other than what Jaalah already noted 👍🏼 
- [x]  CI passes -- the package passes typechecking and unit testing

.github/workflows/ci.yml
bnussman-akamai, line 1: Can we add a job that typechecks the new UI package?

packages/ui/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "@emotion/react": "^11.11.1",
+    "@emotion/styled": "^11.11.0",
+    "@mui/material": "^5.14.7",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
+  },
+  "scripts": {
+    "start": "tsc -w --preserveWatchOutput",
+    "lint": "eslint . --quiet --ext .js,.ts,.tsx",
+    "typecheck": "tsc --noEmit true --emitDeclarationOnly false",
```
bnussman-akamai, line 28: I think we can remove the flags here. It should just use the options from our tsconfig

```suggestion
    "typecheck": "tsc",
```

bnussman-akamai (approved):
General setup looks good! 

I do think our themes should move to this package, but I imagine that requires some significant changes. We can do this as soon as it's feasible

README.md
hkhalil-akamai: Since we're planning to publish this package later, I created M3-8729 to capture this.

packages/ui/vitest.config.ts
hkhalil-akamai: I have some reservations about `&&`ing these two commands together since it might make it hard to read the output of the test command. For example, failing Manager tests would be hidden somewhere in the middle of the command output.

I wonder if there is a way to instruct `vitest` to run both test suites as if they are one -- curious if @bnussman-akamai has any suggestions here.

bnussman-akamai: I also am anti-`&&`

I think we should just add commands like `yarn test:manager` and `yarn test:ui`

I don't really know of anything else more elegant we can do 🫤

.github/workflows/ci.yml
hkhalil-akamai: [Added](https://github.com/linode/manager/pull/11057/commits/e23a7fad29bcc1270e0742f4186e2ea2702adf6f). Side note: why don't we typecheck the api-v4 and validation packages?

packages/ui/src/components/Chip.tsx
hkhalil-akamai: Figured this was small enough to include in this PR [07e44](https://github.com/linode/manager/pull/11057/commits/07e44509a8a5885d601be7e80cf2bd8f75e4277f)

packages/ui/vitest.config.ts
mjac0bs: > For example, failing Manager tests would be hidden somewhere in the middle of the command output.

@hkhalil-akamai Could you elaborate on what you mean by this? Wouldn't the commands run in order -`manager` tests would run first and `ui` tests would run second? That's what I was seeing when I ran it that way, at least.

Using separate commands is okay with me, though - I do just eventually want a way to run all the tests from the root.

mjac0bs (approved):

.github/workflows/ci.yml
bnussman-akamai: The build steps for api-v4 and validation (`build-validation`, `build-sdk`) inherently  type-check because they run `tsc` during the build process

mjac0bs:
Test failures in CI are not related to this PR; merging this on Hussain's behalf.

PR #11062, opened 2024-10-08
Upcoming: [DI-20844] - Tooltip for widget level filters and icons, beta feedbacks and CSS changes for CloudPulse

venkymano-akamai:
## Description 📝
We have added Tooltip for all of our icons and widget level filters. Also addressed some of the beta demo feedbacks like
CamelCase for Aggregation Function dropdown label, Zeroth state of time duration and CSS updates for legend rows inside the widget.

## Changes  🔄

1. Added Tooltip for all icons and widget level filters
2. CamelCase labels for Aggregation Function dropdown in widget
3. Zero state update for time duration to move to Last 30 Minutes
4. Padding and minHeight adjustment for legend rows inside widget
5. Widget Unit scale up display logic update
6. CloudpulseResourceSelection, publish the selected values only if the autocomplete closes since it is a multiselect component and reduce metrics call fired.
7. Interval selection should default to AUTO if there are no preferences in its initial state

## Target release date 🗓️
10-10-2024

## Preview 📷
| Before  | After   |
| ------- | ------- |
|Screenshot 2024-10-08 at 3 54 12 PM|Screenshot 2024-10-08 at 3 54 31 PM|
|Screenshot 2024-10-08 at 3 54 34 PM|Screenshot 2024-10-08 at 3 54 36 PM|
|Screenshot 2024-10-08 at 3 54 48 PM|Screenshot 2024-10-08 at 3 54 51 PM|
|Screenshot 2024-10-08 at 3 55 04 PM|Screenshot 2024-10-08 at 3 55 22 PM|
|Screenshot 2024-10-08 at 7 00 20 PM|Screenshot 2024-10-08 at 4 40 51 PM|
|Screenshot 2024-10-08 at 6 58 49 PM|Screenshot 2024-10-08 at 3 26 54 PM|

## How to test 🧪

1. Login as mock user as some endpoints are not ready
2. Navigate to Monitor page
3. Select dashboard and necessary filters to view the widgets
4. You can see tooltips for icons and widget level filters
5. Also you can see aggregate function dropdown displaying labels in camelcase
9. Once you clear the selected dashboard time duration returns to Last 30 Minutes

*Check all that apply*

- [x] 👀 Doing a self review


packages/manager/src/features/CloudPulse/Widget/CloudPulseWidget.tsx
```diff
@@ -140,6 +140,7 @@ export const CloudPulseWidget = (props: CloudPulseWidgetProperties) => {
     widget: widgetProp,
   } = props;
   const flags = useFlags();
+  const scaledWidgetUnit = React.useRef(unit === 'Bytes' ? 'B' : unit);
```
nikhagra-akamai, line 143:
```suggestion
  const scaledWidgetUnit = React.useRef(generateCurrentUnit(unit));
```

https://github.com/nikhagra-akamai/manager/blob/0bc9d73c4f351b6718061891cc6a9036ca4dff89/packages/manager/src/features/CloudPulse/Utils/unitConversion.ts#L222

nikhagra-akamai (changes requested):

nikhagra-akamai (approved):

packages/manager/src/features/CloudPulse/shared/CloudPulseTimeRangeSelect.tsx
```diff
@@ -58,8 +58,12 @@ export const CloudPulseTimeRangeSelect = React.memo(
           false
         );
       }
+
+      if (item !== selectedTimeRange) {
+        setSelectedTimeRange(item);
+      }
       // eslint-disable-next-line react-hooks/exhaustive-deps
-    }, []); // need to execute only once, during mounting of this component
+    }, [defaultValue]); // need to execute only once, during mounting of this component
```
jaalah-akamai, line 66: The comment here is no longer true. In addition, I believe I was able to get this to work without useEffect, storing state, and without some of the other things in here:

```
export const CloudPulseTimeRangeSelect = React.memo(
  (props: CloudPulseTimeRangeSelectProps) => {
    const { defaultValue, handleStatsChange, savePreferences } = props;

    const defaultOption = React.useMemo(
      () => OPTIONS.find((o) => o.label === defaultValue) || OPTIONS[0],
      [defaultValue]
    );

    return (
      <Autocomplete
        onChange={(_, value: Item<Labels, Labels>) => {
          const timeDuration = TIME_RANGES[value.value];
          handleStatsChange?.(timeDuration, value.value, savePreferences);
        }}
        autoHighlight
        data-testid="cloudpulse-time-duration"
        value={defaultOption}
        disableClearable
        fullWidth
        isOptionEqualToValue={(option, value) => option.value === value.value}
        label="Select a Time Duration"
        options={OPTIONS}
        textFieldProps={{ hideLabel: true }}
      />
    );
  }
);
```

I also moved `getTimeDurationFromTimeRange` to a map lookup:

```
const TIME_RANGES: Record<Labels, TimeDuration> = {
  'Last 7 Days': { unit: 'days', value: 7 },
  'Last 12 Hours': { unit: 'hr', value: 12 },
  'Last 24 Hours': { unit: 'hr', value: 24 },
  'Last 30 Days': { unit: 'days', value: 30 },
  'Last 30 Minutes': { unit: 'min', value: 30 },
};

const OPTIONS: Item<Labels, Labels>[] = Object.keys(TIME_RANGES).map(
  (label) => ({
    label: label as Labels,
    value: label as Labels,
  })
);
```

packages/manager/src/features/CloudPulse/Widget/components/CloudPulseAggregateFunction.tsx
```diff
@@ -52,26 +55,31 @@ export const CloudPulseAggregateFunction = React.memo(
...
-      />
+      <CloudPulseTooltip title={'Aggregation function'}>
+        <StyledWidgetAutocomplete
+          getOptionLabel={(option: { label: string }) => {
+            return convertStringToCamelCasesWithSpaces(option.label); // options needed to be display in Caps first
+          }}
+          isOptionEqualToValue={(option, value) => {
+            return option.label === value.label;
+          }}
+          onChange={(_: any, selectedAggregateFunc: any) => {
```
bnussman-akamai, line 66: Can we fix both of these `any` types?

packages/manager/src/features/CloudPulse/Widget/components/CloudPulseAggregateFunction.tsx
```diff
@@ -52,26 +55,31 @@ export const CloudPulseAggregateFunction = React.memo(
...
-        fullWidth={false}
-        label="Select an Aggregate Function"
-        noMarginTop={true}
-        options={availableAggregateFunc}
-        sx={{ width: '100%' }}
-        value={selectedAggregateFunction}
-      />
+      <CloudPulseTooltip title={'Aggregation function'}>
+        <StyledWidgetAutocomplete
+          getOptionLabel={(option: { label: string }) => {
```
bnussman-akamai, line 60:
```suggestion
          getOptionLabel={(option) => {
```

We shouldn't need to put a type here. The typescript should be able to infer the type based on the Autocomplete's `options`/`value`

packages/manager/src/features/CloudPulse/shared/CloudPulseTooltip.tsx
```diff
@@ -0,0 +1,48 @@
...
+  },
+  [`& .${tooltipClasses.tooltip}`]: {
+    backgroundColor: theme.palette.common.black,
+    color: theme.palette.common.white,
+    fontSize: theme.spacing(1.75),
+    maxHeight: theme.spacing(3.5),
+    maxWidth: theme.spacing(30),
+    padding: theme.spacing(0.75),
+  },
+}));
```
jaalah-akamai, line 48: I recommend we refrain from implementing this change at present. In fact, I strongly suggest removing it entirely. We have a larger initiative planned to align our tooltips with CDS 2.0 across the entire platform. Implementing this now would create additional cleanup work for us later. Moreover, if we were to keep this functionality, it would likely be more appropriate to place it in a different file.

packages/manager/src/features/CloudPulse/Widget/components/Zoomer.tsx
```diff
@@ -20,37 +22,40 @@ export const ZoomIcon = React.memo((props: ZoomIconProperties) => {
   const ToggleZoomer = () => {
     if (props.zoomIn) {
       return (
+        <CloudPulseTooltip placement="bottom-end" title="Minimize">
+          <IconButton
+            sx={{
+              color: theme.color.grey1,
+              fontSize: 'x-large',
+              height: `calc(${theme.spacing(4)} + 2px)`,
```
jaalah-akamai, line 30: Nit: Do we need a height here? The flexbox already aligns it in the center. It just makes the outline look weird when tab focused: 
Screenshot 2024-10-10 at 11 27 06 AM

packages/manager/src/features/CloudPulse/Widget/components/CloudPulseAggregateFunction.tsx
```diff
@@ -52,26 +55,31 @@ export const CloudPulseAggregateFunction = React.memo(
...
-        disableClearable
-        fullWidth={false}
-        label="Select an Aggregate Function"
-        noMarginTop={true}
-        options={availableAggregateFunc}
-        sx={{ width: '100%' }}
-        value={selectedAggregateFunction}
-      />
+      <CloudPulseTooltip title={'Aggregation function'}>
+        <StyledWidgetAutocomplete
```
bnussman-akamai, line 59: There are some serious type-safety issues with this component that need to be fixed. 

If I switch `StyledWidgetAutocomplete` to just a plain `Autocomplete`, it seems to improve things

packages/manager/src/features/CloudPulse/shared/CloudPulseResourcesSelect.tsx
```diff
@@ -79,9 +81,22 @@ export const CloudPulseResourcesSelect = React.memo(
 
     return (
       <Autocomplete
-        onChange={(e, resourceSelections: CloudPulseResources[]) => {
+        onChange={(
+          _: React.SyntheticEvent,
+          resourceSelections: CloudPulseResources[]
```
bnussman-akamai, line 86: It's okay to just let typescript infer the types here

```suggestion
        onChange={(e, resourceSelections) => {
```

packages/manager/src/features/CloudPulse/shared/CloudPulseResourcesSelect.tsx
```diff
@@ -50,6 +50,8 @@ export const CloudPulseResourcesSelect = React.memo(
       CloudPulseResources[]
     >();
 
+    const isAutocompleteOpen = React.useRef(false); // Ref to track the open state of Autocomplete
```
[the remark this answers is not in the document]
bnussman-akamai: We do something similar in `packages/manager/src/components/VLANSelect.tsx`, but we use state instead of a ref. 
I think the pattern is okay if it makes a significant optimization 

If we decide to keep this, maybe we can add a comment explaining the optimization

packages/manager/src/features/CloudPulse/Widget/components/CloudPulseAggregateFunction.tsx
venkymano-akamai: @bnussman-akamai , here the type can be either string or label:string, i removed this and added a check at function level

packages/manager/src/features/CloudPulse/Widget/components/CloudPulseAggregateFunction.tsx
venkymano-akamai: Here we need the styled component, as the widget filters are subject to some styles as per UX design

packages/manager/src/features/CloudPulse/Widget/components/CloudPulseAggregateFunction.tsx
venkymano-akamai: done

packages/manager/src/features/CloudPulse/shared/CloudPulseResourcesSelect.tsx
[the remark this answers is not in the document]
venkymano-akamai: Yes, the useRef here reduces the number of re-renders the component goes through, so we will keep this approach. I have added a comment as well.

packages/manager/src/features/CloudPulse/shared/CloudPulseTimeRangeSelect.tsx
venkymano-akamai: @jaalah-akamai , here I have removed the comments and added appropriate. Implementing this works in product page, whereas in the contextual view it is not working, since we are caching the defaultOption based on defaultValue in useMemo, in the contextual view there is no preferences at all so the default value will always be same and it will always not get recomputed if we change the value in the dropdown

I think we need to spend more time on refactoring this approach. This improves the code whereas it doesn't optimize the user experience, also we can create a ticket to track these refactoring works.

packages/manager/src/features/CloudPulse/shared/CloudPulseTooltip.tsx
venkymano-akamai: @jaalah-akamai , As we don't know the alignment dates for CDS 2.0 and this is an important feature and a UX requirement,  we can keep this for now, it is already kept in a separate file, on CDS 2.0 alignment, only changes in this file should be enough.

packages/manager/src/features/CloudPulse/Widget/components/Zoomer.tsx
venkymano-akamai: removed!!

packages/manager/src/features/CloudPulse/shared/CloudPulseTooltip.tsx
jaalah-akamai: How is a design change that departs from what Cloud Manager is doing today across all other features a requirement? This is a CDS 2.0 design element that has not been rolled out into Cloud Manager yet. This needs to be put on hold until this is tackled as part of the CDS 2.0 rollout effort which is currently underway in this quarter.

packages/manager/src/features/CloudPulse/Widget/components/CloudPulseAggregateFunction.tsx
venkymano-akamai: Upon discussion with @jaalah-akamai  /. @bnussman-akamai , we have agreed to remove styled and use sx prop instead

venkymano-akamai: @jaalah-akamai / @bnussman-akamai  This is done.

jaalah-akamai:
This is looking better - just waiting on the tooltip update before approval

venkymano-akamai:
> This is looking better - just waiting on the tooltip update before approval

@jaalah-akamai , we have moved to the common tooltip

packages/manager/src/features/CloudPulse/shared/CloudPulseTooltip.tsx
venkymano-akamai: @kmuddapo , Thanks, I have already updated to use existing tooltip

packages/manager/src/features/CloudPulse/shared/CloudPulseResourcesSelect.tsx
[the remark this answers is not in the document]
venkymano-akamai: As per discussion, the useRef approach solves our use case better and we will use it.

packages/manager/src/features/CloudPulse/shared/CloudPulseTimeRangeSelect.tsx
venkymano-akamai: As per the discussion, it is agreed we will not go ahead with this approach as it needs more code refactoring, we can take these later and for now will go with defaultValue based useEffect approach.

jaalah-akamai (approved):

packages/manager/src/features/CloudPulse/Widget/components/CloudPulseAggregateFunction.tsx
```diff
@@ -51,27 +57,34 @@ export const CloudPulseAggregateFunction = React.memo(
...
+          textFieldProps={{
+            hideLabel: true,
+          }}
+          autoHighlight
+          disableClearable
+          fullWidth={false}
+          label="Select an Aggregate Function"
+          noMarginTop={true}
+          options={availableAggregateFunc}
+          sx={getAutocompleteWidgetStyles(theme)}
```
bnussman-akamai, line 84: You can actually pass a  function to `sx`.  The theme will be passed automatically, so you can even remove the `useTheme` hook above

```suggestion
          sx={getAutocompleteWidgetStyles}
```

packages/manager/src/features/CloudPulse/Widget/components/CloudPulseIntervalSelect.tsx
```diff
@@ -118,36 +122,42 @@ export const CloudPulseIntervalSelect = React.memo(
...
+              unit: selectedInterval?.unit,
+              value: selectedInterval?.value,
+            });
+          }}
+          textFieldProps={{
+            hideLabel: true,
+          }}
+          autoHighlight
+          disableClearable
+          fullWidth={false}
```
bnussman-akamai, line 153: Same feedback here. Is this nessesaery?

packages/manager/src/features/CloudPulse/Widget/components/CloudPulseIntervalSelect.tsx
```diff
@@ -118,36 +122,42 @@ export const CloudPulseIntervalSelect = React.memo(
...
+            option: IntervalOptions,
+            value: IntervalOptions
+          ) => {
+            return (
+              option?.value === value?.value && option?.unit === value?.unit
+            );
+          }}
+          onChange={(
+            _: React.SyntheticEvent,
+            selectedInterval: IntervalOptions
```
bnussman-akamai, line 140: Now that the type-safety issue is fixed, please remove these explicit types. Allowing typescript to infer the types will be more type-safe

packages/manager/src/features/CloudPulse/Widget/components/CloudPulseIntervalSelect.tsx
```diff
@@ -118,36 +122,42 @@ export const CloudPulseIntervalSelect = React.memo(
...
-        noMarginTop={true}
-        options={[autoIntervalOption, ...availableIntervalOptions]}
-        sx={{ width: { xs: '100%' } }}
-        value={selectedInterval}
-      />
+      <CloudPulseTooltip title={'Data aggregation interval'}>
+        <Autocomplete
+          isOptionEqualToValue={(
+            option: IntervalOptions,
+            value: IntervalOptions
```
bnussman-akamai, line 132: Same thing here. No need to explicitly put the types now that the type-safety is fixed

packages/manager/src/features/CloudPulse/Widget/components/CloudPulseIntervalSelect.tsx
```diff
@@ -118,36 +122,42 @@ export const CloudPulseIntervalSelect = React.memo(
...
+          textFieldProps={{
+            hideLabel: true,
+          }}
+          autoHighlight
+          disableClearable
+          fullWidth={false}
+          label="Select an Interval"
+          noMarginTop={true}
+          options={[autoIntervalOption, ...availableIntervalOptions]}
+          sx={getAutocompleteWidgetStyles(theme)}
```
bnussman-akamai, line 157: Same thing here. `sx` can take a function. This will allow us to remove the `useTheme` call above 
```suggestion
          sx={getAutocompleteWidgetStyles}
```

packages/manager/src/features/CloudPulse/shared/CloudPulseTooltip.tsx
bnussman-akamai, line 1: Should we keep this abstraction or just use `Tooltip` directly? cc @jaalah-akamai  This seems fine, but I feel like one could argue is it an extra abstraction that would be nice to not need if possible

bnussman-akamai (commented):
Just a few small things then I will ✅

packages/manager/src/features/CloudPulse/shared/CloudPulseTooltip.tsx
venkymano-akamai: @bnussman-akamai , thanks for the suggestion, I suggest we can keep this abstraction, so in future any CloudPulse specific changes can go in one tooltip component rather than doing it in all filters.

venkymano-akamai:
@bnussman-akamai , I have addressed the comments you have left. Please check once

bnussman-akamai (approved):

jaalah-akamai:
I had to re-run the CI E2E... when it passes, we'll merge

venkymano-akamai:
@jaalah-akamai , for this PR, in the same branch, in local if run yarn coverage:summary, yarn test, our ACLP cy, everything is passing. I don't know whats with the CI

PR #11064, opened 2024-10-08
upcoming: [DI-21270] - Added the Alerts tab

santoshp210-akamai:
## Description 📝
Added Alert tab under Monitor for the alerting services

## Changes  🔄
- Added Alerts tab under Monitor which is visible when at least one flag of the sub-tabs is accessible(via feature flag)
- Added Recent activity, Alert Definitions, Notification channels tabs under Alert and the tabs visible only when they're accessible through feature flags

## Target release date 🗓️

## Preview 📷

| Before  | After   |
| ------- | ------- |
| Screenshot 2024-09-04 at 14 47 05 | Screenshot 2024-10-08 at 18 28 48 |

## How to test 🧪
- Switch to mock user as APIs are not live and for the feature flags(Legacy MSW handlers)

*Check all that apply*

- [x] 👀 Doing a self review

### DIFF
--- a/packages/manager/src/features/CloudPulse/Alerts/AlertsLanding/AlertsLanding.tsx
+++ b/packages/manager/src/features/CloudPulse/Alerts/AlertsLanding/AlertsLanding.tsx
@@ -15,13 +15,13 @@ import { useFlags } from 'src/hooks/useFlags';
 
 import { AlertDefinitionLanding } from './AlertsDefinitionLanding';
 
-import type { EnabledTab } from '../../CloudPulseTabs';
+import type { EnabledAlertTab } from '../../CloudPulseTabs';
 
 export const AlertsLanding = React.memo(() => {
   const flags = useFlags();
   const { url } = useRouteMatch();
   const { pathname } = useLocation();
-  const tabFlags = React.useMemo<EnabledTab[]>(
+  const alertTabs = React.useMemo<EnabledAlertTab[]>(
     () => [
       {
         isEnabled: Boolean(flags.aclpAlerting?.alertDefinitions),
@@ -32,10 +32,10 @@ export const AlertsLanding = React.memo(() => {
   );
   const accessibleTabs = React.useMemo(
     () =>
-      tabFlags
-        .filter((tabFlag) => tabFlag.isEnabled)
-        .map((tabFlag) => tabFlag.tab),
-    [tabFlags]
+      alertTabs
+        .filter((alertTab) => alertTab.isEnabled)
+        .map((alertTab) => alertTab.tab),
+    [alertTabs]
   );
   const activeTabIndex = React.useMemo(
     () =>
--- a/packages/manager/src/features/CloudPulse/CloudPulseTabs.tsx
+++ b/packages/manager/src/features/CloudPulse/CloudPulseTabs.tsx
@@ -17,7 +17,7 @@ import { CloudPulseDashboardLanding } from './Dashboard/CloudPulseDashboardLandi
 
 import type { Tab } from 'src/components/Tabs/TabLinkList';
 
-export type EnabledTab = {
+export type EnabledAlertTab = {
   isEnabled: boolean;
   tab: Tab;
 };
@@ -25,7 +25,7 @@ export const CloudPulseTabs = () => {
   const flags = useFlags();
   const { url } = useRouteMatch();
   const { pathname } = useLocation();
-  const tabFlags = React.useMemo<EnabledTab[]>(
+  const alertTabs = React.useMemo<EnabledAlertTab[]>(
     () => [
       {
         isEnabled: true,
@@ -50,10 +50,10 @@ export const CloudPulseTabs = () => {
   );
   const accessibleTabs = React.useMemo(
     () =>
-      tabFlags
-        .filter((tabFlag) => tabFlag.isEnabled)
-        .map((tabFlag) => tabFlag.tab),
-    [tabFlags]
+      alertTabs
+        .filter((alertTab) => alertTab.isEnabled)
+        .map((alertTab) => alertTab.tab),
+    [alertTabs]
   );
   const activeTabIndex = React.useMemo(
     () =>


jdamore-linode:
@santoshp210-akamai it looks like this branch might be pretty out-of-date. Would you mind pulling in the latest changes from our `develop` branch so all our latest tests run, etc.?

Thank you!

packages/manager/src/features/CloudPulse/Alerts/AlertsLanding/AlertsLanding.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const accessibleTabs = tabs.filter((tab) => tab.accessible);
+  const location = useLocation();
+  const navToURL = (index: number) => {
+    history.push(accessibleTabs[index].routeName);
+  };
+  const matches = (p: string) => {
+    return Boolean(matchPath(location.pathname, { exact: false, path: p }));
+  };
+  return (
+    <Paper>
```
jaalah-akamai, line 40: Let's add `sx={{padding: 2}}` to the `Paper` as per design

packages/manager/src/features/CloudPulse/Alerts/AlertsLanding/AlertsDefinitionLanding.tsx
```diff
@@ -0,0 +1,25 @@
...
+        exact
+        path="/monitor/cloudpulse/alerts/definitions"
+      />
+    </Switch>
+  );
+};
+
+const AlertDefinition = () => {
+  return (
+    <Paper>
```
jaalah-akamai, line 21: Let's add `sx={(theme) => ({ background: theme.bg.offWhite })}` to the `Paper` as per design

packages/manager/src/features/CloudPulse/Alerts/AlertsLanding/AlertsLanding.tsx
```diff
@@ -0,0 +1,70 @@
...
+      >
+        <Box
+          sx={{
+            aligneItems: 'center',
+            display: 'flex',
+            justifyContent: 'space-between',
+            p: 1,
+            width: '100%',
+          }}
+        >
```
jaalah-akamai, line 54: We can remove this `Box` entirely and the tabs will span the entire content as per design

packages/manager/src/features/CloudPulse/CloudPulseTabs.tsx
```diff
@@ -1,50 +1,69 @@
...
+          />
+          <Route component={AlertsLanding} path={`${props.match.url}/alerts`} />
+          <Redirect
+            exact
+            from="/monitor/cloudpulse"
+            to="/monitor/cloudpulse/dashboards"
+          />
+        </Switch>
       </React.Suspense>
     </StyledTabs>
```
jaalah-akamai, line 68: Let's remove the `StyledTabs` and use `Tabs` directly and just apply `marginTop={0}` as props. No need to turn this into a styled component.

packages/manager/src/features/CloudPulse/CloudPulseTabs.tsx
```diff
@@ -1,50 +1,69 @@
...
 import { Tabs } from 'src/components/Tabs/Tabs';
+import { useFlags } from 'src/hooks/useFlags';
 
+import { AlertsLanding } from './Alerts/AlertsLanding/AlertsLanding';
 import { CloudPulseDashboardLanding } from './Dashboard/CloudPulseDashboardLanding';
 
 import type { RouteComponentProps } from 'react-router-dom';
 type Props = RouteComponentProps<{}>;
 
 export const CloudPulseTabs = React.memo((props: Props) => {
```
jaalah-akamai, line 16: Let's remove the `React.memo`, I feel like this feature is relying on this too much. React's built-in optimizations should be sufficient and we'll optimize when we need to.

packages/manager/src/features/CloudPulse/CloudPulseTabs.tsx
```diff
@@ -1,50 +1,69 @@
...
+      accessible: true,
       routeName: `${props.match.url}/dashboards`,
       title: 'Dashboards',
     },
+    {
+      accessible:
+        flags.aclpAlerting?.alertDefinitions ||
+        flags.aclpAlerting?.recentActivity ||
+        flags.aclpAlerting?.notificationChannels,
+      routeName: `${props.match.url}/alerts`,
```
jaalah-akamai, line 29: Let's switch `props.match.url` to use `url` from the routing hook: const { url } = useRouteMatch();

packages/manager/src/features/CloudPulse/CloudPulseTabs.tsx
```diff
@@ -1,50 +1,69 @@
...
 
+import { AlertsLanding } from './Alerts/AlertsLanding/AlertsLanding';
 import { CloudPulseDashboardLanding } from './Dashboard/CloudPulseDashboardLanding';
 
 import type { RouteComponentProps } from 'react-router-dom';
 type Props = RouteComponentProps<{}>;
 
 export const CloudPulseTabs = React.memo((props: Props) => {
+  const flags = useFlags();
   const tabs = [
```
jaalah-akamai, line 18: Tabs is something you could probably memoize with `useMemo` and `[url, flags.aclpAlerting]` as deps.

packages/manager/src/features/CloudPulse/CloudPulseTabs.tsx
jaalah-akamai: Let's do the same for `pathname`: const { pathname } = useLocation();

packages/manager/src/features/CloudPulse/CloudPulseTabs.tsx
```diff
@@ -1,50 +1,69 @@
...
-    return Boolean(matchPath(p, { path: props.location.pathname }));
+    return Boolean(
+      matchPath(props.location.pathname, { exact: false, path: p })
+    );
   };
 
   const navToURL = (index: number) => {
-    props.history.push(tabs[index].routeName);
+    props.history.push(accessibleTabs[index].routeName);
   };
```
jaalah-akamai, line 42: Remove the `navToURL` function. The TabLinkList component should handle navigation internally.

packages/manager/src/features/CloudPulse/CloudPulseTabs.tsx
```diff
@@ -1,50 +1,69 @@
...
 
   return (
     <StyledTabs
       index={Math.max(
-        tabs.findIndex((tab) => matches(tab.routeName)),
+        accessibleTabs.findIndex((tab) => matches(tab.routeName)),
```
jaalah-akamai, line 47: This is good, but we could memoize this match function probably in a handler above:
```
const activeTabIndex = useMemo(() => 
    Math.max(accessibleTabs.findIndex(tab => pathname.startsWith(tab.routeName)), 0),
  [accessibleTabs, pathname]);
```

packages/manager/src/features/CloudPulse/CloudPulseTabs.tsx
```diff
@@ -1,50 +1,69 @@
...
+    },
   ];
-
+  const accessibleTabs = tabs.filter((tab) => tab.accessible);
   const matches = (p: string) => {
-    return Boolean(matchPath(p, { path: props.location.pathname }));
+    return Boolean(
+      matchPath(props.location.pathname, { exact: false, path: p })
+    );
   };
```
jaalah-akamai, line 38: Shouldn't need `matches` since we can directly compare the current pathname with the route names of our tabs in the `activeTabIndex` function I mentioned below.

packages/manager/src/features/CloudPulse/Alerts/AlertsLanding/AlertsDefinitionLanding.tsx
santoshp210-akamai: @jaalah-akamai This particular component is used as a placeholder for the AlertDefinition page, The contents of the page will be part of a later feature which we are planning to do after the create alert definition flow.

packages/manager/src/features/CloudPulse/Alerts/AlertsLanding/AlertsLanding.tsx
santoshp210-akamai: There is a Create button that is supposed to come after the tabs, so we had initially wrapped everything in the `Box` , too keep PR minimal we just went ahead with this tabs component.

packages/manager/src/features/CloudPulse/Alerts/AlertsLanding/AlertsLanding.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { Tabs } from 'src/components/Tabs/Tabs';
+import { useFlags } from 'src/hooks/useFlags';
+
+import { AlertDefinitionLanding } from './AlertsDefinitionLanding';
+
+export const AlertsLanding = React.memo(() => {
+  const flags = useFlags();
+  const { url } = useRouteMatch();
+  const { pathname } = useLocation();
+  const tabs = React.useMemo(
```
hkhalil-akamai, line 22: Can we add a type parameter to this?
```suggestion
  const tabs = React.useMemo<Tab[]>(
```

Also add to the imports:

```
import type { Tab } from 'src/components/Tabs/TabLinkList';
```

packages/manager/src/features/CloudPulse/CloudPulseTabs.tsx
```diff
@@ -1,57 +1,69 @@
...
-
-  return (
-    <StyledTabs
-      index={Math.max(
-        tabs.findIndex((tab) => matches(tab.routeName)),
+export const CloudPulseTabs = () => {
+  const flags = useFlags();
+  const { url } = useRouteMatch();
+  const { pathname } = useLocation();
+  const tabs = React.useMemo(
```
hkhalil-akamai, line 22: Same comment here.

```suggestion
  const tabs = React.useMemo<Tab[]>(
```

hkhalil-akamai (approved):
One small suggestions, otherwise looks good.

packages/manager/src/features/CloudPulse/Alerts/AlertsLanding/AlertsLanding.tsx
santoshp210-akamai: @hkhalil-akamai , The accessible property added to the tabs here for the visibility might not be the same as the Tab from the TabLinkList , should we create a custom type for this here or is there any other way to do this ?

jaalah-akamai (approved):

packages/manager/src/features/CloudPulse/Alerts/AlertsLanding/AlertsLanding.tsx
jaalah-akamai: Oh good callout Hussain. Yes we want to make changes here. Give me a few and I'll point them out

jaalah-akamai: We don't want to change the Tab interface, so I suggest you create a new interface that's something like:
```
interface AlertsTab {
  isEnabled: boolean;
  tab: Tab;
}
```

```
const tabs = React.useMemo<AlertsTab[]>(
    () => [
      {
        isEnabled: Boolean(flags.aclpAlerting?.alertDefinitions),
        tab: {
          routeName: `${url}/definitions`,
          title: 'Definitions',
        },
      },
      // Add more tabs here as needed
    ],
    [url]
  );
```

Then you can filter based on that
```
const accessibleTabs = React.useMemo(
    () =>
      tabFlags
        .filter((tabFlag) => tabFlag.isTabEnabled)
        .map((tabFlag) => tabFlag.tab),
    [tabFlags, flags]
  );
```

jaalah-akamai (changes requested):

packages/manager/src/features/CloudPulse/Alerts/AlertsLanding/AlertsLanding.tsx
jaalah-akamai: Yea that makes sense. I updated my snippets

jaalah-akamai (approved):

PR #11067, opened 2024-10-08
feat: [M3-8722] - Improve weblish retry behavior

plisiecki1:
## Description 📝

- automatically reconnect weblish sessions that have disconnected
- … unless it fails rapidly (more than 3 times in a 1 minute window)
- … in which case show an error screen with as much information as we can provide
- … where the error screen includes a button to try again.
- focus the terminal element when starting (which is generally helpful but also prevents _losing_ focus after a reconnect)

Previously, a failure would often just result in a spinning circle forever.

## Changes  🔄
- added option “action button” feature to ErrorState
- added some helpers to `Lish.tsx` which may also be useful for similar enhancements to graphical lish (glish) in the future.
- changed `socket` to be mutable and nullable so it can be updated when reconnecting. `origSocket` local variable is used to ensure that “stale” events are properly ignored.
- moved error parsing/handling to the `close` handler, which also eliminates the potential for console output that too closely matches an "expired" error being interpreted as an error and breaking the session immediately.

## Target release date 🗓️
TBD.

## Preview 📷
| Before  | After   |
| ------- | ------- |
| M3-8722-before | M3-8722-error |

### Prerequisites
Running linode(s).

### Reproduction steps
Launch a lish session for a machine, type `^a d kill <enter>`. The session will exit and show a spinning circle.

### Verification steps
- Launch a lish session for a machine, type `^a d kill <enter>`. The session will reconnect.
- Repeat, it should reconnect again.
- Repeat, it should reconnect again.
- Repeat, it should show an error message (if all three attempts are withing 60s).
- Click the "Retry Connection" button. It should reconnect.
- Ensure that the reconnected session actually works, i.e., that interaction with the terminal is successful.

*Check all that apply*

- [x] 👀 Doing a self review


packages/manager/src/components/ErrorState/ErrorState.tsx
```diff
@@ -28,10 +29,19 @@ export interface ErrorStateProps {
...
    */
   typographySx?: SxProps<Theme>;
+  /**
+   * Optional Action Button: Text
+   */
+  actionButtonText?: string;
+  /**
+   * Optional Action Button: Click handler
+   */
+  onActionButtonClick?: ()=>void;
```
hana-linode, line 39: We can make a separate interface to group the two props together and make it optional

```typescript
interface ActionButtonProps {
  onClick: () => void;
  text: string;
}
```

```suggestion
  actionButtonProps?: ActionButtonProps;
```

packages/manager/src/features/Lish/Lish.tsx
```diff
@@ -23,10 +23,48 @@ import type { Tab } from 'src/components/Tabs/TabLinkList';
 
 const AUTH_POLLING_INTERVAL = 2000;
 
+export class RetryLimiter {
```
hana-linode, line 26: We no longer create new components using `class` as that's an old pattern. Can you convert this to a[ functional component](https://github.com/linode/manager/blob/develop/docs/development-guide/02-component-structure.md)?

packages/manager/src/features/Lish/Lish.tsx
```diff
@@ -23,10 +23,48 @@ import type { Tab } from 'src/components/Tabs/TabLinkList';
...
+    error = errObj;
+  } else if (errObj?.reason) {
+    error = `${errObj?.reason}`;
+  } else if (errObj?.errors?.[0]?.reason) {
+    error = `Error code: ${errObj.errors[0].reason}`;
+  }
+  if (errObj?.grn) {
+    error = `${error} (${errObj?.grn})`;
+  }
+  return error;
```
hana-linode, line 61: We can just return without the intermediary error variable
```suggestion
  if (typeof errObj === 'string') {
    return errObj;
  } else if (errObj?.reason) {
    return `${errObj?.reason}`;
  } else if (errObj?.errors?.[0]?.reason) {
    return `Error code: ${errObj.errors[0].reason}`;
  }
  if (errObj?.grn) {
    return `${defaultError} (${errObj?.grn})`;
  }
  return defaultError;
```

packages/manager/src/features/Lish/Weblish.tsx
```diff
@@ -56,13 +62,59 @@ export class Weblish extends React.Component<Props, State> {
...
     this.socket.addEventListener('open', () => {
       if (!this.mounted) {
         return;
       }
-      this.setState({ renderingLish: true }, () => this.renderTerminal());
+      this.setState({ renderingLish: true }, () => this.renderTerminal(origSocket));
+    });
+
+    this.socket.addEventListener('close', (evt) => {
+      if (this.socket != origSocket) {
```
hana-linode, line 81: We use double not equal over single not equal for type safety
```suggestion
      if (this.socket !== origSocket) {
```

packages/manager/src/features/Lish/Weblish.tsx
```diff
@@ -102,10 +159,25 @@ export class Weblish extends React.Component<Props, State> {
...
     const { group, label } = linode;
 
+    if (this.socket === null) {
+      return;
+    }
+    const socket: WebSocket = this.socket;
+
+    if (socket != origSocket) {
+      return;
+    }
```
hana-linode, line 173:
```suggestion
    const socket: WebSocket | null = this.socket;
    if (socket === null || socket !== origSocket) {
      return;
    }
```

packages/manager/src/components/ErrorState/ErrorState.tsx
plisiecki1: Great idea! I have pushed an update.

packages/manager/src/features/Lish/Weblish.tsx
plisiecki1: Done!

packages/manager/src/features/Lish/Weblish.tsx
plisiecki1: I ended up with what I had to try to avoid null-related typing errors. Your suggestion still resolves those errors, but the code is much neater. Thanks!

packages/manager/src/features/Lish/Lish.tsx
plisiecki1: The last case appends to the error already built. I reorganized the code to use two helper functions that construct the two parts separately. The resulting code is much easier to undestand.

packages/manager/src/features/Lish/Weblish.tsx
```diff
@@ -56,24 +62,78 @@ export class Weblish extends React.Component<Props, State> {
...
+    this.socket.addEventListener('close', (evt) => {
+      if (this.socket !== origSocket) {
+        return;
+      }
+      this.socket = null;
+      this.terminal?.dispose();
+      this.setState({ renderingLish: false });
+      if (!this.mounted) {
+        return;
+      }
```
hana-linode, line 89: We can combine these early returns

```suggestion
      if (this.socket !== origSocket || !this.mounted) {
        return;
      }
      this.socket = null;
      this.terminal?.dispose();
      this.setState({ renderingLish: false });
```

packages/manager/src/features/Lish/Weblish.tsx
```diff
@@ -56,24 +62,78 @@ export class Weblish extends React.Component<Props, State> {
...
+          parsed = JSON.parse(evt.reason);
+        } catch {
+        }
+      }
+      if (parsed === null) {
+        try {
+          parsed = JSON.parse(this.lastMessage);
+        } catch {
+        }
+      }
```
hana-linode, line 103: And these try catches
```suggestion
      if (evt?.reason) {
        try {
          parsed = JSON.parse(evt.reason);
          if (parsed === null) {
            parsed = JSON.parse(this.lastMessage);
          }
        } catch {}
      }
```

packages/manager/src/features/Lish/Weblish.tsx
```diff
@@ -56,24 +62,78 @@ export class Weblish extends React.Component<Props, State> {
...
+
+      if (this.retryLimiter.retryAllowed()) {
+        if (parsed?.errors?.[0]?.reason === "expired" ||
+            (parsed?.type === "error" &&
+             typeof parsed?.reason === "string" &&
+             parsed?.reason.toLowerCase() === "your session has expired.")) {
```
hana-linode, line 109: Do we know the `type` of the errors? That is, can we use the `APIError` type here and in the format functions?

packages/manager/src/components/ErrorState/ErrorState.tsx
```diff
@@ -3,10 +3,22 @@ import Grid from '@mui/material/Unstable_Grid2';
...
 
+export interface ActionButtonProps {
+  /**
+   * Text
+   */
+  text: string;
+  /**
+   * Click handler
+   */
+  onClick: ()=>void;
```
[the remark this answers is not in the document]
plisiecki1: Done.

packages/manager/src/components/ErrorState/ErrorState.tsx
```diff
@@ -28,10 +40,14 @@ export interface ErrorStateProps {
    * Styles applied to the error text
    */
   typographySx?: SxProps<Theme>;
+  /**
+   * Optional Action Button
+   */
+  actionButton?: ActionButtonProps;
```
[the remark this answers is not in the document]
plisiecki1: Done.

packages/manager/src/features/Lish/Weblish.tsx
plisiecki1: The intention was to do nothing when `this.socket !== origSocket`, but to run the cleanup whenever `this.socket === origSocket`. But, when `!this.mounted`, we don't want to proceed with the rest of the close handling. So, I believe the code is correct as is.

packages/manager/src/features/Lish/Weblish.tsx
plisiecki1: I made a similar change. (Unlike the original code, the suggestion above does not try to parse `this.lastMessage` when `evt?.reason` is not set. The update that I submitted _also_ moves the check for `evt?.reason` inside the try/catch block which maintains the intended logic while reducing the total amount of code.)

packages/manager/src/features/Lish/Weblish.tsx
plisiecki1: There are two different formats the error might be in, depending on where the error came from, and neither of which conforms to `APIError`:

    {"errors": [{"reason": "expired"}], "grn": "...request trace info..."}
    {"type":"error","reason":"Your session has expired."}

In addition, I would prefer to do all of the explicit typechecking on the incoming value rather than just trust it. It's not clear to me that it's worth creating an entire abstraction around these two formats, which are only accessed in a couple different places.

packages/manager/src/features/Lish/Weblish.tsx
hana-akamai: Gotcha, can we add comment(s) explaining this?

packages/manager/src/components/ErrorState/ErrorState.tsx
```diff
@@ -28,10 +34,14 @@ export interface ErrorStateProps {
    * Styles applied to the error text
    */
   typographySx?: SxProps<Theme>;
+  /**
+   * Optional Action Button
+   */
+  actionButtonProps?: ActionButtonProps;
```
hana-akamai, line 40: The `?` denotes that this is an optional prop so the comment isn't necessary

packages/manager/src/components/ErrorState/ErrorState.tsx
```diff
@@ -3,10 +3,16 @@ import Grid from '@mui/material/Unstable_Grid2';
...
 import * as React from 'react';
 
+import { Button } from 'src/components/Button/Button';
 import { Typography } from 'src/components/Typography';
 
 import { SvgIconProps } from '../SvgIcon';
 
+export interface ActionButtonProps {
+  text: string;
+  onClick: ()=>void;
```
hana-akamai, line 13:
```suggestion
  onClick: () => void;
```

packages/manager/src/features/Lish/Lish.tsx
```diff
@@ -23,10 +23,57 @@ import type { Tab } from 'src/components/Tabs/TabLinkList';
 
 const AUTH_POLLING_INTERVAL = 2000;
 
+export interface RetryLimiterInterface {
+  retryAllowed: ()=>boolean
+  reset: ()=>void
```
hana-akamai, line 28:
```suggestion
  retryAllowed: () => boolean
  reset: () => void
```

packages/manager/src/features/Lish/Weblish.tsx
```diff
@@ -56,24 +62,75 @@ export class Weblish extends React.Component<Props, State> {
...
+          const { refreshToken } = this.props;
+          refreshToken();
+        } else {
+          this.connect();
+        }
+      } else {
+        this.setState({error: formatError(parsed, "Unexpected WebSocket close")});
+      }
```
hana-akamai, line 114: Can we unnest this a bit?
```suggestion
      if (!this.retryLimiter.retryAllowed()) {
        return this.setState({
          error: formatError(parsed, 'Unexpected WebSocket close'),
        });
      }

      if (
        parsed?.errors?.[0]?.reason === 'expired' ||
        (parsed?.type === 'error' &&
          typeof parsed?.reason === 'string' &&
          parsed?.reason.toLowerCase() === 'your session has expired.')
      ) {
        const { refreshToken } = this.props;
        refreshToken();
      } else {
        this.connect();
      }
```

packages/manager/src/features/Lish/Weblish.tsx
plisiecki1: Comments added.

packages/manager/src/components/ErrorState/ErrorState.tsx
plisiecki1: Done.

packages/manager/src/features/Lish/Lish.tsx
plisiecki1: Done.

packages/manager/src/features/Lish/Weblish.tsx
plisiecki1: Flattened. I also refactored the error parsing/formatting some, so hopefully everything is much cleaner & clearer now!

packages/manager/src/features/Lish/Lish.tsx
```diff
@@ -23,10 +23,78 @@ import type { Tab } from 'src/components/Tabs/TabLinkList';
...
+  if (!s) return null;
+
+  let parsed = null;
+  try {
+    parsed = JSON.parse(s);
+  } catch {
+    return null;
+  }
+
+  let grn = (typeof parsed?.grn === "string" ? parsed?.grn : '');
```
hana-akamai, line 66: Since `LishErrorInterface` declares `reason` and `grn` as `string`, we don't need to check that they are strings

```suggestion
  const grn = parsed?.grn ?? '';
```

Can we also address the lint errors?

plisiecki1: Done!

packages/manager/src/features/Lish/Lish.tsx
```diff
@@ -23,10 +23,78 @@ import type { Tab } from 'src/components/Tabs/TabLinkList';
...
+    const reason = parsed?.errors?.[0]?.reason;
+    if (typeof reason === "string") {
+      return {
+        reason: reason,
+        grn: grn,
+        formatted: `Error code: ${reason}${grnFormatted}`,
+        isExpired: reason === "expired"
+      };
+    }
+  }
```
hana-akamai, line 90: What's the difference between these two cases?

plisiecki1: The error may take one of two different formats. `{"type":"error","reason":"..."}` or `{"errors":[{"reason":"..."}]}`. The "expired" case can occur with either format, but the message to check is slightly different. The formatting is also slightly different because the first form tends to output descriptive errors, whereas the second from outputs one-word codes. (Currently, only the second form sends `grn` trace information, but I wanted to leave the option of adding it to the first, so the `grn` is currently handled outside of those two cases.)

hana-akamai (approved):

cpathipa (approved):
Confirming on the verification steps, was able to see error message after couple attempts of ^a+d and able to reconnect by clicking "Retry Connection" button.
image

packages/manager/src/features/Lish/Lish.tsx
```diff
@@ -23,10 +23,76 @@ import type { Tab } from 'src/components/Tabs/TabLinkList';
...
+export const ParsePotentialLishErrorString = (
+  s: null | string
+): LishErrorInterface | null => {
+  if (!s) {
+    return null;
+  }
+
+  let parsed = null;
+  try {
+    parsed = JSON.parse(s);
```
coliu-akamai, line 69: any chance we know what the type/shape of parsed is, instead of using any?

coliu-akamai (approved):
✅ confirmed reconnection x3
✅ confirmed error message after with Retry button
+ repeated this a bunch of times and confirmed it was consistent 🎉 

We'll also want to include a changeset for this

packages/manager/src/features/Lish/Lish.tsx
plisiecki1: Until we do the checks following the quoted code, we don't know whether the incoming string actually conforms to the expected shape yet. The whole purpose of this function is to check the shape, and then if it is indeed in the expected shape, return the information that is useful to the code handling retries and errors. If we added an explicit type, it would just be used to cast away the "any" type once the shape is confirmed, then the code would immediately extract a few fields, and then discard that type. I don't think the additional overhead would add to the clarity or safety of the code.

PR #11079, opened 2024-10-09
fix: Better sanitize Pendo URLs

mjac0bs:
## Description 📝
After looking at Pendo data to observe what we were collecting, we should better sanitize the data in our URLs with regex that covers more cases (e.g. when the id of the entity was not the last part of the path - `/linodes/12345/storage`; when the entity does not have an id, but a name (buckets)).

## Target release date 🗓️
10/14

### Prerequisites
- Follow the instructions here: https://github.com/linode/manager/pull/10982

### Reproduction steps
- Notice in the dev environment that navigating to places like your Linode's storage tab or LKE cluster details page would include IDs  in the normalizedURL because the regex wasn't specific enough.
- Same goes for OBJ details pages - included the region and bucket name. (For example: click on the 'Access' tab.)

### Verification steps
(How to verify changes)
- Confirm the above is fixed by looking at the requests in the browser console filtering on "pendo".

Examples of what you should be seeing:
Screenshot 2024-10-09 at 1 23 53 PM
Screenshot 2024-10-09 at 1 24 09 PM
Screenshot 2024-10-09 at 1 24 27 PM

*Check all that apply*

- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/.changeset/pr-11079-tech-stories-1728504676937.md
+++ b/packages/manager/.changeset/pr-11079-tech-stories-1728504676937.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Tech Stories
+---
+
+Improve Pendo URL sanitization ([#11079](https://github.com/linode/manager/pull/11079))
--- a/packages/manager/src/hooks/usePendo.ts
+++ b/packages/manager/src/hooks/usePendo.ts
@@ -104,17 +104,22 @@ export const usePendo = () => {
               action: 'Replace',
               attr: 'pathname',
               data(url: string) {
-                const idMatchingRegex = /\d+$/;
+                const idMatchingRegex = /(\/\d+)/;
+                const bucketMatchingRegex = /(buckets\/[^\/]+\/[^\/]+)/;
                 const userPathMatchingRegex = /(users\/).*/;
-                const oauthPathMatchingRegex = /oauth\/callback#access_token/;
-                if (
-                  idMatchingRegex.test(url) ||
-                  oauthPathMatchingRegex.test(url)
-                ) {
-                  // Removes everything after the last /
-                  return url.replace(/\/[^\/]*$/, '/');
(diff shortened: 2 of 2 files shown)


jdamore-linode (approved):
Thanks @mjac0bs! Confirmed the fix using the two testing examples you gave. I also spent a few minutes clicking around the app and failed to find any other instances of IDs being present in `normalizedUrl`

packages/manager/.changeset/pr-11079-tech-stories-1728504676937.md
```diff
@@ -0,0 +1,5 @@
+---
+"@linode/manager": Tech Stories
+---
+
+Improve Pendo URL sanitization ([#11079](https://github.com/linode/manager/pull/11079))
```
coliu-akamai, line 5: since Pendo hasn't been released yet, would we need this changeset?

coliu-akamai (approved):
✅ confirmed the fix + tested a bunch of places with /id-of-entity/ urls and confirmed that works as expected
✅ general testing around CM (especially billing/user settings area) and works as expected

Thanks Mariah! 🎉

jaalah-akamai (approved):

packages/manager/.changeset/pr-11079-tech-stories-1728504676937.md
mjac0bs: Technically, no... I listed it as a tech story and not a customer facing 'fix' because it was not released, but since it's a user data-related thing, I'd rather have record of it than not.

mjac0bs:
All tests that failed in the last CI run pass locally with the exception of clone-linode.spec.ts (which is known to be flaky right now). Not quite sure what happened with this CI run, but wouldn't expect these changes to have any impact on UI or tests. Merging. 

Screenshot 2024-10-09 at 2 44 03 PM

PR #11083, opened 2024-10-10
feat: [M3-8703] - Disable VPC Action Buttons for Restricted Users when they do not have access or have read-only access.

hasyed-akamai:
## Description 📝
To prevent unauthorized access to specific flows and provide clearer guidance, we aim to restrict entry to users without the required permissions.

Here, we are disabling VPC Action buttons when they do not have access or have read-only access.

## Changes  🔄
- For restricted users:
  - Disabled VPC Action buttons (edit, delete)

### Before
<video src="https://github.com/user-attachments/assets/78639cc5-219d-4c96-a132-471fab69ed83" controls width="300"></video>

### After
<video src="https://github.com/user-attachments/assets/afcf3c56-c928-43c7-9a30-76441d932aac" controls width="300"></video>

### Prerequisites

- Log into two accounts side by side:
  - An unrestricted admin user account: full access
  - A restricted user account (use Incognito for this)
    - Start with Read Only for everything
    - or Start with None Permission.

### Reproduction steps

- Observe as restricted user:
  - Action buttons will be disabled and will display their own corresponding tooltiptext notice message.

### Verification steps

- After the changes, observe that the action buttons and the tooltips are tailored to the action.

*Check all that apply*

- [x] 👀 Doing a self review

### DIFF
--- a/packages/manager/src/features/VPCs/VPCLanding/VPCRow.tsx
+++ b/packages/manager/src/features/VPCs/VPCLanding/VPCRow.tsx
@@ -9,7 +9,7 @@ import { TableCell } from 'src/components/TableCell';
 import { TableRow } from 'src/components/TableRow';
 import { useRegionsQuery } from 'src/queries/regions/regions';
 import { getRestrictedResourceText } from 'src/features/Account/utils';
-import { useIsResourceRestricted } from 'src/hooks/useIsResourceRestricted';
+import { useGrants } from 'src/queries/profile/profile';
 
 interface Props {
   handleDeleteVPC: () => void;
@@ -27,11 +27,14 @@ export const VPCRow = ({ handleDeleteVPC, handleEditVPC, vpc }: Props) => {
     0
   );
 
-  const isVPCReadOnly = useIsResourceRestricted({
-    grantLevel: 'read_only',
-    grantType: 'vpc',
-    id: vpc.id,
-  });
+  const { data: grants } = useGrants();
+
+  const isVPCReadOnly =
+    grants &&
+    (grants['vpc'].some(
+      (grant) => grant.id === vpc.id && grant.permissions === 'read_only'
+    ) ||
+      !grants['vpc'].some((grant) => grant.id === vpc.id));
 
   const actions: Action[] = [
     {


packages/manager/src/features/VPCs/VPCLanding/VPCRow.tsx
```diff
@@ -27,11 +27,14 @@ export const VPCRow = ({ handleDeleteVPC, handleEditVPC, vpc }: Props) => {
...
-    id: vpc.id,
-  });
+  const { data: grants } = useGrants();
+
+  const isVPCReadOnly =
+    grants &&
+    (grants['vpc'].some(
+      (grant) => grant.id === vpc.id && grant.permissions === 'read_only'
+    ) ||
+      !grants['vpc'].some((grant) => grant.id === vpc.id));
```
abailly-akamai, line 37: I am wondering if this new check could be baked in the `useIsResourceRestricted` logic? It seems reasonable to me

hasyed-akamai: Added new check in  `useIsResourceRestricted` logic

packages/manager/src/hooks/useIsResourceRestricted.ts
```diff
@@ -19,7 +19,9 @@ export const useIsResourceRestricted = ({
   if (!grants) {
     return false;
   }
-  return grants[grantType].some(
-    (grant) => grant.id === id && grant.permissions === grantLevel
+  return (
+    grants[grantType].some(
+      (grant) => grant.id === id && grant.permissions === grantLevel
```
abailly-akamai, line 24: why not doing a strict equality on `read_only`?

hasyed-akamai: As other components are also using `useIsResourceRestricted` to i tried making it dynamic based on the input of `grantLevel` passed in the `useIsResourceRestricted` in file `packages/manager/src/features/VPCs/VPCLanding/VPCRow.tsx`

```
const isVPCReadOnly = useIsResourceRestricted({
    grantLevel: 'read_only',
    grantType: 'vpc',
    id: vpc.id,
  });
```

abailly-akamai: I understand that, but what is your reasoning for a `grantLevel` that's not `read_only`? Can you clarify?

hasyed-akamai: `if` the `grantLevel` is `read_only` in that case the the below code will run:
```grants[grantType].some((grant)` => grant.id === id && grant.permissions === grantLevel)```

`or` if the permission is none for a particular user then the other check will perform whether the `vpc.id`  is present in the `grants['vpc']` or not:
```!grants[grantType].some((grant) => grant.id === id)```

in above both the cases if it is read_only or else it in not present in the `grants['vpc']` array in that case we need to disable the VPCs Action Buttons

grants Arrays:

Screenshot 2024-10-15 at 9 13 54 PM

VPCs:

Screenshot 2024-10-15 at 9 14 05 PM

Permissions:

Screenshot 2024-10-15 at 9 14 21 PM

in the above case `test1` is having `real_only` `grantLevel` but in case of `test2` it does have `none` permission so we just need to check whether the vpcs is present in the `grants['vpc']` or it has the `grantLevel` to be `read_only` which is done by the newly added checks.

abailly-akamai: Got it - thx this should work

abailly-akamai (approved):

dwiley-akamai (approved):
VPC action buttons disabled appropriately for restricted users with no or read-only access ✅

No regressions observed for other entities ✅

pmakode-akamai (approved):

PR #11085, opened 2024-10-10
test: [M3-7863] - Use `happy-dom` instead of `jsdom` in unit tests

bnussman-akamai:
## Description 📝

- Replaces `jsdom` with `happy-dom` 😄 
- I recommend googling `jsdom` vs `happy-dom` to understand the basic differences but here are some extra callouts
  - `happy-dom` should perform better than `jsdom` 🏎️  (Not seeing a huge boost in perf, but it seems to be ~30sec  to 1 min faster in CI)
  - Test stability seems the same (there are no stability regressions with `happy-dom`) 🪨 
  - In my opinion, the main "risk" with `happy-dom` is its lack of maturity although this doesn't seems to have any significant impact on us. This was, for the most part, a drop-in replacement. 👴 
  - `happy-dom` is known to work on the Bun runtime whereas `jsdom` does not work (yet). This doesn't matter at all for us right now, but if we ever consider the Bun runtime in the near future, this change might be nice. 🥟 
  - For whatever reason, `happy-dom` seemed to be more correct with the css that it renders. (In the diff, you'll see some changes to `toHaveStyle` assertions. `happy-dom` seems to be more inline with what the browser does which is good) 🎨 

## Changes Required 💻 
- I had to update some style assertions (`toHaveStyle`)
  - I compared the test output to what I see in the browser and `happy-dom` appears to match what my browser shows, which is a win for `happy-dom`
- I had to update a small number of test assertions to use different selectors. I didn't dig deep into why I had to make the changes I did but all of the changes seemed logical. If you have questions on specific changes, I can comment in more detail
- Some unit tests that tested actions menus never opened the ActionMenu. I had to add a `userEvent.click` in a few tests to open the ActionMenu before making assertions. Not sure how they were passing on `jsdom`
- Had to increase timeout for a Database related test. Don't love this, but not too bad in the grand scheme of things

## How to test 🧪

- Look at diff and make sure changes actually make sense and are not just changes made to satisfy the test 👀 
- Verify the `test-manager` Github Action passes ✅ 
- Run some tests locally and verify they pass
  - This change does not fix timeouts and failures when running a large number of tests locally. This is still an issue. @jdamore-linode mentioned we may be able to increase the global timeout to improve this at some point

- [ ] 👀 Doing a self review


packages/manager/src/features/Databases/DatabaseLanding/DatabaseLanding.test.tsx
```diff
@@ -81,32 +82,25 @@ describe('Database Table', () => {
...
+    const { getByText } = renderWithTheme(<DatabaseLanding />);
 
-    // Loading state should render
-    expect(getByTestId(loadingTestId)).toBeInTheDocument();
-
-    await waitForElementToBeRemoved(getByTestId(loadingTestId));
+    // wait for API data to load
+    await waitFor(() => expect(getByText(database.label)).toBeVisible(), {
+      timeout: 10_000,
+    });
```
bnussman-akamai, line 94: The is the only sketchy change, everything else when pretty smoothly

hkhalil-akamai: Have you tried `await`ing a `findByText`?

packages/manager/src/features/Linodes/CloneLanding/Disks.test.tsx
```diff
@@ -33,7 +33,7 @@ describe('Disks', () => {
     const { getByTestId } = render(wrapWithTheme(<Disks {...props} />));
     disks.forEach((eachDisk) => {
       const checkbox = getByTestId(`checkbox-${eachDisk.id}`).parentNode;
-      fireEvent.click(checkbox as any);
+      fireEvent.click(checkbox!);
```
hkhalil-akamai, line 36: Should we also rewrite this as a `userEvent`?

hkhalil-akamai (approved):
I think this is a positive change, particularly for the performance improvement to our CI. Nice work 👍

packages/manager/src/features/Databases/DatabaseLanding/DatabaseLanding.test.tsx
bnussman-akamai: I did try that but the default timeout wasn't long enough 😖 I might be able to revert to `findByText` but with extra timeout

packages/manager/src/features/Linodes/CloneLanding/Disks.test.tsx
bnussman-akamai: Ideally yes, but I want to keep this PR's scope as narrow as possible so it is representative of what was absolutely necessary to change to get `happy-dom` working

There are so many places where `fireEvent` is used when `userEvent` would be preferable  😣

packages/manager/src/features/Linodes/LinodesDetail/LinodeNetworking/LinodeIPAddressRow.test.tsx
```diff
@@ -40,6 +42,11 @@ describe('LinodeIPAddressRow', () => {
       )
     );
 
+    // open the action menu
+    await userEvent.click(
+      getByLabelText('Action menu for IP Address [object Object]')
```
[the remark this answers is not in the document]
bnussman-akamai: Created M3-8743 🎟️

PR #11104, opened 2024-10-15
fix: `AppSelect.test.tsx` test flake

bnussman-akamai:
## Description 📝

- Attempts to fix the `AppSelect.test.tsx` [test flake we've been seeing](https://github.com/linode/manager/actions/runs/11351685589/job/31573480036?pr=11103) in our code coverage jobs 🔧 
- This flake started to show up when https://github.com/linode/manager/pull/11085 was merged. I don't know exactly why this test got slower. It's annoying, but an increase in timeout should be enough to get is back into a healthy state. 
  - We could consider globally increasing async unit test timeout but we should probably discuss the pros and cons of that before we make the change

## How to test 🧪

- Verify test passes locally
- Verify test passes in Github Actions

```
yarn test AppSelect
```

- [ ] 👀 Doing a self review

### DIFF
--- a/packages/manager/.changeset/pr-11104-tests-1729020207783.md
+++ b/packages/manager/.changeset/pr-11104-tests-1729020207783.md
@@ -1,0 +1,5 @@
+---
+"@linode/manager": Tests
+---
+
+Fix `AppSelect.test.tsx` test flake ([#11104](https://github.com/linode/manager/pull/11104))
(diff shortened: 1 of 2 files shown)


dwiley-akamai (approved):
Passes locally and in CI ✅

mjac0bs (approved):
Thanks for fixing this. Passes in CI and passed locally repeatedly. ✅ 

> We could consider globally increasing async unit test timeout but we should probably discuss the pros and cons of that before we make the change

This makes it 4 tests total that required increases in timeouts so far with happy-dom, right? DatabaseLanding, DatabaseCreate, VolumeCreate, and AppSelect. I'd rather we not increase the global timeout and make our tests potentially run slower when we just made them run faster... 😄 ...but if we continue to find flake due to timeout, maybe. Wonder if it would better to make a spike to look into why these specific tests are slower, though I don't know how deep you went down that rabbit hole.

bnussman-akamai:
> I'd rather we not increase the global timeout and make our tests potentially run slower when we just made them run faster

I agree. I worry that a global increase will open up the door for a slower test suite in general

